PR #3664, opened 2017-04-28
Add Name as tooltip when tab completing process ID

powercode:

### DIFF
--- a/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
+++ b/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
@@ -3201,12 +3201,14 @@ namespace System.Management.Automation
                 var pattern = WildcardPattern.Get(wordToComplete, WildcardOptions.IgnoreCase);
                 foreach (dynamic process in psObjects)
                 {
-                    var completionText = process.Id.ToString();
-                    if (pattern.IsMatch(completionText))
+                    var processId = process.Id.ToString();
+                    if (pattern.IsMatch(processId))
                     {
-                        var listItemText = completionText;
-                        completionText = quote + completionText + quote;
-                        result.Add(new CompletionResult(completionText, listItemText, CompletionResultType.ParameterValue, listItemText));
+                        var processName = process.Name;
+
+                        var idAndName = $"{processId} - {processName}";
+                        processId = quote + processId + quote;
+                        result.Add(new CompletionResult(processId, idAndName, CompletionResultType.ParameterValue, idAndName));
                     }
                 }
 
--- a/test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
+++ b/test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
@@ -90,6 +90,13 @@ Describe "TabCompletion" -Tags CI {
         $res.CompletionMatches.Foreach{$_.CompletionText -in 'A', 'B', 'C' | Should Be $true}
     }
 
+    It 'Should complete "Get-Process -Id " with Id and name in tooltip' {
+        $cmd = 'Get-Process -Id '
+        $res = TabExpansion2 -inputScript $cmd  -cursorColumn $cmd.Length
+        $res.CompletionMatches[0].CompletionText -match '^\d+$' | Should be true
+        $res.CompletionMatches[0].ToolTip -match '^\w' | Should be true
+    }
+
     It 'Should complete keyword' -skip {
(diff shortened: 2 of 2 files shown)


adityapatwardhan (dismissed):
Please add test similar to: https://github.com/PowerShell/PowerShell/blob/master/test/powershell/Language/Parser/ExtensibleCompletion.Tests.ps1#L344

powercode:
The failure seems unrelated to my checkin

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -3205,8 +3205,10 @@ private static void NativeCompletionProcessCommands(string wordToComplete, strin
                     if (pattern.IsMatch(completionText))
                     {
                         var listItemText = completionText;
```
lzybkr: I've always thought the user experience wasn't quite right with this completion.

I wonder if it would be better to use `listItemText` instead of `tooltip` to show the process name **and** the process id.

powercode: I guess you are right, since it is harder to see the tooltip

powercode: There is some weird case here there I have a null/empty process name on macOS. Is that expected or a bug somewhere in the process code?

powercode:
New version with listtext and tooltip "{id} - {name}"

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -1,97 +1,104 @@
...
         $res.CompletionMatches.Count | Should Be 3
         $res.CompletionMatches.Foreach{$_.CompletionText -in 'A', 'B', 'C' | Should Be $true}
-    }     
+    }
+
+    It 'Should complete "Get-Process -Id " with Id and name in tooltip' {
+        $cmd = 'Get-Process -Id '
+        $res = TabExpansion2 -inputScript $cmd  -cursorColumn $cmd.Length
+        $res.CompletionMatches[0].CompletionText -match '^\d+$' | Should be true
+        $res.CompletionMatches[0].ToolTip -match '^\w' | Should be true
```
daxian-dbw: > \w stands for "word character", usually [A-Za-z0-9_]

So `ToolTip -match '^\w'` doesn't guarantee it's in our expected form "<digits - string>".

rkeithhill:
Keep in mind that the PowerShell extension for VSCode uses the TooltipText for the detail text in IntelliSense.  ListItemCompletion is used for sorting and the Label.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
powercode: We don't know that we get the name string. Some tests on mac fail with null (or maybe empty) string. But maybe we could have '^\d+ - '

daxian-dbw: `'^\d+ - '` looks good :)

daxian-dbw:
@powercode can you please update the test with `'^\d+ - '`? Otherwise, the changes look good to me.
@rkeithhill do you have any concerns on the changes?

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -1,97 +1,104 @@
...
+        $res = TabExpansion2 -inputScript 'New-Object -TypeName X -Property @{ ' -cursorColumn 'New-Object -TypeName X -Property @{ '.Length
         $res.CompletionMatches.Count | Should Be 3
         $res.CompletionMatches.Foreach{$_.CompletionText -in 'A', 'B', 'C' | Should Be $true}
-    }     
+    }
+
+    It 'Should complete "Get-Process -Id " with Id and name in tooltip' {
+        $cmd = 'Get-Process -Id '
+        $res = TabExpansion2 -inputScript $cmd  -cursorColumn $cmd.Length
+        $res.CompletionMatches[0].CompletionText -match '^\d+ - ' | Should be true
```
daxian-dbw: You changed the wrong one. It's the `ToolTip` test below that you should change the pattern to `^\d+ - '`

powercode: Shouldn't do this in the middle of the night :) Fixing

powercode:
pushed a rebased version

daxian-dbw (approved):

PR #3678, opened 2017-05-01
Make sure nuget package version suffix applies to all ProjectReference

daxian-dbw:
Due to a breaking change (likely by design) in dotnet-cli, `dotnet pack --version-suffix` doesn't apply the same suffix to all its referenced project anymore. A workaround is used.
See https://github.com/NuGet/Home/issues/4337 and https://github.com/NuGet/Home/issues/3953 for more information.

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -1723,6 +1723,15 @@ function Publish-NuGetFeed
     # Add .NET CLI tools to PATH
     Find-Dotnet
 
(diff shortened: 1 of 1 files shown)


adityapatwardhan (approved):

adityapatwardhan:
@daxian-dbw LGTM

PR #3679, opened 2017-05-02
Disable schema validation in ScriptWriter.cs

daxian-dbw:
Add back the `#if CORECLR` that was removed when moving powershell to `netcoreapp2.0`.
The schema validation fails in powershell core because:
1. The XML Schema file `cmdlets-over-objects.xsd` is missing in Github
2. The resource string `CmdletizationCoreResources.Xml_cmdletsOverObjectsXsd` likely needs to be reworked to make the schema validation work.

The failure causes powershell core to fail to load CDXML modules that worked before moving to `netcoreapp2.0`.

Note that powershell core doesn't validate XML schemas due to the missing of related APIs in .NET Core 1.1 and prior versions. The `#if CORECLR` here was removed when moving to `netcore2.0` without verifying if the schema validation functionality actually works, so it should be reverted.

### DIFF
--- a/src/System.Management.Automation/cimSupport/cmdletization/ScriptWriter.cs
+++ b/src/System.Management.Automation/cimSupport/cmdletization/ScriptWriter.cs
@@ -47,6 +47,10 @@ namespace Microsoft.PowerShell.Cmdletization
             ScriptWriter.s_xmlReaderSettings.MaxCharactersFromEntities = 16384; // generous guess for the upper bound
             ScriptWriter.s_xmlReaderSettings.MaxCharactersInDocument = 128 * 1024 * 1024; // generous guess for the upper bound
 
+#if CORECLR // The XML Schema file 'cmdlets-over-objects.xsd' is missing in Github, and it's likely the resource string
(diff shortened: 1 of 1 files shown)


iSazonov:
@daxian-dbw  Do we need to open a tracking Issue?

daxian-dbw:
#3680 is filed to track enabling the CDXML schema validation.

lzybkr (approved):

iSazonov:
@daxian-dbw Can we exorcise the diable from the title? 😄

daxian-dbw:
@iSazonov Ah, my bad ... Commit message plays a part when calculating a commit hash, so changing it means changing the commit hash and thus I have to `push --force-with-lease`. But people might already pull that commit and start to work on top of it, and they would need to rebase their work if I `push --force-with-lease`.
Given that, I would rather not mess with the commit history in the master branch.

iSazonov:
@daxian-dbw I meant only the PR header.

daxian-dbw:
@iSazonov got it. The PR title is fixed.

PR #3685, opened 2017-05-02
Test change: have native execution tests use df instead of stty

JamesWTruher:
when automating test execution, some environments don't have a tty which causes
the tests to fail. Also, there is a behavioral difference between Linux and OSX.
 df is just as good a test for native execution as stty because it is also present.


SteveL-MSFT (approved):
LGTM

Issue #1903, opened 2016-08-18
Release .deb packages for Debian Stable


joakimlemb:
You have .deb packages for Ubuntu 14.x and 16.x, but not Debian.

In my experience Debian is a more popular for Servers than Ubuntu, so I'm hoping you could also supply .deb packages for Debian or maybe as a "workaround" make sure one of the Ubuntu .deb packages are compatible with the newest Debian Stable.

tupcakes:
I was able to get the 16.x unbuntu release working on debian stretch without much issue. I just did the following:
sudo apt-get install libunwind8
sudo dpkg -i Downloads/powershell_6.0.0-alpha.9-1ubuntu1.16.04.1_amd64.deb

cobracmder:
If you are running Debian Jessie, use the ubuntu 14.04 deb.

voltagex:
https://wiki.debian.org/DontBreakDebian#Don.27t_make_a_FrankenDebian

See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=834756 for the bug to track in Debian,

zaxebo1:
Please make the .deb packages named as:
dotnet-corefx-....deb   , dotnet-coreclr-....deb and dotnet-cli-......deb and powershell-.....deb 

so that the blocking bug https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=779970  for acceptance of powershell  https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=834756 can be removed

voltagex:
@zaxebo1 Unfortunately it isn't that simple. Debian doesn't just take external packages. While the current packages are fine for personal use, they don't meet Debian's standards, nor do they rebuild from source correctly (they require an internet connection on build, which is a no-no for Debian)

zaxebo1:
**i hope that it does get solved, and powershell comes to debian official repository as soon as possible. i think regarding https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=779970 it is the microsoft's powershell team which has to take action now, not the debian side**

Actually  i use ubuntu and debian; and i only use official repository(of ubutu/debian) packages for global installation
For other packages, like powershell and others, which are not in official repo of debian/ubuntu, i only install them for my own personal usage using AppImage or flatpak and **that too in a separate user's home directory without any global privileges.**

**But, without any package being in official repo (of debian/ubuntu), we can not use the software in our official projects for the organization, as per the organization's IT policy**

[...]

PR #3686, opened 2017-05-02
Rename Ubuntu14 package to include Debian8

anmenaga:
This is related to #1903 (Release .deb packages for Debian 8 "jessie").

With this change, package name for Ubuntu14 package will be changed:
from:
powershell_6.0.0-alpha.18-1ubuntu1.14.04.1_amd64.deb
to:
powershell_6.0.0-alpha.18-1ubuntu1.14.04.1-debian8_amd64.deb

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -1623,7 +1623,7 @@ esac
     # iteration is "debian_revision"
     # usage of this to differentiate distributions is allowed by non-standard
     if ($IsUbuntu14) {
-        $Iteration += "ubuntu1.14.04.1"
+        $Iteration += "ubuntu1.14.04.1-debian8"
     } elseif ($IsUbuntu16) {
         $Iteration += "ubuntu1.16.04.1"
     }


SteveL-MSFT:
The change LGTM, but I want @joeyaiello to ack on the filename

joeyaiello:
Per the FrankenDebian post and [this](https://www.debian.org/doc/manuals/debian-faq/ch-pkg_basics.en.html), I think we should have a separate package that follows the guidelines of 7.3

SteveL-MSFT:
@joeyaiello per @anmenaga the current Ubuntu14.04 pkg conforms unless you can point out something specific.  assuming it is kosher, it doesn't make sense to me to have the exact same deb with different names

joeyaiello:
If there's no change to the package, there's no value in renaming the filename. In fact, unless it's 100% following the Debian guidelines, it's potentially inaccurate.

If it is in fact 100% compliant, I think the only thing that would really provide value to Debian users would be to add a renamed package to the Debian 8 repo on packages.microsoft.com. 

For more context, the priority here is to make sure that we're compliant with Debians guidelines to the letter of the law that would allow it to be pulled into an official Debian repo in the future. 

If that's not possible today, I'd I stead recommend updating the install instructions to point Debian 8 at the Ubuntu package with a reference to the FrankenDebian guidance 

Does that sound reasonable?

SteveL-MSFT:
I read through the Debian guidelines and thought `<foo>_<VersionNumber>-<DebianRevisionNumber>_<DebianArchitecture>.deb` meant that we should include the supported Debian version for `<DebianRevisionNumber>`, but looking at the official Debian package repository for Deb8, it doesn't appear to be the case.  They are just named `<foo>_<VersionNumber>_<Arch>.deb`.  So I think we are fine with just the current name and can close this PR.

Issue #1618, opened 2016-08-03
System.Management.Automation.SemanticVersion should support more string constructors


JamesWTruher:
## Steps to reproduce

```
if ( $psversiontable.psversion -gt "5.0" ) {
  "yay"
}
```
## Expected behavior

```
yay
```
## Actual behavior

```
Could not compare "6.0.0-alpha" to "5.0". Error: "Cannot convert value "5.0" to type "System.Management.Automation.SemanticVersion". Error: "Cannot process argument
because the value of argument "version" is not valid. Change the value of the "version" argument and run the operation again.""
At line:1 char:6
+ if ( $psversiontable.psversion -gt "5.0" ) {
+      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : ComparisonFailure

```
## Environment data

```
PS /tmp> $psversiontable

Name                           Value
----                           -----
PSVersion                      6.0.0-alpha
PSEdition                      PowerShellCore
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   3.0.0.0
GitCommitId                    v6.0.0-alpha.7-193-g8f60d488cb4942243d697249c6948d28c56587c7-dirty
CLRVersion
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1

```

markcowl:
Note that this currently blocks installing binary modules that require PowerShell core from being installed from the Gallery.  This is a must for Azure PowerShell modules

edevil:
Has there been any progress on this?

SteveL-MSFT:
Somehow we missed this during triage for beta1.  I have a PR submitted.

lzybkr:
Closed via #3696

PR #3696, opened 2017-05-04
Allow [SemanticVersion] to be constructed with just major or major+minor

SteveL-MSFT:
[SemanticVersion]::new() required major, minor, and patch to be specfied which broke scenarios like:

`$psversiontable.psversion -gt "3.0"` which is used by PowerShellGet to determine if a module is compatible with the current version of PowerShell.

Change is to allow specifying only major or major+minor where the missing segments default to zero.

Addresses https://github.com/PowerShell/PowerShell/issues/1618

### DIFF
--- a/src/System.Management.Automation/engine/PSVersionInfo.cs
+++ b/src/System.Management.Automation/engine/PSVersionInfo.cs
@@ -425,6 +425,25 @@ namespace System.Management.Automation
             Label = null;
         }
 
+        /// <summary>
+        /// Construct a SemanticVersion.
+        /// </summary>
+        /// <param name="major">The major version</param>
+        /// <param name="minor">The minor version</param>
+        /// <exception cref="PSArgumentException">
+        /// If <paramref name="major"/> or <paramref name="minor"/> is less than 0.
+        /// </exception>
+        public SemanticVersion(int major, int minor) : this(major, minor, 0) {}
+
+        /// <summary>
+        /// Construct a SemanticVersion.
+        /// </summary>
+        /// <param name="major">The major version</param>
+        /// <exception cref="PSArgumentException">
+        /// If <paramref name="major"/> is less than 0.
+        /// </exception>
+        public SemanticVersion(int major) : this(major, 0, 0) {}
+
         private const string LabelPropertyName = "PSSemanticVersionLabel";
 
         /// <summary>
@@ -544,24 +563,24 @@ namespace System.Management.Automation
 
             var versionSansLabel = (dashIndex < 0) ? version : version.Substring(0, dashIndex);
             string[] parsedComponents = versionSansLabel.Split(Utils.Separators.Dot);
-            if (parsedComponents.Length != 3)
+            if (parsedComponents.Length > 3)
             {
                 result.SetFailure(ParseFailureKind.ArgumentException);
                 return false;
             }
 
-            int major, minor, patch;
+            int major = 0, minor = 0, patch = 0;
             if (!TryParseComponent(parsedComponents[0], "major", ref result, out major))
             {
                 return false;
             }
 
-            if (!TryParseComponent(parsedComponents[1], "minor", ref result, out minor))
+            if (parsedComponents.Length >= 2 && !TryParseComponent(parsedComponents[1], "minor", ref result, out minor))
             {
                 return false;
             }
 
-            if (!TryParseComponent(parsedComponents[2], "patch", ref result, out patch))
+            if (parsedComponents.Length == 3 && !TryParseComponent(parsedComponents[2], "patch", ref result, out patch))
             {
                 return false;
             }
--- a/test/powershell/engine/SemanticVersion.Tests.ps1
+++ b/test/powershell/engine/SemanticVersion.Tests.ps1
@@ -17,6 +17,20 @@ Describe "SemanticVersion api tests" -Tags 'CI' {
             $v.Patch | Should Be 0
             $v.Label | Should BeNullOrEmpty
             $v.ToString() | Should Be "1.0.0"
+
+            $v = [SemanticVersion]::new("3.0")
+            $v.Major | Should Be 3
+            $v.Minor | Should Be 0
+            $v.Patch | Should Be 0
+            $v.Label | Should BeNullOrEmpty
+            $v.ToString() | Should Be "3.0.0"
+
+            $v = [SemanticVersion]::new("2")
+            $v.Major | Should Be 2
+            $v.Minor | Should Be 0
+            $v.Patch | Should Be 0
+            $v.Label | Should BeNullOrEmpty
+            $v.ToString() | Should Be "2.0.0"
         }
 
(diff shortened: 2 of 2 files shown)


src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -544,24 +544,24 @@ private static bool TryParseVersion(string version, ref VersionResult result)
 
             var versionSansLabel = (dashIndex < 0) ? version : version.Substring(0, dashIndex);
             string[] parsedComponents = versionSansLabel.Split(Utils.Separators.Dot);
-            if (parsedComponents.Length != 3)
+            if (parsedComponents.Length == 0 || parsedComponents.Length > 3)
```
JamesWTruher: >arsedComponents.Length == 0  [](start = 17, length = 28)

i don't know if this condition can exist. Even `[string]::Empty.Split('.').Length will return 1

SteveL-MSFT: I believe that's the ETS .Length returning 1 where we treat scalars as arrays and add both .Length and .Count properties.  If you do [string]::Empty.Split('.')[0].Length, you get 0 as expected

test/powershell/engine/SemanticVersion.Tests.ps1
```diff
@@ -178,10 +199,6 @@ Describe "SemanticVersion api tests" -Tags 'CI' {
             # Revision isn't supported
             { [SemanticVersion]::new([Version]::new(0, 0, 0, 4)) } | Should Throw # PSArgumentException
             { [SemanticVersion]::new([Version]::new("1.2.3.4")) } | Should Throw # PSArgumentException
-
-            # Build is required
```
JamesWTruher: is there a reason why `[SemanticVersion]::new()` is not supported? It's supported for `System.Version`

SteveL-MSFT: I don't know if that was an explicit decision or accidentally left out.  I don't see any reason why it shouldn't be supported as it seems 0.0.0 is a valid semver.  I'll add support since it's a small change even if not related to this change :)

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -413,7 +413,7 @@ public SemanticVersion(int major, int minor, int patch, string label)
         /// <exception cref="PSArgumentException">
         /// If <paramref name="major"/>, <paramref name="minor"/>, or <paramref name="patch"/> is less than 0.
         /// </exception>
-        public SemanticVersion(int major, int minor, int patch)
+        public SemanticVersion(int major, int minor = 0, int patch = 0)
```
lzybkr: We should not use default arguments in public apis. The guidance is to provide multiple overloads for languages that may not support default arguments.

src/System.Management.Automation/engine/PSVersionInfo.cs
lzybkr: This is C# - PowerShell ETS is not relevant, and you are inspecting the length of the first element, not the number of elements in the array. [msdn](https://msdn.microsoft.com/en-us/library/b873y76a(v=vs.110).aspx) does say:
```
If this instance does not contain any of the characters in separator, the returned array consists of a single element that contains this instance.
```

test/powershell/engine/SemanticVersion.Tests.ps1
lzybkr: Under what scenario would we want a default constructor? If you want `0.0.0`, be explicit.

lzybkr (changes requested):
Note that the original implementation matched the specification, specifically: http://semver.org/#spec-item-2

We may want to discuss how best to diverge from the specification in a way that the community finds acceptable.

src/System.Management.Automation/engine/PSVersionInfo.cs
SteveL-MSFT: Will change

src/System.Management.Automation/engine/PSVersionInfo.cs
SteveL-MSFT: So it's a one element array that contains an empty string.  In any case, there is already code that causes an empty string to fail before the split() statement.

test/powershell/engine/SemanticVersion.Tests.ps1
SteveL-MSFT: For now, let's error on the side of simplicity and we can revisit this if there's customer demand for it

SteveL-MSFT:
Regarding the semver.org spec, we are not diverging.  The resulting semver is conformant to semver.

SteveL-MSFT:
@lzybkr I believe I've addressed all of your concerns

lzybkr:
I disagree - the bnf for parsing a semantic version is pretty clear - 3 parts: https://github.com/mojombo/semver/blob/master/semver.md

I found another example where we aren't matching the spec:
```powershell
PS> [System.Management.Automation.SemanticVersion]"01.01.01"

Major Minor Patch Label
----- ----- ----- -----
    1     1     1
```

If we're more permissive than the spec, people will rely on it, and where those uses occur, we can expect issues with more conforming tools.

SteveL-MSFT:
Leading zeros aren't allowed so the resulting semver conforms.  I guess the question is whether the resulting object is what matters for conformance or the input to the constructor.  Seems you prefer a more strict approach and reject 01.0.0 vs producing a valid semver.  This also means that you can't compare a semver like psversion against "3.0" since the latter is not a semver.

lzybkr:
Comparing with System.Version is no problem, but comparing with "3.0" might be, based on my reading of the spec.

It might be worth opening an issue in the repo I linked to.

SteveL-MSFT:
I've opened https://github.com/mojombo/semver/issues/368

SteveL-MSFT:
The response I got indicates it is reasonable and not explicitly defined by semver spec.  Also, this thread https://github.com/mojombo/semver/issues/344 has supporters for both sides with no official stance, but examples of other libraries that have taken both approaches (strict and loose).  I think it's reasonable to allow constructing a semver with defaults to zero.  Cc @PowerShell/powershell-committee

src/System.Management.Automation/engine/PSVersionInfo.cs
lzybkr: @JamesWTruher pointed out that the length can't be 0 and that part of the test should be removed. I agreed. Are saying that it is possible?

lzybkr (commented):
OK - changes look fine except for what looks like an unnecessary test.

src/System.Management.Automation/engine/PSVersionInfo.cs
SteveL-MSFT: Now I see what you are referring to.  I'll remove that part.

SteveL-MSFT:
@lzybkr all fixed

lzybkr (approved):

PR #3698, opened 2017-05-04
Update powershell to use .NET Core 2.0 preview 2 packages

daxian-dbw:
.net core NuGet packages released preview2 packages starting from 4/22. This change makes powershell core target preview2 packages.

The following two dotnet issues are addressed with the latest dotnet-cli tool:
https://github.com/dotnet/corefx/issues/17797
https://github.com/dotnet/cli/issues/6352

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,7 +2,7 @@ version: 6.0.0-alpha.18-{build}
 
 image: Visual Studio 2015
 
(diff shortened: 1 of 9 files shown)


daxian-dbw:
Close this PR as we should move to `netcoreapp` 2.0.0-preview1-002106-00, which is the version dotnet team will use for //Build.

PR #3714, opened 2017-05-06
Fix a pipe char in the table formatting

diev:
Set the proper escaping for a pipe char inside of a cell text in Markdown.


SteveL-MSFT (approved):
LGTM

PR #3719, opened 2017-05-07
Convertions from PSMethod to Func/Action

### LINKED ISSUE #2864: Add conversion from PSMethodInfo to delegate/Func<>
As a step on the way to realize #2226, I think we should start with adding a conversion from PSMethodInfo to System.Func.

Today, we have to do something like this:
```powershell
Class MyClass {
      
    [int] Add47([int] $value) {    
      return 47 + $value
    }
            
    [int] DoStuff([System.Collections.Generic.IEnumerable[int]] $collection) {
        # ugly
        $func = $this.GetType().GetMethod("Add47").CreateDelegate([Func[int,int]], $this)
        return [Linq.Enumerable]::Sum($collection, $func)
    }
}
```

I would like to express that as:
```powershell
Class MyClass {    
    [int] Add47([int] $value) {    
      return 47 + $value
    }
            
    [int] DoStuff([System.Collections.Generic.IEnumerable[int]] $collection) {
        return [Linq.Enumerable]::Sum($collection, $this.Add47)
    }
}
```

What is the right place to do this?

powercode:
Fixes #2864
Underpinnings to make calling of Extension methods /Linq
easier from PowerShell.

Enables the following that previously had to be done via
reflection.

class M {
  [int] Twice([int] $value) { return 2 * $value }

  [int] DoubleSum([int[]] $values) {
     return [Linq.Enumerable]::Sum($values, [M]::Twice)
  }

}


src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -4742,9 +4794,75 @@ internal static object ThrowInvalidConversionException(object valueToConvert, Ty
                 return CacheConversion<object>(fromType, toType, LanguagePrimitives.ConvertIntegerToEnum, ConversionRank.Language);
             }
 
+            if (fromType == typeof(PSMethod) && toType.IsSubclassOf(typeof(MulticastDelegate)))
+            {
+                var psMethod = (PSMethod) fromInstance;
```
lzybkr: As mentioned in long comment - I believe for this approach to work, you would **always** return something inside this `if`. Please do write a test to confirm what I've said, something like:

```PowerShell
class A {
    static [void] ggg([int] $i) {}
}
class B {
    static [void] fff([string] $i) {}
}

& { [Action[int]] [B]::fff } | Should Throw
& { [Action[int]] [A]::ggg } | Should Not Throw
```

test/powershell/Language/Parser/Conversions.Tests.ps1
```diff
@@ -127,3 +127,57 @@ namespace TestTypeResolution {
         $location | Should Be $cmdletDllPath
     }
 }
+
+Describe 'method conversion' -Tags 'CI' {
```
lzybkr: I think you should add tests for varying number of parameters (maybe up to 10?) and also void returns, and probably find some delegates other than Func and Action.

You also need to test overloads - make sure each overload is correctly chosen.

And I would also test multiple classes - making sure we don't call the wrong method, perhaps incorrectly cached from another conversion.

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -4742,9 +4794,75 @@ internal static object ThrowInvalidConversionException(object valueToConvert, Ty
...
+
+                var maybeType = psMethod.instance as Type;
+                var methodInfoCandiates = maybeType != null
+                    ? maybeType.GetMethods(BindingFlags.Static|BindingFlags.Public)
+                    : psMethod.instance.GetType().GetMethods(BindingFlags.Instance | BindingFlags.Public);
+
+                var isAction = toType.FullName.StartsWith("System.Action`");
```
lzybkr: What about delegates other than `Action` that return void?

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -4742,9 +4794,75 @@ internal static object ThrowInvalidConversionException(object valueToConvert, Ty
...
+                var comparator = new FuncActionArgsComparator(isAction, toType.GenericTypeArguments);
```
lzybkr: What about ordinary delegates with no generic type arguments?

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -3215,6 +3215,55 @@ private static PSConverter<bool> CreateNumericToBoolConverter(Type fromType)
...
+                foreach (var candidate in methodInfoCandiates)
+                {
+                    if (candidate.Name != psMethod.Name)
+                    {
+                        continue;
+                    }
+                    if (comparator.SignatureMatches(candidate.ReturnType, candidate.GetParameters()))
+                    {
+                        return maybeType != null
+                            ? candidate.CreateDelegate(resultType)
```
lzybkr: I would expect we create the delegate just once - when figuring the conversion - not every time we're asked to convert (e.g. in a loop).

lzybkr (changes requested):

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -4116,9 +4167,9 @@ private static ConversionData GetConversionData(Type fromType, Type toType)
             }
         }
 
-        internal static ConversionRank GetConversionRank(Type fromType, Type toType)
+        internal static ConversionRank GetConversionRank(Type fromType, Type toType, object fromInstance = null)
```
powercode: can you clarify:

> Instead of returning instances of PSMethod, we return instances of PSMethod

powercode:
Is the approach I took valid at all? Or is there a better way to approach this?

The problem I ran into was that `PSMethod` is a generic container for all signatures, and it seems like a mismatch to conversion of types in general.

src/System.Management.Automation/engine/LanguagePrimitives.cs
lzybkr: Yes, sorry for the confusion.

lzybkr:
Your approach is mostly valid but not great performance wise.

You can't add the instance parameter to FigureConversion as that makes it too easy to introduce bugs.

src/System.Management.Automation/engine/LanguagePrimitives.cs
oising: There's also just plain old non-generic `System.Action`.

oising:
Is this dead?

oising:
With respect to overload resolution and binding to any delegate type for any given `PSMethod`, I prototyped something a while ago that seems to cover a majority of the use cases. Perhaps you might be able to take something out of it to help with some of @lzybkr 's review requests, @powercode ?

oising:
```
PS> [string]::format | Get-Delegate
WARNING: Could not find matching signature for Format with 0 parameter(s).

PS> [string]::format | Get-Delegate -ParameterType string,object

Method                                             Target
------                                             ------
System.String Format(System.String, System.Object)

PS> ([string]::format | Get-Delegate -ParameterType string,object).invoke("hello, {0}", "world")
hello, world

PS> class Foo { [void] OnFoo([object]$sender, [eventargs]$eargs) {} }
PS> $f = new-object foo
PS> $f.OnFoo

OverloadDefinitions
-------------------
void OnFoo(System.Object sender, System.EventArgs eargs)

PS> $f.onfoo | Get-Delegate -DelegateType System.EventHandler

Method                                      Target
------                                      ------
Void OnFoo(System.Object, System.EventArgs) Foo

PS> ($f.onfoo | Get-Delegate -DelegateType System.EventHandler).invoke($null,$null)
```
The script is part of the Poke module on psgallery, and you can see it here: https://github.com/oising/poke/blob/master/delegate.ps1

Key tricks I learned was (ab)using `Type.GetMethodImpl` to do the heavy lifting of overload selection.

powercode:
Not dead, it is only resting :)

Pressure at work will ease up in a few months, and then I'll have more time to work at this.

rkeithhill:
Did you do a rebase in the wrong direction?

powercode:
Have to check. Rebased upstream/master. Not sure what happened here.

markekraus:
@powercode I have no clue what happened there, but I believe you can fix it with this (assuming your upstream for this repo is named `PowerShell`):

```bash
git fetch
git checkout -b 20171020fix PowerShell/master
git cherry-pick 2fb0dee149929d6f94c4e7a7e4906118dbeedef
```
There are some conflicts here, you will need to resolve those first and then continue

```bash
git cherry-pick --continue
git cherry-pick 87ba09b624d77804bdff5b4265f28e52e23253d2
git checkout -b 20171020backup psmethod-func
git checkout psmethod-func
git reset --hard 20171020fix
git push --force
```

I could only find the 2 commits in the history. It possible I missed some. This will at least backup your current branch to `20171020backup` so you can always go back if needed.

TravisEz13:
@powercode Something is wrong with your branch.  @markekraus  has been very good at giving instructions to fix this in the past.

powercode:
@markekraus Thx for the support! 
I think I will close this pull request, and be able to work on it and push changes without triggering review. I have some feedback from @lzybkr about the approach to take that I want to incorporate, and then I will create a new pull request. 

Any objections?

TravisEz13:
@powercode No objections.  FYI, I've seen times if you rebase, you cannot re-open the review.

powercode:
Then I just open a new PR

Issue #3349, opened 2017-03-16
CentOS/Red Hat (RHEL) 7.2/7.3 Support


joeyaiello:
It looks like we've already published our CentOS packages (which hopefully shouldn't need to be different than RHEL packages) to our [RHEL 7](http://packages.microsoft.com/rhel/7/) and [RHEL 7.3](http://packages.microsoft.com/rhel/7.3/) repositories. Going forward, we should:

- [ ] Test these packages on RHEL 7.2 and 7.3 on same cadence to ensure that are working properly 
- [ ] Remove "centos" from the package name to indicate that the RPM is universal across CentOS and RHEL ([as .NET Core has done](http://destek.serbestinternet.com/rhel-repo/rhel-7-server-dotnet-rpms/Packages/))

SteveL-MSFT:
@joeyaiello currently, we name it `powershell-6.0.0_alpha.17-1.el7.centos.x86_64.rpm`, should we change it to `powershell-6.0.0_alpha.18-redhat.x86_64.rpm`?

SteveL-MSFT:
@joeyaiello also, RHEL appears to be paid only (presumably for support), not sure how we should test RHEL regularly

joeyaiello:
@SteveL-MSFT no the recommendation here would be to remove the `centos`/`redhat` qualifier ([`el7` should already be sufficient here](https://unixy.net/secure/knowledgebase/104/What-does-el5-el6-and-el7-mean.html)). 

Also, I did some more research and, while the source has to be available (it's GPL), Red Hat does some stuff with trademarks that prevent redistribution of RHEL binaries. However, the difference between RHEL and CentOS is even less than I previously understood, so I'm now proposing that we just test again CentOS 7.2 and/or 7.3 in addition to 7.1. 

Let's talk offline on which we should do and how we prioritize them.

JamesWTruher:
Travis-CI doesn't support running build/test on CentOS directly. There are some forum threads on running tests on CentOS via docker, but that seems to be quite old (2015).

SteveL-MSFT:
I suspect we'll need VMs for getting regular coverage of distros

kfiresmith:
+1 for as much testing as possible on Centos 7 as we are a prod RHEL 7 shop that will probably end up using Powershell for AzureRM functions on RHEL.

joeyaiello:
[...]

PR #3732, opened 2017-05-08
Fix for issue 3349 - Package name should be el7, not el7.centos

JamesWTruher:
#3349 reference Centos is over specific. It should be simply el7 which is more inclusive

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -1634,7 +1634,7 @@ esac
     # OpenSUSE 42.1 (13.2 might build but is EOL)
     # Also SEE: https://fedoraproject.org/wiki/Packaging:DistTag
     if ($IsCentOS) {
-        $rpm_dist = "el7.centos"
+        $rpm_dist = "el7"
     } elseif ($IsFedora) {
         $version_id = $LinuxInfo.VERSION_ID
         $rpm_dist = "fedora.$version_id"


PaulHigin (approved):
LGTM

daxian-dbw:
@joeyaiello and @SteveL-MSFT please review to see if the name change is good.

joeyaiello (approved):

SteveL-MSFT (approved):

SteveL-MSFT:
We won't delay beta.1 for this change.

build.psm1
```diff
@@ -1634,7 +1634,7 @@ esac
     # OpenSUSE 42.1 (13.2 might build but is EOL)
     # Also SEE: https://fedoraproject.org/wiki/Packaging:DistTag
     if ($IsCentOS) {
-        $rpm_dist = "el7.centos"
+        $rpm_dist = "el7"
```
SteveL-MSFT: I did a quick search and it appears other scripts need to be updated to reflect the name change

SteveL-MSFT (changes requested):

JamesWTruher:
I think I have them all now

daxian-dbw:
The changes to markdown files and dockerfiles will break our users, because we currently don't have the package named `powershell-6.0.0_beta.1-1.el7.x86_64.rpm`.
The changes to files other than `build.psm1` would be reverted, and those changes should be done in the next release.

SteveL-MSFT:
@daxian-dbw so your recommendation is that we stage these changes: change build.psm1 so that our next release will have the correct file name, then update all the readme/scripts to use this new filename after the release, correct?

JamesWTruher:
ok - pulled first commit, running through CI now

build.psm1
SteveL-MSFT: Based on feedback about breaking existing users, we'll need to state the change so related scripts will be changed as part of https://github.com/PowerShell/PowerShell/issues/3888

SteveL-MSFT (approved):
LGTM

PR #3735, opened 2017-05-08
Enable Get-TimeZone for *nix and Mac OS.

### LINKED ISSUE #3605: Port Get-TimeZone cmdlet
.Net Core support [TimeZoneInfo Class](https://apisof.net/catalog/System.TimeZoneInfo) and we can port Get-TimeZone cmdlet on Unix.

adityapatwardhan:
- Get-TimeZone can be enabled now as the required classes are available in .Net Standard 2.0.
- Removed workaround for TimeZoneNotFoundException
- Test fixes
- Fixes #3605

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/TimeZoneCommands.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/TimeZoneCommands.cs
@@ -1,5 +1,3 @@
-#if !UNIX
-
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
@@ -66,15 +64,7 @@ namespace Microsoft.PowerShell.Commands
                     {
                         WriteObject(TimeZoneInfo.FindSystemTimeZoneById(tzid));
                     }
-#if CORECLR
-                    // TimeZoneNotFoundException is thrown by TimeZoneInfo, but not
-                    // publicly visible (so can't be caught), so for now we're catching
-                    // the parent exception time. This should be removed once the more
-                    // specific exception is available.
-                    catch (Exception e)
-#else
                     catch (TimeZoneNotFoundException e)
-#endif
                     {
                         WriteError(new ErrorRecord(e, TimeZoneHelper.TimeZoneNotFoundError,
                             ErrorCategory.InvalidArgument, "Id"));
@@ -103,14 +93,8 @@ namespace Microsoft.PowerShell.Commands
                         {
                             string message = string.Format(CultureInfo.InvariantCulture,
                                 TimeZoneResources.TimeZoneNameNotFound, tzname);
-#if CORECLR
-                        // Because .NET Core does not currently expose the TimeZoneNotFoundException
-                        // we need to throw the more generic parent exception class for the time being.
-                        // This should be removed once the correct exception class is available.
-                        Exception e = new Exception(message);
-#else
+
                             Exception e = new TimeZoneNotFoundException(message);
-#endif
                             WriteError(new ErrorRecord(e, TimeZoneHelper.TimeZoneNotFoundError,
                                 ErrorCategory.InvalidArgument, "Name"));
                         }
@@ -125,6 +109,7 @@ namespace Microsoft.PowerShell.Commands
         }
     }
 
+#if !UNIX
 
     /// <summary>
     /// A cmdlet to set the system's local time zone.
@@ -804,7 +789,7 @@ namespace Microsoft.PowerShell.Commands
 #endregion Win32 interop helper
     }
 
-
+#endif
     /// <summary>
     /// static Helper class for working with system time zones.
     /// </summary>
@@ -862,5 +847,3 @@ namespace Microsoft.PowerShell.Commands
         }
     }
 }
-
-#endif
--- a/src/Modules/Unix/Microsoft.PowerShell.Management/Microsoft.PowerShell.Management.psd1
+++ b/src/Modules/Unix/Microsoft.PowerShell.Management/Microsoft.PowerShell.Management.psd1
@@ -48,5 +48,6 @@ CmdletsToExport=@("Add-Content",
     "Rename-ItemProperty",
     "Resolve-Path",
     "Set-Content",
-    "Set-ItemProperty")
+    "Set-ItemProperty",
+    "Get-TimeZone")
 }
--- a/test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
@@ -14,11 +14,6 @@
            [snippet] Both StandardName and DaylightName are localized according to the current user default UI language.
 #>
 
-try {
-
-    $defaultParamValues = $PSdefaultParameterValues.Clone()
-    $PSDefaultParameterValues["it:skip"] = !$IsWindows
-
 function Assert-ListsSame
 {
     param([object[]] $expected, [object[]] $observed )
@@ -31,15 +26,24 @@ function Assert-ListsSame
     }
 }
 
-Describe "Get-Timezone test case no switches" -Tags "CI" {
+Describe "Get-Timezone test cases" -Tags "CI" {
+
+    BeforeAll {
+        $TimeZonesAvailable = [System.TimeZoneInfo]::GetSystemTimeZones()
+
+        $defaultParamValues = $PSdefaultParameterValues.Clone()
+        $PSDefaultParameterValues["it:skip"] = ($TimeZonesAvailable.Count -eq 0)
+    }
+
+    AfterAll {
+        $global:PSDefaultParameterValues = $defaultParamValues
+    }
+
     It "Call without ListAvailable switch returns current TimeZoneInfo" {
         $observed = (Get-TimeZone).Id
         $expected = ([System.TimeZoneInfo]::Local).Id
-        $observed -eq $expected | Should Be $true
+        $observed | Should Be $expected
     }
-}
-
-Describe "Get-Timezone test cases" -Tags "CI" {
 
     It "Call without ListAvailable switch returns an object of type TimeZoneInfo" {
         $result = Get-TimeZone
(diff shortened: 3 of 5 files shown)


test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
```diff
@@ -57,49 +56,53 @@ Describe "Get-Timezone test cases" -Tags "CI" {
     It "Call with ListAvailable switch returns a list containing TimeZoneInfo.Local" {
         $observedIdList = Get-TimeZone -ListAvailable | Select-Object -ExpandProperty Id
         $oneExpectedId = ([System.TimeZoneInfo]::Local).Id
-        $observedIdList -contains $oneExpectedId | Should Be $true
+        $observedIdList -eq $oneExpectedId | Should Be $oneExpectedId
```
iSazonov: It seems ` $observedIdList -contains $oneExpectedId | Should Be $true` was good.

test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
```diff
@@ -57,49 +56,53 @@ Describe "Get-Timezone test cases" -Tags "CI" {
...
     }
 
     It "Call with ListAvailable switch returns a list containing one returned by Get-TimeZone" {
         $observedIdList = Get-TimeZone -ListAvailable | Select-Object -ExpandProperty Id
         $oneExpectedId = (Get-TimeZone).Id
-        $observedIdList -contains $oneExpectedId | Should Be $true
+        $observedIdList -eq $oneExpectedId | Should Be $oneExpectedId
```
iSazonov: It seems `$observedIdList -contains $oneExpectedId | Should Be $true` was good.

test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
```diff
@@ -57,49 +56,53 @@ Describe "Get-Timezone test cases" -Tags "CI" {
...
     }
 
     It "Call Get-TimeZone using ID param and single item" {
-        (Get-TimeZone -Id "Cape Verde Standard Time").Id -eq "Cape Verde Standard Time" | Should Be $true
+        $selectedTZ = $TimeZonesAvailable | Select-Object -Index ((Get-Random) % $TimeZonesAvailable.Count)
+        (Get-TimeZone -Id $selectedTZ.Id).Id | Should Be $selectedTZ.Id
```
iSazonov: @It seems `Get-Random` isn't good for tests. Maybe cycle for all $TimeZonesAvailable is better and then we can remove next test.

test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
```diff
@@ -57,49 +56,53 @@ Describe "Get-Timezone test cases" -Tags "CI" {
...
     }
 
     It "Call Get-TimeZone using ID param and multiple items" {
-        $idList = @("Cape Verde Standard Time","Morocco Standard Time","Azores Standard Time")
-        $result = (Get-TimeZone -Id $idList).Id
-        Assert-ListsSame $result $idList
+        $randomIndexes = ((Get-Random) % $TimeZonesAvailable.Count), ((Get-Random) % $TimeZonesAvailable.Count),((Get-Random) % $TimeZonesAvailable.Count)
```
iSazonov: The same about `Get-Random`.

test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
```diff
@@ -57,49 +56,53 @@ Describe "Get-Timezone test cases" -Tags "CI" {
...
+        $selectedTZ = ($TimeZonesAvailable | Select-Object -Index $randomIndexes).Id
+        $result = (Get-TimeZone -Id $selectedTZ).Id
+        Assert-ListsSame $result $selectedTZ
     }
 
     It "Call Get-TimeZone using ID param and multiple items, where first and third are invalid ids - expect error" {
-        $null = Get-TimeZone -Id @("Cape Verde Standard","Morocco Standard Time","Azores Standard") `
+        $selectedTZ = ($TimeZonesAvailable | Select-Object -Index ((Get-Random) % $TimeZonesAvailable.Count)).Id
```
iSazonov: The same about `Get-Random`.

test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
```diff
@@ -57,49 +56,53 @@ Describe "Get-Timezone test cases" -Tags "CI" {
...
+        $errVar.Count | Should Be 2
         $errVar[0].FullyQualifiedErrorID | Should Be "TimeZoneNotFound,Microsoft.PowerShell.Commands.GetTimeZoneCommand"
     }
 
     It "Call Get-TimeZone using ID param and multiple items, one is wild card but error action ignore works as expected" {
-        $result = Get-TimeZone -Id @("Cape Verde Standard Time","Morocco Standard Time","*","Azores Standard Time") `
-                               -ErrorAction SilentlyContinue | ForEach-Object Id
-        $expectedIdList = @("Cape Verde Standard Time","Morocco Standard Time","Azores Standard Time")
-        Assert-ListsSame $expectedIdList $result
+        $randomIndexes = ((Get-Random) % $TimeZonesAvailable.Count), ((Get-Random) % $TimeZonesAvailable.Count),((Get-Random) % $TimeZonesAvailable.Count)
```
iSazonov: The same about `Get-Random`.

src/Microsoft.PowerShell.Commands.Management/commands/management/TimeZoneCommands.cs
```diff
@@ -863,4 +848,4 @@ internal static TimeZoneInfo[] LookupSystemTimeZoneInfoByName(string name)
     }
 }
 
-#endif
```
JamesWTruher: additional empty lines?

adityapatwardhan: Removed extra lines at end of file.

test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
adityapatwardhan: On a typical Ubuntu 16.04 machine there are 424 time zones. Checking for all time zones does not provide additional value to the test, but increases test execution time. 

Though Get-Random picks a different time zone when the test is run, Pester will show what the expected output was, and what the actual output is. So, debugging or reproducing the failure should not be a problem. Your thoughts?

The next test uses multiple time zone IDs for the -Id parameter, hence it is a different test.

adityapatwardhan: @iSazonov let me know what you think?

test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
```diff
@@ -35,71 +30,83 @@ Describe "Get-Timezone test case no switches" -Tags "CI" {
...
-        (Get-TimeZone -Id "Cape Verde Standard Time").Id -eq "Cape Verde Standard Time" | Should Be $true
+        $selectedTZ = $TimeZonesAvailable | Select-Object -Index ((Get-Random) % $TimeZonesAvailable.Count)
+        (Get-TimeZone -Id $selectedTZ.Id).Id | Should Be $selectedTZ.Id
     }
 
     It "Call Get-TimeZone using ID param and multiple items" {
-        $idList = @("Cape Verde Standard Time","Morocco Standard Time","Azores Standard Time")
-        $result = (Get-TimeZone -Id $idList).Id
-        Assert-ListsSame $result $idList
+        $randomIndexes = ((Get-Random) % $TimeZonesAvailable.Count), ((Get-Random) % $TimeZonesAvailable.Count),((Get-Random) % $TimeZonesAvailable.Count)
```
JamesWTruher: >Get-Random [](start = 27, length = 10)

also, if you're going to use get-random, it might be more efficient to do:
`get-random -input (1..$TimeZonesAvailable.Count) -count 3`

test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
```diff
@@ -35,71 +30,83 @@ Describe "Get-Timezone test case no switches" -Tags "CI" {
...
-        $oneExpectedId = (Get-TimeZone).Id
-        $observedIdList -contains $oneExpectedId | Should Be $true
+        $observedIdList = Get-TimeZone -ListAvailable | Select-Object -ExpandProperty BaseUtcOffset
+        $oneExpectedOffset = (Get-TimeZone).BaseUtcOffset
+        $observedIdList -eq $oneExpectedOffset | Should Be $oneExpectedOffset
     }
 
     It "Call Get-TimeZone using ID param and single item" {
-        (Get-TimeZone -Id "Cape Verde Standard Time").Id -eq "Cape Verde Standard Time" | Should Be $true
+        $selectedTZ = $TimeZonesAvailable | Select-Object -Index ((Get-Random) % $TimeZonesAvailable.Count)
```
JamesWTruher: >$TimeZonesAvailable | Select-Object -Index ((Get-Random) % $TimeZonesAvailable.Count) [](start = 22, length = 85)

`$TimeZonesAvailable[(Get-Random -max ($TimeZonesAvailable.Count +1))]` avoids a pipeline

test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
iSazonov: I think the tests should be deterministic and I would prefer don't use `Get-Random` because this reproducibly in manual locally _but not in CI restart_.
Perhaps it makes no sense to test all zones but it is very simple and very fast - I believe we may not worry about it here. (Although we may be limited to some selected time zone names/ids.)

adityapatwardhan: @iSazonov I agree on tests being deterministic. Though it seems like a small increase in execution time (from 80ms to 780ms on my machine), it eventually adds up. Since it does not provide additional coverage, I do not think we should go through all of them. 

I cannot use names / IDs because there is no guarantee that the time zone is available on the machine. Time zone names/IDs are different of different operating systems.  Example:

```Powershell
 Get-TimeZone -Name Pacific* | Select-Object -ExpandProperty Id
```
It has different output in Windows v/s Ubuntu.

I cannot use Select-Object -First 3, as there is no guarantee that the system has 3 time zones.

What do you think?



A way to not use Get-Random, would be to use Select-Object -first 3.

iSazonov: Maybe create three separated "testcases" - for Windows, Linux and Mac?

JamesWTruher: I think a reasonable test would be that if you get multiple timezones, check the first *n* (no need for random sample) and the break out of the test. No need to check them all. 
FWIW, this is still not deterministic because it relies on what is installed which may change across systems, and may not even satisfy the CI restart scenario if the config changes

src/Modules/Unix/Microsoft.PowerShell.Management/Microsoft.PowerShell.Management.psd1
```diff
@@ -2,7 +2,7 @@
 GUID="EEFCB906-B326-4E99-9F54-8B4BB6EF3C6D"
 Author="Microsoft Corporation"
 CompanyName="Microsoft Corporation"
-Copyright="� Microsoft Corporation. All rights reserved."
+Copyright="� Microsoft Corporation. All rights reserved."
```
daxian-dbw: Please retain the original character. This unintentional change is usually done by VSCode.

adityapatwardhan: Fixed.

adityapatwardhan:
@iSazonov @JamesWTruher @daxian-dbw I have done the suggested changes.

test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
```diff
@@ -35,71 +30,88 @@ Describe "Get-Timezone test case no switches" -Tags "CI" {
     It "Call without ListAvailable switch returns current TimeZoneInfo" {
         $observed = (Get-TimeZone).Id
         $expected = ([System.TimeZoneInfo]::Local).Id
```
daxian-dbw: Would this test work if `[System.TimeZoneInfo]::GetSystemTimeZones().Count -eq 0`? If not, then it should also be skipped in that case, like what you do for `Describe "Get-Timezone test cases`.

test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
```diff
@@ -35,71 +30,88 @@ Describe "Get-Timezone test case no switches" -Tags "CI" {
...
     }
 }
 
 Describe "Get-Timezone test cases" -Tags "CI" {
 
+    BeforeAll {
+        $TimeZonesAvailable = [System.TimeZoneInfo]::GetSystemTimeZones()
+
+        $defaultParamValues = $PSdefaultParameterValues.Clone()
+        $PSDefaultParameterValues["it:skip"] = ($TimeZonesAvailable.Count -eq 0)
```
daxian-dbw: In what case would `$TimeZonesAvailable.Count -eq 0` be true?

src/Modules/Unix/Microsoft.PowerShell.Management/Microsoft.PowerShell.Management.psd1
```diff
@@ -2,7 +2,7 @@
 GUID="EEFCB906-B326-4E99-9F54-8B4BB6EF3C6D"
 Author="Microsoft Corporation"
 CompanyName="Microsoft Corporation"
-Copyright="� Microsoft Corporation. All rights reserved."
+Copyright="© Microsoft Corporation. All rights reserved."
```
daxian-dbw: When it looks the same but git shows a diff, it's usually due to a special character change. What I would do in this case is to first revert changes to this file (git checkout), and then open it in a different editor (not vscode), and make the `Get-TimeZone` change.
@iSazonov do you happen to know if there is any configuration file we can use to prevent VSCode from messing up special characters?

test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
adityapatwardhan: If the tzdata package is removed.

test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
adityapatwardhan: Fixed. Moved it to the same Describe.

test/powershell/Modules/Microsoft.PowerShell.Management/TimeZone.Tests.ps1
daxian-dbw: good to know. Thanks

daxian-dbw (approved):

Issue #2354, opened 2016-09-24
Can we improve the semantic version implementation for myget?


Jaykul:
> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version.

http://semver.org/
### Two things I'd propose improving.

First, current releases on MyGet are being labelled badly. For instance, SMA is "1.0.0-alpha10" or "1.0.0-alpha9" ... 
1. I'm pretty sure that should be 6.0.0?
2. I think the pre-release section should be -alpha.9 and -alpha.10 

Second, the `System.Management.Automation.SemanticVersion` built into PowerShell 6 can't sort pre-releases into the right order ... because it treats everything on the end as a _single_ text label and sorts them alphabetically (so, like Visual Studio, it thinks the "1.0.0-alpha9" is newer than "1.0.0-alpha10").

SteveL-MSFT:
I would suggest having the SMA.SemanticVersion be a separate issue

vors:
@daxian-dbw can you, please comment on the "." in pre-release part?
Together with alpha.11, we will remove alpha9 package from myget feed to fix the order.

daxian-dbw:
@Jaykul yes, we spotted this awkward issue in powershell-core myget, and I'm sorry for not thinking through when publishing the alpha9 packages.
As for the '.', nuget treats version tags like "1.0.0-alpha.05" as invalid, and package with that version tag cannot be opened by tools like "Nuget Package Explorer".

```
PS:80> cat .\test.nuspec -TotalCount 6
<?xml version="1.0"?>
<package >
    <metadata>
        <id>test</id>
        <version>1.0.0-alpha.05</version>
        <title>test</title>
PS:81> nuget pack .\test.nuspec
Attempting to build package from 'test.nuspec'.
'1.0.0-alpha.05' is not a valid version string.
Parameter name: value
```

Jaykul:
Ok, I just double-checked, and I see that even though your feed on MyGet has support for Semver 2, _after 2 years_, SemVer2 is still _deliberately blocked_ in `nuget pack` ...

So yeah, uhm ... zero padding, and a little bit of planning about how many "-alpha" releases are reasonable 😲 

The [current state](https://github.com/NuGet/Home/wiki/SemVer-2.0.0-support) --for anyone else.

emgarten:
[...]

PR #3759, opened 2017-05-11
Update releasing.md about NuGet package suffix

daxian-dbw:
Per the discussion in #2354, we now use a version like `6.0.0-beta.1` for our NuGet packages. This PR updates `releasing.md` to keep in sync.


adityapatwardhan (approved):
LGTM

PR #3778, opened 2017-05-12
Test HttpListener so that PowerShell Core tests don't have to rely on an internet web service

SteveL-MSFT:
Base for HttpListener module needed for my upcoming changes to web cmdlets to add Link Header traversal.  We should also be able to build upon this to remove dependency on some of the external web sites we use to test the web cmdlets and bring them back to CI instead of Feature.

### DIFF
--- a/test/tools/Modules/HttpListener/HttpListener.psd1
+++ b/test/tools/Modules/HttpListener/HttpListener.psd1
@@ -1,0 +1,10 @@
+@{
+ModuleVersion = '1.0.0'
+GUID = 'e148b26c-0594-4963-99e5-419d4ff302e2'
+Author = 'Steve Lee'
+CompanyName = 'Microsoft'
+Copyright = '(c) Microsoft. All rights reserved.'
+Description = 'Creates a new HTTP Listener for testing purposes'
+RootModule = 'HttpListener.psm1'
+FunctionsToExport = @('Start-HttpListener','Stop-HttpListener')
+}
--- a/test/tools/Modules/HttpListener/HttpListener.psd1
+++ b/test/tools/Modules/HttpListener/HttpListener.psd1
@@ -1,0 +1,28 @@
+@{
+
+# Version number of this module.
+ModuleVersion = '1.0.0'
+
+# ID used to uniquely identify this module
+GUID = 'e148b26c-0594-4963-99e5-419d4ff302e2'
+
+# Author of this module
+Author = 'Steve Lee'
+
+# Company or vendor of this module
+CompanyName = 'Microsoft'
+
+# Copyright statement for this module
+Copyright = '(c) Microsoft. All rights reserved.'
+
+# Description of the functionality provided by this module
+Description = 'Creates a new HTTP Listener for testing purposes'
+
+# Modules to import as nested modules of the module specified in RootModule/ModuleToProcess
+RootModule = 'HttpListener.psm1'
+
+# Functions to export from this module
+FunctionsToExport = @('Start-HttpListener','Stop-HttpListener')
+
+}
+
--- a/test/tools/Modules/HttpListener/HttpListener.psm1
+++ b/test/tools/Modules/HttpListener/HttpListener.psm1
@@ -1,0 +1,153 @@
+Function Stop-HTTPListener {
+    <#
+    .Synopsis
+        Stop HTTP Listener used for PowerShell tests rather than rely on 3rd party websites
+    .Description
+        Sends `exit` command to HTTP Listener causing it to exit.
+    .Parameter Port
+        Port to use, default is 8080
+    .Example
+        Stop-HTTPListener -Port 8080
+    #>
+    Param (
+        [Parameter()]
+        [Int] $Port = 8080
+    )
+
+    Invoke-WebRequest -Uri "http://localhost:$port/PowerShell?test=exit"
+}
+
+Function Start-HTTPListener {
+    <#
+    .Synopsis
+        Creates a new HTTP Listener to be used for PowerShell tests rather than rely on 3rd party websites
+    .Description
+        Creates a new HTTP Listener supporting several test cases intended to be used only by PowerShell tests.
+        This cmdlet requires running from an elevated administrator prompt to open a port.
+
+        Use Ctrl-C to stop the listener.  You'll need to send another web request to allow the listener to stop since
+        it will be blocked waiting for a request.
+    .Parameter Port
+        Port to listen, default is 8080
+    .Example
+        Start-HTTPListener -Port 8080
+        Invoke-WebRequest -Uri "http://localhost:8080/PowerShell/?test=linkheader&maxlinks=5"
+    #>
+    Param (
+        [Parameter()]
+        [Int] $Port = 8080,
+
+        [Parameter()]
+        [switch] $AsJob
+        )
+
+    Process {
+        $ErrorActionPreference = "Stop"
+
+        if ($IsWindows)
+        {
+            $CurrentPrincipal = New-Object Security.Principal.WindowsPrincipal( [Security.Principal.WindowsIdentity]::GetCurrent())
+            if ( -not ($currentPrincipal.IsInRole( [Security.Principal.WindowsBuiltInRole]::Administrator ))) {
+                Write-Error "This script must be executed from an elevated PowerShell session" -ErrorAction Stop
+            }
+        }
+
+        [scriptblock]$script = {
+            param ($Port)
+
+            # HttpListener.QueryString is not being populated, need to follow-up with CoreFx, workaround is to parse it ourselves
+            Function ParseQueryString([string]$url)
+            {
+                Write-Verbose "Parsing: $url"
+                $queryItems = @{}
+                foreach ($segment in $url.Split("&"))
+                {
+                    if ($segment -match "\??(?<name>\w*)(=(?<value>\w*))?$")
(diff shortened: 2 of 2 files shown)


test/tools/Modules/HttpListener/HttpListener.psd1
```diff
@@ -0,0 +1,28 @@
...
+Copyright = '(c) Microsoft. All rights reserved.'
+
+# Description of the functionality provided by this module
+Description = 'Creates a new HTTP Listener for testing purposes'
+
+# Modules to import as nested modules of the module specified in RootModule/ModuleToProcess
+RootModule = 'HttpListener.psm1'
+
+# Functions to export from this module
+FunctionsToExport = @('Start-HttpListener','Stop-HttpListener')
```
iSazonov: Please remove all comments and extra lines - they are useless.

test/tools/Modules/HttpListener/HttpListener.psm1
```diff
@@ -0,0 +1,154 @@
...
+        if ($AsJob)
+        {
+            Start-Job -ScriptBlock $script -ArgumentList $Port
+        }
+        else
+        {
+            & $script -Port $Port
+        }
+    }
+}
```
iSazonov: Please add Newline.

iSazonov (commented):
Thoughts:
1. It seems Windows firewall block the traffic by default. Maybe Unix too.

2. HTTPS?

3. Sometimes we need to use 80 and 443 standard port.

4. System.Net.Sockets.TcpListener don't require elevated privileges.

5. Really is there no lightweight portable web server?

SteveL-MSFT:
@iSazonov 

1. usage should be via local loopback so firewall shouldn't be issue (worked when tested on win10 and ubuntu16)

2. I'll add HTTPS when I remove usage of badssl.com tests

test/tools/Modules/HttpListener/HttpListener.psm1
SteveL-MSFT: fixed

test/tools/Modules/HttpListener/HttpListener.psd1
SteveL-MSFT: fixed

test/tools/Modules/HttpListener/HttpListener.psm1
SteveL-MSFT: Added newline, doesn't show up on GitHub though

SteveL-MSFT:
Looks like you no longer have to be elevated with corefx, so removed that check

iSazonov:
Could you add HTTPS now to avoid wasting time then?

SteveL-MSFT:
HTTPS will require more work as I'll need to figure out how to generate certs for each platform.  For my LinkHeader feature, I don't need HTTPS

iSazonov:
We can generate the client and server certificates in advance and put it in the repo in asserts folder.

SteveL-MSFT:
@iSazonov good idea, I'll work on that

iSazonov:
Sample https://github.com/PowerShell/PowerShell/blob/76de340e790171c16193f3a19ce33410e495c319/test/powershell/Modules/Microsoft.PowerShell.Security/CmsMessage.Tests.ps1

test/tools/Modules/HttpListener/HttpListener.psd1
```diff
@@ -0,0 +1,10 @@
+@{
+ModuleVersion = '1.0.0'
```
anmenaga: Module version supposed to be 4 numbers?

test/tools/Modules/HttpListener/HttpListener.psm1
```diff
@@ -0,0 +1,168 @@
...
+    Invoke-WebRequest -Uri "http://localhost:$port/PowerShell?test=exit"
+}
+
+Function Start-HTTPListener {
+    <#
+    .Synopsis
+        Creates a new HTTP Listener to be used for PowerShell tests rather than rely on 3rd party websites
+    .Description
+        Creates a new HTTP Listener supporting several test cases intended to be used only by PowerShell tests.
+        This cmdlet requires running from an elevated administrator prompt to open a port.
```
anmenaga: Looks like this line can be removed.

test/tools/Modules/HttpListener/HttpListener.psm1
```diff
@@ -0,0 +1,168 @@
...
+                $queryItems = @{}
+                foreach ($segment in $url.Split("&"))
+                {
+                    if ($segment -match "\??(?<name>\w*)(=(?<value>.*?))?$")
+                    {
+                        $name = $matches["name"]
+                        $value = $matches["value"]
+                        if ($value -ne $null)
+                        {
+                            $value = [System.Web.HttpUtility]::UrlDecode($value)
```
anmenaga: Maybe it is better to run entire $url through UrlDecode first, and then to splitting and parsing?

test/tools/Modules/HttpListener/HttpListener.psm1
```diff
@@ -0,0 +1,168 @@
...
+            }
+
+            $listener = [System.Net.HttpListener]::New()
+            $urlPrefix = "/PowerShell"
+            $url = "http://localhost:$Port$urlPrefix/"
+            $listener.Prefixes.Add($url)
+            $listener.AuthenticationSchemes = [System.Net.AuthenticationSchemes]::Anonymous 
+            try {
+                Write-Warning "Note that thread is blocked waiting for a request.  After using Ctrl-C to stop listening, you need to send a valid HTTP request to stop the listener cleanly."
+                Write-Warning "Sending 'exit' command will cause listener to stop"
```
anmenaga: Instead of this "Sending 'exit' command" instruction I would rather see something that could be copy/pasted like:
Stop-HTTPListener / Invoke-WebRequest -Uri "http://localhost:$port/PowerShell?test=exit"

anmenaga (approved):
In general, LGTM.

anmenaga:
@SteveL-MSFT As we are using this in test automation, we should ensure that both tests and this server side are following HTTP spec (specifically, response headers).

test/tools/Modules/HttpListener/HttpListener.psd1
TravisEz13: Doesn't have to be and 3 is SemVer compatible.

test/tools/Modules/HttpListener/HttpListener.psm1
TravisEz13: `elevated` is not capitalized correctly

test/tools/Modules/HttpListener/HttpListener.psm1
TravisEz13: Perhaps use the URI to get the QueryString then do the rest of the parsing.

test/tools/Modules/HttpListener/HttpListener.psm1
SteveL-MSFT: Good suggestion

test/tools/Modules/HttpListener/HttpListener.psm1
SteveL-MSFT: @anmenaga pointed out correctly this comment is no longer valid

test/tools/Modules/HttpListener/HttpListener.psm1
SteveL-MSFT: If I decode first, encoded `=` and `&` will result in incorrect results

SteveL-MSFT:
@anmenaga httplistener class itself should be returning appropriate response headers for basic payloads, it will be up to the tests to add specific response headers depending on the test

SteveL-MSFT:
@iSazonov looks like I have to use httpcfg.exe on Windows to configure use of SSL cert.  Let's defer HTTPS support for now.  I primarily need this in to test my upcoming linkheader traversal change to the webcmdlets as I need to be able to control exactly what is returned in the link header.

iSazonov:
Yes, I agree.
https://blogs.msdn.microsoft.com/jpsanders/2009/09/29/how-to-walkthrough-using-httplistener-or-http-server-unmanaged-code-c-as-an-ssl-simple-server/

SteveL-MSFT:
@iSazonov are you ok with the current changes?  if so, can you approve?

iSazonov (approved):

test/tools/Modules/HttpListener/HttpListener.psm1
TravisEz13: I still recommend using URI to get query

test/tools/Modules/HttpListener/HttpListener.psm1
JamesWTruher: from an automation perspective, I think this will be used more asynchronously, so I would rather see `AsJob` be the default. Perhaps `Blocking` should be the switch and default creates an asynchronous execution. Also, I'm also thinking that it might be better to run in a PowerShell instance rather than a job.

I think that the name of functions are too generic. If this is for our tests, then it should have that flavor, and be named `Start-TestHttpListener` or something along those lines

test/tools/Modules/HttpListener/HttpListener.psm1
SteveL-MSFT: I'll make that change

Issue #3769, opened 2017-05-12
Attribute may silently crash


iSazonov:
Steps to reproduce
------------------

```powershell
function Test-BadAttribute
{
    [CmdletBinding()]
    Param
    (
        # Param1 help description
        [Parameter]
        [ValidateSet("a","b")]
        $Param1,

        [Parameter]
        [ValidateSet()] # Bad attribute - the constructor is failed
        $Param2
    )

    Begin
    {
    }
    Process
    {
    }
    End
    {
    }
}

1. Test-BadAttribute -Param1 <Press Tab>
2. Test-BadAttribute -Param1 <Press Enter>
3. Test-BadAttribute -Param1 "a" <Press Enter>
```

Expected behavior
-----------------

1. IntelliSenese work for Param1 and don't  work for Param2 with error message
or IntelliSenese don't work for both parameters with appropriate error message.
2. Show appropriate error message.
3. Cmdlet work well or don't work with showing appropriate error message.

Actual behavior
---------------
1. No exception. No error message. IntelliSense don't work.
2. Murky error message: it is impossible to understand that there is a problem with _any_ attribute, which is the attribute and which is the parameter.
3. The same as 2. and the cmdlet don't work.

Environment data
----------------

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
BuildVersion                   3.0.0.0
CLRVersion
GitCommitId                    v6.0.0-beta.1-7-g7fa53aedb3c7514cf1c2a19af6c26566661596d9-dirty
OS                             Microsoft Windows 10.0.10240
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

alx9r:
Silent failure for the entire `param()` block also occurs when there is a typo in the attribute.  For example, the following outputs nothing:

```PowerShell
function f { param([NonExistentAttribute()]$x,$y) }
(Get-Command f).Parameters
```

iSazonov:
Possible place for the fix https://github.com/PowerShell/PowerShell/blob/e8bf6ca18a04ee04b290d2f01f5ef01750280d36/src/System.Management.Automation/engine/parser/Compiler.cs#L1702-L1705

vexx32:
[...]

PR #3784, opened 2017-05-15
ValidateSetAttribute enhancement: support set values to be dynamically generated from a custom ValidateSetValueGenerator

### LINKED ISSUE #3744: ValidateSet should support a dynamically generated set
People often use dynamic parameters so they can dynamically generate a `ValidateSet` attribute.

We should provide a simpler way to do this. One option is to accept a scriptblock:

```powershell
param(
    [ValidateSet( { Get-ChildItem -ah | ForEach-Object Name } )]
    $HiddenFileName
)
```

iSazonov:
Close #3744 

Currently `ValidateSetAttribute` accepts only explicit constants as valid values. This is a significant limitation. Sometimes we need to get valid values dynamically, ex., Azure VMs, logged-on users and so on. The PR add follow possibilities:
- pass a _custom type_ (a valid values generator) implementing `IValidateSetValuesGenerator` interface to get valid values dynamically.
- pass a _custom type_ (a valid values generator) derived  from `CachedValidValuesGeneratorBase` abstract class to get valid values dynamically and _cache_ the list to share with other ValidateSetAttribute attributes.

We support valid values generators on PowerShell and C#.


src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1455,6 +1456,43 @@ public ValidateSetAttribute(params string[] validValues)
...
+        /// <summary>
+        /// Initializes a new instance of the ValidateSetAttribute class
+        /// </summary>
+        /// <param name="type">list of valid values</param>
+        /// <exception cref="ArgumentNullException">for null arguments</exception>
+        /// <exception cref="ArgumentOutOfRangeException">for invalid arguments</exception>
+        public ValidateSetAttribute(Type type)
+        {
+            IValidateSetValuesGenerator validValuesGenerator;
+            if (typeof(IValidateSetValuesGenerator).IsAssignableFrom(type))
```
lzybkr: You can use the `as` operator.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1455,6 +1456,43 @@ public ValidateSetAttribute(params string[] validValues)
...
+            {
+                validValuesGenerator = (IValidateSetValuesGenerator)Activator.CreateInstance(type);
+            }
+            else
+            {
+                throw PSTraceSource.NewArgumentNullException("validValues");
```
lzybkr: Cut and paste? I would use `ArgumentException` and please use the correct parameter name.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1455,6 +1456,43 @@ public ValidateSetAttribute(params string[] validValues)
...
+            }
+
+            _validValues = validValuesGenerator.GetValidValues().ToArray();
```
lzybkr: We shouldn't cache the values during construction - the set may change over time.

Also - we may want a way to query other bound parameters from the generator - which would not be possible in the constructor.

test/powershell/Language/Classes/Scripting.Classes.Attributes.Tests.ps1
```diff
@@ -216,3 +218,96 @@ Describe 'Type resolution with attributes' -Tag "CI" {
...
+                    {
+                        var testValues = new string[] {"Test1","TestString","Test2"};
+                        foreach (var value in testValues)
+                        {
+                            yield return value;
+                        }
+                    }
+                }
+
+                /// Implement of test IValidateSetValuesGenerator to return Null
```
lzybkr: This implementation does not return `null` - it returns an empty list.

test/powershell/Language/Classes/Scripting.Classes.Attributes.Tests.ps1
```diff
@@ -216,3 +218,96 @@ Describe 'Type resolution with attributes' -Tag "CI" {
         }
     }
 }
+
+Describe 'ValidateSet support a dynamically generated set' -Tag "CI" {
+    Context 'C# test' {
```
lzybkr: You should also provide a test implementing the interface in PowerShell.

lzybkr (changes requested):

test/powershell/Language/Classes/Scripting.Classes.Attributes.Tests.ps1
iSazonov: Could you help me here?

lzybkr: I recommend learning more about how the `yield` keyword is implemented under the covers, e.g. this is a good article: http://csharpindepth.com/Articles/Chapter6/IteratorBlockImplementation.aspx

src/System.Management.Automation/engine/Attributes.cs
iSazonov: I eliminated the caching.

Please clarify - do you say to put `validValuesGenerator` in a property?

test/powershell/Language/Classes/Scripting.Classes.Attributes.Tests.ps1
iSazonov: Do you have in mind something like?
```powershell
            function Get-TestValidateSet3
            {
                [CmdletBinding()]
                Param
                (
                    [ValidateSet([type]::GetType("GenValuesForParam1"))]
                    $Param1
                )

                $Param1
            }
```

test/powershell/Language/Classes/Scripting.Classes.Attributes.Tests.ps1
iSazonov: Thanks for useful link!

Test was removed.

test/powershell/Language/Classes/Scripting.Classes.Attributes.Tests.ps1
lzybkr: No - I meant:
```powershell
class GenValuesFromParam2 : IValidateSetValuesGenerator { ... }
...
[ValidateSet([GenValuesFromParam2])] ...
```
In other words - no Add-Type.

iSazonov: Following sample don't work - the type passed as string to `public ValidateSetAttribute(params string[] validValues)` constructor - wrong binding 😕 :
```powershell
            class GenValuesForParam2 : System.Management.Automation.IValidateSetValuesGenerator {
                [System.Collections.Generic.IEnumerable[String]] GetValidValues() { return "Test1","TestString","Test2" }
            }

            function Get-TestValidateSet3
            {
                [CmdletBinding()]
                Param
                (
                    [ValidateSet([GenValuesForParam2])]
                    $Param1
                )

                $Param1
            }
```

iSazonov: Fixed.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1455,6 +1453,39 @@ public ValidateSetAttribute(params string[] validValues)
...
+            if (typeof(IValidateSetValuesGenerator).IsAssignableFrom(valuesGeneratorType))
+            {
+                validValuesGenerator = (IValidateSetValuesGenerator)Activator.CreateInstance(valuesGeneratorType);
+            }
+            else
+            {
+                throw PSTraceSource.NewArgumentException("valuesGeneratorType");
+            }
+
+            _validValues = validValuesGenerator.GetValidValues();
```
lzybkr: The constructor should just save `valuesGeneratorType` because we create just one instance of the attribute per use.

For example, if someone added `ValidateSet` to the `Name` parameter on `Get-VM` - you would want the name of each vm whenever you called `Get-VM`, not the names when you first called `Get-VM`.

By constructing the instance and getting the values during validation, the set is dynamically generated when tab completion or validation is happening - which only aligns with construction once (for scripts, attributes are constructed at first invocation) or never (for binary cmdlets, attributes are eagerly constructed).

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1359,7 +1360,7 @@ public ValidateCountAttribute(int minLength, int maxLength)
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
     public sealed class ValidateSetAttribute : ValidateEnumeratedArgumentsAttribute
     {
-        private string[] _validValues;
```
lzybkr: I think this shouldn't change, and I would add a new read-only property for the Type so you can tell the difference on which constructor was used.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1386,7 +1387,7 @@ public IList<string> ValidValues
         {
             get
             {
-                return _validValues;
+                return _validValues.ToList();
```
lzybkr: For the dynamic case - this should probably return a new collection every time it's called.

lzybkr (changes requested):
One other thought - some of the dynamic sets are expensive to compute, e.g. querying Azure or whatever.

One strategy is to cache the set for some time. Nothing in this design prevents such caching, but it would be better to implement the caching in PowerShell so that third party code can use it easily.  Maybe that means adding a `CacheExpiration` parameter that defaults to 0 - immediate expiration.

Still thinking out loud - the cache maybe belongs in the generator, but maybe we provide some helpers to manage it - perhaps as an abstract base class.

iSazonov:
>Still thinking out loud - the cache maybe belongs in the generator, but maybe we provide some helpers to manage it - perhaps as an abstract base class.

Wouldn't that be too intrusive? `IValidateSetValuesGenerator` looks very powerful.
If we want such cache in the generator it seems we should migrate from `IValidateSetValuesGenerator` interface to `ValidateSetValuesGenerator` class and force users to inherite from the class. Maybe we stop on `IValidateSetValuesGenerator`? Now we have 'generic' caching internally in `ValidateSet` so maybe we'll leave the users full freedom in `IValidateSetValuesGenerator`?

lzybkr:
I had feedback in the form of a [PR to TabExpansionPlusPlus](https://github.com/lzybkr/TabExpansionPlusPlus/commit/b1d99d34aac92b07fa76aae3c5aa5a426e75993c)  that caching should be built-in.

iSazonov:
Do you mean that the cache has to work between different `ValidateSet` instances? With `validValuesGeneratorType` key?

lzybkr:
Many cmdlets have parameters that are common, e.g. there are 72 commands in AzureRM.Compute that accept `-ResourceGroupName`.

On my machine, it takes >1s to query the resource group names

So I think a cache that works across `ValidateSet` instances would be very helpful.

powercode:
@lzybkr 
> I had feedback in the form of a PR to TabExpansionPlusPlus that caching should be built-in.
It feels almost more often than not that I implement some sort of custom completion caching. But often with less than stellar aging/expiration.

iSazonov:
I had to rebase to get autoloading of helper modules.

Global valid values cache was added. The only question is. I clean up this cache never. Should we add clean-up and where? Although it seems superfluous.

iSazonov:
@lzybkr Could you please continue the code review?

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1386,6 +1413,37 @@ public IList<string> ValidValues
...
+                        validValuesCache.TryAdd(validValuesGeneratorType, validValuesCacheEntry);
+                    }
+                    if (validValuesCacheEntry.validValues == null)
+                    {
+                        throw new ValidationMetadataException(
+                            "ValidateSetGeneratedValidValuesListIsEmpty",
+                            null,
+                            Metadata.ValidateSetGeneratedValidValuesListIsEmpty);
+                    }
+                    _validValues = validValuesCacheEntry.validValues;
```
lzybkr: Is there a good reason to store the values in the field instead of `return validValuesCacheEntry.validValues;`?

lzybkr: I think it would also be useful to clean up the cache - this could be done with something like `Task.Delay(CacheExpiration).ContinueWith(() => validValuesCacheEntry.validValues = null);`

This might not seem like a big deal, the collection should typically be small, but sometimes it might be large, so we should make the collection collectible if it's truly going to be used again.

lzybkr (changes requested):
This is starting to look pretty good.

I have a feeling that this will make many folks happy, but it will fall short for others, mostly around wanting access to some context, e.g. the cmdlet or some of the already bound parameters.

Hopefully we can get some feedback one way or the other on the usefulness of this as it is though.

src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -739,7 +739,6 @@ static Compiler()
             s_builtinAttributeGenerator.Add(typeof(ParameterAttribute), NewParameterAttribute);
             s_builtinAttributeGenerator.Add(typeof(OutputTypeAttribute), NewOutputTypeAttribute);
             s_builtinAttributeGenerator.Add(typeof(AliasAttribute), NewAliasAttribute);
-            s_builtinAttributeGenerator.Add(typeof(ValidateSetAttribute), NewValidateSetAttribute);
```
iSazonov: I temporary removed that to pass tests but I forgot to ask you about that.
Now fixed.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: >Is there a good reason to store the values in the field instead of  return validValuesCacheEntry.validValues; ?

We only use this in `SetAsString()` for error message. We could use `ValidValues` if we're not afraid that there will be a cache expiration and a recalculation of  `ValidValues` and that there may be other data in the error message.

>  Task.Delay(CacheExpiration)

Different attributes can have different `CacheExpiration`. How can we resolve the problem?

lfshr:
Is this definitely limited to 'binary' cmdlets? Has anyone tested this using a PS declared class?

iSazonov:
@lfshr Scripts is supported - see tests in the PR. We will edit the PR description after completing this work.

lfshr:
@iSazonov absolutely brilliant, v good work! Well done @lzybkr for the solution. (I take full credit for the moan that instigated it 😛)

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Could you please continue?
It seems we should clean up validValuesCache from expired Entries not validValuesCacheEntry.validValues.

lzybkr: Sorry, I thought I had replied.

I now think `CacheExpiration` should **not** be an argument to the attribute because, as you point out, different uses might specify conflicting values.  Instead, the generator should own the cache expiration policy, and the attribute should not cache the results from the generator.

I still like the idea of providing an abstract base class that implements a default caching policy - it would be optional for a generator to derive from it.

iSazonov: Can we add this to  IValidateSetValuesGenerator  interface?

And please comment my "It seems we should clean up validValuesCache from expired Entries (generators) ".

lzybkr: Sure, you could add `CacheExpiration` to `IValidateSetValuesGenerator`, but it would be something extra that some generators have no need for.

I agree we should clean up `validValuesCache` - and ideally, I think that cache is optionally available to generators but not required. This is why I suggested an abstract base class that generators could derive from.

schittli:
This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  thomas.schittli@akros.ch
    Domain jig.ch has exceeded the max emails per hour (209/200 (104%)) allowed.  Message will be reattempted later

------- This is a copy of the message, including all the headers. ------
------ The body of the message is 5430 characters long; only the first
------ 5000 or so are included here.
Received: from o4.sgmail.github.com ([192.254.112.99]:33553)
	by rs111.nsresponse.com with esmtps (TLSv1.2:ECDHE-RSA-AES128-GCM-SHA256:128)
	(Exim 4.89)
	(envelope-from <bounces+848413-aac0-github=jig.ch@sgmail.github.com>)
	id 1dJ3YW-000ecy-9S
	for github@jig.ch; Thu, 08 Jun 2017 14:57:22 -0500
DKIM-Signature: v=1; a=rsa-sha1; c=relaxed/relaxed; d=github.com; 
	h=from:reply-to:to:cc:in-reply-to:references:subject:mime-version:content-type:content-transfer-encoding:list-id:list-archive:list-post:list-unsubscribe; 
	s=s20150108; bh=hShJWcvT9Pb6uxla7Fnd9iixHHM=; b=Y3gswrT8hSHg3Sug
	3nESvwqFy3i0f7kr1bXJ62TWl0X//POn4qUMScz6eJ9NDbdycbPyASBXc4zv8B7T
	nHUY0pDTfrXlDmwYHgDei3Y9FrXvzTi1dsgppciKhx2QZS0Nq19xTbjzEnkFH9Vc
	WoLLO7uWPKx1QBTyMC8ZholprWs=
Received: by filter0986p1mdw1.sendgrid.net with SMTP id filter0986p1mdw1-9547-5939ABF5-38
        2017-06-08 19:56:37.867097011 +0000 UTC
Received: from github-smtp2a-ext-cp1-prd.iad.github.net (github-smtp2a-ext-cp1-prd.iad.github.net [192.30.253.16])
	by ismtpd0005p1iad1.sendgrid.net (SG) with ESMTP id exs5qSK6Sb6QZKnfIy_g2Q
	for <github@jig.ch>; Thu, 08 Jun 2017 19:56:37.840 +0000 (UTC)
Date: Thu, 08 Jun 2017 12:56:37 -0700

src/System.Management.Automation/engine/Attributes.cs
iSazonov: @lzybkr I still don't correct tests however please review my next step - Is that the right direction?

I removed valid values cache and  CacheExpiration  from ValidateSet attribute, enhance a generator cache (add cleanup - the idea is from [CoreFX](https://github.com/stephentoub/corefx/blob/9820567e1bb1972502bfe6345963eb0ab99ef3f9/src/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs)), add an abstract class for generators with valid values cache support.

iSazonov: @lzybkr Now tests adjusted.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1354,13 +1357,156 @@ public ValidateCountAttribute(int minLength, int maxLength)
...
+                throw new ValidationMetadataException(
+                    "ValidateSetGeneratedValidValuesListIsEmpty",
+                    null,
+                    Metadata.ValidateSetGeneratedValidValuesListIsEmpty);
+            }
+
+            if (ValidValuesCacheExpiration > 0)
+            {
+                Task.Delay(ValidValuesCacheExpiration).ContinueWith((task) => _validValues = null);
+                _validValues = _validValuesNoCache;
```
lzybkr: The parameter to `Task.Delay` is in milliseconds, not seconds.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1354,13 +1357,156 @@ public ValidateCountAttribute(int minLength, int maxLength)
...
+        /// Get a valid values.
+        /// </summary>
+        public IEnumerable<string> GetValidValues()
+        {
+            if (_validValues != null)
+            {
+                return _validValues;
+            }
+
+            var _validValuesNoCache = GenerateValidValues();
```
lzybkr: >_validValuesNoCache [](start = 16, length = 19)

Local variables should not use a `_` prefix.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1354,13 +1357,156 @@ public ValidateCountAttribute(int minLength, int maxLength)
...
+
+        /// <summary>
+        /// Get a valid values.
+        /// </summary>
+        public IEnumerable<string> GetValidValues()
+        {
+            if (_validValues != null)
+            {
+                return _validValues;
+            }
```
lzybkr: There is a race condition here.

The sequence to trigger the race is:

1. _validValues != null
2. The background task to release the cache runs, assigning _validValues with null.
3. The return executes, returning null.

The fix is:

```c#
var validValuesLocal = _validValues;
if (validValuesLocal != null)
{
    return validValuesLocal;
}
```

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1354,13 +1357,156 @@ public ValidateCountAttribute(int minLength, int maxLength)
...
     /// </summary>
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
     public sealed class ValidateSetAttribute : ValidateEnumeratedArgumentsAttribute
     {
+        // We can use either static '_validValues'
+        // or dynamic valid values list generated by instance of 'validValuesGeneratorType'.
         private string[] _validValues;
 
+        // The valid values generator cache works across 'ValidateSetAttribute' instances.
+        private static ConcurrentDictionary<Type, ValidValuesGeneratorCacheEntry> s_ValidValuesGeneratorCache = new ConcurrentDictionary<Type, ValidValuesGeneratorCacheEntry>();
```
lzybkr: >ValidValuesGeneratorCacheEntry [](start = 50, length = 30)

I think this can be `IValidateSetValuesGenerator`.
There will be very few types that implement this interface, so I think the extra complexity around cache cleanup on the generator is not needed.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1386,6 +1532,40 @@ public IList<string> ValidValues
...
+                        ValidValuesGeneratorCacheEntry.cacheEntryAccessTime = currentTime;
+                        s_ValidValuesGeneratorCache[validValuesGeneratorType] = ValidValuesGeneratorCacheEntry;
+                    }
+                    else
+                    {
+                        // Add a valid values generator to the cache.
+                        // We don't cache valid values.
+                        // We expect that valid values can be cached in the valid values generator.
+                        var validValuesGenerator = (IValidateSetValuesGenerator)Activator.CreateInstance(validValuesGeneratorType);
+                        var cacheEntryAccessTime = DateTime.Now;
```
lzybkr: I think I'd move creating the generator and storing it in the ConcurrentDictionary to the ValidateSet constructor.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1386,6 +1532,40 @@ public IList<string> ValidValues
...
+                        ValidValuesGeneratorCacheEntry = new ValidValuesGeneratorCacheEntry(validValuesGenerator, cacheEntryAccessTime);
+                        s_ValidValuesGeneratorCache.TryAdd(validValuesGeneratorType, ValidValuesGeneratorCacheEntry);
+                    }
+
+                    _validValues = ValidValuesGeneratorCacheEntry.validValuesGenerator.GetValidValues()?.ToArray();
+
```
lzybkr: As long as we're always call `ToArray`, maybe we should just change the interface to return `string[]`.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Thanks! I thought about it, but I didn't know how to just fix it.

Fixed.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Do we really want tochange _public_ interface? Is the breaking change?

lzybkr: Sorry for not being clear - I meant the interface introduced in this PR - `IValidateSetValuesGenerator` - so not a breaking change.

src/System.Management.Automation/engine/Attributes.cs
lzybkr: I'm a little confused. The is cache cleanup not controlled by the user - so restarting PowerShell is a practical necessity.

src/System.Management.Automation/engine/Attributes.cs
lzybkr: We could, but I think that seconds is a more practical unit for the timeout.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: I agree - the more simple for scripts.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: We can use `ValidValuesGeneratorCacheExpiration` to implicitly remove expired values from the cache.
Main question is - will we be happy with the PowerShell restart?

lzybkr: Restarting PowerShell is an occasional annoyance, but I'm not sure this will make a difference.

For C# implementations, you need to restart anyway.
For script implementations - I think there are acceptable ways to develop your generator and test it without restarting PowerShell.

So my preference is to keep the code as simple as possible and only introduce additional complexity after gaining more experience with the feature and determining it really is necessary.

iSazonov: I agree that we can wait a feedback from users.
The cache is removed.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Yes, now the cache is removed and we can create the generator in the constructor.
Fixed.

iSazonov:
@lzybkr Could you please continue the code review?

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1354,13 +1357,82 @@ public ValidateCountAttribute(int minLength, int maxLength)
     }
 
     /// <summary>
+    /// Base class for all valid values generators with cache support.
```
lzybkr: I'd change this comment to something like:

```
Optional base class for <see cref="IValidateSetValuesGenerator"/> implementations that want a default implementation to cache values.
```

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1354,13 +1357,82 @@ public ValidateCountAttribute(int minLength, int maxLength)
...
+            {
+                return _ValidValuesCacheExpiration / 1000;
+            }
+            set
+            {
+                _ValidValuesCacheExpiration = value * 1000;
+            }
+        }
+
+        private int _ValidValuesCacheExpiration;
```
lzybkr: `_v`, and add a blank line.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1354,13 +1357,82 @@ public ValidateCountAttribute(int minLength, int maxLength)
...
+    public abstract class CachedValidValuesGeneratorBase : IValidateSetValuesGenerator
+    {
+        // Cached valid values.
+        private string[] _validValues;
+
+        /// <summary>
+        /// Sets a time interval in seconds to reset the '_validValues' dynamic valid values cache.
+        /// By default 'ValidValuesCacheExpiration = 0' and no valid values is cached (is re-evaluated every time it is used).
+        /// </summary>
+        public int ValidValuesCacheExpiration {
```
lzybkr: I think this class should have a single constructor:
```C#
public CachedValidValuesGeneratorBase(int cacheExpirationInSeconds) { _validValuesCacheExpiration = cacheExpirationInSeconds; }
```

I don't think the property should be exposed - as changing the value implies something (e.g. that I can extend or shorten the cache expiration for values already cached.)

lzybkr: Maybe we do need a method to clear the cache though - I can see some external event happening and the generator wanting to drop the cache.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1455,6 +1542,39 @@ public ValidateSetAttribute(params string[] validValues)
...
+        /// <param name="valuesGeneratorType">class that implements the 'IValidateSetValuesGenerator' interface</param>
+        /// <exception cref="ArgumentException">for null arguments</exception>
+        public ValidateSetAttribute(Type valuesGeneratorType)
+        {
+            if (!typeof(IValidateSetValuesGenerator).IsAssignableFrom(valuesGeneratorType))
+            {
+                throw PSTraceSource.NewArgumentException("valuesGeneratorType");
+            }
+
+            validValuesGeneratorType = valuesGeneratorType;
```
lzybkr: Why store the type? If we stored the instance, we only look up in `s_ValidValuesGeneratorCache` once per use of the attribute instead of once per invocation of the command using the attribute.

src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -1298,15 +1298,35 @@ private static Attribute NewAliasAttribute(AttributeAst ast)
...
-                args[i] = _attrArgToStringConverter.Target(_attrArgToStringConverter,
-                    ast.PositionalArguments[i].Accept(cvv));
+                if (TypeResolver.TryResolveType(generatorTypeAst.TypeName.FullName, out Type generatorType))
+                {
+                    result = new ValidateSetAttribute(generatorType);
+                }
+                else
+                {
+                    throw InterpreterError.NewInterpreterException(ast, typeof(RuntimeException), ast.Extent,
+                        "CustomValidValuesGeneratorTypeNotFound", ParserStrings.CustomValidValuesGeneratorTypeNotFound, generatorTypeAst.TypeName.FullName, typeof(System.Management.Automation.IValidateSetValuesGenerator).FullName);
```
lzybkr: It's probably better to reuse `TypeNotFound` instead of adding a new error message.

src/System.Management.Automation/resources/Metadata.resx
```diff
@@ -168,6 +168,9 @@
   <data name="ValidateSetFailure" xml:space="preserve">
     <value>The argument "{0}" does not belong to the set "{1}" specified by the ValidateSet attribute. Supply an argument that is in the set and then try the command again.</value>
   </data>
+  <data name="ValidateSetGeneratedValidValuesListIsEmpty" xml:space="preserve">
```
lzybkr: The ErrorId doesn't match the message - is it `null` or empty?

src/System.Management.Automation/resources/ParserStrings.resx
```diff
@@ -759,6 +759,9 @@ Possible matches are</value>
   <data name="CustomAttributeTypeNotFound" xml:space="preserve">
     <value>Cannot find the type for custom attribute '{0}'. Make sure that the assembly that contains this type is loaded.</value>
   </data>
+  <data name="CustomValidValuesGeneratorTypeNotFound" xml:space="preserve">
```
lzybkr: I think I'd reuse `TypeNotFound` instead of adding a new message.

lzybkr (dismissed):
Great - this is looking pretty good now, just a few little things to clean up I think.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Done.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/parser/Compiler.cs
iSazonov: Done.

src/System.Management.Automation/resources/Metadata.resx
iSazonov: Null.

Fixed.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Sorry, I don't understand the question.
We use `validValuesGeneratorType` in `ValidValues` property.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: The property is not available for users from outside the class because we create an instance of the class by type with _default_ constructor only. We need way to set `_validValuesCacheExpiration`. See a sample from our tests below. So we cannot remove the property but we can make it as `protected`. 

```powershell
            class GenValuesWithExpiration : System.Management.Automation.CachedValidValuesGeneratorBase {
                GenValuesWithExpiration() : base() {
                    $this.ValidValuesCacheExpiration = 2
                }

                Static [bool] $temp = $true;

                [String[]] GenerateValidValues() {

                    if ([GenValuesWithExpiration]::temp) {
                        [GenValuesWithExpiration]::temp = $false
                        return [string[]]("Test1","TestString1","Test2")
                    } else {
                        [GenValuesWithExpiration]::temp = $true
                        return [string[]]("Test1","TestString2","Test2")
                    }
                    
                }
            }
```

lzybkr: I'm suggesting:

```C#
    class GenValuesWithExpiration : System.Management.Automation.CachedValidValuesGeneratorBase {
        GenValuesWithExpiration() : base(2) {}
        # Rest is the same
```

The PowerShell engine is calling the default constructor, but the abstract base doesn't need a default constructor.

src/System.Management.Automation/engine/Attributes.cs
lzybkr: You only need to store the type if you defer creating the instance. If you create and store the instance in the constructor, I think that's better overall.

src/System.Management.Automation/engine/parser/Compiler.cs
lzybkr: If we didn't have this special case code, you would have gotten `TypeNotFound`, which is exactly the problem - it has nothing to do with `IValidateSetValuesGenerator`. Besides, if you don't implement the interface, that error is reported when constructing the attribute.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Thanks for clarify! Is it good UX? For me, this is not as intuitive as using the property name explicitly.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Thanks for clarify! I see now.
Won't there be a problem if we add a cache cleanup?

lzybkr: I don't anticipate any problems - these objects are small assuming they leave caching to us.

src/System.Management.Automation/engine/Attributes.cs
lzybkr: UX is bad if you provide an option to set the property but semantics of setting are fuzzy. I think it's better to not expose the setter in this case.

src/System.Management.Automation/engine/parser/Compiler.cs
iSazonov: Thanks! 

Reuse  TypeNotFound done.

src/System.Management.Automation/resources/ParserStrings.resx
iSazonov: Removed.

src/System.Management.Automation/engine/parser/Compiler.cs
iSazonov: Sorry, VS Code removed trailing spaces.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Thanks!

The property is removed.

iSazonov: `ClearValidValuesGeneratorCache` was added.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Done.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1354,19 +1357,90 @@ public ValidateCountAttribute(int minLength, int maxLength)
...
         private string[] _validValues;
+        private IValidateSetValuesGenerator validValuesGenerator = null;
+
+        // The valid values generator cache works across 'ValidateSetAttribute' instances.
+        private static ConcurrentDictionary<Type, IValidateSetValuesGenerator> s_ValidValuesGeneratorCache = new ConcurrentDictionary<Type, IValidateSetValuesGenerator>();
+
+        /// <summary>
+        /// Clear the valid values generator cache.
+        /// </summary>
+        public static void ClearValidValuesGeneratorCache()
```
lzybkr: I'm not sure this is useful - because each instance of the attribute will also contain a reference to the `IValidateSetValuesGenerator` and we can't clear that.

When I suggested a method to clear the cache, I was referring to `CachedValidValuesGeneratorBase`, but I think it's probably fine to not add that - one can simply use a smaller timeout value or implement their own custom caching logic.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1354,19 +1357,90 @@ public ValidateCountAttribute(int minLength, int maxLength)
...
+            {
+                throw new ValidationMetadataException(
+                    "ValidateSetGeneratedValidValuesListIsNull",
+                    null,
+                    Metadata.ValidateSetGeneratedValidValuesListIsNull);
+            }
+
+            if (_validValuesCacheExpiration > 0)
+            {
+                Task.Delay(_validValuesCacheExpiration * 1000).ContinueWith((task) => _validValues = null);
```
lzybkr: There is an extremely unlikely race condition here - the task could complete before the assignment on the next line. I can't see it ever happening except maybe on a single core CPU where PowerShell gets suspended (say machine goes to sleep) at just the right time, and when it resumes, the background task gets to run first.

So I'd swap these two lines - I don't see any downsides to doing that.

lzybkr (commented):
Thanks for addressing all of my feedback, I think this is almost ready.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: I swaped the lines.

Please clarify. If you mean that the race condition is related with clearing `_validValues` - we add a protect above (validValuesLocal) and return validValuesNoCache below - so clearing `_validValues` should not cause problems.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Sorry, my attention is focused on the current virus attacks. 
I remember that we rejected the clear already.

Removed.

iSazonov:
> I have a feeling that this will make many folks happy, but it will fall short for others, mostly around wanting access to some context, e.g. the cmdlet or some of the already bound parameters.

@lzybkr Shall we continue with the "some context"?

iSazonov:
It seems we catch a race condition on slow OSX. 😕

src/System.Management.Automation/engine/Attributes.cs
lzybkr: This would be a different race - we wouldn't actually clear the cache.

lzybkr:
Wasn't there a CLR bug we were hitting on Mac related to defining classes? Maybe @daxian-dbw remembers.

lzybkr:
If you want to investigate options for providing more information to the generator - I could imagine wanting to know the command, maybe the command ast, the actual parameter the attribute is applied to, and maybe the already bound parameters.

iSazonov:
I tried to think about it. If we cache and share the generator, we can't store the context in it and should pass it as a parameter.  Even if we send `engineIntrinsics` from `Validate()` to `ValidateElement` we cannot send it in a generator - we get valid values with the `ValidValues` _property_, and we use the property in Intellisense completers. Although `engineIntrinsics` contains current cmdlet info it does not contain current binding parameter info.
That way requires too much effort. 😕 Is there another way?

lzybkr:
We can add an interface:

```c#
public interface IExtendedValidateArguments
{
    void Validate(object arguments, EngineIntrinsics engineIntrinsics, /* extra args*/);
}
```

Then:
```C#
    public abstract class ValidateArgumentsAttribute : CmdletMetadataAttribute
    {
        //...

        internal void InternalValidate(object o, EngineIntrinsics engineIntrinsics, /* extra args */)
        {
            if (var extended = this as IExtendedValidateArguments) {
                extended.Validate(o, engineIntrinsics, /* extra args */ );
            } else {
                Validate(o, engineIntrinsics);
            }
        }
```

And lastly implement that interface for ValidateSetAttribute.

Alternatively, we can move the type test to the parameter binder to avoid passing extra arguments if that is expensive somehow.

iSazonov:
Thanks! I still don't see how fix the `ValidValues`  property for completers. It seems we need a new method in the interface?

daxian-dbw (commented):
> It seems we catch a race condition on slow OSX.
Wasn't there a CLR bug we were hitting on Mac related to defining classes?

I'm not aware of any CLR bug that affects PowerShell class except for the crossgen one that causes powershell to hang. We don't crossgen for CI builds, so that shouldn't affect it here.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1354,19 +1357,82 @@ public ValidateCountAttribute(int minLength, int maxLength)
...
+    public abstract class CachedValidValuesGeneratorBase : IValidateSetValuesGenerator
+    {
+        // Cached valid values.
+        private string[] _validValues;
+
+        /// <summary>
+        /// Initializes a new instance of the CachedValidValuesGeneratorBase class.
+        /// </summary>
+        /// <param name="cacheExpirationInSeconds">Sets a time interval in seconds to reset the '_validValues' dynamic valid values cache.</param>
+        public CachedValidValuesGeneratorBase(int cacheExpirationInSeconds)
```
iSazonov: Fixed.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1354,19 +1357,82 @@ public ValidateCountAttribute(int minLength, int maxLength)
...
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
     public sealed class ValidateSetAttribute : ValidateEnumeratedArgumentsAttribute
     {
+        // We can use either static '_validValues'
+        // or dynamic valid values list generated by instance of 'validValuesGenerator'.
         private string[] _validValues;
+        private IValidateSetValuesGenerator validValuesGenerator = null;
+
+        // The valid values generator cache works across 'ValidateSetAttribute' instances.
+        private static ConcurrentDictionary<Type, IValidateSetValuesGenerator> s_ValidValuesGeneratorCache = new ConcurrentDictionary<Type, IValidateSetValuesGenerator>();
```
iSazonov: Above I tried to add cleanup but @lzybkr concluded that it didn't matter - we don't expect a huge number of generators.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1380,13 +1446,28 @@ public sealed class ValidateSetAttribute : ValidateEnumeratedArgumentsAttribute
...
         {
             get
             {
-                return _validValues;
+                if (validValuesGenerator == null)
+                {
+                    return _validValues;
+                }
+
+                var validValuesLocal = validValuesGenerator.GetValidValues();
```
iSazonov: It seems related #3769 - completers silently falls back to paths.
I tested a generator with `throw "Test throw!"` - parameter binder handles any exceptions:
```powershell
Get-TestValidateSetPS4 : Cannot validate argument on parameter 'Param1'. Test throw!
At line:1 char:32
+ Get-TestValidateSetPS4 -param1 fdfd
+                                ~~~~
    + CategoryInfo          : InvalidData: (:) [Get-TestValidateSetPS4], ParameterBindingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationError,Get-TestValidateSetPS4
```
If the generator throws good formatted exception users get it in InnerException.
So it seems redundant to add extra level to catch the exception.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1455,6 +1533,44 @@ public ValidateSetAttribute(params string[] validValues)
...
+
+        /// <summary>
+        /// Initializes a new instance of the ValidateSetAttribute class.
+        /// Valid values is returned dynamically from a custom class implementing 'IValidateSetValuesGenerator' interface.
+        /// </summary>
+        /// <param name="valuesGeneratorType">class that implements the 'IValidateSetValuesGenerator' interface</param>
+        /// <exception cref="ArgumentException">for null arguments</exception>
+        public ValidateSetAttribute(Type valuesGeneratorType)
+        {
+            if (!typeof(IValidateSetValuesGenerator).IsAssignableFrom(valuesGeneratorType))
```
iSazonov: It is useful for developers only - so maybe we skip the reflection?

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1455,6 +1533,44 @@ public ValidateSetAttribute(params string[] validValues)
...
+            {
+                throw PSTraceSource.NewArgumentException("valuesGeneratorType");
+            }
+
+            s_ValidValuesGeneratorCache.TryGetValue(valuesGeneratorType, out IValidateSetValuesGenerator ValidValuesGeneratorCacheEntry);
+
+            if (ValidValuesGeneratorCacheEntry == null)
```
iSazonov: Fixed.

src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -1357,7 +1377,7 @@ internal static Attribute GetAttribute(AttributeAst attributeAst)
             if (attributeType == null)
             {
                 throw InterpreterError.NewInterpreterException(attributeAst, typeof(RuntimeException), attributeAst.Extent,
-                    "CustomAttributeTypeNotFound", ParserStrings.CustomAttributeTypeNotFound, attributeAst.TypeName.FullName);
+                    "TypeNotFound", ParserStrings.TypeNotFound, attributeAst.TypeName.FullName);
```
iSazonov: It is @lzybkr request https://github.com/PowerShell/PowerShell/pull/3784#discussion_r124377253

daxian-dbw: > I think I'd reuse TypeNotFound instead of adding a new message.

But `CustomAttributeTypeNotFound` is not a new resource string, and it looks to me fit pretty well in the `if (attributeType == null)` situation:
`Cannot find the type for custom attribute '{0}'. Make sure that the assembly that contains this type is loaded.`

src/System.Management.Automation/engine/Attributes.cs
daxian-dbw: I think basic validations are necessary. For example, do you allow people to use an internal/private type as an argument? `Activator.CreateInstance` works on non-public types but I don't think we should support non-public types.

daxian-dbw: And I think when validation fails, the thrown exception should be informational. `NewArgumentException("valuesGeneratorType")` doesn't indicate what is the problem.

src/System.Management.Automation/engine/Attributes.cs
daxian-dbw: Agreed that wrapping the exception is not necessary.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1455,6 +1532,42 @@ public ValidateSetAttribute(params string[] validValues)
...
+            if (!s_ValidValuesGeneratorCache.TryGetValue(valuesGeneratorType, out IValidateSetValuesGenerator ValidValuesGeneratorCacheEntry))
+            {
+                // Add a valid values generator to the cache.
+                // We don't cache valid values.
+                // We expect that valid values can be cached in the valid values generator.
+                ValidValuesGeneratorCacheEntry = (IValidateSetValuesGenerator)Activator.CreateInstance(valuesGeneratorType);
+                s_ValidValuesGeneratorCache.TryAdd(valuesGeneratorType, ValidValuesGeneratorCacheEntry);
+            }
+
+            validValuesGenerator = ValidValuesGeneratorCacheEntry;
```
daxian-dbw: How about replacing the whole `if {}` block with 
```
validValuesGenerator = s_ValidValuesGeneratorCache.GetOrAdd(valuesGeneratorType, (key) =>(IValidateSetValuesGenerator)Activator.CreateInstance(key))
```

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1455,6 +1533,44 @@ public ValidateSetAttribute(params string[] validValues)
...
+            }
+
+            s_ValidValuesGeneratorCache.TryGetValue(valuesGeneratorType, out IValidateSetValuesGenerator ValidValuesGeneratorCacheEntry);
+
+            if (ValidValuesGeneratorCacheEntry == null)
+            {
+                // Add a valid values generator to the cache.
+                // We don't cache valid values.
+                // We expect that valid values can be cached in the valid values generator.
+                ValidValuesGeneratorCacheEntry = (IValidateSetValuesGenerator)Activator.CreateInstance(valuesGeneratorType);
```
daxian-dbw: #Close

src/System.Management.Automation/engine/Attributes.cs
daxian-dbw: #Closed

src/System.Management.Automation/engine/Attributes.cs
iSazonov: I added a check IsNonPublic.

We use the same pattern ` NewArgumentException` in all constructors in the file. And the exception is intended for developers only not users.

src/System.Management.Automation/engine/parser/Compiler.cs
iSazonov: @lzybkr  @daxian-dbw Please make a final conclusion - I can't throw a coin. :-)

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Good catch!

Fixed.

powercode:
After thinking a bit more about this, what I would want is `[ValidateCompleterSet]`

It would combine two features: `IArgumentCompleter` for getting values, and `ValidateSet` for limiting what can be passed to the function/cmdlet.

That way we could also get the tooltips etc, and have the validation based on the `CompletionText` of the `CompletionResult`.

I understand I'm a bit late to this party...

iSazonov:
@lzybkr I believe we should stop and merge because here we cannot resolve the problem with ValidValues public property while we share valid values generators.
It seems the @powercode's idea (thanks!) opens alternative way.

src/System.Management.Automation/engine/parser/Compiler.cs
lzybkr: My comment was for a different place in the code and I stand by that comment. This location should not be changed.

lzybkr:
I'm fine with merging soon - but the test failure needs to be worked out.

I'll be very busy for the next 1.5 weeks, much of the time offline, so don't be too concerned if you don't see any response from me.

src/System.Management.Automation/engine/Attributes.cs
daxian-dbw: #Closed

test/powershell/Language/Classes/Scripting.Classes.Attributes.Tests.ps1
```diff
@@ -216,3 +216,263 @@ Describe 'Type resolution with attributes' -Tag "CI" {
...
+        }
+
+        It 'Can implement CachedValidValuesGeneratorBase in PowerShell' {
+            Get-TestValidateSetPS4 -Param1 "TestString1" -ErrorAction SilentlyContinue | Should BeExactly "TestString1"
+        }
+
+        It 'Can implement CachedValidValuesGeneratorBase with cache expiration in PowerShell' {
+            Get-TestValidateSetPS5 -Param1 "TestString1" -ErrorAction SilentlyContinue | Should BeExactly "TestString1"
+            Get-TestValidateSetPS5 -Param1 "TestString1" -ErrorAction SilentlyContinue | Should BeExactly "TestString1"
+            Start-Sleep 2
```
daxian-dbw: I think the failure in CI build is due to this. You can either change to `base(1)`, or make it sleep for 3 seconds (not sure which is more reliable).

daxian-dbw:
I think this is almost ready to be merged -- 2 more comments to be addressed.
> I believe we should stop and merge because here we cannot resolve the problem with ValidValues public property while we share valid values generators.

I don't have the context of the `'the problem with ValidValues'`, can you please elaborate it a bit? But no matter what, it should not block this PR and the further discussion should be captured in an issue.

iSazonov:
@daxian-dbw
From lzybkr optional idea:
> I could imagine wanting to know the command, maybe the command ast, the actual parameter the attribute is applied to, and maybe the already bound parameters.
...
We can add an interface:
public interface IExtendedValidateArguments

The idea is to pass a context to a valid value generator. 
Using the new interface we can pass the context to `Validate()` but we have public property ValidValues - I don`t see how we can pass the context to the property. The root of the problem is that we cache and share the generator. If we add "context" to the cache, the effectiveness of the cache will decrease to 0.

src/System.Management.Automation/engine/parser/Compiler.cs
iSazonov: Sorry, I lost the thread.

Reverted.

test/powershell/Language/Classes/Scripting.Classes.Attributes.Tests.ps1
iSazonov: I am not happy with the solution but set 3 seconds.

daxian-dbw (approved):
LGTM.

daxian-dbw:
I will merge this PR soon. I assume #3744 is fixed by this PR, so for further enhancement discussion, please open a new issue to continue.

@iSazonov now that the design and implementation are complete, could you please update the PR description with a summary of the design and new functionalities? We need that information for the documentation.

iSazonov:
@daxian-dbw I added the PR description - please review.

daxian-dbw:
@iSazonov Perfect! Thank you for the great work!

iSazonov:
Many thanks @lzybkr for ideas and helps!

Menelion:
Hi everyone,  
I found this PR when searching for exactly the same solution: how to implement `validateSet` from a dynamically generated array of subfolders in a PS script. I see that this is merged, but is there any documentation about this? If not, could you please either point me to the test where I can see it in action or just drop a couple lines to illustrate it? And one more question: if I run Windows 10 1709, do I need to update PowerShell to have this working?  
Thanks!

markekraus:
@Menelion Generally you should open a new issue for this kind of thing rather than comment on a closed PR you can reference a PR in an issue by doing `#3784` and it will appear like this: #3784 

You can find some documentation of the API here https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.ivalidatesetvaluesgenerator?view=pscore-6.0.0

This features was added in PowerShell Core 6.0.0. To use, your PowerShell code must run in PowerShell Core 6.0.0 or newer. 

You can find example usage here in this PR or at https://github.com/PowerShell/PowerShell/blob/master/test/powershell/Language/Classes/Scripting.Classes.Attributes.Tests.ps1

iSazonov:
>  where I can see it in action or just drop a couple lines to illustrate it?

@Menelion You could see our tests which was added in the PR.

felixfbecker:
I am trying to use this, but I always get `The type or namespace name 'IValidateSetValuesGenerator' could not be found (are you missing a using directive or an assembly reference?)`

I am using the latest `PowerShellStandard.Library` (`5.1.0-preview-06`). How can I reference this interface?

lzybkr:
That interface was introduced in `6.0.0`, so you would need to reference to https://powershell.myget.org/feed/powershell-core/package/nuget/System.Management.Automation

PR #3785, opened 2017-05-15
Do not send telemetry in CI

JamesWTruher:
This removes the telemetry semaphore file in CI for both Travis-CI and AppVeyor

These are done in the travis-cs and appveyor scripts because I only want to remove the telemetry from our CI systems. I want to continue to gather telemetry from local builds which tells us how many are using custom vs release builds.

### DIFF
--- a/tools/appveyor.psm1
+++ b/tools/appveyor.psm1
@@ -311,6 +311,12 @@ function Invoke-AppVeyorTest
         Write-Host -Foreground Green 'Running all CoreCLR tests..'
     }
 
+    # Remove telemetry semaphore file in CI
+    $telemetrySemaphoreFilepath = Join-Path $env:CoreOutput DELETE_ME_TO_DISABLE_CONSOLEHOST_TELEMETRY
+    if ( Test-Path "${telemetrySemaphoreFilepath}" ) {
+        Remove-Item -Force ${telemetrySemaphoreFilepath}
+    }
+
     Start-PSPester -bindir $env:CoreOutput -outputFile $testResultsNonAdminFile -Unelevate -Tag @() -ExcludeTag ($ExcludeTag + @('RequireAdminOnWindows'))
     Write-Host -Foreground Green 'Upload CoreCLR Non-Admin test results'
     Update-AppVeyorTestResults -resultsFile $testResultsNonAdminFile
--- a/tools/travis.ps1
+++ b/tools/travis.ps1
@@ -123,6 +123,12 @@ if ($isFullBuild) {
     $pesterParam['ThrowOnFailure'] = $true
 }
 
+# Remove telemetry semaphore file in CI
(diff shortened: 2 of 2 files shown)


tools/appveyor.psm1
```diff
@@ -311,6 +311,12 @@ function Invoke-AppVeyorTest
         Write-Host -Foreground Green 'Running all CoreCLR tests..'
     }
 
+    # Remove telemetry semaphore file in CI
+    $TelemetrySemaphoreFilePath = join-path $env:CoreOutput DELETE_ME_TO_DISABLE_CONSOLEHOST_TELEMETRY
```
SteveL-MSFT: We should stick with Pascal casing for cmdlet calls to make it consistent and easier to read

SteveL-MSFT (changes requested):

tools/appveyor.psm1
JamesWTruher: do you mean `join-path` should be `Join-Path`? Shall I fix all of these in this file?

iSazonov: VS Code shows tons of warnings for Build.psm1 - if we want fix it all we should a separate PR.

tools/appveyor.psm1
```diff
@@ -311,6 +311,12 @@ function Invoke-AppVeyorTest
...
+    if ( test-path "${TelemetrySemaphoreFilePath}" ) {
+        remove-item -force ${TelemetrySemaphoreFilePath}
```
iSazonov: We can exclude `test-path` - `remove-item -force` will make everything we need.

tools/travis.ps1
```diff
@@ -123,6 +123,12 @@ if ($isFullBuild) {
     $pesterParam['ThrowOnFailure'] = $true
 }
 
+# Remove telemetry semaphore file in CI
+$TelemetrySemaphoreFilePath = join-path $output DELETE_ME_TO_DISABLE_CONSOLEHOST_TELEMETRY
+if ( test-path "${TelemetrySemaphoreFilePath}" ) {
+    remove-item -force ${TelemetrySemaphoreFilePath}
```
iSazonov: The same about `test-path`.

tools/appveyor.psm1
SteveL-MSFT: We should have a separate issue for other scripts/modules.  For this file, we should just fix the casing on the new code.

SteveL-MSFT: Created https://github.com/PowerShell/PowerShell/issues/3791

JamesWTruher:
i think it should be ok now

SteveL-MSFT (approved):
LGTM

TravisEz13 (approved):

PR #3801, opened 2017-05-17
Make ConsoleHost honor NoEcho on Unix platforms

### LINKED ISSUE #3183: ReadKey does not honor NoEcho on OSX.
Steps to reproduce
------------------

Run the following and enter any printable character.

```powershell
$host.UI.RawUI.ReadKey("NoEcho,IncludeKeyDown")                                                                     
```

Expected behavior
-----------------

Entered character is not echoed.

```powershell
$host.UI.RawUI.ReadKey("NoEcho,IncludeKeyDown")

VirtualKeyCode Character ControlKeyState KeyDown
-------------- --------- --------------- -------
            87         w               0    True
```

Above is from PS 5.1.14393.693 and v6.0.0-alpha.16 on Windows 10.

Actual behavior
---------------
Character is echoed. 

```powershell
$host.UI.RawUI.ReadKey("NoEcho,IncludeKeyDown")
w
VirtualKeyCode Character ControlKeyState KeyDown
-------------- --------- --------------- -------
            87         w               0    True
```

Environment data
----------------

```powershell
GitCommitId                    v6.0.0-alpha.16                                                                                     
CLRVersion                                                                                                                         
```

```powershel
PS /Users/thezim> sw_vers                                                                                                             
ProductName:	Mac OS X
ProductVersion:	10.12.3
BuildVersion:	16D32
PS /Users/thezim>
```

JamesWTruher:
Fix #3183
this also affects Linux as well.

This does not provide support for IncludeKeyDown or IncludeKeyUp, but does not
echo the character when provided. In order to support IncludeKeyDown and IncludeKeyUp a
more substantial rewrite will be needed as Console.ReadKey doesn't support this behavior.

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostRawUserInterface.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostRawUserInterface.cs
@@ -1745,10 +1745,10 @@ namespace Microsoft.PowerShell
         /// from the keyboard device, blocking processing until a keystroke is
         /// typed that matches the specified keystroke options.
         /// </summary>
-        /// <param name="options">Unused</param>
+        /// <param name="options">Only NoEcho is supported.</param>
         public override KeyInfo ReadKey(ReadKeyOptions options)
         {
-            ConsoleKeyInfo key = Console.ReadKey();
+            ConsoleKeyInfo key = Console.ReadKey((options & ReadKeyOptions.NoEcho) != 0);
             return new KeyInfo((int)key.Key, key.KeyChar, new ControlKeyStates(), true);
         }
 


JamesWTruher:
I should add that I haven't been able to figure out a way to actually test this other than manually. I investigated how the corefx guys test Console.ReadKey, and they do it manually. That is, their automation calls a test which then requires that a user type keys which are then validated by the user.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostRawUserInterface.cs
```diff
@@ -1745,10 +1746,16 @@ public override string ToString()
         /// from the keyboard device, blocking processing until a keystroke is
         /// typed that matches the specified keystroke options.
         /// </summary>
-        /// <param name="options">Unused</param>
+        /// <param name="options">Only NoEcho is supported.</param>
         public override KeyInfo ReadKey(ReadKeyOptions options)
         {
-            ConsoleKeyInfo key = Console.ReadKey();
+            ConsoleKeyInfo key;
+            if ((options & ReadKeyOptions.NoEcho) == 0) {
```
andschwa: `key = Console.ReadKey(options.HasFlag(NoEcho))`?

JamesWTruher: fixed

daxian-dbw: Actually, we should avoid `HasFlag`, because it's much, much more expensive than a simple bitwise test. `HasFlag` calls into a native implementation -- see the implementation here: https://github.com/dotnet/coreclr/blob/e67851210d1c03d730a3bc97a87e8a6713bbf772/src/vm/reflectioninvocation.cpp#L3699

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostRawUserInterface.cs
```diff
@@ -763,6 +763,7 @@ public override
                                 if ((((options & ReadKeyOptions.IncludeKeyDown) != 0) &&
                                         inputRecords[0].KeyEvent.KeyDown) ||
                                     (((options & ReadKeyOptions.IncludeKeyUp) != 0) &&
+
```
daxian-dbw: Could you please remove the extra empty line?

daxian-dbw:
Title updated -- it's recommended to briefly describe what is fixed in the PR title, and put "Fix #<issue-number>" in the PR's body. See the `contribution.md` at https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request-submission

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostRawUserInterface.cs
```diff
@@ -1745,10 +1746,10 @@ public override string ToString()
         /// from the keyboard device, blocking processing until a keystroke is
         /// typed that matches the specified keystroke options.
         /// </summary>
-        /// <param name="options">Unused</param>
+        /// <param name="options">Only NoEcho is supported.</param>
         public override KeyInfo ReadKey(ReadKeyOptions options)
         {
-            ConsoleKeyInfo key = Console.ReadKey();
+            ConsoleKeyInfo key = Console.ReadKey(options.HasFlag(ReadKeyOptions.NoEcho));;
```
lzybkr: Please avoid the use of the `HasFlag` api - it is much, much slower than the simple bit test - see the implementation [here](https://github.com/dotnet/coreclr/blob/13e7c4368da664a8b50228b1a5ef01a660fbb2dd/src/vm/reflectioninvocation.cpp#L3064). The expensive part is mostly boxing, but you can see it does other work before the simple bitwise test.

JamesWTruher: it would be better to get uniform input - one person's opinion vs another. My original code was requested to change to this

lzybkr: Sorry, I don't see that feedback, so I can't comment on that.

It's not an opinion that the api is significantly slower, e.g. see [here](https://stackoverflow.com/questions/7368652/what-is-it-that-makes-enum-hasflag-so-slow) - that's the second hit for me on this [search](https://www.bing.com/search?q=Enum.HasFlag)

My preference is to avoid the api completely because:
1. We do have places where the api would make PowerShell measurably slower
2. People working in our code base should be comfortable reading and writing code that is equivalent to this api
3. People tend to copy the style, so avoiding the api helps avoid usage creeping into places that matter.

Again, this is just feedback - I'm not blocking the PR.

JamesWTruher: i agree completely, there's even an open issue in corefx (https://github.com/dotnet/corefx/issues/15453). My original code did not use hasflag, which I will put back.

JamesWTruher:
I think all the feedback has been addressed @daxian-dbw ?

daxian-dbw (approved):

PR #3804, opened 2017-05-18
Removed automatically adding year to copyright message

adityapatwardhan:
Removed automatically adding year to copyright message for file generated by New-ModuleManifest and New-PSRoleCapabilityFile.

### DIFF
--- a/src/System.Management.Automation/engine/Modules/NewModuleManifestCommand.cs
+++ b/src/System.Management.Automation/engine/Modules/NewModuleManifestCommand.cs
@@ -930,7 +930,7 @@ namespace Microsoft.PowerShell.Commands
 
                 if (string.IsNullOrEmpty(_copyright))
                 {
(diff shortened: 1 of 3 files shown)


iSazonov:
It seems new tests not needed - why validate an entity what is absent? Or alternatively we would check a exact format of the copyright string.

adityapatwardhan:
@iSazonov fixed.

lzybkr:
I agree with @iSazonov - we don't really need a test for this.

If the code were to change, maintainers will want to understand the reasoning behind the change before merging - regardless of the existence of a test.

And you also added a file in your last commit that shouldn't be there.

adityapatwardhan:
@iSazonov  @lzybkr Removed tests and files added inadvertently. 
We were recommended by CELA (legal) to remove year from copyright strings.

SteveL-MSFT (approved):
LGTM

iSazonov (approved):
LGTM

PR #3806, opened 2017-05-18
Make Get-ChildItem continue enumeration when encountering error on contained item (#2856)

### LINKED ISSUE #2856: Get-ChildItem on Linux terminates abruptly when accessing directory fails with access denied or file not found
```powershell
PS > /home/<user>/ dir

dir : Access to path '/home/<user>/.gvfs' is denied

PS >
```

Result:
------
Enumeration of paths and files terminates on access denied error

Expected
------
Get-ChildItem should skip inaccessible paths (with error) and continue enumerating files and paths that are accessible.

jeffbi:
Fix #2856 

Added try/catch within the enumeration loop to allow the enumeration to continue after encountering an error such as an item within the directory being deleted or renamed.

To assist in testing, two new internal test hooks have been added which cause Get-ChildItem to either delete or rename a specific file (file name hard-coded) when encountered during enumeration.

### DIFF
--- a/src/System.Management.Automation/engine/Utils.cs
+++ b/src/System.Management.Automation/engine/Utils.cs
@@ -1581,10 +1581,11 @@ namespace System.Management.Automation.Internal
         internal static bool StopwatchIsNotHighResolution;
 
         // Will be either "delete" or "rename" during tests.
-        internal static string GciEnumerationAction = null;
+        internal static bool GciEnumerationActionDelete = false;
+        internal static bool GciEnumerationActionRename = false;
 
         /// <summary>This member is used for internal test purposes.</summary>
-        public static void SetTestHook(string property, object value)
+        public static void SetTestHook(string property, bool value)
         {
             var fieldInfo = typeof(InternalTestHooks).GetField(property, BindingFlags.Static | BindingFlags.NonPublic);
             if (fieldInfo != null)
--- a/src/System.Management.Automation/engine/Utils.cs
+++ b/src/System.Management.Automation/engine/Utils.cs
@@ -1580,8 +1580,9 @@ namespace System.Management.Automation.Internal
         // Simulate 'System.Diagnostics.Stopwatch.IsHighResolution is false' to test Get-Uptime throw
         internal static bool StopwatchIsNotHighResolution;
 
-        // Will be either "delete" or "rename" during tests.
+        // Used in the FileSystemProvider to simulate deleting a file during enumeration in Get-ChildItem
         internal static bool GciEnumerationActionDelete = false;
+        // Used in the FileSystemProvider to simulate renaming a file during enumeration in Get-ChildItem
         internal static bool GciEnumerationActionRename = false;
 
         /// <summary>This member is used for internal test purposes.</summary>
--- a/src/System.Management.Automation/engine/Utils.cs
+++ b/src/System.Management.Automation/engine/Utils.cs
@@ -13,6 +13,7 @@ using Microsoft.Win32;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Collections;
 using System.Collections.ObjectModel;
 using System.Collections.Generic;
 using System.Collections.Concurrent;
@@ -1580,8 +1581,13 @@ namespace System.Management.Automation.Internal
         // Simulate 'System.Diagnostics.Stopwatch.IsHighResolution is false' to test Get-Uptime throw
         internal static bool StopwatchIsNotHighResolution;
 
+        // Name of a file to either delete or rename during directory enumeration.
+        internal static string GciEnumerationActionFilename = null;
+        // New name of the above file when renaming. Used only when GciEnumerationActionFilename is set.
+        internal static string GciEnumerationActionRename = null;
+
         /// <summary>This member is used for internal test purposes.</summary>
-        public static void SetTestHook(string property, bool value)
+        public static void SetTestHook(string property, object value)
         {
             var fieldInfo = typeof(InternalTestHooks).GetField(property, BindingFlags.Static | BindingFlags.NonPublic);
             if (fieldInfo != null)
--- a/src/System.Management.Automation/engine/Utils.cs
+++ b/src/System.Management.Automation/engine/Utils.cs
@@ -1580,10 +1580,8 @@ namespace System.Management.Automation.Internal
         // Simulate 'System.Diagnostics.Stopwatch.IsHighResolution is false' to test Get-Uptime throw
         internal static bool StopwatchIsNotHighResolution;
 
-        // Name of a file to either delete or rename during directory enumeration.
-        internal static string GciEnumerationActionFilename = null;
-        // New name of the above file when renaming. Used only when GciEnumerationActionFilename is set.
-        internal static string GciEnumerationActionRename = null;
+        // Will be either "delete" or "rename" during tests.
+        internal static string GciEnumerationAction = null;
 
         /// <summary>This member is used for internal test purposes.</summary>
         public static void SetTestHook(string property, object value)
--- a/src/System.Management.Automation/engine/Utils.cs
+++ b/src/System.Management.Automation/engine/Utils.cs
@@ -13,7 +13,6 @@ using Microsoft.Win32;
 using System.Globalization;
 using System.IO;
 using System.Linq;
-using System.Collections;
 using System.Collections.ObjectModel;
 using System.Collections.Generic;
 using System.Collections.Concurrent;
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -1659,50 +1659,79 @@ namespace Microsoft.PowerShell.Commands
                             return;
                         }
 
-                        bool attributeFilter = true;
-                        bool switchAttributeFilter = true;
-                        bool filterHidden = false;           // "Hidden" is specified somewhere in the expression
-                        bool switchFilterHidden = false;     // "Hidden" is specified somewhere in the parameters
-
-                        if (null != evaluator)
+                        // Internal test code, run only if the
+                        // "GciEnumerationActionFilename" test hook is set
+                        var testActionFilename = InternalTestHooks.GciEnumerationActionFilename;
+                        if (filesystemInfo.Name == testActionFilename)
                         {
-                            attributeFilter = evaluator.Evaluate(filesystemInfo.Attributes);  // expressions
-                            filterHidden = evaluator.ExistsInExpression(FileAttributes.Hidden);
-                        }
-                        if (null != switchEvaluator)
-                        {
-                            switchAttributeFilter = switchEvaluator.Evaluate(filesystemInfo.Attributes);  // switch parameters
-                            switchFilterHidden = switchEvaluator.ExistsInExpression(FileAttributes.Hidden);
-                        }
-
-                        bool hidden = false;
-                        if (!Force) hidden = (filesystemInfo.Attributes & FileAttributes.Hidden) != 0;
-
-                        // if "Hidden" is explicitly specified anywhere in the attribute filter, then override
-                        // default hidden attribute filter.
-                        // if specification is to return all containers, then do not do attribute filter on
-                        // the containers.
-                        bool attributeSatisfy =
-                            ((attributeFilter && switchAttributeFilter) ||
-                                ((returnContainers == ReturnContainers.ReturnAllContainers) &&
-                                ((filesystemInfo.Attributes & FileAttributes.Directory) != 0)));
-
-                        if (attributeSatisfy && (filterHidden || switchFilterHidden || Force || !hidden))
-                        {
-                            if (nameOnly)
+                            var fullName = Path.Combine(directory.FullName, filesystemInfo.Name);
+                            var newFilename = InternalTestHooks.GciEnumerationActionRename;
+                            if (String.IsNullOrEmpty(newFilename))
                             {
-                                WriteItemObject(
-                                    filesystemInfo.Name,
-                                    filesystemInfo.FullName,
-                                    false);
+                                File.Delete(fullName);
                             }
                             else
                             {
-                                if (filesystemInfo is FileInfo)
-                                    WriteItemObject(filesystemInfo, filesystemInfo.FullName, false);
-                                else
-                                    WriteItemObject(filesystemInfo, filesystemInfo.FullName, true);
+                                var newFullName = Path.Combine(directory.FullName, newFilename);
+                                File.Move(fullName, newFullName);
                             }
+                        }
+
+                        try
+                        {
+                            bool attributeFilter = true;
+                            bool switchAttributeFilter = true;
+                            bool filterHidden = false;           // "Hidden" is specified somewhere in the expression
+                            bool switchFilterHidden = false;     // "Hidden" is specified somewhere in the parameters
+
+                            if (null != evaluator)
+                            {
+                                attributeFilter = evaluator.Evaluate(filesystemInfo.Attributes);  // expressions
+                                filterHidden = evaluator.ExistsInExpression(FileAttributes.Hidden);
+                            }
+                            if (null != switchEvaluator)
+                            {
+                                switchAttributeFilter = switchEvaluator.Evaluate(filesystemInfo.Attributes);  // switch parameters
+                                switchFilterHidden = switchEvaluator.ExistsInExpression(FileAttributes.Hidden);
+                            }
+
+                            bool hidden = false;
+                            if (!Force) hidden = (filesystemInfo.Attributes & FileAttributes.Hidden) != 0;
+
+                            // if "Hidden" is explicitly specified anywhere in the attribute filter, then override
+                            // default hidden attribute filter.
+                            // if specification is to return all containers, then do not do attribute filter on
+                            // the containers.
+                            bool attributeSatisfy =
+                                ((attributeFilter && switchAttributeFilter) ||
+                                    ((returnContainers == ReturnContainers.ReturnAllContainers) &&
+                                    ((filesystemInfo.Attributes & FileAttributes.Directory) != 0)));
+
+                            if (attributeSatisfy && (filterHidden || switchFilterHidden || Force || !hidden))
+                            {
+                                if (nameOnly)
+                                {
+                                    WriteItemObject(
+                                        filesystemInfo.Name,
+                                        filesystemInfo.FullName,
+                                        false);
+                                }
+                                else
+                                {
+                                    if (filesystemInfo is FileInfo)
+                                        WriteItemObject(filesystemInfo, filesystemInfo.FullName, false);
+                                    else
+                                        WriteItemObject(filesystemInfo, filesystemInfo.FullName, true);
+                                }
+                            }
+                        }
+                        catch (System.IO.FileNotFoundException ex)
+                        {
+                            WriteError(new ErrorRecord(ex, "DirIOError", ErrorCategory.ReadError, directory.FullName));
+                        }
+                        catch (UnauthorizedAccessException ex)
+                        {
+                            WriteError(new ErrorRecord(ex, "DirUnauthorizedAccessError", ErrorCategory.PermissionDenied, directory.FullName));
                         }
                     }// foreach
                 }// foreach
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -1659,7 +1659,6 @@ namespace Microsoft.PowerShell.Commands
                             return;
                         }
 
-#if DEBUG
                         // Internal test code, run only if the
                         // 'GciEnumerationActionFilename' test hook is set
                         var testActionFilename = InternalTestHooks.GciEnumerationActionFilename;
@@ -1677,7 +1676,6 @@ namespace Microsoft.PowerShell.Commands
                                 File.Move(fullName, newFullName);
                             }
                         }
-#endif
 
                         try
                         {
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -1659,22 +1659,22 @@ namespace Microsoft.PowerShell.Commands
                             return;
                         }
 
-                        // Internal test code, run only if the
-                        // 'GciEnumerationAction' test hook is set
-                        if (InternalTestHooks.GciEnumerationAction == "delete")
+                        // Internal test code, run only if one of the
+                        // 'GciEnumerationAction' test hooks are set.
+                        if (InternalTestHooks.GciEnumerationActionDelete)
                         {
-                            if (filesystemInfo.Name == "c")
+                            if (filesystemInfo.Name == "c283d143-2116-4809-bf11-4f7d61613f92")
                             {
                                 var fullName = Path.Combine(directory.FullName, filesystemInfo.Name);
                                 File.Delete(fullName);
                             }
                         }
-                        else if (InternalTestHooks.GciEnumerationAction == "rename")
+                        else if (InternalTestHooks.GciEnumerationActionRename)
                         {
-                            if (filesystemInfo.Name == "B")
+                            if (filesystemInfo.Name == "B1B691A9-B7B1-4584-AED7-5259511BEEC4")
                             {
                                 var fullName = Path.Combine(directory.FullName, filesystemInfo.Name);
-                                var newFullName = Path.Combine(directory.FullName, "Z");
+                                var newFullName = Path.Combine(directory.FullName, "77efd2bb-92aa-4ad3-979a-18936a4bd565");
                                 File.Move(fullName, newFullName);
                             }
                         }
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -1659,8 +1659,9 @@ namespace Microsoft.PowerShell.Commands
                             return;
                         }
 
+#if DEBUG
                         // Internal test code, run only if the
-                        // "GciEnumerationActionFilename" test hook is set
+                        // 'GciEnumerationActionFilename' test hook is set
                         var testActionFilename = InternalTestHooks.GciEnumerationActionFilename;
                         if (filesystemInfo.Name == testActionFilename)
                         {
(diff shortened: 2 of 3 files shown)


iSazonov:
@jeffbi I added link on Issue

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -13,6 +13,7 @@
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Collections;
```
iSazonov: Is it really need?

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,80 @@ protected override void InvokeDefaultAction(string path)
...
-                        bool switchFilterHidden = false;     // "Hidden" is specified somewhere in the parameters
-
-                        if (null != evaluator)
-                        {
-                            attributeFilter = evaluator.Evaluate(filesystemInfo.Attributes);  // expressions
-                            filterHidden = evaluator.ExistsInExpression(FileAttributes.Hidden);
-                        }
-                        if (null != switchEvaluator)
+                        // Internal test code, run only if the
+                        // "GciEnumerationActionFilename" test hook is set
```
iSazonov: Usually we use single quota.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,80 @@ protected override void InvokeDefaultAction(string path)
...
+                        var testActionFilename = InternalTestHooks.GciEnumerationActionFilename;
```
iSazonov: For discussion - We never do that but it seems we should mask test hooks by `#if Debug`.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,80 @@ protected override void InvokeDefaultAction(string path)
...
-                                ((filesystemInfo.Attributes & FileAttributes.Directory) != 0)));
-
-                        if (attributeSatisfy && (filterHidden || switchFilterHidden || Force || !hidden))
+                        try
                         {
-                            if (nameOnly)
+                            bool attributeFilter = true;
+                            bool switchAttributeFilter = true;
+                            bool filterHidden = false;           // "Hidden" is specified somewhere in the expression
+                            bool switchFilterHidden = false;     // "Hidden" is specified somewhere in the parameters
```
iSazonov: The same about single quota.
And please put comments on separate lines before codes.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,80 @@ protected override void InvokeDefaultAction(string path)
...
+
+                            if (null != evaluator)
                             {
-                                WriteItemObject(
-                                    filesystemInfo.Name,
-                                    filesystemInfo.FullName,
-                                    false);
+                                attributeFilter = evaluator.Evaluate(filesystemInfo.Attributes);  // expressions
```
iSazonov: Please remove the comment or put on separate line before the code and expand it.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,80 @@ protected override void InvokeDefaultAction(string path)
...
+                                filterHidden = evaluator.ExistsInExpression(FileAttributes.Hidden);
                             }
-                            else
+                            if (null != switchEvaluator)
                             {
-                                if (filesystemInfo is FileInfo)
-                                    WriteItemObject(filesystemInfo, filesystemInfo.FullName, false);
+                                switchAttributeFilter = switchEvaluator.Evaluate(filesystemInfo.Attributes);  // switch parameters
```
iSazonov: Please remove the comment or put on separate line before the code and expand it.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,80 @@ protected override void InvokeDefaultAction(string path)
...
+                                switchFilterHidden = switchEvaluator.ExistsInExpression(FileAttributes.Hidden);
+                            }
+
+                            bool hidden = false;
+                            if (!Force) hidden = (filesystemInfo.Attributes & FileAttributes.Hidden) != 0;
+
+                            // if "Hidden" is explicitly specified anywhere in the attribute filter, then override
```
iSazonov: The same about single quota.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,80 @@ protected override void InvokeDefaultAction(string path)
...
+                            if (null != switchEvaluator)
                             {
-                                if (filesystemInfo is FileInfo)
-                                    WriteItemObject(filesystemInfo, filesystemInfo.FullName, false);
+                                switchAttributeFilter = switchEvaluator.Evaluate(filesystemInfo.Attributes);  // switch parameters
+                                switchFilterHidden = switchEvaluator.ExistsInExpression(FileAttributes.Hidden);
+                            }
+
+                            bool hidden = false;
+                            if (!Force) hidden = (filesystemInfo.Attributes & FileAttributes.Hidden) != 0;
```
iSazonov: Please use full pattern:
```c#
if (...)
{
    ...
}
```

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,80 @@ protected override void InvokeDefaultAction(string path)
...
+
+                            // if "Hidden" is explicitly specified anywhere in the attribute filter, then override
+                            // default hidden attribute filter.
+                            // if specification is to return all containers, then do not do attribute filter on
```
iSazonov: Typos - begin with capital letters.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1
```diff
@@ -60,6 +60,28 @@ Describe "Get-ChildItem" -Tags "CI" {
             $file.Count | Should be 1
             $file.Name | Should be "pagefile.sys"
         }
+        It "Should continue enumerating a directory when a contained item is deleted" -Skip:($IsWindows) {
```
iSazonov: Why we skip Windows?

jeffbi:
@iSazonov Thanks for adding the link. I keep thinking that putting it in the title works, but of course it doesn't.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1
jeffbi: Because the error condition doesn't appear on Windows. Unlike Unix, deleting or renaming a file does not have an adverse effect on the enumeration---the file is listed in its original form.

src/System.Management.Automation/engine/Utils.cs
jeffbi: No, removed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Fixed

test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1
iSazonov: So the test is ok on Windows too?

jeffbi: No. On Windows the test will fail because the enumeration will succeed with no errors emitted in the middle of the process, and `$Error.Count` will be zero.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: @iSazonov  I think I need to revert this change. Wrapping that code in `#if debug` is causing the error condition to not occur, and thus the tests and CI to fail.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1
iSazonov: We need to change those tests so that they work the same on all platforms.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: @daxian-dbw @lzybkr  Could you please clarify should we leave test codes in Release build?

test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1
jeffbi: But the underlying behavior is not the same on all platforms. On Unix, if a file is deleted or renamed during enumeration over the result of `DirectoryInfo.EnumerateFiles()`an exception is thrown when attempting to access a property on the `FileSystemInfo` object. On Windows, no exception is thrown, the property is successfully accessed, and the enumeration goes happily on.

The test hook code does not force an error. It sets up the _condition_ which may or may not trigger an error. If the same condition triggers an error on one platform but not on another, how are the tests for _how the error is handled_ expected to behave the same on both platforms?

jeffbi: Sorry, I think I just realized what you meant. You want to change the test so that `if ($IsWindows)` the error count is zero, otherwise it is one. Is that right?

iSazonov: Original Issue:
>Expected
Get-ChildItem should skip inaccessible paths (with error) and continue enumerating files and paths that are accessible.

Our tests must be simple as that expected: dir without throw return expected list (different on different platforms).

src/System.Management.Automation/namespaces/FileSystemProvider.cs
lzybkr: We do leave test hooks in all builds because we primarily test release builds./

test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1
jeffbi: I hope this is what you're looking for. On Unix the tests check that the error was emitted and that the deleted/renamed item was left out of the list. On Windows the tests check that no error was emitted and that the list is complete.

I have also removed the `#if DEBUG` to allow the CI checks to succeed. I would rather the hook code be in debug-only, but the CI and `Start-PSBuild`builds seem to be release builds.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,85 @@ protected override void InvokeDefaultAction(string path)
...
+                        var testActionFilename = InternalTestHooks.GciEnumerationActionFilename;
+                        if (filesystemInfo.Name == testActionFilename)
                         {
-                            switchAttributeFilter = switchEvaluator.Evaluate(filesystemInfo.Attributes);  // switch parameters
-                            switchFilterHidden = switchEvaluator.ExistsInExpression(FileAttributes.Hidden);
+                            var fullName = Path.Combine(directory.FullName, filesystemInfo.Name);
+                            var newFilename = InternalTestHooks.GciEnumerationActionRename;
+                            if (String.IsNullOrEmpty(newFilename))
+                            {
+                                File.Delete(fullName);
```
lzybkr: This test hook frightens me - please think through the security implications of this.

iSazonov: We can hard code the test file names and call InternalTestHooks.GciEnumerationAction to only enable/disable the hook.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
daxian-dbw: I would love to have those test hooks readonly in the release build, so they cannot be messed around to change powershell behavior. Problem is that some tests won't be able to run in a release build, which is not wanted.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1
jeffbi: Here's a version in which the hook now only allow the test script to select what actions to take, but the hook code internally is hard-coded to use specific file names.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1580,8 +1580,11 @@ public static class InternalTestHooks
         // Simulate 'System.Diagnostics.Stopwatch.IsHighResolution is false' to test Get-Uptime throw
         internal static bool StopwatchIsNotHighResolution;
 
+        // Will be either "delete" or "rename" during tests.
+        internal static string GciEnumerationAction = null;
+
         /// <summary>This member is used for internal test purposes.</summary>
-        public static void SetTestHook(string property, bool value)
+        public static void SetTestHook(string property, object value)
```
iSazonov: We can leave the parameter as bool and define two variables `internal static bool GciEnumerationActionDelete = false;` and `internal static bool GciEnumerationActionRename = false;`

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,87 @@ protected override void InvokeDefaultAction(string path)
...
-                        bool switchFilterHidden = false;     // "Hidden" is specified somewhere in the parameters
-
-                        if (null != evaluator)
+                        // Internal test code, run only if the
+                        // 'GciEnumerationAction' test hook is set
+                        if (InternalTestHooks.GciEnumerationAction == "delete")
                         {
-                            attributeFilter = evaluator.Evaluate(filesystemInfo.Attributes);  // expressions
-                            filterHidden = evaluator.ExistsInExpression(FileAttributes.Hidden);
+                            if (filesystemInfo.Name == "c")
```
iSazonov: We need to try to avoid a match with real names. Maybe use Guid-based names for test file names?

src/System.Management.Automation/engine/Utils.cs
jeffbi: I feel like it would be better to not allow both to be set at the same time, which the string property accomplishes.

iSazonov: I talked about two variables - you can enable hooks at different times.

jeffbi: Changed. If both are set at the same time, only the delete will be done.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Fixed. Using GUID-based names.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1
```diff
@@ -34,25 +40,25 @@ Describe "Get-ChildItem" -Tags "CI" {
...
+            $files[3].Name     | Should Be $item_c
+            $files[4].Name     | Should Be $item_D
         }
 
         It "Should list hidden files as well when 'Force' parameter is used" {
             $files = Get-ChildItem -path $TestDrive -Force
             $files | Should not be $null
             $files.Count | Should be 6
-            $files.Name.Contains(".F")
+            $files.Name.Contains($item_F)
```
jeffbi: No doubt. Fixed.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1
```diff
@@ -4,12 +4,18 @@ Describe "Get-ChildItem" -Tags "CI" {
 
         BeforeAll {
             # Create Test data
-            $null = New-Item -Path $TestDrive -Name "a" -ItemType "File" -Force
-            $null = New-Item -Path $TestDrive -Name "B" -ItemType "File" -Force
-            $null = New-Item -Path $TestDrive -Name "c" -ItemType "File" -Force
-            $null = New-Item -Path $TestDrive -Name "D" -ItemType "File" -Force
-            $null = New-Item -Path $TestDrive -Name "E" -ItemType "Directory" -Force
-            $null = New-Item -Path $TestDrive -Name ".F" -ItemType "File" -Force | %{$_.Attributes = "hidden"}
+            $item_a = "a3fe710a-31af-4834-bc29-d0b584589838"
```
jeffbi: The original filenames were `a`, `B`, `c`, `D`, `E`, and `.F`, in that letter casing, which appeared to have been a deliberate choice. To avoid disrupting any existing tests, I elected to keep the same letter casing in the GUID filenames.

It does look like I made a copy/paste error when I was generating GUID-based filenames. The filename for `$Item_E` should have begun with the letter/hex digit `E`. It now does.

jeffbi: The reason I used lower-case in the variable names was to call out that casing was significant.

iSazonov:
LGTM.

@jeffbi Thanks! Great work!

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1580,6 +1580,10 @@ public static class InternalTestHooks
         // Simulate 'System.Diagnostics.Stopwatch.IsHighResolution is false' to test Get-Uptime throw
         internal static bool StopwatchIsNotHighResolution;
 
+        // Will be either "delete" or "rename" during tests.
```
daxian-dbw: The comment should describe where these hooks are used, and used for what.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,87 @@ protected override void InvokeDefaultAction(string path)
...
+                        // Internal test code, run only if one of the
+                        // 'GciEnumerationAction' test hooks are set.
+                        if (InternalTestHooks.GciEnumerationActionDelete)
                         {
-                            attributeFilter = evaluator.Evaluate(filesystemInfo.Attributes);  // expressions
-                            filterHidden = evaluator.ExistsInExpression(FileAttributes.Hidden);
+                            if (filesystemInfo.Name == "c283d143-2116-4809-bf11-4f7d61613f92")
+                            {
+                                var fullName = Path.Combine(directory.FullName, filesystemInfo.Name);
+                                File.Delete(fullName);
```
jeffbi: That was a safety choice. What actually triggers the error is invoking `filesystemInfo.Attributes` further down in the code. I wanted to avoid invoking any properties that might involve lazy evaluation, to avoid having the error occur while running test-hook code.

If we can be sure that using `filesystemInfo.FullName` won't cause the file system to be hit I would be happy to use it instead.

daxian-dbw: I changed it to `filesystemInfo.FullName` and the test passed on windows. I didn't try on Linux or OSX. Are you seeing issue with `filesystemInfo.FullName`?

jeffbi: I haven't seen an issue, no. I was just trying to avoid one. I'll go ahead and make the change to `FullName`.

daxian-dbw: Just tried on Ubuntu 16.04, `filesystemInfo.FullName` also works fine 😄

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,87 @@ protected override void InvokeDefaultAction(string path)
...
 
-                        bool attributeFilter = true;
-                        bool switchAttributeFilter = true;
-                        bool filterHidden = false;           // "Hidden" is specified somewhere in the expression
-                        bool switchFilterHidden = false;     // "Hidden" is specified somewhere in the parameters
-
-                        if (null != evaluator)
+                        // Internal test code, run only if one of the
+                        // 'GciEnumerationAction' test hooks are set.
+                        if (InternalTestHooks.GciEnumerationActionDelete)
```
jeffbi: I'm concerned about hoisting this out of the loop. The point is to have a file go away while the enumeration is in progress. If we do the test outside the loop I don't think we've met the requirement---we've acquired the enumerator but have not yet started the enumeration. That's why we've chosen files that will not be the first in the enumeration.

iSazonov: It seems we can put this **after** `IEnumerable<FileSystemInfo> sortedChildList = childList.OrderBy(c => c.Name, StringComparer.CurrentCultureIgnoreCase);`

jeffbi: I'm not sure that solves the problem. Again, we have an enumerable object but enumeration doesn't start until the `foreach`.

iSazonov: I think we get already the list in `foreach (IEnumerable<FileSystemInfo> childList in target)` - so after that we can safely delete/remove test file.

jeffbi: `target` is just a `List` of `IEnumerable` objects, containing only 1 or 2 entries---an `IEnumerable` for directory names and an `IEnumerable` for file names. I don't think the act of getting a reference to an enumerable object from a list causes enumeration to begin.

daxian-dbw: @jeffbi This is a very good point. Let's keep it as is.

iSazonov: I catched an enumaration issue second time in last week - need to go to school. :-)

The last question is whether to make a test for such normal runtime error?

src/System.Management.Automation/engine/Utils.cs
jeffbi: Fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,87 @@ protected override void InvokeDefaultAction(string path)
...
+                            }
                         }
-                        if (null != switchEvaluator)
+                        else if (InternalTestHooks.GciEnumerationActionRename)
                         {
-                            switchAttributeFilter = switchEvaluator.Evaluate(filesystemInfo.Attributes);  // switch parameters
-                            switchFilterHidden = switchEvaluator.ExistsInExpression(FileAttributes.Hidden);
+                            if (filesystemInfo.Name == "B1B691A9-B7B1-4584-AED7-5259511BEEC4")
+                            {
+                                var fullName = Path.Combine(directory.FullName, filesystemInfo.Name);
```
jeffbi: Fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,87 @@ protected override void InvokeDefaultAction(string path)
...
+                                File.Delete(fullName);
+                            }
                         }
-                        if (null != switchEvaluator)
+                        else if (InternalTestHooks.GciEnumerationActionRename)
                         {
-                            switchAttributeFilter = switchEvaluator.Evaluate(filesystemInfo.Attributes);  // switch parameters
-                            switchFilterHidden = switchEvaluator.ExistsInExpression(FileAttributes.Hidden);
+                            if (filesystemInfo.Name == "B1B691A9-B7B1-4584-AED7-5259511BEEC4")
```
jeffbi: Fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1659,51 +1659,85 @@ protected override void InvokeDefaultAction(string path)
...
-                        bool switchFilterHidden = false;     // "Hidden" is specified somewhere in the parameters
-
-                        if (null != evaluator)
+                        // Internal test code, run only if one of the
+                        // 'GciEnumerationAction' test hooks are set.
+                        if (InternalTestHooks.GciEnumerationActionDelete)
                         {
-                            attributeFilter = evaluator.Evaluate(filesystemInfo.Attributes);  // expressions
-                            filterHidden = evaluator.ExistsInExpression(FileAttributes.Hidden);
+                            if (string.Equals(filesystemInfo.Name, "c283d143-2116-4809-bf11-4f7d61613f92", StringComparison.Ordinal))
```
daxian-dbw: The `string.Equals` check below uses `StringComparison.InvariantCulture` and this one uses `StringComparison.Ordinal`. It's better to be consistent.

daxian-dbw:
@jeffbi thanks for the thorough thinking on the fix and test. One minor comment and it should be good to go once that's addressed.

daxian-dbw (approved):

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: D'oh! Leftover artifact, thanks for spotting that.

Fixed.

PR #3808, opened 2017-05-18
Fix 'Get-Content -Delimiter' to not include the delimiter in the returned lines

### LINKED ISSUE #3706: Get-Content -Delimiter unexpectedly keeps the delimiter in the lines returned
Note: It may be too late to change this behavior, or perhaps it falls into [Bucket 3: Unlikely Grey Area](https://github.com/PowerShell/PowerShell/blob/master/docs/dev-process/breaking-change-contract.md#bucket-3-unlikely-grey-area).
If the former, perhaps a `-TrimDelimiter` option could be added.

By default, `Get-Content` strips the newline character [sequence] from the lines it returns.

The generalization of this concept is to use the `-Delimiter` parameter, which allows specifying a custom line (record) delimiter (terminator / separator).

However, unlike with the default delimiter - a newline - whatever `-Delimiter` argument you specify is _kept_ in the lines returned, which is:

* an inconsistency

* an inconvenience, because the delimiter - which by its nature is not part of the data itself - must be stripped explicitly before further processing.

Steps to reproduce
------------------

I'd expect the following 2 commands to be equivalent:

```powershell
1,2 > t.txt; get-content                                     t.txt | % { "[$_]" }
1,2 > t.txt; get-content -delimiter ([environment]::newline) t.txt | % { "[$_]" }
```

Expected behavior
-----------------

```PowerShell
[1]
[2]
[1]
[2]
```

Actual behavior
---------------

```PowerShell
[1]
[2]
[1
]
[2
]
```

Note how the trailing newlines were retained in the output from the 2nd command.

Environment data
----------------

```powershell
PowerShell Core v6.0.0-alpha (v6.0.0-alpha.18) on macOS 10.12.4
```

mklement0:
Fix #3706

### DIFF
--- a/src/System.Management.Automation/namespaces/FileSystemContentStream.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemContentStream.cs
@@ -725,7 +725,19 @@ namespace Microsoft.PowerShell.Commands
 
             // We've reached the end of file or end of line.
             if (content.Length > 0)
-                blocks.Add(content.ToString());
+            {
+                // Add the block read to the ouptut array list, trimming a trailing delimiter, if present.
+                // Note: If -Tail was specified, we get here in the course of 2 distinct passes:
+                //  - Once while reading backward simply to determine the appropriate *start position* for later forward reading, ignoring the content of the blocks read (in reverse).
+                //  - Then again during forward reading, for regular output processing; it is only then that trimming the delimiter is necessary.
+                //    (Trimming it during backward reading would not only be unnecessary, but could interfere with determining the correct start position.)
+                string contentString = content.ToString();
+                blocks.Add(
+                    !readBackward && contentString.EndsWith(actualDelimiter, StringComparison.Ordinal)
+                        ? contentString.Substring(0, content.Length - actualDelimiter.Length)
+                        : contentString
+                );
+            }
 
             int peekResult = readBackward ? _backReader.Peek() : _reader.Peek();
             if (peekResult != -1)
--- a/test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
@@ -1,24 +1,24 @@
 Describe "Get-Content" -Tags "CI" {
-  $testString = "This is a test content for a file"
-  $nl         = [Environment]::NewLine
-  $firstline  = "Here's a first line "
-  $secondline = " here's a second line"
-  $thirdline  = "more text"
-  $fourthline = "just to make sure"
-  $fifthline  = "there's plenty to work with"
-  $testString2 = $firstline + $nl + $secondline + $nl + $thirdline + $nl + $fourthline + $nl + $fifthline
-  $testPath   = Join-Path -Path $TestDrive -ChildPath testfile1
-  $testPath2  = Join-Path -Path $TestDrive -ChildPath testfile2
+    $testString = "This is a test content for a file"
+    $nl         = [Environment]::NewLine
+    $firstline  = "Here's a first line "
+    $secondline = " here's a second line"
+    $thirdline  = "more text"
+    $fourthline = "just to make sure"
+    $fifthline  = "there's plenty to work with"
+    $testString2 = $firstline + $nl + $secondline + $nl + $thirdline + $nl + $fourthline + $nl + $fifthline
+    $testPath   = Join-Path -Path $TestDrive -ChildPath testfile1
+    $testPath2  = Join-Path -Path $TestDrive -ChildPath testfile2
 
-  BeforeEach {
-    New-Item -Path $testPath -ItemType file -Force -Value $testString
-    New-Item -Path $testPath2 -ItemType file -Force -Value $testString2
-  }
-  AfterEach {
-    Remove-Item -Path $testPath -Force
-    Remove-Item -Path $testPath2 -Force
-  }
-  It "Should throw an error on a directory  " {
+    BeforeEach {
+        New-Item -Path $testPath -ItemType file -Force -Value $testString
+        New-Item -Path $testPath2 -ItemType file -Force -Value $testString2
+    }
+    AfterEach {
+        Remove-Item -Path $testPath -Force
+        Remove-Item -Path $testPath2 -Force
+    }
+    It "Should throw an error on a directory  " {
         try {
             Get-Content . -ErrorAction Stop
             throw "No Exception!"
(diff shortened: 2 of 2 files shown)


iSazonov (commented):
@mklement0  It seems you don't fix "tail".

/cc @jeffbi Could you please review?

mklement0:
@iSazonov Thanks for catching the `-tail` issue - oddly, the tests pass on _macOS_ (which is where I ran them before submitting the PR). I'll look into it.

iSazonov:
I believe the problem is in platform difference of Newline.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
```diff
@@ -63,11 +63,11 @@ Describe "Get-Content" -Tags "CI" {
...
-	  pushd env:
+	  Push-Location env:
 	  $expectedoutput = [Environment]::GetEnvironmentVariable("PATH");
     { Get-Content PATH } | Should Not Throw
 	  Get-Content PATH     | Should Be $expectedoutput
-  	popd
+  	Pop-Location
   }
   #[BugId(BugDatabase.WindowsOutOfBandReleases, 906022)]
   It "should throw 'PSNotSupportedException' when you set-content to an unsupported provider" -Skip:($IsLinux -Or $IsOSX) {
```
adityapatwardhan: Any reason to skip on Linux or OSX?

adityapatwardhan (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
mklement0: I didn't write these tests (and the only modification I made was to resolve aliases to their full cmdlet names), but I can look into it.

mklement0: Again, I can't speak to these tests, but perhaps `BugId(BugDatabase.WindowsOutOfBandReleases, 906022)` holds the key? 

Separately, if you have any pointers for why the PR doesn't work on _Windows_ (have yet to dig into it), that would be helpful.

SteveL-MSFT: Rather than change it here, I suggest we open a new issue to refactor these test cases to use `-TestCase`

SteveL-MSFT: unfortunately, I don't know if that bug database even exists anymore.  that comment is basically worthless now.

mirichmo: I filed Issue #3945 to cover this comment.

mirichmo:
@adityapatwardhan I don't see any more outstanding comments. Do you have any additional concerns?

adityapatwardhan (approved):

mirichmo:
@mklement0 Please address the failing test:

 [-] should Get-Content with a variety of -Tail and -ReadCount values 68ms
   Expected: {2}
   But was:  {3}
   at line: 104 in C:\projects\powershell\test\powershell\Modules\Microsoft.PowerShell.Management\Get-Content.Tests.ps1
   104:     $result.Length    | Should Be 2

mklement0:
@mirichmo: I'll take a look soon. In the meantime, if you have a high-level pointer as to how the platform-specific newline sequence might come into play here (it works in the Unix world, but not on Windows), please let me know.

iSazonov:
@mklement0 Test failed for `-Tail` parameter. If we use the parameter we call `ReadDelimited()` twice - (1) to seek start position - backward read, (2) read forward and parse the file. In the case we can not simply remove the delimiter character from the stream because we use it in SeekItemsBackward() to correct start position. It seems there is still problem places - leave the research to you 😄  Please continue the PR.

Also I see the code is not optimal and we should refactor this in another PR.

mklement0:
Thanks for the tips, @iSazonov.
I hope to get to this soon (I know, deja vu).

mklement0:
Note that the only changes to `Get-Content.Tests.ps1` since the original commit were whitespace normalization (conversion to all-spaces, though I now realize that there are some inconsistencies).

Thanks to @iSazonov  for pointing me to what I hope is the right fix, which turned out to be simple in the end.

SteveL-MSFT (approved):

src/System.Management.Automation/namespaces/FileSystemContentStream.cs
```diff
@@ -725,7 +725,18 @@ private bool ReadDelimited(bool waitChanges, ArrayList blocks, bool readBackward
...
+            {
+                // Add the block read to the ouptut array list, trimming a trailing delimiter, if present.
+                // Note: If -Tail was specified, we get here in the course of 2 distinct passes:
+                //  - Once while reading backward simply to determine the appropriate *start position* for later forward reading, ignoring the content of the blocks read (in reverse).
+                //  - Then again during forward reading, for regular output processing; it is only then that trimming the delimiter is necessary.
+                //    (Trimming it during backward reading would not only be unnecessary, but could interfere with determining the correct start position.)
+                blocks.Add(
+                    !readBackward && content.ToString().EndsWith(actualDelimiter, StringComparison.Ordinal)
+                        ? content.ToString().Substring(0, content.Length - actualDelimiter.Length)
+                        : content.ToString()
```
iSazonov: Here we evaluate `content.ToString()` twice - maybe do this one before `block.Add()`.

iSazonov:
@mklement0 Could you please split functional and formatting changes to two commits? This will help us avoid error skipping.
Use soft reset, then  git add -p.

mklement0:
@iSazonov: I've made the changes and split them into 2 commits (functional vs. formatting) that replaced the earlier ones. 

I think the Travis CI failure is a transient one, unrelated to my changes (it got stuck while setting up the environment). 

Is there a way to re-initiate the test?

SteveL-MSFT:
@mklement0 I restarted that job for you

mklement0:
Thanks, @SteveL-MSFT. It failed again; it looks like there's a fundamental problem with the macOS builds at the moment: it gets stuck at `make install` for the `curl` Homebrew download.

iSazonov:
@mklement0 You can "rename" last commit to restart CI jobs.

mklement0:
Thanks, @iSazonov, but how do you rename a commit? Amend the commit message and push again?

daxian-dbw:
Thanks all! Since we got 2 approvals and @iSazonov's comments are also addressed, I will rebase and merge this PR soon (in about one hour). Please let me know if there are any concerns.

iSazonov:
@mklement0 Yes, you can amend or rebase+reword.

Issue #3463, opened 2017-03-31
When running in a docker container PSReadline crashes


TravisEz13:
Steps to reproduce
------------------

* Install docker
* Run the following in powershell (**note:** it might take me a day or two to get the docker image published, see #3462.)

```powershell
docker run -it microsoft/powershell:windowsservercore
````
* type a command

Expected behavior
-----------------
It should work without error

Actual behavior
---------------
```
PS C:\>
Oops, something went wrong.  Please report this bug with the details below.
Report on GitHub: https://github.com/lzybkr/PSReadLine/issues/new
-----------------------------------------------------------------------
Last 3 Keys:
 UpArrow Enter
 r

Exception:
System.ComponentModel.Win32Exception: Incorrect function
-----------------------------------------------------------------------
remove-module psreadline

PS C:\> $Error| fl * -Force


PSMessageDetails      :
Exception             : System.Management.Automation.MethodInvocationException: Exception calling "ReadLine" with "2"
                        argument(s): "Failed to get console font information." ---> System.Exception: Failed to get
                        console font information. ---> System.ComponentModel.Win32Exception: Incorrect function
                           --- End of inner exception stack trace ---
                           at Microsoft.PowerShell.Internal.ConhostConsole.GetConsoleFontInfo(SafeFileHandle
                        consoleHandle)
                           at Microsoft.PowerShell.Internal.ConhostConsole.StartRender()
                           at Microsoft.PowerShell.PSConsoleReadLine.ReallyRender()
                           at Microsoft.PowerShell.PSConsoleReadLine.Render()
                           at Microsoft.PowerShell.PSConsoleReadLine.InvokePrompt(Nullable`1 key, Object arg)
                           at Microsoft.PowerShell.PSConsoleReadLine.ReadLine(Runspace runspace, EngineIntrinsics
                        engineIntrinsics)
                           at CallSite.Target(Closure , CallSite , Type , Object , Object )
                           --- End of inner exception stack trace ---
                           at
                        System.Management.Automation.ExceptionHandlingOps.ConvertToMethodInvocationException(Exception
                        exception, Type typeToThrow, String methodName, Int32 numArgs, MemberInfo memberInfo)
                           at CallSite.Target(Closure , CallSite , Type , Object , Object )
                           at System.Management.Automation.Interpreter.DynamicInstruction`4.Run(InterpretedFrame frame)
                           at
[...]

PR #3815, opened 2017-05-18
Enable PSReadline to work on NanoServer

SteveL-MSFT:
PSReadline uses two native Win32 apis to determine font and unicode information which is not available on NanoServer.  Fix is to assume the input is one byte in character in size and not using a truetype font.  This should work for most use cases.

Manually tested this change with our published dockerfile for NanoServer.

Fix https://github.com/PowerShell/PowerShell/issues/3463

### DIFF
--- a/src/Microsoft.PowerShell.PSReadLine/ConsoleKeyChordConverter.cs
+++ b/src/Microsoft.PowerShell.PSReadLine/ConsoleKeyChordConverter.cs
@@ -324,7 +324,7 @@ namespace Microsoft.PowerShell
             // get corresponding character  - may be 0, 1 or 2 in length (diacritics)
             var chars = new char[2];
             int charCount = NativeMethods.ToUnicode(
-                        virtualKey, scanCode, state, chars, chars.Length, NativeMethods.MENU_IS_INACTIVE);
+                virtualKey, scanCode, state, chars, chars.Length, NativeMethods.MENU_IS_INACTIVE);
 
             // TODO: support diacritics (charCount == 2)
             if (charCount == 1)
--- a/src/Microsoft.PowerShell.PSReadLine/ConsoleKeyChordConverter.cs
+++ b/src/Microsoft.PowerShell.PSReadLine/ConsoleKeyChordConverter.cs
@@ -1,4 +1,4 @@
-﻿/********************************************************************++
+/********************************************************************++
 Copyright (c) Microsoft Corporation.  All rights reserved.
 --********************************************************************/
 
@@ -323,8 +323,16 @@ namespace Microsoft.PowerShell
 
             // get corresponding character  - may be 0, 1 or 2 in length (diacritics)
             var chars = new char[2];
-            int charCount = NativeMethods.ToUnicode(
-                virtualKey, scanCode, state, chars, chars.Length, NativeMethods.MENU_IS_INACTIVE);
+            int charCount = 1;
+            try
+            {
+                charCount = NativeMethods.ToUnicode(
+                    virtualKey, scanCode, state, chars, chars.Length, NativeMethods.MENU_IS_INACTIVE);
+            }
+            catch (System.EntryPointNotFoundException)
+            {
+                // assume not unicode
+            }
 
             // TODO: support diacritics (charCount == 2)
             if (charCount == 1)
--- a/src/Microsoft.PowerShell.PSReadLine/ConsoleKeyChordConverter.cs
+++ b/src/Microsoft.PowerShell.PSReadLine/ConsoleKeyChordConverter.cs
@@ -231,7 +231,6 @@ namespace Microsoft.PowerShell
             return valid;
         }
 
-#if UNIX
         // this is borrowed from the CoreFX internal System.IO.StdInReader class
         // https://github.com/dotnet/corefx/blob/5b2ae6aa485773cd5569f56f446698633c9ad945/src/System.Console/src/System/IO/StdInReader.cs#L222
         private static ConsoleKey GetKeyFromCharValue(char x, out bool isShift, out bool isCtrl)
@@ -303,9 +302,8 @@ namespace Microsoft.PowerShell
 
             return default(ConsoleKey);
         }
-#else
-        private static bool _toUnicodeApiAvailable = true;
 
+#if !UNIX
         internal static char GetCharFromConsoleKey(ConsoleKey key, ConsoleModifiers modifiers)
         {
             // default for unprintables and unhandled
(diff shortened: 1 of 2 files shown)


src/Microsoft.PowerShell.PSReadLine/ConsoleKeyChordConverter.cs
```diff
@@ -323,8 +323,16 @@ internal static char GetCharFromConsoleKey(ConsoleKey key, ConsoleModifiers modi
 
             // get corresponding character  - may be 0, 1 or 2 in length (diacritics)
             var chars = new char[2];
-            int charCount = NativeMethods.ToUnicode(
-                virtualKey, scanCode, state, chars, chars.Length, NativeMethods.MENU_IS_INACTIVE);
+            int charCount = 1;
+            try
+            {
+                charCount = NativeMethods.ToUnicode(
```
lzybkr: We should find a better way - this will cause an exception on every character typed, not good for performance and extremely annoying when debugging.

There is some one-time initialization code - so one less ugly option is to add a P/Invoke call there just to detect if the api is present or not, setting a flag that is checked here.

src/Microsoft.PowerShell.PSReadLine/ConsoleLib.cs
```diff
@@ -691,7 +691,15 @@ internal static CONSOLE_FONT_INFO_EX GetConsoleFontInfo(SafeFileHandle consoleHa
 
             CONSOLE_FONT_INFO_EX fontInfo = new CONSOLE_FONT_INFO_EX();
             fontInfo.cbSize = Marshal.SizeOf(fontInfo);
-            bool result = NativeMethods.GetCurrentConsoleFontEx(consoleHandle.DangerousGetHandle(), false, ref fontInfo);
+            bool result = true;
+            try
+            {
+                result = NativeMethods.GetCurrentConsoleFontEx(consoleHandle.DangerousGetHandle(), false, ref fontInfo);
```
lzybkr: We should do the same here - detect once so we don't have frequent exceptions.

lzybkr (changes requested):

src/Microsoft.PowerShell.PSReadLine/ConsoleKeyChordConverter.cs
SteveL-MSFT: Will make the change

src/Microsoft.PowerShell.PSReadLine/ConsoleLib.cs
SteveL-MSFT: Will change

src/Microsoft.PowerShell.PSReadLine/ConsoleKeyChordConverter.cs
SteveL-MSFT: I looked into putting the check in Readline.Initialize() but since it was a different class I would have to expose two additional members.  Seemed simpler to keep the state within the classes needing those apis.

SteveL-MSFT:
@lzybkr fixed

src/Microsoft.PowerShell.PSReadLine/ConsoleKeyChordConverter.cs
```diff
@@ -323,8 +325,19 @@ internal static char GetCharFromConsoleKey(ConsoleKey key, ConsoleModifiers modi
 
             // get corresponding character  - may be 0, 1 or 2 in length (diacritics)
             var chars = new char[2];
-            int charCount = NativeMethods.ToUnicode(
-                virtualKey, scanCode, state, chars, chars.Length, NativeMethods.MENU_IS_INACTIVE);
+            int charCount = 1;
+            if (_toUnicodeApiAvailable)
```
lzybkr: After a closer look, I don't think this is the right fix for this case - custom key bindings need a valid results, you'll just return `\0`.

I think we'll need the Unix variant of this code as a fallback, or potentially use it unconditionally, though that might break some more obscure key bindings.

SteveL-MSFT: There's a comment in the code that GetCharFromConsoleKey() is needed to support chords like Ctrl+[.  Seems like the best approach here is to defer to the Unix code if this API isn't available.  I'll make that change.

SteveL-MSFT:
Manually verified, you can use something like this:

```
docker run -it --entrypoint cmd -v C:\users\slee\repos\PowerShell\src\powershell-win-core\bin\Debug\netcoreapp2.0\win10-x64\publish:c:\powershell microsoft/powershell:nanoserver
```

After the container starts in cmd, you can start powershell using `c:\powershell\powershell`

SteveL-MSFT:
@lzybkr good to go?

src/Microsoft.PowerShell.PSReadLine/ConsoleLib.cs
```diff
@@ -386,13 +403,16 @@ public static string ToGestureString(this ConsoleKeyInfo key)
...
             }
 
-#if UNIX
             char c = key.KeyChar;
-#else
-            // Windows cannot use KeyChar as some chords (like Ctrl+[) show up as control characters.
-            char c = ConsoleKeyChordConverter.GetCharFromConsoleKey(key.Key,
-                (mods & ConsoleModifiers.Shift) != 0 ? ConsoleModifiers.Shift : 0);
+#if !UNIX
+            if (_toUnicodeApiAvailable)
```
lzybkr: If this api is not available, based on the comment, I would expect to see some garbage in the output of `Get-PSReadlineKeyHandler -Bound`, e.g. on this line:
```
Ctrl+]                GotoBrace               Go to matching brace
```
(or if you are using Emacs mode):
```
Ctrl+]           CharacterSearch         Read a character and move the cursor to the next occurence of that character
Ctrl+Alt+]       CharacterSearchBackward Read a character and move the cursor to the previous occurence of that char...
```

SteveL-MSFT: I can check next week when back in office where I have Docker and a Nano container, but not sure what you are proposing here when the api is not available.

lzybkr: I'm proposing not degrading the default user experience if possible - which might require a few special cases. I would guess a more general fix isn't possible because this api is that general fix.

SteveL-MSFT: Seems like this is a decision the Nano team should make.  If they want the full experience, they need those apis on Nano.  I'll certainly let them know of the limitation.

SteveL-MSFT: On a Nano container that line says:
```
Ctrl+Oem6             GotoBrace               Go to matching brace
```

lzybkr: Indeed, not very helpful :)

SteveL-MSFT: If you want, I can add specialized code so that if those apis aren't available, we don't have those bindings by default.  Seems like over optimizing for what is today an edge case.  Alternatively, I could have PSReadline fail to load if those apis aren't available even though most of the PSReadline capability seems to run fine on Nano.  Seems like the best option is to document this and have Nano team make the call to add that api.

lzybkr: A few special cases is not over optimizing and is a practical assuming those apis aren't really generally useful.
The default output of `Get-PSReadlineKeyBindings` should not contain `Oem`.

SteveL-MSFT: I created https://github.com/PowerShell/PowerShell/issues/3890 to track this issue, so we can at least unblock Nano for now.

lzybkr (approved):

PR #3828, opened 2017-05-19
Support Link Header pagination in WebCmdlets

### LINKED ISSUE #3041: Invoke-RestMethod should support Link header based pagination
https://tools.ietf.org/html/rfc5988#page-6 defines a standard for the Link header for navigation and standard Link Relation Types: first, last, next, prev for pagination purposes.  We should expose a parameter to automatically retrieve additional pages rather than having every script implement this logic

SteveL-MSFT:
Implements https://github.com/PowerShell/PowerShell-RFC/blob/master/2-Draft-Accepted/RFC0021-Link-header-based-pagination-for-WebCmdlets.md

When the response includes a Link Header (https://tools.ietf.org/html/rfc5988#page-6), for Invoke-WebRequest we create a RelationLink property that is a Dictionary representing the URLs and rel attributes and ensure the URLs are absolute to make it easier for the developer to use.  For Invoke-RestMethod, we expose a -FollowRelLink switch to automatically follow 'next' rel links to the end until we hit the optional -MaximumFollowRelLink parameter value.

Since we will eventually remove the FullClr code, I didn't make the effort to add the same changes to the CoreClr sources to the FullClr sources.

Fix https://github.com/PowerShell/PowerShell/issues/3041

Doc update https://github.com/PowerShell/PowerShell-Docs/pull/1232

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
@@ -37,6 +37,29 @@ namespace Microsoft.PowerShell.Commands
             set { base.CustomMethod = value; }
         }
 
+        /// <summary>
+        /// enable automatic following of rel links
+        /// </summary>
+        [Parameter]
+        [Alias("FL")]
+        public SwitchParameter FollowRelLink
+        {
+            get { return base._followRelLink; }
+            set { base._followRelLink = value; }
+        }
+
+        /// <summary>
+        /// gets or sets the maximum number of rel links to follow
+        /// </summary>
+        [Parameter]
+        [Alias("ML")]
+        [ValidateRange(0, Int32.MaxValue)]
+        public int MaximumFollowRelLink
+        {
+            get { return base._maximumFollowRelLink; }
+            set { base._maximumFollowRelLink = value; }
+        }
+
         #endregion Parameters
 
         #region Helper Methods
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
@@ -53,7 +53,7 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         [Parameter]
         [Alias("ML")]
-        [ValidateRange(0, Int32.MaxValue)]
+        [ValidateRange(1, Int32.MaxValue)]
         public int MaximumFollowRelLink
         {
             get { return base._maximumFollowRelLink; }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
@@ -16,6 +16,9 @@ using System.Globalization;
 using System.Security.Cryptography;
 using System.Threading;
 using System.Xml;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using System.Linq;
 
 namespace Microsoft.PowerShell.Commands
 {
@@ -60,6 +63,21 @@ namespace Microsoft.PowerShell.Commands
         /// Cancellation token source
         /// </summary>
         private CancellationTokenSource _cancelToken = null;
+
+        /// <summary>
+        /// Automatically follow Rel Links
+        /// </summary>
+        internal bool _followRelLink = false;
+
+        /// <summary>
+        /// Automatically follow Rel Links
+        /// </summary>
+        internal Dictionary<string, string> _relationLink = null;
+
+        /// <summary>
+        /// Maximum number of Rel Links to follow
+        /// </summary>
+        internal int _maximumFollowRelLink = -1;
 
         private HttpMethod GetHttpMethod(WebRequestMethod method)
         {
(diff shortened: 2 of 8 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -619,6 +663,40 @@ internal long SetRequestContent(HttpRequestMessage request, IDictionary content)
...
+                _relationLink = new Dictionary<string, string>();
+            }
+            else
+            {
+                _relationLink.Clear();
+            }
+
+            // we only support the URL in angle brackets and `rel`, other attributes are ignored
+            // user can still parse it themselves via the Headers property
+            Regex regex = new Regex("<(?<url>.*?)>;\\srel=\"(?<rel>.*?)\"");
```
lzybkr: You should not create a new `Regex` like this, instead you should prefer the static method, see the [best practices](https://msdn.microsoft.com/en-us/library/gg578045(v=vs.110).aspx).

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -374,91 +392,117 @@ protected override void ProcessRecord()
...
+                            if (_followRelLink)
                             {
-                                ErrorRecord er = new ErrorRecord(new InvalidOperationException(), "MaximumRedirectExceeded", ErrorCategory.InvalidOperation, request);
-                                er.ErrorDetails = new ErrorDetails(WebCmdletStrings.MaximumRedirectionCountExceeded);
-                                WriteError(er);
+                                if (!_relationLink.ContainsKey("next"))
+                                {
+                                    return;
+                                }
+                                Uri = new Uri(_relationLink["next"]);
```
lzybkr: I think it we be better to use a local variable instead of modifying the cmdlet parameter.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -374,91 +392,117 @@ protected override void ProcessRecord()
...
+                                        }
+                                    }
+                                    if (!String.IsNullOrEmpty(detailMsg))
+                                    {
+                                        er.ErrorDetails = new ErrorDetails(detailMsg);
+                                    }
+                                    ThrowTerminatingError(er);
+                                }
+
+                                ParseLinkHeader(response, Uri);
```
lzybkr: Should you skip parsing the header when we're not following pages?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
```diff
@@ -37,6 +37,29 @@ public override string CustomMethod
...
+            get { return base._followRelLink; }
+            set { base._followRelLink = value; }
+        }
+
+        /// <summary>
+        /// gets or sets the maximum number of rel links to follow
+        /// </summary>
+        [Parameter]
+        [Alias("ML")]
+        [ValidateRange(0, Int32.MaxValue)]
```
SteveL-MSFT: Yes, 1 makes sense

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -374,91 +392,117 @@ protected override void ProcessRecord()
...
-                            "received {0}-byte response of content type {1}",
-                            response.Content.Headers.ContentLength,
-                            contentType);
-                        WriteVerbose(respVerboseMsg);
-
-                        if (!response.IsSuccessStatusCode)
+                        if (followedRelLink > 0)
+                        {
+                            string linkVerboseMsg = string.Format(CultureInfo.CurrentCulture,
+                                "Following rel link {0}",
```
SteveL-MSFT: Since there's only 3 I see in this file, I'll fix them with this PR

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
SteveL-MSFT: Will fix

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
SteveL-MSFT: Will fix

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebResponseObject.CoreClr.cs
```diff
@@ -41,6 +41,11 @@ public partial class WebResponseObject
             }
         }
 
+        /// <summary>
+        /// gets the RelationLink property
+        /// </summary>
+        public Dictionary<string, string> RelationLink { get; set; }
```
SteveL-MSFT: Created https://github.com/PowerShell/PowerShell/issues/3842 for the `Headers` issue

SteveL-MSFT: Will make the setter internal

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
SteveL-MSFT: Will fix

lzybkr (approved):

TravisEz13:
Docs update PR?

SteveL-MSFT:
@TravisEz13 added link to doc PR

PR #3843, opened 2017-05-22
Fix check for Wix toolset

adityapatwardhan:
Wix toolset is part of the AppVeyor base image. The daily build did not generate a MSI since Wix toolset was not found.
AppVeyor updated the Wix version in the base image and our checks failed.
The fix removes the hardcoded version from path. It also guards against multiple side-by-side versions.

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -2462,13 +2462,18 @@ function New-MSIPackage
 
     )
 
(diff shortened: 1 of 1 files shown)


raghushantha (approved):

adityapatwardhan:
@PowerShell/powershell-maintainers Can someone have a look? This is blocking daily builds from creating MSI package.

Issue #2900, opened 2016-12-18
Invoke-item fails when a space in the path


rcw5:
I am running PowerShell on my Mac - there is an issue not fixed in 6.0.0 alpha 14 where invoke-item fails when there is a space in the path. It appears to treat the space as a separator as it attempts to open two paths instead of one.

Steps to reproduce
------------------
Create a file, /tmp/My Folder/test.txt

`mkdir '/tmp/My Folder'`
`"Hello" | Out-File '/tmp/My Folder/test.txt' -Force`

cd to that folder then call `Invoke-Item`

`cd '/tmp/My Folder'`
`invoke-item ./test.txt`

Expected behavior
-----------------
Text file should open in TextEdit (at least on my Mac, although I expect default editors vary)

Actual behavior
---------------
Error written to PoSH console:

`The files /tmp/My and /Users/rcw5/Folder/test.txt do not exist.`

If you rename `My Folder` to `MyFolder` then invoke-item works as expected.

Environment data
----------------

```powershell
> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.0.0-alpha
PSEdition                      Core
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   3.0.0.0
GitCommitId                    v6.0.0-alpha.14
CLRVersion
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

PR #3850, opened 2017-05-23
Fix "Invoke-Item" to accept a file path with spaces on Unix platforms

daxian-dbw:
Address #2900

Root Cause
------------
On Unix platforms, we are currently depending on `xdg-open` on Linux and `open` on OSX to open the default program for a registered file type, so the file path is served as an argument to `xdg-open` and `open`. However, we didn't handle the case when path contains spaces.

Fix
---
Use the method `NativeCommandParameterBinder.NeedQuotes`, which is used by powershell native command processor, to check if quote is needed. If yes, add quotes in the same way as our native command processor (see the code [here](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/NativeCommandParameterBinder.cs#L198)).
There is another problem with the current `Invoke-Item` on Linux and OSX -- it cannot invoke an executable properly. The fix is to first run `Process.Start` directly with the file path, and if it fails, then try `xdg-open` or `open`.

Additional Info
----------------
We are focusing on 2 scenarios for "Invoke-Item":
- it can start an executable
- it can open a file with the default program that is registered to the file type.

### DIFF
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -1359,7 +1359,13 @@ namespace Microsoft.PowerShell.Commands
                     ShellExecuteHelper.Start(invokeProcess.StartInfo);
                 }
 #else
-                finally { /* Nothing to do in FullCLR */ }
+                finally
+                {
+                    // Nothing to do in FullCLR.
+                    // This empty 'finally' block is just to match the 'try' block above so that the code can be organized
+                    // in a clean way without too many if/def's.
+                    // Empty finally block will be ignored in release build, so there is no performance concern.
+                }
 #endif
             }
         } // InvokeDefaultAction
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -1327,27 +1327,39 @@ namespace Microsoft.PowerShell.Commands
             {
                 System.Diagnostics.Process invokeProcess = new System.Diagnostics.Process();
 
-#if UNIX
-                invokeProcess.StartInfo.FileName = Platform.IsLinux ? "xdg-open" : /* OS X */ "open";
-                invokeProcess.StartInfo.Arguments = path;
-                invokeProcess.Start();
-#elif CORECLR
                 try
                 {
-                    // Try Process.Start first. This works for executables even on headless SKUs.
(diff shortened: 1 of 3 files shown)


src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1327,27 +1327,39 @@ protected override void InvokeDefaultAction(string path)
...
+                    // Error code 193 -- BAD_EXE_FORMAT (not a valid Win32 application).
                     // If it's headless SKUs, rethrow.
                     if (Platform.IsNanoServer || Platform.IsIoT) { throw; }
                     // If it's full Windows, then try ShellExecute.
                     ShellExecuteHelper.Start(invokeProcess.StartInfo);
                 }
 #else
-                invokeProcess.StartInfo.FileName = path;
-                invokeProcess.Start();
+                finally { /* Nothing to do in FullCLR */ }
```
adityapatwardhan: Do we need a empty finally block?

adityapatwardhan (approved):
Looks good, except one small comment.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
daxian-dbw: The `finally` block is kept there for `FullCLR` section only to match the `try` block above, so that we don't need to enclose `try {` and `}` in `if/def`. I will add one more comment to make the purpose more clear.
An empty finally block will be ignored during compilation (release version), so there won't be any performance concern. See the code below:
```
static void Main(string[] args)
{
    Process p = null;
    try { p = new Process(); } finally {  }
}
```
**Corresponding IL**
```
.method private hidebysig static void  Main(string[] args) cil managed
{
  .entrypoint
  // Code size       7 (0x7)
  .maxstack  8
  IL_0000:  newobj     instance void [System.Diagnostics.Process]System.Diagnostics.Process::.ctor()
  IL_0005:  pop
  IL_0006:  ret
} // end of method Program::Main
```

iSazonov:
The cmdlet seems to have the most unpredictable behavior😕

daxian-dbw:
@iSazonov I think we are focusing on 2 scenarios for this cmdlet:
- it can start an executable
- it can open a file with the default program that is registered to the file type.

iSazonov:
@daxian-dbw Thanks for clarify. Could you move the comment to the PR description for future docs?

Issue #2653, opened 2016-11-09
Get-Help does not list all 'About*' topics 


MaximoTrinidad:
Steps to reproduce
------------------
PS /home/maxt> Get-Help About* | Select name 

Expected behavior
-----------------
To provide a list of all available 'About_*' topics.

Actual behavior
---------------
Is showing nothing to list.
![gethelpnoaboutlist](https://cloud.githubusercontent.com/assets/19892435/20153899/f2769f08-a692-11e6-96fa-2bdf6a36d777.png)

Environment data
----------------
OS = Ubuntu 16.04.1 LTS Desktop

```powershell
PS /home/maxt> $PSVersionTable                                                  

Name                           Value                                           
----                           -----                                           
PSVersion                      6.0.0-alpha                                     
PSEdition                      Core                                            
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                         
BuildVersion                   3.0.0.0                                         
GitCommitId                    v6.0.0-alpha.12                                 
CLRVersion                                                                     
WSManStackVersion              3.0                                             
PSRemotingProtocolVersion      2.3                                             
SerializationVersion           1.1.0.1

```

joeyaiello:
That's cause we haven't shipped them yet. :)

There's a bunch of problems with about_ topics that need to be addressed by our new publishing pipeline. They'll likely all get fixed at once (or very close together), but I don't have an ETA yet, unfortunately...

iSazonov:
Is the original Issue only on Unix? I cannot reproduce on Windows.

@SteveL-MSFT  If we want fix this in 6.0.0-beta1 we need more details about what @joeyaiello  said. Or will it be an internal solution?

gonhidi:
I have experienced this issue using v6.0.0-alpha.17, both on macOS 10.12.3 Sierra and on Ubuntu 16.04 LTS (using the AppImage): Get-Help about_* lists no topics but Get-Help about_Variables does return the appropriate help page.

SteveL-MSFT:
This doesn't repro on Win10, but does repro on Ubuntu16 with alpha.17 for me.

iSazonov:
@chunqingchen Do we ready to close the Issue?

SteveL-MSFT:
This still repros with #3528.  This one won't make it for beta1.

joeyaiello:
Next step here is to list any missing about_* topics (it should be a short or non-existent list), so we can validate before closing.

[...]

PR #3852, opened 2017-05-24
Fix the issue Get-Help does not support string pattern under Unix

chunqingchen:
Fix issue #2653

Summary of the issue:
------------------------
Under Unix system, get-help would not support wildcast pattern such as about*

Root cause of the issue:
--------------------------
Under unix, GetFiles() is using filePath.IndexOf() to search string pattern, this api actually doesn't support pattern.

Fix:
---
Use regex expression to match *, ? if the search string contains such wildcast.  which are the only two cases are supported.

### DIFF
--- a/src/System.Management.Automation/help/MUIFileSearcher.cs
+++ b/src/System.Management.Automation/help/MUIFileSearcher.cs
@@ -7,6 +7,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Globalization;
+using System.Text.RegularExpressions;
 
 namespace System.Management.Automation
 {
@@ -120,11 +121,25 @@ namespace System.Management.Automation
             // extra logic to select the files that match the given pattern.
             ArrayList result = new ArrayList();
             string[] files = Directory.GetFiles(path);
+
+            string regexPattern = pattern.Replace(".", @"\.");
+            regexPattern = regexPattern.Replace("*", ".*");
+            regexPattern = regexPattern.Replace("?", ".?");
+
             foreach (string filePath in files)
             {
                 if (filePath.IndexOf(pattern, StringComparison.OrdinalIgnoreCase) >= 0)
                 {
                     result.Add(filePath);
+                    break;
+                }
+                // If the input is pattern instead of string, we need to use Regex expression.
+                if (pattern.Contains("*") || pattern.Contains("?"))
+                {             
+                    if (Regex.IsMatch(filePath, regexPattern))
+                    {
+                        result.Add(filePath);
+                    }
                 }
             }
             return (String[])result.ToArray(typeof(string));
--- a/test/powershell/engine/Help/HelpSystem.Tests.ps1
+++ b/test/powershell/engine/Help/HelpSystem.Tests.ps1
@@ -226,3 +226,45 @@ Describe "Get-Help should find help info within help files" -Tags @('CI', 'Requi
         }
     }
 }
+
+Describe "Get-Help should find pattern help files" -Tags "CI" {
+    
+    # There is a bug specific to Travis CI that hangs the test if "get-help" is used to search pattern string. This doesn't repro locally. 
+    # This occurs even if Unix system just returns "Directory.GetFiles(path, pattern);" as the windows' code does.
+    # Since there's currently no way to get the vm from Travis CI and the test PASSES locally on both Ubuntu and MacOS, excluding pattern test under Unix system.
+  
+  BeforeAll {
+    $helpFile1 = "about_testCase1.help.txt"
+    $helpFile2 = "about_testCase.2.help.txt"
+    $culture = (Get-Culture).Name
+    $helpFolderPath = Join-Path $PSHOME $culture
+    $helpFilePath1 = Join-Path $helpFolderPath $helpFile1
+    $helpFilePath2 = Join-Path $helpFolderPath $helpFile2
+    $null = New-Item -ItemType Directory -Path $helpFolderPath -ErrorAction SilentlyContinue -Force
+    # Create at least one help file matches "about*" pattern
+    $null = New-Item -ItemType File -Path $helpFilePath1 -Value "about_test1" -ErrorAction SilentlyContinue
+    $null = New-Item -ItemType File -Path $helpFilePath2 -Value "about_test2" -ErrorAction SilentlyContinue 
+  }
+
+  # Remove the test files
+  AfterAll {
+    Remove-Item $helpFilePath1 -Force -ErrorAction SilentlyContinue
+    Remove-Item $helpFilePath2 -Force -ErrorAction SilentlyContinue
+  }
+
+  $testcases = @(
+                  @{command = {Get-Help about_testCas?1}; testname = "test ? pattern"; result = "about_test1"}
+                  @{command = {Get-Help about_testCase.?}; testname = "test ? pattern with dot"; result = "about_test2"}
+                  @{command = {(Get-Help about_testCase*).Count}; testname = "test * pattern"; result = "2"}
+                  @{command = {Get-Help about_testCas?.2*}; testname = "test ?, * pattern with dot"; result = "about_test2"}
+               )
+
+    It "Get-Help should find pattern help files - <testname>" -TestCases $testcases -Pending: (-not $IsWindows){
+            param (
+            $command,
+            $result
+        )
+        $command.Invoke() | Should Be $result
+    }
+
+}


src/System.Management.Automation/help/MUIFileSearcher.cs
```diff
@@ -115,22 +115,9 @@ private void SearchForFiles()
...
-                    result.Add(filePath);
-                }
-            }
-            return (String[])result.ToArray(typeof(string));
-#else
-            return Directory.GetFiles(path, pattern);
-#endif
+            // On Linux, file names are case sensitive, so we need to put both string to lower case.
+            // We can not use StringComparison.OrdinalIgnoreCase here as api supports OrdinalIgnoreCase won't support pattern
+            return Directory.GetFiles(path.ToLower(), pattern.ToLower());
```
adityapatwardhan: ToLower() is not required for path, as GetFiles considers it as case insensitive.

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -225,4 +225,29 @@ Describe "Get-Help should find help info within help files" -Tags @('CI', 'Requi
...
+        if (!(Test-Path $helpFolderPath))
+        {
+            $null = New-Item -ItemType Directory -Path $helpFolderPath -ErrorAction SilentlyContinue
+        }
+
+        try
+        {
+            # Create at least one help file matches "about*" pattern
+            $null = New-Item -ItemType File -Path $helpFilePath -Value "about_test" -ErrorAction SilentlyContinue
+            (Get-ChildItem $PSHOME\$helpFileWithPattern -Recurse).Count | Should Be (Get-Help about*).Count
```
adityapatwardhan: Can you also verify that the about_testCase.help.txt is one of the files that is returned by Get-Help?

adityapatwardhan: Get-Help can get help content from other paths under PSModulePath. So if there are other about help topics not under $PSHome, the counts would not match.

adityapatwardhan (changes requested):

daxian-dbw:
@chunqingchen The PR title and body should not be a copy of the issue. Please update the PR title and body based on the [contribution.md](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request-submission).

**General Guidelines**

1. Use a meaningful title for the PR to describe what is changed in this PR.
   - Do not include issue number in the title. Put it in the description instead, like "Fix #5".
   - Do not use the issue title as the PR title.
     An issue title is to briefly describe what went wrong, while a PR title is to briefly describe what is changed.
1. The PR body shall provide an additional summary of what/why went wrong and how it's fixed in the PR description.
    - An example: https://github.com/dotnet/corefx/pull/19988

src/System.Management.Automation/help/MUIFileSearcher.cs
chunqingchen: resolved

test/powershell/engine/Help/HelpSystem.Tests.ps1
chunqingchen: resolved by only get the specified files start with about_testcase*

chunqingchen: resolved

chunqingchen:
@daxian-dbw your comment is resolved
@adityapatwardhan your comments are resolved

src/System.Management.Automation/help/MUIFileSearcher.cs
```diff
@@ -113,25 +114,42 @@ private void SearchForFiles()
             }
         }
 
-        private string[] GetFiles(string path, string pattern)
-        {
+		private string[] GetFiles(string path, string pattern)
+		{
```
SteveL-MSFT: looks like a formatting mistake here

src/System.Management.Automation/help/MUIFileSearcher.cs
```diff
@@ -113,25 +114,42 @@ private void SearchForFiles()
...
+                    if (r.Match(fileName).Success)
+                    {
+                        result.Add(filePath);
+                    }
                 }
             }
             return (String[])result.ToArray(typeof(string));
 #else
-            return Directory.GetFiles(path, pattern);
+			return Directory.GetFiles(path, pattern);
```
SteveL-MSFT: why the formatting change?

src/System.Management.Automation/help/MUIFileSearcher.cs
```diff
@@ -113,25 +114,42 @@ private void SearchForFiles()
...
 #if UNIX
             // On Linux, file names are case sensitive, so we need to add
             // extra logic to select the files that match the given pattern.
             ArrayList result = new ArrayList();
             string[] files = Directory.GetFiles(path);
+
+            string regexPattern = pattern.Replace(".", @"\.");
+            regexPattern = regexPattern.Replace("*", ".*");
+            regexPattern = regexPattern.Replace("?", ".?");
+            Regex r = new Regex(regexPattern);
```
SteveL-MSFT: Since this matching is infrequent, you should use an interpreted regex instead of a compiled one

https://msdn.microsoft.com/en-us/library/gg578045(v=vs.110).aspx

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -225,4 +225,32 @@ Describe "Get-Help should find help info within help files" -Tags @('CI', 'Requi
...
+        {
+            $null = New-Item -ItemType Directory -Path $helpFolderPath -ErrorAction SilentlyContinue
+        }
+
+        try
+        {
+            # Create at least one help file matches "about*" pattern
+            $null = New-Item -ItemType File -Path $helpFilePath1 -Value "about_test1" -ErrorAction SilentlyContinue
+            $null = New-Item -ItemType File -Path $helpFilePath2 -Value "about_test2" -ErrorAction SilentlyContinue
+            (Get-Help about_testCase*).Count | Should Be 2
```
SteveL-MSFT: This test case doesn't cover the code change, you explicitly handle *, ?, and .

Seems like you should also have a test case with multiple wildcards

SteveL-MSFT (changes requested):

adityapatwardhan (changes requested):

src/System.Management.Automation/help/MUIFileSearcher.cs
```diff
@@ -113,25 +114,42 @@ private void SearchForFiles()
...
-        {
+		private string[] GetFiles(string path, string pattern)
+		{ 
 #if UNIX
             // On Linux, file names are case sensitive, so we need to add
             // extra logic to select the files that match the given pattern.
             ArrayList result = new ArrayList();
             string[] files = Directory.GetFiles(path);
+
+            string regexPattern = pattern.Replace(".", @"\.");
```
chunqingchen: resolved

chunqingchen: This actually should not be used. RegEx.Escape will also escape the '*' and '?' which we want them as normal character to replace later.

src/System.Management.Automation/help/MUIFileSearcher.cs
chunqingchen: The formatting is fixed. It shows normal under IDE, I have to delete and re tab to fix it.

src/System.Management.Automation/help/MUIFileSearcher.cs
chunqingchen: resolved

src/System.Management.Automation/help/MUIFileSearcher.cs
chunqingchen: resolved

test/powershell/engine/Help/HelpSystem.Tests.ps1
chunqingchen: resolved.

src/System.Management.Automation/help/MUIFileSearcher.cs
```diff
@@ -113,25 +114,42 @@ private void SearchForFiles()
...
+            regexPattern = regexPattern.Replace("*", ".*");
+            regexPattern = regexPattern.Replace("?", ".?");
+            Regex r = new Regex(regexPattern);
+
             foreach (string filePath in files)
```
chunqingchen: return Directory.GetFiles(path).Where(f => Regex.IsMatch(f, regExPattern.ToString())).ToArray();  is not implementing the logic of string which is not using a pattern

chunqingchen:
@adityapatwardhan @SteveL-MSFT Thanks Aditya and Steve. Your comments are resolved.

I have spent much time try to figure out why there's time out exception on Unix test runs on my pull request. It ended up not related to my code and also the code and tests get passed on local unix os (both Ubuntu and MacOS). See my comments with the test case. I have to put a skip on the test case if it is not windows os.

chunqingchen:
@adityapatwardhan @SteveL-MSFT Thanks Aditya and Steve. Your comments are resolved.

I have spent much time try to figure out why there's time out exception on Unix test runs on my pull request. It ended up not related to my code and also the code and tests get passed on local unix os (both Ubuntu and MacOS). See my comments with the test case. I have to put a skip on the test case if it is not windows os.

src/System.Management.Automation/help/MUIFileSearcher.cs
```diff
@@ -114,17 +115,31 @@ private void SearchForFiles()
         }
 
         private string[] GetFiles(string path, string pattern)
-        {
+        {
```
adityapatwardhan: Extra space after {

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -225,4 +225,38 @@ Describe "Get-Help should find help info within help files" -Tags @('CI', 'Requi
             Remove-Item $helpFilePath -Force -ErrorAction SilentlyContinue
         }
     }
+
+    # There is a bug specified to Tranvis CI that hangs the test if "get-help" is used to search pattern string. This doesn't repro locally.
```
adityapatwardhan: Typos: 
Tranvis -> Travis
specified -> specific

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -225,4 +225,38 @@ Describe "Get-Help should find help info within help files" -Tags @('CI', 'Requi
...
+    # Since there's currently no way to get the vm from Tranvis CI and the test PASSES locally on both Ubuntu and MacOS, excluding pattern test under Unix system.
+    It "Get-Help should find pattern help files" -skip: (-not $IsWindows){
+        $helpFile1 = "about_testCase1.help.txt"
+        $helpFile2 = "about_testCase.2.help.txt"
+        $culture = (Get-Culture).Name
+        $helpFolderPath = Join-Path $PSHOME $culture
+        $helpFilePath1 = Join-Path $helpFolderPath $helpFile1
+        $helpFilePath2 = Join-Path $helpFolderPath $helpFile2
+
+        if (!(Test-Path $helpFolderPath))
```
adityapatwardhan: The if block can be replace by the following. If it already exists, it is not modified. 
```Powershell
$null = New-Item -ItemType Directory -Path $helpFolderPath -Force
```

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -225,4 +225,38 @@ Describe "Get-Help should find help info within help files" -Tags @('CI', 'Requi
...
+        {
+            $null = New-Item -ItemType Directory -Path $helpFolderPath -ErrorAction SilentlyContinue
+        }
+
+        try
+        {
+            # Create at least one help file matches "about*" pattern
+            $null = New-Item -ItemType File -Path $helpFilePath1 -Value "about_test1" -ErrorAction SilentlyContinue
+            $null = New-Item -ItemType File -Path $helpFilePath2 -Value "about_test2" -ErrorAction SilentlyContinue           
+            Get-Help about_testCas?1 | Should Match "about_test1"
```
adityapatwardhan: For all the Shoulds, use Be or BeExactly. Match uses regular expression comparison. It is not required here.

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -225,4 +225,38 @@ Describe "Get-Help should find help info within help files" -Tags @('CI', 'Requi
...
+            Get-Help about_testCase.? | Should Match "about_test2"
+            (Get-Help about_testCase*).Count | Should Be 2
+            Get-Help about_testCas?.2* | Should Match "about_test2"
```
adityapatwardhan: Ideally, each It should have only one Should. This ensures that if the first should fails, the rest are not ignored. Take a look at: http://www.powershellmagazine.com/2015/06/04/pester-triangulation-and-reusing-test-cases/
You can pass the values for Get-Help as testcases and values for Should as expected values. Using testcases parameter of It also allows you to give better names for each variation.

adityapatwardhan (changes requested):

test/powershell/engine/Help/HelpSystem.Tests.ps1
SteveL-MSFT: Agree with @adityapatwardhan, this is a good example to use the `-TestCases` parameter of Pester.  Look at other test code in this repo as a sample.

src/System.Management.Automation/help/MUIFileSearcher.cs
chunqingchen: resolved

test/powershell/engine/Help/HelpSystem.Tests.ps1
chunqingchen: resolved

test/powershell/engine/Help/HelpSystem.Tests.ps1
chunqingchen: resolved

test/powershell/engine/Help/HelpSystem.Tests.ps1
chunqingchen: resolved

test/powershell/engine/Help/HelpSystem.Tests.ps1
chunqingchen: resolved with triangulation format. a good way !

chunqingchen:
@adityapatwardhan @SteveL-MSFT  your comments are resolved

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -225,4 +225,45 @@ Describe "Get-Help should find help info within help files" -Tags @('CI', 'Requi
...
+    # Since there's currently no way to get the vm from Tranvis CI and the test PASSES locally on both Ubuntu and MacOS, excluding pattern test under Unix system.
+
+    $testcases = @(
+                    @{command = "Get-Help about_testCas?1"; testname = "test ? pattern"; result = "about_test1"}
+                    @{command = "Get-Help about_testCase.?"; testname = "test ? pattern with dot"; result = "about_test2"}
+                    @{command = "(Get-Help about_testCase*).Count"; testname = "test * pattern"; result = "2"}
+                    @{command = "Get-Help about_testCas?.2*"; testname = "test ?, * pattern with dot"; result = "about_test2"}
+                 )
+
+    It "Get-Help should find pattern help files - <testname>" -TestCases $testcases -skip: (-not $IsWindows){
```
SteveL-MSFT: unless the testcases are used in other tests, I would prefer to have these inline to the `It` statement as it makes it easier to associate the two

SteveL-MSFT (changes requested):

test/powershell/engine/Help/HelpSystem.Tests.ps1
adityapatwardhan: A good example of using test cases would be: https://github.com/PowerShell/PowerShell/blob/02b5f357a20e6dee9f8e60e3adb9025be3c94490/test/powershell/Modules/Microsoft.PowerShell.Management/Get-Service.Tests.ps1#L15

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -225,4 +225,45 @@ Describe "Get-Help should find help info within help files" -Tags @('CI', 'Requi
...
+        $helpFolderPath = Join-Path $PSHOME $culture
+        $helpFilePath1 = Join-Path $helpFolderPath $helpFile1
+        $helpFilePath2 = Join-Path $helpFolderPath $helpFile2
+
+        $null = New-Item -ItemType Directory -Path $helpFolderPath -ErrorAction SilentlyContinue -Force
+
+        try
+        {
+            # Create at least one help file matches "about*" pattern
+            $null = New-Item -ItemType File -Path $helpFilePath1 -Value "about_test1" -ErrorAction SilentlyContinue
```
adityapatwardhan: Can these be done in beforeall and cleanup in afterall?

adityapatwardhan (changes requested):

test/powershell/engine/Help/HelpSystem.Tests.ps1
chunqingchen: resolved

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -226,3 +226,45 @@ Describe "Get-Help should find help info within help files" -Tags @('CI', 'Requi
...
+                  @{command = "(Get-Help about_testCase*).Count"; testname = "test * pattern"; result = "2"}
+                  @{command = "Get-Help about_testCas?.2*"; testname = "test ?, * pattern with dot"; result = "about_test2"}
+               )
+
+    It "Get-Help should find pattern help files - <testname>" -TestCases $testcases -skip: (-not $IsWindows){
+            param (
+            $command,
+            $result
+        )
+        Invoke-Expression $command | Should Be $result
```
adityapatwardhan: Don't use Invoke-Expression instead use splatting instead. Look at: https://github.com/PowerShell/PowerShell/commit/28ec9a351197ff35390aea7b1174c97035805bfb

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -226,3 +226,45 @@ Describe "Get-Help should find help info within help files" -Tags @('CI', 'Requi
...
+  }
+
+  $testcases = @(
+                  @{command = "Get-Help about_testCas?1"; testname = "test ? pattern"; result = "about_test1"}
+                  @{command = "Get-Help about_testCase.?"; testname = "test ? pattern with dot"; result = "about_test2"}
+                  @{command = "(Get-Help about_testCase*).Count"; testname = "test * pattern"; result = "2"}
+                  @{command = "Get-Help about_testCas?.2*"; testname = "test ?, * pattern with dot"; result = "about_test2"}
+               )
+
+    It "Get-Help should find pattern help files - <testname>" -TestCases $testcases -skip: (-not $IsWindows){
```
adityapatwardhan: This should be marked as Pending on Non-Windows and not skipped. So we can revisit later.

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -226,3 +226,45 @@ Describe "Get-Help should find help info within help files" -Tags @('CI', 'Requi
         }
     }
 }
+
+Describe "Get-Help should find pattern help files" -Tags "CI" {
+    
+    # There is a bug specific to Travis CI that hangs the test if "get-help" is used to search pattern string. This doesn't repro locally. 
+    # This occurs even if Unix system just returns "Directory.GetFiles(path, pattern);" as the windows' code does.
+    # Since there's currently no way to get the vm from Tranvis CI and the test PASSES locally on both Ubuntu and MacOS, excluding pattern test under Unix system.
```
adityapatwardhan: Typo: Tranvis -> Travis.

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -226,3 +226,45 @@ Describe "Get-Help should find help info within help files" -Tags @('CI', 'Requi
...
+    $null = New-Item -ItemType Directory -Path $helpFolderPath -ErrorAction SilentlyContinue -Force
+    # Create at least one help file matches "about*" pattern
+    $null = New-Item -ItemType File -Path $helpFilePath1 -Value "about_test1" -ErrorAction SilentlyContinue
+    $null = New-Item -ItemType File -Path $helpFilePath2 -Value "about_test2" -ErrorAction SilentlyContinue 
+  }
+
+  # Remove the test files
+  AfterAll {
+    Remove-Item $helpFilePath1 -Force -ErrorAction SilentlyContinue
+    Remove-Item $helpFilePath2 -Force -ErrorAction SilentlyContinue
```
adityapatwardhan: $helpFolderPath does not seem to be cleaned up? Maybe you can just delete $helpFolderPath recursively.

adityapatwardhan (changes requested):

test/powershell/engine/Help/HelpSystem.Tests.ps1
chunqingchen: resolved

test/powershell/engine/Help/HelpSystem.Tests.ps1
chunqingchen: $helpFolderPath is an existed folder contains other help files.

test/powershell/engine/Help/HelpSystem.Tests.ps1
chunqingchen: resolved

test/powershell/engine/Help/HelpSystem.Tests.ps1
chunqingchen: In that case (Get-Help about_testCase*).Count would needs to be explicitly written in the test case one more time. If u think this is better than invoke-expression i will update later.

SteveL-MSFT (approved):

test/powershell/engine/Help/HelpSystem.Tests.ps1
adityapatwardhan: After talking to Jim, in general Invoke-Expression is not a best practice. An alternative approach would be to make the command as script block. So the test would look like:

```Powershell
$testcases = @( 
                  @{command = {Get-Help about_testCas?1}; testname = "test ? pattern"; result = "about_test1"} 
                  @{command = {Get-Help about_testCase.?}; testname = "test ? pattern with dot"; result = "about_test2"} 
                  @{command = {(Get-Help about_testCase*).Count}; testname = "test * pattern"; result = "2"} 
                  @{command = {Get-Help about_testCas?.2*}; testname = "test ?, * pattern with dot"; result = "about_test2"} 
               ) 
```

And replace Invoke-Expression with 

```Powershell
$command.Invoke() | Should be $result
```

adityapatwardhan (changes requested):

chunqingchen:
@adityapatwardhan your comment has been resolved

adityapatwardhan (approved):

src/System.Management.Automation/help/MUIFileSearcher.cs
```diff
@@ -120,11 +121,25 @@ private string[] GetFiles(string path, string pattern)
...
             {
                 if (filePath.IndexOf(pattern, StringComparison.OrdinalIgnoreCase) >= 0)
                 {
                     result.Add(filePath);
+                    break;
+                }
+                // If the input is pattern instead of string, we need to use Regex expression.
+                if (pattern.Contains("*") || pattern.Contains("?"))
+                {             
+                    if (Regex.IsMatch(filePath, regexPattern))
```
lzybkr: Why are we using regular expression matching instead of wildcard matching?

FWIW - wildcard matching already converts the wildcard to a regex - so this code is probably doing something similar but subtly different.

daxian-dbw: I agree. We basically need to implementation `Directory.GetFiles(path, pattern);` in a case-insensitive way, and the code at https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs#L5961 is doing the same thing.

daxian-dbw: #4053 is opened for this.

PR #3863, opened 2017-05-25
Bug: Start-Trace handles file paths incorrectly

IISResetMe:
`Start-Trace` neglects qualifying arguments to the -o and -pf command line
switches, causing logman (and subsequently Start-Trace) to exit with error code
`0x80070057` and return:

```
Error:
The argument is incorrect.
```

This is easy to reproduce on Windows:

```
PS> $dir = mkdir "a foldername with spaces" 
PS> $OutFile = Join-Path $dir trace.log
PS> Start-Trace test -OutputFilePath $OutFile -ETS
Argument 'foldername' is unknown.
Argument 'with' is unknown.
Argument 'spaces\trace.log' is unknown.
Argument 'pf' requires additional parameters.

Error:
The parameter is incorrect.
```

This commit moves basic input validation for output file paths to the
ProviderFilePath and OutputFilePath parameter definitions, and adds text
qualifiers (double-quotes) around input arguments


PaulHigin (changes requested):
Please include a test for this.

TravisEz13:
Will close if no response to review within a week.

IISResetMe:
Apologies for the delay, but I'm unsure of how to structure the test. I can't find any precedent or existing examples for tests with dependencies on Windows-specific external executables. Any guidance would be much appreciated

PaulHigin:
I think all you need to do is test Start-Trace with an output file path having spaces (per your repro steps).  You can use "$IsWindows" to limit the test to only run on Windows platforms.  Do a search for "$IsWindows" to see how other tests restrict to Windows only platforms.

@JamesWTruher Do you have any other advice such as where the test should reside?

PaulHigin (approved):
LGTM.  I don't think a test for this is all that important.  This is a good fix to get in.

TravisEz13:
@IISResetMe Please try to write tests in the future.  I've filed an issue for the missing tests.  Thanks for the contribution.

Issue #2240, opened 2016-09-12
Port Web Cmdlets to CoreCLR


jeffbi:
Port these cmdlets to CoreCLR:

- [ ] ConvertTo-Html
- [ ] Send-MailMessage
- [ ] New-WebServiceProxy

Two of these are mentioned in #2123

thezim:
Is Send-MailMessage waiting on [corefx](https://github.com/dotnet/corefx/issues/1006)?

zippy1981:
New-WebServiceProxy would be great.

SteveL-MSFT:
`New-WebServiceProxy` depends on System.Web.Services.dll which isn't part of corefx

Glober777:
I have a lot of scripts depending on New-WebServiceProxy so can't easily migrate them over to PowerShell 6

markekraus:
@Glober777 I looked into making something work for this. We can make something that works similar in PowerShell core, but it would not be backwards compatible in any way. The newer APIs for web services in .NET Framework and .NET Core behave differently than the deprecated `System.Web.Services` API. The reality is that `New-WebServiceProxy` will likely never be ported to PowerShell core, but, a new cmdlet may be created to work with SOAP and WSDL.

jonathanmedd:
@markekraus That would be great. My [Brickset module](https://github.com/jonathanmedd/BricksetModule) is based around New-WebServiceProxy and I would like to update it to support PowerShell Core.

markekraus:
@jonathanmedd For clarity, by "we" I mean Microsoft and the Community. SOAP is not something I personally have any need for, so this is very very low on my priority stack and not on on any of my roadmaps. So, I'm really wanting someone who actually needs the commands to lead the charge on this one. You game?

jonathanmedd:
@markekraus totally understood, wasn't intentional to convey any expectation that you were personally going to deal with it ;-) Potentially I could be, would need to give it some thought. I guess I would need to learn C# first, although this might be an impetus that finally got me there.....

iSazonov:
> I guess I would need to learn C# first

First we need RFC for the new cmdlet. You could design the cmdlet if you have an expirience in web services. This will be a big step in solving this Issue.

Sarafian:
[...]

Issue #2123, opened 2016-08-29
Several Cmdlets from Microsoft.PowerShell.Utility are missing


ffeldhaus:
The following Function and Cmdlets are missing from Microsoft.PowerShell.Utility

``` powershell
Function        ConvertFrom-SddlString                             3.1.0.0    Microsoft.PowerShell.Utility
Cmdlet          ConvertFrom-String                                 3.1.0.0    Microsoft.PowerShell.Utility
Cmdlet          Convert-String                                     3.1.0.0    Microsoft.PowerShell.Utility
Cmdlet          ConvertTo-Html                                     3.1.0.0    Microsoft.PowerShell.Utility
Cmdlet          Export-PSSession                                   3.1.0.0    Microsoft.PowerShell.Utility
Cmdlet          Import-PSSession                                   3.1.0.0    Microsoft.PowerShell.Utility
Cmdlet          Out-GridView                                       3.1.0.0    Microsoft.PowerShell.Utility
Cmdlet          Out-Printer                                        3.1.0.0    Microsoft.PowerShell.Utility
Cmdlet          Send-MailMessage                                   3.1.0.0    Microsoft.PowerShell.Utility
Cmdlet          Show-Command                                       3.1.0.0    Microsoft.PowerShell.Utility
Cmdlet          Unblock-File                                       3.1.0.0    Microsoft.PowerShell.Utility
Cmdlet          Update-List                                        3.1.0.0    Microsoft.PowerShell.Utility
```

It would be great if at least the `Convert-FromString`, `Convert-String`, `ConvertTo-Html`, `Send-MailMessage` and `Update-List` Cmdlets get implemented. The `Out-GridView`and `Show-Command` Cmdlets may be difficult due to graphical interface requirements, similar for `Out-Printer` with external printer requirement. A removal notice (ideally including explanation) should be added to the release notes for the other Cmdlets if they will be permanantly removed.

vors:
[Command availability table](https://github.com/PowerShell/PowerShell/blob/master/docs/KNOWNISSUES.md#command-availability) need to be updated.

MaximoTrinidad:
I just coming up with these missing ones.  But specially I was need the ConvertTo-Html cmdlet.

paxpacis:
+1 on the missing ConvertTo-Html

edvail:
Missing Send-MailMessage

paxpacis:
[...]

PR #3869, opened 2017-05-26
Enable Send-MailMessage on CoreCLR

jeffbi:
Contributes to the completion of #2123 and #2240.

Now that the requisite components are implemented in corefx, compilation of the Send-MailMessage cmdlet has been enabled.

This PR includes a test script that allows the cmdlet to be tested against a working local SMTP server, but only on Linux and under specific conditions. A separate issue will be opened for improving test coverage and including additional platforms.

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
@@ -1,0 +1,135 @@
+Describe "Basic Send-MailMessage tests" -Tags CI {
+    function read-mail
+    {
+        Param(
+            [parameter(Mandatory=$true)]
+            [String]
+            $mailBox
+        )
+
+        $state = "init"
+        $mail = Get-Content $mailBox
+        $rv = @{}
+        foreach ($line in $mail)
+        {
+            switch ($state)
+            {
+                "init"
+                {
+                    if ($line.Length -gt 0)
+                    {
+                        $state = "headers"
+                    }
+                }
+                "headers"
+                {
+                    if ($line.StartsWith("From: "))
+                    {
+                        $rv.From = $line.Substring(6)
+                    }
+                    elseif ($line.StartsWith("To: "))
+                    {
+                        if ($rv.To -eq $null)
+                        {
+                            $rv.To = @()
+                        }
+
+                        $rv.To += $line.Substring(4)
+                    }
+                    elseif ($line.StartsWith("Subject: "))
+                    {
+                        $rv.Subject = $line.Substring(9);
+                    }
+                    elseif ($line.Length -eq 0)
+                    {
+                        $state = "body"
+                    }
+                }
+                "body"
+                {
+                    if ($line.Length -eq 0)
+                    {
+                        $state = "done"
+                        continue
+                    }
+
+                    if ($rv.Body -eq $null)
+                    {
+                        $rv.Body = @()
+                    }
+
+                    $rv.Body += $line
+                }
+            }
+        }
+
+        return $rv
+    }
+
+    BeforeAll {
+        $PesterArgs = @{ Name = "Can send mail message from user to self"}
+        $alreadyHasMail = $true
+
+        if (-not $IsLinux)
+        {
+            $PesterArgs["Skip"] = $true
+            $PesterArgs["Name"] += " (skipped: not Linux)"
+            return
+        }
+
+        $user = "jeff"
+        $inPassword = Select-String "^${user}:" /etc/passwd -ErrorAction SilentlyContinue
+        if (-not $inPassword)
+        {
+            $PesterArgs["Pending"] = $true
+            $PesterArgs["Name"] += " (pending: user not in /etc/passwd)"
+            return
+        }
+
+        $domain = "Chinstrap"
+        if ($domain -ne [System.Environment]::MachineName)
+        {
+            $PesterArgs["Pending"] = $true
+            $PesterArgs["Name"] += " (pending: machine name not '$domain')"
+            return
+        }
+        $address = "$user@$domain"
+        $mailStore = "/var/mail"
+        $mailBox = Join-Path $mailStore $user
+        $mailBoxFile = Get-Item $mailBox -ErrorAction SilentlyContinue
+        if ($mailBoxFile -ne $null -and $mailBoxFile.Length -gt 2)
+        {
+            $PesterArgs["Pending"] = $true
+            $PesterArgs["Name"] += " (pending: mailbox not empty)"
+            return
+        }
+        $alreadyHasMail = $false
+        Set-Content -Value "" -Path $mailBox -Force -ErrorAction SilentlyContinue
+        $mailBoxFile = Get-Item $mailBox -ErrorAction SilentlyContinue
+        if ($mailBoxFile -eq $null -or $mailBoxFile.Length -gt 2)
+        {
+            $PesterArgs["Pending"] = $true
+            $PesterArgs["Name"] += " (pending: did not clear or create mailbox)"
+            return
+        }
+    }
+    AfterAll {
+       if (-not $alreadyHasMail)
+       {
+           Set-Content -Value "" -Path $mailBox -Force -ErrorAction SilentlyContinue
+       }
+    }
+
+    It @PesterArgs {
+        $body = "Greetings from me."
+        $subject = "Test message"
+        Send-MailMessage -To $address -From $address -Subject $subject -Body $body -SmtpServer 127.0.0.1
+        Test-Path -Path $mailBox | Should Be $true
+        $mail = read-mail $mailBox
+        $mail.From | Should BeExactly $address
+        $mail.To.Count | Should BeExactly 1
+        $mail.To[0] | Should BeExactly $address
+        $mail.Body.Count | Should BeExactly 1
+        $mail.Body[0] | Should BeExactly $body
+    }
+}
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
@@ -1,72 +1,90 @@
 Describe "Basic Send-MailMessage tests" -Tags CI {
-    function read-mail
-    {
-        Param(
-            [parameter(Mandatory=$true)]
-            [String]
-            $mailBox
-        )
-
-        $state = "init"
-        $mail = Get-Content $mailBox
-        $rv = @{}
-        foreach ($line in $mail)
+    BeforeAll {
+        function test-smtpserver
         {
-            switch ($state)
+            $rv = $false
+
+            try
             {
-                "init"
-                {
-                    if ($line.Length -gt 0)
-                    {
-                        $state = "headers"
-                    }
-                }
-                "headers"
-                {
-                    if ($line.StartsWith("From: "))
-                    {
-                        $rv.From = $line.Substring(6)
-                    }
-                    elseif ($line.StartsWith("To: "))
-                    {
-                        if ($rv.To -eq $null)
-                        {
-                            $rv.To = @()
-                        }
-
-                        $rv.To += $line.Substring(4)
-                    }
-                    elseif ($line.StartsWith("Subject: "))
-                    {
-                        $rv.Subject = $line.Substring(9);
-                    }
-                    elseif ($line.Length -eq 0)
-                    {
-                        $state = "body"
-                    }
-                }
-                "body"
-                {
-                    if ($line.Length -eq 0)
-                    {
-                        $state = "done"
-                        continue
-                    }
-
-                    if ($rv.Body -eq $null)
-                    {
-                        $rv.Body = @()
-                    }
-
-                    $rv.Body += $line
-                }
+                $tc = New-Object -TypeName System.Net.Sockets.TcpClient -ArgumentList "localhost", 25
+                $rv = $tc.Connected
+                $tc.Close()
             }
+            catch
+            {
+                $rv = false
+            }
+
+            return $rv
         }
 
-        return $rv
-    }
+        function read-mail
+        {
+            Param(
+                [parameter(Mandatory=$true)]
+                [String]
+                $mailBox
+            )
 
-    BeforeAll {
+            $state = "init"
+            $mail = Get-Content $mailBox
+            $rv = @{}
+            foreach ($line in $mail)
+            {
+                switch ($state)
+                {
+                    "init"
+                    {
+                        if ($line.Length -gt 0)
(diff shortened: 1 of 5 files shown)


iSazonov:
@jeffbi You need add new record in test/powershell/engine/DefaultCommands.Tests.ps1

jeffbi:
@iSazonov Thanks, I wasn't aware of this.

I see that there already is a record for Send-MailMessage. Do I just need to add `-or $CoreWindows -or $CoreUnix` into it?

iSazonov:
Yes, you need edit and add -or $CoreWindows -or $CoreUnix

jeffbi:
Done.

iSazonov:
@jeffbi The tests is skipped on both CI - I believe we should remove its at all.

iSazonov:
@jeffbi The tests is skipped on both CI - I believe we should remove its at all.

Maybe we can merge without tests if the cmdlet passes manual tests locally.

Related https://github.com/tedious/DovecotTesting/blob/master/resources/Scripts/Provision.sh#L28

jeffbi:
@iSazonov I would guess that the test is being marked as Pending due to the machine name not matching the one specified in the test script. The test does require some pretty specific configuration.

I have been able to manually send mail to myself on Linux with the cmdlet. I don't have an SMTP server for my Windows box to test with.

@JamesWTruher, @SteveL-MSFT, do you have thoughts on this?

jeffbi:
@iSazonov So you know, when the CI checks failed, the error was pointing to `test/powershell/engine/Remoting/SSHRemotingAPI.Tests.ps1` as the culprit, not my tests. It looks to me like that test script is trying to import a module that does not exist. Updating the entry in `DefaultCommands.Tests.ps1` appears to have fixed the problem, but I don't quite see how.

iSazonov:
I did not saw problem with  test/powershell/engine/Remoting/SSHRemotingAPI.Tests.ps1  - I saw only  failed DefaultCommands.Tests.ps1 that is expected.

I believe it is very expensive install a mail server on CI VMs.

iSazonov:
I did not saw problem with  test/powershell/engine/Remoting/SSHRemotingAPI.Tests.ps1  - I saw only  failed DefaultCommands.Tests.ps1 that is expected.

I believe it is very expensive install a mail server on CI VMs.

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+    AfterAll {
+       if (-not $alreadyHasMail)
+       {
+           Set-Content -Value "" -Path $mailBox -Force -ErrorAction SilentlyContinue
+       }
+    }
+
+    It @PesterArgs {
+        $body = "Greetings from me."
+        $subject = "Test message"
```
SteveL-MSFT: shouldn't you also validate the subject?

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+
+        if (-not $IsLinux)
+        {
+            $PesterArgs["Skip"] = $true
+            $PesterArgs["Name"] += " (skipped: not Linux)"
+            return
+        }
+
+        $user = "jeff"
+        $inPassword = Select-String "^${user}:" /etc/passwd -ErrorAction SilentlyContinue
```
SteveL-MSFT: seems like user 'jeff' will almost always not exist so this test is always `pending`?

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+        if (-not $inPassword)
+        {
+            $PesterArgs["Pending"] = $true
+            $PesterArgs["Name"] += " (pending: user not in /etc/passwd)"
+            return
+        }
+
+        $domain = "Chinstrap"
```
SteveL-MSFT: similar to above, this will almost always fail and thus be `Pending`, seems like if `mail` is installed you should be able to rely on root@computername?

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+        $mailStore = "/var/mail"
+        $mailBox = Join-Path $mailStore $user
+        $mailBoxFile = Get-Item $mailBox -ErrorAction SilentlyContinue
+        if ($mailBoxFile -ne $null -and $mailBoxFile.Length -gt 2)
+        {
+            $PesterArgs["Pending"] = $true
+            $PesterArgs["Name"] += " (pending: mailbox not empty)"
+            return
+        }
+        $alreadyHasMail = $false
```
SteveL-MSFT: I don't understand this, you set `$alreadyHasMail` to $true above and then it's always set to $false here

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        return $rv
+    }
+
+    BeforeAll {
+        $PesterArgs = @{ Name = "Can send mail message from user to self"}
+        $alreadyHasMail = $true
+
+        if (-not $IsLinux)
```
SteveL-MSFT: do we need to check if a smtp server is installed or is that always the case on Linux/Mac?

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
jeffbi: On Linux, an SMTP server is not installed by default, since there are at least two to choose from, perhaps more.

I'll have to look into how to detect from PowerShell if anyone's listening on port 25.

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
jeffbi: What it actually should be is `current_user$computername`. While the test takes steps to avoid deleting an existing mailbox, it still made me nervous, so I left those hard-coded names in, partially to call attention to the situation.

To use `root@computername`, the test would need to be running as root in order to access `root`'s mailbox. Is that what the CI environment is?

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
jeffbi: This is one of those steps to avoid deleting an existing mailbox. In the `AfterAll` block, the `$alreadyHasMail` variable is checked before deleting the mailbox. The early return at line 104 keeps that set to `$true` if the mailbox is not empty.

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+        $mailBox = Join-Path $mailStore $user
+        $mailBoxFile = Get-Item $mailBox -ErrorAction SilentlyContinue
+        if ($mailBoxFile -ne $null -and $mailBoxFile.Length -gt 2)
+        {
+            $PesterArgs["Pending"] = $true
+            $PesterArgs["Name"] += " (pending: mailbox not empty)"
+            return
+        }
+        $alreadyHasMail = $false
+        Set-Content -Value "" -Path $mailBox -Force -ErrorAction SilentlyContinue
```
JamesWTruher: do permissions need to be set on this mailbox? I seem to remember that the permissions for mailbox files are special

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+       if (-not $alreadyHasMail)
+       {
+           Set-Content -Value "" -Path $mailBox -Force -ErrorAction SilentlyContinue
+       }
+    }
+
+    It @PesterArgs {
+        $body = "Greetings from me."
+        $subject = "Test message"
+        Send-MailMessage -To $address -From $address -Subject $subject -Body $body -SmtpServer 127.0.0.1
```
JamesWTruher: one more check needed above to be sure that there's an smtp server running on localhost

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+        {
+            $PesterArgs["Pending"] = $true
+            $PesterArgs["Name"] += " (pending: did not clear or create mailbox)"
+            return
+        }
+    }
+    AfterAll {
+       if (-not $alreadyHasMail)
+       {
+           Set-Content -Value "" -Path $mailBox -Force -ErrorAction SilentlyContinue
```
JamesWTruher: should this just be delete the file?

JamesWTruher (changes requested):
the important check to be added is to be sure that an SMTP server is installed and running on the system

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+        $alreadyHasMail = $true
+
+        if (-not $IsLinux)
+        {
+            $PesterArgs["Skip"] = $true
+            $PesterArgs["Name"] += " (skipped: not Linux)"
+            return
+        }
+
+        $user = "jeff"
```
jeffbi: Is `testuser` the username the CI tests will run as?

The permission bits on a user's mailbox are `rw-------` so a test might be able to send mail to another user, but it won't be able to read the mailbox to verify that the mail was sent.

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
jeffbi: They don't seem to be special, just  `rw-------`

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
SteveL-MSFT: current user is probably correct

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
jeffbi: Unless we're running as root or as a member of the `mail` group, we can write to the file but we cannot delete it.

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
jeffbi: Fixed.

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
jeffbi: Fixed.

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
jeffbi: Changed to current user

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
jeffbi: Changed to current user on current machine

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
jeffbi: Fixed

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
iSazonov: We could use an environment variables:
 - AppVeyor - [APPVEYOR_ACCOUNT_NAME ](https://www.appveyor.com/docs/environment-variables/)
 - Travis - [USER](https://docs.travis-ci.com/user/environment-variables/)

iSazonov:
Can we use  [.Net Fake SMTP Server ](https://github.com/cmendible/netDumbster)?

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
```diff
@@ -0,0 +1,147 @@
+Describe "Basic Send-MailMessage tests" -Tags CI {
```
mirichmo: Does this test require Administrator on Windows or root on all non-Windows platforms?

mirichmo:
@SteveL-MSFT @JamesWTruher Have all of you concerns been addressed? Do you have any additional comments?

test/powershell/Modules/Microsoft.PowerShell.Utility/Send-MailMessage.Tests.ps1
jeffbi: Currently this test runs only on Linux. So long as the sender/receiver of the email is the currently logged-in user, running as root is not required.

I've been looking into ways to make tests more cross-platform, using a suggestion from @iSazonov, but it's going to take some effort.

My personal feeling is that I'd like to create a separate issue for testing `Send-MailMessage` and have this PR be taken as-is, but I don't know if that works with testing policies.

SteveL-MSFT: @jeffbi create a separate issue for that and link it to this

SteveL-MSFT (approved):

mirichmo:
@JamesWTruher Do you have any additional comments?

JamesWTruher:
LGTM

Issue #3400, opened 2017-03-23
Build powershell core in Visual Studio 2017


daxian-dbw:
With #3398, we are now building powershell core via MSBuild with our building script. However, it seems not working properly in Visual Studio 2017. We need to make it possible to build powershell in VS2017 as well.

powercode:
Should we think about what is a Configuration and what is a Platform?

Linux is a config today, but it seems more reasonable to treat it as a platform.

I have a branch, https://github.com/powercode/PowerShell/tree/vs2017-build,  where I can build with vs2017, but it is untested on Linux/Mac and from the command line.

daxian-dbw:
The term "Platform" sounds like the right one to use, but it seems always indicating the CPU architecture based on my limited MSBuild experience, for example, `Any CPU`, `x64`, and `x86`.

And we have another problem -- with the current project files, `dotnet pack` will generate a NuGet package that only contains assemblies built for windows, which cannot be used reliably on Linux. I wonder if it's possible that we can change the .csproj somehow to make `dotnet pack` produce a single NuGet package that contains both win and unix assemblies. #3417 is tracking it.

Is changing to "Platform" a possible direction? Do you have some resources on the differences between "Configuration" and "Project" in the new MSBuild from .NET Core SDK?

powercode:
Platform may be what we choose. LinuxX64 for example.
They are just monikers to give builders an easy way to set different settings based on Config and platform. 
C++ has had Win32 and x64 while .NET has used AnyCPU, x86 and x64. But I think this traditionally has been Windows specific, so instead of saying Win-x86, they just went with the shorter alternative.
I haven't had much time to dig into the changes with the new msbuild. Seems nice and clean though.

iSazonov:
I wonder how .Net Core resolve the same for itself packages? Can we use the same?

iSazonov:
@powercode Could you make PR with your vs2017 sln file?

oising:
Please, please, this. I am loathe to return to building software like it's 1993. I already had that life. I just wasted three hours trying to make anything work at all in VS 2017 15.3-pre (supposedly it understands net standard 2.x) and it's just a confusing mess.

iSazonov:
I've already started taking little steps in this direction, but @daxian-dbw  too busy these days to do the PR review.

BobbyCannon-Quantros:
It's been a few months... any progress on this?

oising:
Ping pong poong.

[...]

PR #3870, opened 2017-05-26
Move Start-TypeGen logic from Build.psm1 to SDK csproj file

iSazonov:
Related #3400
(Also it is one step to unblock #3690)

Now Microsoft.PowerShell.SDK.csproj:
1. Generate Reference Assembly List ("RefAssemblyList.inc" file) in project local sub directory "gen" (Target Name="TypeCatalogGen"):
 - Check if Reference Assembly List is updated and only then overwrite RefAssemblyList.inc
 - If RefAssemblyList.inc is updated then generate new CorePsTypeCatalog.cs 
2. Support Clean target to remove RefAssemblyList.inc (for rebuild)

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -535,8 +535,8 @@ Fix steps:
 
     # publish netcoreapp2.0 reference assemblies
     try {
-        Push-Location "$PSScriptRoot/src/TypeCatalogGen"
-        $refAssemblies = Get-Content -Path $incFileName | Where-Object { $_ -like "*microsoft.netcore.app*" } | ForEach-Object { $_.TrimEnd(';') }
+        Push-Location "$PSScriptRoot/src/Microsoft.PowerShell.SDK/gen"
+        $refAssemblies = Get-Content -Path "RefAssemblyList.inc" | Where-Object { $_ -like "*microsoft.netcore.app*" } | ForEach-Object { $_.TrimEnd(';') }
         $refDestFolder = Join-Path -Path $publishPath -ChildPath "ref"
 
         if (Test-Path $refDestFolder -PathType Container) {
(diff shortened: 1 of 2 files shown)


iSazonov:
@daxian-dbw I think I'm a little late with the PR, but it could still be useful for a while.

daxian-dbw:
@iSazonov We still need to have `TypeCatalogGen` in our build. With the type catalog, you can use a .NET type without caring whether the assembly that contains it is already loaded because powershell is able to look it up in the type catalog and load the assembly automatically. So it would be a breaking change if we remove the type catalog. And `TypeCatalogGen` will be useful for new features too, for example, the extension methods support @powercode is working on -- we will need to do some analysis of all reference assemblies at build time to build a cache of all .NET extension methods, which will improve the runtime performance. So this PR is not late 😄
BTW, I'm busy on other tasks recently and thus is slow on code review, sorry for that.

iSazonov:
@TravisEz13 @daxian-dbw Could you please continue the code review?

TravisEz13:
@daxian-dbw Are you ok with this change now?

daxian-dbw:
@TravisEz13 I haven't got the time to carefully review this yet. Will do it soon.

iSazonov:
@daxian-dbw Could you please review?

daxian-dbw:
Will the `TypeCatalogGen` target execute when we running `dotnet restore` or `dotnet build` on `Microsoft.PowerShell.SDK.csproj`?

iSazonov:
We call `dotnet msbuild .\Microsoft.PowerShell.SDK.csproj /t:"Clean;TypeCatalogGen" "/property:DesignTimeBuild=true" ` in Build.psm1 in the same point as before - so the behavior must be the same.

Ah, sorry - your question of a direct call - it needs to be checked explicitly because dependencies is deeply.

daxian-dbw:
Great. My concern was that the type catalog (both `.inc` and `.cs`) would be generated when just running `Start-PSBuild` for a build. I will continue the review.
(BTW, I took Monday off, so please don't be concerned if I didn't reply promptly.)

TravisEz13:
@daxian-dbw Do you want to review?

iSazonov:
Should I fix or close the PR?

/cc @daxian-dbw

daxian-dbw:
Sorry that I completely lost the context here. I won't have much time for code review recently. I'm fine we close this PR for now and we can revive it when needed.

PR #3871, opened 2017-05-26
Officially support creating PS class instance that is not bound to any Runspace

### LINKED ISSUE #3651: PowerShell class instance methods cannot run concurrently
This behavior changed between PowerShell versions 5.0 and 5.1.  The change was to make PowerShell class method script blocks bound to the runspace in which they were created.  If a class instance method is run from within a different runspace then it gets "marshaled" back to the original runspace (via engine events).  The result is that if a single class instance is shared among multiple runspaces (e.g., a runspace pool) with the intention to execute instance methods concurrently, the methods run serially instead.

I don't know why this changed between 5.0 and 5.1 but it looks to be by design.  @vors  can you please comment?

Repro steps:
```powershell
PS > (measure-command { .\RunspaceInstanceTest.ps1 }).TotalSeconds

RunspaceInstanceTest.ps1 file script:
class Writer
{
    static WriteLineS([string] $Msg)
    {
        for ($i=0; $i -lt 10; $i++)
        {
            [Console]::WriteLine("Static: Loop $i - $Msg")
            Start-Sleep -MilliSeconds 100
        }
    }

    WriteLineI([string] $msg)
    {
        for ($i=0; $i -lt 10; $i++)
        {
            [Console]::WriteLine("Instance: Loop $i - $Msg")
            Start-Sleep -MilliSeconds 100
        }
    }
}

$script = @'
    param ([object] $writerInstance, [string] $Title)

    class Writer
    {
        static WriteLineS([string] $Msg)
        {
            for ($i=0; $i -lt 10; $i++)
            {
                [Console]::WriteLine("Static: Loop $i - $Msg")
                Start-Sleep -MilliSeconds 100
            }
        }

        WriteLineI([string] $msg)
        {
            for ($i=0; $i -lt 10; $i++)
            {
                [Console]::WriteLine("Instance: Loop $i - $Msg")
                Start-Sleep -MilliSeconds 100
            }
        }
    }

    # Create new instance
    #$writerInstance = [Writer]::new()

    # Instance write
    $writerInstance.WriteLineI("$Title")

    # Static write
    #[Writer]::WriteLineS("$Title")
'@

$writer = [Writer]::new()

$rsp = [runspacefactory]::CreateRunspacePool(1, 10, $host)
$rsp.Open()

class Task
{
    [powershell] $powershell
    [System.IAsyncResult] $Async
}

$tasks = @()

1..5 | foreach {

    $task = [Task]::new()
    $tasks += $task

    $task.powershell = [powershell]::Create()
    $task.powershell.RunspacePool = $rsp
    $task.Async = $task.powershell.AddScript($script).AddArgument($writer).AddArgument("Task $_").BeginInvoke()
}

foreach ($task in $tasks)
{
    $task.powershell.EndInvoke($task.Async)
    $task.powershell.Dispose()
}

$rsp.Dispose()

```
Expected:
5 concurrently running 1 second loops (with sleep) should take about 1 second to run.

Actual Result:
It takes about 5 seconds to run indicating that the concurrent scripts are running serially.

daxian-dbw:
Fix #3651

Issue Summary
-----------------
In PSv5.0, an instance of a PowerShell class is not bound to any Runspace. In PSv5.1, a PowerShell class instance by default is bound to the Runspace where the class is defined. This change was for safety and reliability purpose, but it does break applications that depend on the PSv5.0 behavior. So we need to support a way to allow users to intentionally create a PowerShell class instance that is not bound to any Runspace.

Fix
---
To create an instance not pre-bound to any Runspace, just create the instance from a Runspace other than the one where the corresponding powershell class is defined.
To officially support this approach, we need to remove the `Diagnostics.Assert` that guards this situation, and also add tests to cover this scenario.

### DIFF
--- a/src/System.Management.Automation/engine/runtime/Operations/ClassOps.cs
+++ b/src/System.Management.Automation/engine/runtime/Operations/ClassOps.cs
@@ -70,12 +70,17 @@ namespace System.Management.Automation.Internal
         /// It's not intended to be a public API, but because we generate type in a different assembly it has to be public.
         /// Return type should be SessionStateInternal, but it violates accessibility consistency, so we use object.
         /// </summary>
+        /// <remarks>
+        /// By default, PowerShell class instantiation usually happens in the same Runspace where the class is defined. In
+        /// that case, the created instance will be bound to the session state used to define that class in the Runspace.
+        /// However, if the instantiation happens in a different Runspace where the class is not defined, then the created
+        /// instance won't be bound to any session state.
+        /// </remarks>
         /// <returns>SessionStateInternal</returns>
         public object GetSessionState()
         {
             SessionStateInternal ss = null;
-            bool found = _stateMap.TryGetValue(Runspace.DefaultRunspace, out ss);
-            Diagnostics.Assert(found, "We always should be able to find corresponding SessionState");
+            _stateMap.TryGetValue(Runspace.DefaultRunspace, out ss);
             return ss;
         }
     }
--- a/test/powershell/Language/Classes/Scripting.Classes.MiscOps.Tests.ps1
+++ b/test/powershell/Language/Classes/Scripting.Classes.MiscOps.Tests.ps1
@@ -43,4 +43,75 @@ Describe 'Misc Test' -Tags "CI" {
                 [C1]::new().Bar() | should be "1;2;3;"
         }
     }
+
+    Context 'Class instantiation' {
+        Class C1 {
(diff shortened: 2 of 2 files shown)


SteveL-MSFT (approved):
LGTM

PaulHigin:
Adding link to related issue.  #3651 This looks like a good workaround and this issue can be closed once this PR is accepted.

lzybkr (approved):

PR #3872, opened 2017-05-27
Use different HttpListener for Invoke-WebRequest and Invoke-RestMethod tests

SteveL-MSFT:
On Mac, when the httplistener is stopped, it takes time for the system to clean it up.  It also appears that on Mac, http reservation

is at the port and not the URL.  Fix is to have the two different httplisteners use different ports.  Otherwise due to a race condition
the second listener tries to start on the same port as the previous one that stopped but isn't cleaned up yet by the system so fails.

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -146,11 +146,11 @@ function GetTestData
 Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
     BeforeAll {
-        $null = Start-HttpListener -AsJob
+        $null = Start-HttpListener -AsJob -Port 8080
     }
 
     AfterAll {
-        $null = Stop-HttpListener
+        $null = Stop-HttpListener -Port 8080
     }
 
     # Validate the output of Invoke-WebRequest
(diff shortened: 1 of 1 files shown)


iSazonov:
Can we reuse first created httplistener? Can create it at helper module load time and reuse in all tests?

SteveL-MSFT:
The Webcmdlets tests have two Describe blocks, so it's probably not a good practice to start one at the beginning of the script and cleanup at the end outside of Pester.  Having a global one means someone has to cleanup still.

iSazonov:
And if we were to use a full web server for the tests,  did we worry about it?

SteveL-MSFT:
@iSazonov the problem was installing a full web server for different platforms and the corresponding different code if we end up using, for example, IIS on Windows, and Apache on Linux.

iSazonov:
Sorry for the inaccuracy. I wanted to compare the variants of a httplistener and a http server. If we had a server installed on all platforms, we wouldn't be worried about cleaning it up. The same is true for a httplistener - we download it on demand in the module and use it in more than one test. Now we have to use the new port for each new test which can cause problems.

SteveL-MSFT:
@iSazonov this would probably be less of a problem if we split the webcmdlets tests into separate scripts for invoke-webrequest and invoke-restmethod although there is still a chance that after one finishes and stops the listener, the OS may not have cleaned up the endpoint before the next one starts and tries to create it on the same port

TravisEz13 (approved):

PR #3885, opened 2017-05-30
Automatically strip Authorization header on redirects

### LINKED ISSUE #2227: Invoke-RestMethod does not strip Authorization Headers
## Steps to reproduce

Attempt to download something using `Invoke-RestMethod` that requires an `Authorization` header AND generates a 302 redirect to the final location of that resource.

One such example is downloading a build artifact from an AppVeyor project.
E.g from the example [here](https://www.appveyor.com/docs/api/samples/download-artifacts-ps/)

```
Invoke-RestMethod -Method Get -Uri "$apiUrl/buildjobs/$jobId/artifacts/$artifactFileName" `
-OutFile $localArtifactPath -Headers @{ "Authorization" = "Bearer $token" }
```
## Expected behavior

The request should succeed like it does on Windows
## Actual behavior

The request fails with the following error

```
Invoke-RestMethod : Response status code does not indicate success: 400 (Authentication information is not given in the correct
format. Check the value of Authorization header.).
At /Users/dave/Desktop/appveyor.ps1:31 char:1
+ Invoke-RestMethod -Method Get -Uri "$apiUrl/buildjobs/$jobId/artifact ...
    + CategoryInfo          : InvalidOperation: (Method: GET, Re...rShell/6.0.0
}:HttpRequestMessage) [Invoke-RestMethod], HttpRequestException
    + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeRestMethodCommand
```

Having debugged this with AppVeyor support it seems they send a 302 redirect to a location in Azure. On Windows, I've verified with Fiddler that the `Authorization` header is stripped which allows this command to succeed. On Mac it would appear the `Authorization` header is not stripped, causing Azure to generate the error.
## Environment data

``` powershell
> $PSVersionTable

CLRVersion

```

dantraMSFT:
Fix #2227

Update Invoke-WebRequest and Invoke-RestMethod cmdlets to strip an Authorization header on redirect.

The FullCLR implementation uses WebRequest to perform the request which silently strips the Authorization header when a redirect occurs. The CoreCLR implementation uses HttpClient to perform the request which does not strip the authorization header. The change explicitly handles the initial redirect, removes the authorization header and submits the request to location in the response.
A new switch, PreserveAuthorizationOnRedirect, disables this handling.

See https://github.com/PowerShell/PowerShell-Docs/pull/1269 for the doc change to document the new switch -PerserveAuthorizationOnRedirect.


src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -48,7 +48,23 @@ public HttpResponseException (string message, HttpResponseMessage response) : ba
     /// </summary>
     public abstract partial class WebRequestPSCmdlet : PSCmdlet
     {
-        #region Abstract Methods
+
+        /// <summary>
+        /// gets or sets the PreserveAuthorizationOnRedirect property
+        /// </summary>
+        /// <remarks>
+        /// This property overrides compatability with web requests on Windows.
```
SteveL-MSFT: Typo: compatibility

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -57,7 +73,7 @@ public abstract partial class WebRequestPSCmdlet : PSCmdlet
         /// <param name="response">Instance of a WebResponse object to be processed</param>
         internal abstract void ProcessResponse(HttpResponseMessage response);
 
-        #endregion Abstract Methods
+#endregion Abstract Methods
```
SteveL-MSFT: Generally, you should not mix formatting changes with functional changes.  Have them as separate PRs to make them more readable and reviewable.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -367,23 +390,50 @@ internal virtual void FillRequestStream(HttpRequestMessage request)
...
+                // if explicit count was provided, reduce it for this redirection.
+                if (WebSession.MaximumRedirection > 0)
+                {
+                    WebSession.MaximumRedirection--;
+                }
+                // recreate the HttpClient with redirection enabled since the first call suppressed redirection
+                using (client = GetHttpClient(false))
+                using (HttpRequestMessage redirectRequest = GetRequest(response.Headers.Location))
+                {
+                    string msg = string.Format("{0}->{1}", request.RequestUri.AbsoluteUri, response.Headers.Location.AbsoluteUri);
```
SteveL-MSFT: You should put this in WebCmdletStrings.resx

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -223,6 +223,43 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         $jsonContent.headers.'User-Agent' | Should Match "WindowsPowerShell"
     }
 
+    # Verifies Invoke-WebRequest with -PreserveAuthorizationOnRedirect preserves
+    # the authorization header whe a 302 redirection
+    It "Validate Invoke-WebRequest with -PreserveAuthorizationOnRedirect preserves the authorization header on 302 redirects" {
```
SteveL-MSFT: Don't we need to support other 3xx redirect codes?

https://en.wikipedia.org/wiki/URL_redirection#HTTP_status_codes_3xx

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -223,6 +223,43 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
...
+        try {
+            Start-HttpListener -AsJob
```
SteveL-MSFT: There's a race condition on Mac where stopping the httplistener and then starting a new one, the OS may not have cleaned up the actual underlying http service so the next start fails as the reservation still exists.  In anycase, you should rebase against upstream/master as this script already starts and stops httplistener for all tests.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -223,6 +223,43 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
...
+
+            $headers = @{"Authorization" = "test"}
+            $response = Invoke-WebRequest -Uri "http://localhost:8080/PowerShell?test=redirect" -Headers $headers -PreserveAuthorizationOnRedirect
```
SteveL-MSFT: Seems like we should test multiple redirects to make sure it's stripped off the first one and subsequent redirects get handled correctly (3 may be sufficient)

SteveL-MSFT: Should also have a test case where `-PreserveAuthorizationOnRedirect` is specified, but no Authorization header was supplied

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -367,23 +390,50 @@ internal virtual void FillRequestStream(HttpRequestMessage request)
...
-            return client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, _cancelToken.Token).GetAwaiter().GetResult();
+            HttpResponseMessage response = client.SendAsync(request, HttpCompletionOption.ResponseHeadersRead, _cancelToken.Token).GetAwaiter().GetResult();
+
+            if (
+                System.Net.HttpStatusCode.Found == response.StatusCode
+                &&
+                stripAuthorization 
+                )
+            {
+                _cancelToken.Cancel();
```
PaulHigin: > _cancelToken.Cancel(); [](start = 15, length = 23)

I am a little confused by this logic.  It looks like an asynchronous call is being cancelled here.  Is there a possibility of a race condition?  Can you add a comment indicating what is being cancelled?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
dantraMSFT: I'll remove it. Not appropriate for release.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
dantraMSFT: Fixed

PaulHigin:
This seems to have a lot of unrelated changes (26 files now changed) now.

SteveL-MSFT:
@dantraMSFT looks like you accidentally pulled in other commits, you need to do a `rebase -i` to clean this up.  Let me know if you need help with this.

dantraMSFT:
fixed

---
In reply to: [306282201](https://github.com/PowerShell/PowerShell/pull/3885#issuecomment-306282201) [](ancestors = 306282201)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -48,6 +48,22 @@ public HttpResponseException (string message, HttpResponseMessage response) : ba
...
+        /// This property overrides compatibility with web requests on Windows.
+        /// On FullCLR (WebRequest), authorization headers are stripped during redirect.
+        /// CoreCLR (HTTPClient) does not have this behavior so web reqeusts that work on
+        /// PowerShell/FullCLR can fail with PowerShell/CoreCLR.  To provide compatibility,
+        /// we'll detect request with an Authorization header and automatically strip
+        /// the header.  This switch turns off this logic for edge cases where the
+        /// authorization header needs to be preserved across redirects.
+        /// </remarks>
+        [Parameter]
+        public virtual SwitchParameter PreserveAuthorizationOnRedirect { get; set; }
```
SteveL-MSFT: You need to make the corresponding doc change in PowerShell/PowerShell-Docs and mention it here to link the two

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -367,13 +394,77 @@ internal virtual void FillRequestStream(HttpRequestMessage request)
...
-        internal virtual HttpResponseMessage GetResponse(HttpClient client, HttpRequestMessage request)
+        // Returns true if the status code is one of the supported redirection codes.
+        static bool IsRedirectCode(HttpStatusCode code)
+        {
+            int intCode = (int) code;
+            return
+            (
+                (intCode >= 300 && intCode < 304)
+                ||
+                intCode == 307
```
SteveL-MSFT: isn't 308 also valid?

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -143,6 +142,125 @@ function GetTestData
...
+function Convert-ResponseRawContent
+{
+    param
+    (
+        [Parameter(Mandatory)]
+        [ValidateNotNull()]
+        [Microsoft.PowerShell.Commands.WebResponseObject] $Response
+    )
+
+    $index = $Response.RawContent.IndexOf("{")
```
SteveL-MSFT: why isn't $response.content | convertfrom-json sufficient?  this code seems unnecessary

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -143,6 +142,125 @@ function GetTestData
...
+        [switch] $UsePost,
+
+        [switch] $PreserveAuthorizationOnRedirect
+    )
+    $result = [PSObject]@{Output = $null; Error = $null; Content = $null}
+
+    try
+    {
+        $headers = @{"Authorization" = "test"}
+        if ($UsePost)
```
SteveL-MSFT: It would be better to have this as a `-testcase` where one of the parameters is $method and you can then pass Get and Post

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -143,6 +142,125 @@ function GetTestData
...
+    }
+    catch
+    {
+        $result.Error = $_
+    }
+
+    return $result
+}
+
+Describe "Invoke-WebRequest redirect tests" -Tags ("Feature", "Redirect") {
```
SteveL-MSFT: Is there a specific reason you want to add a new Describe rather than having your tests part of the existing Invoke-WebRequest tests?

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -143,6 +142,125 @@ function GetTestData
...
+    AfterAll {
+        $null = Stop-HttpListener -Port 8082
+    }
+
+    It "Validates Invoke-WebRequest with -PreserveAuthorizationOnRedirect preserves the authorization header 302 (Found) redirect" {
+        $response = ExecuteRedirectRequest -Uri 'http://localhost:8082/PowerShell?test=redirect&type=Found' -PreserveAuthorizationOnRedirect
+        $response.Error | Should BeNullOrEmpty
+
+        # ensure Authorization header has been preserved.
+        $response.Content.Headers -contains "Authorization" | Should Be $true
```
SteveL-MSFT: should validate the value as well

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -143,6 +142,125 @@ function GetTestData
...
+    }
+
+    It "Validates Invoke-WebRequest preserves the authorization header on multiple redirects." {
+        $response = ExecuteRedirectRequest -Uri "http://localhost:8082/PowerShell?test=redirect&type=Found&multiredirect=true" -PreserveAuthorizationOnRedirect
+
+        $response.Error | Should BeNullOrEmpty
+        # ensure Authorization header was stripped
+        $response.Content.Headers -contains "Authorization" | Should Be $true
```
SteveL-MSFT: validate the value

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -143,6 +142,125 @@ function GetTestData
...
+      @{redirectType = 'RedirectMethod'}
+      @{redirectType = 'TemporaryRedirect'}
+    ) {
+        param($redirectType)
+
+        $response = ExecuteRedirectRequest -Uri "http://localhost:8082/PowerShell?test=redirect&type=$redirectType"
+        $response.Error | Should BeNullOrEmpty
+
+        # ensure user-agent is present (i.e., no false positives )
+        $response.Content.Headers -contains "User-Agent" | Should Be $true
```
SteveL-MSFT: why not validate all the headers are there except authorization?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
SteveL-MSFT: Since you have this specific checks it seems you should have TestCases that go through the different status codes

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: Because I just want the json part of the raw content. The rest is extraneous for my purposes and it is also not valid json.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: I placed the tests in a new describe so I can run them in isolation while developing them.  it turns out appveyor doesn't allow extra tag so I'll be removing it shortly.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: This is only a choice between GET and POST. No other methods are being used.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
dantraMSFT: It's in progress.  Will submit a PR shortly and update the description to reference it.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
dantraMSFT: 308 isn't currently supported in CoreCLR - HttpStatusCode doesn't have a define for it so there's no way for the code to see it.

dantraMSFT: An exhaustive -PreserveAuthroizationOnRedirect won't add any value but if you feel strongly enough, I can make these tests cover all redirections as well.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: Will fix

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: I don't know explicitly what other headers 'should' be there so I don't have a source of truth to compare against what is echoed in the response.  I know the authorization header should or should not be in a specific echoed response so I can verify it.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: Will fix.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: Actually, I can't fix it without fixing ConvertTo-Json.   $request.Headers is a WebHeaderCollection which is a NamedValueCollection (ICollection).  ConvertTo-Json (and Select-Object for that matter) treat it as a collection and would have to special case the type.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
dantraMSFT: Refactored all tests to cover each redirect status code.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
SteveL-MSFT: lines 182 and 186 below only differ by use of `-Method`.  You should either change `$UsePost` to `$Method` or use splatting so there's only a single Invoke-WebRequest line.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
SteveL-MSFT: Looking at your usage below, you pass in the output of Invoke-WebRequest to this helper.  This helper isn't necessary.  Line 189 below can just be:

```powershell
$result.Content = $result.Output.Content | ConvertFrom-Json
```

This is assuming the output you create in httplistener is json (and if not, perhaps it should be).

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -143,6 +142,143 @@ function GetTestData
...
+
+      @{redirectType = 'redirectMethod'; redirectedMethod='GET'}
+      @{redirectType = 'SeeOther'; redirectedMethod='GET'} # Synonym for RedirectMethod
+
+      @{redirectType = 'TemporaryRedirect'; redirectedMethod='GET'}
+      @{redirectType = 'RedirectKeepVerb'; redirectedMethod='GET'} # Synonym for TemporaryRedirect
+    )
+
+    It "Validates Invoke-WebRequest with -PreserveAuthorizationOnRedirect preserves the authorization header on redirect" -TestCases $redirectTests {
+        param($redirectType, $redirectedMethod)
```
SteveL-MSFT: Not entirely sure about the Pester syntax, but do you have to have $redirectedMethod here since you don't use it?  Or does Pester complain?

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: $response.Content is a byte array, not string ($result.Output.Content is this byte array).  $response.RawContent has a few lines of text inserted by the underlying API followed by the json form of the original request; inserted by the httplistener.  The purpose of the function is to extract the json from the RawContent and return it explicitly since it is needed to verify the incoming request.

dantraMSFT: I'll see if I can drop the unused $redirectedMethod params.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: I'll update the function to use a ValidateSet and default to GET. I don't want to accept arbitrary methods  since I'm explicitly testing the POST->GET logic on redirects.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: I'm remove the unused parameters.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: The listener does add the output as json and I've even tried setting ContentType and ContentEncoding with no change in behavior. It appears it may be an issue on the client side. For now, I'll update the function to detect byte[] versus  string and convert the byte[] to string before calling ConvertFrom-Json so it will work with arbitrary servers.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -143,6 +142,121 @@ function GetTestData
...
+    }
+    catch
+    {
+        $result.Error = $_
+    }
+
+    return $result
+}
+
+Describe "Invoke-WebRequest redirect tests" -Tags "Feature" {
```
SteveL-MSFT: I thought you were going to put these tests into an existing Describe?

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
SteveL-MSFT: Also, do we need this capability with invoke-restmethod?

SteveL-MSFT (changes requested):

dantraMSFT:
Please squash when merged.

SteveL-MSFT (approved):
LGTM

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -217,6 +240,10 @@ internal virtual HttpRequestMessage GetRequest(Uri uri)
                     }
                     else
                     {
+                        if (stripAuthorization && entry.Key == HttpKnownHeaderNames.Authorization.ToString())
+                        {
```
lzybkr: We try to avoid `==` with string comparisons and instead use `string.Equals(a,b, StringComparison.Ordinal)` or the appropriate comparison - this makes it clearer that the author considered culture and case when writing the code.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -48,6 +48,22 @@ public HttpResponseException (string message, HttpResponseMessage response) : ba
...
     public abstract partial class WebRequestPSCmdlet : PSCmdlet
     {
+
+        /// <summary>
+        /// gets or sets the PreserveAuthorizationOnRedirect property
+        /// </summary>
+        /// <remarks>
+        /// This property overrides compatibility with web requests on Windows.
+        /// On FullCLR (WebRequest), authorization headers are stripped during redirect.
+        /// CoreCLR (HTTPClient) does not have this behavior so web reqeusts that work on
```
lzybkr: >reqeusts [](start = 68, length = 8)

Spelling.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -410,7 +511,7 @@ protected override void ProcessRecord()
                             WriteVerbose(linkVerboseMsg);
                         }
 
-                        using (HttpRequestMessage request = GetRequest(uri))
+                        using (HttpRequestMessage request = GetRequest(uri, false))
```
lzybkr: >false [](start = 76, length = 5)

Just a suggestion - we like to use named parameters when using bool constants - it makes it much easier to understand the code, so this could be `GetResult(uri, stripAuthorization: false)`.

lzybkr (changes requested):
There were a couple of minor issues I'd like to see addressed before I merge.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
dantraMSFT: Will fix.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
dantraMSFT: Will change.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
dantraMSFT: Will fix

dantraMSFT:
@lzybkr Would you review my changes so I can get this merged before I go OOF.
Thanks.

lzybkr (approved):

lzybkr:
I'll squash, but I'll ask you to write the commit message that follows our [guidelines](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#common-engineering-practices) - you can just comment here or squash the changes yourself and update the PR, either way is fine.

dantraMSFT:
Description updated.

Issue #2334, opened 2016-09-21
Transcription for the host stops, when ANY runspace in the host get closed


vors:
## Steps to reproduce

``` powershell
Start-Transcript -Path foo.txt -Force
$rs = [system.management.automation.runspaces.runspacefactory]::CreateRunspace()
$rs.open()
$ps = [System.Management.Automation.PowerShell]::Create()
$ps.Runspace = $rs
$ps.Dispose()
"Before Dispose"
$rs.Dispose()
"After Dispose"
Stop-Transcript
```
## Expected behavior

No errors, the whole transcript (including "After Dispose") should be in the `foo.txt` file
## Actual behavior

```
Stop-Transcript : An error occurred stopping transcription: The host is not currently transcribing.
At line:1 char:1
+ Stop-Transcript
+ ~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Stop-Transcript], PSInvalidOperationException
    + FullyQualifiedErrorId : InvalidOperation,Microsoft.PowerShell.Commands.StopTranscriptCommand
```

"After Dispose" is not in the `foo.txt` file
## Environment data

Windows.

``` powershell
> $PSVersionTable
Name                           Value
----                           -----
PSEdition                      Core
PSVersion                      6.0.0-alpha
CLRVersion
GitCommitId                    v6.0.0-alpha.10
WSManStackVersion              3.0
BuildVersion                   3.0.0.0
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

 Also verfied in inbox

```
Name                           Value
----                           -----
PSVersion                      5.1.14393.187
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.14393.187
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

cc @adityapatwardhan @LeeHolmes @PaulHigin

Fitzroy87:
@vors - does this issue account for the undesired behaviour documented in this uservoice post? https://windowsserver.uservoice.com/forums/301869-powershell/suggestions/14745750-ps-5-0-bug-transcript-stops-after-restarting-a-re

vors:
[...]

PR #3896, opened 2017-05-31
Only stop transcription when all runspaces are closed

chunqingchen:
Fix issue #2334

Summary of the issue:

If user create any runspace within transcription and later close the runspace within the transcription, transcription get closed

Root cause of the issue:

When any runspace get closed, StopAllTranscribing() get called and the transcription get closed
Fix:

Check and make sure the we are closing the last runspace to stop transcription. otherwise user still need to explicitly call stop-transcript to close the transcription.

### DIFF
--- a/src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
+++ b/src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
@@ -863,12 +863,24 @@ namespace System.Management.Automation.Runspaces
                 }
 
                 if ((hostRunspace == null) || (this == hostRunspace))
-                {
+                {                   
+                    // We should close transcripting only if we are closing the last opened runspace.
+                    foreach (Runspace runspace in RunspaceList)
+                    {
+                        // At this stage, the last opened runspace should be at closing state.
+                        if (runspace.RunspaceStateInfo.State == RunspaceState.Opened)
+                        {
+                            return;
+                        }
+                    }
+
                     PSHostUserInterface host = executionContext.EngineHostInterface.UI;
                     if (host != null)
                     {
                         host.StopAllTranscribing();
                     }
+
+                    AmsiUtils.Uninitialize();
                 }
             }
 
@@ -913,10 +925,7 @@ namespace System.Management.Automation.Runspaces
 
             //Log engine lifecycle event.
             MshLog.LogEngineLifecycleEvent(_engine.Context, EngineState.Stopped);
-
-            // Uninitialize the AMSI scan interface
-            AmsiUtils.Uninitialize();
-
+  
             //All pipelines have been canceled. Close the runspace.
             _engine = null;
             _commandFactory = null;
--- a/test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
@@ -106,4 +106,34 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
         $expectedError = "CannotStartTranscription,Microsoft.PowerShell.Commands.StartTranscriptCommand"
         ValidateTranscription -scriptToExecute $script -outputFilePath $null -expectedError $expectedError
     }
+    It "Transcription should remain active if other runspace in the host get closed" {
+        try{
+            $ps = [powershell]::Create()
+            $ps.addscript("Start-Transcript -path $transcriptFilePath").Invoke()
+            $ps.addscript('$rs = [system.management.automation.runspaces.runspacefactory]::CreateRunspace()').Invoke()      
+            $ps.addscript('$rs.open()').Invoke()
+            $ps.addscript('$rs.Dispose()').Invoke()
+            $ps.addscript('Write-Host "After Dispose"').Invoke()
+            $ps.addscript("Stop-Transcript").Invoke()
+            } finally {
+                if ($ps -ne $null) {
+                    $ps.Dispose()
+                }
+            }
+
+        
+        Test-Path $transcriptFilePath | Should be $true
+        $transcriptFilePath | Should contain "After Dispose"
+    }
+
+    It "Transcription should be closed if the only runspace gets closed" {
+        $powerShellPath = [System.Diagnostics.Process]::GetCurrentProcess().Path
+        $powerShellCommand = $powerShellPath + ' "start-transcript $transcriptFilePath; Write-Host ''Before Dispose'';"'
+        Invoke-Expression $powerShellCommand
+        
+        Test-Path $transcriptFilePath | Should be $true
+        $transcriptFilePath | Should contain "Before Dispose"
+        $transcriptFilePath | Should contain "Windows PowerShell transcript end"
+    }
+
 }


src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
```diff
@@ -863,7 +863,13 @@ private void DoCloseHelper()
                 }
 
                 if ((hostRunspace == null) || (this == hostRunspace))
-                {
+                {                   
+                    // We should not close transcripting if we are not under default runspace.
```
SteveL-MSFT: I was confused by the code and the comment as I misread the comment initially.  I think you should change the comment to avoid a double negative.

```
We should close transcripting only if we are running in default runspace
```

SteveL-MSFT (changes requested):

chunqingchen:
@SteveL-MSFT your comment is resolved.

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
chunqingchen: resolved

adityapatwardhan:
I tried fixing this in Windows Powershell 5.1, but it broke with the following repro. Powershell crashes. Try running the following with a debugger attached.

```Powershell
powershell.exe -c "start-transcript"
```

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
```diff
@@ -863,7 +863,13 @@ private void DoCloseHelper()
                 }
 
                 if ((hostRunspace == null) || (this == hostRunspace))
-                {
+                {                   
+                    // We should close transcripting only if we are running in default runspace.
+                    if (executionContext.CurrentRunspace != null && DefaultRunspace != null && executionContext.CurrentRunspace.Id != DefaultRunspace.Id)
+                    {
+                        return;
```
SteveL-MSFT: We need a test case to cover the implicit closing of transcription

SteveL-MSFT (changes requested):

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
```diff
@@ -863,7 +863,13 @@ private void DoCloseHelper()
                 }
 
                 if ((hostRunspace == null) || (this == hostRunspace))
-                {
+                {                   
+                    // We should close transcripting only if we are running in default runspace.
+                    if (executionContext.CurrentRunspace != null && DefaultRunspace != null && executionContext.CurrentRunspace.Id != DefaultRunspace.Id)
```
PaulHigin: This is tricky because DefaultRunspace is a thread static property and so there can be multiple DefaultRunspaces for multiple threads.  It is really hard to know when all runspaces are closed.  We do now keep a list of all (non-disposed) local runspaces.  See Connection.cs https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/hostifaces/Connection.cs#L799
You could search through this list and if there is only one runspace left in the "open" state (meaning this is the last runspace) then do the stop transcribing.
This list only shows runspaces that have not yet been disposed.  But a runspace can be non-open and still be in the list if it was not disposed, so you need to check the state of each runspace.

PaulHigin (changes requested):

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
PaulHigin: Also I just realized that the "AmsiUtils.Uninitialize()" call should be done at the last runspace being closed as well.  Can you please move that call into this logic too?

chunqingchen:
@adityapatwardhan I tried your repro and powershell isn't crashing.

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
chunqingchen: resolved

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
chunqingchen: resolved.

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
```diff
@@ -106,4 +106,51 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
...
+                    $ps.Dispose()
+                }
+            }
+
+        
+        Test-Path $transcriptFilePath | Should be $true
+        $transcriptFilePath | Should contain "After Dispose"
+    }
+
+    It "Transcription should remain active if other runspace in the host get closed" {
```
SteveL-MSFT: Isn't this exactly the same as above?

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
chunqingchen: whoops~ ! deleted.

chunqingchen:
@PaulHigin @SteveL-MSFT your comments are resolved.

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
```diff
@@ -863,12 +863,24 @@ private void DoCloseHelper()
...
+                        }
+                    }
+
                     PSHostUserInterface host = executionContext.EngineHostInterface.UI;
                     if (host != null)
                     {
                         host.StopAllTranscribing();
                     }
+
+                    AmsiUtils.Uninitialize();
```
PaulHigin: AmsiUtils.Uninitialize() is still being called at line 930 below.  Please remove that call so that uninitialize only happens when the last runspace closes.

PaulHigin (changes requested):

adityapatwardhan (changes requested):

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
```diff
@@ -106,4 +106,51 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
...
+                }
+            }
+
+        
+        Test-Path $transcriptFilePath | Should be $true
+        $transcriptFilePath | Should contain "After Dispose"
+    }
+
+    It "Transcription should be closed if the only runspace gets closed" {
+        powershell "start-transcript $transcriptFilePath; Write-Host 'Before Dispose';"
```
SteveL-MSFT: Even on Linux, it will default to the installed one and not the running one.  See https://github.com/PowerShell/PowerShell/blob/master/test/powershell/Host/ConsoleHost.Tests.ps1#L12

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
```diff
@@ -106,4 +106,51 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
...
+        
+        Test-Path $transcriptFilePath | Should be $true
+        $transcriptFilePath | Should contain "Before Dispose"
+        $transcriptFilePath | Should contain "Windows PowerShell transcript end"
+    }
```
chunqingchen: resolved

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
chunqingchen: resolved

chunqingchen:
@PaulHigin @SteveL-MSFT @adityapatwardhan your comments are resolved.

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
TravisEz13: This problem is still in the file the image at the end of the review indicates there is not trailing newline.

SteveL-MSFT (approved):

adityapatwardhan (approved):
After the newline at the end of the file is fixed.

PaulHigin (approved):
LGTM

chunqingchen:
@TravisEz13 Hi Travis, please merge the pr

Issue #3738, opened 2017-05-09
ScriptsToProcess not honored when Import-Module is passed -Version param


Matteo-T:
Steps to reproduce
------------------
- Install SqlServer module from PSGallery (I doubt the issue is specific to my module, though)
- Open PowerShell (full .Net, I'm on Win10 using the PowerShell that comes with the OS)
   PowerShell -executionpolicy bypass
- Import-Module SqlServer -Verbose
  (observe that SqlServer.ps1, listed in the SqlServer.psd1 manifest file, is loaded during import)
- Close PowerShell
- Restart PowerShell (same as above)
- Import-Module SqlServer -Verbose -Version 21.0.17099

Expected behavior
-----------------
The SqlServer.ps1 script is loaded

Actual behavior
---------------
The SqlServer.ps1 script is NOT loaded

Environment data
----------------
```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      5.1.16192.0
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.16192.0
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

iSazonov:
@Matteo-T Thanks for your report! Could you check on latest PowerShell Core?

Matteo-T:
My module does not work on PowerShell Core, so I'm not sure how that would help... I'm not sure I'll have time to create a dummy repro (i.e. a dummy module) that works on PowerShell Core any time soon, but if I have to...

megamorf:
This repo is for PowerShell Core only. Please report regular PowerShell issues on the uservoice website:

https://windowsserver.uservoice.com/forums/301869-powershell

iSazonov:
@Matteo-T Please test after PowerShell Core Beta1 release (in the coming days).
And yes, we can report on UserVoice.

Matteo-T:
Sorry, I did not know this was for "Core" only. I was told to come here, but I guess I was told incorrectly. I'll try the UserVoice thing...

joeyaiello:
@megamorf @Matteo-T for bugs that are almost certainly applicable to both editions of PowerShell, it's perfectly fine to file them here. I suspect that this one repros on Core 6 just fine, and given the importance of `Import-Module` we should probably get it fixed (assuming it *does* repro). 

[...]

PR #3897, opened 2017-05-31
Fixed ScriptsToProcess when -Version param is used in Import-Module

anmenaga:
This is fix for #3738 "ScriptsToProcess not honored when Import-Module is passed -Version param".

The bug was cased by existing check in LoadModule functions:

> If  MinimumVersion/RequiredVersion/MaximumVersion has been specified, then only try to process manifest modules...
> If the -Version flag was specified, don't look for non-manifest modules.

When loading usual module files, this check is valid;
However, same LoadModule function is used to load ScriptsToProcess - and this when this check was 
triggered and ScriptsToProcess were not executed.

The fix is to work around that condition for ScriptsToProcess in the same way it is currently done for NestedModules - temporary reset the version info for the duration of processing ScriptsToProcess / NestedModules.

Test results before the fix:
befoerfix

Test results after the fix:
afterfix

### DIFF
--- a/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
@@ -2769,37 +2769,57 @@ namespace Microsoft.PowerShell.Commands
 
             if (scriptsToProcess != null)
             {
-                foreach (string scriptFile in scriptsToProcess)
+                Version savedBaseMinimumVersion = BaseMinimumVersion;
+                Version savedBaseMaximumVersion = BaseMaximumVersion;
+                Version savedBaseRequiredVersion = BaseRequiredVersion;
+                Guid? savedBaseGuid = BaseGuid;
+
+                try
                 {
-                    bool found = false;
-                    PSModuleInfo module = LoadModule(scriptFile,
-                        moduleBase,
-                        string.Empty, // prefix (-Prefix shouldn't be applied to dot sourced scripts)
-                        null,
-                        ref options,
-                        manifestProcessingFlags,
-                        out found);
+                    BaseMinimumVersion = null;
+                    BaseMaximumVersion = null;
+                    BaseRequiredVersion = null;
+                    BaseGuid = null;
 
-                    // If we're in analysis, add the detected exports to this module's
-                    // exports
-                    if (found && (ss == null))
+                    foreach (string scriptFile in scriptsToProcess)
                     {
-                        foreach (string detectedCmdlet in module.ExportedCmdlets.Keys)
-                        {
-                            manifestInfo.AddDetectedCmdletExport(detectedCmdlet);
-                        }
+                        bool found = false;
+                        PSModuleInfo module = LoadModule(scriptFile,
+                            moduleBase,
+                            string.Empty, // prefix (-Prefix shouldn't be applied to dot sourced scripts)
+                            null,
+                            ref options,
+                            manifestProcessingFlags,
+                            out found);
 
-                        foreach (string detectedFunction in module.ExportedFunctions.Keys)
+                        // If we're in analysis, add the detected exports to this module's
+                        // exports
+                        if (found && (ss == null))
                         {
-                            manifestInfo.AddDetectedFunctionExport(detectedFunction);
-                        }
+                            foreach (string detectedCmdlet in module.ExportedCmdlets.Keys)
+                            {
+                                manifestInfo.AddDetectedCmdletExport(detectedCmdlet);
+                            }
 
-                        foreach (string detectedAlias in module.ExportedAliases.Keys)
-                        {
-                            manifestInfo.AddDetectedAliasExport(detectedAlias,
-                                module.ExportedAliases[detectedAlias].Definition);
+                            foreach (string detectedFunction in module.ExportedFunctions.Keys)
+                            {
+                                manifestInfo.AddDetectedFunctionExport(detectedFunction);
+                            }
+
+                            foreach (string detectedAlias in module.ExportedAliases.Keys)
+                            {
+                                manifestInfo.AddDetectedAliasExport(detectedAlias,
+                                    module.ExportedAliases[detectedAlias].Definition);
+                            }
                         }
                     }
+                }
+                finally
+                {
+                    BaseMinimumVersion = savedBaseMinimumVersion;
+                    BaseMaximumVersion = savedBaseMaximumVersion;
+                    BaseRequiredVersion = savedBaseRequiredVersion;
+                    BaseGuid = savedBaseGuid;
                 }
             }
 
--- a/test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
@@ -38,7 +38,6 @@ Describe "Import-Module with ScriptsToProcess" -Tags "CI" {
 
     AfterAll {
         Pop-Location
-        #Remove-Item $moduleRootPath -Recurse -Force -ErrorAction SilentlyContinue
     }
 
     BeforeEach {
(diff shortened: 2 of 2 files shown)


src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -2768,38 +2768,58 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
...
-                        foreach (string detectedFunction in module.ExportedFunctions.Keys)
+                        // If we're in analysis, add the detected exports to this module's
+                        // exports
+                        if (found && (ss == null))
                         {
-                            manifestInfo.AddDetectedFunctionExport(detectedFunction);
-                        }
+                            foreach (string detectedCmdlet in module.ExportedCmdlets.Keys)
+                            {
+                                manifestInfo.AddDetectedCmdletExport(detectedCmdlet);
```
adityapatwardhan: AddDetected* methods from PSModuleInfo in most cases seem to be called multiple times using a for loop. Will it make sense to add an overload which accepts IEnumerable?

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -22,3 +22,52 @@
...
+    It "Verify ScriptsToProcess are executed for top-level module when -Version is specified" {
+        Import-Module .\module1.psd1 -Version 1.0
+        Get-Content out.txt | Should Be '1'
+    }
+
+    It "Verify ScriptsToProcess are executed for top-level and nested module when -Version is specified" {
+        Import-Module .\module2.psd1 -Version 1.0
+        Get-Content out.txt | Should Be '21'
+    }
+}
```
adityapatwardhan: Add new line at the end.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -22,3 +22,52 @@
...
+
+        "1 | Out-File out.txt -Append -NoNewline" | Out-File script1.ps1
+        "2 | Out-File out.txt -Append -NoNewline" | Out-File script2.ps1
+        New-ModuleManifest module1.psd1 -ScriptsToProcess script1.ps1
+        New-ModuleManifest module2.psd1 -ScriptsToProcess script2.ps1 -NestedModules module1.psd1
+    }
+
+    AfterAll {
+        Pop-Location
+        #Remove-Item $moduleRootPath -Recurse -Force -ErrorAction SilentlyContinue
```
adityapatwardhan: Remove commented line.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -22,3 +22,52 @@
...
+        New-Item out.txt -ItemType File -Force | Out-Null
+    }
+
+    AfterEach {
+        $m = @('module1','module2','script1','script2')
+        remove-module $m -Force -ErrorAction SilentlyContinue
+        Remove-Item out.txt -Force -ErrorAction SilentlyContinue
+    }
+    
+    It "Verify ScriptsToProcess are executed for top-level module" {
```
adityapatwardhan: You may want to take a look at TestCases parameter of Pester It.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -22,3 +22,52 @@
         (Get-Module -Name $moduleName).Name | Should Be $moduleName
     }
 }
+
+Describe "Import-Module with ScriptsToProcess" -Tags "CI" {
+    
+    BeforeAll {
+        $moduleRootPath = Join-Path $TestDrive 'TestModules'
+        New-Item $moduleRootPath -ItemType Directory -Force | Out-Null
+        Push-Location $moduleRootPath
```
adityapatwardhan: Why do you need to push location? Can you use full paths for all cmdlets like New-Item, New-ModuleManifest etc?

adityapatwardhan (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
anmenaga: Fixed.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
anmenaga: Fixed.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
anmenaga: Thanks for suggestion; I did give it a try. I believe that semantic difference between these tests is larger than what TestCases parameter is aimed for; plus I don't like that TestCases parameter has worse logging (same test name repeated several times vs individual, descriptive test names in case of individual 'It's).

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
anmenaga: I chose push location specifically because I didn't want to deal with full paths. :)

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -2768,38 +2768,58 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
...
+                    BaseGuid = null;
+
+                    foreach (string scriptFile in scriptsToProcess)
                     {
-                        foreach (string detectedCmdlet in module.ExportedCmdlets.Keys)
-                        {
-                            manifestInfo.AddDetectedCmdletExport(detectedCmdlet);
-                        }
+                        bool found = false;
+                        PSModuleInfo module = LoadModule(scriptFile,
```
anmenaga: That is actually existing code; I didn't change it; looks like Github's diff'er has room for improvement :)  - see picture below.
Anyway, there are multiple LoadModule calls like this one throughout the file, so if refactoring is to be done - it has to be done to all of them - and that is out of scope of this bugfix.

diff

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
anmenaga: Maybe. But as previous comment - this would be refactoring of unrelated code - out of scope of this bug fix.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
adityapatwardhan: You can have custom test case names for individual test cases. Have look at: https://github.com/PowerShell/PowerShell/blob/02b5f357a20e6dee9f8e60e3adb9025be3c94490/test/powershell/engine/ParameterBinding/NullableBooleanDCR.Tests.ps1

Here the parameters will be parameter splat for Import-Module, and expected result, test name.

adityapatwardhan (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
anmenaga: That is actually cool; thank you!
I've updated tests.

anmenaga: Good idea; I've updated tests.

adityapatwardhan (approved):

PR #3900, opened 2017-05-31
Added functionality to load Windows Powershell Modules

adityapatwardhan:
A lot of binary modules written for Windows Powershell (Full cLR) use GAC for their assemblies.
CoreCLR does not have a GAC and hence fails to locate the assemblies.

Added functionality to search the well known location of GAC on Windows to load the assemblies.
This functionality can be opted-in by adding a key-value to PowershellProperties.json file at $PSHOME.
The key-value pair for enabling GAC loading is: "AllowGacLoading" : "true"

### DIFF
--- a/src/System.Management.Automation/engine/ExecutionContext.cs
+++ b/src/System.Management.Automation/engine/ExecutionContext.cs
@@ -1318,7 +1318,6 @@ namespace System.Management.Automation
             }
         }
 
-
         [SuppressMessage("Microsoft.Reliability", "CA2001:AvoidCallingProblematicMethods", MessageId = "System.Reflection.Assembly.LoadWithPartialName")]
         [SuppressMessage("Microsoft.Reliability", "CA2001:AvoidCallingProblematicMethods", MessageId = "System.Reflection.Assembly.LoadFrom")]
         internal static Assembly LoadAssembly(string name, string filename, out Exception error)
@@ -1327,14 +1326,11 @@ namespace System.Management.Automation
 
             Assembly loadedAssembly = null;
             error = null;
-
             string fixedName = null;
             if (!String.IsNullOrEmpty(name))
             {
                 // Remove the '.dll' if it's there...
-                fixedName = name.EndsWith(".dll", StringComparison.OrdinalIgnoreCase)
-                                ? Path.GetFileNameWithoutExtension(name)
-                                : name;
+                fixedName = GetAssemblyFixedName(name);
 
                 var assemblyString = Utils.IsPowerShellAssembly(fixedName)
                                          ? Utils.GetPowerShellAssemblyStrongName(fixedName)
@@ -1346,7 +1342,21 @@ namespace System.Management.Automation
                 }
                 catch (FileNotFoundException fileNotFound)
                 {
-                    error = fileNotFound;
+                    if(Platform.IsWindows && ConfigPropertyAccessor.Instance.GetAllowGacLoading())
+                    {
+                        // Try to load dependencies from GAC.
+                        loadedAssembly = LoadAssemblyFromGac(name);
+
+                        if(loadedAssembly == null)
+                        {
+                            error = fileNotFound;
+                            return null;
+                        }
+                    }
+                    else
+                    {
+                        error = fileNotFound;
+                    }
                 }
                 catch (FileLoadException fileLoadException)
                 {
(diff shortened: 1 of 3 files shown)


daxian-dbw:
Talked to @adityapatwardhan offline. The GAC probing logic should be put in AssemblyLoadContext and hook up with the `Resolving` event of the default loader. `Assembly.Load` will trigger the `Resolving` event if the default loader cannot resolve the assembly request. Implicit assembly loading is also taken care of by `Assembly.Load` in this way.

src/System.Management.Automation/engine/PropertyAccessor.cs
```diff
@@ -681,6 +697,15 @@ internal override void SetDefaultSourcePath(string defaultPath)
         }
 
         /// <summary>
+        /// Allow loading assemblies by searching in GAC.
+        /// </summary>
+        /// <returns>Always true, as this is for Full CLR.</returns>
```
daxian-dbw: We don't need to do this for FullCLR. Assembly.Load already takes care of GAC in FullCLR.

adityapatwardhan: I had two options.. 
1. Add it to the interface, hence need to implement it in RegistryAccessor and JsonConfigFileAccessor.
2. Implement it as a static method on JsonConfigFileAccessor and break the strategy pattern.

I ended up going with number 1.

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1429,6 +1439,88 @@ internal static Assembly LoadAssembly(string name, string filename, out Exceptio
...
+                    return null;
+                }
+            }
+
+            var fileToSearch = fixedName + ".dll";
+            var fileList = Directory.GetFiles(gacPath.ToString(), fileToSearch, System.IO.SearchOption.AllDirectories);
+
+            if(fileList.Length > 0)
+            {
+                // GAC only has 1 file in the folder, so we can pick index 0.
```
anmenaga: Several versions of same assembly?
What if a binary module references v2 of an assembly, when GAC has v1 and v2 of that assembly installed?

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1429,6 +1439,88 @@ internal static Assembly LoadAssembly(string name, string filename, out Exceptio
...
+
+        // Search the GAC for the assembly.
+        // First try MSIL and then try the GAC_<PROCESS_ARCH>
+        private static string GetValidGacPath(string assemblyName)
+        {
+            // Get name of assembly without '.dll'. This is the folder name in GAC.
+            // In some cases '.dll' might not be present at the end.
+            string fixedName = GetAssemblyFixedName(assemblyName);
+
+            string winDirPath = Environment.GetEnvironmentVariable("WINDIR");
```
SteveL-MSFT: seems like this may be called often enough to cache

src/System.Management.Automation/engine/PropertyAccessor.cs
```diff
@@ -318,6 +324,16 @@ internal override void SetDefaultSourcePath(string defaultPath)
...
         }
 
+        /// <summary>
+        /// Allow loading assemblies by searching in GAC.
+        /// </summary>
+        /// <returns>Boolean indicating whether assemblies should be loaded from the GAC</returns>
+        internal override bool GetAllowGacLoading()
+        {
+            string fileName = Path.Combine(psHomeConfigDirectory, configFileName);
+            return ReadValueFromFile<bool>(fileName, "AllowGacLoading");
```
SteveL-MSFT: @joeyaiello did we agree to make this opt-out vs opt-in?  I think that for most customers, they won't know to add a setting to a config file and will give up on PSCore6.  Perhaps we should default to opt-out and document this is best effort?

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -22,3 +22,27 @@
...
 }
+
+Describe "Import-Module for Binary Modules in GAC" -Tags 'CI' {
+
+    BeforeAll {
+        ##To enable loading assemblies from GAC, we need to add an entry to PowershellProperties.json
+        $PropertiesFilePath = Join-Path $pshome 'PowershellProperties.json'
+        $savePSProperties = Get-Content $PropertiesFilePath -Raw
+        $PSProperties = ConvertFrom-Json $savePSProperties
+        $PSProperties | Add-Member -MemberType NoteProperty -Name 'AllowGacLoading' -Value 'true'
```
SteveL-MSFT: there should be a test case to validate the default config and behavior

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -22,3 +22,27 @@
...
+    }
+
+    AfterAll {
+        #Reset the PowershellProperties.json file.
+        $savePSProperties | Out-File $PropertiesFilePath -Force -ErrorAction SilentlyContinue
+        Remove-Module PSScheduledJob -Force -ErrorAction SilentlyContinue
+    }
+
+    It "Load PSScheduledJob from Windows Powershell Modules folder" -Skip:(-not $IsWindows) {
+        $modulePath = Join-Path $env:windir "System32/WindowsPowershell/v1.0/Modules/PSScheduledJob"
```
SteveL-MSFT: can we validate the x86 path as well?

SteveL-MSFT (changes requested):

src/System.Management.Automation/engine/PropertyAccessor.cs
adityapatwardhan: @SteveL-MSFT @joeyaiello I believe it was discussed to be opt-in due to performance implications. From what I see, the performance difference is not noticeable. So should I change it to opt-out?

joeyaiello: Let's push the issue while we're in beta by making it default. I'd rather people complained about performance than have no one discover this feature. We can always roll it back out and do strong documentation later. 

So yeah, I say the default should be on, and then we should get customer feedback that tells us to turn it off.

SteveL-MSFT: Agree, we can revisit based on feedback

adityapatwardhan:
Closing this PR and will start a new one for changes in AssemblyLoadContext.

Issue #2321, opened 2016-09-20
Ctrl+C doesn't work as expected with SSH Remoting


PaulHigin:
In PowerShell Ctrl+C should stop a running command.  But in an SSH interactive remote session Ctrl+C terminates the SSH process and closes the session.  This behavior is the same for both Windows and Linux clients.

This is happening because both PowerShell and SSH processes are attached to the console and so they both get the SIGINT signal.  PowerShell sends a PSRP stop signal over the channel as expected but the SSH process also handles the signal and terminates the process which closes the remote session.
## Steps to reproduce

``` powershell
PS> Enter-PSSession -HostName <hostName> -UserName <userName>
[<hostName>]: PS > 1..100 % { sleep 1; "Output $_" }
Output 1
Output 2
# While script is running type Ctrl+C to stop it
```
## Expected behavior

Running command stops and remote session prompt is displayed.
## Actual behavior

Session is ended with error:

``` powershell
System.IO.IOException: Pipe is broken
```

Also the exception is unhandled causing PowerShell process to crash.
## Environment data

``` powershell
Name                           Value
----                           -----
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
SerializationVersion           1.1.0.1
PSRemotingProtocolVersion      2.3
BuildVersion                   3.0.0.0
WSManStackVersion              3.0
PSVersion                      6.0.0-alpha
CLRVersion
GitCommitId                    v6.0.0-alpha.10-8-gdd2394b1e6088f6c82054f5cae5520e532010fd4-dirty
PSEdition                      Core
```

PaulHigin:
This has been fixed as part of the work for issue #2827 for a Windows client.  When creating the ssh.exe process we can use the CREATE_NEW_PROCESS_GROUP process creation flag which prevents the console Ctrl+C signal from propagating to that process.  But there is no support for this in managed code used on Linux clients so it is not fixed there yet.

vors:
I'm hitting this problem.
Would be really useful to get fix for unix.

```
Unhandled Exception: System.IO.IOException: Broken pipe
   at System.IO.UnixFileStream.CheckFileCall(Int64 result, Boolean ignoreNotSupported)
   at System.IO.UnixFileStream.WriteNative(Byte[] array, Int32 offset, Int32 count)
   at System.IO.UnixFileStream.FlushWriteBuffer()
   at System.IO.UnixFileStream.Flush(Boolean flushToDisk)
[...]

PR #3901, opened 2017-05-31
Linux platform SSH process creation for remoting

PaulHigin:
Update:
--------
The technique is to now use sigaction to create an SSH client process that ignores SIGINT signals.  It remains in the PowerShell process group and has access to the terminal, which means it can prompt on the command line as before.  So SSH_ASKPASS dependency is not needed.

-------

This change addresses Issue #2321.  The original idea for this fix was to follow what I did in Windows and create the SSH process in its own process group.  That way a Ctrl+C typed signal in PowerShell (the parent process) wouldn't be handled by the child SSH process (causing it to end), and instead would allow PowerShell to handle it through its protocol (PSRP).

But this ended up disabling the /dev/tty terminal for the SSH process so it could not prompt the user for a password, if that was how SSH was set up.

It turns out that SSH will detect if there is no tty available and fall back to using SSH_ASKPASS program if available.  This will happen if the SSH child process is created in its own session.  So this change does that, it creates the SSH process (used for SSH remoting) in its own session.

The downside to this is that now the SSH_ASKPASS program needs to be installed if we want to support any SSH prompting.  We will need to make sure it gets installed with Enable-SSHRemoting (a separate issue/work item).

Creating a Linux process in its own session does not seem like a common need and so I felt it was not necessary to add this functionality to DotNet CoreFx.  Instead I pulled some of the UNIX process creation code over to PowerShell and added the ability to create in a separate session.  So this change has a fair amount of native and platform interop code.

I will update the license document to indicate that PowerShell is using this DotNet CoreFx code under the MIT license.


daxian-dbw:
@PaulHigin Please add reviewers for your PR and assign it to an appropriate maintainer.

iSazonov:
I leave a [comment in the original Issue](https://github.com/PowerShell/PowerShell/issues/2321#issuecomment-305687171)

PaulHigin:
@iSazonov Please put comments here rather than providing a link.  The SSH -t option to create a pseudo tty prevents using subsystems which is needed to host PowerShell for SSH remoting.  So it won't work here as far as I can tell.  If there was some way to create and use the child SSH process such that it is not part of the process group (so that it doesn't get Ctrl+C signals) but still can use a tty for user prompt, then that would be great.  But I don't currently see a way to do this.  However, I am definitely open to any ideas as this would be preferable to using SSH_ASKPASS!

iSazonov:
@PaulHigin I don't read in depth but it is interesting https://unix.stackexchange.com/questions/266866/how-to-prevent-ctrlc-to-break-ssh-connection

iSazonov:
Oh, I think about a remote session all the time, but the problem is on the local side, isn't it? If so there is a simple test:
1. run Bash
2. run stty intr undef
3. run PowerShell and test Ctrl-C in remote connection

If this works well we can make the same internally in PowerShell - disable ISIG by means of [TERMIOS(3)](http://man7.org/linux/man-pages/man3/termios.3.html) at startup time.

PaulHigin:
This looks interesting, but I don't understand what the fix would be.  I don't want to disable the ISIG for the PowerShell /dev/tty but instead want to keep the child SSH process from getting it.  I only want PowerShell to handle it.  It is not clear to me that disabling ISIG via termios can be done per process.  But if so that would certainly solve our problem!

BTW I did look at your previous link as a way to turn off ISIG processing on the server side, but I wasn't able to get it to work.  "set -m"  didn't seem to do anything.

PaulHigin:
Hmm, sigaction() looks interesting.  It should be possible to suppress a signal on a specific process.  I'll look into it.

PaulHigin:
Sigaction() works great and obviates the need for SSH_ASKPASS since the SSH process can be created in its parent group/session but still be made to ignore interrupt signals (SIGINT).

I have updated the PR to use sigaction().

daxian-dbw:
Assign this PR to @mirichmo because he would have more context as one of the reviewers.

iSazonov:
@PaulHigin Good catch!  I reread the discussion and it looks that `sigaction`  is exactly what we wanted. 

I believe we should consider [sigprocmask](http://man7.org/linux/man-pages/man2/sigprocmask.2.html) 

Before the code review I still have thoughts how make it easy. Because we search just a workaround for Unix we can: 
1. Use simple wrapper to run "ssh" like [this](https://stackoverflow.com/questions/4515274/externally-disabling-signals-for-a-linux-program) 
2. Ask OpenSSH team to add new option for ignoring SIGINT.
3. If none of the preceding is satisfied we continue the PR as is.

I'd rather not complicate the code and use (1) option.

PaulHigin:
@iSazonov Thanks for you help!
I don't want to use "mask" because the documentation indicates that it masks signals for the entire process group.  Instead I just want the SSH process to ignore the signal.
1. That is what this fix is!  The only difference is that we configure the SSH process start to ignore the signal instead of having the SSH process do it.
2. Or do you mean to move the fix from PowerShell into OpenSSH?  I agree that we should pursue this in the long term but there are a couple of problems:
 - The change may not be accepted or it may take a long time to get accepted as the OpenSSH group must think about wider consequences and compatibility issues.
 - Even if we get the change approved it will still leave many systems with older versions of OpenSSH and without the needed fix.  Fixing this in PowerShell ensures it will just work after installing PowerShell.

The nice thing about open source is that the code can be refactored by anyone if/when a better idea comes along.  But for now I feel this is the best solution.

PaulHigin:
Removing "breaking change" tag since SSH_ASKPASS is no longer needed and behavior remains the same.

iSazonov:
@PaulHigin 
I agree that the old OpenSSH versions will not solve the problem even though new versions get new option to mask SIGINT, but I should have said this option. :-)

I don't see in docs that sigprocmask masks signals for the entire process group - it seems it affects only the [current process](http://man7.org/linux/man-pages/man2/sigprocmask.2.html):
>A child created via fork(2) inherits a copy of its parent's signal
       mask; the signal mask is preserved across execve(2).

Now your code is doing what we need. I suggest to discuss an option that is dear to the Unix world:
- bring the code to external wrapper  `SSHrunner`: to fork - sigprocmask - execvp - ssh
- use our statdard process call to run the wrapper

I believe it's much simpler than use internally CoreFX code. What do you think?

SteveL-MSFT:
I agree with @iSazonov that a wrapper may be simpler than having a copy of the dotnet code.  Should we reuse powershell.exe for this purpose?

iSazonov:
> Should we reuse powershell.exe for this purpose?

I think yes, we can add new "secret" switch for that `powershell -startSSH`.

PaulHigin:
Have two processes to run SSH?  This seems like a waste of processes to me.  For fan-out scenarios it means having twice as many processes on the client for each supported session.  This does not sound like a good design to me.  The code to create a new process is quite small and can be made much smaller if it is meant to be SSH specific.  I just copied the dotNet general process creation code but we don't need all of it.  And I really dislike using powershell.exe to be a process host wrapper since it is pretty heavy weight with the CLR.  I'll work at making the launch code smaller and specific but I feel we should stick with the current process launch design.

iSazonov:
>Have two processes to run SSH? 

No, we run a wrapper - (1) fork and (2) fork to SSH - so we have one parent process - PowerShell, and one child process - SSH. It is not Windows, It is Unix and it is fast and best practice for Unix.

>I really dislike using powershell.exe to be a process host wrapper since it is pretty heavy weight with the CLR

I agree but we can fork in Main(string[] args) before PowerShell initialization code. So it should be fast.

PaulHigin:
Yes, but there is no reason to do it.  Even if the wrapper doesn't "run" it takes up resources, and I assume the scheduler has to account for it.  Also std pipes must be routed.  Why do this when we can easily create a single process?

iSazonov:
>Also std pipes must be routed

Fork inherits its.

I have no more thoughts and PG can make an conclusion.

PaulHigin:
Nevertheless my position remains "why do it when it is not necessary?".  Its funny but I just watched some presentations about performance yesterday.  A big theme from the presenter was that best practices and accepted patterns were many times poor choices performance wise.  One of the top things the presenter looked for when investigating performance problems was unnecessary work such as memory allocations, thread creation, uses of locks, and other system resource usage.  So after that the idea of creating an unneeded process when it is just as easy to create a single process seems undesirable.

daxian-dbw:
Very interesting discussion here and I believe calling `sigactions` before `execve` is the best solution.

I changed to code in `ForkAndExecProcess` a bit on CoreFx to fix issue https://github.com/dotnet/corefx/issues/13447, I think you should use the latest code instead. I will leave a comment there.

src/libpsl-native/src/createprocess.cpp
```diff
@@ -0,0 +1,197 @@
...
+    // format of the target; such errors will emerge via the child process' exit code.
+    if (access(filename, X_OK) != 0)
+    {
+        success = false;
+        goto done;
+    }
+
+    // Open pipes for any requests to redirect stdin/stdout/stderr
+    if ((redirectStdin && pipe(stdinFds) != 0) || (redirectStdout && pipe(stdoutFds) != 0) ||
+        (redirectStderr && pipe(stderrFds) != 0))
```
daxian-dbw: The code in this function was changed a bit to address https://github.com/dotnet/corefx/issues/13447, you can see the latest code at [`pal_process.cpp`](https://github.com/dotnet/corefx/blob/release/2.0.0/src/Native/Unix/System.Native/pal_process.cpp#L142). The change was basically to make all pipe file descriptors close-on-exec. If not doing this here, then it's possible that the ssh process will continue to block on waiting for input even if the redirected standard input has been closed in powershell.

iSazonov:
@PaulHigin @daxian-dbw I want to get clarify about  `sigactions` vs `sigprocmask`. I still belive we can use sigprocmask after fork so OS will not trigger the sigactions empty SIGINT handler in SSH process.

src/System.Management.Automation/CoreCLR/CorePsPlatform.cs
```diff
@@ -602,7 +604,190 @@ internal static int NonWindowsGetProcessParentPid(int pid)
...
+        //
+        // This code is based on GitHub DotNet CoreFx
+        // It is specific to launching the SSH process for use in
+        // SSH based remoting, and is not intended to be general
+        // process creation code.
+        //
+
+#if UNIX
+        private static readonly UTF8Encoding s_utf8NoBom =
+            new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
```
iSazonov: It seems we can use this everywhere. So maybe put this in "right place" as internal?

src/System.Management.Automation/CoreCLR/CorePsPlatform.cs
```diff
@@ -602,7 +604,190 @@ internal static int NonWindowsGetProcessParentPid(int pid)
...
+            return childPid;
+        }
+
+        /// <summary>Opens a stream around the specified file descriptor and with the specified access.</summary>
+        /// <param name="fd">The file descriptor.</param>
+        /// <param name="access">The access mode.</param>
+        /// <returns>The opened stream.</returns>
+        private static FileStream OpenStream(int fd, FileAccess access)
+        {
+            Debug.Assert(fd >= 0);
```
iSazonov: Should we add an usefull message?

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -1999,25 +1999,21 @@ internal override BaseClientSessionTransportManager CreateClientSessionTransport
...
             out StreamReader stdOutReaderVar,
             out StreamReader stdErrReaderVar)
         {
             string filePath = string.Empty;
-#if !UNIX
             var context = Runspaces.LocalPipeline.GetExecutionContextFromTLS();
             if (context != null)
             {
-                var cmdInfo = context.CommandDiscovery.LookupCommandInfo("ssh.exe", CommandOrigin.Internal) as ApplicationInfo;
+                var cmdInfo = context.CommandDiscovery.LookupCommandInfo("ssh", CommandOrigin.Internal) as ApplicationInfo;
```
iSazonov: Why we remove ".exe" for Windows? We can catch .bat or .com file.

src/libpsl-native/src/createprocess.cpp
PaulHigin: I don't think this is applicable in this situation.  The pipes are only used for PSRP messages for a single remoting session.  PowerShell only closes the pipe at the end of the session at which time it also terminates the SSH process.

src/System.Management.Automation/CoreCLR/CorePsPlatform.cs
PaulHigin: I am not sure what you mean, I only see this being used here in my branch.

src/System.Management.Automation/CoreCLR/CorePsPlatform.cs
PaulHigin: Sure, but I do see the point of DotNet CoreFx developers who originated this code.  The assert condition is pretty explanatory...

src/libpsl-native/src/createprocess.cpp
daxian-dbw: Without `close-on-exec`, the pipe file descriptors will be leaked to all subsequent processes created by powershell. The symptom I described above is one result this would cause. Although it's not applicable to the SSH remoting scenario, my worry is that there may be other symptoms resulted by the leak that would cause a problem we don't know yet which might be hard to diagnose.

src/System.Management.Automation/CoreCLR/CorePsPlatform.cs
daxian-dbw: I think @iSazonov means that `s_utf8NoBom` should be put at some common place so that a single instance can be used in all places in our code base.

daxian-dbw:
@iSazonov I see this in the man page of `SIGPROCMASK`
> The use of sigprocmask() is unspecified in a multithreaded process;

This makes me feel safer to use `sigactions`.

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
PaulHigin: I was hoping to simply the code a bit between platforms.  But I see your point and will fix.

src/System.Management.Automation/CoreCLR/CorePsPlatform.cs
PaulHigin: Oh I see.  Yes this does seem like something that would be useful.  I'll move it to PsUtils.cs

src/libpsl-native/src/createprocess.cpp
PaulHigin: Ok, the leak definitely sounds bad.  I'll add the close-on-exec.  So that I understand more clearly, can you provide an example where the Fd is leaked even when SSH process is terminated?

daxian-dbw: Take the write end of the redirected stdin pipe as an instance (similar to the redirected stdout/stderr pipes), a FD pointing to the write end of that pipe will be created in powershell process after creating the SSH process. Assuming powershell then creates some other processes before SSH process terminates (I'm thinking of the scenario where I create a SSH PSSession but no an interactive remote session, so it's possible to run native commands locally while the SSH process is still alive), then that FD will be inherited by all those processes, so they are all potentially able to write to that pipe or affect that pipe in possible ways.

In this case, the SSH process is the only reader of that pipe, so when SSH terminates, "_the only reader dies, and then the writer gets notified about this by getting a SIGPIPE or at least an EPIPE error (depending on how signals are defined)._" (quoted from [here](https://stackoverflow.com/questions/19265191/why-should-you-close-a-pipe-in-linux))

I'm not saying it will definitely cause a problem, but just that if the leaked FD cause trouble to us, it might be very hard to figure out the real cause.

PaulHigin: Thanks, this definitely helps and could be an issue for the fan-out scenario.  I'll have this change in the PR update.

src/libpsl-native/src/createprocess.cpp
```diff
@@ -0,0 +1,225 @@
...
+        success = false;
+        goto done;
+    }
+
+    if (processId == 0) // processId == 0 if this is child process
+    {
+        // Close the child's copy of the parent end of any open pipes
+        CloseIfOpen(stdinFds[WRITE_END_OF_PIPE]);
+        CloseIfOpen(stdoutFds[READ_END_OF_PIPE]);
+        CloseIfOpen(stderrFds[READ_END_OF_PIPE]);
```
daxian-dbw: There are some additional changes for the close-on-exec -- calls to `CloseIfOpen` are removed from the `if (processId == 0)` block, see the CoreFx code [here](https://github.com/dotnet/corefx/blob/release/2.0.0/src/Native/Unix/System.Native/pal_process.cpp#L167).

I may be easier to see the differences in the PR change: https://github.com/dotnet/corefx/commit/6d204a6a83e3eca283535ca930157350d4adabd6

daxian-dbw:
@PaulHigin Could you please update the PR description? A great discussion happened in this PR that led to a better solution. So the original description is out-dated and we'd better capture the solution of using `SIGPROCMASK` in the PR description.

iSazonov:
LGTM.

Sorry for the long discussion - my last code for Unix was exactly 20 years ago. I'm glad I can remember just something. 😄

mirichmo (changes requested):

src/libpsl-native/src/createprocess.cpp
PaulHigin: Fixed

PaulHigin:
@iSazonov Thank you for your help on this.

src/System.Management.Automation/CoreCLR/CorePsPlatform.cs
```diff
@@ -602,7 +604,188 @@ internal static int NonWindowsGetProcessParentPid(int pid)
...
+        #region UNIX Create SSH Process
 
+        //
+        // This code is based on GitHub DotNet CoreFx
+        // It is specific to launching the SSH process for use in
+        // SSH based remoting, and is not intended to be general
+        // process creation code.
+        //
+
+#if UNIX
```
PaulHigin: Done

src/System.Management.Automation/CoreCLR/CorePsPlatform.cs
```diff
@@ -799,6 +982,12 @@ internal static UnixTm DateTimeToUnixTm(DateTime date)
                 [return: MarshalAs(UnmanagedType.I1)]
                 internal static extern bool IsSameFileSystemItem([MarshalAs(UnmanagedType.LPStr)]string filePathOne,
                                                                  [MarshalAs(UnmanagedType.LPStr)]string filePathTwo);
+
+                [DllImport(psLib, CharSet = CharSet.Ansi, SetLastError = true)]
```
PaulHigin: Done

SteveL-MSFT:
@iSazonov can you formally approve if you're good with the changes?  Thanks!

iSazonov (approved):

iSazonov:
@SteveL-MSFT Approved.

PaulHigin:
@mirichmo Is there any reason why this cannot be merged?

src/System.Management.Automation/CoreCLR/CorePsPlatform.cs
```diff
@@ -8,6 +8,8 @@
 using Microsoft.Win32;
 using Microsoft.Win32.SafeHandles;
 using System.IO;
+using System.Diagnostics;
```
mirichmo: Please clean this up

mirichmo (changes requested):

mirichmo (approved):

PaulHigin:
@mirichmo  Thanks!

PR #3887, opened 2017-05-30
Move to latest .NET Core and enable `-SkipCertificateCheck` on OSX

daxian-dbw:
Fix #3648

Summary
----------
A workaround has been made in .NET Core to allow accept any certificate: https://github.com/dotnet/corefx/issues/19709
To enable `-SkipCertificateCheck` in web cmdlets, we need to move to the latest .NET Core, and also use `HttpClientHandler.DangerousAcceptAnyServerCertificateValidator` as the callback delegate.

New Issues
------------
There are 2 problems with the latest .NET Core:
1. Behavior change in `FileSystemInfo.Attribute` causes 2 tests to fail on Unix platforms. https://github.com/dotnet/corefx/issues/20456 was filed to track the behavior change. These 2 tests are currently marked as `-Pending` for now.
2. VSCode C# extension doesn't support .NET Core `2.0.0-preview2-*` SDK. When using `2.0.0-preview2` packages and dotnet-cli, you will get a `MissingMethodException` opening the project in VSCode (see below). This is tracked by https://github.com/OmniSharp/omnisharp-vscode/issues/1495, and hopefully will be addressed this week according to this comment: https://github.com/OmniSharp/omnisharp-vscode/issues/1495#issuecomment-304547648
```
[fail]: OmniSharp.MSBuild.ProjectFile.ProjectFileInfo
        The "ResolvePackageDependencies" task failed unexpectedly.
System.MissingMethodException: Method not found: 'System.Collections.Generic.IList`1<NuGet.Packaging.Core.PackageDependency> NuGet.ProjectModel.LockFileTargetLibrary.get_Dependencies()'.
   at Microsoft.NET.Build.Tasks.ResolvePackageDependencies.GetPackageDependencies(LockFileTargetLibrary package, String targetName, Dictionary`2 resolvedPackageVersions, HashSet`1 transitiveProjectRefs)
   at Microsoft.NET.Build.Tasks.ResolvePackageDependencies.GetPackageAndFileDependencies(LockFileTarget target)
   at Microsoft.NET.Build.Tasks.ResolvePackageDependencies.RaiseLockFileTargets()
   at Microsoft.NET.Build.Tasks.ResolvePackageDependencies.ExecuteCore()
   at Microsoft.NET.Build.Tasks.TaskBase.Execute()
   at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
[...]

PR #3903, opened 2017-06-01
Refactor PowerShell Core to use the default CoreCLR loader instead

### LINKED ISSUE #3649: Avoid using powershell assembly load context with netcoreapp2.0
Issue Summary
----------------
The API `AppDomain.GetAssemblies` is brought back in .NET Core 2.0 which returns the loaded assemblies from the default loader. Therefore it's possible now for powershell to just depend on the default CoreCLR loader without having our own assembly load context getting in the picture. This would greatly simplify the scenario of hosting powershell in applications.

CoreFX Fixes needed
-----------------------
- [x] https://github.com/dotnet/corefx/issues/18989 - `AssemblyLoadContext.GetLoadContext` crashes when pass in a dynamic assembly. **[Verified Fixed]**
- [x] https://github.com/dotnet/corefx/issues/18877 - `AppDomain.GetAssemblies` doesn't return dynamic assemblies that are emitted on the fly. **[Verified Fixed]**
- [x] https://github.com/dotnet/corefx/issues/18791 - For an assembly loaded via 'Assembly.LoadFrom', the assembly it referenced cannot be loaded even if they are in the same folder. **[Verified Fixed]**
- [x] OmniSharp/omnisharp-vscode#1495 - Preview 2 builds are not yet supported by VS Code C#. **[Verified Fixed] -- `v1.11.0-Beta4` VSCode C# extension has the fix and it needs to be installed manually`**
- [x] https://github.com/dotnet/corefx/issues/21095 - [Regression] ResourceManager.GetString(string) throw System.IO.FileNotFoundException. **[Verified Fixed]**

Tasks
-----
- [x] Refactor the code to make PowerShell Core depend on the default CoreCLR loader. Fixed via #3903
- [x] Update the hosting code in [pwrshclrhost.cpp](https://github.com/PowerShell/PowerShell/blob/309bf603f9aff9009dce7e725d42b2d4f99f58c9/src/powershell-native/nativemsh/pwrshplugin/pwrshclrhost.cpp#L62)
- [x] Update the host code in [Shell.cs](https://github.com/PowerShell/psl-omi-provider/blob/master/src/Shell.c#L446) in psl-omi-provider (submitted PR https://github.com/PowerShell/psl-omi-provider/pull/99, but since the CI build depends on powershell release package to run tests, the PR CI will fail until after beta.3 is out.)
- [ ] Remove `Microsoft.PowerShell.CoreCLR.AssemblyExtensions.LoadFrom(string assemblyPath)` once the `PackageManagement` module is migrated to .NET Core 2.0.
- [x] Move PowerShell Core to the latest .NET Core to get all fixes we need. (~via #3887~ Ran into a regression in the latest .NET Core, so moved back to 2.0.0-preview1 via #4026)
- [ ] Think about removing `Microsoft.PowerShell.CoreCLR.AssemblyLoadContext.dll` completely.

daxian-dbw:
Partially fix #3649

Issue Summary
----------------
The API `AppDomain.GetAssemblies` is brought back in .NET Core 2.0 which returns the loaded assemblies from the default loader. Therefore it's possible now for powershell to just depend on the default CoreCLR loader without having our own assembly load context getting in the picture. This would greatly simplify the scenario of hosting powershell in applications.

Fix
---
Remove the code that spins up our own assembly load context. Keep the code that registers our `Resolve` method to the default loader's `Resolving` event, so that we can continue to do special assembly resolution as needed (such as the GAC probing logic needed for consuming FullCLR PS modules).

Essentially, the assembly `Microsoft.PowerShell.CoreCLR.AssemblyLoadContext.dll` is not needed anymore, the remaining code should be moved to S.M.A.dll. However, that will break DSC and other native hosts that are hosting powershell. So this assembly is kept for now.

CoreFX Fixes needed
-----------------------
https://github.com/dotnet/corefx/issues/18989
https://github.com/dotnet/corefx/issues/18877
https://github.com/dotnet/corefx/issues/18791

This PR is not blocked by those CoreFX issues, but it's incomplete until we have the fixes. Those issues have already been addressed, and #3887 will get those fixes for us.
Note that this PR is NOT blocked by #3887.

Follow-up work
-----------------
This PR addresses the first task listed in #3649. Three tasks are remaining.

### DIFF
--- a/src/Microsoft.PowerShell.CoreCLR.AssemblyLoadContext/CoreCLR/CorePsAssemblyLoadContext.cs
+++ b/src/Microsoft.PowerShell.CoreCLR.AssemblyLoadContext/CoreCLR/CorePsAssemblyLoadContext.cs
@@ -31,11 +31,7 @@ namespace System.Management.Automation
         // When the first attempt fails, we again need to retrieve the resource string to construct another exception, which ends up with an infinite loop.
         private const string BaseFolderDoesNotExist = "The base directory '{0}' does not exist.";
         private const string ManifestDefinitionDoesNotMatch = "Could not load file or assembly '{0}' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference.";
-        private const string AssemblyPathDoesNotExist = "Could not load file or assembly '{0}' or one of its dependencies. The system cannot find the file specified.";
-        private const string InvalidAssemblyExtensionName = "Could not load file or assembly '{0}' or one of its dependencies. The file specified is not a DLL file.";
-        private const string AbsolutePathRequired = "Absolute path information is required.";
         private const string SingletonAlreadyInitialized = "The singleton of PowerShellAssemblyLoadContext has already been initialized.";
-        private const string UseResolvingEventHandlerOnly = "PowerShellAssemblyLoadContext was initialized to use its 'Resolving' event handler only.";
 
         #endregion Resource_Strings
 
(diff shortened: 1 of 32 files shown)


SteveL-MSFT (approved):
LGTM

src/Microsoft.PowerShell.CoreCLR.AssemblyLoadContext/CoreCLR/CorePsAssemblyLoadContext.cs
```diff
@@ -31,11 +31,7 @@ internal partial class PowerShellAssemblyLoadContext : AssemblyLoadContext
         // When the first attempt fails, we again need to retrieve the resource string to construct another exception, which ends up with an infinite loop.
         private const string BaseFolderDoesNotExist = "The base directory '{0}' does not exist.";
         private const string ManifestDefinitionDoesNotMatch = "Could not load file or assembly '{0}' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference.";
-        private const string AssemblyPathDoesNotExist = "Could not load file or assembly '{0}' or one of its dependencies. The system cannot find the file specified.";
-        private const string InvalidAssemblyExtensionName = "Could not load file or assembly '{0}' or one of its dependencies. The file specified is not a DLL file.";
-        private const string AbsolutePathRequired = "Absolute path information is required.";
         private const string SingletonAlreadyInitialized = "The singleton of PowerShellAssemblyLoadContext has already been initialized.";
```
adityapatwardhan: What is the localization story for these messages?

adityapatwardhan (approved):
Approved with 1 comment. Probably we need not worry about it right now, as this code might be merged into SMA.dll in the future.

src/Microsoft.PowerShell.CoreCLR.AssemblyLoadContext/CoreCLR/CorePsAssemblyLoadContext.cs
daxian-dbw: Sorry for the late response. These strings are currently not localized because the loading of a satellite resource DLL when a loading exception happens causes a recursion. It's explained in the comment above the first resource string. Eventually, after moving the remaining code from `PowerShellAssemblyLoadContext.dll` to `SMA.dll`, we can reconsider the localization.

daxian-dbw:
Minor update to add back `Microsoft.PowerShell.CoreCLR.AssemblyExtensions.LoadFrom(string assemblyPath)` for the short term. This is because PackageManagement depends on this API and thus removing this API will instantly break PackageManagement.

Since the `LoadFrom` functionality has been removed from our assembly load context, we just call `Assembly.LoadFrom(string)` in the above API. I also added `ObsoleteAttribute` to this API to prevent new code from using it.

I have verified that PackageManagement works with this update.
I also verified that host powershell in an application is the same as in FullCLR after this change.

PR #3794, opened 2017-05-16
Add documentation requirements to CONTRIBUTING.md

joeyaiello:
This is intended to address the underlying issue behind #2566 by adding a note in `CONTRIBUTING.md` to create a corresponding pull request in PowerShell/PowerShell-Docs when making code changes that warrant it.

It's also going to take some enforcement from @PowerShell/powershell-maintainers, which I know will be difficult because our documentation isn't completely up-to-date--we're playing catch up here. However, it should be analogous to the test debt that we've had while enforcing requirements for new tests.

I'm also not opposed to creating a label as well to track those PRs where doc changes are required but where they haven't been addressed. It's really up to the @PowerShell/powershell-maintainers here, I'm open to whatever.

PR #3908, opened 2017-06-01
Add code cleanup and documentation labels to issue-management.md

SteveL-MSFT:
Added descriptions for `Issue-Code Cleanup` and `Review - Need Documentation` labels

### DIFF
--- a/docs/maintainers/issue-management.md
+++ b/docs/maintainers/issue-management.md
@@ -75,7 +75,6 @@ The following labels are used on PRs:
 * `Review - Abandoned` : The PR was not updated for significant number of days (the exact number could vary over time).
   Maintainers should look into such PRs and re-evaluate them.
 * `Review - Committee` : The PR/Issue needs a review from [powershell-committee](../community/governance.md#powershell-committee)
-* `Review - Need Documentation` : The PR has changes that require a documentation change or new documentation added to [PowerShell-Docs](http://github.com/powershell/powershell-docs)
 
 ### Miscellaneous labels
 
@@ -90,3 +89,4 @@ The following labels are used on PRs:
   because they more directly affect the usability of a particular feature or area.
 * `Changelog Needed`: The PR requires an addition to the changelog,
   and should be removed when it has been added.
+* `Documentation Needed` : The PR has changes that require a documentation change or new documentation added to [PowerShell-Docs](http://github.com/powershell/powershell-docs)
--- a/docs/maintainers/issue-management.md
+++ b/docs/maintainers/issue-management.md
@@ -8,6 +8,7 @@ Issues are opened for many different reasons.
 We use the following labels for issue classifications:
 
 * `Issue-Bug`: the issue is reporting a bug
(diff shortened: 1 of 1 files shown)


docs/maintainers/issue-management.md
```diff
@@ -8,6 +8,7 @@ Issues are opened for many different reasons.
 We use the following labels for issue classifications:
 
 * `Issue-Bug`: the issue is reporting a bug
+* `Issue-Code Cleanup`: the issue is for cleaning up the code with no impact on functionality
```
iSazonov: Maybe remove the space in the label? This makes it easier to write manual filters in GitHub.

Please clarify - usually we have PRs with code/test cleanup not Issues.

SteveL-MSFT: I originally had no space, but looking at some of the other labels we have, I decided to have the space for consistency.

`Issue-Code Cleanup` is to identify known issues in the code where we should clean up (I've already labeled a number of such issues) that isn't a priority to address so they aren't automatically PRs

docs/maintainers/issue-management.md
```diff
@@ -74,6 +75,7 @@ The following labels are used on PRs:
 * `Review - Abandoned` : The PR was not updated for significant number of days (the exact number could vary over time).
   Maintainers should look into such PRs and re-evaluate them.
 * `Review - Committee` : The PR/Issue needs a review from [powershell-committee](../community/governance.md#powershell-committee)
+* `Review - Need Documentation` : The PR has changes that require a documentation change or new documentation added to [PowerShell-Docs](http://github.com/powershell/powershell-docs)
```
joeyaiello: Wouldn't this be "Needs"? If you think that's pedantic, whatever, I'm fine with what we've got, but it sounds funny to me.

joeyaiello: In fact, now that I type it out, I do think it would be useful for "Documentation" to be more upfront in the label for easy ID. Can we go with `Review _ Documentation Needed`?

In fact, his `Review` even needed? See my uber-comment below...

joeyaiello (changes requested):
Generally, this LGTM, but it's unclear to me what action the "Review - Needs Documentation" is actually supposed to trigger (which is why I hadn't done this in my other PR yet). What is actually supposed to happen when someone adds that label? Do we block the PR merge? Do we go generate an issue in PowerShell-Docs?

iSazonov:
I believe we shouldn't block PRs, we should full async, PowerShell Doc Writers should be active side and monitor the label to start a documentation process in accordance with _their_ workflow.

docs/maintainers/issue-management.md
SteveL-MSFT: Intent is to label PRs that have doc implications, but we agreed we wouldn't block accepting PRs due to lack of doc.  I put it with the `Review` group of labels as it seemed appropriate to get attention.  I'm fine with standalone `Documentation Needed`

SteveL-MSFT:
@joeyaiello fixed

joeyaiello (approved):
Going to update #3794 to link up with this.

PR #3917, opened 2017-06-02
Set assembly versions by MSBuild

iSazonov:
Related #3400

Motivation
---
We inherited old assembly file versions from Windows PowerShell.
Currently we shouldn't set versions of assemblies statically in `AssemblyInfo.cs` files, we should set versions of assemblies dynamically by MSBuild as discussed in #3690.

Fix
---
1. Remove `AssemblyVersion` and `AssemblyFileVersionAttribute` attributes from `AssemblyInfo.cs` files.
2. Put all common properties in PowerShellCommon.props including version related properties.
3. Import `PowerShellCommon.props` in all csproj files.

Follow-Up Work
---
In the PR we temporarily hard code assembly versions as "6.0.0-beta.2".
Later we should set a dll version based on GitCommitId.


src/Microsoft.Management.Infrastructure.CimCmdlets/Microsoft.Management.Infrastructure.CimCmdlets.csproj
```diff
@@ -1,17 +1,13 @@
...
-    <VersionPrefix>6.0.0</VersionPrefix>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <DelaySign>true</DelaySign>
+    <Description>PowerShell Core's Microsoft.Management.Infrastructure.CimCmdlets project</Description>
+    <NoWarn>$(NoWarn);CS1570;CS1572;CS1573;CS1574;CS1584;CS1587;CS1591</NoWarn>
     <AssemblyName>Microsoft.Management.Infrastructure.CimCmdlets</AssemblyName>
-    <AssemblyOriginatorKeyFile>../signing/visualstudiopublic.snk</AssemblyOriginatorKeyFile>
-    <SignAssembly>true</SignAssembly>
-    <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
```
mirichmo: All of the GenerateAssembly* entries are missing. I'm assuming it was intentional. Can you please explain?

mirichmo (changes requested):

src/Microsoft.Management.Infrastructure.CimCmdlets/Microsoft.Management.Infrastructure.CimCmdlets.csproj
iSazonov: Do you mean "<GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>" ?

We explicitly had AssemblyInfo attrubutes in `AssemblyInfo.cs` files and we had to disable auto generating its by "<GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>"
Now we use auto generating and should remove the "false".

src/Microsoft.Management.Infrastructure.CimCmdlets/Microsoft.Management.Infrastructure.CimCmdlets.csproj
```diff
@@ -1,17 +1,13 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <VersionPrefix>6.0.0</VersionPrefix>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-    <DelaySign>true</DelaySign>
+    <Description>PowerShell Core's Microsoft.Management.Infrastructure.CimCmdlets project</Description>
+    <NoWarn>$(NoWarn);CS1570;CS1572;CS1573;CS1574;CS1584;CS1587;CS1591</NoWarn>
```
iSazonov: I'm surprised, too. We do not receive these warnings with the original csproj files, but they appear after we add "*.props".
I guess this is a "feature" of `Sdk="Microsoft.NET.Sdk"` templates.
I believe that some of these warnings can be easily corrected in separate PRs.

iSazonov: I review the warnings - all belong to XML comments and we have TONs warnings - It takes a lot of work to fix it.

src/Microsoft.Management.Infrastructure.CimCmdlets/Microsoft.Management.Infrastructure.CimCmdlets.csproj
mirichmo: My comment applied to all of those flags as well since I referred to them as "GenerateAssembly*". I am not particularly familiar with them and would like to make sure that we are not inadvertently dropping a flag that is necessary for PowerShell.

iSazonov: All these attributes were statically installed in the `AssemblyInfo.cs` files and that their automatic generation was turned off in csproj files. Now we've removed static attributes and allowed them to be automatically generated. In other words, attributes are present, but their values have changed - now they are always up-to-day. If you build the branch you can find the attributes in auto generated `*.AssemblyInfo.cs` files in OBJ directory.

iSazonov:
@mirichmo @daxian-dbw Could you please continue with the PR?

src/Microsoft.Management.Infrastructure.CimCmdlets/Microsoft.Management.Infrastructure.CimCmdlets.csproj
powercode: I would put the import of the `PowerShell.common.props` before the property section in the csproj files.

That way you can do per project overrides if there is something that needs to be project specific.

Maybe even have .before.props and .after.props (or with some other naming).

`Before` contains defaults, and `After` contains common things you want to have done, but parameterized by each project properties.

iSazonov: @powercode Could you please give an example?

src/Microsoft.Management.Infrastructure.CimCmdlets/Microsoft.Management.Infrastructure.CimCmdlets.csproj
lzybkr: I'm fine with disabling the xml doc warnings.

src/Microsoft.Management.Infrastructure.CimCmdlets/Microsoft.Management.Infrastructure.CimCmdlets.csproj
lzybkr: I agree with @powercode - simply move the `Import` so that it's before any `PropertyGroup`.

One example - we might move NoWarn into the `PowerShellCommon.props`, but in some project, we want to more/less warnings disabled.

iSazonov: Thanks for clarify. It's good.
Done.

PowerShellCommon.props
```diff
@@ -0,0 +1,22 @@
...
+  <PropertyGroup>
+    <Product>PowerShell Core</Product>
+    <Company>Microsoft Corporation</Company>
+    <Copyright>(c) Microsoft Corporation. All rights reserved.</Copyright>
+
+    <FileVersion>6.0.0.0</FileVersion>
+    <AssemblyVersion>6.0.0.0</AssemblyVersion>
+
+    <ProductVersion>6.0.0-beta.2</ProductVersion>
+    <InformationalVersion>6.0.0-beta.2</InformationalVersion>
```
mirichmo: beta.3 please since we created the release today. Otherwise, it looks fine to me.

daxian-dbw: Why is `6.0.0-beta.2` hard coded?

iSazonov: The PR only targets assemblies and make multiple cleanups. I didn't want to complicate the PR and hard coded the tag.
In next PR I plan use git to get the tag dynamically. We discussed this previously in #3690.

mirichmo (approved):

src/System.Management.Automation/AssemblyInfo.cs
```diff
@@ -2,9 +2,6 @@
 using System.Reflection;
 using System.Resources;
 
-[assembly: AssemblyFileVersionAttribute("3.0.0.0")]
-[assembly: AssemblyVersion("3.0.0.0")]
```
daxian-dbw: Change the assembly version from 3.0.0.0 to 6.0.0.0 will break all existing assemblies that reference to S.M.A 3.0.0.0, for example the existing PakcageManagement module.

src/Microsoft.Management.Infrastructure.CimCmdlets/Microsoft.Management.Infrastructure.CimCmdlets.csproj
```diff
@@ -1,15 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
+  <Import Project="..\..\PowerShellCommon.props"/>  
+
   <PropertyGroup>
-    <VersionPrefix>6.0.0</VersionPrefix>
```
daxian-dbw: `<VersionPrefix>` tag is needed to generate NuGet packages with the version like `6.0.0-alpha.3`,

daxian-dbw:
@mirichmo I don't think this PR is ready to merge yet. Please revert this merge.
@PowerShell/powershell-maintainers I didn't get the time to thoroughly review all PRs related to .csproj changes. Let's hold on merging them.

daxian-dbw:
@mirichmo hold on the revert. Let's see if we can quickly fix the issues.

src/System.Management.Automation/AssemblyInfo.cs
daxian-dbw: image

src/Microsoft.Management.Infrastructure.CimCmdlets/Microsoft.Management.Infrastructure.CimCmdlets.csproj
daxian-dbw: Here are the NuGet packages generated:
```
PS:7> dir -Name
Microsoft.PowerShell.Commands.Diagnostics.1.0.0-beta.4.nupkg
Microsoft.PowerShell.Commands.Diagnostics.1.0.0-beta.4.symbols.nupkg
Microsoft.PowerShell.Commands.Management.1.0.0-beta.4.nupkg
Microsoft.PowerShell.Commands.Management.1.0.0-beta.4.symbols.nupkg
Microsoft.PowerShell.Commands.Utility.1.0.0-beta.4.nupkg
...
```
They becomes **1.0.0**

src/System.Management.Automation/AssemblyInfo.cs
daxian-dbw: I think bumping assembly version number needs to be reviewed by powershell committee before the change.

daxian-dbw:
@mirichmo I don't have a quick fix. I suggest reverting this merge for now and submit another PR later.

daxian-dbw:
This merge has been reverted via #4091. Thanks @mirichmo for the quick turnaround.
@iSazonov Sorry for the troubles ☹️ Lesson learned: if you think a PR is risky but don't yet have the time to thoroughly review, leave a comment to point out that you need more time to review so that everyone is on the same page. I will open a new PR for this change.

iSazonov:
😕 It's a very strange side effect. I thought PowerShell Core assembles is not Windows PowerShell assemblies, and we could be free in the change. CoreFX assemblies don't keep old versions too. Maybe a right way is to be based on packages versions?

daxian-dbw:
@iSazonov With `netstandard2.0` we expect most of the existing Windows powershell modules to work on powershell core without change. They were built against 3.0.0.0 S.M.A.dll, so we want to make sure nothing will break there.

By the way, I think you have to submit a new PR for this. If I submit it, it will be from my fork and thus you won't be able to make changes. It's very easy to check out a PR with the tool [`hub`](https://github.com/github/hub/releases). Once you get that tool, you can do things like this to get the changes from this PR:
```
## It will get the changes from this PR to a local branch named iSazonov-csproj-common-props
hub checkout https://github.com/PowerShell/PowerShell/pull/3917
## Then rebase this branch and push it to your fork
## Then submit a new PR with that remote branch
```
Again, sorry for the troubles and thanks for your understanding.

iSazonov:
@daxian-dbw Should I restore the branch for you?

iSazonov:
> I think you have to submit a new PR for this.

Ok!

Issue #2882, opened 2016-12-14
Dotnet-resgen now support <lang>


iSazonov:
Based on dotnet/cli#819 we could make improvements here with strongly-typed auto-generated C# bindings for resources.

@andschwa @vors  If the community can help with this, could you publish a road map?

Related #745 #746

lzybkr:
Most of our resources have been localized by the Windows localization team, so I don't think it makes much sense to ask for community help until we figure out the best way to pull in those translations.

vors:
@lzybkr I don't see how that's related. The `<lang>` in the title refers to C# language, not the localizations.

@iSazonov yes, it would be awesome to remove our own res-gen. Theoretically it should be a drop-in replacement. We are currently pinning version of dotnet, so I think it make sense to combine this effort with switch to csproj.

iSazonov:
I tried to build with target "ResGen" and get error in diagnostic log "Generating strongly typed resource files not currently supported on .NET Core MSBuild". It seems we are still blocked by .Net Core 2.0 Preview.

iSazonov:
Tracking Issue in MSBuild repo https://github.com/Microsoft/msbuild/issues/2272

iSazonov:
It seems we get this in .Net Core 3 and 5.
See https://github.com/microsoft/msbuild/issues/2272#issuecomment-532268770

andschwa:
@iSazonov oh, that's awesome!

iSazonov:
I pulled PR to use dotnet ResGen.

PR #3918, opened 2017-06-02
Enhance ResGen tool to support a command line argument

iSazonov:
Related #3400

Motivation
---
ResGen tool process all resx files in PowerShell Repo at a time. It is not problem because it works very fast. But updating multiple cs files causes all Repo projects to be recompiled. It really slows down.

We still can not use native resgen from dotnet-cli. See https://github.com/PowerShell/PowerShell/issues/2882#issuecomment-304300044
So we should enhance our ResGen tool, use MSBuild capabilities and process resx files one at a time to avoid unnecessary rebuilds.

Fix
---
We can specify a resx file path parameter on the command line and process resx files one at a time.

Follow-Up work
---
We will clean up Build.psm1 and modify csproj files to use the new feature of ResGen tool with MSBuild for faster project build.

### DIFF
--- a/src/ResGen/Program.cs
+++ b/src/ResGen/Program.cs
@@ -10,12 +10,30 @@ namespace ConsoleApplication
     {
         public static void Main(string[] args)
         {
-            // we are assuming resgen is run with 'dotnet run'
-            // so we can use relative paths
-            foreach (string folder in Directory.EnumerateDirectories(".."))
+            IEnumerable<string> dirs;
+            string fileFilter;
+
+            if (args.Length == 1)
+            {
+                // We are assuming resgen is run with 'dotnet run pathToResxFile.resx'.
+                fileFilter = Path.GetFileName(args[0]);
+                string moduleDirectory = Directory.GetParent(Path.GetDirectoryName(args[0])).FullName;
+                dirs = new List<string>() { moduleDirectory };
+            }
+            else
+            {
+                // We are assuming resgen is run with 'dotnet run'
+                // so we can use relative path to get a parent directory
+                // to process all *.resx files in all project subdirectories.
+                fileFilter = "*.resx";
+                dirs = Directory.EnumerateDirectories("..");
+            }
+
+            foreach (string folder in dirs)
             {
                 string moduleName = Path.GetFileName(folder);
                 string resourcePath = Path.Combine(folder, "resources");
+
                 if (Directory.Exists(resourcePath))
                 {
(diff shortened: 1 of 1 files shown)


vors:
I think it's a good idea.

> We will clean up Build.psm1 and modify csproj files to use the new feature of ResGen tool for faster project build.

Couple comments: I don't think we should try to be smart and auto-detect set of files for resgen in build.psm1. We should just figure out the way to switch to msbuild resgen sooner or later.
Let's just expose this functionality thru `Start-ResGen` optional parameter in the same PR.

iSazonov:
Yes, it is all for MSBuild - I updated the comment.

vors:
oh, do you mean that you know how to plug it into csproj files so it's called automatically?

iSazonov:
Yes. You can see sample in #3870

TravisEz13 (approved):

iSazonov:
@TravisEz13 Is the PR ready to merge?

src/ResGen/Program.cs
```diff
@@ -10,12 +10,30 @@ public class Program
...
-            // so we can use relative paths
-            foreach (string folder in Directory.EnumerateDirectories(".."))
+            IEnumerable<string> dirs;
+            string fileFilter;
+
+            if (args.Length == 1)
+            {
+                // We are assuming resgen is run with 'dotnet run pathToResxFile.resx'.
+                fileFilter = Path.GetFileName(args[0]);
+                string moduleDirectory = Directory.GetParent(Path.GetDirectoryName(args[0])).FullName;
```
daxian-dbw: > string moduleDirectory = Directory.GetParent(Path.GetDirectoryName(args[0])).FullName;
```
PS:39> [System.IO.Directory]::GetParent

OverloadDefinitions
-------------------
static System.IO.DirectoryInfo GetParent(string path)
```
A `DirectoryInfo` object cannot be converted to a string ...

daxian-dbw: Why not `Path.GetDirectoryName(Path.GetDirectoryName(args[0]))`?

daxian-dbw (approved):

daxian-dbw:
LGTM except for a minor comment.

vors (approved):

src/ResGen/Program.cs
iSazonov: Good catch! I'll fix on next week.

iSazonov: Done.

PR #3407, opened 2017-03-23
Remove inapplicable documentation and build helpers

vors:
Since we are not build FullCLR version anymore, I removed the following:

- map.json files
- map.json related stuff from build.psm1
- windows-full build docs

You can keep this PR around until FullCLR would be completely removed.

PR #3921, opened 2017-06-02
Make 'Start-PSBuild' and 'Start-PSPackage' accept a release tag argument

### LINKED ISSUE #3909: We should be able to complete all tasks for a release in a branch before creating the release tag
Ideally, we should create a release branch and complete all release tasks using that branch, including generating all packages and update all docs, and then push the release tag and merge the release branch to master.

However, this cannot be done currently because our build depends on the release tag to generate `GitCommitId` for `$PSVersionTable`. and thus the release tag has to be pushed before starting the release tasks.

To break this dependency, we should update `Start-PSBuild` to accept a GitCommitId string and update the Jenkins build script so that the GitCommitId string can be passed to the build pipeline.

Tasks
------
- [x] Update `Start-PSBuild` and `Start-PSPackage` to accept a `ReleaseTag` argument, which will override the git commit id retrieved from `git describe`. Resolved via #3921.
- [x] Update our VSTS builder scripts to honor the `-ReleaseTag` flag. Resolved by @TravisEz13 via https://github.com/PowerShell/PSRelease/pull/13

daxian-dbw:
Partially fix #3909

Issue Summary
---------------
Currently a release build depends on the release tag to be pushed. We need to break this dependency, so that creating the release tag can be made the last step of a release -- do it only if all preparation of the release is done.

Fix
---
Add a new parameter `-ReleaseTag` to `Start-PSBuild` and `Start-PSPackage` to allow a release tag to be passed in. When the release tag is specified at the command line, it will override the git commit id retrieved from `git describe`.

Example
--------
```
# Build for v6.0.0-beta.3 release
Start-PSBuild -Clean -CrossGen -PSModuleRestore -ReleaseTag v6.0.0-beta.3

# Create package for v6.0.0-beta.3 release
Start-PSPackage -ReleaseTag v6.0.0-beta.3 
```

Follow-up work
---------------
Our VSO build script need to update to somehow accept a release tag value and pass it along when it's specified.

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -1863,6 +1863,7 @@ function Start-DevPowerShell {
 <#
 .EXAMPLE
 PS C:> Copy-MappedFiles -PslMonadRoot .\src\monad
+
 copy files FROM .\src\monad (old location of submodule) TO src/<project> folders
 #>
 function Copy-MappedFiles {
--- a/build.psm1
+++ b/build.psm1
@@ -1859,6 +1859,83 @@ function Start-DevPowerShell {
     }
 }
 
+
+<#
+.EXAMPLE
+PS C:> Copy-MappedFiles -PslMonadRoot .\src\monad
+copy files FROM .\src\monad (old location of submodule) TO src/<project> folders
+#>
+function Copy-MappedFiles {
+
+    [CmdletBinding()]
+    param(
+        [Parameter(ValueFromPipeline=$true)]
+        [string[]]$Path = "$PSScriptRoot",
+        [Parameter(Mandatory=$true)]
+        [string]$PslMonadRoot,
+        [switch]$Force,
+        [switch]$WhatIf
+    )
+
+    begin {
+        function MaybeTerminatingWarning {
+            param([string]$Message)
+
+            if ($Force) {
+                Write-Warning "$Message : ignoring (-Force)"
+            } elseif ($WhatIf) {
+                Write-Warning "$Message : ignoring (-WhatIf)"
+            } else {
+                throw "$Message : use -Force to ignore"
+            }
+        }
+
+        if (-not (Test-Path -PathType Container $PslMonadRoot)) {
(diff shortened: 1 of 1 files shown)


build.psm1
```diff
@@ -1206,8 +1216,14 @@ function Start-PSBootstrap {
 function Start-PSPackage {
     [CmdletBinding()]param(
         # PowerShell packages use Semantic Versioning http://semver.org/
+        [Parameter(ParameterSetName = "Version")]
         [string]$Version,
 
+        [Parameter(ParameterSetName = "ReleaseTag")]
+        [ValidatePattern("^v\d+\.\d+\.\d+-\w+\.\d+$")]
```
lzybkr: The `-\w+\.\d+` part should be optional - an official release won't have that part I think.

build.psm1
```diff
@@ -1838,84 +1859,6 @@ function Start-DevPowerShell {
...
 }
 
-
-<#
-.EXAMPLE
-PS C:> Copy-MappedFiles -PslMonadRoot .\src\monad
-
-copy files FROM .\src\monad (old location of submodule) TO src/<project> folders
-#>
-function Copy-MappedFiles {
```
lzybkr: Does this belong in a distinct change for another PR? Maybe also removing the json files and other functions like `Get-Mappings`?

build.psm1
daxian-dbw: True. Wil update the reg expression.

build.psm1
daxian-dbw: Sure thing, Actually @vors submitted a PR to do so: #3407. Maybe we need committee's approval to go ahead with that PR?

TravisEz13 (approved):

build.psm1
iSazonov: Such clean-ups will help us to move to MSBuild.

PR #3928, opened 2017-06-03
Update launch.json Attach config to use process picker.

rkeithhill:
Using the process picker is much better since if you are using VSCode you are likely to have more than one PowerShell process running.  With the picker, just type `pow` and the process list will filter down to the PowerShell processes.  Just pick the desired process to attach to based on the PID (or command line).

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -17,7 +17,7 @@
             "type": "coreclr",
             "request": "attach",
             "justMyCode": false,
-            "processName": "powershell"
+            "processId": "${command:pickProcess}"
         }
     ]
 }


iSazonov:
Great!

Is it possible to add a regex filter like "powershell*"?

rkeithhill:
Unfortunately, I don't believe so.  This is a feature of the C#/.NET Core debugger and if you specify `processId`, then you are not supposed to specify `processName`. 

Also, eventually you might want to attach to a host named something other than PowerShell.  :-)  Besides, when you press <kbd>F5</kbd> to attach, the process list will appear, type `pow` and the list will immediately filter down to just the powershell processes.

iSazonov:
LGTM.

Issue #3924, opened 2017-06-02
Start-PSBuild -SMAOnly not working


anmenaga:
-SMAOnly flag does Not work when building using Start-PSBuild.

Steps to reproduce
------------------
`Start-PSBuild -SMAOnly`

Expected behavior
-----------------
Build successful.

Actual behavior
---------------
Error:

> Assets file 'System.Management.Automation\obj\project.assets.json' doesn't have a target for '.NETCoreApp,Version=v2.0/win10-x64'

iSazonov:
I started to publish PRs to move the logic from Build.psm1 to csproj files and take advantages of MSBuild. This way, we get an optimized and fast re-build process.

vors:
This should be fixed already, not sure why it didn't auto close this issue before.

PR #3930, opened 2017-06-03
Fix Start-PSBuild -SMAOnly with the latest dotnet cli

vors:
dotnet starts to error out because runtime is passed

I was trying to find a good place in the docs to document `Start-PSBuild -SMAOnly`, but cannot find it. @SteveL-MSFT do you have suggestions?

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -239,7 +239,11 @@ function Start-PSBuild {
 
     $Arguments += "--configuration", $Options.Configuration
     $Arguments += "--framework", $Options.Framework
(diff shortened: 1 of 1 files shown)


vors:
Fixes #3924

SteveL-MSFT:
@vors we should probably have comment based help

vors:
Yes, probably that's the right way to go about it.
Does this PR need any changes?

daxian-dbw (approved):

PR #3933, opened 2017-06-05
Add 'CSharpVersion7' member into public 'Language' enum

### LINKED ISSUE #3853: Add-Type: add CSharpVersion7
Should we add new CSharpVersion7 option in [Language](https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs#L43) enum in Add-Type cmdlet?

And why we use "#if CORECLR" in the enum?

iSazonov:
Close #3853 

Public enum 'LanguageVersion' from 'Microsoft.CodeAnalysis.CSharp'
already contains 'CSharp7' member.
After moving PowerShell to .Net Core 2.0 we should add 'CSharpVersion7'
member into public 'Microsoft.PowerShell.Commands.Language' enum too.

This will allow 'Add-Type -Language CSharpVersion7'

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/Add-Type.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/Add-Type.Tests.ps1
@@ -1,6 +1,10 @@
 $guid = [Guid]::NewGuid().ToString().Replace("-","")
 
 Describe "Add-Type" -Tags "CI" {
+    It "Public 'Language' enumeration contains all members" {
+        [Enum]::GetNames("Microsoft.PowerShell.Commands.Language") -join "," | Should Be "CSharp,CSharpVersion7,CSharpVersion6,CSharpVersion5,CSharpVersion4,CSharpVersion3,CSharpVersion2,CSharpVersion1,VisualBasic,JScript"
+    }
+
     It "Should not throw given a simple class definition" {
         { Add-Type -TypeDefinition "public static class foo { }" } | Should Not Throw
     }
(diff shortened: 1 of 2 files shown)


test/powershell/Modules/Microsoft.PowerShell.Utility/Add-Type.Tests.ps1
```diff
@@ -1,6 +1,10 @@
 $guid = [Guid]::NewGuid().ToString().Replace("-","")
 
 Describe "Add-Type" -Tags "CI" {
+    It "Public 'Language' enumeration contains all members" {
+        [Enum]::GetNames("Microsoft.PowerShell.Commands.Language") -join "," | Should Be "CSharp,CSharpVersion7,CSharpVersion6,CSharpVersion5,CSharpVersion4,CSharpVersion3,CSharpVersion2,CSharpVersion1,VisualBasic,JScript"
+    }
+
```
Francisco-Gamino: Should we add a test case for 'Add-Type -Language CSharpVersion7'?

Francisco-Gamino: Also, should there be two commits (one for the product change and one for the test)?

iSazonov: We checked it manually in the original Issue.
I didn't touch the functionality of this cmdlet. Besides, we don't even have tests for it right now. So I believe this is a separate huge job.
It seems @daxian-dbw is working on Add-Type. It would be good to agree on the distribution of work on tests.

I can say nothing about separate commits. 😕

iSazonov:
@mirichmo Could you please continue with the PR?

test/powershell/Modules/Microsoft.PowerShell.Utility/Add-Type.Tests.ps1
mirichmo: Two commits (one for code changes and one for tests) is not a strict requirement.

mirichmo (approved):

PR #3934, opened 2017-06-05
Enhance '#requires' directive to check OS

### LINKED ISSUE #3751: Add support for OS check to the #requires statement
PowerShell Core is a cross-platform, so it makes sense to add support for operating system check to the #requires statement in case you want to write an OS-specific script.

iSazonov:
Close #3751

Motivation
---
PowerShell Core is a cross-platform, so a module author can write a module for only specific platform.

Enhancement
---
Now in a module import time '#requires' support:

#requires -OS Windows
#requires -OS Linux
#requires -OS OSX
#requires -OS Unix
#requires -OS Inbox
#requires -OS IoT
#requires -OS Nano

Here:
 - Unix is Linux or OSX
 - InBox is IoT or Nano
 - Nano is Nano Server

Additional considerations
---
Our CI systems don't run tests for IoT and Nano so we have to run it in a special way.

### DIFF
--- a/src/System.Management.Automation/engine/CommandDiscovery.cs
+++ b/src/System.Management.Automation/engine/CommandDiscovery.cs
@@ -558,11 +558,55 @@ namespace System.Management.Automation
         // #Requires -Module
         internal static void VerifyScriptRequirements(ExternalScriptInfo scriptInfo, ExecutionContext context)
         {
+            VerifyOS(scriptInfo);
             VerifyElevatedPrivileges(scriptInfo);
             VerifyPSVersion(scriptInfo);
             VerifyPSEdition(scriptInfo);
             VerifyRequiredModules(scriptInfo, context);
         }
+
+        internal static void VerifyOS(ExternalScriptInfo scriptInfo)
+        {
+            string requiredOS = scriptInfo.RequiredOS;
+            bool trueOS = false;
+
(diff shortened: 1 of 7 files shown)


rkeithhill:
Does this allow for:
```posh
#Requires -OS Windows,Linux
```
Also, I thought the preferred term is `macOS` instead of `OSX`?

iSazonov:
> #Requires -OS Windows,Linux

No, while only single value. It is good for a discussion.

>  macOS  instead of  OSX 

I use names that we already use (`$isOSX`). No problem move to `macOS` if it will be approved.

iSazonov:
@SteveL-MSFT Have we conclusion from PowerShell-Committee?

SteveL-MSFT:
@iSazonov we had to cancel the committee meeting last week due to too many committee members being out, we're meeting today

joeyaiello:
@PowerShell/powershell-committee discussed this and we'd rather expose this sort of behavior through other means. I still have a half-written RFC that has yet to come out (and I apologize for the delay there), but it's unlikely that we want this behavior. (Among other reasons, we don't want to completely block execution, e.g. of `#requires -OS Linux` on macOS.)

Again, until that RFC comes out, I think wrapping everything in an `$IsLinux` conditional is an acceptable workaround.

Issue #3212, opened 2017-02-27
Rewrite ConvertFrom-SddlString in C#


lzybkr:
See #2718.

faraazahmad:
I would like to take this up

vexx32:
@lzybkr the code here mentions that certain libraries/methods/classes used by the code are not available in CoreCLR. Would we need to reimplement that, or is this pending support from .NET Core itself?

SteveL-MSFT:
@vexx32 I looked at the script, there are some specific rights not supported in .Net Core.  I think we would just skip those and add a comment in the code in case they do show up in .Net Core in the future.

SteveL-MSFT:
This was done as part of https://github.com/PowerShell/PowerShell/pull/8341

PR #3936, opened 2017-06-05
Rewrite convertfrom sddlstring in csharp

zacateras:
#3212


src/System.Management.Automation/System.Management.Automation.csproj
```diff
@@ -25,6 +25,7 @@
     <PackageReference Include="System.Security.AccessControl" Version="4.4.0-preview1-25302-01" />
     <PackageReference Include="System.Security.Cryptography.Pkcs" Version="4.4.0-preview1-25302-01" />
     <PackageReference Include="System.Text.Encoding.CodePages" Version="4.4.0-preview1-25302-01" />
+    <PackageReference Include="System.Threading.AccessControl" Version="4.4.0-preview1-25302-01" />
```
zacateras: this is required by types moved from System.Security.AccessControl to System.Threading.AccessControl (System.Security.AccessControl.MutexRights), am I allowed to add new references here ?

daxian-dbw: It looks to me the package `System.Threading.AccessControl` is required by `Microsoft.PowerShell.Commands.Utility` instead of `System.Management.Automation`. If that's the case, this `PackageReference` entry should be added to `Microosft.PowerShell.Commands.Utility` instead. And please remove the same entry from `Microsoft.PowerShell.SDK.csproj`.

daxian-dbw:
@iSazonov if you have time, could you please review this PR?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
```diff
@@ -0,0 +1,187 @@
...
+            // enumerations have the same value but with different names.
+            HashSet<long> foundValues = new HashSet<long>();
+
+            // Go through the entries in the different right types, and see if they apply to the
+            // provided access mask. If they do, then add that to the result.
+            foreach (var rightType in typesToExamine)
+            {
+                foreach (var accessFlag in Enum.GetNames(rightType))
+                {
+                    long longKeyValue = (long)Enum.Parse(rightType, accessFlag);
```
iSazonov: Oh, we re-evaluate Enum.GetNames and Enum.Parse again and again. We should refactor "GetAccessRights" to use cache(s).

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            List<string> aceStrings = new List<string>();
+
+            foreach (var ace in acl.OfType<KnownAce>())
+            {
+                string aceString = ConvertToNtAccount(ace.SecurityIdentifier) + ": ";
+
+                if (ace is QualifiedAce qualifiedAce)
+                {
+                    aceString += qualifiedAce.AceQualifier;
```
iSazonov: We should use StringBuilder because `aceString` is changed multiple times.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
```diff
@@ -0,0 +1,187 @@
...
+                    }
+                }
+
+                aceStrings.Add(aceString);
+            }
+
+            return aceStrings;
+        }
+    }
+}
```
iSazonov: Please add Newline.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/UtilityCommon.cs
```diff
@@ -88,6 +88,11 @@ public static class UtilityResources
         /// The file '{0}' could not be parsed as a PowerShell Data File.
         /// </summary>
         public static string CouldNotParseAsPowerShellDataFile { get { return UtilityCommonStrings.CouldNotParseAsPowerShellDataFile; } }
+
+        /// <summary>
+        /// '{0}' is not supported in this system.
+        /// </summary>
+        public static string TypeNotSupported { get { return UtilityCommonStrings.TypeNotSupported; } }
```
iSazonov: It seems we can directly use `UtilityCommonStrings.TypeNotSupported`

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
```diff
@@ -0,0 +1,187 @@
...
+            List<string> foundAccess = new List<string>();
+
+            // Store the access types we've already seen, so that we don't report access
+            // flags that are essentially duplicate. Many of the access values in the different
+            // enumerations have the same value but with different names.
+            HashSet<long> foundValues = new HashSet<long>();
+
+            // Go through the entries in the different right types, and see if they apply to the
+            // provided access mask. If they do, then add that to the result.
+            foreach (var rightType in typesToExamine)
```
iSazonov: @SteveL-MSFT @daxian-dbw We should make a conclusion about the algorithm. SDDL is always "object-specific" but we go through all "types". I believe that's a mistake. An user always know the object type (where he did get the SDDL), should set it explicit and the cmdlet must decode only against this type.

iSazonov (commented):
Leave a comment

iSazonov:
@zacateras Thanks for your work! I have question about the algorithm that I addressed to PowerShell team - please wait the conclusion before continue.

Also we haven't tests for the cmdlet - you should add its and pass CI before we can merge.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
lzybkr: We could write a script to generate C# and avoid reflection entirely.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
lzybkr: I don't understand this domain enough to be confident here - but in principle I fully agree with @iSazonov - the current algorithm is strange.

lzybkr (commented):
Thanks for the contribution - I'll be very happy to see this one merged.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
iSazonov: @lzybkr Could you please ping an expert in this domain?

src/Modules/Shared/Microsoft.PowerShell.Utility/Microsoft.PowerShell.Utility.psm1
```diff
@@ -1,174 +0,0 @@
-
```
zacateras: this file name is referenced 3 times:
* /src/Modules/map.json (remove entry?)
* /src/System.Management.Automation/engine/InitialSessionState.cs (replace with some other build result file?)
* /src/System.Management.Automation/security/Authenticode.cs (replace with some other build result file?)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
```diff
@@ -0,0 +1,187 @@
...
+            CommonSecurityDescriptor rawSecurityDescriptor = new CommonSecurityDescriptor(false, false, Sddl);
+
+            string owner = ConvertToNtAccount(rawSecurityDescriptor.Owner); 
+            string group = ConvertToNtAccount(rawSecurityDescriptor.Group);
+            List<string> discretionaryAcl = ConvertToAceString(rawSecurityDescriptor.DiscretionaryAcl, Type);
+            List<string> systemAcl = ConvertToAceString(rawSecurityDescriptor.SystemAcl, Type);
+
+            PSObject result = new PSObject();
+
+            result.Properties.Add(new PSNoteProperty("Owner", owner));
```
lzybkr: Following up on my own question and after playing around with this cmdlet a little, we definitely want strongly typed output.

src/Modules/Shared/Microsoft.PowerShell.Utility/Microsoft.PowerShell.Utility.psm1
lzybkr: Thanks for catching that.
Remove the entry in `map.json` - that's perfectly safe.
I think you can change the other two references to `Microsoft.PowerShell.Utility.psd1`.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
lzybkr: The [documentation for SDDL](https://msdn.microsoft.com/en-us/library/windows/desktop/aa379567(v=vs.85).aspx) looks pretty good, I think if we have concerns about this algorithm, we can answer them ourselves, but I don't have time for that now.

SteveL-MSFT (changes requested):
Can you add some tests?  I just cloned this branch and a basic attempt to use this cmdlet failed:

```powershell
PS C:\users\slee\repos\zacateras> $sddl = "O:BAG:BAD:(A;CI;CCDCLCSWRPWPRCWD;;;BA)(A;CI;CCDCRP;;;NS)(A;CI;CCDCRP;;;LS)(A;
CI;CCDCRP;;;AU)"
PS C:\users\slee\repos\zacateras> ConvertFrom-SddlString -Sddl $sddl
ConvertFrom-SddlString : Unable to cast object of type 'System.Security.AccessControl.FileSystemRights' to type
'System.Int64'.
At line:1 char:1
+ ConvertFrom-SddlString -Sddl $sddl
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [ConvertFrom-SddlString], InvalidCastException
    + FullyQualifiedErrorId : System.InvalidCastException,Microsoft.PowerShell.Commands.ConvertFromSddlStringCommand
```

Which worked with the old script cmdlet:

```powershell
PS C:\users\slee\repos\PowerShell> $sddl = "O:BAG:BAD:(A;CI;CCDCLCSWRPWPRCWD;;;BA)(A;CI;CCDCRP;;;NS)(A;CI;CCDCRP;;;LS)(A
;CI;CCDCRP;;;AU)"
PS C:\users\slee\repos\PowerShell> ConvertFrom-SddlString -Sddl $sddl


Owner            : BUILTIN\Administrators
Group            : BUILTIN\Administrators
DiscretionaryAcl : {NT AUTHORITY\Authenticated Users: AccessAllowed (ListDirectory, WriteData,
                   WriteExtendedAttributes), NT AUTHORITY\LOCAL SERVICE: AccessAllowed (ListDirectory, WriteData,
                   WriteExtendedAttributes), NT AUTHORITY\NETWORK SERVICE: AccessAllowed (ListDirectory, WriteData,
                   WriteExtendedAttributes), BUILTIN\Administrators: AccessAllowed (ChangePermissions,
                   CreateDirectories, ExecuteKey, ListDirectory, ReadExtendedAttributes, ReadPermissions, Traverse,
                   WriteData, WriteExtendedAttributes, WriteKey)}
SystemAcl        : {}
RawDescriptor    : System.Security.AccessControl.CommonSecurityDescriptor
```

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
zacateras: @lzybkr what kind of code generation script do you mean? t4 template?

lzybkr: No, I just meant a PowerShell script that generates switch statements or something like that.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
iSazonov: I re-read the document and made any tests - My conclusion is full discard the algorithm:

1. From the [docs](https://msdn.microsoft.com/en-us/library/windows/desktop/aa379557%28v=vs.85%29.aspx):
> Each type of securable object defines its own set of specific access rights and its own mapping of generic access rights. For information about the specific and generic access rights for each type of securable object, see the overview for that type of object.

- the conclusion is  an SDDL has no meaning without object entity - the cmdlet should get the object type as argument.

2. The cmdlet calls  [Security.AccessControl.CommonSecurityDescriptor]::new($false,$false,$Sddl)
[The constructor](https://github.com/dotnet/corefx/blob/0e4095bc5da407df6cbe0aa350e1b2ccf32696a4/src/System.Security.AccessControl/src/System/Security/AccessControl/SecurityDescriptor.cs#L949) in Core FX

I tested the cmdlet with (1)  [Security.AccessControl.CommonSecurityDescriptor]::new($false,$false,$Sddl) and (2)  [Security.AccessControl.CommonSecurityDescriptor]::new($true,$false,$Sddl) for path "c:\test" (it is container) and got different results.

- the conclusion is the cmdlet should take into account `isContainer` (folders/containers) and `isDS` (directory objects and containers).

3. CoreFX can throw exceptions (ArgumentException) - maybe the cmdlet should catch them and re-throw (put CoreFX exception in InnerException)

4.  The cmdlet should have strongly typed output. In current PSCustomObject we should add "Type".

5. "Type" - we can define it as Enum ( remove unsupported member by #if !Unix ) and exclude ValidateSet at all.

6. We should add tests for all object types.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
```diff
@@ -0,0 +1,187 @@
...
+        public string Sddl { get; set; }
+
+        /// <summary>
+        /// The type of rights that this SDDL string represents, if any.
+        /// </summary>
+        [Parameter()]
+        [ValidateSet(
+            "FileSystemRights", "RegistryRights", "ActiveDirectoryRights",
+            "MutexRights", "SemaphoreRights", "CryptoKeyRights",
+            "EventWaitHandleRights")]
```
zacateras: used custom enum instead

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
```diff
@@ -0,0 +1,187 @@
...
+        public string Type { get; set; }
+
+        /// <summary>
+        /// Implements the BeginProcessing method for the ConvertFrom-SddlString command.
+        /// </summary>
+        protected override void BeginProcessing()
+        {
+#if CORECLR
+            if (Type == "CryptoKeyRights" || Type == "ActiveDirectoryRights")
```
zacateras: thanks, done

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
```diff
@@ -0,0 +1,187 @@
...
+            return null;
+        }
+
+        /// <summary>
+        /// Gets the access rights that apply to an access mask, preferring right types
+        /// of 'Type' if specified.
+        /// </summary>
+        private List<string> GetAccessRights(long accessMask, string type)
+        {
+            IDictionary<string, Type> rightTypes = new Dictionary<string, Type>
```
zacateras: added static cache field

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
```diff
@@ -0,0 +1,187 @@
...
+            {
+                { "FileSystemRights", typeof(System.Security.AccessControl.FileSystemRights) },
+                { "RegistryRights", typeof(System.Security.AccessControl.RegistryRights) },
+#if !CORECLR
+                { "ActiveDirectoryRights", typeof(System.DirectoryServices.ActiveDirectoryRights) },
```
zacateras: done

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
zacateras: added ps1 generator

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
```diff
@@ -0,0 +1,187 @@
...
+                        foundValues.Add(longKeyValue);
+                        if ((accessMask & longKeyValue) == longKeyValue)
+                        {
+                            foundAccess.Add(accessFlag);
+                        }
+                    }
+                }
+            }
+
+            foundAccess.Sort();
```
zacateras: used StringComparer.OrdinalIgnoreCase

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
zacateras: done

src/Microsoft.PowerShell.Commands.Utility/commands/utility/UtilityCommon.cs
zacateras: removed obsolete property

src/Modules/Shared/Microsoft.PowerShell.Utility/Microsoft.PowerShell.Utility.psm1
zacateras: done

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
zacateras: removed ValidateSet, used custom enum instead

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
zacateras: ok

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
zacateras: @iSazonov I also have doubts about current method of converting ACE AccessMask into access flag enum based strings. There are situations when flag cannot be converted without additional logic i.e. [FileSystemRights](https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem.AccessControl/ref/System.IO.FileSystem.AccessControl.cs#L83&L85) contains entries with duplicate flag values - some applies only for files, other only for containers. I guess that this won't be solved by requiring Type parameter.

iSazonov: @zacateras Please see (2) point from my conclusion.
> •the conclusion is the cmdlet should take into account  isContainer  (folders/containers) and  isDS  (directory objects and containers).

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
iSazonov: @lzybkr Will this require recompilation if there will are changes in CoreFX? We could initilize this in static cache on Begin step.

SteveL-MSFT (approved):

iSazonov:
@SteveL-MSFT The algorithm is broken! Please see my comment above https://github.com/PowerShell/PowerShell/pull/3936#discussion_r121588232 I believe we cannot merge the PR.

SteveL-MSFT (changes requested):
Reviewing algorithm

iSazonov:
@daxian-dbw Can we use [System.DirectoryService](https://dotnet.myget.org/feed/dotnet-core/package/nuget/System.DirectoryServices) or before we should move to 4.5.0 version ?

iSazonov:
@zacateras Do you plan to continue?

zacateras:
@iSazonov yep. I thought that we are waiting for some external review.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFrom-SddlString.cs
zacateras: @iSazonov Apart from fixing CommonSecurityDescriptor constructor call (by requiring isContainer and isDS flags) we should also take care about correct parsing of AccessMasks to enum names. FileSystemRights and ActiveDirectoryRights contain several names that use the same value, eg. FileSystemRights contain AppendData & CreateDirectories which have [the same value (4)](https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem.AccessControl/ref/System.IO.FileSystem.AccessControl.cs#L83&L85). In consequence Directories with access flag (4) can get AppendData right, what is incorrect.

As a solution we can add custom enums (or at least custom mapping dictionaries) containing distinct subsets of allowed values - one for containers (ContainerSystemRights) and one for other objects (FileSystemRights). Alternatively we can discard CommonSecurityDescriptor/Right-Enum based parsing algorithm entirely and try to implement text parser using sddl docs. What do you think ?

SteveL-MSFT:
Now that Beta.5 is out, I'll spend time on this tomorrow

iSazonov:
We should follow MSDN docs - see my comment above.

iSazonov:
@SteveL-MSFT Could you please continue?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromSddlStringCommand.RightType.Generate.ps1
```diff
@@ -0,0 +1,119 @@
+$RightTypes = @(
```
SteveL-MSFT: Was this file meant to be checked in?  It's not part of the build to run and the resulting source doesn't indicate it shouldn't be edited.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromSddlStringCommand.RightType.cs
```diff
@@ -0,0 +1,189 @@
...
+            /// <see cref="System.Security.AccessControl.MutexRights"/> type rights.
+            /// </summary>
+            MutexRights,
+
+            /// <summary>
+            /// <see cref="System.Security.AccessControl.SemaphoreRights"/> type rights.
+            /// </summary>
+            SemaphoreRights,
+
+#if !CORECLR
```
SteveL-MSFT: since we're officially removing fullclr support, we should not include this

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromSddlStringCommand.RightType.cs
```diff
@@ -0,0 +1,189 @@
...
+                        { 2, "Modify" },
+                        { 65536, "Delete" },
+                        { 131072, "ReadPermissions" },
+                        { 262144, "ChangePermissions" },
+                        { 524288, "TakeOwnership" },
+                        { 1048576, "Synchronize" },
+                        { 2031619, "FullControl" },
+                    }
+                },
+#if !CORECLR
```
SteveL-MSFT: since we're officially removing fullclr support, we should not include this

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertFrom-SddlString.Tests.ps1
```diff
@@ -0,0 +1,21 @@
+Describe "ConvertFrom-SddlString" -Tags "CI" {
+    It "Should convert without type" -Skip:($IsLinux -Or $IsOSX) {
```
SteveL-MSFT: Since this cmdlet is not supported on non-Windows, you should skip them in bulk using this pattern https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/WritingPesterTests.md#skipping-tests-in-bulk (that way if additional tests get added they get auto-skipped)

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertFrom-SddlString.Tests.ps1
```diff
@@ -0,0 +1,21 @@
...
+        $SddlStringInfo = ConvertFrom-SddlString "O:BAG:BAD:(A;CI;CCDCLCSWRPWPRCWD;;;BA)(A;CI;CCDCRP;;;NS)(A;CI;CCDCRP;;;LS)(A;CI;CCDCRP;;;AU)"
+
+        $SddlStringInfo.Owner | Should Be "BUILTIN\Administrators"
+        $SddlStringInfo.Group | Should Be "BUILTIN\Administrators"
+        $SddlStringInfo.DiscretionaryAcl.Length | Should BeGreaterThan 0
+        $SddlStringInfo.SystemAcl.Length | Should Be $null
+        $SddlStringInfo.RawDescriptor | Should BeOfType [System.Security.AccessControl.CommonSecurityDescriptor]
```
SteveL-MSFT: Should be some tests for `-Type` parameter

src/Microsoft.PowerShell.Commands.Utility/Microsoft.PowerShell.Commands.Utility.csproj
```diff
@@ -80,6 +80,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="2.2.0" />
+    <PackageReference Include="System.DirectoryServices" Version="4.0.0-preview3-25423-02" />
```
SteveL-MSFT: This doesn't appear to be a supported package, perhaps we should not support AD rights for now.

SteveL-MSFT (changes requested):

SteveL-MSFT:
Sorry this took so long, got randomized by other things.  I also compared the result with 

```powershell
ConvertFrom-SddlString "D:(A;;FA;;;SY)(A;;FA;;;BA)"
```

```
# old script version
DiscretionaryAcl : {NT AUTHORITY\SYSTEM: AccessAllowed (ChangePermissions, CreateDirectories, Delete, DeleteSubdirectoriesAndFiles, ExecuteKey, FullControl,
                   FullControl, FullControl, FullControl, ListDirectory, Modify, Read, ReadAndExecute, ReadAttributes, ReadExtendedAttributes,
                   ReadPermissions, Synchronize, TakeOwnership, Traverse, Write, WriteAttributes, WriteData, WriteExtendedAttributes, WriteKey),
                   BUILTIN\Administrators: AccessAllowed (ChangePermissions, CreateDirectories, Delete, DeleteSubdirectoriesAndFiles, ExecuteKey, FullControl,
                   FullControl, FullControl, FullControl, ListDirectory, Modify, Read, ReadAndExecute, ReadAttributes, ReadExtendedAttributes,
                   ReadPermissions, Synchronize, TakeOwnership, Traverse, Write, WriteAttributes, WriteData, WriteExtendedAttributes, WriteKey)}
```

```powershell
# c# version
DiscretionaryAcl : {NT AUTHORITY\SYSTEM: AccessAllowed (AppendData, ChangePermissions, CreateFiles, Delete, DeleteSubdirectoriesAndFiles, ExecuteFile,
                   FullControl, FullControl, FullControl, FullControl, Modify, Read, ReadAndExecute, ReadAttributes, ReadData, ReadExtendedAttributes,
                   ReadKey, ReadPermissions, Synchronize, TakeOwnership, Write, WriteAttributes, WriteExtendedAttributes, WriteKey), BUILTIN\Administrators:
                   AccessAllowed (AppendData, ChangePermissions, CreateFiles, Delete, DeleteSubdirectoriesAndFiles, ExecuteFile, FullControl, FullControl,
                   FullControl, FullControl, Modify, Read, ReadAndExecute, ReadAttributes, ReadData, ReadExtendedAttributes, ReadKey, ReadPermissions,
                   Synchronize, TakeOwnership, Write, WriteAttributes, WriteExtendedAttributes, WriteKey)}
```

And the newer version has more permissions (although not sure if the old one was wrong).

iSazonov:
@SteveL-MSFT Did you look [my comment](https://github.com/PowerShell/PowerShell/pull/3936#discussion_r121588232) above? Main is "SDDL has no meaning without object entity - the cmdlet should get the object type as argument."

src/Microsoft.PowerShell.Commands.Utility/Microsoft.PowerShell.Commands.Utility.csproj
iSazonov: Official package is  [System.DirectoryService](https://dotnet.myget.org/feed/dotnet-core/package/nuget/System.DirectoryServices). It is 4.5.0 version. @daxian-dbw said that we should review the API before use.

SteveL-MSFT:
@iSazonov Seems like if you have the object, you would be using `Get-Acl`.  `ConvertFrom-SDDLString` is more likely used when you have just the SDDL and just want to understand (perhaps roughly) what it represents.  Not sure how useful this is if you require the object.

iSazonov:
@SteveL-MSFT My idea is that we should make `-Type` mandatory.

zacateras:
@iSazonov If AccessMask is presented using [well known aliases](https://github.com/tpn/winsdk-10/blob/master/Include/10.0.10240.0/shared/sddl.h#L126) then it can be parsed entirely without the knowledge of securable object type (Type parameter).

In case of numerical AccessMasks (eg. 0x02142512) we can translate only higher 16-32bits [generic part](https://msdn.microsoft.com/pl-pl/library/windows/desktop/aa374896(v=vs.85).aspx) if the Type is not provided (lower bits can remain unparsed as a reminder). If the Type parameter is present then specific rights scheme could be used - according to constant definitions in [winnt.h](https://source.winehq.org/source/include/winnt.h). Also more securable object types can be considered - according to the list in [docs](https://msdn.microsoft.com/en-us/library/aa379557(VS.85).aspx).

I started an implementation of SDDL parser using the documentation. The code can be found [here](https://github.com/zacateras/sddl-parser) (with examples [here](https://github.com/zacateras/sddl-parser/blob/master/test/Sddl.Parser.Tests/SddlTests.cs)). If we want to use parts of the implemented approach then I can integrate its simplified form into the command.

What do you think about such solution?

iSazonov:
@zacateras I still don't understand the need for this heuristic. For which scenarios is it addressed? As for me it looks like a trasformation enum1 -> int -> enum2 - what's confusing.

daxian-dbw:
This PR has been left unattended for a long time. I will close this PR for now.
@zacateras Please feel free to reopen it if you decide to continue on it at a later time.

PR #3939, opened 2017-06-05
set default parameter set of Start-PSPackge

TravisEz13:
Restore previous behavior when no parameter is specified.


daxian-dbw (approved):

SteveL-MSFT (approved):
LGTM

Issue #3789, opened 2017-05-16
New-ModuleManifest creates files in UTF-16


andrewducker:
Steps to reproduce
------------------
1) Run New-ModuleManifest
2) Discover that it's created in UTF-16
3) And that this means that Git will treat it as a binary, and not diff it.

Expected behavior
-----------------
New-ModuleManifest creates files in UTF8

Actual behavior
---------------
![image](https://cloud.githubusercontent.com/assets/91378/26094938/ad30b090-3a13-11e7-93f9-7c6b38f8556e.png)

iSazonov:
@andrewducker Thanks for the report!

Related #1559

BrucePay:
Perhaps this should be included/mentioned in the encoding RFC.

anmenaga:
Submitted doc PR that mentions New-ModuleManifest in encoding RFC.

joeyaiello:
Blocked on changes from @JamesWTruher

anmenaga:
@joeyaiello this was fixed and merged previously. Jim's changes ( #4119 ) are separate and overwrite this.

Issue #1979, opened 2016-08-20
Decouple the bundled modules to seperate Repo's


kilasuit:
Now that we have PowerShell Open Sourced we should really move to decouple all of the modules that are shipped "inbox" into their own repos so that updates to these can be make separately to the core engine and can also be pulled from the PowerShell Gallery

However this brings a number of additional challenges that warrants a detailed discussion and a plan of action as not to break the current build process.

Modules that should be decoupled and pulled in as submodules
- [x]  - PowerShellGet **Priority**
- [x]  - PackageManagement (this may already be out there in the OneGet Repo)
- [x]  - Microsoft.PowerShell.ODataUtils
- [x]  - PSReadline (already decoupled just needs to become a submodule)
- [x]  - PowerShell.Archive (as above)
- [x]  - Pester (as above but submodule to point to Pester not current submodule)
- [ ]  - CimCmdlets
- [ ]  - Microsoft.PowerShell.Diagnostics
- [ ]  - Microsoft.PowerShell.LocalAccounts
- [ ]  - Microsoft.PowerShell.Management
- [ ]  - Microsoft.PowerShell.Security
- [ ]  - Microsoft.WSMan.Management
- [ ]  - PowerShell.Utility
- [ ]  - PSDiagnostics
- [ ]  - PSScheduledJob
- [ ]  - PSWorkflow
- [ ]  - PSWorkflowUtility 

This is especially more important as in the differing subfolders in https://github.com/PowerShell/PowerShell/tree/master/src/Modules there are a number of code duplications due to the modules residing in multiple folders which is rather unnecessary and adds unneeded complexity for management of this repository going forward.

lzybkr:
We have definite plans to do this for some of the modules like PackageManagement, PowerShellGet, Pester, and PSReadline.

Utility and Management might not be worth the hassle because of implicit or explicit circular dependencies.

kilasuit:
I thought it may have been "on the plans" but couldn't see an issue for it so felt it best to raise one for visibility

Carringguns:
Thanks im not free and clear yet. Little cleanup still. But want go to the source and offer you the data. and its running state im keeping in a loop.

Carringguns:
Its design to eexplore when mission complete. Love this one. Better than most ive seen

vors:
I will take care about 

- [ ] PSReadLine
- [ ] Archive

The current proposition is the following:
Make sure that upstream version is suitable as a drop-in replacement.
[...]

PR #3940, opened 2017-06-05
Change New-ModuleManifest encoding to UTF8NoBOM on non-Windows platforms

anmenaga:
This addresses #3789 
Currently New-ModuleManifest creates psd1 manifests in UTF-16 with BOM; this creates problems for Linux tools and Git diff'er.
This fix changes encoding of New-ModuleManifest to be UTF8 (no BOM).
Also mentioned New-ModuleManifest in "[RFC 0020 DefaultFileEncoding](https://github.com/PowerShell/PowerShell-RFC/blob/master/1-Draft/RFC0020-DefaultFileEncoding.md)".

### DIFF
--- a/src/System.Management.Automation/engine/Modules/NewModuleManifestCommand.cs
+++ b/src/System.Management.Automation/engine/Modules/NewModuleManifestCommand.cs
@@ -941,7 +941,7 @@ namespace Microsoft.PowerShell.Commands
                 PathUtils.MasterStreamOpen(
                     this,
                     filePath,
-                    EncodingConversion.Unicode,
+                    new UTF8Encoding(false), // UTF-8, no BOM
                     /* defaultEncoding */ false,
                     /* Append */ false,
                     /* Force */ false,
--- a/src/System.Management.Automation/engine/Modules/NewModuleManifestCommand.cs
+++ b/src/System.Management.Automation/engine/Modules/NewModuleManifestCommand.cs
@@ -941,7 +941,11 @@ namespace Microsoft.PowerShell.Commands
                 PathUtils.MasterStreamOpen(
                     this,
                     filePath,
+#if UNIX
                     new UTF8Encoding(false), // UTF-8, no BOM
+#else
+                    EncodingConversion.Unicode, // UTF-16 with BOM
+#endif
                     /* defaultEncoding */ false,
                     /* Append */ false,
                     /* Force */ false,
@@ -1167,5 +1171,5 @@ namespace Microsoft.PowerShell.Commands
         }
     }
 
-    #endregion
+#endregion
 } // Microsoft.PowerShell.Commands
--- a/test/powershell/engine/Module/NewModuleManifest.Tests.ps1
+++ b/test/powershell/engine/Module/NewModuleManifest.Tests.ps1
@@ -20,6 +20,12 @@ Describe "New-ModuleManifest tests" -tags "CI" {
         $module.PrivateData.PSData.ProjectUri | Should BeExactly $absoluteUri
     }
 
+    It "Verify module manifest encoding" {
+        New-ModuleManifest -Path $testModulePath
+        # verify first line of the manifest - 3 bytes - '#' '\r' '\n' - in UTF-8 no BOM this should be @(35,13,10)
+        Get-Content -Encoding Byte -Path $testModulePath -TotalCount 3 | Should Be @(35,13,10)
+    }
+
     It "Relative URIs are not allowed" {
         $testUri = [Uri]"../foo"
 
--- a/test/powershell/engine/Module/NewModuleManifest.Tests.ps1
+++ b/test/powershell/engine/Module/NewModuleManifest.Tests.ps1
@@ -35,9 +35,9 @@ Describe "New-ModuleManifest tests" -tags "CI" {
 
     It "Verify module manifest encoding on non-Windows " -Skip:($IsWindows) {
         
-        # verify first line of the manifest - 3 characters - '#' '\r' '\n'
-        # On non-Windows platforms - in UTF-8 no BOM - this should be @(35,13,10)
-        TestNewModuleManifestEncoding -expected @(35,13,10)
+        # verify first line of the manifest - 2 characters - '#' '\n'
+        # On non-Windows platforms - in UTF-8 no BOM - this should be @(35,10)
+        TestNewModuleManifestEncoding -expected @(35,10)
     }
 
     It "Relative URIs are not allowed" {
--- a/test/powershell/engine/Module/NewModuleManifest.Tests.ps1
+++ b/test/powershell/engine/Module/NewModuleManifest.Tests.ps1
@@ -8,6 +8,17 @@ Describe "New-ModuleManifest tests" -tags "CI" {
         Remove-Item -Recurse -Force -ErrorAction SilentlyContinue testdrive:/module
     }
 
+    BeforeAll {
+        if ($IsWindows)
+        {
+            $ExpectedManifestBytes = @(255,254,35,0,13,0,10,0)
+        }
+        else
+        {
+            $ExpectedManifestBytes = @(35,10)
+        }
+    }
+
     It "Uris with spaces are allowed and escaped correctly" {
         $testUri = [Uri]"http://foo.com/hello world"
         $absoluteUri = $testUri.AbsoluteUri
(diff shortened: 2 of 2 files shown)


src/System.Management.Automation/engine/Modules/NewModuleManifestCommand.cs
```diff
@@ -941,7 +941,7 @@ protected override void EndProcessing()
                 PathUtils.MasterStreamOpen(
                     this,
                     filePath,
-                    EncodingConversion.Unicode,
+                    new UTF8Encoding(false), // UTF-8, no BOM
```
Francisco-Gamino: Do we need to update the documentation for this change?

joeyaiello:
As it currently stands, this is a breaking change (appropriate label added). 

Based on our current POR, we're okay changing the default *on non-Windows platforms*, but this shouldn't be changed (yet) on Windows until we've had the breaking change conversation. /cc @JamesWTruher

joeyaiello:
I understand that this currently breaks the `git` scenario, we just have to be thoughtful about this.

src/System.Management.Automation/engine/Modules/NewModuleManifestCommand.cs
SteveL-MSFT: @anmenaga says he opened a doc issue, can you link it here?

vors:
Are there plans to add UTF8-NoBom to powershell supported Encodings?
I.e. for `Set-Content` and friends.

Changing the most important scenarios is good, but it would be great to address the problem with encodings broader and consistently to help shape the conversation about breaking changes.

iSazonov:
@vors UTF8noBOM is discussed in [RFC 0020 DefaultFileEncoding](https://github.com/PowerShell/PowerShell-RFC/blob/master/1-Draft/RFC0020-DefaultFileEncoding.md).

iSazonov:
New-PSSessionConfigurationFile uses `EncodingConversion.Unicode` too.

joeyaiello:
@vors: first, yes. We're adding that (and making it the default across the board on Linux). The hard part is what to do with Windows (change the defaults? leave a legacy enum value for the "traditional" disparate defaults?) Which is why I want to have the discussion.

src/System.Management.Automation/engine/Modules/NewModuleManifestCommand.cs
anmenaga: [This PR](https://github.com/PowerShell/PowerShell-RFC/pull/95) in RFC repo.

anmenaga:
@joeyaiello good point;
PR updated to change encoding only for non-Windows.

SteveL-MSFT:
Is there an actual problem on Windows using UTF8NoBOM?  If only English is used, it's equivalent to an ascii file and Windows tools handle that just fine and I expect that to be the majority of use.

TravisEz13:
Restarted AppVeyor due to MyGet connectivity issue

joeyaiello:
@anmenaga if you could sync with @JamesWTruher before merging this, that'd be awesome. Assuming the @PowerShell/powershell-committee votes to accept tomorrow, we want to move to UTF8NoBOM everywhere.

iSazonov:
@SteveL-MSFT 
>Is there an actual problem on Windows using UTF8NoBOM? 

Users can add non-English comments in the manifest.

SteveL-MSFT:
@iSazonov absolutely, but I believe the majority of usage is English.  Even with non-ascii characters, my understanding is that notepad has heuristics to determine encoding to display it correctly.  VSCode seems to support UTF8NoBom just fine.

iSazonov:
>notepad has heuristics to determine encoding to display it correctly

It seem works on last versions only not on Windows 7/8(?).

TravisEz13:
@joeyaiello  Are you saying this is blocked on a sign off from @JamesWTruher ?

TravisEz13:
@joeyaiello @JamesWTruher  Do we want to document the output format of Cmdlets?

joeyaiello:
@TravisEz13 yes and yes. I think the uber-point is that @JamesWTruher is doing so much work here, that it might get changed anyway (and it might just muddle his extensive notes). 

Ultimately up to him, though. I just wanted to make sure everyone was on the same page. 

Regarding docs though: heck yes. We'll have a `Legacy` value for Encoding, and that will be well-documented.

JamesWTruher:
This is fine for now, I'll be changing all the cmdlets in response to https://github.com/PowerShell/PowerShell-RFC/blob/master/1-Draft/RFC0020-DefaultFileEncoding.md.

JamesWTruher (approved):
This is fine for now, I'll be changing all the cmdlets which provide for encoding as part of https://github.com/PowerShell/PowerShell-RFC/blob/master/1-Draft/RFC0020-DefaultFileEncoding.md. This doesn't muddy the waters for me.

test/powershell/engine/Module/NewModuleManifest.Tests.ps1
```diff
@@ -20,6 +20,26 @@ Describe "New-ModuleManifest tests" -tags "CI" {
         $module.PrivateData.PSData.ProjectUri | Should BeExactly $absoluteUri
     }
 
+    function TestNewModuleManifestEncoding {
+        param ([byte[]]$expected)
+        New-ModuleManifest -Path $testModulePath
+        (Get-Content -Encoding Byte -Path $testModulePath -TotalCount $expected.Length) -join ',' | Should Be ($expected -join ',')
+    }
+
+    It "Verify module manifest encoding on Windows " -Skip:(-not $IsWindows) {
```
TravisEz13: Could you just calculate the expected value in a `beforeall` then have one test for all platforms?

TravisEz13 (commented):
Please respond to my comment on your code.

TravisEz13:
@anmenaga Also, as a PowerShell team member, you need to submit the PR to update the docs before I merge this.  Please add a link to the PR here.  Thanks.

iSazonov:
For encoding changes, you might want to delay documenting as we expect big changes in this.

test/powershell/engine/Module/NewModuleManifest.Tests.ps1
anmenaga: Good idea; updated.

anmenaga:
@TravisEz13
1) [This PR](https://github.com/PowerShell/PowerShell-RFC/pull/95) in RFC repo was merged;
2) I've ran a search in PowerShell-Docs and didn't find any mentioning of encoding for New-ModuleManifest, so there is nothing to update there. Besides, this PR will be anyway overwritten by Jim's encoding work that is arriving soon.

SteveL-MSFT (dismissed):
We should update https://github.com/PowerShell/PowerShell/blob/master/docs/FAQ.md with the recommended settings in VSCode for automatic removing of trailing whitespace for resx files

TravisEz13:
@anmenaga Per the conversation in this PR, we want to add encoding documentation to the CmdLet docs.

anmenaga:
@TravisEz13 Here is the doc PR: https://github.com/PowerShell/PowerShell-Docs/pull/1405

kilasuit:
My only issue here is that although this is a change that I agree with, I think that this is where I have concerns of this causing a disparate and inconsistent behaviour between all versions of PowerShell below v6 due to the fact that New-ModuleManifest is baked into the **Microsoft.PowerShell.Core** module.

I personally would **much** prefer to see that this update to the cmdlet gets released in a manner that **_fully_** supports being used on downlevel systems where PowerShell v6 is unable to be installed.

However the only real solutions that I can really see available to be able to make that to happen would be to 
- Investigate the decoupling of a number of the cmdlets from the **Microsoft.PowerShell.Core** module into their own separate individual modules or a condensed updated singular module that has all the updates to any module in **Microsoft.PowerShell.Core** (not an overly fantastic solution) - similar in the idea put forward in #1979 
- Look to recreate this cmdlet within the PowerShellGet module.  
- Alongside the v6 updates look to release a completely updated and downlevel compatible version of the **Microsoft.PowerShell.Core** module to the PowerShell Gallery, which isn't really an overly viable solution either.

If this change could be easily replicated for downlevel versions of PowerShell I think that would be the best outcome all around.

TravisEz13:
@kilasuit Thanks for the feedback.  I think most of the concerns raised would be best tracked in #1979 or a new issue.

Issue #3743, opened 2017-05-09
Align PowerShell's CLI / startup behavior with POSIX-like shells such as Bash - command-line arguments


mklement0:
<sup>This is a follow-up to #3600.</sup>

To make PowerShell fit into the world of POSIX-like shells on Unix platforms, so that it can be a viable alternative, it must: 
* support certain command-line parameters (options) 
* exhibit certain default behaviors on startup

The [POSIX spec for the `sh` utility](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html) prescribes both.

Definitely not everything there is applicable to PowerShell, but I think it's important to at least fundamentally exhibit the same behavior and to document to what extent PowerShell is compliant.

A step has been made in the right direction by adding support for `-i` to start an interactive session - see #3558 

More work is needed, _some of which would result in breaking changes_.

The list is not complete, but hopefully covers the most important aspects.

* POSIX-like shells default to _not_ loading any initialization files _when commands are passed_ (whether via stdin or via `-c`), whereas PowerShell _always_ loads its profiles, the only exception being  explicit suppression with `-NoProfile`.

   * The POSIX behavior is the more sensible default: it makes for a more predictable execution environment and doesn't incur the cost of sourcing the initialization files.
   * Aligning PowerShell with this behavior would be a _breaking change_.
   * Arguably, when commands are combined with `-NoExit`, the profiles should still be loaded by default (note that this ability to execute commands on startup _and_ keep the shell open for interactive use is unique to PowerShell: POSIX-like shells invariably _exit_ after executing commands).
   * See #992, which has a [green-lit, but never-implemented committee decision](https://github.com/PowerShell/PowerShell/issues/992#issuecomment-384455760) to implement a _separate_ CLI entry point that doesn't load profiles, `pwsh-np`

* PowerShell already reads commands from stdin by default, and also when `-Command -` (which arguably should be `-File -`) is specified, but:
   * `-s` is the POSIX option for explicitly specifying standard input, and should be supported too.
   * `-s` allows combining stdin command input with passing _parameters_, which PowerShell doesn't support; for instance, `bash` supports something like:   
`echo 'echo "$# params"' | bash -s one two`, which yields `2 params`.

* `-Command` and `-File` issues:
  * They exhibit problematic interactive shell-like behaviors (with or without `-NonInteractive`) and lack consistent argument support - see #3223.

[...]

Issue #3223, opened 2017-02-27
Reconsider the behavior of powershell.exe -Command -: it shouldn't require terminating multi-line commands with an empty line and/or consider proper support for `-File -`


mklement0:
This not about a bug, but about possibly unexpected behavior.

A consequence of treating stdin-based input to `-Command -` as if it were entered _interactively_ is the need to terminate _multi-line_ commands with an _empty line_ that parallels the extra ENTER keystroke needed to interactively terminate a multi-line command (_except_ if `PSReadLine` is loaded, which obviates this need, but it's not involved when `-Command -` input is parsed).

Thus, the following command does _not_ work:

```powershell
@'
"first"
"hi" |
  % { "$_ there" }
"last"
'@ | powershell -noprofile -command -
```
The above only yields `first`, because the end of the multi-line command is never detected.

As evidence that this happens in the real world, see [this Stack Overflow question](http://stackoverflow.com/a/42475326/45375).

Inserting an empty line after the multi-line command fixes the problem:

```powershell
@'
"first"
"hi" |
  % { "$_ there" }

"last"
'@ | powershell -noprofile -command -
```

The above now yields both `first` and `last`.

**What is the benefit of treating the `-Command` argument as if its commands were entered _interactively_?**.

**Similarly, it could be argued that it is  `-File -` should be used to provide a _script_ via stdin.**

`File -` isn't currently documented when you run `powershell -?`, but it currently _sort of_ works: it works like `-Command -`, except that it exhibits even more behavior only desirable in an _interactive_ session: it prints the prompt string between commands and the commands themselves before their output, and even adding `-NonInteractive` doesn't help.

Additionally  **`-File -` would then allow enabling passing _arguments_ to the script**, which `-Command -` explicitly disallows (presumably, due to ambiguity over how to pass those arguments).

iSazonov:
At first look I'd expect that "-Command -" accept only one line command but how input multi-line one? And it is not clear that is semantic of "-File script.ps1 -" 

Also I'd expect that we read console input if there is not explicit redirection.

mklement0:
> At first look I'd expect that `-Command -` accept only one line command but how input multi-line one? 

As argued in https://github.com/PowerShell/PowerShell/issues/9494#issuecomment-487621948, there's really no point in supporting `-Command -` altogether, if `-File -` worked properly.

[...]

PR #3963, opened 2017-06-07
shebang without .ps1 extension ends up in recursive loop calling powershell

### LINKED ISSUE #1103: PowerShell Scripts require .ps1 extension
Scripts (on linux) don't function unless they have the extension `.ps1` (generally, languages do not require specific extensions in Linux when using shebang to start them `#! /usr/bin/powershell`
## Steps to reproduce

`myscript`

``` powershell
#!/usr/bin/powershell
echo hi
```

``` bash
chmod a+x ./myscript

# hangs!
./myscript 


mv ./myscript ./myscript.ps1

# works
./myscript.ps1
hi

```
## Expected behavior

On linux, this should work even if the extension isn't `.ps1`
## Actual behavior

**EDIT** 
it turns out it didn't hang, it fails:

``` text
System.TypeInitializationException: The type initializer for 'System.Management.Automation.PSVersionInfo' threw an exception. ---> System.IO.FileLoadException: Could not load file or assembly 'System.Reflection.Extensions, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. An internal error occurred.
 (Exception from HRESULT: 0x8007054F)
   at System.Reflection.Internal.LightUpHelper.GetMethod(Type type, String name, Type[] parameterTypes)
   at System.Reflection.Internal.MemoryMapLightUp.TryLoadMembers()
   at System.Reflection.Internal.MemoryMapLightUp.get_IsAvailable()
   at System.Reflection.Internal.StreamMemoryBlockProvider..ctor(Stream stream, Int64 imageStart, Int32 imageSize, Boolean isFileStream, Boolean leaveOpen)
   at System.Reflection.PortableExecutable.PEReader..ctor(Stream peStream, PEStreamOptions options, Nullable`1 sizeOpt)
   at System.Diagnostics.FileVersionInfo.TryLoadManagedAssemblyMetadata()
   at System.Diagnostics.FileVersionInfo.GetVersionInfo(String fileName)
   at System.Management.Automation.PSVersionInfo.GetBuildVersion()
   at System.Management.Automation.PSVersionInfo..cctor()
   --- End of inner exception stack trace ---
   at System.Management.Automation.PSVersionInfo.get_PSVersion()
   at System.Management.Automation.PSSnapInReader.ReadRegistryInfo(Version& assemblyVersion, String& publicKeyToken, String& culture, String& architecture, String& applicationBase, Version& psVersion)
   at System.Management.Automation.PSSnapInReader.ReadCoreEngineSnapIn()
   at System.Management.Automation.Runspaces.InitialSessionState.ImportCorePSSnapIn()
   at System.Management.Automation.Runspaces.InitialSessionState.CreateDefault2()
   at Microsoft.PowerShell.UnmanagedPSEntry.Start(String consoleFilePath, String[] args, Int32 argc)
   at Microsoft.PowerShell.ManagedPSEntry.Main(String[] args)

```
## Environment data

v0.4.0

### LINKED ISSUE #3959: Launching with shebang does not work (hangs)
Powershell Beta 2 on Ubuntu 16.04

Steps to reproduce
------------------
Create a file hello
#!/opt/microsoft/powershell/6.0.0-beta.2/powershell
gps i*

----------
Then chmod +x it
and run it ./hello

It hangs

Expected behavior
-----------------
Not hanging :-)

Actual behavior
---------------

Environment data
----------------

```powershell
> $PSVersionTable

```

SteveL-MSFT:
Fix https://github.com/PowerShell/PowerShell/issues/1103

On Linux, when a shebang is encountered (first two bytes are #!), it looks for the interpreter to pass the pass to.  In this case, if it is powershell it calls powershell with a path to the script.  If the script doesn't
have a .ps1 extension, powershell treats it like a native command and executes it as such.  The OS
sees the shebang and again calls powershell with the script.  This causes a hang (and will eventually consume all the OS resources as it keeps spawning new powershells).

Fix is to inspect the file to see if it contains the shebang magic number.  If so, we check if the running
powershell is the interpreter.  If so, we treat it the same as a .ps1 script.  Otherwise, we execute it as
a native command and expect the OS to find the correct interpreter (could be different version of powershell, for example).

### DIFF
--- a/src/System.Management.Automation/engine/CommandSearcher.cs
+++ b/src/System.Management.Automation/engine/CommandSearcher.cs
@@ -635,7 +635,41 @@ namespace System.Management.Automation
                     break;
                 }
 
-                if (String.Equals(extension, StringLiterals.PowerShellScriptFileExtension, StringComparison.OrdinalIgnoreCase))
+                // handle case where we are called by the OS to handle a shebang script using us as the interpreter
+                bool isShebangScript = false;
+                try
+                {
+                    using(FileStream fileStream = new FileStream(path, FileMode.Open, FileAccess.Read))
+                    {
+                        byte[] bytes = new byte[2];
+                        int bytesRead = fileStream.Read(bytes, 0, 2);  // read the first two bytes to determine if shebang
+                        if (bytesRead == 2)
+                        {
+                            if (bytes[0] == '#' && bytes[1] == '!')
+                            {
+                                // see if we are supposed to be the interpreter
+                                using(StreamReader file = new StreamReader(fileStream))
+                                {
+                                    string interpreter = file.ReadLine();
+                                    System.Reflection.Assembly assembly = System.Reflection.Assembly.GetEntryAssembly();
+                                    // this returns path to powershell.dll
+                                    string powershellPath = assembly.Location.Replace(".dll","");
+                                    // need to handle both powershell and powershell.exe
+                                    if (interpreter.Split(' ',2)[0].Replace(".exe","") == powershellPath)
+                                    {
+                                        isShebangScript = true;
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+                catch(Exception)
+                {
+                    // If we can't read the file, safe to assume it's not a script
+                }
+
+                if (isShebangScript || String.Equals(extension, StringLiterals.PowerShellScriptFileExtension, StringComparison.OrdinalIgnoreCase) || isShebangScript)
                 {
                     if ((_commandTypes & CommandTypes.ExternalScript) != 0)
                     {
--- a/test/powershell/Host/ConsoleHost.Tests.ps1
+++ b/test/powershell/Host/ConsoleHost.Tests.ps1
@@ -176,6 +176,18 @@ Describe "ConsoleHost unit tests" -tags "Feature" {
             # no extraneous output
             $observed | should be $currentVersion
         }
+
+        $testCases =
+            @{interpreter = "#!${pshome}/powershell"},
+            @{interpreter = "#!${pshome}/powershell.exe"}
+
+        It "Should properly execute shebang script that doesn't have .ps1 extension" -TestCases $testCases {
+            param($Interpreter)
+            $scriptPath = "~/shebangtest"
+            Set-Content -Path $scriptPath -Value $Interpreter
+            Add-Content -Path $scriptPath -Value "'hello'"
+            & $powershell $scriptPath | Should BeExactly 'hello'
+        }
     }
 
     Context "Pipe to/from powershell" {


mklement0:
I suggest taking a step back, in the context of our discussion to align PowerShell's CLI with that of POSIX-like shells (#3743):

POSIX-like shells interpret an operand (in POSIX speak: an argument other than an option / option-argument) as a _script file to execute_ rather than an _arbitrary shell command_.

Translated into PowerShell terms that means:

* A positional argument should bind to `-File`, not `-Command`.

If this change were made, no extra accommodations would have to made for shebang line - _except_ that `-File <file-without-ps1-extension>` doesn't currently work:

    Processing -File '...' failed because the file does not have a '.ps1' extension.

SteveL-MSFT:
@mklement0 that would certainly be a breaking change, although it would align semantics with POSIX (but break many of our tests that expect the positional parameter to be a command).    We'll discuss the implications at the next committee meeting (next week).

The Mac failures aren't related to this change, but I'll spend time getting those fixed anyways on separate PRs.

src/System.Management.Automation/engine/CommandSearcher.cs
```diff
@@ -635,7 +635,41 @@ private CommandInfo GetInfoFromPath(string path)
                     break;
                 }
 
-                if (String.Equals(extension, StringLiterals.PowerShellScriptFileExtension, StringComparison.OrdinalIgnoreCase))
+                // handle case where we are called by the OS to handle a shebang script using us as the interpreter
+                bool isShebangScript = false;
+                try
+                {
+                    using(FileStream fileStream = new FileStream(path, FileMode.Open, FileAccess.Read))
```
iSazonov: There's a lot of code! Can we just read the line (with length < Path.Max) and check Shebang with `RegEx.IsMatch()`? And it seems we're going to have less trouble with BOM.

src/System.Management.Automation/engine/CommandSearcher.cs
```diff
@@ -635,7 +635,41 @@ private CommandInfo GetInfoFromPath(string path)
...
+                        int bytesRead = fileStream.Read(bytes, 0, 2);  // read the first two bytes to determine if shebang
+                        if (bytesRead == 2)
+                        {
+                            if (bytes[0] == '#' && bytes[1] == '!')
+                            {
+                                // see if we are supposed to be the interpreter
+                                using(StreamReader file = new StreamReader(fileStream))
+                                {
+                                    string interpreter = file.ReadLine();
+                                    System.Reflection.Assembly assembly = System.Reflection.Assembly.GetEntryAssembly();
```
iSazonov: Maybe we should have an internal flag `isInterprter`?

iSazonov (commented):
Leave a comment

src/System.Management.Automation/engine/CommandSearcher.cs
SteveL-MSFT: I wanted to limit how much gets read (first two bytes) since every file will get analyzed

src/System.Management.Automation/engine/CommandSearcher.cs
SteveL-MSFT: We should certainly cache this since this if using shebang with PowerShell becomes popular.  I have to explicitly check if the interpreter is exactly the same as the running PowerShell otherwise I let the OS handle it as we would want to support scripts that target specific versions of PowerShell instead of just the system default one.  Actually this reminds me that the current code won't work with

```
#!/usr/bin/env powershell
```

src/System.Management.Automation/engine/CommandSearcher.cs
iSazonov: In either case, the entire file will be read by OS for native processing or PowerShell processing. The question is only in this local buffer. I guess 64K is not a problem.

SteveL-MSFT:
the more I think about this, I agree with @mklement0.  there's really only two approaches:

1. we keep command as the positional parameter and only support scripts with .ps1 extension
2. we make file be the positional parameter, which is a breaking change, but we treat any given file as a script.  this simplifies the shebang as it's back to just being a comment and no need to parse and we let the OS resolve the interpreter

src/System.Management.Automation/engine/CommandSearcher.cs
iSazonov: I open PR #3969 to cache the reflection - we can use `Utils.GetApplicationBaseDefaultPowerShell()`

iSazonov: [Interesting tests](https://www.in-ulm.de/~mascheck/various/shebang/) 
1. interpreter can not be supported on IOS
2. First line effective length is 512 on IOS and 127 on Linux.

iSazonov:
Can we use a shebang like:
```#!/usr/bin/powershell -f```

**Comment by lzybkr** [comment]
I'm not sure this is the right fix.

First, the difference between `-File` and `-Command` is significant and important here in terms of setting the exit code.

Second, I don't think we want shebang PowerShell scripts to run in process, which this change would allow.

I believe the correct fix will involve detecting that we're invoked via shebang, which might mean looking at `arg[0]` in the console host entry point - it probably isn't `powershell`.

**Comment by SteveL-MSFT** [comment]
@iSazonov `#!/usr/bin/powershell -f` is possible, but only if the script has the `.ps1` extension

**Comment by mklement0** [comment]
@lzybkr:

Fundamentally: Is aligning with the CLI of POSIX-like shells a goal in principle?

If so:

* breaking changes are inevitable.

* the awkward emulation of the shebang-line mechanism considered here isn't needed.

* supporting scripts that do _not_ have extension (suffix)  `.ps1` is a must, given that extensions have no intrinsic meaning on Unix platforms and the executability of a file is solely governed by its permission bits relating to its ownership and the invoking user.

POSIX-like shells have no analog to the `-File` option, but they consistently interpret their 1st operand (POSIX speak for: positional argument, as distinct from an option or option-argument) as:

*  the _path to a script file_ to execute
* in a non-interactive shell (session)
* that exits when the script exits
* with the script's exit code getting passed through.

The shebang mechanism - built into the _kernel_ of Unix platforms - relies on that very behavior:

The path to the script file (exactly as specified when the script was directly invoked) is passed as the (last) argument to the interpreter executable specified by full name in the shebang line.
(Platforms differ in how many additional, fixed arguments they support as part of the shebang line; only _one_ is guaranteed to work portably.)

In terms of the command line that is invoked by the kernel, and the `argv[0]` (`$0`, in POSIX shell speak) reported in the resulting process:

* _Unless explicitly changed_, `$0` reflects the path to the script file exactly as invoked.
   * `argv[0]` / `$0` can be set to arbitrary values (see below).

* `ps -o args= $$` tells you the _command line_ as invoked (`$$` is the current PID in POSIX-like shells; I _think_ what that `ps` command reports can't be tampered with, but I'm not sure; in any event, necessary shell quoting around the original arguments will _not_ be reflected).

A simple example: Let's assume script `t`, placed in the current directory and made executable with `chmod +x t`:

```sh
#!/bin/sh

echo "\$0: [$0]"
echo "Originating command line: [$(ps -o args= $$)]"
```

Let's invoke that script directly, letting the kernel interpret the shebang line:

```none
$ ./t
$0: [./t]
Originating command line: [/bin/sh ./t]
```

Now let's see how we can change `argv[0]` / `$0` (and also demonstrate how `ps` doesn't preserve shell quoting):

```none
$ sh -c '. ./t' 'foo bar'
$0: [foo bar]
Originating command line: [sh -c . ./t foo bar]
```

Note when a POSIX-like shell is explicitly invoked with a script filename operand - e.g., `sh ./t` - it:

 * does _not_ care whether the script file is executable - all that's needed is the ability to _read_ the file.
 * _ignores_ the shebang line (which may or may not specify the shell the script is being passed to), because it is a _comment_ from the shell's perspective.

---

>  I don't think we want shebang PowerShell scripts to run in process

It would run in-process in  the `powershell` process, and that's exactly what's expected (and what happens when you run `sh ./t`, for instance).
In-process considerations apply only when you have a parent session from which you launch _other_ commands, but here there's only 1 session here, whose sole purpose is to run the script, and - at least by default - immediately exit afterwards.

lzybkr:
@mklement0 - if you run `./t` and `sh ./t`, you are starting a new process in both cases.

With the proposed change, running `./t` from bash would start a new PowerShell process, but it would not when running the same command line from PowerShell.

This should be a intentional design choice, not an accidental side effect of fixing the recursive hang. My inclination is that starting a new process is desirable (so perhaps more POSIX-like), but I'm open to a discussion, especially given PowerShell's slow startup cost.

As for `-File` - I mentioned that only in regards to semantics - we want shebang to behave as though `-File` was specified, in part so the exit code is set correctly.

And as for `$0`, I was hinting at what I thought it would take to implement what I propose above - that we have a new code path that behaves as though you ran `powershell -File foo` when `argv[0] == 'foo'`.

mklement0:
> but it would not when running the same command line from PowerShell.

Again, if we're talking about aligning with POSIX-like shells:  
It certainly _should_, because that invocation should be treated as the invocation of an _external utility_ (delegated to a system call such as `execv` on Unix). We're not talking about _sourcing _ the invocation (the `.` is part of the script path).

I get that this is not how it currently works with invoking `*.ps1` files, which do run in the same process, but perhaps the extension / lack thereof could be the distinguishing factor:

* Continue to run `*.ps1` files in-process from within a PowerShell session, for backward compatibility.

* Treat files _without_ that extension like any other external utility - invoke it via the platform's system calls (you don't even need to know / care what they are and whether their shebang line _happens_ to also target PowerShell).

   * If the intent is to run such files in-process, dot-sourcing is an option (which _already_ works with files that do _not_ have a `.ps1` extension.

What interpreter ultimately processes a given executable file that happens to have a shebang line should really be considered an implementation detail, and there's _no_ reason to reflect that interpreter in the filename extension (even though, sadly, there's a widespread bad habit in the Unix world of naming shell scripts `*.sh`).

iSazonov:
What is the behavior for another interpreters (bash, perl, python, etc.)? Run a native script in-process?

SteveL-MSFT:
[perl](https://linux.die.net/man/1/perl) follows the POSIX convention of having the positional parameter point to a script and using an explicit `-c` for a command
[bash](https://linux.die.net/man/1/bash), of course, does the same
[python](https://linux.die.net/man/1/python) accepts the first positional as the script and subsequent positional parameters passed as args to the script

We should strongly consider adopting the POSIX conventions for PowerShell (which means closing this PR):

1. we've already seen that existing Linux tools that work with shells have expectations on parameter semantics and syntax and if we don't adopt them, it means there are some tools people use that aren't compatible with PowerShell
2. if we stay different, this will impact our adoption on Linux
3. since this is a breaking change, if we ever do this, we need to do this with the 6.0.0 release

iSazonov:
My question was how they behave internally: if we type in Bash "bash-script-file-name" - do Bash run it "in-process"?

mklement0:
@SteveL-MSFT: An (interfaith) amen to that.

> accepts the first positional as the script and subsequent positional parameters passed as args to the script

Interpreting all remaining arguments after the script filename as the arguments to pass through to the _script_ is also part of "shebang-line compatibility", so all POSIX-like shells and Perl support it too; Ruby are Node.js are other examples.

The ability to use `-c` (per POSIX) for passing a command _string_  (the _contents_ of a script) is less universally supported: among the interpreters mentioned, Perl, Ruby, and Node.js require `-e`, unfortunately.

Combining `-c`  with positional arguments _is_ supported and [_required by POSIX_](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html), but involves a _pitfall_:

The first argument after the command string binds to `$0` (the equivalent of `$myInvocation.MyCommand.Name`), not `$1`, which means that it isn't reflected in the `$@` (`$*`) array of positional parameters.


```sh
$ sh -c 'printf "%s " "$@"' ARG0 one two
one two    # !! Note how ARG0 is not listed.
```

PowerShell does _not_ support this, currently:

```powershell
# !! BROKEN
> powershell -noprofile -command '$args' one two
Unexpected token 'one' in expression or statement.

# !! BROKEN
> powershell -noprofile -command '"`$args: $args"' one two
$args : The term '$args' is not recognized as the name of a cmdlet, function, script file, or operable program
```

It looks like PowerShell simply _string-appends_ the additional arguments _directly to the command string_, which, of course, only works in limited cases, whereas `-c`  in POSIX-like shells basically allows passing an _entire script_ ad-hoc, with arguments passed as positional parameters, as to a script.

The only way to get this to work currently appears to be:

```powershell
$ powershell -noprofile -command '& { "`$args: $args" }' one two
$args: one two
```

_From within Powershell_:

* This actually BREAKS:

```powershell
# !! BROKEN when invoked from PowerShell
> powershell -noprofile -command '& { "`$args: $args" }' one two
$args: : The term '$args:' is not recognized as the name of a cmdlet, function, script file, or operable program.
```

* Using an actual script block (not a string representation) breaks too, because arguments are apparently not supported:

```powershell
# !! DOES NOT WORK - arguments not supported.
> powershell -noprofile -command { "`$args: $args" } one two
(prints CLI help text)

# Only works *without arguments*:
> powershell -noprofile -command { "`$args: [$args]" }
$args: []
```

Other problematic aspects of `-Command` are discussed in #3223.

mklement0:
@iSazonov:

> if we type in Bash "bash-script-file-name" - do Bash run it "in-process"?

No: The only way to make a POSIX-like shell run a script run in-process of an existing session is to use `.` (or its nonstandard, effective alias, `source`).

Conversely, anything that is neither a shell builtin or a shell-language statement is considered an _external utility_ and _runs in a child process_, and it is _irrelevant_ (a) whether the script invoked is indeed a script or a binary and (b) if the former, what shell/interpreter will process it.

Note, however, that this only works with _executable_ shell scripts (whereas invocation with `.` doesn't require this, but will, of course, only work if the script "speaks the same language" as the invoking shell).

An otherwise well-formed script with shebang line that isn't executable [by the current user] will simply result in a `Permission denied` error.

If it is executable, a _system_ call is made to invoke the external utility, at which point the kernel's shebang-line processing will kick in.

As a _courtesy_, if a script is executable but lacks a shebang line, POSIX-like shells will fall back to processing the script _themselves_, but - again - in a separate instance in a _child process_.

mklement0:
To more explicitly contrast the invoke-a-script-written-in-the-shell's-own-language scenarios:

* In POSIX-like shells, invoking a script is an all-or-nothing choice:

  * Invoked with `.` (`source`), _everything_ the script does potentially affects the current shell; (e.g., a non-local variable created in the script will linger).

  * Invoked by file path only, it will run in a _child process_ that _cannot affect the calling shell's environment at all_.

* In PowerShell, the distinctions are more nuanced:

  * Invoked with `.` (`source`), the same applies as in POSIX-like shells: , _everything_ the script does potentially affects the current session and its global state.

  * Invoked by file path only, the script still runs in-process, but in a _child scope_:

    * While that commendably localizes variables and function definitions,

    * it still allows for explicit and - pitfall alert - implicit modification of the session's global state - _something that will catch Unix shell users by surprise_; e.g.:
        * Simply changing the current location in the script (using `Set-Location` the equivalent of `cd`) changes the location _globally_.
        * By contrast, changing a global variable, for instance, requires a more deliberate effort.
        
_Personally, I think the implicit and invariable global location-changing behavior is problematic._

powercode:
Another area is static members on classes.
They will also be shared implicitly.  

Maybe we should have syntactic sugar for invoking scripts easily out-of-proc.
like 
```powershell
.\a.ps1
. .\a.ps1
! .\a.ps1 # -> powershell.exe -noprofile -file .\a.ps1
```

powercode:
Or change the defaults before PowerShell v6 ships, so that you explicitly have to opt in to run in-proc.

mklement0:
> Another area is static members on classes.

Good point, and definitely worth documenting, though probably less of a problem in practice.

> Maybe we should have syntactic sugar

Introducing another symbol-based operator may not be worth the effort (as an aside: I assume `!` was just an example, but to clear: it's already taken), especially that, once the CLI has been harmonized with POSIX, all you'd need to do to run out-of-process is:

    powershell ./foo.ps1

 > so that you explicitly have to opt in to run in-proc.

As long as it doesn't come with surprises, I think that running in-process by default is a great _strength_: 

  * No startup cost (no child-process creation, no PowerShell startup cost)
  * Integration with the calling session with respect to input/output _types_ (no need for serializing / deserializing of objects to to/from _strings_ on in-/output).

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and agree that the right change is to have -File be the positional parameter instead of -Command.  Also allow scripts used with -File (or implicit) to not require a .ps1 extension.  A new PR will be submitted.

Issue #3912, opened 2017-06-01
WebRequest tests should be rewritten to use a new runspace rather than a job 


JamesWTruher:
Because the test httplistener is run as a job, the job may be still starting when the tests are run which may cause the tests to fail on busy systems. we should avoid using jobs for performance reasons and ensure that the listener is responding before we return to run the tests.

SteveL-MSFT:
This appears to be impacting test runs on CentOS with current tests

JamesWTruher:
merged and closed

PR #3966, opened 2017-06-07
Alter test httplistener to use a new runspace rather than a job

JamesWTruher:
fix for #3912 

This improves the performance of the listener by not relying on a new
process starting up to run the listener. It improves the debugability
of tests by providing more direct access to the session executing
the listener. It also reverses the blocking nature of starting the
listener. By default, Start-HttpListener will no longer block, you need
to use -Foreground to have Start-HttpListener block. Lastly, create
a way to catch errors if the listener has a problem. If code in the
listener throws, it emits an error record.

fix the tests which use the listener

### DIFF
--- a/test/tools/Modules/HttpListener/HttpListener.psm1
+++ b/test/tools/Modules/HttpListener/HttpListener.psm1
@@ -37,7 +37,7 @@ Function Start-HTTPListener {
         [Int] $Port = 8080,
 
         [Parameter()]
-        [switch] $AsJob
+        [switch] $Foreground
         )
 
     Process {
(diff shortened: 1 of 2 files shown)


test/tools/Modules/HttpListener/HttpListener.psm1
```diff
@@ -214,19 +214,32 @@ Function Start-HTTPListener {
...
         else
         {
-            & $script -Port $Port
+            $ps = [powershell]::Create()
+            $null = $ps.AddScript($script)
+            $null = $ps.AddParameter("port",$port)
+            $AsyncResponse = $ps.BeginInvoke()
+            [pscustomobject]@{
+                PowerShell = $ps
+                AsyncResponse  = $AsyncResponse
```
SteveL-MSFT: AsyncResponse doesn't appear to actually get used, why not just return $ps?

JamesWTruher: the async response can be used for debugging purposes. It's better to keep track of it.

SteveL-MSFT: Perhaps add a comment then

SteveL-MSFT (approved):

test/tools/Modules/HttpListener/HttpListener.psm1
JamesWTruher: comment added

PR #3968, opened 2017-06-08
Fix non-fatal pester test failures

SteveL-MSFT:
On another PR, I see a bunch of red error messages, but they appear to not actually fail the tests, but makes it hard to determine what actually failed.  Cleaned up the tests.  It appears that with the refactoring of the test modules, some lines of code were missed.  There were also cut and paste issues in the test code so cleanup doesn't happen correctly.


iSazonov (approved):
LGTM.

SteveL-MSFT:
@mirichmo can you merge?

lzybkr (approved):

PR #3969, opened 2017-06-08
Add cache for GetApplicationBase

iSazonov:
Motivation
---
We call Utils.GetApplicationBase() multiple times at startup and later so we need cache the one to exclude a reflection.

Fix
---
1. Add a static getter-only property named `DefaultPowerShellAppBase` in `Utils.cs`.
2. Refactor code to directly use cached value.

Additional considerations
---
Code that uses a ShellId from Context has not been changed, although it is only associated with the Windows PowerShell (FullCLR).

### DIFF
--- a/src/System.Management.Automation/engine/Utils.cs
+++ b/src/System.Management.Automation/engine/Utils.cs
@@ -238,20 +238,35 @@ namespace System.Management.Automation
         }
 #endif
 
+        private static string s_GetApplicationBaseDefaultPowerShell = InitGetApplicationBaseDefaultPowerShell();
+
+        private static string InitGetApplicationBaseDefaultPowerShell()
+        {
+            return GetApplicationBase(Utils.DefaultPowerShellShellID);
+        }
+
         /// <summary>
-        /// Gets the application base for current monad version
+        /// Gets the application base for current PowerShell version.
         /// </summary>
         /// <returns>
-        /// applicationbase path for current monad version installation
+        /// Applicationbase path for current PowerShell version installation.
         /// </returns>
         /// <exception cref="SecurityException">
-        /// if caller doesn't have permission to read the key
+        /// if caller doesn't have permission to read the key.
         /// </exception>
+        internal static string GetApplicationBaseDefaultPowerShell()
+        {
+            return s_GetApplicationBaseDefaultPowerShell;
+        }
+
         internal static string GetApplicationBase(string shellId)
         {
 #if CORECLR
-            // Use the location of SMA.dll as the application base
-            // Assembly.GetEntryAssembly and GAC are not in CoreCLR.
+            if (s_GetApplicationBaseDefaultPowerShell != null)
+            {
+                return s_GetApplicationBaseDefaultPowerShell;
+            }
+            // Use the location of SMA.dll as the application base.
             Assembly assembly = typeof(PSObject).GetTypeInfo().Assembly;
             return Path.GetDirectoryName(assembly.Location);
 #else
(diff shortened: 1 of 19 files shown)


src/System.Management.Automation/engine/Utils.cs
```diff
@@ -238,20 +238,35 @@ internal static string GetApplicationBaseFromRegistry(string shellId)
         }
 #endif
 
+        private static string s_GetApplicationBaseDefaultPowerShell = InitGetApplicationBaseDefaultPowerShell();
+
+        private static string InitGetApplicationBaseDefaultPowerShell()
+        {
+            return GetApplicationBase(Utils.DefaultPowerShellShellID);
+        }
+
```
daxian-dbw: How about having an internal static property named `DefaultPowerShellAppBase` like this:
```CSharp
internal static string DefaultPowerShellAppBase { get; } = GetApplicationBase(DefaultPowerShellShellID);
```
And then use this get-only property in places where `GetApplicationBase(DefaultPowerShellShellID)` is used.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -238,20 +238,35 @@ internal static string GetApplicationBaseFromRegistry(string shellId)
...
+        /// Applicationbase path for current PowerShell version installation.
         /// </returns>
         /// <exception cref="SecurityException">
-        /// if caller doesn't have permission to read the key
+        /// if caller doesn't have permission to read the key.
         /// </exception>
+        internal static string GetApplicationBaseDefaultPowerShell()
+        {
+            return s_GetApplicationBaseDefaultPowerShell;
+        }
```
daxian-dbw: With that getter-only property, you won't need this method.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -238,20 +238,35 @@ internal static string GetApplicationBaseFromRegistry(string shellId)
...
         internal static string GetApplicationBase(string shellId)
         {
 #if CORECLR
-            // Use the location of SMA.dll as the application base
-            // Assembly.GetEntryAssembly and GAC are not in CoreCLR.
+            if (s_GetApplicationBaseDefaultPowerShell != null)
+            {
+                return s_GetApplicationBaseDefaultPowerShell;
+            }
+            // Use the location of SMA.dll as the application base.
```
daxian-dbw: With that getter-only property, you won't need to add this.

daxian-dbw:
@iSazonov Sorry about the late review 😦

src/System.Management.Automation/engine/Utils.cs
iSazonov: Good catch!

Done.

src/System.Management.Automation/engine/Utils.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/Utils.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -238,20 +238,16 @@ internal static string GetApplicationBaseFromRegistry(string shellId)
...
+
         internal static string GetApplicationBase(string shellId)
         {
 #if CORECLR
-            // Use the location of SMA.dll as the application base
-            // Assembly.GetEntryAssembly and GAC are not in CoreCLR.
+            if (DefaultPowerShellAppBase != null)
+            {
+                return DefaultPowerShellAppBase;
+            }
```
daxian-dbw: This is a bit confusing -- looks like a recursion would happen even though it's not. I suggest removing this if block.

iSazonov: Fixed.

daxian-dbw (approved):

iSazonov:
@daxian-dbw  Thanks for the code review!

Issue #3967, opened 2017-06-07
Wildcard matching should be used in help file search


lzybkr:
In https://github.com/PowerShell/PowerShell/pull/3852 - the change treats the parameter like a wildcard, but uses a regex. Also, the fix repeatedly checks (inside the loop) if it's a pattern, that can be done once.

I would have expected a fix more like:

```C#
            var wildcardPattern = WildcardPattern.ContainsWildcardCharacters(pattern)
                ? WildcardPatter.Get(pattern, WildcardOptions.IgnoreCase)
                : null;

            foreach (string filePath in files)
            {
                if (filePath.IndexOf(pattern, StringComparison.OrdinalIgnoreCase) >= 0)
                {
                    result.Add(filePath);
                    break;
                }

                if (wildcardPattern.?IsMatch(pattern))
                {             
                    result.Add(filePath);
                }
            }
```

PR #3971, opened 2017-06-08
Fix the issue Wildcard matching should be used in help file search

chunqingchen:
Fix issue #3967

Summary of the issue:

According to Jason's suggestion, replace the regex expression with wildcard.

Root cause of the issue:

This is an code improvement. The existed test cases are covering the scenario

### DIFF
--- a/src/System.Management.Automation/help/MUIFileSearcher.cs
+++ b/src/System.Management.Automation/help/MUIFileSearcher.cs
@@ -122,9 +122,9 @@ namespace System.Management.Automation
             ArrayList result = new ArrayList();
             string[] files = Directory.GetFiles(path);
 
-            string regexPattern = pattern.Replace(".", @"\.");
-            regexPattern = regexPattern.Replace("*", ".*");
-            regexPattern = regexPattern.Replace("?", ".?");
+            var wildcardPattern = WildcardPattern.ContainsWildcardCharacters(pattern)
+                ? WildcardPattern.Get(pattern, WildcardOptions.IgnoreCase)
+                : null;
 
             foreach (string filePath in files)
             {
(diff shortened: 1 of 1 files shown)


src/System.Management.Automation/help/MUIFileSearcher.cs
```diff
@@ -133,13 +133,11 @@ private string[] GetFiles(string path, string pattern)
                     result.Add(filePath);
                     break;
                 }
-                // If the input is pattern instead of string, we need to use Regex expression.
-                if (pattern.Contains("*") || pattern.Contains("?"))
+
+                string fileName = Path.GetFileName(filePath);
```
lzybkr: `fileName` isn't used when `wildcardPattern` is null, so we should avoid the call (which copies and allocates memory) unless we'll actually use the value.

chunqingchen:
@lzybkr Thanks Jason, your comment has been resolved

lzybkr (approved):

src/System.Management.Automation/help/MUIFileSearcher.cs
gonhidi: Actually, shouldn't `fileName` also be used when there are no wildcards? Currently
```PowerShell
if (filePath.IndexOf(pattern, StringComparison.OrdinalIgnoreCase) >= 0)
```
is searching the whole path for a match. If it is desirable to only target the file name when performing the wildcard match, shouldn't it also be appropriate here? (Actually, given that an exact match is expected — that's how `GetFiles` would work in the non-UNIX case for a non-wildcard pattern, if I am not mistaken — shouldn't the test be against `fileName.Equals(pattern, StringComparison.OrdinalIgnoreCase)` instead of using `IndexOf`?)

(An unrelated minor issue I believe exists with the logic of `GetFiles` is using `break` instead of `continue` when a non-wildcard match is found. This might limit the matches in the (rare) case in which a file name equals a wildcard search pattern, e.g. foo?, though I haven't tested how `GetFiles`would behave in that scenario, and it might be nitpicking if the expected directory and file names are curated.

Also, sorry for butting in, but since I wasn't sure enough about my observations I didn't want to open an issue, so I didn't know a better way to proceed.)

lzybkr: @gonhidi - thanks for the extra scrutiny - it is definitely appreciated.

It does seem like there should be consistency between wildcards and non-wildcards. Feel free to open an issue.

PR #3972, opened 2017-06-08
adding type inference for Foreach-Object -MemberName  

### LINKED ISSUE #2596: Make type inference understand Foreach-Object -MemberName <propertyname>
```powershell
#this works
Get-Process | % {$_.MainModule} | % Com<Tab>
#this doesn't
Get-Process | % MainModule | % Com<Tab>

```
It would be nice if the type inference could interpret -Membername as a MemberInvokeExpression on $_.

Steps to reproduce
------------------
```powershell
Get-Process | % MainModule | % Com<Tab>
```
Expected behavior
-----------------
Completion to 
```powershell
Get-Process | % MainModule | % Company
```
Actual behavior
---------------
No completion

Environment data
----------------

```powershell

```

powercode:
Fixes #2596

This PR adds type inference support for the 'MemberName' parameterset for Foreach-Object.

Before this PR:
```
#this works
Get-Process | % {$_.MainModule} | % Com<Tab>
#this doesn't
Get-Process | % MainModule | % Com<Tab>
```
After this PR:
```
#this works (completes Company)
Get-Process | % {$_.MainModule} | % Com<Tab>
Get-Process | % MainModule | % Com<Tab> 
```

### DIFF
--- a/src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
+++ b/src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
@@ -1028,7 +1028,6 @@ namespace System.Management.Automation
             {
                 if (type.Type == typeof(PSObject))
                 {
-                    
                     continue;
                 }
                 var members = _context.GetMembersByInferredType(type, isStatic, filter: null);
--- a/src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
+++ b/src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
@@ -122,6 +122,9 @@ namespace System.Management.Automation
             Helper = new PowerShellExecutionHelper(powerShell);
         }
 
+        // used to infer types in script properties attached to an object,
+        // to be able to determine the type of $this in the scripts properties
+        public PSTypeName CurrentThisType { get; set; }
 
         public TypeDefinitionAst CurrentTypeDefinitionAst { get; set; }
 
@@ -868,7 +871,7 @@ namespace System.Management.Automation
             // foreach-object - yields the type of it's script block parameters
             if (cmdletInfo.ImplementingType == typeof(ForEachObjectCommand))
             {
-                foreach (var foreachType in InferTypesFromForeachCommand(pseudoBinding))
+                foreach (var foreachType in InferTypesFromForeachCommand(pseudoBinding, commandAst))
                 {
                     yield return foreachType;
                 }
@@ -895,33 +898,66 @@ namespace System.Management.Automation
             yield return new PSTypeName(typeof(CimInstance));
         }
 
-        private IEnumerable<PSTypeName> InferTypesFromForeachCommand(PseudoBindingInfo pseudoBinding)
+        private IEnumerable<PSTypeName> InferTypesFromForeachCommand(PseudoBindingInfo pseudoBinding, CommandAst commandAst)
         {
-            AstParameterArgumentPair argument;
-            if (pseudoBinding.BoundArguments.TryGetValue("Begin", out argument))
+            var previousPipelineElement = GetPreviousPipelineCommand(commandAst);
+            if (previousPipelineElement == null)
             {
-                foreach (var type in GetInferredTypeFromScriptBlockParameter(argument))
+                yield break;
+            }
+            foreach (var t in InferTypes(previousPipelineElement))
+            {
+                var thisToRestore = _context.CurrentThisType;
+                try
                 {
-                    yield return type;
+                    _context.CurrentThisType = t;
+                    AstParameterArgumentPair argument;
+                    if (pseudoBinding.BoundArguments.TryGetValue("MemberName", out argument))
+                    {
+                        var memberName = (((AstPair) argument).Argument as StringConstantExpressionAst)?.Value;
+                        if (memberName != null)
+                        {
+                            var types = _context.GetMembersByInferredType(t, false, null);
+                            bool maybeWantDefaultCtor = false;
+                            var res = new List<PSTypeName>(10);
+                            foreach (var type in GetTypesOfMembers(memberName, types, ref maybeWantDefaultCtor, isInvokeMemberExpressionAst: false))
+                            {
+                                yield return type;
+                            }
+                        }
+                    }
+
+                    if (pseudoBinding.BoundArguments.TryGetValue("Begin", out argument))
+                    {
+                        foreach (var type in GetInferredTypeFromScriptBlockParameter(argument))
+                        {
+                            yield return type;
+                        }
+                    }
+
+                    if (pseudoBinding.BoundArguments.TryGetValue("Process", out argument))
+                    {
+                        foreach (var type in GetInferredTypeFromScriptBlockParameter(argument))
+                        {
+                            yield return type;
+                        }
+                    }
+
+                    if (pseudoBinding.BoundArguments.TryGetValue("End", out argument))
+                    {
+                        foreach (var type in GetInferredTypeFromScriptBlockParameter(argument))
+                        {
+                            yield return type;
+                        }
+                    }
+                }
+                finally
+                {
+                    _context.CurrentThisType = thisToRestore;
                 }
             }
 
-            if (pseudoBinding.BoundArguments.TryGetValue("Process", out argument))
-            {
-                foreach (var type in GetInferredTypeFromScriptBlockParameter(argument))
-                {
-                    yield return type;
-                }
             }
-
-            if (pseudoBinding.BoundArguments.TryGetValue("End", out argument))
-            {
-                foreach (var type in GetInferredTypeFromScriptBlockParameter(argument))
-                {
-                    yield return type;
-                }
-            }
-        }
 
         private IEnumerable<PSTypeName> InferTypesFromWhereAndSortCommand(CommandAst commandAst)
         {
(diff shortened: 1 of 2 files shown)


iSazonov:
@powercode Please add a PR description that we can use later for documentation. (What do you want to see in an absolutely great documentation? 😄 )

powercode:
Better?

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -895,33 +898,66 @@ private static IEnumerable<PSTypeName> InferTypesFromCimCommand(PseudoBindingInf
...
+            foreach (var t in InferTypes(previousPipelineElement))
             {
-                foreach (var type in GetInferredTypeFromScriptBlockParameter(argument))
+                var thisToRestore = _context.CurrentThisType;
+                try
                 {
-                    yield return type;
-                }
-            }
+                    _context.CurrentThisType = t;
```
lzybkr: I think this should be moved to where we recurse on the script method/property - you should know the type there as well, and it limits the scope more clearly.

You should also include some tests that depend on this value being set correctly.

Last - because it's effectively a global, you probably need a test that covers potential recursion where `$this` is used in two different script properties on completely different types and completion depends on correctly inferring the type both times.

lzybkr (changes requested):
Thanks for this fix, I'm sure I've noticed it not working before.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -895,33 +898,66 @@ private static IEnumerable<PSTypeName> InferTypesFromCimCommand(PseudoBindingInf
             yield return new PSTypeName(typeof(CimInstance));
         }
 
-        private IEnumerable<PSTypeName> InferTypesFromForeachCommand(PseudoBindingInfo pseudoBinding)
+        private IEnumerable<PSTypeName> InferTypesFromForeachCommand(PseudoBindingInfo pseudoBinding, CommandAst commandAst)
         {
-            AstParameterArgumentPair argument;
-            if (pseudoBinding.BoundArguments.TryGetValue("Begin", out argument))
+            var previousPipelineElement = GetPreviousPipelineCommand(commandAst);
+            if (previousPipelineElement == null)
```
powercode: Correct. Fixing.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -895,33 +898,66 @@ private static IEnumerable<PSTypeName> InferTypesFromCimCommand(PseudoBindingInf
...
             {
-                foreach (var type in GetInferredTypeFromScriptBlockParameter(argument))
-                {
-                    yield return type;
-                }
+                yield break;
             }
-
-            if (pseudoBinding.BoundArguments.TryGetValue("Process", out argument))
+            foreach (var t in InferTypes(previousPipelineElement))
```
powercode: Fixing

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
powercode: Agree about the test, but it will require some refactoring to pass the type down to the scriptblock handling. Worth it?

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -968,16 +1004,17 @@ private IEnumerable<PSTypeName> InferTypesFrom(MemberExpressionAst memberExpress
...
             var memberAsStringConst = memberCommandElement as StringConstantExpressionAst;
             if (memberAsStringConst == null)
-                yield break;
+                return new PSTypeName[0];
 
             var exprType = GetExpressionType(expression, isStatic);
             if (exprType == null || exprType.Length == 0)
             {
-                yield break;
+                return new PSTypeName[0];
```
powercode: Thx. Didn't know about it. Could have assumed it existed though :)

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -987,154 +1024,179 @@ private IEnumerable<PSTypeName> InferTypesFrom(MemberExpressionAst memberExpress
...
-                            }
-                            case FieldInfo fieldInfo:       // .net field
-                            {
-                                if (fieldInfo.Name.EqualsOrdinalIgnoreCase(memberName) && !isInvokeMemberAst)
-                                {
-                                    yield return new PSTypeName(fieldInfo.FieldType);
-                                }
-                                continue;
-                            }
+                        goto NextMemberNameToCheck;
```
powercode: Remnant of refactorings. Should definitely be 'break'.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
lzybkr: In my experience, it's probably worth it.

Personally, I would write a test covering the recursion I mentioned. If the test fails, I'm done, I know the refactoring was necessary.

If the test passes, I step through the code to convince myself my test really is testing what I thought it was so I can make an informed decision on whether it's worth refactoring or not.

powercode: Good point

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -987,154 +1024,179 @@ private IEnumerable<PSTypeName> InferTypesFrom(MemberExpressionAst memberExpress
...
+        }
 
-                            case DotNetAdapter.MethodCacheEntry methodCacheEntry: // .net method
+        private bool TryGetTypeFromMember(object member, string memberName, ref bool maybeWantDefaultCtor, bool isInvokeMemberExpressionAst, List<PSTypeName> result, List<string> memberNamesToCheck)
+        {
+            switch (member)
+            {
+                case PropertyInfo propertyInfo: // .net property
+                {
+                    if (propertyInfo.Name.EqualsOrdinalIgnoreCase(memberName) && !isInvokeMemberExpressionAst)
```
powercode: Why?

Should I change it back? 
Personally, I think it improves readability.

lzybkr: When casually reading code, I care more about the strings being compared than the comparer, I might skip over the comparer completely. Written this way, I stumble over the long name before getting to what I was looking for.

I suppose I could get used to this name, but only if it was used more widely.

---
In reply to: [122571659](https://github.com/PowerShell/PowerShell/pull/3972#discussion_r122571659) [](ancestors = 122571659)

test/powershell/engine/Api/TypeInference.Tests.ps1
```diff
@@ -302,6 +301,70 @@ Describe "Type inference Tests" -tags "CI" {
...
+                $this.Value = "TheValue"
+            }
+        }
+        class YSP {
+            [XSP] $X
+            YSP() {$this.X = [XSP]::new()}
+        }
+        Update-TypeData -TypeName XSP -MemberName TheValue -MemberType ScriptProperty -Value { return $this.Value } -Force
+        Update-TypeData -TypeName YSP -MemberName XVal -MemberType ScriptProperty -Value {return $this.X } -Force
+
```
lzybkr: The type data should be removed so it doesn't impact some future test. That, or the type name should be much more random.

test/powershell/engine/Api/TypeInference.Tests.ps1
```diff
@@ -302,6 +301,70 @@ Describe "Type inference Tests" -tags "CI" {
...
+        $ast = {[YSP]::new() | Foreach-Object -MemberName XVal | ForEach-Object -MemberName TheValue}.Ast
+        $typeNames = [AstTypeInference]::InferTypeof($ast, [TypeInferenceRuntimePermissions]::AllowSafeEval)
+        $typeNames.Count | Should be 1
+        $typeNames[0] | Should be 'System.String'
+    }
+
+    # Not yet implementet
+    It 'Infers typeof Foreach-Object -Member when using dependent scriptproperties on pscustomobject' -Skip {
```
lzybkr: Spelling, and to be honest, I don't think we want tests for features that might never get implemented.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -968,173 +1007,213 @@ private IEnumerable<PSTypeName> InferTypesFrom(MemberExpressionAst memberExpress
...
 
             // We use a list of member names because we might discover aliases properties
             // and if we do, we'll add to the list.
             var memberNameList = new List<string> { memberAsStringConst.Value };
             foreach (var type in exprType)
             {
+                if (type.Type == typeof(PSObject))
+                {
+                    
+                    continue;
```
lzybkr: Extra blank line.

lzybkr (changes requested):
OK, I think this is almost ready, just a couple of small things to address first.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
powercode: fixed.

test/powershell/engine/Api/TypeInference.Tests.ps1
powercode: removed

test/powershell/engine/Api/TypeInference.Tests.ps1
powercode: fixed

lzybkr (approved):

iSazonov:
@powercode Great work! Thanks!

Issue #2561, opened 2016-10-28
PowerShell doesn't kill native child processes when StopProcessing is called on pipeline


SteveL-MSFT:
The native command processor doesn't correctly kill child processes (only the parent) when Stop()/BeginStop() is called:

build createchildprocess from test\tools\CreateChildProcess

``` powershell
$cmd = "pathto\createchildprocess.exe"
$ps = [powershell]::create().Addcommand($cmd)
$ps.AddParameter(2) # this creates two children
$ps.BeginInvoke()
# verify that 3 instances of createchildprocess has started
$ps.Stop()
```

Expected:
parent and child process to be killed

Actual:
parent is killed and children are orphaned (they will eventually timeout and exit)

Repros on PS5.1, PSCore 6.0 Windows/Linux

``` powershell
Name                           Value
----                           -----
PSVersion                      5.1.14931.1000
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.14931.1000
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

SteveL-MSFT:
@joeyaiello see if this impacts our partners

iSazonov:
Related [test ](https://github.com/SteveL-MSFT/PowerShell/blob/52e3916fab229378e118ce8fb32d8d01364309d6/test/powershell/Language/Scripting/NativeExecution/NativeCommandProcessor.Tests.ps1#L42) "Should kill native process tree" should be enabled after the fix.
Also we could replace "while (((Get-Date) - $startTime).TotalSeconds -lt 5)" with Wait-Process in the test.

PR #3982, opened 2017-06-09
Consolidated the two test exes into one

SteveL-MSFT:
In preparation for another PR where I need another test exe, it made sense to have just a single one rather than multiple.  Combined the previous echoargs and createchildprocess test exes into a single TestExe that can be extended for other tests.


test/powershell/Language/Scripting/NativeExecution/NativeCommandArguments.Tests.ps1
```diff
@@ -1,41 +1,41 @@
...
-        $lines[1] | Should Be 'Arg 1 is <a"b c"d>'
-    }
-}
+Describe "Native Command Arguments" -tags "CI" {
+    # Find where test/powershell is so we can find the testexe command relative to it
+    $powershellTestDir = $PSScriptRoot
+    while ($powershellTestDir -notmatch 'test[\\/]powershell$') {
+        $powershellTestDir = Split-Path $powershellTestDir
+    }
+    $testexe = Join-Path (Split-Path $powershellTestDir) tools/TestExe/bin/testexe
```
iSazonov: I see the "find testsexe" code is used multiple times in tests. Maybe move it in HelpersCommon.psm1 as a function?
And we should use BeforeAll here.

build.psm1
```diff
@@ -681,13 +681,13 @@ function Publish-PSTestTools {
 
     Find-Dotnet
 
-    $tools = @("$PSScriptRoot/test/tools/EchoArgs", "echoargs"), @("$PSScriptRoot/test/tools/CreateChildProcess", "createchildprocess")
+    $tools = ,@("${PSScriptRoot}/test/tools/TestExe", "testexe")
```
iSazonov: TestExe vs testexe - We need to be consistent and cautious with the case of letters.
TestExe may be a bad name on Linux. Maybe `BinaryTestRunner` or something like?

test/powershell/Language/Scripting/NativeExecution/NativeCommandProcessor.Tests.ps1
```diff
@@ -42,20 +42,20 @@ Describe "Native Command Processor" -tags "Feature" {
     It "Should kill native process tree" -Pending {
 
         # make sure no test processes are running
-        # on Linux, the Process class truncates the name so filter using Where-Object
-        Get-Process | Where-Object {$_.Name -like 'createchildproc*'} | Stop-Process
+        Get-Process testexe -ErrorAction SilentlyContinue | Stop-Process
```
iSazonov: Docs say that by default, Stop-Process prompts for confirmation - we should add "-Force".

test/powershell/Language/Scripting/NativeExecution/NativeCommandProcessor.Tests.ps1
```diff
@@ -72,7 +72,7 @@ Describe "Native Command Processor" -tags "Feature" {
                 break
```
iSazonov: It seems we can use Wait-Process -Timeout instead of "while" cycle.

test/powershell/Language/Scripting/NativeExecution/NativeCommandProcessor.Tests.ps1
```diff
@@ -72,7 +72,7 @@ Describe "Native Command Processor" -tags "Feature" {
...
             }
         }
-        $childprocesses = Get-Process | Where-Object {$_.Name -like 'createchildproc*'}
+        $childprocesses = Get-Process testexe
         $count = $childprocesses.count
         $childprocesses | Stop-Process
```
iSazonov: Again "Stop-Process -Force"

test/powershell/Language/Scripting/NativeExecution/NativeStreams.Tests.ps1
```diff
@@ -1,70 +1,70 @@
-Describe "Native streams behavior with PowerShell" -Tags 'CI' {
```
iSazonov: Why was the Describe removed and re-added?

test/powershell/Language/Scripting/NativeExecution/NativeStreams.Tests.ps1
```diff
@@ -1,70 +1,70 @@
...
+        $powershellTestDir = Split-Path $powershellTestDir
+    }
+    $testexe = Join-Path (Split-Path $powershellTestDir) tools/TestExe/bin/testexe
+
+    It 'doesn''t throw any exceptions, when we are piping to the closed executable' {
+        1..3 | % {
+            Start-Sleep -Milliseconds 100
+            # yeild some multi-line formatted object
+            @{'a' = 'b'}
+        } | & $testexe -echoargs | Should Be $null
```
iSazonov: I don't understand the test. It seems it does not match its description.

-

test/tools/TestExe/TestExe.cs
```diff
@@ -0,0 +1,47 @@
...
+    class TestExe
+    {
+        static void Main(string[] args)
+        {
+            if (args.Length > 0)
+            {
+                switch(args[0])
+                {
+                    case "-echoargs":
+                        for (int i = 1; i < args.Length; i++)
```
iSazonov: Maybe put the test code in each case in a function if we plan to add multiple tests?
And it would be good to add comments describing the purpose of the test code.

iSazonov (commented):
Leave a comment

iSazonov:
@SteveL-MSFT Sorry, most of my comments go beyond your changes.

test/powershell/Language/Scripting/NativeExecution/NativeCommandArguments.Tests.ps1
SteveL-MSFT: Good suggestion.  I was only modifying the existing code for the new exe without changing any existing logic.  Similar to how we add the test modules to PSModulePath, I'm thinking of just adding the testexe to PATH.

build.psm1
SteveL-MSFT: I was trying to follow the existing convention for EchoArgs and CreateChildProcess.  Since DotNet calls it an exe, I think TestExe is appropriate.  "TestRunner" to me means something that runs tests.

test/powershell/Language/Scripting/NativeExecution/NativeCommandProcessor.Tests.ps1
SteveL-MSFT: Docs must be wrong, it doesn't prompt by default unless you specify `-confirm`.  I'll submit a doc fix.

test/powershell/Language/Scripting/NativeExecution/NativeStreams.Tests.ps1
SteveL-MSFT: Not sure why the diff is showing up this way.  I made edits directly replacing EchoArgs in this case.

test/powershell/Language/Scripting/NativeExecution/NativeStreams.Tests.ps1
SteveL-MSFT: I didn't make changes to the test.  I think if you have concerns about the test, we should handle that as a separate issue and PR.  I'm only trying to converge to a single test executable.

test/tools/TestExe/TestExe.cs
SteveL-MSFT: I'm fine with that, will change.

test/powershell/Language/Scripting/NativeExecution/NativeCommandProcessor.Tests.ps1
SteveL-MSFT: Doc is wrong

test/powershell/Language/Scripting/NativeExecution/NativeCommandProcessor.Tests.ps1
SteveL-MSFT: I didn't make changes to the test. I think if you have concerns about the test, we should handle that as a separate issue and PR. I'm only trying to converge to a single test executable.

test/powershell/Language/Scripting/NativeExecution/NativeCommandProcessor.Tests.ps1
SteveL-MSFT: https://github.com/PowerShell/PowerShell-Docs/pull/1276

test/powershell/Language/Scripting/NativeExecution/NativeStreams.Tests.ps1
SteveL-MSFT: I think it's because VSCode change the newline from LF to CRLF

build.psm1
iSazonov: Closed.

test/powershell/Language/Scripting/NativeExecution/NativeCommandProcessor.Tests.ps1
iSazonov: Closed.

test/powershell/Language/Scripting/NativeExecution/NativeCommandProcessor.Tests.ps1
iSazonov: Closed.

test/powershell/Language/Scripting/NativeExecution/NativeStreams.Tests.ps1
iSazonov: Closed.

test/tools/TestExe/TestExe.cs
iSazonov: Closed.

test/powershell/Language/Scripting/NativeExecution/NativeCommandProcessor.Tests.ps1
iSazonov: I left a comment in #2561.

Closed.

test/powershell/Language/Scripting/NativeExecution/NativeStreams.Tests.ps1
iSazonov: Closed.

test/powershell/Language/Scripting/NativeExecution/NativeCommandArguments.Tests.ps1
iSazonov: Good catch!

Closed.

iSazonov (approved):

SteveL-MSFT:
@lzybkr can you merge?  thanks

test/tools/TestExe/TestExe.cs
```diff
@@ -0,0 +1,64 @@
...
+namespace TestExe
+{
+    class TestExe
+    {
+        static void Main(string[] args)
+        {
+            if (args.Length > 0)
+            {
+                switch(args[0])
+                {
```
lzybkr: Maybe add `.ToLowerInvariant` to save folks some hassle if they use mixed case.

build.psm1
```diff
@@ -681,25 +681,29 @@ function Publish-PSTestTools {
...
     if ($Options -eq $null)
```
lzybkr: This shouldn't be an array of arrays, it should be an array of custom objects.

SteveL-MSFT: How about an array of hashtables?

test/tools/TestExe/TestExe.cs
SteveL-MSFT: sounds good

build.psm1
lzybkr: That works too, but `[pscustomobject]@{Path="${PSScriptRoot}/test/tools/TestExe";Output="testexe"}` expresses the intent more clearly to me.

---
In reply to: [121528687](https://github.com/PowerShell/PowerShell/pull/3982#discussion_r121528687) [](ancestors = 121528687)

PR #4006, opened 2017-06-14
Enhance formatting to exclude explicit newlines when writing errors

### LINKED ISSUE #1491: We shouldn't add explicit newlines when writing errors
## Steps to reproduce

On Windows

```
PS> throw ("1234567890"*([Console]::BufferWidth/10 + 10))
```

then with the mouse, select the text of the error message, paste on something like notepad
## Expected behavior

no unnecessary line breaks
## Actual behavior

line breaks at the buffer width
## Environment data

Win10 any version

### LINKED ISSUE #3813: StdErr is wrapped as ErrorRecord producing undesired effects
Steps to reproduce
------------------

Run some executable that outputs a long line to stderr and redirect it.
For example

```
python -c "from __future__ import print_function`nimport sys`nprint(range(100), file=sys.stderr)" *> 1
```

Real-life scenario that's highly impacted by that problem is searching in the logs.
When log file can be arbitrary split on the search tokens, you cannot reliable find anything in the logs.

Expected behavior
-----------------

One line in the file `1`

Actual behavior
---------------

Several lines in the output file.
Compare it with the redirection of `stdout` which works as expected.

```
PS /Users/svorobev/junk> python -c "from __future__ import print_function`nimport sys`nprint(range(100), file=sys.stdout)" *> 1                           
PS /Users/svorobev/junk> cat 1 | measure                                                                                                                  


Count    : 1
Average  : 
Sum      : 
Maximum  : 
Minimum  : 
Property : 



PS /Users/svorobev/junk> python -c "from __future__ import print_function`nimport sys`nprint(range(100), file=sys.stderr)" *> 1                           
PS /Users/svorobev/junk> cat 1 | measure                                                                                                                  


Count    : 3
Average  : 
Sum      : 
Maximum  : 
Minimum  : 
Property : 
```

Environment data
----------------

```powershell
> $PSVersionTable
CLRVersion                                                                                                                                               
GitCommitId                    v6.0.0-beta.1                                                                                                             
OS                             Darwin 16.6.0 Darwin Kernel Version 16.6.0: Fri Apr 14 16:21:16 PDT 2017; root:xnu-3789.60.24~6/RELEASE_X86_64            
```

iSazonov:
Fix #3813 
Fix #1491

Problem
---
Newlines are added by console width when ErrorRecord formatted.

Solution
---
Override the format width with Int32.MaxValue for ErrorRecord type.

### DIFF
--- a/src/System.Management.Automation/commands/utility/FormatAndOutput/common/DefaultFormatters/PowerShellCore_format_ps1xml.cs
+++ b/src/System.Management.Automation/commands/utility/FormatAndOutput/common/DefaultFormatters/PowerShellCore_format_ps1xml.cs
@@ -717,7 +717,7 @@ namespace System.Management.Automation.Runspaces
         private static IEnumerable<FormatViewDefinition> ViewsOf_System_Management_Automation_ErrorRecord()
         {
             yield return new FormatViewDefinition("ErrorInstance",
-                CustomControl.Create(outOfBand: true)
+                CustomControl.Create(outOfBand: true, width: Int32.MaxValue)
                     .StartEntry()
                         .AddScriptBlockExpressionBinding(@"
                                     if (($_.FullyQualifiedErrorId -ne ""NativeCommandErrorMessage"" -and $_.FullyQualifiedErrorId -ne ""NativeCommandError"") -and $ErrorView -ne ""CategoryView"")
--- a/src/System.Management.Automation/commands/utility/FormatAndOutput/common/DisplayDatabase/displayDescriptionData.cs
+++ b/src/System.Management.Automation/commands/utility/FormatAndOutput/common/DisplayDatabase/displayDescriptionData.cs
@@ -546,6 +546,11 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
         internal bool outOfBand;
 
         /// <summary>
+        /// Override default console host width.
+        /// </summary>
+        internal int? width;
+
+        /// <summary>
         /// Set if the view is for help output, used so we can prune the view from Get-FormatData
         /// because those views are too complicated and big for remoting.
         /// </summary>
@@ -712,6 +717,11 @@ namespace System.Management.Automation
         /// regardless of previous objects that may have selected the shape.
         /// </summary>
         public bool OutOfBand { get; set; }
+
+        /// <summary>
+        /// Set explicit limit for Width. Default width is a console host width.
+        /// </summary>
+        public int? Width { get; set; }
 
         internal abstract void WriteToXml(FormatXmlWriter writer);
 
--- a/src/System.Management.Automation/engine/serialization.cs
+++ b/src/System.Management.Automation/engine/serialization.cs
@@ -7428,6 +7428,7 @@ namespace Microsoft.PowerShell
 
             result.GroupBy = GetPropertyValue<PSControlGroupBy>(deserializedControl, "GroupBy", RehydrationFlags.MissingPropertyOk);
             result.OutOfBand = GetPropertyValue<bool>(deserializedControl, "OutOfBand", RehydrationFlags.MissingPropertyOk);
+            result.Width = GetPropertyValue<int?>(deserializedControl, "Width", RehydrationFlags.MissingPropertyOk);
             return result;
         }
 
--- a/src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs
+++ b/src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs
@@ -540,6 +540,8 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
 
         private void ProcessOutOfBandPayload(FormatEntryData fed)
         {
+            var screenColumnWidth = fed.width ?? _lo.ColumnNumber;
+
             // try if it is raw text
             RawTextFormatEntry rte = fed.formatEntryInfo as RawTextFormatEntry;
             if (rte != null)
@@ -548,7 +550,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 {
                     ComplexWriter complexWriter = new ComplexWriter();
 
-                    complexWriter.Initialize(_lo, _lo.ColumnNumber);
+                    complexWriter.Initialize(_lo, screenColumnWidth);
                     complexWriter.WriteString(rte.text);
                 }
                 else
@@ -565,7 +567,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             {
                 ComplexWriter complexWriter = new ComplexWriter();
 
-                complexWriter.Initialize(_lo, _lo.ColumnNumber);
+                complexWriter.Initialize(_lo, screenColumnWidth);
                 complexWriter.WriteObject(cve.formatValueList);
 
                 return;
@@ -579,7 +581,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 _lo.WriteLine("");
 
                 string[] properties = ListOutputContext.GetProperties(lve);
-                listWriter.Initialize(properties, _lo.ColumnNumber, _lo.DisplayCells);
+                listWriter.Initialize(properties, screenColumnWidth, _lo.DisplayCells);
                 string[] values = ListOutputContext.GetValues(lve);
                 listWriter.WriteProperties(values, _lo);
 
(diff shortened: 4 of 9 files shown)


src/System.Management.Automation/commands/utility/FormatAndOutput/common/DefaultFormatters/PowerShellCore_format_ps1xml.cs
```diff
@@ -717,7 +717,7 @@ private static IEnumerable<FormatViewDefinition> ViewsOf_System_Management_Autom
         private static IEnumerable<FormatViewDefinition> ViewsOf_System_Management_Automation_ErrorRecord()
         {
             yield return new FormatViewDefinition("ErrorInstance",
-                CustomControl.Create(outOfBand: true)
+                CustomControl.Create(outOfBand: true, width: Int32.MaxValue)
                     .StartEntry()
```
lzybkr: I know I've tried something similar, and the formatting system decided to pad out the width with spaces - I ended up with huge log files.

Can you see if that's still a problem?

src/System.Management.Automation/commands/utility/FormatAndOutput/common/DisplayDatabase/displayDescriptionData.cs
```diff
@@ -713,6 +718,11 @@ public abstract class PSControl
         /// </summary>
         public bool OutOfBand { get; set; }
 
+        /// <summary>
+        /// Set explicit limit for Width. Default width is a console host width.
+        /// </summary>
+        public int? Width { get; set; }
+
```
lzybkr: New properties in the public Format api surface require special code in deserialization for remoting - see [here](https://github.com/PowerShell/PowerShell/blob/8731747c3e412edf537c3dc0b4ce6a09e6fd7d95/src/System.Management.Automation/engine/serialization.cs#L7382)

Note that you need to use `RehydrationFlags.MissingPropertyOk`, otherwise remoting interop with Windows PowerShell will not work.

It's possible you might also need to update [this method](https://github.com/PowerShell/PowerShell/blob/8731747c3e412edf537c3dc0b4ce6a09e6fd7d95/src/System.Management.Automation/commands/utility/FormatAndOutput/common/DisplayDatabase/displayDescriptionData.cs#L723) but it's possible serialization of this property won't have any effect on Windows PowerShell, so it might be harmless.

src/System.Management.Automation/commands/utility/FormatAndOutput/common/DefaultFormatters/PowerShellCore_format_ps1xml.cs
iSazonov: Your test code:
```powershell
throw ("1234567890"*([Console]::BufferWidth/10 + 10))
```
don't pad to Int32.MaxValue lines if you meant this.

vors:
@iSazonov This is awesome, thank you for tacking this! 🎉

src/System.Management.Automation/commands/utility/FormatAndOutput/common/DisplayDatabase/displayDescriptionData.cs
iSazonov: We still don't have tests for this area. 😕 
I added `Width` in `RehydratePSControl`. I don't changed 'CompatibleWithOldPowerShell'.
I still don't understand how serialization works in depth so I rely to your experience.

src/System.Management.Automation/commands/utility/FormatAndOutput/common/DefaultFormatters/PowerShellCore_format_ps1xml.cs
lzybkr: Thanks - I tried your code and agree that it seems to work in a reasonable way. I'm still concerned though :)

src/System.Management.Automation/commands/utility/FormatAndOutput/common/DisplayDatabase/displayDescriptionData.cs
lzybkr: I believe @JamesWTruher was working on remoting interop tests - hopefully we'll get those at some point.

src/System.Management.Automation/engine/serialization.cs
```diff
@@ -7428,6 +7428,7 @@ private static PSControl RehydratePSControl(PSObject deserializedControl)
 
             result.GroupBy = GetPropertyValue<PSControlGroupBy>(deserializedControl, "GroupBy", RehydrationFlags.MissingPropertyOk);
             result.OutOfBand = GetPropertyValue<bool>(deserializedControl, "OutOfBand", RehydrationFlags.MissingPropertyOk);
+            result.Width = GetPropertyValue<int?>(deserializedControl, "Width", RehydrationFlags.MissingPropertyOk);
```
lzybkr: I think this needs to be `int`, not `int?`.

src/System.Management.Automation/commands/utility/FormatAndOutput/common/DisplayDatabase/displayDescriptionData.cs
lzybkr: Oh, thinking about the property some more, I had some concerns. If we're thinking about files, width should really be `uint64`. To keep the code simpler, we could disallow values > `int.MaxValue`, but this would future proof the api to avoid adding `Width64`.

lzybkr (changes requested):
I've thought about this some more and have some concerns with this approach and I'd like to ask you to explore other approaches.

Today, when writing an error to the console, we sometimes truncate (I see that with `Write-Error`), but we also word wrap - breaking long lines at word boundaries.

This proposed change stops doing both of those things.

One concern I have about this approach - it introduces a new formatting feature that is opt-in, but the real problem is that any custom formatting incorrectly uses the console width when output is going to a file.

In my mind - nobody should have to specify anything in their custom format specification - long lines should simply not get wrapped when output goes to a file.  Errors might be the most common example, but there must be others.

That said, the formatting engine is complex and I don't know how we would go about detecting the output target.

Another concern - you've added a property to the base of formatting, but only tested it with a complex control. How does this property work with tables? Lists? And for these formats, does it even make sense to override the console width? I believe that enables users to create formats that don't work well for some users.

src/System.Management.Automation/commands/utility/FormatAndOutput/common/DisplayDatabase/displayDescriptionData.cs
iSazonov: Current limit is [Short.MaxValue](https://github.com/dotnet/corefx/blob/ec2a6190efa743ab600317f44d757433e44e859b/src/Common/src/System/IO/PathInternal.Windows.cs#L52).  My thoughts were that we had a good reserve with `Int32`. But I agree that we can forget this forever with `uint64`.

src/System.Management.Automation/commands/utility/FormatAndOutput/common/DefaultFormatters/PowerShellCore_format_ps1xml.cs
iSazonov: Only with the third attempt, I started to understand how the formating system works. Although it could be a hoax. 😄

iSazonov:
>Another concern - you've added a property to the base of formatting, but only tested it with a complex control.

I think we should make the "Width" property and the constructor "public static CustomControlBuilder Create(bool outOfBand = false, int? width = null)" as `internal` - now I don't see real scenarios where users can use this. So we solve the problem with error formatting and we should not consider these (list and table) options yet, pending feedback. (A possible scenario is an output to "fixed-width" text files.)

>nobody should have to specify anything in their custom format specification - long lines should simply not get wrapped when output goes to a file.

Good point! It seems the bevavior based on the [LineOutput property ](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs#L597)and [LineOutput class ](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/commands/utility/FormatAndOutput/common/ILineOutput.cs#L138) - we could make it more generic to support not only a host but also files. Right now I don't really know where is a right place to initialize it. Maybe it's in internal [Out-LineOutput](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/commands/utility/FormatAndOutput/out-textInterface/OutTextInterface.cs#L19) cmdlet
Here we fix only error formating for host. For file output formating it would be good to open a new Issue. It'll give us more time to think.

vors:
What's the status of this PR?

> I'd like to ask you to explore other approaches.

We are not going ahead with the current approach, correct?

iSazonov:
>What's the status of this PR?

Under review and discussion.  @lzybkr is very busy...

lzybkr:
I'm going to close this PR because I don't think it's the right approach. Wrapping should be automatic in fixed width output devices, but should not wrap otherwise (e.g. files or infinite width hosts like the ISE).

This behavior should be automatic - not requiring opt-in via a new property.

PR #4010, opened 2017-06-14
Fixing Import-Module on Linux for special cases of NestedModules/RootModule path format

### LINKED ISSUE #3693: NestedModules does not work on Linux if the path starts with '.' e.g. '.\folder\library.dll'
Steps to reproduce
------------------
In psd1 have the following lines.
NestedModules = @('.\netstandard1.6\Microsoft.Azure.Commands.Profile.dll') 
FormatsToProcess = '.\netstandard1.6\Microsoft.Azure.Commands.Profile.format.ps1xml'

Expected behavior
-----------------
Both files could be found and loaded throug import-module

Actual behavior
---------------
ps1xml has no issues while Microsoft.Azure.Commands.Profile.dll file fails with obscure error message:

PS /usr/local/src> Import-Module AzureRM.Profile.Netcore -Verbose -Debug
VERBOSE: Loading module from path
'/opt/microsoft/powershell/6.0.0-alpha.18/Modules/AzureRM.Profile.Netcore/1.0.0/AzureRM.Profile.Netcore.psd1'.
VERBOSE: Loading 'FormatsToProcess' from path
'/opt/microsoft/powershell/6.0.0-alpha.18/Modules/AzureRM.Profile.Netcore/1.0.0/netstandard1.6/Microsoft.Azure.Commands.Profile
.format.ps1xml'.
VERBOSE: Populating RepositorySourceLocation property for module AzureRM.Profile.Netcore.
Import-Module : **The given assembly name or codebase was invalid. (Exception from HRESULT: 0x80131047)**
At line:1 char:1
+ Import-Module AzureRM.Profile.Netcore -Verbose -Debug
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Import-Module], FileLoadException
    + FullyQualifiedErrorId : System.IO.FileLoadException,Microsoft.PowerShell.Commands.ImportModuleCommand 

Environment data
----------------
Red Hat Enterprise Linux 7

```powershell
> $PSVersionTable
GitCommitId                    v6.0.0-alpha.18
CLRVersion
```

## Current Workaround
Remove '.\' at the beginning of each NestedModule and everything works fine on Windows, Linux and Mac OS. 
It is inconsistent and the worst thing is the error message.

anmenaga:
NestedModules/RootModule are fields in a module manifest (.psd1) that reference other modules using string paths. Assuming that a module is written to be cross-platform, the same psd1 module manifest should be successfully imported on both Windows and Linux platforms.
The bug is that in some cases Import-Module on Linux fails for a totally valid module. It depends on format of NestedModules/RootModule paths.

The root of the problem is using simple Path.Combine for constructing a path to the submodule, that is then tested using NativeItemExists methods, resulting in calls like
`Unix.NativeMethods.IsFile("/home/testuser/TestRootModule/.\SubModule\SubModule.psm1")`
which fails and Import-Module returns an error.

The fix is to resolve submodule paths using existing FileSystem provider utilities.

Test results on Ubuntu 14 after the fix:
afterfix

Test results on Ubuntu 14 before the fix:
beforefix

Fix #3693


test/powershell/engine/Module/SubmodulePathInManifest.Tests.ps1
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    It "Test if RootModule path is <SubModulePath>" -TestCases $testCases {
+        param($SubModulePath)
+    
+        New-ModuleManifest $moduleFilePath -RootModule $SubModulePath
+        Import-Module $moduleFilePath
+        (Get-Module $moduleName).ExportedCommands.Keys.Contains('TestModuleFunction') | Should Be $true
+    }
+}
```
SteveL-MSFT: The little red icon means you should have a trailing linefeed at the end of the file

anmenaga: I always forget about that newline...   Updated.

SteveL-MSFT (approved):
LGTM

mirichmo (approved):

PR #4019, opened 2017-06-15
Change positional parameter for powershell.exe from -Command to -File

SteveL-MSFT:
Previously powershell.exe treated unknown arguments as a command line to execute.  To align with POSIX so that things like shebang scripts work correctly, we are changing powershell.exe so that it treats unknown arguments (aka positional argument) as a file.  This means that `powershell foo` will now attempt to use `foo` as a PowerShell script whereas previously `foo` would be treated as a command to execute.  This doesn't affect existing usage of either `-File` nor `-Command`.  Fixed tests that didn't explicitly use `-Command` parameter.

Note that some of the changes below for trailing whitespace was due to VSCode setting.

Fix https://github.com/PowerShell/PowerShell/issues/1103
Fix https://github.com/PowerShell/PowerShell/issues/3959

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
@@ -499,10 +499,10 @@ namespace Microsoft.PowerShell
 
                 if (!SpecialCharacters.IsDash(switchKey[0]) && switchKey[0] != '/')
                 {
-                    // then its a command
+                    // then its a file
 
                     --i;
-                    ParseCommand(args, ref i, noexitSeen, false);
+                    ParseFile(args, ref i, noexitSeen);
                     break;
                 }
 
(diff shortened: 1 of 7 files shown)


rkeithhill:
@SteveL-MSFT You might find limiting that settting to just PowerShell files to be handy - especially for this repo e.g.:
```
    "[powershell]": {
        "files.trimTrailingWhitespace": true
    },
```

lzybkr:
You probably should fix the resx file to only include your intended change - the other whitespace changes do have a visible effect in the binary, which is probably fine, but should be in a different commit.

---
In reply to: [308830296](https://github.com/PowerShell/PowerShell/pull/4019#issuecomment-308830296) [](ancestors = 308830296)

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -1820,7 +1820,17 @@ private void DoRunspaceInitialization(bool importSystemModules, bool skipProfile
...
 
                 Pipeline tempPipeline = exec.CreatePipeline();
-                Command c = new Command(filePath, false, false);
+                Command c;
+                // if file doesn't have .ps1 extension, we read the contents and treat it as a script to support shebang with no .ps1 extension usage
+                if (!Path.GetExtension(filePath).Equals(".ps1", StringComparison.OrdinalIgnoreCase))
+                {
+                    string script = File.ReadAllText(filePath);
+                    c = new Command(script, isScript: true, useLocalScope: false);
+                }
```
lzybkr: Does this set the exit code correctly? You don't have a test - you should add one as well.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -967,6 +851,122 @@ private void ParseExecutionPolicy(string[] args, ref int i, ref string execution
             executionPolicy = args[i];
         }
 
+        private bool ParseFile(string[] args, ref int i, bool noexitSeen)
+        {
```
lzybkr: C# now supports local functions - that would make this a bit simpler because you wouldn't the parameters.

alexandair:
Isn't this a breaking change? It should be documented in the release notes.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: Making ParseFile a local function would make it inconsistent with similar methods like ParseCommand(), ParseFormat(), ParseExecutionPolicy(), etc...

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
SteveL-MSFT: Will add test

SteveL-MSFT:
@lzybkr any other feedback, would like to get this into beta.3 since it's a breaking change and get the feedback

lzybkr:
:shipit:

lzybkr (approved):

PR #4020, opened 2017-06-15
Make Get-ChildItem follow symlinks on demand, with checks for link loops

### LINKED ISSUE #3951: Need the ability to opt into / out of following symlinks with Get-ChildItem -Recurse, behavior differs between editions
Currently, the behavior with respect to following symlinks to directories with `Get-ChildItem -Recurse`:

* differs between editions

* is _invariable_ in either case.

_Windows_ PowerShell _always_ follows directory symlinks, whereas PowerShell _Core_ _never_ does  - due to #3780.
Note that the Core behavior also applies on Windows.

Given that `Get-ChildItem` is not just the analog of `ls`, but also of `find`, there should at least be a way to _opt in_ with respect to symlink following, as with `find -L`.

Note that reintroducing symlink recursion would reintroduce the need to detect symlink _loops_, as reported in #1875.

Unless infeasible due to backward compatibility concerns, it's also worth considering changing the Windows edition's default behavior to match the Core edition's.

Steps to reproduce
------------------

Note: On _Windows_, you must run the code from an _elevated_ console.

```powershell
# Create 2 temp. dirs.
$tmpDirRoot = [IO.Path]::GetTempPath()
$tmpDir1 = join-path $tmpDirRoot "$PID-a"
$tmpDir2 = join-path $tmpDirRoot "$PID-b"
$null = New-Item -Type Directory -Path $tmpDir1, $tmpDir2

Push-Location $tmpDir1

# Create 1 file each.
'' > "$tmpDir1/f1"
'' > "$tmpDir2/f2"

# In tmpDir1, create a symlink to tmpDir2
$null = New-Item -Type SymbolicLink dirlink -Value $tmpDir2

# Recurse and count the items.
# If the directory symlink is followed, 3 items are reported, otherwise 2.
$count = (Get-ChildItem -Recurse).Count

[pscustomobject] @{ 'Actual count' = $count; 'Count w/ following' = 3; 'Count w/o following' = 2 }

# Comment out the following 2 lines to prevent cleanup.
Pop-Location
Remove-Item -Force -Recurse $tmpDir1, $tmpDir2
```

Actual behavior
---------------

Windows PowerShell:

```none
Actual count Count w/ following Count w/o following
------------ ------------------ -------------------
           3                  3                   2
```

PowerShell Core:

```none
Actual count Count w/ following Count w/o following
------------ ------------------ -------------------
           2                  3                   2
```

Environment data
----------------

```powershell
PowerShell Core v6.0.0-beta (v6.0.0-beta.2) on macOS 10.12.5
PowerShell Core v6.0.0-beta (v6.0.0-beta.2) on Ubuntu 16.04.1 LTS
PowerShell Core v6.0.0-beta (v6.0.0-beta.2) on Microsoft Windows 10 Pro (64-bit; v10.0.14393)
Windows PowerShell v5.1.14393.1198 on Microsoft Windows 10 Pro (64-bit; v10.0.14393)
```

jeffbi:
Fixes #3951

    Add the dynamic parameter -FollowSymlink to Get-ChildItem.
    Add a mechanism for tracking visited directories.
    Add native code to get device/inode information on Unix/Windows.
    Add warning when refusing to enter an already-visited directory.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/GetChildrenCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/GetChildrenCommand.cs
@@ -172,25 +172,6 @@ namespace Microsoft.PowerShell.Commands
         }
 
         /// <summary>
-        /// Gets or sets whether symbolic links are to be followed when recursing.
-        /// </summary>
-        [Parameter]
-        public SwitchParameter FollowSymlink
-        {
-            get
-            {
-                return _followSymlink;
-            }
-            set
-            {
-                _followSymlink = value;
-                if (value == true)
-                    Recurse = true;
-            }
-        }
-        private bool _followSymlink = false;
-
-        /// <summary>
         /// Gets or sets the force property
         /// </summary>
         ///
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/GetChildrenCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/GetChildrenCommand.cs
@@ -184,6 +184,8 @@ namespace Microsoft.PowerShell.Commands
             set
             {
                 _followSymlink = value;
+                if (value == true)
+                    Recurse = true;
             }
         }
         private bool _followSymlink = false;
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/GetChildrenCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/GetChildrenCommand.cs
@@ -172,6 +172,23 @@ namespace Microsoft.PowerShell.Commands
         }
 
         /// <summary>
+        /// Gets or sets whether symbolic links are to be followed when recursing.
+        /// </summary>
+        [Parameter]
+        public SwitchParameter FollowSymlink
+        {
+            get
+            {
+                return _followSymlink;
+            }
+            set
+            {
+                _followSymlink = value;
+            }
+        }
+        private bool _followSymlink = false;
+
+        /// <summary>
         /// Gets or sets the force property
         /// </summary>
         ///
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -31,56 +31,6 @@ using Microsoft.PowerShell.CoreClr.Stubs;
 
 namespace Microsoft.PowerShell.Commands
 {
-    #region InodeTracker
-    /// <summary>
-    /// Tracks visited files/directories by caching their device IDs and inodes.
-    /// </summary>
-    internal class InodeTracker
-    {
-        private Dictionary<(UInt64, UInt64), bool>  _visitations;
-
-        /// <summary>
-        /// Construct a new InodeTracker with an initial path
-        /// </summary>
-        internal InodeTracker(string path)
-        {
-            _visitations = new Dictionary<(UInt64, UInt64), bool>();
-            var inodeData = (0UL, 0UL);
-
-            if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out inodeData))
-            {
-                _visitations[inodeData] = true;
-            }
-        }
-
-        /// <summary>
-        /// Try to mark a path as having been visited.
-        /// </summary>
-        /// <param name="path">Path to the file or directory to be marked as visited.</param>
-        /// <returns>
-        /// False if the path has already been visited.
-        /// True if the path has not yet been visited.
-        /// </returns>
-        /// <remarks>
-        /// If the path has not yet been visited then before this method returns it marks
-        /// the path as having been visited.
-        /// </remarks>
-        internal bool TryVisitPath(string path)
-        {
-            bool rv = true;
-            var inodeData = (0UL, 0UL);
-
-            if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out inodeData))
-            {
-                rv = _visitations.TryAdd(inodeData, true);
-            }
-
-            return rv;
-        }
-    }
-
-    #endregion
-
     #region FileSystemProvider
 
     /// <summary>
@@ -1591,14 +1541,17 @@ namespace Microsoft.PowerShell.Commands
                     DirectoryInfo directory = new DirectoryInfo(path);
                     InodeTracker tracker = null;
 
-                    GetChildDynamicParameters fspDynamicParam = DynamicParameters as GetChildDynamicParameters;
-                    if (fspDynamicParam != null && fspDynamicParam.FollowSymlink)
+                    if (recurse)
                     {
-                        tracker = new InodeTracker(directory.FullName);
+                        GetChildDynamicParameters fspDynamicParam = DynamicParameters as GetChildDynamicParameters;
+                        if (fspDynamicParam != null && fspDynamicParam.FollowSymlink)
+                        {
+                            tracker = new InodeTracker(directory.FullName);
+                        }
                     }
 
                     // Enumerate the directory
-                    Dir(directory, recurse, depth, tracker, nameOnly, returnContainers);
+                    Dir(directory, recurse, depth, nameOnly, returnContainers, tracker);
                 }
                 else
                 {
@@ -1666,9 +1619,9 @@ namespace Microsoft.PowerShell.Commands
             DirectoryInfo directory,
             bool recurse,
             uint depth,
-            InodeTracker tracker,   // This will be non-null only if the user invoked the -FollowSymLinks switch parameter.
             bool nameOnly,
-            ReturnContainers returnContainers)
+            ReturnContainers returnContainers,
+            InodeTracker tracker)   // tracker will be non-null only if the user invoked the -FollowSymLinks switch parameter.
         {
             List<IEnumerable<FileSystemInfo>> target = new List<IEnumerable<FileSystemInfo>>();
 
@@ -1868,17 +1821,14 @@ namespace Microsoft.PowerShell.Commands
                                         continue;
                                     }
                                 }
-                                else
+                                else if (tracker.IsPathVisited(recursiveDirectory.FullName, true))
                                 {
-                                    if (!tracker.TryVisitPath(recursiveDirectory.FullName))
-                                    {
-                                        WriteWarning(StringUtil.Format(FileSystemProviderStrings.AlreadyListedDirectory,
-                                                                    recursiveDirectory.FullName));
-                                        continue;
-                                    }
+                                    WriteWarning(StringUtil.Format(FileSystemProviderStrings.AlreadyListedDirectory,
+                                                                   recursiveDirectory.FullName));
+                                    continue;
                                 }
 
-                                Dir(recursiveDirectory, recurse, depth - 1, tracker, nameOnly, returnContainers);
+                                Dir(recursiveDirectory, recurse, depth - 1, nameOnly, returnContainers, tracker);
                             }
                         }//foreach
                     }//if
@@ -7392,6 +7342,69 @@ namespace Microsoft.PowerShell.Commands
             [MarshalAs(UnmanagedType.LPWStr)]
             public string Provider;
         }
+
+        #region InodeTracker
+        /// <summary>
+        /// Tracks visited files/directories by caching their device IDs and inodes.
+        /// </summary>
+        private class InodeTracker
+        {
+            private HashSet<(UInt64, UInt64)> _visitations;
+
+            /// <summary>
+            /// Construct a new InodeTracker with an initial path
+            /// </summary>
+            internal InodeTracker(string path)
+            {
+                _visitations = new HashSet<(UInt64, UInt64)>();
+
+                if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out (UInt64, UInt64) inodeData))
+                {
+                    _visitations.Add(inodeData);
+                }
+            }
+
+            /// <summary>
+            /// Determine if a path has been visited. Optionally mark the path as
+            /// having been visited if it has not already
+            /// </summary>
+            /// <param name="path">
+            /// Path to the file system item to be checked.
+            /// </param>
+            /// <param name="visitIfNotAlready">
+            /// Flag to indicate whether the path should be marked as visited if
+            /// it has not already been visited.
+            /// </param>
+            /// <returns>
+            /// True if the path had already been visited, false otherwise.
+            /// </returns>
+            /// <remarks>
+            /// If the visitIfNotAlready parameter is true and the path had not
+            /// already been visited, this method returns false but marks the
+            /// path as visited so that subsequent calls with a path to the same
+            /// file system item will return true.
+            /// </remarks>
+            internal bool IsPathVisited(string path, bool visitIfNotAlready)
+            {
+                bool isPathVisited = false;
+
+                if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out (UInt64, UInt64) inodeData))
+                {
+                    if (visitIfNotAlready)
+                    {
+                        isPathVisited = !_visitations.Add(inodeData);
+                    }
+                    else
+                    {
+                        isPathVisited = _visitations.Contains(inodeData);
+                    }
+                }
+
+                return isPathVisited;
+            }
+        }
+
+        #endregion
     } // class FileSystemProvider
 
     internal static class SafeInvokeCommand
@@ -7556,7 +7569,7 @@ namespace Microsoft.PowerShell.Commands
 
         /// <summary>
         /// Gets or sets the flag to follow symbolic links when recursing.
-        ///
+        /// </summary>
         [Parameter]
         public SwitchParameter FollowSymlink { get; set; }
 
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -1821,7 +1821,7 @@ namespace Microsoft.PowerShell.Commands
                                         continue;
                                     }
                                 }
-                                else if (tracker.IsPathVisited(recursiveDirectory.FullName, true))
+                                else if (!tracker.TryVisitPath(recursiveDirectory.FullName))
                                 {
                                     WriteWarning(StringUtil.Format(FileSystemProviderStrings.AlreadyListedDirectory,
                                                                    recursiveDirectory.FullName));
@@ -7365,42 +7365,25 @@ namespace Microsoft.PowerShell.Commands
             }
 
             /// <summary>
-            /// Determine if a path has been visited. Optionally mark the path as
-            /// having been visited if it has not already
+            /// Attempt to mark a path as having been visited.
             /// </summary>
             /// <param name="path">
-            /// Path to the file system item to be checked.
-            /// </param>
-            /// <param name="visitIfNotAlready">
-            /// Flag to indicate whether the path should be marked as visited if
-            /// it has not already been visited.
+            /// Path to the file system item to be visited.
             /// </param>
             /// <returns>
-            /// True if the path had already been visited, false otherwise.
+            /// True if the path had not been previously visited and was
+            /// successfully marked as visited, false otherwise.
             /// </returns>
-            /// <remarks>
-            /// If the visitIfNotAlready parameter is true and the path had not
-            /// already been visited, this method returns false but marks the
-            /// path as visited so that subsequent calls with a path to the same
-            /// file system item will return true.
-            /// </remarks>
-            internal bool IsPathVisited(string path, bool visitIfNotAlready)
+            internal bool TryVisitPath(string path)
             {
-                bool isPathVisited = false;
+                bool returnValue = false;
 
                 if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out (UInt64, UInt64) inodeData))
                 {
-                    if (visitIfNotAlready)
-                    {
-                        isPathVisited = !_visitations.Add(inodeData);
-                    }
-                    else
-                    {
-                        isPathVisited = _visitations.Contains(inodeData);
-                    }
+                    returnValue = _visitations.Add(inodeData);
                 }
 
-                return isPathVisited;
+                return returnValue;
             }
         }
 
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -67,7 +67,6 @@ namespace Microsoft.PowerShell.Commands
         /// Mark a path as visited.
         /// </summary>
         /// <param name="path">Path to the file or directory to be marked as visited.</param>
-        /// <returns>True if the path was successfully added to the visited list, false otherwise</returns>
         internal void VisitPath(string path)
         {
             var inodeData = (0UL, 0UL);
@@ -1674,7 +1673,9 @@ namespace Microsoft.PowerShell.Commands
             ReturnContainers returnContainers)
         {
             if (tracker != null)
+            {
                 tracker.VisitPath(directory.FullName);
+            }
 
             List<IEnumerable<FileSystemInfo>> target = new List<IEnumerable<FileSystemInfo>>();
 
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -1621,7 +1621,7 @@ namespace Microsoft.PowerShell.Commands
             uint depth,
             bool nameOnly,
             ReturnContainers returnContainers,
-            InodeTracker tracker)   // tracker will be non-null only if the user invoked the -FollowSymLinks switch parameter.
+            InodeTracker tracker)   // tracker will be non-null only if the user invoked the -FollowSymLinks and -Recurse switch parameters.
         {
             List<IEnumerable<FileSystemInfo>> target = new List<IEnumerable<FileSystemInfo>>();
 
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -54,14 +54,12 @@ namespace Microsoft.PowerShell.Commands
         /// <returns>True if the path has been visited, false otherwise</returns>
         public bool Visited(string path)
         {
-            UInt64 dev;
-            UInt64 inode;
+            var inodeData = (0UL, 0UL);
 
-            if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out dev, out inode))
+            if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out inodeData))
             {
-                return _visitations.ContainsKey((dev, inode));
+                return _visitations.ContainsKey(inodeData);
             }
-
             return false;
         }
 
@@ -72,12 +70,10 @@ namespace Microsoft.PowerShell.Commands
         /// <returns>True if the path was successfully added to the visited list, false otherwise</returns>
         public bool Visit(string path)
         {
-            UInt64 dev;
-            UInt64 inode;
-
-            if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out dev, out inode))
+            var inodeData = (0UL, 0UL);
+            if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out inodeData))
             {
-                _visitations[(dev, inode)] = true;
+                _visitations[inodeData] = true;
 
                 return true;
             }
@@ -8321,16 +8317,17 @@ namespace Microsoft.PowerShell.Commands
             return false;
         }
 
-        internal static bool GetInodeData(string path, out UInt64 dev, out UInt64 inode)
+        internal static bool GetInodeData(string path, out System.ValueTuple<UInt64, UInt64> inodeData)
         {
 #if UNIX
-            return Platform.NonWindowsGetInodeData(path, out dev, out inode);
+            bool rv = Platform.NonWindowsGetInodeData(path, out inodeData);
 #else
-            return WinGetInodeData(path, out dev, out inode);
+            bool rv = WinGetInodeData(path, out inodeData);
 #endif
+            return rv;
         }
 
-        internal static bool WinGetInodeData(string path, out UInt64 dev, out UInt64 inode)
+        internal static bool WinGetInodeData(string path, out System.ValueTuple<UInt64, UInt64> inodeData)
         {
             var access = FileAccess.Read;
             var share = FileShare.Read;
@@ -8349,16 +8346,14 @@ namespace Microsoft.PowerShell.Commands
                         tmp <<= 32;
                         tmp |= info.FileIndexLow;
 
-                        dev = info.VolumeSerialNumber;
-                        inode = tmp;
+                        inodeData = (info.VolumeSerialNumber, tmp);
 
                         return true;
                     }
                 }
             }
 
-            dev = 0;
-            inode = 0;
+            inodeData = (0, 0);
             return false;
         }
 
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -42,7 +42,7 @@ namespace Microsoft.PowerShell.Commands
         /// <summary>
         /// Construct a new InodeTracker
         /// </summary>
-        public InodeTracker()
+        internal InodeTracker()
         {
             _visitations = new Dictionary<(UInt64, UInt64), bool>();
         }
@@ -52,7 +52,7 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         /// <param name="path">Path to the file or directory to be checked.</param>
         /// <returns>True if the path has been visited, false otherwise</returns>
-        public bool Visited(string path)
+        internal bool IsPathVisited(string path)
         {
             var inodeData = (0UL, 0UL);
 
@@ -68,19 +68,13 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         /// <param name="path">Path to the file or directory to be marked as visited.</param>
         /// <returns>True if the path was successfully added to the visited list, false otherwise</returns>
-        public bool Visit(string path)
+        internal void VisitPath(string path)
         {
             var inodeData = (0UL, 0UL);
 
             if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out inodeData))
             {
                 _visitations[inodeData] = true;
-
-                return true;
-            }
-            else
-            {
-                return false;
             }
         }
     }
@@ -1597,9 +1591,13 @@ namespace Microsoft.PowerShell.Commands
                     DirectoryInfo directory = new DirectoryInfo(path);
                     InodeTracker tracker = null;
 
-                    if (recurse && Context.MyInvocation.BoundParameters.ContainsKey("FollowSymlink"))
+                    GetChildDynamicParameters fspDynamicParam = DynamicParameters as GetChildDynamicParameters;
+                    if (fspDynamicParam != null)
                     {
-                        tracker = new InodeTracker();
+                        if (fspDynamicParam.FollowSymlink)
+                        {
+                            tracker = new InodeTracker();
+                        }
                     }
 
                     // Enumerate the directory
@@ -1676,7 +1674,7 @@ namespace Microsoft.PowerShell.Commands
             ReturnContainers returnContainers)
         {
             if (tracker != null)
-                tracker.Visit(directory.FullName);
+                tracker.VisitPath(directory.FullName);
 
             List<IEnumerable<FileSystemInfo>> target = new List<IEnumerable<FileSystemInfo>>();
 
@@ -1868,7 +1866,7 @@ namespace Microsoft.PowerShell.Commands
                             }
                             else
                             {
-                                if (tracker.Visited(recursiveDirectory.FullName))
+                                if (tracker.IsPathVisited(recursiveDirectory.FullName))
                                 {
                                     WriteWarning(StringUtil.Format(FileSystemProviderStrings.AlreadyListedDirectory,
                                                                 recursiveDirectory.FullName));
@@ -7561,6 +7559,12 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         [Parameter]
         public FlagsExpression<FileAttributes> Attributes { get; set; }
+
+        /// <summary>
+        /// Gets or sets the flag to follow symbolic links when recursing.
+        ///
+        [Parameter]
+        public SwitchParameter FollowSymlink { get; set; }
 
         /// <summary>
         /// Gets or sets the filter directory flag
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -71,6 +71,7 @@ namespace Microsoft.PowerShell.Commands
         public bool Visit(string path)
         {
             var inodeData = (0UL, 0UL);
+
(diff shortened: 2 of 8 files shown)


jeffbi:
@iSazonov, @daxian-dbw can someone look at the AppVeyor failure? Looking at the details, It appears to be failing _very_ early in the run

iSazonov:
@jeffbi I believe it is temporary error - please restart CI by reward your commit.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -8238,6 +8330,47 @@ internal static bool WinIsSameFileSystemItem(string pathOne, string pathTwo)
...
+            var creation = FileMode.Open;
+            var attributes = FileAttributes.BackupSemantics | FileAttributes.PosixSemantics;
+
+            using (var sf = AlternateDataStreamUtilities.NativeMethods.CreateFile(path, access, share, IntPtr.Zero, creation, (int)attributes, IntPtr.Zero))
+            {
+                if (!sf.IsInvalid)
+                {
+                    BY_HANDLE_FILE_INFORMATION info;
+
+                    if (GetFileInformationByHandle(sf.DangerousGetHandle(), out info))
```
iSazonov: Is it supported on Windows 7?

jeffbi: It's supported as early as XP.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1539,8 +1605,15 @@ protected override void InvokeDefaultAction(string path)
                 if (isDirectory)
                 {
                     DirectoryInfo directory = new DirectoryInfo(path);
+                    InodeTracker tracker = null;
+
+                    if (recurse && Context.MyInvocation.BoundParameters.ContainsKey("FollowSymlink"))
```
jeffbi: There are two parameter sets for `Get-ChildItem`: `Items` and `LiteralItems`. The `-Recurse` and `-FollowSymlink` parameters are in both parameter sets.

The `Depth` parameter, which is also in both, implies `-Recurse` by setting the `Recurse` property. Should `-FollowSymlink` follow that pattern? That would remove the check of `recurse` here.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1608,9 +1681,13 @@ protected override void InvokeDefaultAction(string path)
             DirectoryInfo directory,
             bool recurse,
             uint depth,
+            InodeTracker tracker,   // This will be non-null only if the user invoked the -FollowSymLinks switch parameter.
             bool nameOnly,
             ReturnContainers returnContainers)
         {
+            if (tracker != null)
+                tracker.Visit(directory.FullName);
```
jeffbi: This function is called recursively, so I think we need to leave these lines here.

iSazonov: `tracker = new InodeTracker();` - It's not expensive. We can always create this on top level and not check many times during recursion.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: Closed.

src/libpsl-native/src/getinodedata.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include <assert.h>
+#include <errno.h>
+#include <locale.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <string>
+
+//! @brief GetInodeData retrieves a file's device and inode information.
```
jeffbi: This one describes the function. The one at the top of the file describes the file.

I used the existing `getlinkcount.cpp` as the pattern for this.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: In the `Dir` function, the `tracker` object also acts as a flag that the user gave the `-FollowSymlink` parameter.

src/libpsl-native/src/getinodedata.cpp
iSazonov: Closed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: I tried to refactor code and had no benefits.

Closed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1788,12 +1865,27 @@ protected override void InvokeDefaultAction(string path)
...
-                            // Once the recursion process has begun by being in this function,
-                            // we do not want to further recurse into directory symbolic links
-                            // so as to prevent the possibility of an endless symlink loop.
-                            // This is the behavior of both the Unix 'ls' command and the Windows
-                            // 'DIR' command.
-                            if (!InternalSymbolicLinkLinkCodeMethods.IsReparsePoint(recursiveDirectory))
+                            bool enterDir = true;
+                            if (tracker != null)
+                            {
+                                if (tracker.Visited(recursiveDirectory.FullName))
```
iSazonov: It seems we can exclude ` enterDir`:

```c#
                            if (tracker != null)
                            {
                                if (tracker.Visited(recursiveDirectory.FullName))
                                {
                                    WriteWarning(StringUtil.Format(FileSystemProviderStrings.AlreadyListedDirectory,
                                                                    recursiveDirectory.FullName));
                                    continue;
                                }
                            }
                            else
                            {
                                if (InternalSymbolicLinkLinkCodeMethods.IsReparsePoint(recursiveDirectory))
                                {
                                    continue;
                                }
                            }

                            bool hidden = false;
                            if (!Force)
                            {
                                hidden = (recursiveDirectory.Attributes & FileAttributes.Hidden) != 0;
                            }

                            // if "Hidden" is explicitly specified anywhere in the attribute filter, then override
                            // default hidden attribute filter.
                            if (Force || !hidden || isFilterHiddenSpecified || isSwitchFilterHiddenSpecified)
                            {
                                Dir(recursiveDirectory, recurse, depth - 1, tracker, nameOnly, returnContainers);
                            }
```

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1788,12 +1865,27 @@ protected override void InvokeDefaultAction(string path)
...
+                                }
+                            }
+                            // We only want to recurse into symlinks if a) the user has asked to with
+                            // the -FollowSymLinks switch parameter and b) the directory the symlink
+                            // points to has not already been visited, preventing symlink loops.
+                            if (enterDir && InternalSymbolicLinkLinkCodeMethods.IsReparsePoint(recursiveDirectory))
+                            {
+                                if (tracker == null)
+                                {
+                                    enterDir = false;
```
iSazonov: Closed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: We can not define explicitly mandatory sub-parameters 😕 That would be a useful feature.
So it is more safe to leave "recurse" here.

If `-Depth`  implies  `-Recurse` I believe we can do the same with `-FollowSymlink `.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -31,6 +31,72 @@
...
+        public bool Visited(string path)
+        {
+            UInt64 dev;
+            UInt64 inode;
+
+            if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out dev, out inode))
+            {
+                if (_visitations.ContainsKey(dev))
+                {
+                    return _visitations[dev].ContainsKey(inode);
```
jeffbi: Part of that discussion ([https://stackoverflow.com/questions/2877660/composite-key-dictionary#comment14110652_2877820](https://stackoverflow.com/questions/2877660/composite-key-dictionary#comment14110652_2877820)) suggests that the nested dictionary approach may be faster.

As for the C#7 pattern, I thought I read somewhere in another code review that C#7 code should be avoided. Is that not the case?

iSazonov: >C#7 pattern

We have no strict documentation. Different maintainers may have different requirements. My understanding: New patterns are welcome for the new code. Existing patterns (not relevant to the PR) should not be changed without maintainer's requiring.

> the nested dictionary approach may be faster.

There, the main answer has recently been updated and there is a reference to [gist with the tests](https://gist.github.com/ljw1004/61bc96700d0b03c17cf83dbb51437a69). The conclusion is now Tuple keys is fast.
We both think that performance is important here. So maybe it makes sense to spend time testing cases.
Another thought is, in most cases, the user will have one volume. We could use one dictionary until we find a reference to another volume. Although it will complicate the code and it's not clear how much we win. Ooops, for other volume we can make separate recursive `Dir()` call - in the case we have to switch Dictionary (tracker) only on "top" level when we meet new volume. In other words, we will have two dictionaries but they will not be nested and the number of comparisons will be less - we'll be faster.

And maybe rename `Dir()` in something like `DirectoryVisitor()` ?

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Modified the `-FollowSymlink` parameter to set `-Recurse`, and the `-Depth` parameter does.

src/Microsoft.PowerShell.Commands.Management/commands/management/GetChildrenCommand.cs
```diff
@@ -172,6 +172,23 @@ public uint Depth
...
+        [Parameter]
+        public SwitchParameter FollowSymlink
+        {
+            get
+            {
+                return _followSymlink;
+            }
+            set
+            {
+                _followSymlink = value;
```
jeffbi: Modified the `-FollowSymlink` parameter to set `-Recurse`, and the `-Depth` parameter does. Thus we cannot use auto-property here.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: It may be that on Windows the user will have only one volume, but that is much less the case on Unix. For example, on my Linux box the root (`/`) directory has device ID 2049, the `/dev` directory has ID 6, and the `/proc` directory has ID 4. Symlinks can cross device/volume boundaries.

I've changed the code to use the C#7 ValueTuple mechanism.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Refactored code to eliminate `enterDir`.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1788,12 +1865,27 @@ protected override void InvokeDefaultAction(string path)
...
+                                if (tracker.Visited(recursiveDirectory.FullName))
+                                {
+                                    enterDir = false;
+                                    WriteWarning(StringUtil.Format(FileSystemProviderStrings.AlreadyListedDirectory,
+                                                                    recursiveDirectory.FullName));
+                                }
+                            }
+                            // We only want to recurse into symlinks if a) the user has asked to with
+                            // the -FollowSymLinks switch parameter and b) the directory the symlink
+                            // points to has not already been visited, preventing symlink loops.
```
jeffbi: Modified the comment.

src/System.Management.Automation/resources/FileSystemProviderStrings.resx
```diff
@@ -339,4 +339,7 @@
   <data name="SymlinkItemExists" xml:space="preserve">
     <value>Cannot create symbolic link because the path {0} already exists.</value>
   </data>
+  <data name="AlreadyListedDirectory" xml:space="preserve">
+    <value>Directory {0} has already been listed.</value>
```
jeffbi: Re-formatted the message to put the path at the end.

src/Microsoft.PowerShell.Commands.Management/commands/management/GetChildrenCommand.cs
iSazonov: Closed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: > Context.MyInvocation.BoundParameters.ContainsKey("FollowSymlink")) 

Can we use `FollowSymlink` property?

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -8238,6 +8321,47 @@ internal static bool WinIsSameFileSystemItem(string pathOne, string pathTwo)
             return false;
         }
 
+        internal static bool GetInodeData(string path, out UInt64 dev, out UInt64 inode)
```
iSazonov: Now we don't use dev and inode and we could out Tuple.

src/System.Management.Automation/resources/FileSystemProviderStrings.resx
```diff
@@ -339,4 +339,7 @@
   <data name="SymlinkItemExists" xml:space="preserve">
     <value>Cannot create symbolic link because the path {0} already exists.</value>
   </data>
+  <data name="AlreadyListedDirectory" xml:space="preserve">
+    <value>Not listing already-listed directory {0}.</value>
```
iSazonov: Can we exclude repetition  "listing ... listed"?

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: FollowSymlink is a property of `GetChildItemCommand`, not of the `FileSystemProvider`.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: We do still use dev and inode. That's how we uniquely identify a file system object.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: Thanks for clarify!

Closed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: Closed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: Closed.

src/System.Management.Automation/resources/FileSystemProviderStrings.resx
iSazonov: Closed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: I see the `GetInodeData` is only called in `Visit` and `Visited` method.

jeffbi: Yes. Did you want to move the implementation up into the `InodeTracker` class?

iSazonov: I meant simple change like ``` internal static bool GetInodeData(string path, out (dev, inode) var) ```

src/System.Management.Automation/resources/FileSystemProviderStrings.resx
jeffbi: Changed "already-listed" to "already-visited"

iSazonov: Closed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: Closed.

src/Microsoft.PowerShell.Commands.Management/commands/management/GetChildrenCommand.cs
```diff
@@ -172,6 +172,25 @@ public uint Depth
...
+                if (value == true)
+                    Recurse = true;
```
jeffbi: It looks like that was in response to a specific bug. In our case it's just part of a new feature.

iSazonov:
@daxian-dbw Could you please assign a reviewer?

src/Microsoft.PowerShell.Commands.Management/commands/management/GetChildrenCommand.cs
iSazonov: I'm afraid we could catch the same bug here. Maybe maintainers clarify.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -31,6 +31,61 @@
...
+
+        /// <summary>
+        /// Mark a path as visited.
+        /// </summary>
+        /// <param name="path">Path to the file or directory to be marked as visited.</param>
+        /// <returns>True if the path was successfully added to the visited list, false otherwise</returns>
+        public bool Visit(string path)
+        {
+            var inodeData = (0UL, 0UL);
+            if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out inodeData))
```
jeffbi: Thanks for catching that. New line inserted.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -8238,6 +8317,46 @@ internal static bool WinIsSameFileSystemItem(string pathOne, string pathTwo)
...
+            {
+                if (!sf.IsInvalid)
+                {
+                    BY_HANDLE_FILE_INFORMATION info;
+
+                    if (GetFileInformationByHandle(sf.DangerousGetHandle(), out info))
+                    {
+                        UInt64 tmp = info.FileIndexHigh;
+                        tmp <<= 32;
+                        tmp |= info.FileIndexLow;
```
jeffbi: Made the calculation more concise.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: Closed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: Closed.

iSazonov:
@jeffbi Great work! Thanks!
I leave two questions opened for maintainers.

LGTM.

daxian-dbw:
@jeffbi and @iSazonov thanks for the good work and review! I will quickly go through the changes before merging.

src/System.Management.Automation/resources/FileSystemProviderStrings.resx
```diff
@@ -339,4 +339,7 @@
   <data name="SymlinkItemExists" xml:space="preserve">
     <value>Cannot create symbolic link because the path {0} already exists.</value>
   </data>
+  <data name="AlreadyListedDirectory" xml:space="preserve">
+    <value>Not listing already-visited directory {0}.</value>
```
daxian-dbw: I think `Skip already-visited directory {0}` might be better.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -8238,6 +8318,45 @@ internal static bool WinIsSameFileSystemItem(string pathOne, string pathTwo)
             return false;
         }
 
+        internal static bool GetInodeData(string path, out System.ValueTuple<UInt64, UInt64> inodeData)
+        {
```
daxian-dbw: I see that you are following a pattern like `IsSameFileSystemItem` or `IsHardLink`. Ideally, code here should be refactored to consolidate Unix/Windows implementations in one place and eliminate wrapper functions like `IsSameFileSystemItem` and `IsHardLink`. Instead they should be
```CSharp
internal static bool IsSameFileSystemItem(string pathOne, string pathTwo)
{
#if UNIX
   // Unix implementation goes here
#else
   // Windows implementation goes here
#endif
}
```
I'm not asking you to do this change for this PR, just want to call out that it should be this way moving forward. I opened #4086 to track the refactoring task.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -31,6 +31,62 @@
...
+            }
+            return false;
+        }
+
+        /// <summary>
+        /// Mark a path as visited.
+        /// </summary>
+        /// <param name="path">Path to the file or directory to be marked as visited.</param>
+        /// <returns>True if the path was successfully added to the visited list, false otherwise</returns>
+        public bool Visit(string path)
```
daxian-dbw: Correspondingly, maybe this name can be changed to `VisitPath`.

daxian-dbw: And, does this method really need to return anything? If `GetInodeData` failed, then we just don't skip anything.

src/System.Management.Automation/resources/FileSystemProviderStrings.resx
jeffbi: Should this be "Skip"? Or perhaps "Skipping" or "Skipped"?

src/Microsoft.PowerShell.Commands.Management/commands/management/GetChildrenCommand.cs
```diff
@@ -172,6 +172,25 @@ public uint Depth
         }
 
         /// <summary>
+        /// Gets or sets whether symbolic links are to be followed when recursing.
+        /// </summary>
+        [Parameter]
+        public SwitchParameter FollowSymlink
```
jeffbi: So then `-FollowSymlink` should not imply `-Recurse`? It has meaning only if `-Recurse` is also explicitly given?

iSazonov: powercode initiated #4102 for disscusion of parameter sets selection. So it is design case whether or not imply "mandatory" parameter by subparameter.

jeffbi: Event without `-FollowSymlink`, the cmdlet will follow symlinks specified in `-Path`

iSazonov: Should we follow w/o FollowSymlink ?

daxian-dbw: Maybe it will turn out that making `-FollowSymlink` implicitly imply `-Recurse` is better, but for now here are my 2 cents:
1. It's always better to force the user to be explicit about their intent, so `dir -Recurse -FollowSymlink` is more readable than `dir -FollowSymlink` in my opinion.
2. For simplicity reason, this PR can just add the `-FollowSymlink` parameter without worrying about its dependency to `-Recurse`. In my understanding (correct me if I'm wrong) the related code won't execute when it's not recursively iterating file system items. Then we can get feedback from users and base on that to decide if we should make `-FollowSymlink` implicitly imply `-Recurse`.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -31,6 +31,62 @@
...
+    /// Tracks visited files/directories by caching their device IDs and inodes.
+    /// </summary>
+    internal class InodeTracker
+    {
+        private Dictionary<(UInt64, UInt64), bool>  _visitations;
+
+        /// <summary>
+        /// Construct a new InodeTracker
+        /// </summary>
+        public InodeTracker()
```
jeffbi: Fixed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -31,6 +31,62 @@
...
+        {
+            _visitations = new Dictionary<(UInt64, UInt64), bool>();
+        }
+
+        /// <summary>
+        /// Determine if a path has been visited.
+        /// </summary>
+        /// <param name="path">Path to the file or directory to be checked.</param>
+        /// <returns>True if the path has been visited, false otherwise</returns>
+        public bool Visited(string path)
```
jeffbi: Fixed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Fixed both.

src/Microsoft.PowerShell.Commands.Management/commands/management/GetChildrenCommand.cs
jeffbi: Changed to dynamic parameter. `-FollowSymlink` no longer implies `-Recurse`.

src/System.Management.Automation/resources/FileSystemProviderStrings.resx
jeffbi: Fixed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -68,19 +68,13 @@ public bool Visited(string path)
         /// </summary>
         /// <param name="path">Path to the file or directory to be marked as visited.</param>
         /// <returns>True if the path was successfully added to the visited list, false otherwise</returns>
```
iSazonov: Now we return nothing. Please remove the line.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1676,7 +1674,7 @@ protected override void InvokeDefaultAction(string path)
             ReturnContainers returnContainers)
         {
             if (tracker != null)
-                tracker.Visit(directory.FullName);
+                tracker.VisitPath(directory.FullName);
```
iSazonov: Wrong pattern.
We can use `tracker?.VisitPath(directory.FullName); ` or `If (...) { ... }`.

iSazonov: Closed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: Closed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1788,26 +1854,39 @@ protected override void InvokeDefaultAction(string path)
...
+                            }
+                            else
+                            {
+                                if (tracker.IsPathVisited(recursiveDirectory.FullName))
                                 {
-                                    Dir(recursiveDirectory, recurse, depth - 1, nameOnly, returnContainers);
+                                    WriteWarning(StringUtil.Format(FileSystemProviderStrings.AlreadyListedDirectory,
+                                                                recursiveDirectory.FullName));
+                                    continue;
                                 }
```
iSazonov: Sorry, I again under 'tracker == null' doulble check pressure.
It seem we can remove one if remove lines 1675-1678, add its before first `Dir` call and here add:
```c#
                                else
                                {
                                    tracker.VisitPath(directory.FullName);
                                }
```

Next, we can exclude `VisitPath` at all if make "visit" in `IsPathVisited` (we should rename it maybe as `TryPathVisit`).

jeffbi: I don't agree with this. If we mark the directory has having been visited here, then if the test at line 1883 evaluates to false the `Dir` method will not be called and we will not visit the directory even though we have already claimed to have done so.

iSazonov: Do you mean "hidden" directories? If so we can safely mark "hidden" directories as "visited" - in both cases they is really skipped.

jeffbi: A directory might be hidden. A symbolic link (or several such) might not be.

iSazonov: You are right. We can move the hidden check above line 1861 - it is less expensive then double call 'GetInodeData'.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -40,41 +40,49 @@ internal class InodeTracker
...
+        /// </remarks>
+        internal bool TryVisitPath(string path)
         {
+            bool rv = true;
             var inodeData = (0UL, 0UL);
 
             if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out inodeData))
             {
-                _visitations[inodeData] = true;
+                if (_visitations.ContainsKey(inodeData))
```
iSazonov: Now CoreFX has a new method [Dictionary.TryAdd](https://github.com/dotnet/corefx/blob/3d4d8038d1a3f0c56d3046bc7494f30ecc35e67b/src/Common/tests/System/Collections/DictionaryExtensions.cs) - Could we use them?

jeffbi: Used new method

iSazonov:
LGTM.

@daxian-dbw Could you please continue the code review?

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -31,6 +31,56 @@
...
+        /// <param name="path">Path to the file or directory to be marked as visited.</param>
+        /// <returns>
+        /// False if the path has already been visited.
+        /// True if the path has not yet been visited.
+        /// </returns>
+        /// <remarks>
+        /// If the path has not yet been visited then before this method returns it marks
+        /// the path as having been visited.
+        /// </remarks>
+        internal bool TryVisitPath(string path)
```
daxian-dbw: This method goes against the convention of `Try<Verb>` pattern. Take `TryAdd` as an instance, if added, then return true; if not added, then return false. Following that conversion, `TryVisitPath` should be: if visited, then return true; if not visited, then return false. So `TryVisitPath` makes it confusing understanding the code.

How about making it `IsPathVisited(bool visitIfNotAlready = false)`?
```CSharp
internal bool IsPathVisited(bool visitIfNotAlready = false)
{
    bool isPathVisited = false;
    if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out (UInt64, UInt64) inodeData))
    {
        if (visitIfNotAlready)
        {
            isPathVisited = !_visitations.Add(inodeData)
        }
        else
        {
            isPathVisited = _visitations.Contains(inodeData)
        }
    }
    return isPathVisited;
}
```

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -31,6 +31,56 @@
 
 namespace Microsoft.PowerShell.Commands
 {
+    #region InodeTracker
+    /// <summary>
+    /// Tracks visited files/directories by caching their device IDs and inodes.
+    /// </summary>
+    internal class InodeTracker
```
daxian-dbw: This class seems to be a helper that will only be used in FileSystemProvider code, so can we make it a private nested class within FileSystemProvider? (just like `private static class NativeMethods` in FileSystemProvider)

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -31,6 +31,56 @@
...
+
+        /// <summary>
+        /// Construct a new InodeTracker with an initial path
+        /// </summary>
+        internal InodeTracker(string path)
+        {
+            _visitations = new Dictionary<(UInt64, UInt64), bool>();
+            var inodeData = (0UL, 0UL);
+
+            if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out inodeData))
```
daxian-dbw: Minor comment: maybe use the following pattern to avoid `var inodeData = (0UL, 0UL);`?
```CSharp
if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out (UInt64, UInt64) inodeData))
```

daxian-dbw (commented):
Thanks for switching to the dynamic parameter approach! See some more comments below.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -31,6 +31,56 @@
 
 namespace Microsoft.PowerShell.Commands
 {
+    #region InodeTracker
+    /// <summary>
+    /// Tracks visited files/directories by caching their device IDs and inodes.
+    /// </summary>
+    internal class InodeTracker
+    {
+        private Dictionary<(UInt64, UInt64), bool>  _visitations;
```
iSazonov: We use new Dictionary.TryAdd() but HashSet.Add() works the same - so we can use HashSet.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: Looks good.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: With HashSet it will looks just better.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Fixed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: HashSet is much better. Fixed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Fixed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
jeffbi: Fixed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1539,8 +1589,16 @@ protected override void InvokeDefaultAction(string path)
                 if (isDirectory)
                 {
                     DirectoryInfo directory = new DirectoryInfo(path);
+                    InodeTracker tracker = null;
+
+                    GetChildDynamicParameters fspDynamicParam = DynamicParameters as GetChildDynamicParameters;
```
jeffbi: I thought I had a check for recurse at one time. It must have gotten lost in the shuffle. Fixed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1608,6 +1666,7 @@ protected override void InvokeDefaultAction(string path)
             DirectoryInfo directory,
             bool recurse,
             uint depth,
+            InodeTracker tracker,   // This will be non-null only if the user invoked the -FollowSymLinks switch parameter.
```
jeffbi: Fixed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1788,25 +1847,38 @@ protected override void InvokeDefaultAction(string path)
...
+                                    }
                                 }
-
-                                // if "Hidden" is explicitly specified anywhere in the attribute filter, then override
-                                // default hidden attribute filter.
-                                if (Force || !hidden || isFilterHiddenSpecified || isSwitchFilterHiddenSpecified)
+                                else
                                 {
-                                    Dir(recursiveDirectory, recurse, depth - 1, nameOnly, returnContainers);
+                                    if (!tracker.TryVisitPath(recursiveDirectory.FullName))
```
jeffbi: Fixed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -7483,6 +7555,12 @@ internal sealed class GetChildDynamicParameters
         public FlagsExpression<FileAttributes> Attributes { get; set; }
 
         /// <summary>
+        /// Gets or sets the flag to follow symbolic links when recursing.
+        ///
```
jeffbi: Fixed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -7320,6 +7342,69 @@ private struct NetResource
...
+            /// <summary>
+            /// Determine if a path has been visited. Optionally mark the path as
+            /// having been visited if it has not already
+            /// </summary>
+            /// <param name="path">
+            /// Path to the file system item to be checked.
+            /// </param>
+            /// <param name="visitIfNotAlready">
+            /// Flag to indicate whether the path should be marked as visited if
+            /// it has not already been visited.
```
iSazonov: Should we add the parameter if we don't use them?

jeffbi: Normally I would say no. In this case the existence of the parameter and its name give a clue about the side effect of marking the path as visited.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1609,7 +1620,8 @@ protected override void InvokeDefaultAction(string path)
             bool recurse,
             uint depth,
             bool nameOnly,
-            ReturnContainers returnContainers)
+            ReturnContainers returnContainers,
+            InodeTracker tracker)   // tracker will be non-null only if the user invoked the -FollowSymLinks switch parameter.
```
daxian-dbw: Minor comment: this comment is not accurate now -- `tracker` will be null when `-FollowSymlinks` is specified but `-Recurse` is not.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
daxian-dbw: I give this some more thinking, and believes that the `TryVisitPath` signature is more natural in this case. But we need to follow the `Try<>` pattern.

Again, take `HashSet.Add(object)` as an example:
1. If the object is not in the set, add it and return true.
2. If the object is already in the set, return false which means it was not added in this call because it's already there.

Similarly, `InodeTracker.TryVisitPath(path)` should be
1. If the path is not visited yet, visit it and return true.
2. If the path is already visited, return false which means it was not visited in this call because it's already visited.

The code would be like:
```CSharp
internal bool TryVisitPath(string path)
{
    bool returnValue = false;
    if (InternalSymbolicLinkLinkCodeMethods.GetInodeData(path, out (UInt64, UInt64) inodeData))
    {
        returnValue = _visitations.Add(inodeData);
    }
    return returnValue;
}
```

@jeffbi @iSazonov what do you think?

daxian-dbw (commented):
2 more comments

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: I like the TryVisitPath more the IsVisitPath.

jeffbi: Works for me. I'll make the change.

daxian-dbw: Awesome. Thank you both!

daxian-dbw (approved):
@jeffbi and @iSazonov Thank you both for the great work and thorough review!

iSazonov:
@jeffbi Many thanks for the great work!

jeffbi:
@iSazonov and @daxian-dbw Thanks for the review.

sba923:
Very good job! 

P.S. I would probably have called the new parameter `-FollowSymlinks`

PR #3800, opened 2017-05-17
Update format-hex tests to include -TestCase parameter

MiaRomero:
Addressing issue #3382 'Use -TestCases parameter in FormatHex.Tests.ps1'.

* Replaced the 'foreach' with the -TestCases parameter in the test helper functions.

PR #4021, opened 2017-06-15
Fixed test to not create a file called `1` when run

SteveL-MSFT:
When using `Start-PSPester` a file called `1` was being created.  This is due to incorrect usage of a redirection operator `>` when it should have been the comparison operator `-gt`.  After fixing this, a new
code path started executing which had incorrect assumption as the Format-Hex cmdlet works with
16-bytes at a time, so fixed that logic.


MiaRomero:
I fixed this same thing in PR #3800 but it doesn't look like it was merged.

SteveL-MSFT:
@MiaRomero didn't see your PR, closing this one

Issue #1395, opened 2016-07-16
$PSVersionTable.CLRVersion property should be removed in PowerShell Core


alexandair:
## Steps to reproduce

I'm on Windows 10 with November update:

``` powershell
PS C:\> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      5.0.10586.122
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.10586.122
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

When I install PowerShell 0.6.0 by using powershell-0.6.0-win81-x64.zip or PowerShell_0.6.0.0.msi, output of $PSVersionTable is not correct. 
## Expected behavior

PSEdition property should be Desktop, not Core.
CLRVersion property should not be empty.
## Actual behavior

``` powershell
PS C:\Program Files\PowerShell\0.6.0.0> $PSVersionTable

Name                           Value
----                           -----
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
PSVersion                      5.1.10032.0
PSRemotingProtocolVersion      2.3
PSEdition                      Core
GitCommitId                    v0.6.0
CLRVersion
BuildVersion                   3.0.0.0
```

andschwa:
This is purposefully not set because it is supposed to be `[Environment]::Version`, but that is not available in .NET Core (at least not in the current stable release; it *is coming back* per https://github.com/dotnet/corefx/commit/fe5f4fbebc9e85c8e512a273567719dca15ac764, which will be available in the next stable release). So while it is not available, it is set to null.

Please note that the following will be confusing:

```powershell
~/src/PowerShell |-/ [environment]::Version 

Major  Minor  Build  Revision
-----  -----  -----  --------
4      0      30319  42000   
```

This is due to `Version` being set in the environment extensions class for compatibility, and is not the actual CLR version.

So `$PSVersionTable` is currently correct; but should be updated with the next release of .NET Core.

Closing for now.

/cc @daxian-dbw

lzybkr:
If the api is coming back, this issue should be left open so we remember to fix the code.

If the api never comes back, we should consider either removing the entry or fill it in explicitly. An empty value is useless.

joeyaiello:
[...]

PR #4027, opened 2017-06-15
Remove ClrVersion property from $PSVersionTable

SteveL-MSFT:
The ClrVersion property of $PSVersionTable is not useful with CoreCLR and end users should not be using
that value to determine compatibility.  Recommendation from dotnet team is to remove that property.

Fix https://github.com/PowerShell/PowerShell/issues/1395

### DIFF
--- a/src/System.Management.Automation/engine/PSVersionInfo.cs
+++ b/src/System.Management.Automation/engine/PSVersionInfo.cs
@@ -72,7 +72,6 @@ namespace System.Management.Automation
 #if CORECLR
             s_psVersionTable[PSOSName] = Runtime.InteropServices.RuntimeInformation.OSDescription.ToString();
 #else
-            s_psVersionTable[PSCLRVersionName] = Environment.Version;
             s_psVersionTable[PSOSName] = Environment.OSVersion.ToString();
 #endif
         }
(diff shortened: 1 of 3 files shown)


iSazonov (commented):
I believe we should remove [internal static Version CLRVersion](https://github.com/SteveL-MSFT/PowerShell/blob/645c87554d3a4e2797c02614c982e1a50ba8f1de/src/System.Management.Automation/engine/PSVersionInfo.cs#L164) and [internal const string PSCLRVersionName](https://github.com/SteveL-MSFT/PowerShell/blob/645c87554d3a4e2797c02614c982e1a50ba8f1de/src/System.Management.Automation/engine/PSVersionInfo.cs#L25)

SteveL-MSFT:
@iSazonov will remove those

iSazonov (approved):

iSazonov:
LGTM.

SteveL-MSFT:
@mirichmo can you merge?

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -162,14 +160,6 @@ internal static string GitCommitId
             }
         }
 
-        internal static Version CLRVersion
```
mirichmo: What is our stance on removing FullCLR compatibility of the code? If we are dropping that requirement and only focusing on CoreCLR, then you should remove PSCLRVersionName from line 77 as well.

SteveL-MSFT: Removed it since we removed the other members anyways.  After beta.3 goes out, I intend to revisit the discussion about removing FullCLR code.

mirichmo (approved):

PR #4039, opened 2017-06-17
Start-Sleep: add 'ms' alias

### LINKED ISSUE #3991: Suggestion: Alias Start-Sleep's -Milliseconds parameter to -ms
"ms" is a standard abbreviation for "milliseconds", so it make sense to allow `-ms` as an alias for `-MilliSeconds`.

```powershell
Start-Sleep -ms 100  # should be short for Start-Sleep -Milliseconds 100
```

```powershell
PowerShell Core v6.0.0-beta.2
```

Tadas:
Fixes #3991

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/Start-Sleep.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/Start-Sleep.Tests.ps1
@@ -9,6 +9,7 @@ Describe "Start-Sleep DRT Unit Tests" -Tags "CI" {
         Start-Sleep -Seconds 1
         $watch.Stop()
         $watch.ElapsedMilliseconds | Should BeGreaterThan 950
+        $watch.ElapsedMilliseconds | Should BeLessThan 1050
     }
 
     It "Should work properly when sleeping with Milliseconds" {
@@ -16,6 +17,7 @@ Describe "Start-Sleep DRT Unit Tests" -Tags "CI" {
         Start-Sleep -Milliseconds 1000
         $watch.Stop()
         $watch.ElapsedMilliseconds | Should BeGreaterThan 950
+        $watch.ElapsedMilliseconds | Should BeLessThan 1050
     }
 
     It "Should work properly when sleeping with ms alias" {
@@ -23,6 +25,7 @@ Describe "Start-Sleep DRT Unit Tests" -Tags "CI" {
         Start-Sleep -ms 1000
         $watch.Stop()
         $watch.ElapsedMilliseconds | Should BeGreaterThan 950
+        $watch.ElapsedMilliseconds | Should BeLessThan 1050
     }
 }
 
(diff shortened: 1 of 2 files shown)


test/powershell/Modules/Microsoft.PowerShell.Utility/Start-Sleep.Tests.ps1
```diff
@@ -17,6 +17,13 @@ Describe "Start-Sleep DRT Unit Tests" -Tags "CI" {
         $watch.Stop()
         $watch.ElapsedMilliseconds | Should BeGreaterThan 950
     }
+
+    It "Should work properly when sleeping with ms alias" {
+        $watch = [System.Diagnostics.Stopwatch]::StartNew()
+        Start-Sleep -ms 1000
+        $watch.Stop()
+        $watch.ElapsedMilliseconds | Should BeGreaterThan 950
```
SteveL-MSFT: Seems like you should also make sure it's less than 1050

SteveL-MSFT (changes requested):

Tadas:
What about the `Describe "Start-Sleep" -Tags "CI" {` test, isn't it redundant?

test/powershell/Modules/Microsoft.PowerShell.Utility/Start-Sleep.Tests.ps1
iSazonov: Should we test parameter aliases? It seems it is a general feature and we should test the feature in special test but not every cmdlet and every alias.

SteveL-MSFT: In general, we should validate the alias is correct, but shouldn't repeat every test for the parameter and its alias.  `-ms` is a pretty simple case, so it's hard to not repeat the test (using -TestCases is probably better here).  For parameters with more complex behavior, I think it would be sufficient if some cases used the parameter while others used the alias.

lzybkr: @iSazonov - As a matter of principle - we avoid testing the parameter binder in cmdlet specific tests.

Personally I don't ask for any tests for an alias - the test isn't covering any new code, but I understand the concerns of those asking for the test, so I don't complain about it either.

SteveL-MSFT (approved):

test/powershell/Modules/Microsoft.PowerShell.Utility/Start-Sleep.Tests.ps1
iSazonov: If we're worried about parameter's aliases, we could make a common test for all of them, just like we did for the cmdlets and their aliases as < cmdlet, parameter, alias, platform(?) >.

lzybkr: It's not a big problem, people mostly never remove code anyway and if they did, it should be caught during the PR, but it's a nice idea.

iSazonov:
@Tadas Thanks for your contribution!

PR #4044, opened 2017-06-18
Cmdlets: Set -Subject parameter to non-mandatory on Send-MailMessage command

### LINKED ISSUE #4043: Send-MailMessage $Subject Mandatory=$true
Not a bug, but a request for "Send-MailMessage" cmdlet

```powershell
 $From = $email
    $To = $Number
    $Body = "Something"
    $Subject = $null
    $SMTPServer = "smtp.gmail.com"
    $SMTPPort = "587"
    Send-MailMessage -From $From -to $To -Subject $Subject `
        -Body $Body -SmtpServer $SMTPServer -port $SMTPPort -UseSsl `
    -Credential $Cred
```

RESULT:
```
Send-MailMessage : Cannot validate argument on parameter 'Subject'. The argument is null or empty. Provide an argument that is not null or empty, and then try the command again.
At line:8 char:52
+       -Body $Body -SmtpServer $SMTPServer -Subject $Subject -port $SM ...
+                                                    ~~~~~~~~
    + CategoryInfo          : InvalidData: (:) [Send-MailMessage], ParameterBindingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationError,Microsoft.PowerShell.Commands.SendMailMessage
```

CAUSE:
```
[Parameter(Mandatory=$true, Position=1)]
        [Alias('sub')]
        [ValidateNotNullOrEmpty()]
        [string]
        ${Subject},
```

--------
Emails do not require subjects to be present.  However, in the Send-MailMessage cmdlet, the $subject parameter is set to "Mandatory=$true".  Since an email does not require a subject, this should really be set to "$false". 

I've created my own workaround this, but I feel that this should be changed since a subject is not imperative for an email to be sent.

pcgeek86:
Fixes #4043 

Cheers,
Trevor Sullivan

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/Send-MailMessage.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/Send-MailMessage.cs
@@ -191,7 +191,7 @@ namespace Microsoft.PowerShell.Commands
         /// <summary>
         /// Specifies the  subject of the email message.
         /// </summary>
-        [Parameter(Mandatory = true, Position = 1)]
+        [Parameter(Mandatory = false, Position = 1)]
         [Alias("sub")]
         [ValidateNotNullOrEmpty]
         public String Subject


pcgeek86:
FYI @joeyaiello @stevel-msft

SteveL-MSFT:
Should it output a warning if no subject?

pcgeek86:
Hmm, good thought but my vote would be no.

BitteringAgent:
Now that I think about it.  It should work like any other email application where there is a warning for no subject.  So there should be a warning, but also there should be a -Force to push past the warning of having no subject.

SteveL-MSFT:
@BitteringAgent `-force` wouldn't be needed as a warning is just informational

BitteringAgent:
Ah, I haven't dealt with too many cmdlets with warnings other than maybe doing a remote session into Exchange Online.  Thanks.

iSazonov:
Missing Subject is a very bad practice. By default, we must ask for it to present. Maybe add `-IgnoreSubject`?

SteveL-MSFT:
Another option is to prompt and allow `-force`

TravisEz13:
restarted mac tests due to random failure.

TravisEz13:
I agree with @iSazonov , but prefer @SteveL-MSFT 's proposal to fix the issue.

iSazonov:
@pcgeek86 Could you please continue?

dantraMSFT:
I think a better user experience would be to leave Subject as mandatory and place it in a 'Subject' parameter set.  Add a -NoSubject switch and place it in a 'NoSubject' parameter set, then set the default parameter set to 'Subject'.
This ensures the best practice is the default, avoids having to use an ambiguous -Force switch while still requiring an explicit choice to send without an subject.

TravisEz13:
@pcgeek86 comments haven't been addressed in a long time.  At some point, if comments are not addressed, I'll close the PR.

PR #4045, opened 2017-06-19
Check a dotnet version in Start-PSBuild

### LINKED ISSUE #4018: Add dotnet version check to build.psm1 module
Currently it is not obvious when PowerShell takes a dependency on a new version of dotnet, which can cause the build to fail.  It would be great if Start-PSBuild would check dotnet existing and required versions and provide a notification when the existing version is out of date.  The notification can tell the user how to update:
a) Remove the old dotnet version
b) Run Install-Dotnet or Start-PSBootstrap

iSazonov:
Close #4018

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -19,7 +19,8 @@ try {
     catch { }
 }
 
-$dotnetRequiredVersion = "2.0.0-preview1-005952"
+$dotnetCLIChannel = "preview"
+$dotnetCLIRequiredVersion = "2.0.0-preview1-005952"
 
 # On Unix paths is separated by colon
 # On Windows paths is separated by semicolon
@@ -191,8 +192,22 @@ function Start-PSBuild {
     $precheck = precheck 'dotnet' "Build dependency 'dotnet' not found in PATH. Run Start-PSBootstrap. Also see: https://dotnet.github.io/getting-started/"
 
     $dotnetIntalledVersion = (dotnet --version)
-    If ( $dotnetIntalledVersion -ne $dotnetRequiredVersion ) {
-        Write-Warning ".Net Core version mismatch: installed version is $dotnetIntalledVersion, required version is $dotnetRequiredVersion.`nRun Start-PSBootstrap and Start-PSBuild -Clean -Restore.`nAlso see: https://dotnet.github.io/getting-started/"
+    If ( $dotnetIntalledVersion -ne $dotnetCLIRequiredVersion ) {
+        Write-Warning @"
+The currently installed .Net Core version is not the required version.
+
+Installed version: $dotnetIntalledVersion
+Required version: $dotnetCLIRequiredVersion
+
+Fix steps:
+
+1. Remove the installed version from:
+    - on windows '`$env:LOCALAPPDATA\Microsoft\dotnet'
+    - on osx and linux '`$env:HOME/.dotnet'
+2. Run Start-PSBootstrap or Install-Dotnet
+3. Start-PSBuild -Clean -Restore
+`n
+"@
         return
     }
 
@@ -995,8 +1010,8 @@ function Start-PSxUnit {
 function Install-Dotnet {
     [CmdletBinding()]
     param(
-        [string]$Channel = "preview",
-        [string]$Version = $dotnetRequiredVersion,
+        [string]$Channel = $dotnetCLIChannel,
+        [string]$Version = $dotnetCLIRequiredVersion,
         [switch]$NoSudo
     )
 
@@ -1055,10 +1070,10 @@ function Start-PSBootstrap {
         SupportsShouldProcess=$true,
         ConfirmImpact="High")]
     param(
-        [string]$Channel = "preview",
+        [string]$Channel = $dotnetCLIChannel,
         # we currently pin dotnet-cli version, and will
         # update it when more stable version comes out.
-        [string]$Version = $dotnetRequiredVersion,
+        [string]$Version = $dotnetCLIRequiredVersion,
         [switch]$Package,
         [switch]$NoSudo,
         [switch]$Force
--- a/build.psm1
+++ b/build.psm1
@@ -19,6 +19,8 @@ try {
     catch { }
 }
 
+$dotnetRequiredVersion = "2.0.0-preview1-005952"
+
 # On Unix paths is separated by colon
 # On Windows paths is separated by semicolon
 $TestModulePathSeparator = ':'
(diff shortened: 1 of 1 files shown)


build.psm1
```diff
@@ -188,6 +190,12 @@ function Start-PSBuild {
     # verify we have all tools in place to do the build
     $precheck = precheck 'dotnet' "Build dependency 'dotnet' not found in PATH. Run Start-PSBootstrap. Also see: https://dotnet.github.io/getting-started/"
 
+    $dotnetIntalledVersion = (dotnet --version)
+    If ( $dotnetIntalledVersion -ne $dotnetRequiredVersion ) {
+        Write-Warning ".Net Core version mismatch: installed version is $dotnetIntalledVersion, required version is $dotnetRequiredVersion.`nRun Start-PSBootstrap and Start-PSBuild -Clean -Restore.`nAlso see: https://dotnet.github.io/getting-started/"
```
PaulHigin: I usually have to manually remove dotnet, so I think we should mention that manually removing the existing version of dotnet may be necessary.

PaulHigin (changes requested):

build.psm1
iSazonov: How the message should sound? I believe short step-by-step will be useful.

lzybkr: The message should also print out the path to `dotnet` - a user might have multiple versions installed and is picking up the wrong one.

PaulHigin: How about?
```
WARNING: The currently installed .Net Core version is not the required version.
Installed version: 2.0.0-preview1-beta
Installed path:  /home/user/dotnet
Required version: 2.0.0-preview2-006388
1. Manually remove the old version from this location: /home/user/dotnet
2. Run Start-PSBootstrap and Start-PSBuild -Clean -Restore.
   Also see: https://dotnet.github.io.getting-started/
```

build.psm1
```diff
@@ -19,6 +19,8 @@ try {
     catch { }
 }
 
+$dotnetRequiredVersion = "2.0.0-preview1-005952"
```
daxian-dbw: It's the version of the `.NET Command Line Tools`, not the .NET Core version. Could you please update the name to be `dotnetCLIRequiredVersion`?

build.psm1
```diff
@@ -188,6 +190,12 @@ function Start-PSBuild {
     # verify we have all tools in place to do the build
     $precheck = precheck 'dotnet' "Build dependency 'dotnet' not found in PATH. Run Start-PSBootstrap. Also see: https://dotnet.github.io/getting-started/"
 
+    $dotnetIntalledVersion = (dotnet --version)
```
daxian-dbw: same here, should be `$dotnetCLIIntalledVersion`

build.psm1
daxian-dbw: `Install-Dotnet` installs dotnet-cli to the default location --
- on windows, it's `$env:LOCALAPPDATA\Microsoft\dotnet`
- on osx and linux, it's `"$env:HOME/.dotnet"`

How about we remove the folder from the default location in `Install-Dotnet` before installing the new one?

> https://dotnet.github.io/getting-started/

I think we can skip this link, it's not very useful as we are using `2.0.0-preview`. And also, running `Install-Dotnet` will suffice. You don't need to run `Start-PSBootstrap`.

PaulHigin: Do we own Install-Dotnet?  Maybe we should have Install-DotNet -Force to automatically remove any existing?

build.psm1
iSazonov: Done.

build.psm1
iSazonov: Done.

build.psm1
iSazonov: I re-format the message, add `$dotnetCLIChannel` and now waiting conclusion about auto removing old dotnet version. I don't mind.

iSazonov:
@SteveL-MSFT CI OSX failed in Start-Sleep.Tests.ps1 - should we fix the test?
```
Description: Should work properly when sleeping with Milliseconds
Name:        Start-Sleep DRT Unit Tests.Should work properly when sleeping with Milliseconds
message:
Expected {1051} to be less than {1050}
stack-trace:
at line: 20 in /Users/travis/build/PowerShell/PowerShell/test/powershell/Modules/Microsoft.PowerShell.Utility/Start-Sleep.Tests.ps1
$watch.ElapsedMilliseconds | Should BeLessThan 1050
```

PaulHigin (approved):
LGTM

SteveL-MSFT:
@iSazonov already have a PR for start-sleep tests https://github.com/PowerShell/PowerShell/pull/4049

build.psm1
daxian-dbw: > now waiting conclusion about auto removing old dotnet version

That can be done with a separate PR. Actually removing the existing `.dotnet` folder from the default location might not be necessary because it seems `dotnet-install.ps1` handles it properly.

build.psm1
daxian-dbw: this comment is not addressed.

build.psm1
```diff
@@ -188,6 +191,26 @@ function Start-PSBuild {
...
+    If ( $dotnetIntalledVersion -ne $dotnetCLIRequiredVersion ) {
+        Write-Warning @"
+The currently installed .Net Core version is not the required version.
```
daxian-dbw: > The currently installed .Net Core version

Should be `The currently installed .NET Command Line Tools is not the required version.`

build.psm1
```diff
@@ -188,6 +191,26 @@ function Start-PSBuild {
...
+Installed version: $dotnetIntalledVersion
+Required version: $dotnetCLIRequiredVersion
+
+Fix steps:
+
+1. Remove the installed version from:
+    - on windows '`$env:LOCALAPPDATA\Microsoft\dotnet'
+    - on osx and linux '`$env:HOME/.dotnet'
+2. Run Start-PSBootstrap or Install-Dotnet
+3. Start-PSBuild -Clean -Restore
```
daxian-dbw: `Start-PSBuild -Clean` will be sufficient.

iSazonov: Fixed.

build.psm1
iSazonov: Fixed.

build.psm1
iSazonov: Sorry, lost a commit.

Fixed.

daxian-dbw (approved):

iSazonov:
Should we repair .nuget\packages folder in an user profile folder?

daxian-dbw:
@iSazonov I'm not sure I understand what you are suggesting. Can you please elaborate?

iSazonov:
In my  .nuget\packages folder I see packages of multiple versions. I'm afraid that this can lead to side effects, especially when we are restoring a previous dotnet version.

daxian-dbw:
It's fine. dotnet-cli is designed to work side by side, and it's totally legitimate to have different projects built against different versions of .NET Core.

PR #4049, opened 2017-06-19
Travis-CI nightly build failing due to start-sleep tests

SteveL-MSFT:
Tests were failing in CI nightly build as the timing was off past 1050ms (was actually 1053ms).  Adding a slightly bigger window.


daxian-dbw (approved):

SteveL-MSFT:
@daxian-dbw can we merge this as it's affecting several of the CI runs

Issue #3014, opened 2017-01-18
Move Windows Native Components to a Sub-project or Separate Project


mirichmo:
The WinRM plugin for PowerShell introduces the Visual Studio and Windows 10 SDK dependencies. These are heavyweight components that slow down the build time. Moving the native Windows components to a separate project or sub-project would decouple this dependency and improve build times, CI passes, etc.

iSazonov:
What is Sub-project or Separate Project? New GitHub Repo? If so only maintainers can do this.

/cc @SteveL-MSFT

SteveL-MSFT:
Current plan is to just pull the native binaries from the last official release at build time.  Since the native code hasn't changed much at all, no need for a separate repo for now.

SteveL-MSFT:
Feedback from @daxian-dbw that makes sense to me:

- I think we should by default continue to build native component for OSX and Linux. Unlike pwrshplugin.dll, libpsl is much more crucial to PSCore (it has important native implementations that PS engine needs to call in via P/inovke), and it changes far more frequently than the WinRM plugin dll -- for example, @jeffbi made changes to it to fix the cyclic symbolic link detection issue in file system provider; @PaulHigin made changes to it to create the SSH process for remoting.
- For pwrshplugin.dll, pwrshplugin.pdb and Install-PowerShellRemoting.ps1, they should be made into a NuGet package and published to powershell-core feed. So we can depend on dotnet-cli to handle them and it's also more robust and cleaner. When we need to make changes to pwrshplugin.dll, we just publish a new version of the NuGet package.

iSazonov:
Native components don't seem so slow to build, I don't have a problem with that. Also we have some opened Issues for remoting and console and  a Issue for refactoring native components.

daxian-dbw:
@iSazonov yes, the building itself doesn't take much time, the problem here is `Start-PSBootstrap` for Windows -- building `pwrshplugin.dll` requires to install win10 SDK and Visual Studio, and that's too heavy. If we can move WinRM remoting components to NuGet package, then a regular user don't need to install them on bootstrap anymore.

iSazonov:
Locally we don't install SDK every time. For CI we could use a build cache.

PR #4051, opened 2017-06-19
Updated Start-PSBuild to not require building native binaries by default

SteveL-MSFT:
With this change, `Start-PSBuild` only builds the managed code.  Since the native code rarely changes, the script picks it up from the currently running PowerShell.  Expectation is that the developer is running latest PowerShell.  If there is a need to build the native binaries, user will run `Start-PSBootStrap -BuildNative` at least once to ensure the native build tools (and Win10 SDK on Windows) gets installed.  Then `Start-PSBuild -BuildNative` is run to explicitly build the native sources.

Fix https://github.com/PowerShell/PowerShell/issues/3014

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -92,6 +92,7 @@ function Start-PSBuild {
         [switch]$TypeGen,
         [switch]$Clean,
         [switch]$PSModuleRestore,
+        [switch]$BuildNative,
 
         # this switch will re-build only System.Management.Automation.dll
         # it's useful for development, to do a quick changes in the engine
@@ -188,7 +189,7 @@ function Start-PSBuild {
     # verify we have all tools in place to do the build
     $precheck = precheck 'dotnet' "Build dependency 'dotnet' not found in PATH. Run Start-PSBootstrap. Also see: https://dotnet.github.io/getting-started/"
 
-    if ($IsWindows) {
+    if ($IsWindows -and $BuildNative) {
         # cmake is needed to build powershell.exe
         $precheck = $precheck -and (precheck 'cmake' 'cmake not found. Run "Start-PSBootstrap -BuildNative". You can also install it from https://chocolatey.org/packages/cmake')
 
@@ -211,9 +212,9 @@ function Start-PSBuild {
             $msbuildConfiguration = 'Release'
         }
 
-    } elseif ($IsLinux -or $IsOSX) {
+    } elseif (($IsLinux -or $IsOSX) -and $BuildNative) {
         foreach ($Dependency in 'cmake', 'make', 'g++') {
-            $precheck = $precheck -and (precheck $Dependency "Build dependency '$Dependency' not found. Run 'Start-PSBootstrap -BuildNative'.")
+            $precheck = $precheck -and (precheck $Dependency "Build dependency '$Dependency' not found. Run 'Start-PSBootstrap -BuildNative'")
         }
     }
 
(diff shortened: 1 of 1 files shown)


iSazonov:
Could we Test-Path native binaries and force build if its absent?

SteveL-MSFT:
@iSazonov it's not only building, but also bootstrapping all the build dependencies.  I think in general, the engineering team will know if the native binaries need to be rebuilt and will explicitly do so for a release, otherwise I expect we won't be making many changes to the native code so it won't need to be built much

build.psm1
```diff
@@ -286,122 +287,191 @@ function Start-PSBuild {
...
-                Get-ChildItem $nativeResourcesFolder -Filter "*.mc" | % {
-                    $command = @"
-cmd.exe /C cd /d "$currentLocation" "&" "$($vcVarsPath)\vcvarsall.bat" "$NativeHostArch" "&" mc.exe -o -d -c -U "$($_.FullName)" -h "$nativeResourcesFolder" -r "$nativeResourcesFolder"
+                # Compile native resources
+                $currentLocation = Get-Location
+                @("nativemsh/pwrshplugin") | % {
+                    $nativeResourcesFolder = $_
+                    Get-ChildItem $nativeResourcesFolder -Filter "*.mc" | % {
+                        $command = @"
+    cmd.exe /C cd /d "$currentLocation" "&" "$($vcVarsPath)\vcvarsall.bat" "$NativeHostArch" "&" mc.exe -o -d -c -U "$($_.FullName)" -h "$nativeResourcesFolder" -r "$nativeResourcesFolder"
```
andschwa: I don't think this was meant to be indented. I doubt it'll break anything, but it looks accidental.

build.psm1
```diff
@@ -286,122 +287,191 @@ function Start-PSBuild {
...
-                $location = Get-Location
+    # Disabling until I figure out if it is necessary
+    #                $overrideFlags = "-DCMAKE_USER_MAKE_RULES_OVERRIDE=$PSScriptRoot\src\powershell-native\windows-compiler-override.txt"
+                    $overrideFlags = ""
+                    $location = Get-Location
 
-                $command = @"
-cmd.exe /C cd /d "$location" "&" "$($vcVarsPath)\vcvarsall.bat" "$NativeHostArch" "&" cmake "$overrideFlags" -DBUILD_ONECORE=$OneCoreValue -DBUILD_TARGET_ARCH=$NativeHostArch -G "$cmakeGenerator" . "&" msbuild ALL_BUILD.vcxproj "/p:Configuration=$msbuildConfiguration"
+                    $command = @"
+    cmd.exe /C cd /d "$location" "&" "$($vcVarsPath)\vcvarsall.bat" "$NativeHostArch" "&" cmake "$overrideFlags" -DBUILD_ONECORE=$OneCoreValue -DBUILD_TARGET_ARCH=$NativeHostArch -G "$cmakeGenerator" . "&" msbuild ALL_BUILD.vcxproj "/p:Configuration=$msbuildConfiguration"
```
andschwa: Ditto.

build.psm1
```diff
@@ -286,122 +287,191 @@ function Start-PSBuild {
...
+    }
+    else
+    {
+        # use native binaries from $PSHOME, assume it is latest
+        $nativeBinaries = @()
+        if ($IsWindows)
+        {
+            if ($PSVersionTable.PSEdition -ne "Core")
+            {
+                # if running under Windows PowerShell, we'll need to get the binaries from GitHub release
```
andschwa: At this point, should we maybe just build the native binaries rather than download them? I get for the rest of the scenarios this PR makes us use existing native binaries on the machine, but here, in Windows PowerShell, why download a PowerShell package over building the binaries?

I could see a reason: installing the tools to build the native binaries on Windows can be an annoying task, but perhaps we should document the reason we avoid this.

andyleejordan (approved):

build.psm1
SteveL-MSFT: The problem is incurring the install of chocolatey, then the Win10 SDK, and the required build tools.  I think ideally we should get `start-psbootstrap` to work in PSCore6 so Windows PowerShell isn't needed and thus we can just pick it up from the running PSCore6.  This code was just a fallback as on AppVeyor we are using Windows PowerShell to build.

daxian-dbw: Here are my 2 cents:
1. I think we should by default continue to build native component for OSX and Linux. Unlike `pwrshplugin.dll`, `libpsl` is much more crucial to PSCore (it has important native implementations that PS engine needs to call in via P/inovke), and it changes far more frequently than the WinRM plugin dll -- for example, @jeffbi made changes to it to fix the cyclic symbolic link detection issue; @PaulHigin made changes to it to create the SSH process for remoting.
2. For `pwrshplugin.dll`, `pwrshplugin.pdb` and `Install-PowerShellRemoting.ps1`, they should be made into a NuGet package and published to powershell-core feed. So we can depend on dotnet-cli to handle them and it's also more robust and clean. When we need to make changes to `pwrshplugin.dll`, we just publish a new version of the NuGet package.

SteveL-MSFT: @daxian-dbw if we're still making changes to libpsl, then we should build it by default.  agree it would be better to make `pswrshplugin` into a NuGet pkg.  I'll close this PR.

PR #4052, opened 2017-06-19
Remove nightly docker images

TravisEz13:
Nightly docker images have not been working for months.  
Removing and cleaning up docs.
If we want to get this working again, we can recreate this.  
The script should be written in PowerShell anyway because we need to support all OSs.


daxian-dbw:
Please update .spelling to fix the test failure.

daxian-dbw (approved):

PR #3001, opened 2017-01-13
Remove trailing whitespace

lzybkr:
I was concerned about #2725 changing file encodings and possibly other characters. If the number of changes was small, I would have just reviewed it, but there a ton of changes.

I took the script written by @powercode and modified it some to verify we only deleted spaces and tabs, the result is [here](https://gist.github.com/lzybkr/f040f18d1fbfff9eaf3f4533e24126fe).

With this extra check, I found 25 or so files where #2725 would have changed characters we might not have wanted changed. Those files are not included in the PR, I'll fix those in another PR - and at the same time fix those funny characters or fix the file encoding to be true ascii.

PR #4054, opened 2017-06-20
Add description of recommended settings for VS Code

iSazonov:


iSazonov:
CI OSX temporary failed

docs/learning-powershell/using-vscode.md
```diff
@@ -71,6 +71,22 @@ If you wish to use a specific installation of PowerShell with Visual Studio Code
 3. Replace the setting with the path to the desired PowerShell executable
 4. Save the settings file and restart Visual Studio Code
 
+#### Configuration settings for Visual Studio Code
+
+By using the steps in the previous paragraph you can add onfiguration settings in `settings.json`.
+
+We recommend the following  configuration settings for Visual Studio Code:
```
SteveL-MSFT: Looks like a double space between `following` and `configuration`

docs/learning-powershell/using-vscode.md
```diff
@@ -71,6 +71,22 @@ If you wish to use a specific installation of PowerShell with Visual Studio Code
...
+
+```json
+{
+    "csharp.suppressDotnetRestoreNotification": true,
+    "editor.renderWhitespace": "all",
+    "editor.renderControlCharacters": true,
+    "omnisharp.projectLoadTimeout": 120,
+    "files.trimTrailingWhitespace": true
```
SteveL-MSFT: should we just limit this to resx and powershell script files for now or are all the c# files clean?

.github/CONTRIBUTING.md
```diff
@@ -61,6 +61,10 @@ To run the spellchecker, follow the steps as follows:
 
 ## Contributing to Code
 
+### Code Editors
+
+[You can use multiplatform Visual Studio Code (VS Code)][use-vscode-editor]
```
SteveL-MSFT: I think it should say: You should use the multi-platform ...

And make just the `Visual Studio Code (VSCode)` the hyperlink instead of the whole line

SteveL-MSFT (changes requested):

docs/learning-powershell/using-vscode.md
iSazonov: Fixed.

docs/learning-powershell/using-vscode.md
iSazonov: C# files was fixed in #3001 by lzybkr.

.github/CONTRIBUTING.md
iSazonov: Fixed.

SteveL-MSFT (approved):
LGTM

iSazonov:
@rkeithhill   Any thoughts?

rkeithhill:
@iSazonov  The problem with putting this in the docs is that few people read the docs.  I suggest that at some point we either add a `.vscode\settting.json` for this project or perhaps even better, add a `.editorconfig` file in the root dir.  This is a config file that supports multiple editors.  VS 2017 supports this file and VSCode does if you install the `EditorConfig` extension.  

This brings me to my next suggestion.  Consider adding a `.vscode\extensions.json` file which will prompt VSCode users to install the "recommended" extensions for this project.  Perhaps with contents like this:
```
{
    // See http://go.microsoft.com/fwlink/?LinkId=827846
    // for the documentation about the extensions.json format
    "recommendations": [
        "ms-vscode.csharp",
        "ms-vscode.powershell",
        "EditorConfig.editorconfig"
    ]
}
```

iSazonov:
@rkeithhill Thanks for great feedback!

I think we should be more conservative - other developers might have their preferences and we shouldn't overwrite them.

rkeithhill:
Keep in mind that the use of a "workspace" specific settings file only overrides "user" specific settings for that workspace.  And that is typically what you want.  That is, a user might prefer tabs to spaces but this project uses spaces.  So adding that setting to an editor config file will ensure that spaces are used for this user in only this workspace.  In his own projects, tabs will still be used.

iSazonov:
@rkeithhill The PR is for docs only. I believe your suggestions is great for new PR. Could you please open it?

daxian-dbw (approved):

PR #4058, opened 2017-06-20
Use -c when calling powershell.exe in tests to avoid issues when argument positions change

SteveL-MSFT:
Updated tests to explicitly use `-c` to indicate a command


TravisEz13 (approved):

SteveL-MSFT:
@TravisEz13 can you merge?

Issue #4060, opened 2017-06-20
Get-AuthenticodeSignature -FilePath does not provide the TimeStamperCertificate in the returned signature


PaulHigin:
This only happens when using the newer catalog file signing APIs, which is normally not reproducible with SxS GitHub PowerShell which code path uses the older APIs.  Fix is to refactor the older code path that does obtain the time stamp cert so that the newer catalog signing APIs can also use it.

Repro
------
```powershell
Get-AuthenticodeSignature C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe | select TimeStamperCertificate
```
Result
------
Time stamp certificate property is empty

Expected
----------
Time stamp certificate property correctly filled

PaulHigin:
Fixed with PR #4061

PR #4061, opened 2017-06-20
Add missing timestamper cert to the newer catalog signing APIs

PaulHigin:
This fixes Issue #4060

Code to obtain the file signature time stamp cert was missing from the newer catalog signing APIs, with the result that Get-AuthenticodeSignature would not include the time stamp cert in the Signature object.

Fix is to refactor and use the older API code for obtaining this.

I didn't add a test for this because no catalog signing tests are in GitHub yet and porting them will be a separate work item.

### DIFF
--- a/src/System.Management.Automation/security/Authenticode.cs
+++ b/src/System.Management.Automation/security/Authenticode.cs
@@ -363,10 +363,23 @@ namespace System.Management.Automation
                             DWORD error = GetErrorFromSignatureState(sigInfo.nSignatureState);
 
                             X509Certificate2 cert = null;
+
                             if (ppCertContext != IntPtr.Zero)
                             {
                                 cert = new X509Certificate2(ppCertContext);
-                                signature = new Signature(filename, error, cert);
+
+                                // Get the time stamper certificate if available
+                                IntPtr pProvSigner = IntPtr.Zero;
+                                X509Certificate2 timestamperCert = null;
+                                TryGetProviderSigner(phStateData, ref pProvSigner, ref timestamperCert);
+                                if (timestamperCert != null)
+                                {
+                                    signature = new Signature(filename, error, cert, timestamperCert);
+                                }
+                                else
+                                {
+                                    signature = new Signature(filename, error, cert);
+                                }
 
                                 switch (sigInfo.nSignatureType)
                                 {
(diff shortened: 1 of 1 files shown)


src/System.Management.Automation/security/Authenticode.cs
```diff
@@ -578,70 +591,81 @@ private static X509Certificate2 GetCertFromChain(IntPtr pSigner)
...
+                        signature = new Signature(filePath,
+                                                  error,
+                                                  signerCert);
+                    }
+
+                    signature.SignatureType = SignatureType.Authenticode;
+                }
+            }
+
+            Diagnostics.Assert(((error == 0) && (signature != null)) || (error != 0), "GetSignatureFromWintrustData: general crypto failure");
```
TravisEz13: Question about this case, will sync offline

TravisEz13 (approved):

src/System.Management.Automation/security/Authenticode.cs
```diff
@@ -578,70 +591,81 @@ private static X509Certificate2 GetCertFromChain(IntPtr pSigner)
...
+            if ((signature == null) && (error != 0))
+            {
+                signature = new Signature(filePath, error);
+            }
+
+            return signature;
+        }
 
+        [ArchitectureSensitive]
+        private static bool TryGetProviderSigner(IntPtr wvtStateData, ref IntPtr pProvSigner, ref X509Certificate2 timestamperCert)
```
daxian-dbw: It seems to me both `pProvSigner` and `timestamperCert` are better to be an out parameter instead of a ref parameter.

src/System.Management.Automation/security/Authenticode.cs
```diff
@@ -363,10 +363,23 @@ private static Signature GetSignatureFromCatalog(string filename)
...
+                                X509Certificate2 timestamperCert = null;
+                                TryGetProviderSigner(phStateData, ref pProvSigner, ref timestamperCert);
+                                if (timestamperCert != null)
+                                {
+                                    signature = new Signature(filename, error, cert, timestamperCert);
+                                }
+                                else
+                                {
+                                    signature = new Signature(filename, error, cert);
+                                }
```
daxian-dbw: Once the parameters `pProvSigner` and `timestamperCert` are changed to `out` parameters, you can use the new language feature in C# for out parameters:
```CSharp
if (int.TryParse(input, out int result))
    WriteLine(result);
else
    WriteLine("Could not parse input");
```
So this code block can be rewritten to:
```CSharp
if (TryGetProviderSigner(phStateData, out IntPtr pProvSigner, out X509Certificate2 timestamperCert) && timestamperCert != null)
{
    signature = new Signature(filename, error, cert, timestamperCert);
}
else
{
    signature = new Signature(filename, error, cert);
}
```

src/System.Management.Automation/security/Authenticode.cs
```diff
@@ -578,70 +591,81 @@ private static X509Certificate2 GetCertFromChain(IntPtr pSigner)
             DWORD error,
             NativeMethods.WINTRUST_DATA wtd)
         {
+            s_tracer.WriteLine("GetSignatureFromWintrustData: error: {0}", error);
+
             Signature signature = null;
-            X509Certificate2 signerCert = null;
+            IntPtr pProvSigner = IntPtr.Zero;
             X509Certificate2 timestamperCert = null;
+            if (TryGetProviderSigner(wtd.hWVTStateData, ref pProvSigner, ref timestamperCert))
```
daxian-dbw: Same here. Once change to out parameters, you can use the new feature in C# to make the code cleaner.

PaulHigin: Good idea.  Will change.

daxian-dbw (approved):

mirichmo:
I created issue #4087 to document the lack of tests

PaulHigin:
@mirichmo Is there anything preventing this from being merged?  I also need to port this to internal branches.  Thanks!

Issue #2909, opened 2016-12-20
The output of help command is one letter per line


yanxyz:
Run help command:

```powershell
> help cd
```

![20161220_194627](https://cloud.githubusercontent.com/assets/737120/21349818/f9360bb8-c6ee-11e6-9f2c-25c58b863e9b.png)

The one on the right is the latest PowerShell:

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.0.0-alpha
BuildVersion                   3.0.0.0
SerializationVersion           1.1.0.1
PSRemotingProtocolVersion      2.3
WSManStackVersion              3.0
CLRVersion
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
GitCommitId                    v6.0.0-alpha.14
PSEdition                      Core
```

The one on the left is system embeded PowerShell:

```powershell
> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      5.1.14393.576
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.14393.576
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

My OS is Windows 10 x64 Chinese language. 

The `Get-Help`  cmdlet has no problem.

randomvariable:
Just looking at #2809. Is this because on Windows, the help file gets piped into more, and this doesn't support unicode properly, and is it an issue for the command prompt team?

chunqingchen:
This only repros on cmd running powershell.exe, it won't repro on powershell console running Core powershell.exe

chunqingchen:
@yanxyz can we close the issue?

SteveL-MSFT:
@chunqingchen I don't think that's a viable workaround.  if `more` has an issue, then we should have `help` have a dependency on `more`.

yanxyz:
@chunqingchen thanks. I have no idea the inner detail. As a normal user I'm just curious about why it doesn't work after update. I have tried the second method you mentioned, it works. 

You could close this issue when you think it is OK.

iSazonov:
@chunqingchen Can we close the Issue as resolved?

PR #4074, opened 2017-06-22
Redefine the unicode output condition for nanoserver for "more" function

chunqingchen:
Fix issue #2909 

I haven't included a test case with this fix because 
1. unicode issue is better to be judged by eye.
2. this issue only repros in a non-english windows system with ps running under cmd window. Our current test environment won't touch this scenario.

Thus, one time manual verification might be enough.

Summary of the issue:

Under non-english OS, if user start ps core under comandline and type "help <cmdletName>", the  output will be one character per line

Root cause of the issue:
Only Nano needs to use Unicode, but Windows and Linux need the default encoding. The way we judge if the current os is Nano server has been outdated.
Fix:
Update the condition we use to judge nano server.

### DIFF
--- a/src/System.Management.Automation/engine/InitialSessionState.cs
+++ b/src/System.Management.Automation/engine/InitialSessionState.cs
@@ -5329,7 +5329,7 @@ end
         internal const string DefaultMoreFunctionText = @"
 param([string[]]$paths)
 # Nano needs to use Unicode, but Windows and Linux need the default
-$OutputEncoding = if ($IsWindows -and $IsCoreCLR) {
+$OutputEncoding = if ([System.Management.Automation.Platform]::IsNanoServer -or [System.Management.Automation.Platform]::IsIoT) {
     [System.Text.Encoding]::Unicode
 } else {
     [System.Console]::OutputEncoding


src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -5325,7 +5325,7 @@ internal static SessionStateAliasEntry[] BuiltInAliases
         internal const string DefaultMoreFunctionText = @"
 param([string[]]$paths)
 # Nano needs to use Unicode, but Windows and Linux need the default
-$OutputEncoding = if ($IsWindows -and $IsCoreCLR) {
+$OutputEncoding = if ($IsNanoServer) {
```
daxian-dbw: There is no such variable `$IsNanoServer` in powershell core.
You need to look at `[System.Management.Automation.Platform]::IsIoT` and `[System.Management.Automation.Platform]::IsNanoServer`

daxian-dbw:
I still would like to have a test for this fix. If it's really impossible, can you please at least have a screenshot to show the behavior difference before and after the fix?

chunqingchen:
Before the fix:
image

After the fix:
image

chunqingchen:
@daxian-dbw thanks! the comment has been resolved and the screenshots are attached.

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -5329,7 +5329,7 @@ internal static SessionStateAliasEntry[] BuiltInAliases
         internal const string DefaultMoreFunctionText = @"
 param([string[]]$paths)
 # Nano needs to use Unicode, but Windows and Linux need the default
-$OutputEncoding = if ($IsWindows -and $IsCoreCLR) {
+$OutputEncoding = if ([System.Management.Automation.Platform]::IsNanoServer) {
```
daxian-dbw: You also need to check `[System.Management.Automation.Platform]::IsIoT`. IoT is very similar to NanoServer.

chunqingchen:
@daxian-dbw resolved

daxian-dbw (approved):

PR #4078, opened 2017-06-22
Fixed BaseDirectory tests by adding -c instead of positional argument

adityapatwardhan:
Fix the test failing on Travis CI, due to change 41f65de57c540a9b6e058c67baabbe5b10ce18e8
PowerShell now expects the positional parameter as a file for POSIX compliance. The test needed to be updated accordingly.


SteveL-MSFT (approved):
LGTM

PR #4083, opened 2017-06-22
Address static analysis issues in SSH remoting

SteveL-MSFT:
- remoteRunspaces is not used TFS:10618323
- not checking return value of ResumeThread() TFS:10618344

### DIFF
--- a/src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
+++ b/src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
@@ -2290,7 +2290,7 @@ namespace System.Management.Automation.Runspaces
                 Process result = Process.GetProcessById(lpProcessInformation.dwProcessId);
                 uint returnValue = PlatformInvokes.ResumeThread(lpProcessInformation.hThread);
 
-                if (returnValue == 4294967295) // same as (DWORD)-1
+                if (returnValue == PlatformInvokes.RESUME_THREAD_FAILED)
                 {
                     throw new Win32Exception(Marshal.GetLastWin32Error());
                 }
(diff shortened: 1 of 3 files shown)


src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -2288,7 +2288,12 @@ private string GetCurrentUserName()
 
                 // At this point, we should have a suspended process.  Get the .Net Process object, resume the process, and return.
                 Process result = Process.GetProcessById(lpProcessInformation.dwProcessId);
-                PlatformInvokes.ResumeThread(lpProcessInformation.hThread);
+                uint returnValue = PlatformInvokes.ResumeThread(lpProcessInformation.hThread);
+
+                if (returnValue == 4294967295) // same as (DWORD)-1
```
mirichmo: We should not use magic numbers in the code. Please define a descriptive constant for this value.

mirichmo (changes requested):

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
SteveL-MSFT: fixed

SteveL-MSFT:
@mirichmo fixed

PaulHigin (approved):
LGTM

src/System.Management.Automation/utils/PlatformInvokes.cs
```diff
@@ -701,6 +701,7 @@ public SECURITY_ATTRIBUTES()
 
         [DllImport(PinvokeDllNames.ResumeThreadDllName, CharSet = CharSet.Unicode, SetLastError = true)]
         public static extern uint ResumeThread(IntPtr threadHandle);
+        internal static uint RESUME_THREAD_FAILED = 4294967295; // (DWORD)-1
```
mirichmo: Please use System.UInt32.MaxValue instead

mirichmo (changes requested):

src/System.Management.Automation/utils/PlatformInvokes.cs
SteveL-MSFT: will fix, seems obvious now :)

mirichmo (approved):

Issue #2634, opened 2016-11-05
Add ValidatePositive and ValidateNonNegative attributes


iSazonov:
From @lzybkr #2492 
> it seems like we should have `ValidatePositive` and `ValidateNonNegative` attributes.

as replacement of `[ValidateRange((int)1, Int32.MaxValue)] ` and `[ValidateRange((int)0, Int32.MaxValue)] `

dchristian3188:
Picking this one up in PR: 
https://github.com/PowerShell/PowerShell/pull/4084

lzybkr:
Closed via #4084

PR #4084, opened 2017-06-22
Added Addtional Validation Attributes

dchristian3188:
First pass of additional parameter validations attributes as mentioned here: https://github.com/PowerShell/PowerShell/issues/2634

New attributes include:
ValidatePositiveAttribute
ValidateNonNegativeAttribute
ValidateNegativeAttribute 
ValidateNonPositiveAttribute


test/powershell/engine/ValidateAttributes.Tests.ps1
```diff
@@ -25,4 +25,111 @@ Describe 'Validate Attributes Tests' -Tags 'CI' {
             { function Local:foo { param([ValidateCount(2, 4)] [string[]] $bar) }; foo 1,2,3,4 } | Should Not Throw
         }
     }
+
+    Context "ValidatePositive" {
+        BeforeAll {
+           $testCases = @(
+               @{ sb = { function Local:foo { param([ValidatePositive()] [int] $bar) }; foo -1  }; FullyQualifiedErrorId = "ParameterArgumentValidationError,foo";  InnerErrorId = "ValidateRangeTooSmall" }
```
lzybkr: `local:` is unnecessary.

lzybkr: Also - this is hard to read with all keys on the same line - I suggest each key on it's own line.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1098,6 +1098,194 @@ private static Type GetCommonType(Type minType, Type maxType)
...
+    /// <seealso cref="ValidatePositiveAttribute"/>,
+    /// <seealso cref="ValidateNonNegativeAttribute"/>
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
+    public class ValidatePositiveBaseAttribute : ValidateEnumeratedArgumentsAttribute
+    {
+        /// <summary>
+        /// Minimum allowed value for the attribute
+        /// </summary>
+        public int minValue;
```
lzybkr: What about other data types? Smaller should just work, but what about `long`, `float`, `double` or `bigint`?

lzybkr: It's possible `dynamic` helps you here

In validate element, you would write:

```c#
dymamic val = element;
if (val > minValue) ...
```

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1098,6 +1098,194 @@ private static Type GetCommonType(Type minType, Type maxType)
...
     /// <summary>
+    /// Base class for postive number valididation attributes
+    /// </summary>
+    /// <remarks>
+    /// Base class used by postive and non negative attribute
+    /// <seealso cref="ValidatePositiveAttribute"/>,
+    /// <seealso cref="ValidateNonNegativeAttribute"/>
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
+    public class ValidatePositiveBaseAttribute : ValidateEnumeratedArgumentsAttribute
```
lzybkr: This base class is not generally useful, so I'd prefer avoiding introducing a new public type.

lzybkr (changes requested):
We have existing cmdlet parameters where these new attributes should be used.

I think it make sense to start using the new attribute right away.

We also need to think through usage on all data types, and avoid introducing unnecessary new public apis (the base classes).

test/powershell/engine/ValidateAttributes.Tests.ps1
dchristian3188: I'll get this cleaned up. I was trying to follow the existing format of the file. Is it cool to update those other test as part of this PR as well?

src/System.Management.Automation/engine/Attributes.cs
dchristian3188: Will remove

src/System.Management.Automation/engine/Attributes.cs
dchristian3188: Any tips on how to include the larger numbers? I was thinking if checking if the element has a ```CompareTo()``` but not sure if this is correct. For example, string also have compare to which would be an invalid type.

iSazonov:
@dchristian3188 Please put a message in #2634 that you are starting the fix to exclude conflicts.

iSazonov:
>We also need to think through usage on all data types, and avoid introducing unnecessary new public apis (the base classes).

We could exclude the new apis at all - if we want usage on all data types we should generalize existing `ValidateElement()`. So we can generalize `ValidateRangeAttribute.ValidateElement()`  and add new constructor to support:
```c#
ValidateRange(Range="Positive")
```

src/System.Management.Automation/engine/Attributes.cs
lzybkr: Try my suggestion to use `dynamic` - it might just work. Behind the scenes, the C# compiler will generate code calling the correct operator *after* testing the type of `val`.

lzybkr:
@iSazonov - I was saying we don't need any unnecessary apis like the abstract base class used to share a tiny bit of code - that could easily be done without inheritance.

Your suggestion might look like:
```C#
public enum RangeKind {
    Positive, NonNegative, Negative, NonPositive
}

    // In ValidateRangeAttribute
    RangeKind? _rangeKind;
    public ValidateRange(RangeKind kind) { _rangeKind = kind; }

        // in ValidateElement
        if (_rangeKind.HasValue) { ... }
        else { /* existing code */ }
```

That might be fine too, I'm not sure I have a preference either way.

iSazonov:
@lzybkr Thanks for you prototype. I like it. We use the similar approach for enhancing ValidateSet with valid values generator. First locally I used inheritance, but your idea (to add new constructor) was better. So I believe we should use the same pattern here. Users will find it easier to know and apply the enhancements than the new apis.

src/System.Management.Automation/engine/Attributes.cs
dchristian3188: sorry, never used this before. Updated to use dynamic

dchristian3188:
I think you guys are right. Extending Validate range does feel cleaner.

test/powershell/engine/ValidateAttributes.Tests.ps1
```diff
@@ -22,7 +46,270 @@ Describe 'Validate Attributes Tests' -Tags 'CI' {
...
+            param($sb)
+                $sb | Should Not Throw
+        }
+    }
+
+    Context "ValidateRange - Positive" {
+        BeforeAll {
+           $testCases = @(
+               @{ 
+                   sb = { function foo { param([ValidateRange('Positive')] [int] $bar) }; foo -1 }
```
iSazonov: Please add test with the enum:
```
ValidateRange([ValidateRangeKind]::Positive)
```
I guess that's enough replace 'Positive' in one test and don't duplicate all tests.

test/powershell/engine/ValidateAttributes.Tests.ps1
```diff
@@ -22,7 +46,270 @@ Describe 'Validate Attributes Tests' -Tags 'CI' {
...
+                   sb = { function foo { param([ValidateRange('Positive')] $bar) }; foo "one" }
+                   FullyQualifiedErrorId = "ParameterArgumentValidationError,foo"
+                   InnerErrorId = "" 
+                }
+           )
+
+           $validTestCases = @(
+               @{ 
+                   sb = { function foo { param([ValidateRange('Positive')] [int] $bar) }; foo 15 }
+                   TestValue = 15
```
iSazonov: We use `LanguagePrimitives.TryConvertTo` so we very "generic" here and we should add tests for non-numeric values.

test/powershell/engine/ValidateAttributes.Tests.ps1
```diff
@@ -22,7 +46,270 @@ Describe 'Validate Attributes Tests' -Tags 'CI' {
...
+            }
+        }
+
+        It 'No Exception: positive argument "<TestValue>"' -TestCases $validTestCases {
+            param($sb, $testValue)
+                $sb | Should Not Throw
+        }
+    }
+    
+    Context "ValidateRange - NonNegative" {
```
iSazonov: I think we can combine all ValidateRangeKind-s in one TestCases:
```
sb = { function foo { param([ValidateRange(   '<TestCase-Parameter-Here>'    )] [int] $bar) }; foo -1 } 
```

iSazonov (commented):
Leave a comment

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -896,6 +896,32 @@ public ValidateLengthAttribute(int minLength, int maxLength) : base()
...
+        
+        /// <Summary>
+        /// Range is less than 0
+        /// </Summary>
+        Negative,
+
+        /// <Summary>
+        /// Range is less than or equal to 0
+        /// </Summary>        
+        NonPositive
```
dchristian3188: My Vote is for 

1 - Positive, Negative, NonPostive, NonNegative
2 - ValueGTZero, ValueGEZero, ValueGTZero, ValueGEZero

test/powershell/engine/ValidateAttributes.Tests.ps1
dchristian3188: I tried this but it failed since ValidateRangeKind is not exposed as a type accelerator. Should i use the full namespace or add [ValidateRangeKind]?

iSazonov: Yes.

test/powershell/engine/ValidateAttributes.Tests.ps1
iSazonov: Closed.

test/powershell/engine/ValidateAttributes.Tests.ps1
iSazonov: Please add the tests.

test/powershell/engine/ValidateAttributes.Tests.ps1
```diff
@@ -3,12 +3,36 @@ Describe 'Validate Attributes Tests' -Tags 'CI' {
...
+                }
+               @{
+                   sb = { function foo { param([ValidateCount(1,-1)] [string[]] $bar) }; foo }
+                   FullyQualifiedErrorId = "ExceptionConstructingAttribute"          
+                   InnerErrorId = "" 
+                }
+               @{ 
+                   sb = { function foo { param([ValidateCount(2, 1)] [string[]] $bar) }; foo }
+                    FullyQualifiedErrorId = "ValidateRangeMaxLengthSmallerThanMinLength"
+                    InnerErrorId = ""
```
iSazonov: Closed.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -943,43 +971,98 @@ protected override void ValidateElement(object element)
...
-            var o = element as PSObject;
-            if (o != null)
+            if(_rangeKind.HasValue)
             {
-                element = o.BaseObject;
+                dynamic val = element;
+                switch (_rangeKind)
+                {
+                    case ValidateRangeKind.Positive:
+                        if(val <= 0)
```
iSazonov: Closed.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -943,43 +971,98 @@ protected override void ValidateElement(object element)
                         Metadata.ValidateNotNullFailure);
             }
 
-            var o = element as PSObject;
-            if (o != null)
+            if(_rangeKind.HasValue)
             {
-                element = o.BaseObject;
+                dynamic val = element;
```
iSazonov: Closed.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -896,6 +896,32 @@ public ValidateLengthAttribute(int minLength, int maxLength) : base()
...
     }
 
+    /// <Summary>
+    /// predefined range kind to use with 
+    /// ValidateRangeAttribute
+    /// </Summary>
+    public enum ValidateRangeKind{
+        
+        /// <Summary>
+        /// Range is greater than 0
```
iSazonov: Closed.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -896,6 +896,32 @@ public ValidateLengthAttribute(int minLength, int maxLength) : base()
         }
     }
 
+    /// <Summary>
+    /// predefined range kind to use with 
+    /// ValidateRangeAttribute
+    /// </Summary>
+    public enum ValidateRangeKind{
```
iSazonov: Closed.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -896,6 +896,32 @@ public ValidateLengthAttribute(int minLength, int maxLength) : base()
         }
     }
 
+    /// <Summary>
+    /// predefined range kind to use with 
+    /// ValidateRangeAttribute
```
iSazonov: Closed.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -943,43 +970,100 @@ protected override void ValidateElement(object element)
                         Metadata.ValidateNotNullFailure);
             }
 
-            var o = element as PSObject;
-            if (o != null)
+            if(_rangeKind.HasValue)
             {
-                element = o.BaseObject;
+                // Using dynamic to handle different type of numbers such as
+                // int, double, decimal and float.
```
iSazonov: CompareTo() allow us be more general. I think we should enhance the comment.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -943,43 +970,100 @@ protected override void ValidateElement(object element)
...
+                var o = element as PSObject;
+                if (o != null)
                 {
-                    element = resultValue;
+                    element = o.BaseObject;
                 }
-                else
+
+                // minRange and maxRange have the same type, so we just need
+                // to compare to one of them
```
iSazonov: Please correct the comment formatting.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -943,43 +970,100 @@ protected override void ValidateElement(object element)
...
-                    null, Metadata.ValidateRangeGreaterThanMaxRangeFailure,
-                    element.ToString(), MaxRange.ToString());
+                if (_maxComparable.CompareTo(element) < 0)
+                {
+                    throw new ValidationMetadataException("ValidateRangeTooBig",
+                        null, Metadata.ValidateRangeGreaterThanMaxRangeFailure,
+                        element.ToString(), MaxRange.ToString());
+                }
             }
+
```
iSazonov: Please remove the extra line.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -943,43 +971,98 @@ protected override void ValidateElement(object element)
...
             {
-                element = o.BaseObject;
+                dynamic val = element;
+                switch (_rangeKind)
+                {
+                    case ValidateRangeKind.Positive:
+                        if(val <= 0)
+                        {
+                            throw new ValidationMetadataException(
+                                "ValidateRangeTooSmall",
```
lzybkr: It's convenient but not important for the ErrorId to match the name of the property in the resx file.

Note that we may have a few legacy places where we do use the ErrorId to load the resource - in those cases, it is important that they match, but it would be even better to convert those locations to this pattern where the ErrorId and format string are separate parameters.

---
In reply to: [123883907](https://github.com/PowerShell/PowerShell/pull/4084#discussion_r123883907) [](ancestors = 123883907)

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -896,6 +896,31 @@ public ValidateLengthAttribute(int minLength, int maxLength) : base()
         }
     }
 
+    /// <Summary>
+    /// Predefined range kind to use with ValidateRangeAttribute.
+    /// </Summary>
+    public enum ValidateRangeKind {
```
lzybkr: Please follow the conventions of the file you are editing, e.g. braces normally go on their own line, and you have no space after `if`.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -943,43 +970,100 @@ protected override void ValidateElement(object element)
...
+            if(_rangeKind.HasValue)
             {
-                element = o.BaseObject;
+                // Using dynamic to handle different type of numbers such as
+                // int, double, decimal and float. 
+                dynamic val = element;
+                switch (_rangeKind)
+                {
+                    case ValidateRangeKind.Positive:
+                        if(val.CompareTo(0) <= 0)
```
lzybkr: If the argument does not implement `IComparable`, the error message will not be good.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -943,43 +970,100 @@ protected override void ValidateElement(object element)
...
             }
 
-            var o = element as PSObject;
-            if (o != null)
+            if(_rangeKind.HasValue)
             {
-                element = o.BaseObject;
+                // Using dynamic to handle different type of numbers such as
+                // int, double, decimal and float. 
+                dynamic val = element;
```
lzybkr: This might be too general - e.g. strings implement `IComparable`, but strings aren't positive or negative.

lzybkr: I also see you omit the test for `PSObject` - you'll probably want it for your code as well - and you should also ensure you have tests to cover simple (non-wrapped) arguments and PSObject wrapped arguments.

lzybkr (changes requested):
I see you're reusing existing error messages - one reason for this feature was to improve the error message.

And once we've settled on naming, I'd really like to see this new form of the attribute used in our own code base - there are 80 or so places that could use this.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: >This might be too general - e.g. strings implement  IComparable , but strings aren't positive or negative.

In general, we cannot know if the type is comparable with zero in advance. Should we test this?

dchristian3188: How do you feel about reusing the ```GetCommonType``` method from the validate range class? 
We could use the incoming object and ```0```. I believe this will also throw the correct message if the user attempts to pass an incompatable object such as string

lzybkr: That seems reasonable.

dchristian3188: Reusing ```GetCommonType``` looked promising but failed in certain scenarios. 

For example, if the incoming element was ```-1```, this value was interpreted as a string. This would cause the ```GetCommonType``` method to fail since it could not find the common group between ```0``` and the string.

I've resorted to using the ```LanguagePrimitives.TryConvertTo``` method and casting the incoming value as a double since this is the largest number type currently supported by the function. 

Is there a cleaner way to test if the object is comparable with zero that I might be missing?

lzybkr: Can you share an example where the element was a string and you expected an integer?

dchristian3188: Here's what i had for ```ValidateElement```.

```c#
protected override void ValidateElement(object element)
{
    if (element == null)
    {
        throw new ValidationMetadataException(
                "ArgumentIsEmpty",
                null,
                Metadata.ValidateNotNullFailure);
    }

    var o = element as PSObject;
    if (o != null)
    {
        element = o.BaseObject;
    }
    
    if (_rangeKind.HasValue)
    {

        Type commonType = GetCommonType((0).GetType(),element.GetType());

        ...
```

Then in PowerShell

```powershell
function foo { param([ValidateRange('Positive')] $bar) }

foo -1
```

At this point, element will be a string when we enter ```GetCommonType```.

What's interesting is that this does work.

```powershell
function foo { param([ValidateRange('Positive')] $bar) }

foo (-1)
```

I checked the existing code and they solve this by first getting the common type of the min and max values. They then try to cast element to this new type before comparison operations. 

Since I don't have a common type, I was thinking of trying the cast to double.

iSazonov: It is interesting. Maybe it is by design for `-1` but confusing:
```powershell
PS > function foo { param($bar) $bar.Gettype() }
PS > foo
You cannot call a method on a null-valued expression.
At line:1 char:29
+ function foo { param( $bar) $bar.Gettype() }
+                             ~~~~~~~~~~~~~~
    + FullyQualifiedErrorId : InvokeMethodOnNull

PS > foo 0

True     True     Int32                                    System.ValueType


PS > foo -1

PS > foo 1

True     True     Int32                                    System.ValueType
```

iSazonov: In general case, a custom class can implement `CompareTo()` and allow `(CustomClass)0`. If the class don't  implement this we get an exception - it seems this is acceptable here.

lzybkr: The real problem here is the parameter is not typed - so I think this is by design. If the parameter was typed as `[int]`, then it would have worked as expected.

Based on this, I think it's still reasonable to use `GetCommonType`.

dchristian3188:
Ok, i added the check for the common type. Also tried to make the error messages a little more relevant.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1148,6 +1068,134 @@ public ValidateRangeAttribute(ValidateRangeKind kind) : base()
...
+        {
+            Type commonType = GetCommonType((0).GetType(),element.GetType());
+            if (commonType == null)
+            {
+                    throw new ValidationMetadataException(
+                    "ValidationRangeElementType",
+                    null, 
+                    Metadata.ValidateRangeElementType,
+                    element.GetType().Name, 
+                    0.GetType().Name);
```
iSazonov: `0.GetType()` and `0.GetType().Name` is good condidates for static read only constant.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1148,6 +1068,134 @@ public ValidateRangeAttribute(ValidateRangeKind kind) : base()
...
+
+            dynamic elementValue = element;
+
+            bool throwTypeError = true;
+            object dynamicZero = 0;
+            object resultValue;
+
+            if (LanguagePrimitives.TryConvertTo(element, commonType, out resultValue))
+            {
+                elementValue = resultValue as IComparable;
```
iSazonov: What if we get elementValue == null ?

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -970,100 +970,20 @@ protected override void ValidateElement(object element)
...
-                                "-1");
-                        }
-                        break;
-                }
+                element = o.BaseObject;
+            }
+            
+            if (_rangeKind.HasValue)
+            {
+                ValidatePredefinedRange(element);
```
iSazonov: Comparing is the same for any range kinds (including "user defined"). I believe we should create a common method to compare elements and call it with different arguments for different ranges.

dchristian3188:
I get what you're saying about the methods. I've consolidated them into one method with overload. Let me know how it looks.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -923,6 +948,11 @@ public sealed class ValidateRangeAttribute : ValidateEnumeratedArgumentsAttribut
         /// </summary>
         private Type _promotedType;
 
+        ValidateRangeKind? _rangeKind;
+
+        private readonly static Type _zeroType = 0.GetType();
+        private readonly static string _zeroTypeName = 0.GetType().Name;
```
iSazonov: Please use s_zeroType for static constant.
And maybe `_zeroTypeName = s_zeroType.Name; `

iSazonov (commented):
I meant that a range is a range and we can have one method for all range kinds. ~~We need only add 'ExpliciteRange' in enum.~~

dchristian3188:
Hey iSazonov, I'm not sure I follow on adding the additional range type. If we were to add the ```ExpliciteRange``` in the enum, wouldn't this allow a user to do something like this:

```powershell
function foo { param([ValidateRange("ExpliciteRange")] [int] $bar) }; foo 0 }
```

iSazonov:
@dchristian3188 I agree - we should follow HasValue(). That's enough to make a choice in general method.

iSazonov:
We could enhance and unify Ranges. Currently Min and Max values are allowed but user can want to exclude them. Ex., "Positive" exclude zero. So the "Positive" range looks:
```powershell
ValidateRange(0, int.MaxValue, RangeOption.ExcludeMin)
```
"Negative":
```powershell
ValidateRange( -int.MaxValue, 0, RangeOption.ExcludeMax)
```
Exclude both:
```powershell
ValidateRange( 0, 1, RangeOption.ExcludeMin+RangeOption.ExcludeMax)
```

And we can move `switch (rangeKind)` to a constructor (with assigning a specific error message) and implement general ValidateElement method.

dchristian3188:
Hi iSazonov,

I think expanding the range functionality is a cool idea, but might be out of scope for this PR. Do you want me to open a separate issue for tracking?

Are there other updates I should make to the validate positive, negative, etc?

iSazonov:
@dchristian3188 Please clarify do you plan continue in new PR or want cut off?

dchristian3188:
Hi iSazonov,

I would continue the work you mentioned in the new PR.

iSazonov:
@lzybkr  Could you please continue the code review?

lzybkr (changes requested):

ArieHein:
will this make the value zero validated both as NonNegative and NonPositive ?  
or did I miss something ?

dchristian3188:
Yes 0 will fall into both NonNegative and NonPositive. 
The breakdown is:

Positive -gt 0
NonNegative -ge 0
Negative -lt 0
NonPostive -le 0

dchristian3188:
@lzybkr 

Latest PR removes static variables. Also reworked to get rid of the dynamic variable.

Let me know if everything looks ok.

iSazonov:
@dchristian3188 For ` 0.GetType();` request was
>you should just use  typeof(int)  instead.

dchristian3188:
hi @iSazonov , sorry missed that one.

Corrections have been made. Please review

ArieHein:
Thers also the issue is 0 defined as positive or negative.

I need to test my DSC scripts to make sure if I use these validations I get the desired logic.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1055,6 +1059,142 @@ public ValidateRangeAttribute(object minRange, object maxRange) : base()
...
+
+            if (LanguagePrimitives.TryConvertTo(element, commonType, out resultValue))
+            {
+                element = (IComparable)resultValue;
+                
+                if (LanguagePrimitives.TryConvertTo(0, commonType, out resultValue))
+                {
+                    dynamicZero = (IComparable)resultValue;
+                    throwTypeError = false;
+                }
```
iSazonov: We can add `else` and exclude `throwTypeError`.

src/System.Management.Automation/resources/Metadata.resx
```diff
@@ -123,6 +123,18 @@
   <data name="ValidateRangeElementType" xml:space="preserve">
     <value>The argument cannot be validated because its type "{0}" is not the same type ({1}) as the maximum and minimum limits of the parameter. Make sure the argument is of type {1} and then try the command again.</value>
   </data>
+  <data name="ValidateRangePositiveFailure" xml:space="preserve">
+    <value>The argument "{0}" cannot be validated because its value is not greater than zero.</value>
```
iSazonov: We can "validate" but cannot "accept". Maybe better change this. Below too.

iSazonov (commented):
Leave a comment

src/System.Management.Automation/resources/Metadata.resx
dchristian3188: You're right, this does make more sense. It also follows the same verbiage as existing Validate Range

dchristian3188:
Updates made in latest PR

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1055,6 +1059,139 @@ public ValidateRangeAttribute(object minRange, object maxRange) : base()
...
+                }                    
+            }
+            else
+            {
+                throw new ValidationMetadataException(
+                    "ValidationRangeElementType",
+                    null, 
+                    Metadata.ValidateRangeElementType,
+                    element.GetType().Name, 
+                    typeof(int).Name);
```
lzybkr: Shouldn't this be `typeof(commonType).Name`?

src/System.Management.Automation/resources/Metadata.resx
```diff
@@ -123,6 +123,18 @@
   <data name="ValidateRangeElementType" xml:space="preserve">
     <value>The argument cannot be validated because its type "{0}" is not the same type ({1}) as the maximum and minimum limits of the parameter. Make sure the argument is of type {1} and then try the command again.</value>
   </data>
+  <data name="ValidateRangePositiveFailure" xml:space="preserve">
+    <value>The argument "{0}" cannot be accepted because its value is not greater than zero.</value>
```
lzybkr: For consistency, validation errors use `The argument "{0} cannot be validated` - the `cannot be accepted` phrase is not commonly used in our code base, basically just in this resx file, and maybe we should change those messages.

test/powershell/engine/ValidateAttributes.Tests.ps1
```diff
@@ -3,12 +3,36 @@ Describe 'Validate Attributes Tests' -Tags 'CI' {
...
-               @{ sb = { function Local:foo { param([ValidateCount(2, 1)] [string[]] $bar) }; foo         }; FullyQualifiedErrorId = "ValidateRangeMaxLengthSmallerThanMinLength"; InnerErrorId = "" }
-               @{ sb = { function Local:foo { param([ValidateCount(2, 2)] [string[]] $bar) }; foo 1       }; FullyQualifiedErrorId = "ParameterArgumentValidationError,foo";       InnerErrorId = "ValidateCountExactFailure" }
-               @{ sb = { function Local:foo { param([ValidateCount(2, 3)] [string[]] $bar) }; foo 1       }; FullyQualifiedErrorId = "ParameterArgumentValidationError,foo";       InnerErrorId = "ValidateCountMinMaxFailure" }
-               @{ sb = { function Local:foo { param([ValidateCount(2, 3)] [string[]] $bar) }; foo 1,2,3,4 }; FullyQualifiedErrorId = "ParameterArgumentValidationError,foo";       InnerErrorId = "ValidateCountMinMaxFailure" }
+                @{
+                    sb                       = { function foo { param([ValidateCount(-1,2)] [string[]] $bar) }; foo }
```
lzybkr: I realize you didn't introduce this, but `sb` is not a great name - it is often used for StringBuilder as well. `$ScriptBlock` is probably better.

I won't block the PR if you want to leave it though.

lzybkr (changes requested):
Just a few little things now - I think it's just about ready.

lzybkr (approved):

PR #4103, opened 2017-06-26
MacOS test failing for read-host

SteveL-MSFT:
We changed powershell.exe to require -c to execute command

### DIFF
--- a/test/powershell/Host/assets/Read-Host.Output.expect
+++ b/test/powershell/Host/assets/Read-Host.Output.expect
@@ -7,7 +7,7 @@ set powershell [lindex $argv 0]
 
 set timeout 60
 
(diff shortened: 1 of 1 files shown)


daxian-dbw (approved):
I have verified that the test passes with this fix.

andyleejordan (approved):
LGTM though I would use `-command` over `-c` for consistency.

andschwa:
@SteveL-MSFT it looks like your commit is getting created with `slee@ntdev.microsoft.com`, which I don't think you meant, as it's not associated with your GitHub account.

iSazonov:
It seems we should add the check in ScriptAnalizer.

SteveL-MSFT:
@andschwa yeah, used a different machine, let me see if I can fix that

PR #4110, opened 2017-06-26
Minimize writing ANSI escape sequences in PSReadline

### LINKED ISSUE #4075: broken terminal using powershell over ssh from windows
It appears powershell over SSH going to a Windows console is broken strangely. If you use powershell on Windows to ssh and then powershell, you get the last line repeated everytime you press a character. If you use `cmd`, it basically keeps writing a single line in place rather than ever moving.

Any other linux tool, nano, vim, etc over SSH work fine. So not sure what powershell is doing in particular that the windows powershell terminal cannot handle while handling more complex terminal abusing applications.

Steps to reproduce
------------------

Setup for example the "OpenSSH for Windows" or a msys2 ssh implementation (it doesn't matter which one).

Launch powershell on Windows. Execute the ssh client to a debian 8 machine which has powershell installed via the deb packages.

Execute powershell.

Try typing.

It ends up broken like this:

PS /home/mroszko> PS /home/mroszko> $
                  PS /home/mroszko> $P
                  PS /home/mroszko> $PS
                  PS /home/mroszko> $PSV
                  PS /home/mroszko> $PSVE
                  PS /home/mroszko> $PSVER
                  PS /home/mroszko> $PSVERS
                  PS /home/mroszko> $PSVERSI
                  PS /home/mroszko> $PSVERSIO
                  PS /home/mroszko> $PSVERSION
                  PS /home/mroszko> $PSVERSIONT
                  PS /home/mroszko> $PSVERSIONTA
                  PS /home/mroszko> $PSVERSIONTAB
                  PS /home/mroszko> $PSVERSIONTABL
                  PS /home/mroszko> $PSVERSIONTABLE

When I was simply trying to type $PSVERSIONTABLE, it automatically kept regenerating the line.
Other strange artifacts can occur instead too, cursor not moving, lines jumping, buffer bleeding.

Environment data
----------------
`
CLRVersion
GitCommitId                    v6.0.0-beta.2
OS                             Linux 3.16.0-4-amd64 #1 SMP Debian 3.16.39-1 (2016-12-30)
`

What I've tried:

Cmder/Conemu which wrap Cmd.exe and PowerShell.exe

Powershell.exe standalone with OpenSSH for Windows and msys2 ssh separately
Cmd.exe standalone with OpenSSH for Windows and msys2 ssh separately

lzybkr:
Instead of changing Console.*groundColor on every character, only set
the color when the color changes - this avoids writing out ANSI escape
sequences after every character.

Fixes #4075

### DIFF
--- a/src/Microsoft.PowerShell.PSReadLine/ConsoleLib.cs
+++ b/src/Microsoft.PowerShell.PSReadLine/ConsoleLib.cs
@@ -1075,11 +1075,11 @@ namespace Microsoft.PowerShell.Internal
                 // TODO: use escape sequences for better perf
                 var nextForegroundColor = buffer[i].ForegroundColor;
                 var nextBackgroundColor = buffer[i].BackgroundColor;
-                if (lastForegroundColor != nextForegroundColor)
+                if (nextForegroundColor != lastForegroundColor)
                 {
                     Console.ForegroundColor = lastForegroundColor = nextForegroundColor;
                 }
-                if (lastBackgroundColor != nextBackgroundColor)
+                if (nextBackgroundColor != lastBackgroundColor)
                 {
                     Console.BackgroundColor = lastBackgroundColor = nextBackgroundColor;
                 }
@@ -1087,13 +1087,13 @@ namespace Microsoft.PowerShell.Internal
                 Console.Write(buffer[i].UnicodeChar);
             }
 
-            if (backgroundColor != lastBackgroundColor)
-            {
-                Console.BackgroundColor = backgroundColor;
-            }
             if (foregroundColor != lastForegroundColor)
             {
                 Console.ForegroundColor = foregroundColor;
+            }
+            if (backgroundColor != lastBackgroundColor)
+            {
+                Console.BackgroundColor = backgroundColor;
             }
         }
 
--- a/src/Microsoft.PowerShell.PSReadLine/ConsoleLib.cs
+++ b/src/Microsoft.PowerShell.PSReadLine/ConsoleLib.cs
@@ -1068,17 +1068,33 @@ namespace Microsoft.PowerShell.Internal
 
             Console.SetCursorPosition(0, (top>=0) ? top : 0);
 
+            var lastForegroundColor = (ConsoleColor)(-1);
+            var lastBackgroundColor = (ConsoleColor)(-1);
             for (int i = 0; i < buffer.Length; ++i)
             {
                 // TODO: use escape sequences for better perf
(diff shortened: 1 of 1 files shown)


src/Microsoft.PowerShell.PSReadLine/ConsoleLib.cs
```diff
@@ -1068,17 +1068,33 @@ public void WriteBufferLines(BufferChar[] buffer, ref int top, bool ensureBottom
...
+            var lastForegroundColor = (ConsoleColor)(-1);
+            var lastBackgroundColor = (ConsoleColor)(-1);
             for (int i = 0; i < buffer.Length; ++i)
             {
                 // TODO: use escape sequences for better perf
-                Console.ForegroundColor =  buffer[i].ForegroundColor;
-                Console.BackgroundColor =  buffer[i].BackgroundColor;
+                var nextForegroundColor = buffer[i].ForegroundColor;
+                var nextBackgroundColor = buffer[i].BackgroundColor;
+                if (lastForegroundColor != nextForegroundColor)
```
SteveL-MSFT: Small nitpick, but it seems that `if (nextForegroundColor != lastForegroundColor)` is slightly more correct improving readability

src/Microsoft.PowerShell.PSReadLine/ConsoleLib.cs
```diff
@@ -1068,17 +1068,33 @@ public void WriteBufferLines(BufferChar[] buffer, ref int top, bool ensureBottom
...
+                {
+                    Console.ForegroundColor = lastForegroundColor = nextForegroundColor;
+                }
+                if (lastBackgroundColor != nextBackgroundColor)
```
SteveL-MSFT: Same as above

src/Microsoft.PowerShell.PSReadLine/ConsoleLib.cs
```diff
@@ -1068,17 +1068,33 @@ public void WriteBufferLines(BufferChar[] buffer, ref int top, bool ensureBottom
...
+                {
+                    Console.BackgroundColor = lastBackgroundColor = nextBackgroundColor;
+                }
 
                 Console.Write(buffer[i].UnicodeChar);
             }
 
-            Console.BackgroundColor = backgroundColor;
-            Console.ForegroundColor = foregroundColor;
+            if (backgroundColor != lastBackgroundColor)
```
SteveL-MSFT: Pretty much everywhere else you have foreground first, then background.  Another nitpick, but I would reorder these two `if` statements

SteveL-MSFT (approved):

PR #3467, opened 2017-03-31
Fix Default/OEM encoding behavior PowerShell Core

iSazonov:
Based on conclusion in discussion #3248

- [x] - OEM

We need internal fix because .Net Core haven't "OEM" term.
The fix should provide PowerShell Core behavior:
   - on Windows - as Windows PowerShell based on GetOEMCP() and legacy encodings (System.Text.Encoding.CodePages)
   - on Unix - the same as default encoding in PowerShell Core

- [x] - Default

We need internal fix because [CoreFX default](https://github.com/dotnet/coreclr/blob/ab0fbce6b629d081c1102d353abab0d8febd4487/src/mscorlib/src/System/Text/Encoding.cs#L1244) for [all platforms is UTF8](https://github.com/dotnet/coreclr/blob/ab0fbce6b629d081c1102d353abab0d8febd4487/src/mscorlib/src/System/Text/Encoding.cs#L1235)
Until we have completed Encoding RFC we should restore behaivor as in Windows PowerShell.
The fix should provide PowerShell Core behavior:
   - on Windows - as Windows PowerShell based on GetACP() and legacy encodings (System.Text.Encoding.CodePages)
   - on Unix - UTF-8 without BOM (We are expecting that the same will be [in CoreFX](https://github.com/dotnet/coreclr/issues/10643))

- [ ] - Tests

Now we don't have a full set of encoding tests (only for redirections). I believe we need to create them during future RFC implementation, not now.

Issue #3248, opened 2017-03-03
Set-Content/Add-Content/Get-Content use an 8-bit character encoding by default, but the help topics state ASCII; problematic Core default file encoding


mklement0:
This issue has two distinct aspects:
* discussion of an existing _documentation bug_
* discussion of the problematic fixed default file encoding currently (alpha16) chosen for _Core_. 

Steps to reproduce
------------------

```powershell
'ö' | Set-Content -NoNewline -Encoding ASCII tmp.txt 
'ö' | Add-Content -Encoding ASCII -NoNewline tmp.txt 
Get-Content -Encoding ASCII tmp.txt
(Get-Content -Encoding Byte -TotalCount 2 tmp.txt) | % { '0x{0:x}' -f $_ }
'--'
'ö' | Set-Content -NoNewline tmp.txt   # use default encoding
'ö' | Add-Content -NoNewline tmp.txt   # use default encoding
Get-Content tmp.txt                    # use default encoding
(Get-Content -Encoding Byte -TotalCount 2 tmp.txt) | % { '0x{0:x}' -f $_ }
```

Expected behavior
-----------------

```
??
0x3f
0x3f
--
??
0x3f
0x3f
```

Actual behavior
---------------

```
??
0x3f
0x3f
--
öö
0xf6
0xf6
```

That is, ASCII encoding turns a non-ASCII character into _literal_ `?` (`0x3f`)

The fact that `Set-Content` without an `-Encoding` argument resulted in `ö` on reading implies that ASCII encoding wasn't used, and the specific byte value of `0xf6` further implies that that a single-byte, extended-ASCII encoding was used:

* For _Windows_ PowerShell, it is the _respective_ system's legacy codepage ("ANSI"), such as [Windows-1252](https://en.wikipedia.org/wiki/Windows-1252) on US-English systems, or [Windows-1251](https://en.wikipedia.org/wiki/Windows-1251) on Russian systems. In other words: the specific encoding is, to put it in Unix terms, _locale-dependent_.

* For PowerShell _Core_, _as of alpha 16_, it is [ISO-8859-1](https://en.wikipedia.org/wiki/ISO/IEC_8859-1), as @iSazonov helpfully points out (see [his comment below](https://github.com/PowerShell/PowerShell/issues/3248#issuecomment-284038694) for the source-code links).
   * **Using a _fixed_ encoding that is limited to 256 code points is problematic, however.**
[...]

PR #4119, opened 2017-06-27
Unify file encoding when a cmdlet creates a file

JamesWTruher:
Implementation of https://github.com/PowerShell/PowerShell-RFC/blob/master/1-Draft/RFC0020-DefaultFileEncoding.md

Cmdlets now create files with consistent encoding (UTF8 without BOM) on all platforms.
A new preference variable `PSDefaultFileEncoding` is now available to enable users to set encoding for cmdlets. By setting `$PSDefaultFileEncoding = "WindowsLegacy"` users can select the encoding which exists in PowerShell5 for each specific cmdlet.

Provider and cmdlet encoding methods have been centralized and are now common.


src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,387 @@
...
+    public static class PowerShellEncoding
+    {
+
+        /// <summary>
+        /// Return the default PowerShell encoding which is UTF8 without a BOM.
+        /// There is no distinction between platforms
+        /// </summary>
+        public static Encoding GetDefaultEncoding()
+        {
+            return new UTF8Encoding(false);
```
iSazonov: We can use static cache for `UTF8Encoding(false);` here and in Line 142 too - the cache is in Line 311.

JamesWTruher: good point

lzybkr:
Please fix the conflict and rebase.

iSazonov:
If the new test batch covers all the old encoding variants, it would be good to merge the tests before the PR - then we'll be sure to see that the PR has full backward compatibility.

src/System.Management.Automation/utils/PathUtils.cs
```diff
@@ -17,6 +18,42 @@ namespace System.Management.Automation
...
+            bool isLiteralPath
+            )
+        {
+            Encoding resolvedEncoding = PowerShellEncoding.GetEncoding(cmdlet, encoding);
+
+            MasterStreamOpen(cmdlet, filePath, resolvedEncoding, defaultEncoding, Append, Force, NoClobber, out fileStream, out streamWriter, out readOnlyFileInfo, isLiteralPath);
+        }
+
+        /*
         /// <summary>
```
adityapatwardhan: Delete the commented block?

src/System.Management.Automation/utils/PathUtils.cs
```diff
@@ -188,6 +226,13 @@ internal static void ReportFileOpenFailure(Cmdlet cmdlet, string filePath, Excep
             cmdlet.ThrowTerminatingError(errorRecord);
         }
 
+        internal static StreamReader OpenStreamReader(PSCmdlet command, string filePath, FileEncoding encoding, bool isLiteralPath)
+        {
+            FileStream fileStream = OpenFileStream(filePath, command, isLiteralPath);
+            return new StreamReader(fileStream, PowerShellEncoding.GetEncoding(command, encoding));
+        }
+
+        /*
```
adityapatwardhan: Delete commented block

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,387 @@
...
+                return FileEncoding.Byte;
+            }
+
+            // we couldn't determine anything from direct examination,
+            // return UTF8 without a BOM which should be good for both Windows and Non-Windows
+            return FileEncoding.UTF8NoBOM;
+        }
+    }
+
+}
```
adityapatwardhan: Can you check the code coverage with the newly added tests for this file? So we add tests now if we are missing any.

adityapatwardhan (changes requested):

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,387 @@
...
+                { "microsoft.powershell.commands.exportpssessioncommand", Encoding.UTF8 }, // with BOM
+                { "microsoft.powershell.commands.formathex", Encoding.ASCII },
+                { "microsoft.powershell.commands.newmodulemanifestcommand", Encoding.Unicode },
+                { "microsoft.powershell.commands.getcontentcommand", Encoding.ASCII },
+                { "microsoft.powershell.commands.importcsvcommand", Encoding.ASCII },
+                { "microsoft.powershell.commands.outfilecommand", Encoding.Unicode }, // This includes redirection
+                { "microsoft.powershell.commands.setcontentcommand", Encoding.ASCII },
+                // Providers are handled here
+                { "microsoft.powershell.commands.filesystemprovider", Encoding.ASCII },
+
```
JamesWTruher: fixed

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,387 @@
...
+            };
+
+        internal static Encoding GetWindowsLegacyEncoding(string name)
+        {
+            if ( legacyEncodingMap.ContainsKey(name))
```
JamesWTruher: ok

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,387 @@
...
+        }
+
+        /// <summary>
+        /// Retrieve the encoding in a provider context
+        /// </summary>
+        public static Encoding GetProviderEncoding(CmdletProvider provider, FileEncoding encoding)
+        {
+            Encoding resolvedEncoding = GetDefaultEncoding();
+            FileEncoding encodingPreference = GetEncodingPreference(provider.SessionState);
+            if ( encoding == FileEncoding.Unknown && encodingPreference != FileEncoding.Unknown )
```
JamesWTruher: yes - because of the provider cmdlets which leave the encoding to the underlying provider as a dynamic parameter it needs to be handled via a different path than checking against the name of the cmdlet.

src/System.Management.Automation/utils/Encoding.cs
JamesWTruher: now that we can, i will :)

src/System.Management.Automation/utils/PathUtils.cs
JamesWTruher: fixed - I missed that one

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -206,8 +207,7 @@ public SwitchParameter NoClobber
         /// Encoding optional flag
         /// </summary>
         [Parameter()]
-        [ValidateSetAttribute(new string[] { "Unicode", "UTF7", "UTF8", "ASCII", "UTF32", "BigEndianUnicode", "Default", "OEM" })]
-        public string Encoding { get; set; }
+        public FileEncoding Encoding { get; set; } = FileEncoding.Unknown;
```
dantraMSFT: Why aren't you using a ValidateSet with the FileEncoding values?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -571,8 +571,7 @@ public SwitchParameter UseCulture
         /// Encoding optional flag
         /// </summary>
         [Parameter()]
-        [ValidateSetAttribute(new[] { "Unicode", "UTF7", "UTF8", "ASCII", "UTF32", "BigEndianUnicode", "Default", "OEM" })]
-        public string Encoding { get; set; }
+        public FileEncoding Encoding { get; set; } = FileEncoding.Unknown;
```
dantraMSFT: As above, Why aren't you using a ValidateSet with the FileEncoding values

src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
```diff
@@ -1200,17 +1200,7 @@ public SwitchParameter AllMatches
...
-            EncodingConversion.Unicode,
-            EncodingConversion.Utf7,
-            EncodingConversion.Utf8,
-            EncodingConversion.Utf32,
-            EncodingConversion.Ascii,
-            EncodingConversion.BigEndianUnicode,
-            EncodingConversion.Default,
-            EncodingConversion.OEM })]
-        public string Encoding { get; set; }
+        public FileEncoding Encoding { get; set; } = FileEncoding.Unknown;
```
dantraMSFT: The default should be FileEncoding.UTF8. See BeginProcessing.

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,387 @@
...
+            }
+            return encodingPreference;
+        }
+
+        /// <summary>
+        /// Retrieve the encoding in a provider context
+        /// </summary>
+        public static Encoding GetProviderEncoding(CmdletProvider provider, FileEncoding encoding)
+        {
+            Encoding resolvedEncoding = GetDefaultEncoding();
```
dantraMSFT: Suggest checking for encoding != FileEncoding.Unknown first; none of the other code paths are executed for that case.

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,387 @@
...
+                {
+                    resolvedEncoding = GetEncoding(encoding);
+                }
+            }
+            else
+            {
+                // the parameter is not specifically set, so check the preference variable
+                encodingPreference = GetEncodingPreference(cmdlet.Context.SessionState);
+                // If set to unknown, we accept that it is unset
+                preferenceSetAndValid = encodingPreference != FileEncoding.Unknown;
```
dantraMSFT: dead code?

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,387 @@
...
+        /// <summary>
+        /// Retrieve the encoding based on the Cmdlet and the Encoding
+        /// <param name="cmdlet">The cmdlet of interest</param>
+        /// <param name="encoding">The Encoding parameter value</param>
+        /// </summary>
+        public static Encoding GetEncoding(Cmdlet cmdlet, FileEncoding encoding)
+        {
+            Encoding resolvedEncoding = GetDefaultEncoding();
+            FileEncoding encodingPreference = FileEncoding.Unknown;
+            bool preferenceSetAndValid = false;
```
dantraMSFT: dead code?

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,387 @@
...
+            byte[] initialBytes = new byte[100];
+            int bytesRead = 0;
+
+            try
+            {
+                using (FileStream stream = System.IO.File.OpenRead(path))
+                {
+                    using (BinaryReader reader = new BinaryReader(stream))
+                    {
+                        bytesRead = reader.Read(initialBytes, 0, 100);
```
dantraMSFT: use initialBytes.Length instead of 100.

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,387 @@
...
+
+        // take a look at the file contents and guess at the best encoding
+        internal static FileEncoding GetEncoding(string path)
+        {
+            if (!File.Exists(path))
+            {
+                return FileEncoding.Default;
+            }
+
+            byte[] initialBytes = new byte[100];
```
dantraMSFT: Do you really need to read 100 bytes?

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,387 @@
...
+        internal static char[] nonPrintableCharacters = {
+            (char) 0, (char) 1, (char) 2, (char) 3, (char) 4, (char) 5, (char) 6, (char) 7, (char) 8,
+            (char) 11, (char) 12, (char) 14, (char) 15, (char) 16, (char) 17, (char) 18, (char) 19, (char) 20,
+            (char) 21, (char) 22, (char) 23, (char) 24, (char) 25, (char) 26, (char) 28, (char) 29, (char) 30,
+            (char) 31, (char) 127, (char) 129, (char) 141, (char) 143, (char) 144, (char) 157 };
+
+        internal static readonly UTF8Encoding utf8NoBom = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
+
+        // take a look at the file contents and guess at the best encoding
+        internal static FileEncoding GetEncoding(string path)
```
dantraMSFT: Suggest naming this GetFileEncoding for clarity.

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,387 @@
...
+        {
+            if (!File.Exists(path))
+            {
+                return FileEncoding.Default;
```
dantraMSFT: I think this should probably throw; not return Default.

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,387 @@
...
+            }
+
+            // Test for four-byte preambles
+            string preamble = null;
+            FileEncoding foundEncoding = FileEncoding.Default;
+
+            if (bytesRead > 3)
+            {
+                preamble = String.Join("-", initialBytes[0], initialBytes[1], initialBytes[2], initialBytes[3]);
```
dantraMSFT: I suspect the string building/lookups appears overkill.  Consider testing the bytes directly?

src/System.Management.Automation/commands/utility/FormatAndOutput/common/FormatXMLWriter.cs
```diff
@@ -37,7 +38,7 @@ internal class FormatXmlWriter
             StreamWriter streamWriter;
             FileStream fileStream;
             FileInfo fileInfo;
-            PathUtils.MasterStreamOpen(cmdlet, filepath, "ascii", true, false, force, noclobber,
+            PathUtils.MasterStreamOpen(cmdlet, filepath, PowerShellEncoding.GetEncoding(FileEncoding.Ascii), true, false, force, noclobber,
```
dantraMSFT: This appears to be the only external call to Encoding overload of PathUtils.MasterStreamOpen.  Suggest using FileEncoding.Ascii here and merge the two MasterStreamOpen overloads.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
iSazonov: We could use #3784 - dynamic validate set, don't expose `FileEncoding` and  allow enhancement to use legacy charsets.

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -4824,6 +4825,7 @@ .ForwardHelpCategory Cmdlet
         internal const ActionPreference defaultVerbosePreference = ActionPreference.SilentlyContinue;
         internal const ActionPreference defaultWarningPreference = ActionPreference.Continue;
         internal const ActionPreference defaultInformationPreference = ActionPreference.SilentlyContinue;
+        internal const Microsoft.PowerShell.FileEncoding defaultFileEncodingPreference = FileEncoding.Unknown;
```
dantraMSFT: I can find no references to this field other than setting the session state variable. the variable. How is it used and how does it control behavior? Currently, it appears to be a NOP.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -8,6 +8,7 @@
 using System.Management.Automation.Internal;
 using System.Management.Automation.Security;
 using System.Reflection;
+using Microsoft.PowerShell;
```
dantraMSFT: Is this actually needed? The only other change to this file is the removal of the GetEncoding logic.

src/System.Management.Automation/engine/Modules/ScriptAnalysis.cs
```diff
@@ -95,7 +96,7 @@ internal static string ReadScript(string path)
         {
             using (FileStream readerStream = new FileStream(path, FileMode.Open, FileAccess.Read))
             {
-                Encoding defaultEncoding = ClrFacade.GetDefaultEncoding();
+                Encoding defaultEncoding = PowerShellEncoding.GetDefaultEncoding();
```
dantraMSFT: Shouldn't this be using PowerShellEncoding.GetEncoding

src/System.Management.Automation/utils/ClrFacade.cs
```diff
@@ -209,7 +210,7 @@ internal static Encoding GetOEMEncoding()
             if (s_oemEncoding == null)
             {
 #if UNIX        // PowerShell Core on Unix
-                s_oemEncoding = GetDefaultEncoding();
+                s_oemEncoding = PowerShellEncoding.GetDefaultEncoding();
 #elif CORECLR   // PowerShell Core on Windows
```
dantraMSFT: I think the CORECLR branches could use commenting in both GetDefaultEncoding and GetOEMEncoding. The logic reasoning for GetACP and GetOEMCP is opaque.

src/System.Management.Automation/utils/ClrFacade.cs
```diff
@@ -355,6 +356,7 @@ private static SecurityZone ReadFromZoneIdentifierDataStream(string filePath)
                                             FileAccess.Read, FileShare.Read);
 
                 // If we successfully get the zone data stream, try to read the ZoneId information
+                // use the method in this class not PowerShellEncoding.
```
dantraMSFT: Why  isn't this detecting the encoding of the stream data?

src/System.Management.Automation/utils/PathUtils.cs
```diff
@@ -17,6 +18,42 @@ namespace System.Management.Automation
...
+        /// <param name="encoding">encoding (this method will convert the command line string to an Encoding instance)</param>
+        /// <param name="defaultEncoding">if <c>true</c>, then we will use default .NET encoding instead of the encoding specified in <paramref name="encoding"/> parameter</param>
+        /// <param name="Append"></param>
+        /// <param name="Force"></param>
+        /// <param name="NoClobber"></param>
+        /// <param name="fileStream">Result1: <see cref="FileStream"/> opened for writing</param>
+        /// <param name="streamWriter">Result2: <see cref="StreamWriter"/> (inherits from <see cref="TextWriter"/>) opened for writing</param>
+        /// <param name="readOnlyFileInfo">Result3: file info that should be used to restore file attributes after done with the file (<c>null</c> is this is not needed)</param>
+        /// <param name="isLiteralPath">True if wildcard expansion should be bypassed.</param>
+        internal static void MasterStreamOpen(
```
dantraMSFT: Suggest merging this with the Encoding overload. There appears to be only 1 caller to the later overload and it starts converts a FileEncoding to an Encoding to accomplish it.

src/System.Management.Automation/utils/PathUtils.cs
```diff
@@ -490,7 +537,7 @@ internal static Encoding Convert(Cmdlet cmdlet, string encoding)
                 return System.Text.Encoding.UTF32;
 
             if (string.Equals(encoding, Default, StringComparison.OrdinalIgnoreCase))
-                return ClrFacade.GetDefaultEncoding();
+                return PowerShellEncoding.GetDefaultEncoding();
```
dantraMSFT: Suggest using a dictionary for these cascading 'if' tests.

src/System.Management.Automation/utils/Encoding.cs
dantraMSFT: Dead code.

src/System.Management.Automation/utils/Encoding.cs
dantraMSFT: Dead code.

src/System.Management.Automation/utils/ClrFacade.cs
iSazonov: The discussion was in #3248 and fixed in #3467. Maybe it should be reviewed again. In short we returned behavior Windows PowerShell.

test/powershell/engine/Encoding.Tests.ps1
```diff
@@ -0,0 +1,173 @@
...
+            (Get-Content $file -Encoding byte | Select-Object -First $count) -Join "-"
+        }
+
+        function Get-NewLineBytes
+        {
+            param ( [Microsoft.PowerShell.FileEncoding]$encoding )
+            $encoder = [Microsoft.PowerShell.PowerShellEncoding]::GetEncoding($encoding)
+            $encoder.GetBytes([Environment]::NewLine) -Join "-"
+        }
+        function Test-GetEncoding
```
dantraMSFT: This function is not referenced.

test/powershell/engine/Encoding.Tests.ps1
```diff
@@ -0,0 +1,173 @@
...
+            param (
+                [Microsoft.PowerShell.FileEncoding]$Encoding
+            )
+            END {
+                [Microsoft.PowerShell.PowerShellEncoding]::GetCmdletEncoding($pscmdlet, $encoding)
+            }
+        }
+
+        $preambleTests =
+            @{ Name = 'Ascii'; Preamble = '' },
```
dantraMSFT: The values that have an empty Preamble mask possible bugs and unexpected preamble may be present but it won't be detected resulting in a false positive. For example, the test won't detect that a UTF32 encoding was encountered when ASCII was expected.
I suggest coding the test data to have a well known sequence following the BOM (or lack of BOM) and have the Preamble strings include this sequence. This will help reduce potential false positives.

test/powershell/engine/Encoding.Tests.ps1
```diff
@@ -0,0 +1,173 @@
...
+            @{ Name = 'UTF7'; Preamble = '' },
+            @{ Name = 'UTF8'; Preamble = '' },
+            @{ Name = 'UTF8BOM'; Preamble = '239-187-191' },
+            @{ Name = 'UTF8NoBOM'; Preamble = '' },
+            @{ Name = 'Unicode'; Preamble = '255-254' },
+            @{ Name = 'Unknown'; Preamble = '' },
+            @{ Name = 'WindowsLegacy'; Preamble = '' }
+
+        $contentTests =
+            @{ Name = 'Ascii'; Bytes = "116-63-115-116-" + (Get-NewLineBytes Ascii) },
```
dantraMSFT: Minor nit; I suggest using Encoding instead of Name.

test/powershell/engine/Encoding.Tests.ps1
```diff
@@ -0,0 +1,173 @@
...
+        $PSDefaultFileEncoding = $Name
+        $testString | out-file $testFile
+        Get-FileBytes $testFile | should be $Bytes
+    }
+
+    It "Explicit encoding is not overridden by setting PSDefaultFileEncoding to '<Name>'" -TestCase $contentTests {
+        param ( $Name, $Bytes )
+        $PSDefaultFileEncoding = $Name
+        $testString | out-file -encoding ascii $testFile
+        Get-FileBytes $testFile | should be ("116-63-115-116-" + (Get-NewLineBytes ASCII))
```
dantraMSFT: A brief comment explaining why this is handled specially with the leading bytes spelled out would be useful for this test and subsequent tests.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
JamesWTruher: I don't want to use validateset here at all. Especially since this is so pervasive in so many areas, I get the same more easily with an enum

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
JamesWTruher: same as above

src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
JamesWTruher: i want to be sure that I discover the encoding. If I set it to something that is likely to be provided, I won't be able to determine whether the user explicitly set it. By setting it to unknown, I'll calculate the appropriate value and I can tell that it wasn't set by the user

src/System.Management.Automation/utils/Encoding.cs
JamesWTruher: fixed

src/System.Management.Automation/utils/Encoding.cs
JamesWTruher: this was part of a refactor where this code has been around for a long time, and I didn't really want to change it. I'm certainly open reducing the size, but it's not too much, not too little and any number here would be a guess anyway, right?

src/System.Management.Automation/utils/Encoding.cs
JamesWTruher: fixed

test/powershell/engine/Encoding.Tests.ps1
JamesWTruher: will add

JamesWTruher: i decided to refactor these so I could avoid hardcoding these strings more than once

test/powershell/engine/Encoding.Tests.ps1
JamesWTruher: will change

test/powershell/engine/Encoding.Tests.ps1
JamesWTruher: removed

src/System.Management.Automation/utils/Encoding.cs
JamesWTruher: removed

test/powershell/engine/Encoding.Tests.ps1
JamesWTruher: removed

src/System.Management.Automation/utils/ClrFacade.cs
JamesWTruher: I'll add comments about this, we can change it later after we have live-fire feedback.

src/System.Management.Automation/engine/InitialSessionState.cs
JamesWTruher: it's used down in line 4915 where it essentially lays claim to the variable

src/System.Management.Automation/commands/utility/FormatAndOutput/common/FormatXMLWriter.cs
JamesWTruher: yah - I was worried about removing the overload as it was a public interface and thought to do less harm, just in case someone else was using it somehow.

src/System.Management.Automation/utils/Encoding.cs
JamesWTruher: this was code which was refactored from another location, I didn't really want to change it

src/System.Management.Automation/utils/Encoding.cs
JamesWTruher: not necessarily - the file could be later created, so I need to pass back something - this code was refactored from another location, so i wanted to do as little as possible

src/System.Management.Automation/commands/utility/FormatAndOutput/common/FormatXMLWriter.cs
dantraMSFT: MasterStreamOpen is internal.

JamesWTruher: I'll investigate

JamesWTruher: the encoding overload is used by CSV cmdlets which do their own determination of the encoding (in the append case), and pass that along to the open. I'd rather not refactor all that code if I don't need to.

src/System.Management.Automation/engine/Utils.cs
JamesWTruher: removed

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -381,6 +298,101 @@ internal static FileEncoding GetEncoding(string path)
...
+                { "microsoft.powershell.commands.outfilecommand", Encoding.Unicode }, // This includes redirection
+                { "microsoft.powershell.commands.setcontentcommand", Encoding.ASCII },
+                // Providers are handled here
+                { "microsoft.powershell.commands.filesystemprovider", Encoding.ASCII },
+            };
+
+        /// Get the Windows legacy encoding from our encoding map
+        internal static Encoding GetWindowsLegacyEncoding(string name)
+        {
+            if ( legacyEncodingMap.ContainsKey(name))
```
iSazonov: Can we use `return TryGetValue(name, out Encoding value) ? value : Encoding.Default`?

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -151,7 +152,9 @@ public static Encoding GetEncoding(FileEncoding TextEncoding)
                     break;
 
                 case FileEncoding.BigEndianUTF32:
-                    result = Encoding.BigEndianUnicode;
+                    // This can possibly throw, but if so, we can't provide
+                    // the encoding which the user requested, so we should fail
+                    result = Encoding.GetEncoding("utf-32BE");
```
iSazonov: Should we cache this in  internal static readonly constant as for `internal static readonly UTF8Encoding utf8NoBom`?

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -381,6 +298,101 @@ internal static FileEncoding GetEncoding(string path)
...
+
+        /// <summary>
+        /// Retrieve the PSDefaultFileEncoding preference value if set
+        /// <summary>
+        public static FileEncoding GetEncodingPreference(SessionState sessionState)
+        {
+            FileEncoding encodingPreference = FileEncoding.Unknown;
+            try
+            {
+                // It doesn't matter if this fails or throws, we will return unknown in that case
```
iSazonov: Please replace `unknown` on 'FileEncoding.Unknown'.

src/System.Management.Automation/utils/Encoding.cs
iSazonov: Ah, we can use `GetValueOrDefault`
```c#
return legacyEncodingMap.GetValueOrDefault(name, Encoding.Default);
```

JamesWTruher:
@adityapatwardhan code coverage with latest tests is about 98%

adityapatwardhan (approved):
LGTM

test/powershell/engine/Encoding.Tests.ps1
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    It "Explicit encoding set to unknown and preference variable set to unicode creates unicode file" {
+        $PSDefaultFileEncoding = "Unicode"
+        $testString | set-content -encoding unknown $testfile
+        Get-FileBytes $testFile | should be $testStringEncodedBytes['Unicode']
+    }
+
+    It "getting the encoding for an unknown cmdlet should return utf-8" {
+        $method = [microsoft.powershell.powershellencoding].getmember("GetWindowsLegacyEncoding","NonPublic,Static")
```
lzybkr: We'd like to avoid reflection as much as possible in tests - can you find user accessible way to test this instead, e.g. calling a public api that calls this internal api?

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,399 @@
...
+                object tmp = sessionState.PSVariable.GetValue("PSDefaultFileEncoding");
```
lzybkr: You should probably also test `sessionState != null` to avoid the expensive exception because I can see this method getting called with a null session state.

src/System.Management.Automation/engine/SpecialVariables.cs
```diff
@@ -289,6 +293,7 @@ internal static class SpecialVariables
                                                                     /* WarningPreference */     typeof(ActionPreference),
                                                                     /* InformationPreference */ typeof(ActionPreference),
                                                                     /* ConfirmPreference */     typeof(ConfirmImpact),
+                                                                    /* PSDefaultFileEncoding */ typeof(Microsoft.PowerShell.FileEncoding),
```
lzybkr: This line should be removed. I realize it's not obvious why this array exists, but it is used for variables that are always created, and these preference variables get created in every script cmdlet, but your new preference variable should not.

src/System.Management.Automation/namespaces/FileSystemContentStream.cs
```diff
@@ -1160,8 +1161,10 @@ internal FileStreamBackReader(FileStream fileStream, Encoding encoding)
                 _currentPosition = _stream.Position;
 
                 // Get the oem encoding and system current ANSI code page
-                _oemEncoding = EncodingConversion.Convert(null, EncodingConversion.OEM);
-                _defaultAnsiEncoding = EncodingConversion.Convert(null, EncodingConversion.Default);
+                // _oemEncoding = EncodingConversion.Convert(null, EncodingConversion.OEM);
```
lzybkr: Delete?

lzybkr (changes requested):
The @PowerShell/powershell-committee should probably review the new public apis.

src/System.Management.Automation/namespaces/FileSystemContentStream.cs
JamesWTruher: fixed

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,399 @@
...
+        /// <summary>
+        /// Windows legacy encoding. This requires a cmdlet object to resolve.
+        /// </summary>
+        WindowsLegacy,
+    }
+
+    /// <summary>
+    /// the helper class for determining encodings for PowerShell
+    /// </summary>
+    public static class PowerShellEncoding
```
JamesWTruher: Please suggest something

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,399 @@
...
+        {
+            return utf8NoBom;
+        }
+
+        /// <summary>
+        /// translate a FileEncoding to an actual System.Text.Encoding
+        /// <param name="TextEncoding">The enum value</param>
+        /// <returns>System.Text.Encoding</returns>
+        /// </summary>
+        public static Encoding GetEncoding(FileEncoding TextEncoding)
```
JamesWTruher: fixed

JamesWTruher: i see this as an opportunity to return the default encoding (utf8NoBom) - making this an error doesn't seem to be needed

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,399 @@
...
+        {
+            System.Text.Encoding result = GetDefaultEncoding();
```
JamesWTruher: fixed

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,399 @@
...
+            return result;
+        }
+
+        /// <summary>
+        /// Retrieve the encoding based on the Cmdlet and the Encoding
+        /// <param name="cmdlet">The cmdlet of interest</param>
+        /// <param name="encoding">The Encoding parameter value</param>
+        /// <returns>System.Text.Encoding</returns>
+        /// </summary>
+        public static Encoding GetEncoding(Cmdlet cmdlet, FileEncoding encoding)
```
JamesWTruher: i fixed the public api parameter validation but I disagree about the throw - if I don't have access to session state, I can't retrieve the value I can still return that no encoding preference has been set

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,399 @@
...
+        {
+            Encoding resolvedEncoding = GetDefaultEncoding();
+            FileEncoding encodingPreference = FileEncoding.Unknown;
+            bool preferenceSetAndValid = false;
+            string name = cmdlet.GetType().FullName.ToLower(CultureInfo.InvariantCulture);
```
JamesWTruher: yah, especially since the dictionary is a case insensitive one

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,399 @@
...
+
+            return resolvedEncoding;
+        }
+
+        /// <summary>
+        /// Given a path to a file, attempt to retrieve the encoding
+        /// <param name="path">The path to a file to inspect for an encoding</param>
+        /// <returns>System.Text.Encoding</returns>
+        /// </summary>
+        public static FileEncoding GetFileEncodingFromFile(string path)
```
JamesWTruher: concrete suggestion please? This code has existed for some time (although not public) and has been used for a long time

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,399 @@
...
+                }
+            }
+            catch (IOException)
+            {
+                return FileEncoding.Default;
+            }
+
+            // Test for four-byte preambles
+            string preamble = null;
+            FileEncoding foundEncoding = FileEncoding.Default;
```
JamesWTruher: fixed

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,399 @@
...
+                resolvedEncoding = GetEncoding(encoding);
+            }
+            return resolvedEncoding;
+        }
+
+        // This is the way the encoding is implemented in PowerShell 5 and earlier.
+        // If the user sets the default encoding to WindowsLegacy, we will
+        // be able to encode for that
+        internal static Dictionary<String, Encoding> legacyEncodingMap =
+            new Dictionary<string, Encoding>(StringComparer.OrdinalIgnoreCase)
```
JamesWTruher: i didn't want to use that because this means that those types must be known in the engine, which I don't believe is a good thing.

JamesWTruher: also, with regard to the test, pester doesn't do array comparisons very well. This is a reasonable approach for that problem

test/powershell/engine/Encoding.Tests.ps1
JamesWTruher: this is pretty difficult as the API that is public takes a cmdlet object, so i couldn't think of any other way to do this. I'm happy to hear suggestions

src/System.Management.Automation/engine/SpecialVariables.cs
JamesWTruher: fixed

test/powershell/Language/Parser/RedirectionOperator.Tests.ps1
```diff
@@ -29,22 +29,23 @@ Describe "Redirection operator now supports encoding changes" -Tags "CI" {
...
-        $BOM = [text.encoding]::unicode.GetPreamble()
-        $TXT = [text.encoding]::unicode.GetBytes($asciiString)
-        $CR  = [text.encoding]::unicode.GetBytes($asciiCR)
+        $BOM = $encoder.GetPreamble()
+        $TXT = $encoder.GetBytes($asciiString)
+        $CR  = $encoder.GetBytes($asciiCR)
         $expectedBytes = .{ $BOM; $TXT; $CR }
         $bytes.Count | should be $expectedBytes.count
         for($i = 0; $i -lt $bytes.count; $i++) {
             $bytes[$i] | Should be $expectedBytes[$i]
```
JamesWTruher: fixed

src/System.Management.Automation/utils/Encoding.cs
lzybkr: Don't make it public, or do a ton more testing and fix the issues I raised.

src/System.Management.Automation/utils/Encoding.cs
lzybkr: We don't want to introduce a new circular dependency, but we have implicit knowledge encoded here, either via a string or `typeof` - there is little difference.

Actually removing the implicit knowledge probably isn't worth it, but you could do that via registration - each of these cmdlets would call an internal api to register itself.

test/powershell/engine/Encoding.Tests.ps1
lzybkr: What's wrong with `[cmdlet]::new()` to get that instance?

src/System.Management.Automation/utils/Encoding.cs
lzybkr: Shouldn't I be able to use this api to get an Encoding for a file I've written using `FileEncoding`? That will work for everything but `WindowsLegacy`, so I would say it's better to throw.

src/System.Management.Automation/utils/Encoding.cs
lzybkr: EncodingUtils or something like that?

src/System.Management.Automation/utils/Encoding.cs
JamesWTruher: your example also misses the UTF8 without BOM case and looks return the wrong encoding (ASCII) in the case that it encountered such a file.

lzybkr: Yeah - my sample served it's purpose and nothing more - but I mentioned it to show it's not much code.

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,399 @@
...
+{
+    /// <summary>
+    /// The list of available file encodings
+    /// </summary>
+    public enum FileEncoding
+    {
+        /// <summary>
+        /// No encoding, or unset.
+        /// </summary>
+        Unknown,
```
JamesWTruher: how would this be set back to -1, once set? would the user be stuck with the value?

lzybkr: Internally, we would have a const, similar to what the CLR does for colors (see [here](https://github.com/dotnet/corefx/blob/ef82e74c065a6f839515fee8871e1114e6aac357/src/System.Console/src/System/Console.cs#L170)).

I don't quite follow your question - but in C#, you can assign any value you like (with a cast). In PowerShell it's not as easy because we check an integer value to see if it's a valid enumerator.

We could also consider using the value `0` to mean `Unspecified` as long as the first enum starts at 1 - this has the benefit of being the default value if a field is not assigned.

src/System.Management.Automation/utils/Encoding.cs
JamesWTruher: changed to EncodingUtils

src/System.Management.Automation/utils/Encoding.cs
JamesWTruher: changed to internal

src/System.Management.Automation/utils/Encoding.cs
JamesWTruher: Text.Encoding.UTF7.GetPreamble() returns no bytes

test/powershell/engine/Encoding.Tests.ps1
```diff
@@ -0,0 +1,225 @@
...
+        $preamble = @{
+            Ascii = ''
+            BigEndianUTF32 = '254-255'
+            BigEndianUnicode = '254-255'
+            Byte = '255-254'
+            Default = ''
+            Oem = ''
+            String = '255-254'
+            UTF32 = '255-254-0-0'
+            UTF7 = ''
```
JamesWTruher: Text.Encoding.UTF7.GetPreamble() returns no bytes

lzybkr: I understand (maybe the api designer didn't think it was worth returning an array of possible preambles, and the implementer decided it was better to return nothing rather than choosing one possible preamble.)

Regardless, it's worth a comment on why utf7 is missing an expected bom, after all, we aren't calling it UTF7NoBOM.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and we should have an RFC specific to the new public apis (and generally any new public apis should be RFCs going forward)

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,402 @@
...
+
+            return resolvedEncoding;
+        }
+
+        /// <summary>
+        /// Given a path to a file, attempt to retrieve the encoding
+        /// <param name="path">The path to a file to inspect for an encoding</param>
+        /// <returns>System.Text.Encoding</returns>
+        /// </summary>
+        internal static FileEncoding GetFileEncodingFromFile(string path)
```
SteveL-MSFT: Seems like this should be `GetEncodingFromFile` to keep the naming consistent

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,402 @@
...
+        /// <summary>
+        /// Retrieve the PSDefaultFileEncoding preference value if set
+        /// <param name="sessionState">SessionState to use to retrieve the preference variable if set</param>
+        /// <summary>
+        public static FileEncoding GetEncodingPreference(SessionState sessionState)
+        {
+            FileEncoding encodingPreference = FileEncoding.Unspecified;
+            try
+            {
+                // It doesn't matter if this fails or throws, we will return unknown in that case
```
lzybkr: return Unspecified, not return unknown

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,402 @@
...
+            return result;
+        }
+
+        /// <summary>
+        /// Retrieve the encoding based on the Cmdlet and the Encoding
+        /// <param name="cmdlet">The cmdlet of interest</param>
+        /// <param name="encoding">The Encoding parameter value</param>
+        /// <returns>System.Text.Encoding</returns>
+        /// </summary>
+        public static Encoding GetEncoding(Cmdlet cmdlet, FileEncoding encoding)
```
lzybkr: Maybe this api would be better if we passed a `SessionState` instead of `Cmdlet`.

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,402 @@
...
+            new Dictionary<string, FileEncoding>()
+            {
+                { "255-254", FileEncoding.Unicode },
+                { "254-255", FileEncoding.BigEndianUnicode },
+                { "255-254-0-0", FileEncoding.UTF32 },
+                { "0-0-254-255", FileEncoding.BigEndianUTF32 },
+                { "239-187-191", FileEncoding.UTF8BOM },
+            };
+
+        internal static char[] nonPrintableCharacters = {
```
mklement0: Strictly speaking, `11` (vertical tab) and `12` (form feed) are also printable characters; given their rarity, I assume they were left out intentionally; if so, perhaps a comment would appease drive-by pedants like me.

src/System.Management.Automation/utils/ClrFacade.cs
```diff
@@ -126,15 +127,20 @@ internal static Encoding GetDefaultEncoding()
...
+            // be installed.
             if (s_oemEncoding == null)
             {
 #if UNIX        // PowerShell Core on Unix
                 s_oemEncoding = GetDefaultEncoding();
-#else           // PowerShell Core on Windows
+#else           // PowerShell Core on Windows, which needs provider registration
                 EncodingRegisterProvider();
 
                 uint oemCp = NativeMethods.GetOEMCP();
```
mklement0: Looks like this line is no longer needed.

src/System.Management.Automation/utils/ClrFacade.cs
```diff
@@ -126,15 +127,20 @@ internal static Encoding GetDefaultEncoding()
         /// </summary>
         internal static Encoding GetOEMEncoding()
         {
+            // The OEM code pages are sometimes used by Win32 console applications, and
+            // on non-Windows platforms they still may have uses (if installed) and
+            // could be used if desired.
+            // On non-windows platforms, they have more limited uses, and probably won't
```
mklement0: This sentence is a bit confusing. Doesn't the previous one say everything that needs to be said?

src/System.Management.Automation/utils/Encoding.cs
```diff
@@ -0,0 +1,402 @@
...
+                { "microsoft.powershell.commands.exportpssessioncommand", Encoding.UTF8 }, // with BOM
+                { "microsoft.powershell.commands.formathex", Encoding.ASCII },
+                { "microsoft.powershell.commands.newmodulemanifestcommand", Encoding.Unicode },
+                { "microsoft.powershell.commands.getcontentcommand", Encoding.ASCII },
+                { "microsoft.powershell.commands.importcsvcommand", Encoding.ASCII },
+                { "microsoft.powershell.commands.outfilecommand", Encoding.Unicode }, // This includes redirection
+                { "microsoft.powershell.commands.setcontentcommand", Encoding.ASCII },
+                // Providers are handled here
+                { "microsoft.powershell.commands.filesystemprovider", Encoding.ASCII },
+            };
```
mklement0: Unfortunately, that doesn't appear to be correct.

As for the _writing_ cmdlets:

`Add-Content` and `Set-Content` - despite what the documentation claims - have always used "ANSI" encoding, not ASCII - see https://github.com/PowerShell/PowerShell-Docs/issues/1483.

What about `Send-MailMessage`? The help topics that ASCII encoding is the default.

`New-Item -Type File -Value` currently creates BOM-less(!) UTF-8.

As an aside: While `Export-Csv` without `-Append` indeed creates ASCII files, `Export-Csv -Append` currently blindly appends _UTF-8_ if the existing file's encoding is any of ASCII/UTF-8/"ANSI", but correctly matches UTF-16LE and UTF-16BE.

---

As for the _reading_ cmdlets:

`Get-Content` currently defaults to "ANSI" in the absence of a BOM, as does `Import-PowerShellDataFile`.

`Import-Csv` currently actually assumes UTF-8 in the absence of a BOM, as do `Import-CliXml` and `Select-String`

JamesWTruher:
i am going to take a different route for this change - closing this

Issue #4122, opened 2017-06-27
SSH Remoting session hangs if SSH client process abruptly terminates


PaulHigin:
If the SSH client process that PowerShell is using for the SSH transport terminates abruptly the StreamReader will return null instead of closing the pipe for a normal process exit.

The current error stream reading code ignores null StreamReader values resulting in a hang where the remote session never ends.

Steps to reproduce
------------------

```powershell
PS > New-PSSession -HostName <cnName>
PS > Get-Process SSH | kill -force
```

Expected behavior
-----------------

```none
Session state goes to "Broken" with error message
```

Actual behavior
---------------

```none
Session is broken and hangs on the client end
```

Environment data
----------------

```powershell
Current build
```

PR #4123, opened 2017-06-27
Fix for SSH remoting when SSH client abruptly terminates

PaulHigin:
This PR is for Issue #4122

If the SSH client process that PowerShell is using for the SSH transport terminates abruptly the StreamReader will return null instead of closing the pipe for a normal process exit.

The current error stream reading code ignores null StreamReader values resulting in a hang where the remote session never ends.

Fix is to throw an error when this occurs.

### DIFF
--- a/src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
+++ b/src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
@@ -1514,7 +1514,7 @@ namespace System.Management.Automation.Remoting.Client
                     if (error == null)
                     {
                         // Stream is closed unexpectedly.
-                        throw new PSInvalidOperationException(RemotingErrorIdStrings.SSHTerminated);
+                        throw new PSInvalidOperationException(RemotingErrorIdStrings.SSHAbruptlyTerminated);
                     }
 
                     if (error.Length == 0)
--- a/src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
+++ b/src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
@@ -1514,7 +1514,7 @@ namespace System.Management.Automation.Remoting.Client
                     if (error == null)
                     {
                         // Stream is closed unexpectedly.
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/resources/RemotingErrorIdStrings.resx
```diff
@@ -1636,4 +1636,7 @@ All WinRM sessions connected to Windows PowerShell session configurations, such
   <data name="InvalidRoleCapabilityFileExtension" xml:space="preserve">
     <value>The provided role capability file {0} does not have the required .psrc extension.</value>
   </data>
+  <data name="SSHTerminated" >
```
iSazonov: Maybe SSHabruptlyTerminated ?

PaulHigin: Sure, will change.

src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
```diff
@@ -1548,7 +1559,12 @@ private static string ReadError(StreamReader reader)
             // Blocking read from StdError stream
             string error = reader.ReadLine();
 
-            if (string.IsNullOrEmpty(error) ||
+            if (error == null)
+            {
+                return error;
```
mirichmo: Why did you decide to return null here and throw in the calling function? Why not just throw here and make the ReadError() function always return non-null strings?

mirichmo: Why did you decide to return null here and throw in the calling function? Why not just throw here and make the ReadError() function always return non-null strings?

PaulHigin: No special reason except that I think of ReadError() helper method as a wrapper to StreamReader.  But I agree that it would be cleaner to just throw in ReadError()

mirichmo (approved):

mirichmo:
@iSazonov Do you have any additional comments or concerns?

iSazonov:
LGTM.

PR #4132, opened 2017-06-28
Enable Windows PowerShell PSModulePath by default on Windows

### LINKED ISSUE #4056: PSCore6 should have a way to use Windows PowerShell PSModulePath
With moving to .Net Std 2.0 and the PR to enable PSCore6 to load assemblies from the GAC, the missing piece to enable PSCore6 to replace the need for Windows PowerShell (majority of scenarios anyways) is to make it simpler for users to find and use in-box Windows modules.  Currently, PSCore6 has no knowledge of the legacy Windows PowerShell PSModulePath.  Since we know we don't have 100% compatibility with in-box modules (PSSnapIns, for example are not supported at all), it seems we should consider an opt-in approach.  Alternatively, maybe we should have a `Windows Compat` module that has a whitelist of validated modules that work with PSCore6?

cc @joeyaiello @HemantMahawar

SteveL-MSFT:
Now that we have moved to .Net Std 2.0 compatible CoreClr 2.0 and added capability to search GAC for assemblies, we need to make it easier for users to use Windows PowerShell modules from PSCore6 to get feedback on compatibility.

This change is only for Windows and appends the Windows PowerShell PSModulePath on startup.  Depending on the data/feedback we get, we can decide what to do (opt-in vs opt-out) as we get closer to a release candidate.

Out-Default tests has to be updated to suppress the Warning message by not loading the profile.

Related #4056 

Based on the customer data, we can decide a more permanent solution.


build.psm1
```diff
@@ -196,7 +196,7 @@ function Start-PSBuild {
         Write-Warning @"
 The currently installed .NET Command Line Tools is not the required version.
 
-Installed version: $dotnetCLIIntalledVersion 
+Installed version: $dotnetCLIIntalledVersion
```
adityapatwardhan: Typo in Installed.

src/powershell-win-core/Microsoft.PowerShell_profile.ps1
```diff
@@ -0,0 +1,10 @@
+# Default profile for all users for PowerShell host
+
+if ($IsWindows)
+{
+	# Add Windows PowerShell PSModulePath to make it easier to discover potentially compatible PowerShell modules
+	# If a Windows PowerShell module works or not, please provide feedback at https://github.com/PowerShell/PowerShell/issues/4062
+
+	Write-Warning "Appended Windows PowerShell PSModulePath"
+	$env:psmodulepath += ";${env:userprofile}\Documents\WindowsPowerShell\Modules;${env:programfiles}\WindowsPowerShell\Modules;${env:windir}\system32\WindowsPowerShell\v1.0\Modules\"
```
adityapatwardhan: This only gets the default, if the Windows PSModulePath is modified by installing Azure cmdlets or SQL, then they won't be available. We would need something like this:
```Powershell
$env:PSModulePath += (";" + (& ${env:windir}\system32\windowspowershell\v1.0\powershell.exe '$env:PSModulePath'))
```

adityapatwardhan (changes requested):

build.psm1
SteveL-MSFT: That wasn't part of this PR, but since I only see 3 hits, I'm leaning towards fixing it here instead of a new PR

src/powershell-win-core/Microsoft.PowerShell_profile.ps1
SteveL-MSFT: Calling out to Windows PowerShell can incur a 150-250ms startup penalty.  Since this change isn't intended to be the final solution and just helps to get customer feedback, I think we are ok with it not being 100% accurate.  Advanced customers can update their PSModulePath as needed.

SteveL-MSFT:
Although I intend for this to be temporary, I'll add a test case

build.psm1
```diff
@@ -472,6 +472,12 @@ cmd.exe /C cd /d "$location" "&" "$($vcVarsPath)\vcvarsall.bat" "$NativeHostArch
         Pop-Location
     }
 
+    # copy PowerShell host profile if Windows
+    if ($IsWindows)
+    {
+        Copy-Item -Path "$PSScriptRoot/src/powershell-win-core/Microsoft.PowerShell_profile.ps1" -Destination $publishPath
```
adityapatwardhan: -Force ? When rebuilding without cleaning.

test/powershell/engine/Module/ModulePath.Tests.ps1
```diff
@@ -93,4 +93,19 @@ Describe "SxS Module Path Basic Tests" -tags "CI" {
         $paths -contains $fakePSHomeModuleDir | Should Be $true
         $paths -contains $customeModules | Should Be $true
     }
+
+    It "Default PowerShell profile appends Windows PowerShell PSModulePath only on Windows" {
+
+        $psmodulepath = & $powershell -c '$env:PSModulePath'
```
adityapatwardhan: Like in this test case, would other tests now have the warning message and fail? Can you verify by running all tests?

adityapatwardhan (changes requested):

lzybkr:
Because this is a temporary change - can we have a temporary test that will fail at some **fixed** point in the future - even if it's a nightly failure, that would be fine.

The point here - if we reach that point where the test starts failing, we need to take action, either implement this correctly (not via a profile script), remove it, or something else.

build.psm1
SteveL-MSFT: make sense, will add

test/powershell/engine/Module/ModulePath.Tests.ps1
SteveL-MSFT: I thought I did, but hit a few more failures.  Will fix.

src/powershell-win-core/Microsoft.PowerShell_profile.ps1
SteveL-MSFT: I believe I've fixed them all running CI and Feature tests

SteveL-MSFT:
@lzybkr rather than having a temporary test, I've changed this to not be a 'fix' but related so we'll keep the issue open and will revisit after we get some customer data

adityapatwardhan (approved):

SteveL-MSFT:
@adityapatwardhan looks like I had some changes staged on a different computer that wasn't committed (Console Tests), can you take a quick look?

adityapatwardhan (approved):
LGTM

SteveL-MSFT:
@lzybkr can you merge so we can get this in for beta.4?

SteveL-MSFT:
@TravisEz13 can you merge this since @lzybkr is currently out of office?

daxian-dbw (approved):

alexandair:
This change doesn't make sense on new Nano Server container.
This is what you see when you start microsoft/nanoserver-insider-powershell:

"WARNING: Appended Windows PowerShell PSModulePath"

```
PS C:\> $env:psmodulepath -split ';'
PowerShell\Modules
c:\program files\powershell\Modules
C:\Users\ContainerUser\Documents\WindowsPowerShell\Modules
C:\Program Files\WindowsPowerShell\Modules
C:\Windows\system32\WindowsPowerShell\v1.0\Modules\
```

3 of those 5 paths are unnecessary on new Nano Server.

SteveL-MSFT:
@alexandair you are correct, NanoServer-Insider doesn't have Windows PowerShell Core 5.1 anymore.  Can you add a comment to https://github.com/PowerShell/PowerShell/issues/4056?  The final resolution to this will likely be adding a config setting to the startup json file and checking if full windows.  Thanks.

alexandair:
@SteveL-MSFT Done.

MathiasMagnus:
Is there a way to turn off the warning? It's fairly annoying and I can give feedback without it reminding me on every startup.

SteveL-MSFT:
@MathiasMagnus edit Microsoft.PowerShell_profile.ps1 in $PSHome and remove the Write-Warning line

MathiasMagnus:
@SteveL-MSFT Thanks. :)

As a sidenote: is it instructive to think that these are 1-to-1 mappings?

| Powershell | Bash |
| ------------ | ---- |
| `$PSHOME\Microsoft.PowerShell_profile.ps1` | `/etc/profile` |
| `$PROFILE` | `~/.profile` |

SteveL-MSFT:
@MathiasMagnus yes

PR #4136, opened 2017-06-28
Added comma to replaced characters in assemblyname

TimCurwick:
A script with a native class will not run if it is saved with a name or full path with a comma.

I added the comma to the list of characters that are replaced with a similar character when creating an assemblyName based on a file path. Added pester test to test.

Discussed with Jason Shirk.

### DIFF
--- a/src/System.Management.Automation/engine/parser/PSType.cs
+++ b/src/System.Management.Automation/engine/parser/PSType.cs
@@ -1113,12 +1113,13 @@ namespace System.Management.Automation.Language
             var definedTypes = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
             // First character is a special mark that allows us to cheaply ignore dynamic generated assemblies in ClrFacede.GetAssemblies()
-            // Two replaces at the end are for not-allowed characters. They are replaced by similar-looking chars.
+            // The replaces at the end are for not-allowed characters. They are replaced by similar-looking chars.
             string assemblyName = ClrFacade.FIRST_CHAR_PSASSEMBLY_MARK + (string.IsNullOrWhiteSpace(rootAst.Extent.File)
                                       ? "powershell"
                                       : rootAst.Extent.File
                                       .Replace('\\', (char)0x29f9)
                                       .Replace('/', (char)0x29f9)
+                                      .Replace(',', (char)0x201a)
                                       .Replace(':', (char)0x0589));
 
             var assembly = AssemblyBuilder.DefineDynamicAssembly(new AssemblyName(assemblyName),
--- a/src/System.Management.Automation/engine/parser/PSType.cs
+++ b/src/System.Management.Automation/engine/parser/PSType.cs
@@ -1112,7 +1112,7 @@ namespace System.Management.Automation.Language
 
             var definedTypes = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
-            // First character is a special mark that allows us to cheaply ignore dynamic generated assemblies in ClrFacede.GetAssemblies()
+            // First character is a special mark that allows us to cheaply ignore dynamic generated assemblies in ClrFacade.GetAssemblies()
             // The replaces at the end are for not-allowed characters. They are replaced by similar-looking chars.
             string assemblyName = ClrFacade.FIRST_CHAR_PSASSEMBLY_MARK + (string.IsNullOrWhiteSpace(rootAst.Extent.File)
                                       ? "powershell"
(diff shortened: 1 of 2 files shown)


test/powershell/Language/Classes/Scripting.Classes.RunPath.Tests.ps1
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        finally {
+            Remove-Item -Path $FilePath
+        }
+
+        $Success | Should Be $True
+    }
+
+    It "Script with a class definition can run from a path with a comma" {
```
adityapatwardhan: Same comments as above test

adityapatwardhan: These two test cases can be combined into 1 using the -TestCases parameter for It. Example: https://github.com/PowerShell/PowerShell/blob/02b5f357a20e6dee9f8e60e3adb9025be3c94490/test/powershell/Language/Parser/LanguageAndParser.TestFollowup.Tests.ps1#L147

src/System.Management.Automation/engine/parser/PSType.cs
```diff
@@ -1113,12 +1113,13 @@ internal static Assembly DefineTypes(Parser parser, Ast rootAst, TypeDefinitionA
             var definedTypes = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
             // First character is a special mark that allows us to cheaply ignore dynamic generated assemblies in ClrFacede.GetAssemblies()
```
adityapatwardhan: Typo in comment for ClrFacade.

adityapatwardhan (changes requested):

test/powershell/Language/Classes/Scripting.Classes.RunPath.Tests.ps1
```diff
@@ -0,0 +1,42 @@
+Describe "Script with a class definition run path" -Tags "CI" {
+
+    It "Script with a class definition can run from a path without a comma" {
+
+        $FilePath = '.\MyTest.ps1'
+
+        try {
+            'class MyClass { [string]$MyProperty }; $True' | Out-File -FilePath $FilePath
```
lzybkr: Use `testdrive:` or `$testdrive` - then Pester will ensure the file is deleted when the test completes.

test/powershell/Language/Classes/Scripting.Classes.RunPath.Tests.ps1
```diff
@@ -0,0 +1,42 @@
...
+
+        catch {
+            $Success = $False
+        }
+
+        finally {
+            Remove-Item -Path $FilePath
+        }
+
+        $Success | Should Be $True
```
lzybkr: Avoid `Should Be $True` - this results in a typically unhelpful message in the log.

I suggest a test that actually uses the class, so maybe something like:

Instead, try something like:

```powershell
`@
class MyClass { static [string]$MyProperty = $PSScriptRoot }
[MyClass]::MyProperty
'@ > $testDrive/My,Test.ps1

& $testDrive/My,Test.ps1 | Should Match ".*My,Test.ps1"
```

Or something like that.

lzybkr (changes requested):

test/powershell/Language/Classes/Scripting.Classes.RunPath.Tests.ps1
TimCurwick: $PSScriptRoot doesn't work inside a class.

test/powershell/Language/Classes/Scripting.Classes.RunPath.Tests.ps1
```diff
@@ -1,42 +1,21 @@
-Describe "Script with a class definition run path" -Tags "CI" {
+Describe "Script with a class definition run path" {
```
adityapatwardhan: Missing -Tags 'CI'

adityapatwardhan (changes requested):

test/powershell/Language/Classes/Scripting.Classes.RunPath.Tests.ps1
```diff
@@ -1,42 +1,21 @@
...
-        catch {
-            $Success = $False
-        }
-
-        finally {
-            Remove-Item -Path $FilePath
-        }
-
-        $Success | Should Be $True
+        { . $FilePath } | Should Not Throw
```
lzybkr: This test is not necessary - the next test covers this case.

daxian-dbw:
@adityapatwardhan and @lzybkr could you please take another look? I think the author has addressed all your comments.

adityapatwardhan (approved):

daxian-dbw (approved):

daxian-dbw:
@TimCurwick Thanks for your contribution! 👍

Issue #3460, opened 2017-03-30
Suggestion: Provide a type accelerator for [System.Management.Automation.SemanticVersion]


mklement0:
Now that [semantic version numbers](http://semver.org/) (semver) are becoming ubiquitous, it would be nice to be able to do something like

```powershell
$ [semver] '1.2.3-preview-02+build42'

Major Minor Patch Label         
----- ----- ----- -----         
    1     2     3 preview-02+build42

```

instead of what you currently need to do:

```powershell
$ [System.Management.Automation.SemanticVersion] '1.2.3-preview-02+build42'
```

oising:
Ok, this looks mindlessly simple. Can I take this?

mklement0:
@oising: I can't speak for the team, but I think it's fine to just go ahead, now that you've announced your intent. Thanks for taking it on.

SteveL-MSFT:
@oising consider it yours!  I can only assign to formal team members so don't worry about that

oising:
Just a quick update - I got busy, and then got lost trying to get everything to build in VS2017 - but I'm back on track.

oising:
I don't see anything wrong with allowing [semver] in constrained language mode, so I'm permitting it there. If there's anything outlandish about this, bring it up in the PR review I guess.

iSazonov:
@oising Can we close the Issue as fxed?

oising:
Yes :) My first (and hopefully not last) PR.

iSazonov:
@oising Good start!

PR #4142, opened 2017-06-29
added 'semver' as core type accelerator for S.M.A.SemanticVersion

oising:
As per https://github.com/PowerShell/PowerShell/issues/3460

Added `[semver]` as a core (safe to use in constrained language mode) type accelerator for `System.Management.Automation.SemanticVersion`. 

Individual accelerators are not currently being tested, so no unit tests were added or updated.

### DIFF
--- a/src/System.Management.Automation/engine/parser/TypeResolver.cs
+++ b/src/System.Management.Automation/engine/parser/TypeResolver.cs
@@ -804,6 +804,7 @@ namespace System.Management.Automation
                     { typeof(XmlDocument),                                 new[] { "xml" } },
                     { typeof(CimSession),                                  new[] { "CimSession" } },
                     { typeof(MailAddress),                                 new[] { "mailaddress" } },
+                    { typeof(SemanticVersion),                             new[] { "semver" } },
 #if !CORECLR
                     // Following types not in CoreCLR
                     { typeof(DirectoryEntry),                              new[] { "adsi" } },


iSazonov:
Appveyor temporary failed.

oising:
Yeah, it seems one of the New-TimeSpan tests is failing. It must have already been broken when I merged to rebase.

daxian-dbw:
@PaulHigin currently `SemanticVersion` is permitted in constrained language, please review if that's OK.

PaulHigin:
@oising PowerShell core types are intended to be the absolute minimum needed in a constrained language interactive session.  Can you tell us the scenario where this type is needed and why constrained language is required?

oising:
@PaulHigin I guess I included it in core types for the same reasons you guys included `[ModuleSpecification]`. Given that NuGet packages are a core (no pun intended) building block of CoreFX, and `[semver]` being a large part of their identity, it seemed like something that may be more useful than not. I know this is all a bit fuzzy, but it's just what made sense to me.

daxian-dbw:
@PaulHigin I think the rationale is that `SemanticVersion` is replacing `System.Version` ( the type of `$PSVersionTable.PSVersion` is `SemanticVersion`), so I think it makes sense to allow it in constrained language. But we need your expertise to inspect the implementation of `SemanticVersion` and see if it's safe to be exposed in constrained language.

oising:
@daxian-dbw Yes, also good points.

PaulHigin:
Ok, I think this is reasonable.  I'll review for security safety.  Thanks.

PaulHigin (approved):
LGTM

daxian-dbw:
Filed #4221 to track adding tests for type accelerators in PowerShell.

PR #4175, opened 2017-07-02
Close FileStream opened to read file go get hash.

### LINKED ISSUE #4174: Get-FileHash is leaving file open - can't remove it later
This [issue was originally filed on the Plaster project](https://github.com/PowerShell/Plaster/issues/240) but the problem seems to be with PowerShell Core.

Steps to reproduce
------------------

```powershell
PS C:\Users\Keith> $hash = (Get-FileHash .\remove.psd1).Hash
PS C:\Users\Keith> $hash
E58204D2C5BB9057A085C00D3EEC3ED96390C52A739D4ADCCC02A1B1A4DDF1B8
PS C:\Users\Keith> Remove-Item .\remove.psd1
Remove-Item : The process cannot access the file 'C:\Users\Keith\remove.psd1' because it is being used by another
process.
At line:1 char:1
+ Remove-Item .\remove.psd1
+ ~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : WriteError: (C:\Users\Keith\remove.psd1:FileInfo) [Remove-Item], IOException
    + FullyQualifiedErrorId : RemoveFileSystemItemIOError,Microsoft.PowerShell.Commands.RemoveItemCommand
```

Expected behavior
-----------------

```none
Remove-Item should work.
```

Actual behavior
---------------

```none
Remove-Item fails because apparently the execution of Get-FileHash has resulted in an open file handle on the file.
```

Environment data
----------------

```powershell
> $PSVersionTable
GitCommitId                    v6.0.0-beta.3-20-g27e47f9ee1aa0592bffe17c8f979cfa6513887a4
```

rkeithhill:
Fix #4174

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetHash.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetHash.cs
@@ -121,10 +121,11 @@ namespace Microsoft.PowerShell.Commands
             {
                 byte[] bytehash = null;
                 String hash = null;
+                Stream openfilestream = null;
 
                 try
                 {
-                    Stream openfilestream = File.OpenRead(path);
+                    openfilestream = File.OpenRead(path);
                     bytehash = hasher.ComputeHash(openfilestream);
 
                     hash = BitConverter.ToString(bytehash).Replace("-","");
@@ -137,6 +138,10 @@ namespace Microsoft.PowerShell.Commands
                         ErrorCategory.ObjectNotFound,
                         path);
                     WriteError(errorRecord);
+                }
+                finally
+                {
+                    openfilestream?.Dispose();
                 }
             }
         }


src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetHash.cs
```diff
@@ -121,10 +121,11 @@ protected override void ProcessRecord()
             {
                 byte[] bytehash = null;
                 String hash = null;
+                Stream openfilestream = null;
 
                 try
                 {
-                    Stream openfilestream = File.OpenRead(path);
+                    openfilestream = File.OpenRead(path);
```
SteveL-MSFT: How about 'using' instead?

rkeithhill:
I went with that approach first but with the catch statement if seemed to read better using `try/catch/finally` than `try ( using) / catch`.  That said, I can certainly remove the finally and put a `using` statement inside the `try`.  It's about six of one / half dozen the other.  :-)

iSazonov:
LGTM.

SteveL-MSFT:
@rkeithhill looking at the entire file, keeping it try...finally is probably more consistent with other try...catch usage in this file, so LGTM

SteveL-MSFT (approved):

rkeithhill:
Sure would be nice to get this fix into beta.4.  Got some Plaster users on Linux who are hitting this bug.

PR #4178, opened 2017-07-03
Allow passing $true/$false as parameter to script using -File

### LINKED ISSUE #4036: Executing powershell script with bool parameter doesnt work
Steps to reproduce
------------------
Run from CMD the following command
powershell -ExecutionPolicy Unrestricted -File .\somescript.ps1 -someParameter $True

In the script itself i have the following line
[Parameter(Mandatory=$true)] [bool] $installSQL

I tried playing with that line and the arguments in many ways, all fail.. i have to turn this into a string and make the string into a bool :\

Expected behavior
-----------------
Script starting to work.. :)

Actual behavior
---------------

Error Message Cannot process
argument transformation on parameter 'installSQL'. Cannot convert value "System.String" to type "System.Boolean".
Boolean parameters accept only Boolean values and numbers, such as $True, $False, 1 or 0.
    + CategoryInfo          : InvalidData: (:) [bootstrapImageScript.ps1], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : ParameterArgumentTransformationError,bootstrapImageScript.ps1

Environment data
----------------

PSVersion                      5.1.14393.1198
BuildVersion                   10.0.14393.1198

SteveL-MSFT:
Using powershell.exe to execute a PowerShell script using `-File` currently provides no way to pass $true/$false as parameter values.  Current behavior is that `-File` accumulates passed parameters as strings only.

Fix is to special case this based on discussion with PS-Committee to support $true/$false as parsed values to parameters.  Switch values is also supported as currently documented syntax doesn't work.

Fix https://github.com/PowerShell/PowerShell/issues/4036

Doc change is https://github.com/PowerShell/PowerShell-Docs/pull/1430

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
@@ -858,6 +858,19 @@ namespace Microsoft.PowerShell
             // of the script to evaluate. If -file comes before -command, it will
             // treat -command as an argument to the script...
 
+            bool? GetBoolValue(string arg)
+            {
+                if (arg.Equals("$true", StringComparison.OrdinalIgnoreCase) || arg.Equals("true", StringComparison.OrdinalIgnoreCase))
+                {
+                    return true;
+                }
+                else if (arg.Equals("$false", StringComparison.OrdinalIgnoreCase) || arg.Equals("false", StringComparison.OrdinalIgnoreCase))
+                {
+                    return false;
+                }
+                return null;
+            }
+
             ++i;
             if (i >= args.Length)
             {
@@ -922,7 +935,6 @@ namespace Microsoft.PowerShell
 
                 i++;
 
-                Regex argPattern = new Regex(@"^.\w+\:", RegexOptions.CultureInvariant);
                 string pendingParameter = null;
 
                 // Accumulate the arguments to this script...
@@ -939,17 +951,25 @@ namespace Microsoft.PowerShell
                     }
                     else if (!string.IsNullOrEmpty(arg) && SpecialCharacters.IsDash(arg[0]))
                     {
-                        Match m = argPattern.Match(arg);
-                        if (m.Success)
+                        int offset = arg.IndexOf(':');
+                        if (offset >= 0)
                         {
-                            int offset = arg.IndexOf(':');
                             if (offset == arg.Length - 1)
                             {
                                 pendingParameter = arg.TrimEnd(':');
                             }
                             else
                             {
-                                _collectedArgs.Add(new CommandParameter(arg.Substring(0, offset), arg.Substring(offset + 1)));
+                                string argValue = arg.Substring(offset + 1);
+                                bool? boolValue = GetBoolValue(argValue);
+                                if (boolValue != null)
+                                {
+                                        _collectedArgs.Add(new CommandParameter(arg.Substring(0, offset), boolValue));
+                                }
+                                else
+                                {
+                                        _collectedArgs.Add(new CommandParameter(arg.Substring(0, offset), argValue));
+                                }
                             }
                         }
                         else
@@ -959,7 +979,15 @@ namespace Microsoft.PowerShell
                     }
                     else
                     {
-                        _collectedArgs.Add(new CommandParameter(null, arg));
+                        bool? boolValue = GetBoolValue(arg);
+                        if (boolValue != null)
+                        {
+                            _collectedArgs.Add(new CommandParameter(null, boolValue));
+                        }
+                        else
+                        {
+                            _collectedArgs.Add(new CommandParameter(null, arg));
+                        }
                     }
                     ++i;
                 }
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
@@ -858,17 +858,20 @@ namespace Microsoft.PowerShell
             // of the script to evaluate. If -file comes before -command, it will
             // treat -command as an argument to the script...
 
-            bool? GetBoolValue(string arg)
+            bool TryGetBoolValue(string arg, out bool boolValue)
             {
                 if (arg.Equals("$true", StringComparison.OrdinalIgnoreCase) || arg.Equals("true", StringComparison.OrdinalIgnoreCase))
                 {
+                    boolValue = true;
                     return true;
                 }
                 else if (arg.Equals("$false", StringComparison.OrdinalIgnoreCase) || arg.Equals("false", StringComparison.OrdinalIgnoreCase))
                 {
-                    return false;
+                    boolValue = false;
+                    return true;
                 }
-                return null;
+                boolValue = false;
+                return false;
             }
 
             ++i;
@@ -961,14 +964,14 @@ namespace Microsoft.PowerShell
                             else
                             {
                                 string argValue = arg.Substring(offset + 1);
-                                bool? boolValue = GetBoolValue(argValue);
-                                if (boolValue != null)
+                                string argName = arg.Substring(0, offset);
+                                if (TryGetBoolValue(argValue, out bool boolValue))
                                 {
-                                        _collectedArgs.Add(new CommandParameter(arg.Substring(0, offset), boolValue));
+                                        _collectedArgs.Add(new CommandParameter(argName, boolValue));
                                 }
                                 else
                                 {
-                                        _collectedArgs.Add(new CommandParameter(arg.Substring(0, offset), argValue));
+                                        _collectedArgs.Add(new CommandParameter(argName, argValue));
                                 }
                             }
                         }
@@ -979,15 +982,7 @@ namespace Microsoft.PowerShell
                     }
                     else
                     {
-                        bool? boolValue = GetBoolValue(arg);
-                        if (boolValue != null)
-                        {
-                            _collectedArgs.Add(new CommandParameter(null, boolValue));
-                        }
-                        else
-                        {
-                            _collectedArgs.Add(new CommandParameter(null, arg));
-                        }
+                        _collectedArgs.Add(new CommandParameter(null, arg));
                     }
                     ++i;
                 }
--- a/test/powershell/Host/ConsoleHost.Tests.ps1
+++ b/test/powershell/Host/ConsoleHost.Tests.ps1
@@ -201,7 +201,39 @@ Describe "ConsoleHost unit tests" -tags "Feature" {
             $observed[1] | Should Be "bar"
         }
 
-        It "-File should return exit code from script: <Filename>" -TestCases @(
+        It "-File should be able to pass bool values as strings to parameters: <BoolString>" -TestCases @(
+            @{BoolString = '$truE';BoolValue = 'True'},
+            @{BoolString = '$falsE';BoolValue = 'False'},
+            @{BoolString = 'truE';BoolValue = 'True'},
+            @{BoolString = 'falsE';BoolValue = 'False'}
+         ) {
+            param([string]$BoolString, [string]$BoolValue)
+            Set-Content -Path $testdrive/test.ps1 -Value 'param([bool]$bool) $bool'
+            $observed = & $powershell -NoProfile -Nologo -File $testdrive/test.ps1 -Bool $BoolString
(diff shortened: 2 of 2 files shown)


daxian-dbw:
The commits has changes in `src/Modules/Shared/Pester` which should not be included. I think you need to run `git submodule update` to resolve it.

SteveL-MSFT:
@daxian-dbw fixed

SteveL-MSFT:
@lzybkr if you get a chance, can you review this?  thanks

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -939,8 +951,7 @@ private bool ParseFile(string[] args, ref int i, bool noexitSeen)
                     }
                     else if (!string.IsNullOrEmpty(arg) && SpecialCharacters.IsDash(arg[0]))
                     {
-                        Match m = argPattern.Match(arg);
-                        if (m.Success)
+                        if (arg.Contains(":"))
```
lzybkr: Instead of walking the string twice, move the call to IndexOf to before the `if` and test `offset` >= 0 instead.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -949,7 +960,15 @@ private bool ParseFile(string[] args, ref int i, bool noexitSeen)
                             }
                             else
                             {
-                                _collectedArgs.Add(new CommandParameter(arg.Substring(0, offset), arg.Substring(offset + 1)));
+                                bool? boolValue = GetBoolValue(arg.Substring(offset + 1));
```
lzybkr: In the common case, you are creating the same string twice with `arg.Substring(offiset + 1)`, once here, and once below.

lzybkr: So this can be made much simpler, something like:
```c#
string argValue = arg.Substring(offset + 1);
bool? boolValue = GetBoolValue(argValue);
_collectedArgs.Add(new CommandParameter(arg.Substring(0, offset), boolValue ?? argValue);
```

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -959,7 +978,15 @@ private bool ParseFile(string[] args, ref int i, bool noexitSeen)
                     }
                     else
                     {
-                        _collectedArgs.Add(new CommandParameter(null, arg));
+                        bool? boolValue = GetBoolValue(arg);
```
lzybkr: Again, you can simply a bunch like above with `??`.

test/powershell/Host/ConsoleHost.Tests.ps1
```diff
@@ -201,7 +201,37 @@ Describe "ConsoleHost unit tests" -tags "Feature" {
             $observed[1] | Should Be "bar"
         }
 
-        It "-File should return exit code from script: <Filename>" -TestCases @(
+        It "-File should be able to pass bool values as strings to parameters: <BoolString>" -TestCases @(
```
lzybkr: I don't think these tests reflect the real user scenario.

Passing `$BoolString` is completely different than passing `$true`.

`$BoolString` is getting replaced with `$true`, but `$true` gets replaced with `True`.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -858,6 +858,19 @@ private bool ParseFile(string[] args, ref int i, bool noexitSeen)
             // of the script to evaluate. If -file comes before -command, it will
             // treat -command as an argument to the script...
 
+            bool? GetBoolValue(string arg)
+            {
+                if (arg.Equals("$true", StringComparison.OrdinalIgnoreCase))
```
lzybkr: I think this needs to be `True`, not `$true`.

Or maybe it needs to be both `$true` and `True` - because it depends on the shell how `$true` is expanded.

lzybkr (changes requested):

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: Isn't the use case:
```
powershell -file foo.ps1 -myswitch:$true
```

Not clear to me when we need `True` as I don't think we want to support:

```
powershell -file foo.ps1 -myswitch:true
```

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: good point, will fix

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: will fix

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: will fix

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: Ok, I think I understand what you're getting it by your comment below.  From a non-powershell shell like bash (which I thought would be the primary use case), I would think we want to support PowerShell syntax for passing $true and $false.  From within powershell, I suppose we shouldn't expect the user to escape $true to be just a string.  I can make this change.

test/powershell/Host/ConsoleHost.Tests.ps1
SteveL-MSFT: See comment above regarding usage from different shells.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: null-coalescing operator requires same type on both sides, I can use ternary operator to simplify it instead

SteveL-MSFT: Nevermind, even ternary expects same type

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: same as above

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
iSazonov: We can use out parameter or return argValue as default value.

SteveL-MSFT:
@lzybkr feedback addressed

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
lzybkr: You can still simplify though, something like:
```c#
object argValue = arg.Substring(offset + 1);
 _collectedArgs.Add(new CommandParameter(arg.Substring(0, offset), GetArgPossiblyAsBool(argValue)));
```

And `GetArgPossiblyAsBool` returns `object`.

test/powershell/Host/ConsoleHost.Tests.ps1
lzybkr: We should have a test that accepts a `string` parameter and you pass `true` and `false` - we want to make sure that when we convert to `bool`, the `bool` gets properly converted back to the correct string.

And come to think about it, there is a problem - we'll lose the case that the user specified.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -939,17 +951,25 @@ private bool ParseFile(string[] args, ref int i, bool noexitSeen)
...
                             if (offset == arg.Length - 1)
                             {
                                 pendingParameter = arg.TrimEnd(':');
                             }
                             else
                             {
-                                _collectedArgs.Add(new CommandParameter(arg.Substring(0, offset), arg.Substring(offset + 1)));
+                                string argValue = arg.Substring(offset + 1);
+                                bool? boolValue = GetBoolValue(argValue);
+                                if (boolValue != null)
```
iSazonov: Sorry again -
Can we use `LanguagePrimitives.TryConvertTo` or `LanguagePrimitives.TryConvertArg`?
Can we use TryGetBoolValue pattern?

```c#
                               string argValue1 = arg.Substring(offset + 1); 
                               string argValue0 = arg.Substring(0, offset);
                               if (TryGetBoolValue(argValue1, out bool boolValue))  
                               {  
                                       _collectedArgs.Add(new CommandParameter(argValue0, boolValue));  
                              }  
                               else  
                               {  
                                       _collectedArgs.Add(new CommandParameter(argValue0, argValue1));  
                               }  
```

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: Will change

SteveL-MSFT: Doesn't work, can't implicitly convert object to bool/string

test/powershell/Host/ConsoleHost.Tests.ps1
SteveL-MSFT: Yeah, that's a problem.  I've changed it so it only works with switches and added tests for other cases

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: Sure

test/powershell/Host/ConsoleHost.Tests.ps1
```diff
@@ -201,31 +201,35 @@ Describe "ConsoleHost unit tests" -tags "Feature" {
...
-         ) {
-            param([string]$BoolString, [string]$BoolValue)
-            Set-Content -Path $testdrive/test.ps1 -Value 'param([bool]$bool) $bool'
-            $observed = & $powershell -NoProfile -Nologo -File $testdrive/test.ps1 -Bool $BoolString
-            $observed | Should Be $BoolValue
+        It "-File should be able to pass bool string values as string to positional parameters: <BoolString>" -TestCases @(
+            @{BoolString = '$truE'},
+            @{BoolString = '$falSe'},
+            @{BoolString = 'trUe'},
+            @{BoolString = 'faLse'}
```
iSazonov: Please add a protection comment like "Don't change case!".

SteveL-MSFT: sure

SteveL-MSFT:
@lzybkr any other concerns?

lzybkr (approved):

ChristopherGLewis:
C:\Temp>type test.ps1
Param( [bool] $test )
Echo $test
C:\Temp>pwsh.exe -file test.ps1 -test $false
test.ps1: Cannot process argument transformation on parameter 'test'. Cannot convert value "System.String" to type "System.Boolean". Boolean parameters accept only Boolean values and numbers, such as $True, $False, 1 or 0.

C:\Temp>pwsh.exe -command $PSVersionTable

PSVersion                      7.0.0
GitCommitId                    7.0.0
OS                             Microsoft Windows 10.0.18363
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}

iSazonov:
@ChristopherGLewis See #10838

Issue #4162, opened 2017-07-01
Add tests for -is, -isnot, -contains, -notcontains and -like operators


adityapatwardhan:
Add tests for -is, -isnot, -contains, -notcontains and -like operators

Related code coverage report:

https://codecov.io/gh/PowerShell/PowerShell/src/9e41c647a9407e05d86f4971a6bd25719a7130d0/src/System.Management.Automation/engine/lang/parserutils.cs

Environment data
----------------

```powershell
> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.3
OS                             Microsoft Windows 10.0.16204
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

```

SteveL-MSFT:
@adityapatwardhan is this for the parser or where-object params?  I see tests for -like, -contains, and -notcontains and -is is used by other tests

adityapatwardhan:
@SteveL-MSFT  Added link to the report showing the missing coverage. Not sure why this is not covered.

PR #4179, opened 2017-07-03
added test coverage for -notcontains, -is, -isnot operators

SteveL-MSFT:
-like and -contains already has coverage

Fix https://github.com/PowerShell/PowerShell/issues/4162

### DIFF
--- a/test/powershell/Language/Operators/ComparisonOperator.Tests.ps1
+++ b/test/powershell/Language/Operators/ComparisonOperator.Tests.ps1
@@ -1,79 +1,54 @@
 Describe "ComparisonOperator" -tag "CI" {
-    It "Should be $true for 1 -lt 2" {
-	 1 -lt 2       | Should Be $true
-    }
-	It "Should be $false for 1 -gt 2" {
-	 1 -gt 2       | Should Be $false
-    }
-	It "Should be $true for 1 -le 2" {
-	 1 -le 2       | Should Be $true
-    }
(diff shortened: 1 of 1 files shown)


test/powershell/Language/Operators/ComparisonOperator.Tests.ps1
```diff
@@ -47,9 +47,13 @@ Describe "ComparisonOperator" -tag "CI" {
 	 !0       | Should Be $true
     }
 
-	It "Should be $true for 'Hello','world' -contains 'Hello'" {
-	 $arr= 'Hello','world'
-	 $arr -contains 'Hello'       | Should Be $true
+	It "Should be <result> for 'Hello','world' <operator> <rhs>" -TestCases @(
+        @{result = $true; operator = "-contains"; rhs = "'Hello'"},
```
adityapatwardhan: Test for case sensitive variants? ccontains and cnotcontains? There are some tests below. Can we move them to this It and add it as part of -TestCases array.

adityapatwardhan (changes requested):

test/powershell/Language/Operators/ComparisonOperator.Tests.ps1
SteveL-MSFT: I'll do some general cleanup of the tests

SteveL-MSFT:
@adityapatwardhan addressed

adityapatwardhan (approved):
LGTM

PR #4106, opened 2017-06-26
Refactor MSBuild project files by adding the common property file

iSazonov:
Reopen #3917 
Related #3400

Motivation
---
We inherited old assembly file versions from Windows PowerShell.
Currently we shouldn't set versions of assemblies statically in `AssemblyInfo.cs` files, we should set versions of assemblies dynamically by MSBuild as discussed in #3690.

Fix
---
1. Remove `AssemblyVersion` and `AssemblyFileVersionAttribute` attributes from `AssemblyInfo.cs` files.
2. Put all common properties in PowerShell.Common.props including version related properties.
3. Import `PowerShell.Common.props` in all csproj files.

Follow-Up Work
---
In the PR we temporarily hard code assembly versions as "6.0.0-beta.3".
Later we should set a dll version based on GitCommitId.

Additional considerations
---
In previous PR #3917 we caught some problems and we should fix it here:
 - package versions now corrupted (1.0.0) ([for help](https://stackoverflow.com/questions/42183300/what-is-the-difference-between-various-msbuild-version-properties-such-as-versi)): **[Continue to use 'VersionPrefix' tag. 'Version' will be inferred from this tag and all other versions, such as AssemblyVersion, FileVesion and PackageVersion and etc, will be inferred from 'Version']**
    - set PackageVersion?
    - set Version?
 - SMA.dll backward compatibility: **[Reviewed by powershell-committee and agreed to bump AssemblyVersion to 6.0.0.0]**
    - statically set `AssemblyVersion` 3.0.00  (`AssemblyFileVersionAttribute`?)
    - dynamically set `AssemblyVersion` by MSBuild
    - any tests?

Note
-----
Quoted from @SteveL-MSFT:
> @PowerShell/powershell-committee reviewed this and is fine with taking the assembly version to 6.0.0.0 and taking it now to discover customer impact

PR #4182, opened 2017-07-04
Refactor MSBuild project files to get PowerShell version from git tag

iSazonov:
Related #3400
Continue #3690 and #4106.

Extract information about the release tag, number of commits since the tag and the hash of the latest commit within a MSBuild target, and bake that information into version properties of the assemblies appropriately.

- New target `GetPSCoreVersionFromGit` depends on (BeforeTargets) Restore and Nuget steps so now we get packages with the right version.
- Added generating a version without suffix (for release tag).
- Now ProductVersion is the same as  AssemblyInformationalVersion

### DIFF
--- a/PowerShell.Common.props
+++ b/PowerShell.Common.props
@@ -43,18 +43,19 @@
       <PSCoreFormattedVersion Condition = "'$(ReleaseTag)' != '' or '$(PSCoreAdditionalCommits)' == '0'">$(PSCoreBuildVersion) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
       <PSCoreFormattedVersion Condition = "'$(PSCoreFormattedVersion)' == ''">$(PSCoreBuildVersion) Commits: $(PSCoreAdditionalCommits) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
 
-      <ProductVersion>$(PSCoreFormattedVersion)</ProductVersion>
       <!--
-            Here we define explicitly 'Version' to set 'FileVersion' by 'GetAssemblyVersion' target in 'Microsoft.NET.GenerateAssemblyInfo.targets'.
+            Here we define explicitly 'Version' to set 'FileVersion' and 'AssemblyVersion' by 'GetAssemblyVersion' target in 'Microsoft.NET.GenerateAssemblyInfo.targets'.
             Here we define explicitly 'InformationalVersion' because by default it is defined as 'Version' by 'GetAssemblyVersion' target in 'Microsoft.NET.GenerateAssemblyInfo.targets'.
       -->
       <Version>$(PSCoreBuildVersion)</Version>
       <InformationalVersion>$(PSCoreFormattedVersion)</InformationalVersion>
+      <ProductVersion>$(PSCoreFormattedVersion)</ProductVersion>
 
       <!--
             We have explicitly assign 'PackageVersion'
             because there is a bug: 'PackageVersion' is correctly assigned as 'Version' in 'NuGet.targets'
             but then immediately redefined as '1.0.0'.
+            Tracking Issue https://github.com/dotnet/sdk/issues/1557
       -->
       <PackageVersion>$(PSCoreBuildVersion)</PackageVersion>
 
@@ -71,7 +72,6 @@
                PSCoreMinorVersion = $(PSCoreMinorVersion);
                PSCorePatchVersion = $(PSCorePatchVersion);
                PSCoreLabelVersion = $(PSCoreLabelVersion);
-               GitCDUP = $(GitCDUP);
                RegexGitVersion = $(RegexGitVersion);
                PSCoreFormattedVersion = $(PSCoreFormattedVersion);
                ProductVersion = $(ProductVersion);
--- a/PowerShell.Common.props
+++ b/PowerShell.Common.props
@@ -1,16 +1,16 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <!--
-        We should ~explicitly~ assign versions properties in the target
+        We should *explicitly* assign version properties in this target
         because:
-          1. By default the version properties defined in global PropertyGroup (not target!) in 'Microsoft.NET.DefaultAssemblyInfo.targets'
-          2. Properties in global 'PropertyGroups' is assigned before any targets
+          1. By default the version properties are defined in global 'PropertyGroup' (not target!) in 'Microsoft.NET.DefaultAssemblyInfo.targets'
+          2. Properties in global 'PropertyGroups' are assigned before any targets
           3. and cannot be conditionally excecuted as targets (after this target).
 
-        To assign versions properties we should excecute the target:
-          '_GenerateRestoreProjectSpec' - before 'Restore' target
-          'GenerateNuspec'              - before 'Pack' target
-          'BeforeBuild'                 - before 'Build' target
+        To assign version properties we should excecute this target:
+          before 'Restore' target - '_GenerateRestoreProjectSpec'
+          before 'Pack' target    - 'GenerateNuspec'
+          before 'Build' target   - 'BeforeBuild'
   -->
   <Target Name="GetPSCoreVersionFromGit"
     BeforeTargets="_GenerateRestoreProjectSpec;GenerateNuspec;BeforeBuild"
@@ -33,15 +33,18 @@
       <PSCoreAdditionalCommits>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[2].Value)</PSCoreAdditionalCommits>
       <PSCoreCommitSHA>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[3].Value)</PSCoreCommitSHA>
 
-      <RegexSymVer>^((\d+).(\d+).(\d+))(?:-(.+))?</RegexSymVer>
-      <PSCorePrefixVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[1].Value)</PSCorePrefixVersion>
-      <PSCoreMajorVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[2].Value)</PSCoreMajorVersion>
-      <PSCoreMinorVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[3].Value)</PSCoreMinorVersion>
-      <PSCorePatchVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[4].Value)</PSCorePatchVersion>
-      <PSCoreLabelVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[5].Value)</PSCoreLabelVersion>
-
       <PSCoreFormattedVersion Condition = "'$(ReleaseTag)' != '' or '$(PSCoreAdditionalCommits)' == '0'">$(PSCoreBuildVersion) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
       <PSCoreFormattedVersion Condition = "'$(PSCoreFormattedVersion)' == ''">$(PSCoreBuildVersion) Commits: $(PSCoreAdditionalCommits) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
+
+      <!-- Extract the major, minor and patch version numbers, as well as the preview label.
+           They are currently not used anywhere, so we comment them out for now.
+        <RegexSymVer>^((\d+).(\d+).(\d+))(?:-(.+))?</RegexSymVer>
+        <PSCorePrefixVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[1].Value)</PSCorePrefixVersion>
+        <PSCoreMajorVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[2].Value)</PSCoreMajorVersion>
+        <PSCoreMinorVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[3].Value)</PSCoreMinorVersion>
+        <PSCorePatchVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[4].Value)</PSCorePatchVersion>
+        <PSCoreLabelVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[5].Value)</PSCoreLabelVersion>
+      -->
 
       <!--
             Here we define explicitly 'Version' to set 'FileVersion' and 'AssemblyVersion' by 'GetAssemblyVersion' target in 'Microsoft.NET.GenerateAssemblyInfo.targets'.
@@ -61,7 +64,7 @@
 
     </PropertyGroup>
 
-   <!-- for debug output
+    <!-- Output For Debugging
       <WriteLinesToFile File="targetfile1.txt"
         Lines="ReleaseTag=$(ReleaseTag);
                PowerShellVersion=$(PowerShellVersion);
@@ -79,7 +82,8 @@
                ProjectVersion = '!$(ProjectVersion)!'
                InformationalVersion = $(InformationalVersion);
               "
-        Overwrite="true"/-->
+        Overwrite="true" />
+    -->
 
   </Target>
 
@@ -99,7 +103,6 @@
     <DelaySign>true</DelaySign>
     <AssemblyOriginatorKeyFile>../signing/visualstudiopublic.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
-
   </PropertyGroup>
 
 </Project>
--- a/PowerShell.Common.props
+++ b/PowerShell.Common.props
@@ -7,7 +7,7 @@
           2. Properties in global 'PropertyGroups' is assigned before any targets
           3. and cannot be conditionally excecuted as targets (after this target).
 
-        To assign vertions properties we should excecute the target:
+        To assign versions properties we should excecute the target:
           '_GenerateRestoreProjectSpec' - before 'Restore' target
           'GenerateNuspec'              - before 'Pack' target
           'BeforeBuild'                 - before 'Build' target
@@ -28,14 +28,23 @@
       <GitInfoBaseDir Condition="'$(GitInfoBaseDir)' == ''">$(MSBuildProjectDirectory)/$(GitCDUP)/.git</GitInfoBaseDir>
     </PropertyGroup>
 
-    <Exec Command='$(GitExe) --git-dir="$(GitInfoBaseDir)" describe --abbrev=60 --long' ConsoleToMSBuild="true" EchoOff="true">
+    <Exec Command='$(GitExe) --git-dir="$(GitInfoBaseDir)" describe --abbrev=60 --long'
+          ConsoleToMSBuild="true"
+          EchoOff="true">
       <Output TaskParameter="ConsoleOutput" PropertyName="PowerShellVersion" />
     </Exec>
 
     <PropertyGroup>
       <RegexGitVersion>^v(.+)-(\d+)-g(.+)</RegexGitVersion>
-      <PSCoreBuildVersion>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[1].Value)</PSCoreBuildVersion>
+    </PropertyGroup>
+
+    <PropertyGroup Condition = "'$(ReleaseTag)' != ''">
+      <PSCoreBuildVersion>$(ReleaseTag)</PSCoreBuildVersion>
       <PSCoreAdditionalCommits>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[2].Value)</PSCoreAdditionalCommits>
+    </PropertyGroup>
+
+    <PropertyGroup>
+      <PSCoreBuildVersion Condition = "'$(PSCoreBuildVersion)' == ''">$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[1].Value)</PSCoreBuildVersion>
       <PSCoreCommitSHA>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[3].Value)</PSCoreCommitSHA>
 
       <RegexSymVer>^((\d+).(\d+).(\d+))(?:-(.+))?</RegexSymVer>
@@ -45,7 +54,7 @@
       <PSCorePatchVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[4].Value)</PSCorePatchVersion>
       <PSCoreLabelVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[5].Value)</PSCoreLabelVersion>
 
-      <PSCoreFormattedVersion Condition = "'$(PSCoreAdditionalCommits)' == '0'">$(PSCoreBuildVersion) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
+      <PSCoreFormattedVersion Condition = "'$(ReleaseTag)' != '' or '$(PSCoreAdditionalCommits)' == '0'">$(PSCoreBuildVersion) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
       <PSCoreFormattedVersion Condition = "'$(PSCoreFormattedVersion)' == ''">$(PSCoreBuildVersion) Commits: $(PSCoreAdditionalCommits) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
 
       <ProductVersion>$(PSCoreFormattedVersion)</ProductVersion>
@@ -69,8 +78,9 @@
     </PropertyGroup>
 
    <!-- for debug output
-        WriteLinesToFile File="targetfile1.txt"
-        Lines="PowerShellVersion=$(PowerShellVersion);
+      <WriteLinesToFile File="targetfile1.txt"
+        Lines="ReleaseTag=$(ReleaseTag);
+               PowerShellVersion=$(PowerShellVersion);
                PSCoreBuildVersion = $(PSCoreBuildVersion);
                PSCoreAdditionalCommits = $(PSCoreAdditionalCommits);
                PSCoreCommitSHA = $(PSCoreCommitSHA);
--- a/PowerShell.Common.props
+++ b/PowerShell.Common.props
@@ -16,35 +16,21 @@
     BeforeTargets="_GenerateRestoreProjectSpec;GenerateNuspec;BeforeBuild"
   >
 
-    <PropertyGroup>
-      <GitExe>git</GitExe>
-    </PropertyGroup>
-
-    <Exec Command='$(GitExe) rev-parse --show-cdup' ConsoleToMSBuild="true">
-      <Output TaskParameter="ConsoleOutput" PropertyName="GitCDUP" />
-    </Exec>
-
-    <PropertyGroup>
-      <GitInfoBaseDir Condition="'$(GitInfoBaseDir)' == ''">$(MSBuildProjectDirectory)/$(GitCDUP)/.git</GitInfoBaseDir>
-    </PropertyGroup>
-
-    <Exec Command='$(GitExe) --git-dir="$(GitInfoBaseDir)" describe --abbrev=60 --long'
+    <Exec Command='git describe --abbrev=60 --long'
+          WorkingDirectory=$(MSBuildProjectDirectory)
           ConsoleToMSBuild="true"
-          EchoOff="true">
+          StandardOutputImportance="Low">
       <Output TaskParameter="ConsoleOutput" PropertyName="PowerShellVersion" />
     </Exec>
 
-    <PropertyGroup>
-      <RegexGitVersion>^v(.+)-(\d+)-g(.+)</RegexGitVersion>
-    </PropertyGroup>
-
     <PropertyGroup Condition = "'$(ReleaseTag)' != ''">
       <PSCoreBuildVersion>$(ReleaseTag)</PSCoreBuildVersion>
-      <PSCoreAdditionalCommits>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[2].Value)</PSCoreAdditionalCommits>
     </PropertyGroup>
 
     <PropertyGroup>
+      <RegexGitVersion>^v(.+)-(\d+)-g(.+)</RegexGitVersion>
       <PSCoreBuildVersion Condition = "'$(PSCoreBuildVersion)' == ''">$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[1].Value)</PSCoreBuildVersion>
+      <PSCoreAdditionalCommits>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[2].Value)</PSCoreAdditionalCommits>
       <PSCoreCommitSHA>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[3].Value)</PSCoreCommitSHA>
 
       <RegexSymVer>^((\d+).(\d+).(\d+))(?:-(.+))?</RegexSymVer>
--- a/PowerShell.Common.props
+++ b/PowerShell.Common.props
@@ -1,13 +1,13 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <!--
-        We should *explicitly* assign version properties in this target
-        because:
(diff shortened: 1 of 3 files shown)


iSazonov:
@daxian-dbw Could you please review?

daxian-dbw:
@iSazonov Sure, next item on my list :)

mi-hol:
@daxian-dbw 4 weeks pass by quickly :) , hence I wonder if this PR getting any closer to the top of your list?

daxian-dbw:
Sorry for leaving this PR unattended for such a long time ... The conflicts were resolved.

The changes in `PowerShell.Common.props` is pretty clean and well organized. The comments are good and very helpful -- some minor touch-ups can be made but we can do that when we finalize the changes. I have a few concerns here about the change:
1. We need to be able to produce a release build without actually pushing the release tag. For example, we need to build for `v6.0.0-beta.6` without pushing the git tag `v6.0.0-beta.6`. This is needed because pushing a release tag should be the last step for a release so that any issues we discovered during the preparation of the release can still be fixed before having the release tag pushed. Currently, this is achieved by passing `-ReleaseTag` to `Start-PSBuild` and `Start-PSPackage`. Take `Start-PSBuild` for an instance, when `-ReleaseTag` is specified, it will use that as the release tag instead of turning to `git describe`. Is it possible to continue allowing this with the `GetPSCoreVersionFromGit` target? (maybe by passing the to-be-used tag name to a property and in case that property is not empty, use that instead of running `git describe`)

1. The target `GetPSCoreVersionFromGit` seems to be executed for multiple times (once per `csproj`). Can we make it run only once?

1. When the target runs `git`, the output shows on console (see the screenshot below). Is it possible to suppress it while still capturing the output in `Exec` tasks?

image

iSazonov:
1. I believe we can fix this. We can call MSBuild with explicit tag value and don't re-evaluate it in the target by condition.
2. I guess it is "by design" - MSBuild create new context for every csproj and re-run targets. I'll be happy if I'm wrong. (We can use this as an advantage if we'll want to assign the dll versions more precisely and independently.)
3. The same. It seems MSBuild can catch output external program only from console. Currently I don't know how suppress the noise.

PowerShell.Common.props
```diff
@@ -1,10 +1,100 @@
...
+
+    <Exec Command='$(GitExe) rev-parse --show-cdup' ConsoleToMSBuild="true">
+      <Output TaskParameter="ConsoleOutput" PropertyName="GitCDUP" />
+    </Exec>
+
+    <PropertyGroup>
+      <GitInfoBaseDir Condition="'$(GitInfoBaseDir)' == ''">$(MSBuildProjectDirectory)/$(GitCDUP)/.git</GitInfoBaseDir>
+    </PropertyGroup>
+
+    <Exec Command='$(GitExe) --git-dir="$(GitInfoBaseDir)" describe --abbrev=60 --long' ConsoleToMSBuild="true" EchoOff="true">
```
TravisEz13: why include the `git-dir` if you are just getting it from running git in the current directory anyway?

TravisEz13: During the release process, we manually specify the version.  How do you pick that up?

iSazonov: It seems best practice for csproj to use variables everywhere. Currently we can assign the dir externally `dotnet build /v:GitInfoBaseDir=...`.
We can use the same method to assign a version during the release process - @daxian-dbw ask the same and I'll check this and fix if need.

TravisEz13: I would suggest assigned the git directory externally.  The reason I've added it to all the git  commands I've touch is that sometimes git cannot find the directory, which the current solution will not solve.

iSazonov: Anybody can assign `GitInfoBaseDir` externally - it is supported in the target.

PowerShell.Common.props
```diff
@@ -1,10 +1,110 @@
...
+          'GenerateNuspec'              - before 'Pack' target
+          'BeforeBuild'                 - before 'Build' target
+  -->
+  <Target Name="GetPSCoreVersionFromGit"
+    BeforeTargets="_GenerateRestoreProjectSpec;GenerateNuspec;BeforeBuild"
+  >
+
+    <PropertyGroup>
+      <GitExe>git</GitExe>
+    </PropertyGroup>
```
daxian-dbw: Is this property necessary? dotnet/cli use 'git' directly -- see https://github.com/dotnet/cli/blob/45a1e9e56cc47e3973a49f164e98d41442cedebd/build/GitCommitInfo.targets#L3

PowerShell.Common.props
```diff
@@ -1,10 +1,110 @@
...
+
+    <Exec Command='$(GitExe) rev-parse --show-cdup' ConsoleToMSBuild="true">
```
daxian-dbw: Add `StandardOutputImportance="Low"`, and it seems to suppress the log of the output.

PowerShell.Common.props
```diff
@@ -1,10 +1,110 @@
...
+      <Output TaskParameter="ConsoleOutput" PropertyName="GitCDUP" />
+    </Exec>
+
+    <PropertyGroup>
+      <GitInfoBaseDir Condition="'$(GitInfoBaseDir)' == ''">$(MSBuildProjectDirectory)/$(GitCDUP)/.git</GitInfoBaseDir>
+    </PropertyGroup>
+
+    <Exec Command='$(GitExe) --git-dir="$(GitInfoBaseDir)" describe --abbrev=60 --long'
+          ConsoleToMSBuild="true"
+          EchoOff="true">
```
daxian-dbw: Is `EchoOff` necessary? I don't see a difference when building with or without it ...

PowerShell.Common.props
```diff
@@ -1,10 +1,110 @@
...
+    <Exec Command='$(GitExe) rev-parse --show-cdup' ConsoleToMSBuild="true">
+      <Output TaskParameter="ConsoleOutput" PropertyName="GitCDUP" />
+    </Exec>
+
+    <PropertyGroup>
+      <GitInfoBaseDir Condition="'$(GitInfoBaseDir)' == ''">$(MSBuildProjectDirectory)/$(GitCDUP)/.git</GitInfoBaseDir>
+    </PropertyGroup>
+
+    <Exec Command='$(GitExe) --git-dir="$(GitInfoBaseDir)" describe --abbrev=60 --long'
+          ConsoleToMSBuild="true"
```
daxian-dbw: Add `StandardOutputImportance="Low"`, and it seems to suppress the log of the output.

PowerShell.Common.props
```diff
@@ -1,10 +1,110 @@
...
+    BeforeTargets="_GenerateRestoreProjectSpec;GenerateNuspec;BeforeBuild"
+  >
+
+    <PropertyGroup>
+      <GitExe>git</GitExe>
+    </PropertyGroup>
+
+    <Exec Command='$(GitExe) rev-parse --show-cdup' ConsoleToMSBuild="true">
+      <Output TaskParameter="ConsoleOutput" PropertyName="GitCDUP" />
+    </Exec>
```
daxian-dbw: I think we don't need this step. When running `git describe` below, just set the `WorkingDirectory=$(MSBuildProjectDirectory)`, then it's guaranteed that the `git` command executes from the directory where `.csproj` locals, so `--git-dir` would be unnecessary.

PowerShell.Common.props
```diff
@@ -1,10 +1,110 @@
...
+      <Output TaskParameter="ConsoleOutput" PropertyName="PowerShellVersion" />
+    </Exec>
+
+    <PropertyGroup>
+      <RegexGitVersion>^v(.+)-(\d+)-g(.+)</RegexGitVersion>
+    </PropertyGroup>
+
+    <PropertyGroup Condition = "'$(ReleaseTag)' != ''">
+      <PSCoreBuildVersion>$(ReleaseTag)</PSCoreBuildVersion>
+      <PSCoreAdditionalCommits>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[2].Value)</PSCoreAdditionalCommits>
```
daxian-dbw: This seems wrong -- so the `PSCoreAdditionalCommits` will be set only if `ReleaseTag` is specified, however, it's used only if `ReleaseTag` is not specified.

PowerShell.Common.props
```diff
@@ -1,10 +1,110 @@
...
+    </PropertyGroup>
+
+    <PropertyGroup>
+      <PSCoreBuildVersion Condition = "'$(PSCoreBuildVersion)' == ''">$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[1].Value)</PSCoreBuildVersion>
```
daxian-dbw: It seems to me `RegexGitVersion` can be moved to this `PropertyGroup`.

daxian-dbw (dismissed):

PowerShell.Common.props
```diff
@@ -1,10 +1,110 @@
...
+          EchoOff="true">
+      <Output TaskParameter="ConsoleOutput" PropertyName="PowerShellVersion" />
+    </Exec>
+
+    <PropertyGroup>
+      <RegexGitVersion>^v(.+)-(\d+)-g(.+)</RegexGitVersion>
+    </PropertyGroup>
+
+    <PropertyGroup Condition = "'$(ReleaseTag)' != ''">
+      <PSCoreBuildVersion>$(ReleaseTag)</PSCoreBuildVersion>
```
daxian-dbw: The `ReleaseTag` starts with `'v'` (at least that's the form we are using now in build). Hence I think we need a regex match here.

build.psm1
```diff
@@ -1320,20 +1325,26 @@ function Publish-NuGetFeed
 {
     param(
         [string]$OutputPath = "$PSScriptRoot/nuget-artifacts",
-        [Parameter(Mandatory=$true)]
-        [string]$VersionSuffix
+        [Parameter(ParameterSetName = "ReleaseTag")]
+        [ValidatePattern("^v\d+\.\d+\.\d+(-\w+\.\d+)?$")]
+        [ValidateNotNullOrEmpty()]
+        [string]$ReleaseTag
```
daxian-dbw: Let's make this parameter mandatory. `-VersionSuffix` was mandatory.

build.psm1
```diff
@@ -1320,20 +1325,26 @@ function Publish-NuGetFeed
...
-        ## Workaround:
-        ##   dotnet restore /p:VersionSuffix=<suffix> # Bake the suffix into project.assets.json
-        ##   dotnet pack --version-suffix <suffix>
-        $TopProject = (New-PSOptions).Top
-        dotnet restore $TopProject "/p:VersionSuffix=$VersionSuffix"
+    ## NuGet/Home #3953, #4337 -- dotnet pack - version suffix missing from ProjectReference
+    ## Workaround:
+    ##   dotnet restore /p:VersionSuffix=<suffix> # Bake the suffix into project.assets.json
+    ##   dotnet pack --version-suffix <suffix>
+    ## We update 'project.assets.json' files with new version tag value by 'GetPSCoreVersionFromGit' target.
```
daxian-dbw: The comments is outdated now, as we don't use --version-suffix anymore.

build.psm1
```diff
@@ -1320,20 +1325,26 @@ function Publish-NuGetFeed
...
+    $TopProject = (New-PSOptions).Top
+    if ($ReleaseTag) {
```
daxian-dbw: Let's make `-ReleaseTag` mandatory. So the `if/else` is unnecessary now.

build.psm1
```diff
@@ -1351,8 +1362,8 @@ function Publish-NuGetFeed
 'Microsoft.WSMan.Runtime',
 'Microsoft.PowerShell.SDK'
         ) | ForEach-Object {
-            if ($VersionSuffix) {
-                dotnet pack "src/$_" --output $OutputPath --version-suffix $VersionSuffix /p:IncludeSymbols=true
+            if ($ReleaseTag) {
```
daxian-dbw: Same here.

build.psm1
iSazonov: Are we going to do daily builds? If so we shouldn't make it mandatory.

PowerShell.Common.props
iSazonov: It seems a best practice for csproj is  to use variables everywhere.
Here we can implement the idea - use another git version that is out off PATH. To get this we should do:
```
<GitExe Condition = " GitExe =='' ">git</GitExe>
```
What is your thoughts? Should we make our csproj-s power?

PowerShell.Common.props
iSazonov: It seems we can remove `WorkingDirectory` too because MSBuild make `cd` to project directory.

PowerShell.Common.props
iSazonov: Replaced with `StandardOutputImportance="Low"`

PowerShell.Common.props
iSazonov: Good catch!

build.psm1
```diff
@@ -1320,20 +1325,26 @@ function Publish-NuGetFeed
 {
     param(
         [string]$OutputPath = "$PSScriptRoot/nuget-artifacts",
-        [Parameter(Mandatory=$true)]
-        [string]$VersionSuffix
+        [Parameter(ParameterSetName = "ReleaseTag")]
```
daxian-dbw: This parameter attribute seems not necessary.

build.psm1
daxian-dbw: yes we generate nuget packages from daily build, and you need to change the related code in appveyor.psm1: https://github.com/PowerShell/PowerShell/blob/master/tools/appveyor.psm1#L452

The existing code has predefined suffix for daily build, you need to think how to make it continue to work, or how to alter it according to this PR. /cc @TravisEz13 for input.

PowerShell.Common.props
iSazonov: Sorry. Fixed.

PowerShell.Common.props
daxian-dbw: I think using `git` directly makes the xml file a little simpler. I don't think we need to support using a different `git` that is not in PATH.

PowerShell.Common.props
daxian-dbw: so even I'm running `dotnet build <path-to-s.m.a.csproj>` from a totally different location, it still guarantees that the git command will run in the `src\s.m.a` folder?

PowerShell.Common.props
iSazonov: Yes, I tried minimize `Condition`-s.
Fixed.

PowerShell.Common.props
iSazonov: Removed.

PowerShell.Common.props
iSazonov: Line removed.

PowerShell.Common.props
iSazonov: I run `Publish-NuGetFeed` (it don't change the current directory) from non repo directory and it works. (I can not find link to MSDN docs although I read about this. It seems it was an article about MSBuild targets).
Also we may have an issue here when later we'll implement the building from Visual Studio.

iSazonov: Maybe related https://github.com/dotnet/cli/issues/4473#issuecomment-256502969

build.psm1
iSazonov: Leave only my last comment.

build.psm1
iSazonov: I see there we call `Start-PSPackage` - we need fix it too. And in `Start-PSPackage` we have another code path for nuget packages - I wonder why we have two code path to generate nuget packages? Because `Publish-NuGetFeed` was not moved in Packaging.psm1 I thought it is not used on CI but it is used.

build.psm1
iSazonov: I added this because we could return `[string]$VersionSuffix` to _temporary_ fix daily builds.

build.psm1
daxian-dbw: The nuget package created in `Start-PSPackage` is not individual assembly package to be used for development, but a package that contains all powershell core bits -- it's similar to the `.zip` packages we generated for release and can be installed directly by using `PackageManagement` module.

PowerShell.Common.props
daxian-dbw: > Also we may have an issue here when later we'll implement the building from Visual Studio.

What is the issue? Is it that the VS may use a different default working directory?
I feel this 'default working directory' is too implicit and most of us wouldn't know this. I think we should use `WorkingDirectory=$(MSBuildProjectDirectory)` to make it explicit.

iSazonov: Yes, VS can. Mentioned CLI Issue imply about this.

Will fix.

tools/appveyor.psm1
```diff
@@ -451,24 +451,26 @@ function Invoke-AppveyorFinish
 
         if ($env:APPVEYOR_REPO_TAG_NAME)
         {
-            # ignore the first part of semver, use the preview part
-            $preReleaseVersion = ($env:APPVEYOR_REPO_TAG_NAME).Split('-')[1]
+            $preReleaseVersion = "v$($env:APPVEYOR_REPO_TAG_NAME)"
```
daxian-dbw: Isn't the tag name already has `'v'` prefixed?

tools/appveyor.psm1
```diff
@@ -451,24 +451,26 @@ function Invoke-AppveyorFinish
...
 
-            $preReleaseVersion = "$previewLabel-$($env:APPVEYOR_BUILD_NUMBER.replace('.','-'))"
+            $preReleaseVersion = "$previewPrefix-$previewLabel-$($env:APPVEYOR_BUILD_NUMBER.Replace('.','-'))"
         }
 
         # only publish to nuget feed if it is a daily build and tests passed
         if((Test-DailyBuild) -and $env:TestPassed -eq 'True')
         {
-            Publish-NuGetFeed -OutputPath .\nuget-artifacts -VersionSuffix $preReleaseVersion
+            Publish-NuGetFeed -OutputPath .\nuget-artifacts -ReleaseTag "/property:ReleaseTag=$preReleaseVersion"
```
daxian-dbw: > `-ReleaseTag "/property:ReleaseTag=$preReleaseVersion"`

This seems wrong ...

PowerShell.Common.props
```diff
@@ -1,10 +1,95 @@
...
+      <PSCoreAdditionalCommits>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[2].Value)</PSCoreAdditionalCommits>
+      <PSCoreCommitSHA>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[3].Value)</PSCoreCommitSHA>
+
+      <RegexSymVer>^((\d+).(\d+).(\d+))(?:-(.+))?</RegexSymVer>
+      <PSCorePrefixVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[1].Value)</PSCorePrefixVersion>
+      <PSCoreMajorVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[2].Value)</PSCoreMajorVersion>
+      <PSCoreMinorVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[3].Value)</PSCoreMinorVersion>
+      <PSCorePatchVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[4].Value)</PSCorePatchVersion>
+      <PSCoreLabelVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[5].Value)</PSCoreLabelVersion>
```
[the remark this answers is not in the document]
iSazonov: Yes, currently we don't use but can in future. I added its because we do "strange" 😄  manipulations with version strings in build scripts

PowerShell.Common.props
```diff
@@ -1,10 +1,95 @@
...
+          'BeforeBuild'                 - before 'Build' target
+  -->
+  <Target Name="GetPSCoreVersionFromGit"
+    BeforeTargets="_GenerateRestoreProjectSpec;GenerateNuspec;BeforeBuild"
+  >
+
+    <Exec Command='git describe --abbrev=60 --long'
+          ConsoleToMSBuild="true"
+          StandardOutputImportance="Low">
+      <Output TaskParameter="ConsoleOutput" PropertyName="PowerShellVersion" />
```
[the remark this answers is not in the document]
iSazonov: Yes. Fixed.

PowerShell.Common.props
```diff
@@ -1,10 +1,95 @@
...
+      <PSCorePrefixVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[1].Value)</PSCorePrefixVersion>
+      <PSCoreMajorVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[2].Value)</PSCoreMajorVersion>
+      <PSCoreMinorVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[3].Value)</PSCoreMinorVersion>
+      <PSCorePatchVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[4].Value)</PSCorePatchVersion>
+      <PSCoreLabelVersion>$([System.Text.RegularExpressions.Regex]::Match($(PSCoreBuildVersion), $(RegexSymVer)).Groups[5].Value)</PSCoreLabelVersion>
+
+      <PSCoreFormattedVersion Condition = "'$(ReleaseTag)' != '' or '$(PSCoreAdditionalCommits)' == '0'">$(PSCoreBuildVersion) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
+      <PSCoreFormattedVersion Condition = "'$(PSCoreFormattedVersion)' == ''">$(PSCoreBuildVersion) Commits: $(PSCoreAdditionalCommits) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
+
+      <ProductVersion>$(PSCoreFormattedVersion)</ProductVersion>
```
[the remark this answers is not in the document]
iSazonov: `InformationalVersion` is explicitly defined in `Microsoft.NET.GenerateAssemblyInfo.targets` from `Version`. `ProductVersion` is seems hard coded in AL task. So I explicitly set it here to exclude unpredictable behavior if AL task logic will be changed in any next dotnet update.

PowerShell.Common.props
```diff
@@ -1,10 +1,95 @@
...
+      <!--
+            Here we define explicitly 'Version' to set 'FileVersion' by 'GetAssemblyVersion' target in 'Microsoft.NET.GenerateAssemblyInfo.targets'.
```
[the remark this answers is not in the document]
iSazonov: It seems it will be obvious comment. Or you meant we should explicitly enumerate such properties which we use?

PowerShell.Common.props
```diff
@@ -1,10 +1,95 @@
...
+            Here we define explicitly 'InformationalVersion' because by default it is defined as 'Version' by 'GetAssemblyVersion' target in 'Microsoft.NET.GenerateAssemblyInfo.targets'.
+      -->
+      <Version>$(PSCoreBuildVersion)</Version>
+      <InformationalVersion>$(PSCoreFormattedVersion)</InformationalVersion>
+
+      <!--VersionPrefix>$(PSCorePrefixVersion)</VersionPrefix-->
+      <!--VersionSuffix>$(PSCoreLabelVersion)</VersionSuffix-->
```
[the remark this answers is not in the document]
iSazonov: Removed.

PowerShell.Common.props
```diff
@@ -1,10 +1,95 @@
...
+
+      <!--
+            We have explicitly assign 'PackageVersion'
+            because there is a bug: 'PackageVersion' is correctly assigned as 'Version' in 'NuGet.targets'
+            but then immediately redefined as '1.0.0'.
```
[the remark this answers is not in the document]
iSazonov: I don't open an Issue. What is right place?

build.psm1
```diff
@@ -1320,20 +1325,21 @@ function Publish-NuGetFeed
...
-        ##   dotnet pack --version-suffix <suffix>
-        $TopProject = (New-PSOptions).Top
-        dotnet restore $TopProject "/p:VersionSuffix=$VersionSuffix"
+    ## We update 'project.assets.json' files with new version tag value by 'GetPSCoreVersionFromGit' target.
+    $TopProject = (New-PSOptions).Top
+    if ($ReleaseTag) {
+        $ReleaseTag1 = $ReleaseTag -Replace '^v'
+        dotnet restore $TopProject "/property:ReleaseTag=$ReleaseTag1"
+    } else {
+        dotnet restore $TopProject
```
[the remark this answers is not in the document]
iSazonov: If we set new version git tag we should explicitly run  `dotnet restore` before locally build nuget packages.

build.psm1
```diff
@@ -1351,8 +1357,8 @@ function Publish-NuGetFeed
 'Microsoft.WSMan.Runtime',
 'Microsoft.PowerShell.SDK'
         ) | ForEach-Object {
-            if ($VersionSuffix) {
-                dotnet pack "src/$_" --output $OutputPath --version-suffix $VersionSuffix /p:IncludeSymbols=true
+            if ($ReleaseTag) {
+                dotnet pack "src/$_" --output $OutputPath "/property:ReleaseTag=$ReleaseTag1" /p:IncludeSymbols=true
```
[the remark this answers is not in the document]
iSazonov: Fixed. (I can not test this.)

tools/appveyor.psm1
iSazonov: I don't know - I can not test. I am only tying to follow previous logic.

tools/appveyor.psm1
iSazonov: Sorry. Fixed.

PowerShell.Common.props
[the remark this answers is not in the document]
daxian-dbw: I mean if you mention `FileVersion` in the comment here, then better mention that version properties that use `<Version>` as the default value. Otherwise, the comment means more like `FileVersion` is the only reason we explicitly define `<version>` here.

PowerShell.Common.props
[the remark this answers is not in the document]
daxian-dbw: https://github.com/dotnet/sdk should be the place for dotnet msbuild issues.

build.psm1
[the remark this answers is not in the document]
daxian-dbw: I see your point. You mean `dotnet pack` may use the existing artifacts produced from previous `dotnet restore` run and thus use stale version info. It makes sense. #Closed

PowerShell.Common.props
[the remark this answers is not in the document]
daxian-dbw: I think we only used the individual major, minor and patch numbers when generating MSI package, so it's hard for me to see we will somehow use them in .csproj files. I tend to make our build faster, but probably it won't make much difference. So I will leave this decision up to you.

PowerShell.Common.props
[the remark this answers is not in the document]
daxian-dbw: I guess it's better to be explicit, so let's keep the `<ProductVersion>`. But can we move this tag next to `<InformationalVersion>` instead of having them separated by the comments?

PowerShell.Common.props
```diff
@@ -1,10 +1,95 @@
...
+               PSCoreLabelVersion = $(PSCoreLabelVersion);
+               GitCDUP = $(GitCDUP);
+               RegexGitVersion = $(RegexGitVersion);
+               PSCoreFormattedVersion = $(PSCoreFormattedVersion);
+               ProductVersion = $(ProductVersion);
+               Version = $(Version);
+               ProjectVersion = '!$(ProjectVersion)!'
+               InformationalVersion = $(InformationalVersion);
+              "
+        Overwrite="true"/-->
```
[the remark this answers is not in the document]
daxian-dbw: This debug section still needs to be cleaned up, for example, `GitCDUP` is gone now.

build.psm1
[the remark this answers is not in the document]
daxian-dbw: Why can't this be tested? `IncludeSymbols=true` means symbol packages is demanded as well, so if you use `Publish-NuGetFeed -ReleaseTag1 v6.0.0-beta.10`, the expected result is both `beta.10.nupkg` and `beta.10.symbols.nupkg` are generated.

tools/appveyor.psm1
daxian-dbw: @TravisEz13 can you comment on this one? What would `$env:APPVEYOR_REPO_TAG_NAME` be like?
@iSazonov I think the next commit you submit should be marked with '[Feature]', so that we can see how the `nupkg` generation works in full build.

PowerShell.Common.props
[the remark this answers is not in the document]
iSazonov: We can comment this so that we don't waste time on creating this code if we'll  need it in future.

PowerShell.Common.props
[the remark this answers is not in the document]
iSazonov: Moved.

PowerShell.Common.props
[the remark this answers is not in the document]
iSazonov: I thought there is many properties that depended on `<Version>` but it's not. 
Added ` AssemblyVersion ` in the comment.

PowerShell.Common.props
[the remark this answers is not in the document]
iSazonov: Thanks! Issue https://github.com/dotnet/sdk/issues/1557

PowerShell.Common.props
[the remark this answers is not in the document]
iSazonov: Cleaned up.

tools/appveyor.psm1
iSazonov: [Feature] - done.

iSazonov:
Web tests failed - it seems it is not related with the PR.

PowerShell.Common.props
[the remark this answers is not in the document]
daxian-dbw: That sounds good to me :)

PowerShell.Common.props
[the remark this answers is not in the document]
daxian-dbw: Thank you!

PowerShell.Common.props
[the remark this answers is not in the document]
daxian-dbw: #Closed

daxian-dbw:
The changes to `build.psm1` and `PowerShell.Common.props` are mostly good. I asked @TravisEz13 to review the changes in AppVeyor.psm1 as he is more experienced in the full build.

I did find that the full build run in CI doesn't generate the powershell packages (msi, nupkg and zip) like our daily build (look in the `artifacts` tab). I don't know how to check the individual assembly nuget packages.

TravisEz13:
Packaging is broken:  See https://ci.appveyor.com/project/PowerShell/powershell/build/6.0.0-beta.6-4928#L9098

tools/appveyor.psm1
TravisEz13: > contains tag name for builds started by tag; otherwise this variable is undefined
I don't think we ever do this, unless AppVeyor counts a push with a tag.

daxian-dbw: I see. So if an AppVeyor build is triggered by pushing a tag, then this environment variable will be set to be the tag.
So we don't need to prefix with a `'v'` here,

daxian-dbw:
@TravisEz13 Now that `Publish-NugetFeed` has `-VersionSuffix` replaced by `-ReleaseTag`, the value provided when running `Publish-NugetFeed` has to comply with the release tag format pattern `'^v\d+\.\d+\.\d+(-\w+\.\d+)?$'`. So we cannot use the current daily build suffix now. I changed that to the format like this `v6.0.0-dailybeta5.3453`. Let me know if you have concerns about that.

TravisEz13:
No concerns with that change

TravisEz13:
Windows artifacts look good.

daxian-dbw:
Thanks @TravisEz13. Then I will consider this PR to be almost ready. I will update some of the comments, and then it would be good to go.

PowerShell.Common.props
```diff
@@ -1,10 +1,97 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <!--
+        We should *explicitly* assign version properties in this target
+        because:
+          1. By default the version properties are defined in global 'PropertyGroup' (not target!) in 'Microsoft.NET.DefaultAssemblyInfo.targets'
+          2. Properties in global 'PropertyGroups' are assigned before any targets
+          3. and cannot be conditionally excecuted as targets (after this target).
```
daxian-dbw: > and cannot be conditionally excecuted as targets (after this target).

First, a typo I didn't notice in my last change: excecuted  -> executed.
Second, I have trouble understanding this sentence accurately. Can you please reword this sentence a bit to make it easier to understand?
And one more thing, what does `global PropertyGroup` mean? Is it different from the `PropertyGroup` we used in this file?

daxian-dbw (dismissed):

PowerShell.Common.props
iSazonov: Typo fixed.
Under "global" I meant that they affect the entire project. Removed.
I tryed to reword but ... - feel free to rewrite.

daxian-dbw:
@iSazonov Can you please take a look and see if my re-write reflects what you think?

iSazonov:
Yes, thanks! Look very good.

iSazonov:
Get-ComputerInfo test failed in AppVeyor CI.

daxian-dbw:
The test failure is not related to this change. Comparing with the last successful [full build CI](https://ci.appveyor.com/project/PowerShell/powershell/build/6.0.0-beta.6-5006), the only changes we have are comment changes. So it's OK to merge this PR.

daxian-dbw (approved):

iSazonov:
@daxian-dbw @TravisEz13 Thanks for review and approvement!

Should we continue to improve csproj files? I think it's worth moving a stable logic into them.

daxian-dbw:
@iSazonov Thanks for your hard work on it!
I'm a little hesitating about aggressively moving build logic to `.csproj` files. I think #3690 deserves higher priority -- now that we have version attributes embedded in the assemblies, we can set up the `CommitId` and `PSVersion` in C#.

iSazonov:
@daxian-dbw My question was more about #3690 - I'll continue.

iSazonov:
@daxian-dbw What is your thoughts - should we set up the  CommitId  and  PSVersion based on assembly  version attributes or generate strong typed constant?
The first method is simpler and more [approved](https://github.com/PowerShell/PowerShell/pull/3690#issuecomment-301105191) but not protected like powershell.version file.
Second is more strong but can result in frequent recompilation of SMA.

daxian-dbw:
I vote for getting the needed info from the attributes 😄

Issue #3267, opened 2017-03-06
Webcmdlets should parse the <html><head><meta charset="foo"> attribute for the correct encoding if not in http header


SteveL-MSFT:
Some websites do not populate the `charset` property of the `content-type` header so characters aren't rendered correctly.  Suggestion is to expose a `-charset` parameter, however the user still needs to know the expected charset.  Advanced users today can do the encoding translation in script.  `utf-8` probably works in most cases, so not entirely sure how useful this will be to expect the user to know ahead of time the correct charset.

See discussion from https://github.com/PowerShell/PowerShell/issues/3126 for details on how this came about

iSazonov:
Point for the fix is [here](https://github.com/PowerShell/PowerShell/blob/c74b2a7670c79b02bb526c0998720b93ac385d03/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs#L512)

iSazonov:
I believe we should discuss here a default encoding for Web cmdlets.
Currently [we use RFC 2616 "ISO-8859-1" ](https://github.com/PowerShell/PowerShell/blob/77c622eeea474e21af0ea954836d49c53f71f5b1/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/ContentHelper.Common.cs#L20)
In fact, browsers for a long time use Windows-1252.
And HTML5 default is UTF-8.

Perhaps we should also use Windows 1252.

chuanjiao10:
There are all like powershell fans,do not be perfunctory.
I disagree "-CharSet", do not push this question to the users.
 Think:
 If the browser does not recognize 【meta charset="gb2312"】,every time you need to manually select   ...

1 Receive raw html string
2 Find out after 'charset=' strings.
3 use 【some charset】 encoding raw html，and out to console  
2 and 3 are so hard that powershell can not do it?

iSazonov:
I looked Web cmdlets and found that we use `ContentType` parameter to encode a request (Content-Type header can contains `charset` value). If we don't specify a charset in `ContenType` we use default charset ISO-8859-1.
For decoding a response we use a charset from `ContentType` of the response. If a server return no `ContentType` we use the same default charset ISO-8859-1.

So we should treat `-CharSet` as `-ResponseCharSet`.

[...]

PR #4184, opened 2017-07-04
Added Meta, Charset, and Transitional parameters to ConvertTo-HTML

ergo3114:
Related #3267

Added the Meta, Chartset, and Transitional parameters to ConvertTo-HTML

**Meta** Allows the user to insert `<meta>` tag information such as refresh, author, and keywords
**Charset** Sets the html encoding of the page, denoted by the `<charset>` html tag
**Transitional** A switch parameter that allows for the insertion of XHTML Transitional DTD DOCTYPE, the default being XHTML Strict DTD

Edit: Documentation can be found [here](https://github.com/PowerShell/PowerShell-Docs/pull/1432)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -376,9 +437,33 @@ protected override void BeginProcessing()
...
+                    WriteObject("<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Transitional//EN\"  \"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\">");
+                }
                 WriteObject("<html xmlns=\"http://www.w3.org/1999/xhtml\">");
+                if(_metaSpecified)
+                {
+                    foreach(string s in _meta.Keys)
+                    {
+                        if(s != "content-type" && s != "default-style" && s !="refresh")
+                        {
+                            WriteObject("<meta name=\"" + s + "\" content=\"" + _meta[s] + "\">");
```
TravisEz13: should we encode the key and value?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -376,9 +437,33 @@ protected override void BeginProcessing()
...
+                        }
+                        else {
+                            WriteObject("<meta http-equiv=\"" + s + "\" content=\"" + _meta[s] + "\">");
+                        }
+                    }
+                }
                 WriteObject("<head>");
+                if(_charsetSpecified)
+                {
+                    WriteObject("<meta charset=\"" + _charset + "\">");
```
TravisEz13: Should probably validate or encode the character set.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -376,9 +437,33 @@ protected override void BeginProcessing()
...
+                if(_metaSpecified)
+                {
+                    foreach(string s in _meta.Keys)
+                    {
+                        if(s != "content-type" && s != "default-style" && s !="refresh")
+                        {
+                            WriteObject("<meta name=\"" + s + "\" content=\"" + _meta[s] + "\">");
+                        }
+                        else {
+                            WriteObject("<meta http-equiv=\"" + s + "\" content=\"" + _meta[s] + "\">");
```
TravisEz13: Should we allow these or should it be another parameter like `charset`

TravisEz13:
@JamesWTruher  Can you review the character set parameter here?

TravisEz13 (dismissed):
Thanks for the PR.  I have a couple of comments.  Please review and respond.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
ergo3114: Validation would be great. I don't see too many changes in the charset list but it would be great if the validation were dynamic pulling from w3schools or something. I was trying to find documentation on what the browser does when a non standard charset is specified; such as "foo". I believe it analyzes the html and determines the best or just defaults to UTF-8 for HTML5.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
ergo3114: It looks like you can validate the meta tag. [w3schools](https://www.w3schools.com/tags/tag_meta.asp) shows the schema for the meta tag.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
ergo3114: What would encoding gain us here? I am not familiar with that.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
TravisEz13: Could you use ValidatePattern or ValidateScript to implement it?  https://blogs.technet.microsoft.com/heyscriptingguy/2011/01/11/validate-powershell-parameters-before-running-the-script/

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
TravisEz13: I'm sure there are illegal characters.  Even a validation that validates the user hasn't specified something illegal would be better than nothing.

ergo3114: I will put something in and resubmit it.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -228,6 +228,68 @@ public string[] PreContent
...
+        public Hashtable Meta
+        {
+            get
+            {
+                return _meta;
+            }
+            set
+            {
+                _meta = value;
+                _metaSpecified = true;
```
SteveL-MSFT: I believe the typical pattern is to use [InvocationInfo.BoundParameters](https://msdn.microsoft.com/en-us/library/system.management.automation.invocationinfo.boundparameters(v=vs.85).aspx)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -376,9 +438,33 @@ protected override void BeginProcessing()
...
                 WriteObject("<head>");
+                if(_charsetSpecified)
+                {
+                    WriteObject("<meta charset=\"" + _charset + "\">");
+                }
+                if(_metaSpecified)
+                {
+                    foreach(string s in _meta.Keys)
+                    {
+                        if(s == "content-type" && s == "default-style" && s =="refresh")
```
SteveL-MSFT: Shouldn't this be `||`?  In any case, there's more `http-equiv` than this list.  Perhaps we should just have a separate parameter for `http-equiv` meta tags and let the user put whatever they want in it?  Also, I believe the names are case-insensitive so `Content-Type` is valid.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -228,6 +228,68 @@ public string[] PreContent
...
+        }
+        private Hashtable _meta;
+        private bool _metaSpecified = false;
+
+        /// <summary>
+        /// Specifies the charset encoding for the HTML document
+        /// </summary>
+        [Parameter(ParameterSetName = "Page")]
+        [ValidateNotNullOrEmpty]
+        [ValidateSet("UTF-8", "ISO-8859-1","ASCII","ANSI",IgnoreCase=true)]
```
SteveL-MSFT: Technically, the allowed set of (legacy) encodings is quite large although HTML5 I believe has standardized on UTF-8.  So perhaps this shouldn't have a `ValidateSet`?

SteveL-MSFT:
Please add some tests

SteveL-MSFT (changes requested):
Please add some tests

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
ergo3114: I have removed to the ValidateSet.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
ergo3114: This has been changed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
ergo3114: I was unable to get InvocationInfo.BoundParameters to work. I read over InvocationInfo.cs and ImplicitRemotingCommands.cs to see if I can make sense of it but I am at a lose here.

iSazonov: We can use an auto property and `Meta == null` to check if user assign a value.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -228,6 +228,67 @@ public string[] PreContent
...
+        public string Charset
+        {
+            get
+            {
+                return _charset;
+            }
+            set
+            {
+                _charset = value.ToUpper();
+                _charsetSpecified = true;
```
iSazonov: Why do we convert a charset to upper case?
It seems we can remove this and use auto property.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -228,6 +228,67 @@ public string[] PreContent
...
+        public SwitchParameter Transitional
+        {
+            get
+            {
+                return _transitional;
+            }
+            set
+            {
+                _transitional = value;
+            }
```
iSazonov: We can use an auto property for a switch parameter without getter/setter.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
SteveL-MSFT: That should be fine, too

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
iSazonov: Closed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
iSazonov: Closed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
iSazonov: @ergo3114 Please address this.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -228,6 +228,56 @@ public string[] PreContent
...
+        }
+        private Hashtable _meta;
+        private bool _metaSpecified = false;
+
+        /// <summary>
+        /// Specifies the charset encoding for the HTML document
+        /// </summary>
+        [Parameter(ParameterSetName = "Page")]
+        [ValidateNotNullOrEmpty]
+        public string Charset
```
TravisEz13: Can we add a [ValidatePatternAttribute](https://msdn.microsoft.com/en-us/library/ms714454%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396) and verify that the parameter is a valid Charset.  For example, charset should not have a `"`, `<` or `>` in the parameter.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -376,9 +426,33 @@ protected override void BeginProcessing()
...
+                }
                 WriteObject("<html xmlns=\"http://www.w3.org/1999/xhtml\">");
                 WriteObject("<head>");
+                if(_charsetSpecified)
+                {
+                    WriteObject("<meta charset=\"" + _charset + "\">");
+                }
+                if(_metaSpecified)
+                {
+                    foreach(string s in _meta.Keys)
```
TravisEz13: validate the meta keys and value against a regex before adding them to the HTML.

TravisEz13 (changes requested):
Please remove `/1`.  Binary encoded files should not be added.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
ergo3114: Certainly can. Are you concerned with someone putting HTML into the parameter and injecting malicious code?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
TravisEz13: `http-equiv:refresh` should probably not be allowed as it has no purpose in this scenario

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
TravisEz13: Not necessarily malicious, just malformed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
ergo3114: I see this as two ways to do the same thing. What are you looking to gain from setting a default value versus checking a flag? Just curious.

SteveL-MSFT: A quick search shows we already have inconsistency so I'm fine staying with this pattern

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -376,9 +426,33 @@ protected override void BeginProcessing()
...
+                    {
+                        if(s == "content-type" || s == "default-style" || s =="refresh")
+                        {
+                            WriteObject("<meta http-equiv=\"" + s + "\" content=\"" + _meta[s] + "\">");
+                        }
+                        else if (s == "application-name" || s == "author" || s == "description" || s == "generator" || s == "keywords" || s == "viewport") {
+                            WriteObject("<meta name=\"" + s + "\" content=\"" + _meta[s] + "\">");
+                        }
+                    }
```
SteveL-MSFT: Perhaps we should output a warning for any other meta that we didn't include in case the user made a typo

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
ergo3114: What do you mean?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
ergo3114: I put in an error and have pushed to my repo. If you have an example of a warning I will gladly change it to a warning.

ergo3114:
Documentation updated and added per [PowerShell-Docs #1432](https://github.com/PowerShell/PowerShell-Docs/pull/1432)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
TravisEz13: This should be static html.  Allowing it to be refreshed would have no purpose.  refresh can also be used to redirect.  I also don't see the purpose of that.

ergo3114: I see what you mean and have changed my code to reflect that.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
SteveL-MSFT: Rather than doing the validation above and going through the hashtable twice, I think it might be better to change these lines to a switch statement and have it default to an error

SteveL-MSFT (changes requested):

ergo3114:
Documentation for this PR was merged July 31th. [1432](https://github.com/PowerShell/PowerShell-Docs/pull/1432)

TravisEz13:
Looks better to be, other than the commented code should be removed.

ergo3114:
Removed comments. Keep me posted with anything else you find and I'll be glad to work on your request.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -376,9 +427,44 @@ protected override void BeginProcessing()
...
                 WriteObject("<head>");
+                if(_charsetSpecified)
+                {
+                    WriteObject("<meta charset=\"" + _charset + "\">");
+                }
+                if(_metaSpecified)
+                {
+                    foreach(string s in _meta.Keys)
+                    {
+                        switch(s){
```
SteveL-MSFT: Based on what I could find, meta tag names are not case sensitive (although best practice is to use lowercase), so it may be good to throw in a `ToLower()`, but you're using a hashtable which is case sensitive so the same the same tag could show up twice.  Perhaps add code to check if a particular tag is already used and return an error.

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Html.Tests.ps1
```diff
@@ -116,5 +116,65 @@ After the object
...
+<tr><th>Name</th><th>Age</th><th>Friends</th></tr>
+<tr><td>John Doe</td><td>42</td><td>System.Object[]</td></tr>
+</table>
+</body></html>
+"@
+        $returnString | Should Be $expectedValue
+    }
+
+    It "Test ConvertTo-HTML meta with invalid properties should throw"{
+        { ($customObject | ConvertTo-HTML -Meta @{"authors"="John Doe";"keyword"="PowerShell,PSv6"}) -join $newLine } | Should Throw
```
JamesWTruher: it would be better to check for a specific errorid if one is expected (and consistent)

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Html.Tests.ps1
```diff
@@ -116,5 +116,65 @@ After the object
...
+<tr><th>Name</th><th>Age</th><th>Friends</th></tr>
+<tr><td>John Doe</td><td>42</td><td>System.Object[]</td></tr>
+</table>
+</body></html>
+"@
+        $returnString | Should Be $expectedValue
+    }
+
+    It "Test ConvertTo-HTML transitional"{
+        $returnString = ($customObject | ConvertTo-HTML -Transitional) -join $newLine
```
JamesWTruher: do you want to validate the entire document or just the first line? If the later, would:
```
$result = $customObject | ConvertTo-HTML -Transitional | Select-Object -First 1
$result | should be '<!DOCTYPE html PUBLIC...
```
be sufficient?

ergo3114:
My development computer crashed and I am having issues getting things set back up. I will continue updating this PR when I fix my issues.

TravisEz13:
@ergo3114 Thanks for the update.  It's helpful for us to know the PR is not abandoned.

ergo3114:
I believe I am back in the game with continuing this PR. I will work on items that were left unattended in the coming days.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -228,6 +228,57 @@ public string[] PreContent
...
+        private bool _charsetSpecified = false;
+
+        /// <summary>
+        /// When this switch statement is specified,
+        /// it will change the DOCTYPE to XHTML Transitional DTD
+        /// </summary>
+        /// <returns></returns>
+        [Parameter(ParameterSetName = "Page")]
+        [ValidateNotNullOrEmpty]
+        public SwitchParameter Transitional;
```
dantraMSFT: I think you want { get; set; } here instead of a field.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -364,9 +415,49 @@ protected override void BeginProcessing()
...
+                                case "viewport":
+                                    WriteObject("<meta name=\"" + s + "\" content=\"" + _meta[s] + "\">");
+                                    break;
+                                default:
+                                    Exception exc = new NotSupportedException(StringUtil.Format(ConvertHTMLStrings.MetaPropertyNotFound, s));
+                                    ThrowTerminatingError (new ErrorRecord(exc, "MetaPropertyNotFound", ErrorCategory.ParserError, null));
+                                    break;
+                            }
+                            useditems.Add(s);
+                        }
```
SteveL-MSFT: seems like we should give a warning if we are dropping duplicates

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
ergo3114: I have changed it to a warning and decided to keep the meta tags that don't match the switch.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/convert-HTML.cs
```diff
@@ -461,8 +461,19 @@ protected override void BeginProcessing()
                                     WriteObject("<meta name=\"" + s + "\" content=\"" + _meta[s] + "\">");
                                     break;
                                 default:
-                                    Exception exc = new NotSupportedException(StringUtil.Format(ConvertHTMLStrings.MetaPropertyNotFound, s));
-                                    ThrowTerminatingError (new ErrorRecord(exc, "MetaPropertyNotFound", ErrorCategory.ParserError, null));
+                                    //Exception exc = new NotSupportedException(StringUtil.Format(ConvertHTMLStrings.MetaPropertyNotFound, s));
+                                    MshCommandRuntime mshCommandRuntime = this.CommandRuntime as MshCommandRuntime;
+                                    string Message = "Accepted meta properties are content-type, default-style, application-name, author, description, generator, keywords, x-ua-compatible, and viewport. The meta pair: " + s + " and " + _meta[s] + " may not function correctly.";
```
SteveL-MSFT: Add a new string in the resx file so that it can be localized

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Html.Tests.ps1
```diff
@@ -136,8 +136,9 @@ After the object
         $returnString | Should Be $expectedValue
     }
 
-    It "Test ConvertTo-HTML meta with invalid properties should throw"{
-        { ($customObject | ConvertTo-HTML -Meta @{"authors"="John Doe";"keywords"="PowerShell,PSv6"}) -join $newLine } | Should Throw "authors is not a supported meta property. Accepted meta properties are content-type, default-style, application-name, author, description, generator, keywords, x-ua-compatible, and viewport."
+    It "Test ConvertTo-HTML meta with invalid properties should throw warning" {
+        ($customObject | ConvertTo-HTML -Meta @{"authors"="John Doe";"keywords"="PowerShell,PSv6"} 3>&1) -match "Accepted meta properties are" | Should Be $true
```
SteveL-MSFT: Since the message could be localized, you probably just need to check that the warning contains the Meta tags you used in the test

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Html.Tests.ps1
```diff
@@ -116,5 +116,52 @@ After the object
...
+<tr><th>Name</th><th>Age</th><th>Friends</th></tr>
+<tr><td>John Doe</td><td>42</td><td>System.Object[]</td></tr>
+</table>
+</body></html>
+"@
+        $returnString | Should Be $expectedValue
+    }
+
+    It "Test ConvertTo-HTML meta with invalid properties should throw warning" {
+        ($customObject | ConvertTo-HTML -Meta @{"authors"="John Doe";"keywords"="PowerShell,PSv6"} 3>&1) -match "Accepted meta properties are" | Should Be $true
```
SteveL-MSFT: Since the warning message can be localized, you shouldn't check against the English string.  Instead, check that the warning message contains the meta pair.

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

TravisEz13 (approved):

Issue #3807, opened 2017-05-18
Remove UTC and SQM telemetry code


SteveL-MSFT:
The code base inherited from Windows UTC and SQM telemetry data points that are no longer relevant for cross platform now that we've adopted AppInsights.  This code needs to be removed as it's counting against our code coverage

SteveL-MSFT:
@JamesWTruher we should remove this when we close on new telemetry plan to see what logic in the legacy telemetry we can reuse

JamesWTruher:
thank you

PR #4190, opened 2017-07-05
Remove UTC and SQM telemetry code

JamesWTruher:
Implementation of #3807 

The historical telemetry code uses WindowsUTC which is not applicable to non-Windows platforms and is superseded by our ApplicationInsight telemetry. `#if LEGACYTELEMETRY` was used to make it easier to track when we add our next round of telemetry.

Change should not be visible to the user, unless they were inspecting the eventlog looking for it. Any SQM telemetry appears to have been disabled in June 2016.

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
@@ -46,11 +46,11 @@ namespace Microsoft.PowerShell
         :
         PSHost,
         IDisposable,
+        IHostSupportsInteractiveSession
 #if LEGACYTELEMETRY
+        ,
         IHostSupportsInteractiveSession,
         IHostProvidesTelemetryData
-#else
-        IHostSupportsInteractiveSession
 #endif
     {
         #region static methods
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
@@ -46,12 +46,10 @@ namespace Microsoft.PowerShell
         :
         PSHost,
         IDisposable,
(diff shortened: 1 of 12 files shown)


src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -44,8 +46,12 @@ internal sealed partial class ConsoleHost
         :
         PSHost,
         IDisposable,
+#if LEGACYTELEMETRY
         IHostSupportsInteractiveSession,
         IHostProvidesTelemetryData
```
SteveL-MSFT: Can you rearrange the order to avoid the `#else`?

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -44,8 +46,12 @@ internal sealed partial class ConsoleHost
         :
         PSHost,
         IDisposable,
+        IHostSupportsInteractiveSession
```
SteveL-MSFT: I think this should be:
```c#
#if LEGACYTELEMETRY
  IHostProvidesTelemetryData,
#endif
  IHostSupportsInteractiveSession
```

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):
LGTM

daxian-dbw:
The old telemetry implementation are still kept in code base (PSTelemetryMethods.cs, PSTelemetryWrapper.cs). Shall we delete them in this PR?

SteveL-MSFT:
@daxian-dbw current thinking is to keep it for now.  We'll likely add more telemetry in 6.1.0 and we can see what we can reuse from the old telemetry and clean up the code completely at that time.

Issue #3325, opened 2017-03-13
Unexpected module loading through NuGet


chunqingchen:
Steps to reproduce
------------------
Get VS 2017 RTW. 
Install an “experimental” VSIX of NuGet like the one here by double clicking on it:
\\nuget\nuget\Share\Drops\CI\NuGet.Client\dev\2381\artifacts\VS15\NuGet.Tools.vsix
Open PackageManagerConsole…

Expected behavior
-----------------
correct version of  'NuGet.PackageManagement.PowerShellCmdlets.dll' is loaded

Actual behavior
---------------
the wrong NuGet.PackageManagement.PowershellCmdlets.dll is pulled in

Environment data
----------------
PSVersion                      5.1.14393.693                                                                                                                                                                        
PSEdition                      Desktop                                                                                                                                                                              
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                                                                                              
BuildVersion                   10.0.14393.693                                                                                                                                                                       
CLRVersion                     4.0.30319.42000                                                                                                                                                                      
WSManStackVersion              3.0                                                                                                                                                                                  
PSRemotingProtocolVersion      2.3                                                                                                                                                                                  
SerializationVersion           1.1.0.1            

```

SteveL-MSFT:
Seems like this is an issue with VS2017 which hosts PowerShell and has their own PackageManagement module?

alpaix:
NuGet Package Manager Console (PMC) hosts PowerShell for running NuGet cmdlets and executing install/uninstall/init package scripts. It imports `<extension root>\Modules\NuGet\NuGet.psd1` module which in its turn depends on a nested module `NuGet.PackageManagement.PowerShellCmdlets.dll`. 

The last one is loaded from wrong location when working with "experimental" VS instance.

nkolev92:
Any updates here?
ETA?

daxian-dbw:
Close via #5161

PR #4196, opened 2017-07-07
PS should try to load the assembly from file path first before loading from assemblyName

chunqingchen:
Fix issue #3325 

Summary of the issue:

visual studio cannot import the right assembly contains in their test vsix package through package manage console. It will always load the assembly contains in the vs location

Root cause of the issue:
loadassembly funtion will always try to load the assembly based on the assembly name first, then try the filename. In this case, no matter what assembly path psd1 file gives, ps will ignore and load the assembly in the vs cache.

Fix:
Since the filename is given, we should try it first before going to the assembly name.

The test scenario of this issue is very specific and does not contains in our ps enviroment. Attached the screen shot before/after the fix.

Before the fix:
image

After the fix:
image

### DIFF
--- a/src/System.Management.Automation/engine/ExecutionContext.cs
+++ b/src/System.Management.Automation/engine/ExecutionContext.cs
@@ -1323,14 +1323,37 @@ namespace System.Management.Automation
         [SuppressMessage("Microsoft.Reliability", "CA2001:AvoidCallingProblematicMethods", MessageId = "System.Reflection.Assembly.LoadFrom")]
         internal static Assembly LoadAssembly(string name, string filename, out Exception error)
         {
-            // First we try to load the assembly based on the given name
+            // First we try to load the assembly based on the filename
 
             Assembly loadedAssembly = null;
             error = null;
 
-            string fixedName = null;
-            if (!String.IsNullOrEmpty(name))
+            if (!String.IsNullOrEmpty(filename))
             {
+                try
+                {
+                    loadedAssembly = Assembly.LoadFrom(filename);
+                }
+                catch (FileNotFoundException fileNotFound)
+                {
+                    error = fileNotFound;
+                }
+                catch (FileLoadException fileLoadException)
+                {
+                    error = fileLoadException;
+                }
+                catch (BadImageFormatException badImage)
+                {
+                    error = badImage;
+                }
+                catch (SecurityException securityException)
+                {
+                    error = securityException;
+                }
+            }
+            else if (!String.IsNullOrEmpty(name))
+            {
+                string fixedName = null;
                 // Remove the '.dll' if it's there...
                 fixedName = name.EndsWith(".dll", StringComparison.OrdinalIgnoreCase)
                                 ? Path.GetFileNameWithoutExtension(name)
@@ -1357,76 +1380,15 @@ namespace System.Management.Automation
                 catch (BadImageFormatException badImage)
                 {
                     error = badImage;
-                    return null;
                 }
                 catch (SecurityException securityException)
                 {
                     error = securityException;
-                    return null;
                 }
             }
 
-            if (loadedAssembly != null)
-                return loadedAssembly;
-
-            if (!String.IsNullOrEmpty(filename))
-            {
-                error = null;
-
-                try
-                {
-                    loadedAssembly = Assembly.LoadFrom(filename);
-                    return loadedAssembly;
-                }
-                catch (FileNotFoundException fileNotFound)
-                {
-                    error = fileNotFound;
-                }
-                catch (FileLoadException fileLoadException)
-                {
-                    error = fileLoadException;
-                    return null;
-                }
-                catch (BadImageFormatException badImage)
-                {
-                    error = badImage;
-                    return null;
-                }
-                catch (SecurityException securityException)
-                {
-                    error = securityException;
-                    return null;
-                }
-            }
-
-#if !CORECLR// Assembly.LoadWithPartialName is not in CoreCLR. In CoreCLR, 'LoadWithPartialName' can be replaced by Assembly.Load with the help of AssemblyLoadContext.
-            // Finally try with partial name...
-            if (!String.IsNullOrEmpty(fixedName))
-            {
-                try
-                {
-                    // This is a deprecated API, use of this API needs to be
-                    // reviewed periodically.
-#pragma warning disable 0618
-                    loadedAssembly = Assembly.LoadWithPartialName(fixedName);
-
-                    if (loadedAssembly != null)
-                    {
-                        // In the past, LoadWithPartialName would just return null in most cases when the assembly could not be found or loaded.
-                        // In addition to this, the error was always cleared. So now, clear the error variable only if the assembly was loaded.
-                        error = null;
-                    }
-                    return loadedAssembly;
-                }
-
-                // Expected exceptions are ArgumentNullException and BadImageFormatException. See https://msdn.microsoft.com/en-us/library/12xc5368(v=vs.110).aspx
-                catch (BadImageFormatException badImage)
-                {
-                    error = badImage;
-                }
-            }
-#endif
-            return null;
+            // We either return the loaded Assembly, or return null.
+            return loadedAssembly;
         }
 
         /// <summary>
--- a/test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
@@ -88,3 +88,31 @@ Describe "Import-Module for Binary Modules in GAC" -Tags 'CI' {
         }
     }
 }
+
+Describe "Import-Module for Binary Modules" -Tags 'CI' {
+
+    It "PS should try to load the assembly from file path first" {
+ $src = @"
+using System.Management.Automation;           // Windows PowerShell namespace.
+
+namespace ModuleCmdlets
+{
+  [Cmdlet(VerbsDiagnostic.Test,"BinaryModuleCmdlet1")]   
+  public class TestBinaryModuleCmdlet1Command : Cmdlet
+  {
+    protected override void BeginProcessing()
+    {
+      WriteObject("BinaryModuleCmdlet1 exported by the ModuleCmdlets module.");
+    }
+  }
+}
+"@
+
+    Add-Type -TypeDefinition $src -OutputAssembly $TESTDRIVE\System.dll
+    $assembly = Import-Module $TESTDRIVE\System.dll -Force -PassThru
+
+    #Ignore slash format difference under windows/Unix
+    $path = (Get-ChildItem $TESTDRIVE\System.dll).FullName
+    $assembly.Path | Should be $path
+    }
+ }
--- a/test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
@@ -109,12 +109,12 @@ namespace ModuleCmdlets
 "@
 
     Add-Type -TypeDefinition $src -OutputAssembly $TESTDRIVE\System.dll
-    $assembly = Import-Module $TESTDRIVE\System.dll -Force -PassThru
+    $results = powershell -noprofile -c "`$module = Import-Module $TESTDRIVE\System.dll -Passthru; `$module.ImplementingAssembly.Location; Test-BinaryModuleCmdlet1"
 
     #Ignore slash format difference under windows/Unix
     $path = (Get-ChildItem $TESTDRIVE\System.dll).FullName
-    $assembly.Path | Should be $path
-    Test-BinaryModuleCmdlet1 | Should Be "BinaryModuleCmdlet1 exported by the ModuleCmdlets module."
+    $results[0] | Should Be $path
+    $results[1] | Should BeExactly "BinaryModuleCmdlet1 exported by the ModuleCmdlets module."
     }
  }
 
--- a/test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
@@ -114,5 +114,7 @@ namespace ModuleCmdlets
     #Ignore slash format difference under windows/Unix
     $path = (Get-ChildItem $TESTDRIVE\System.dll).FullName
     $assembly.Path | Should be $path
+    Test-BinaryModuleCmdlet1 | Should Be "BinaryModuleCmdlet1 exported by the ModuleCmdlets module."
     }
  }
+


SteveL-MSFT:
Can you add a synthetic test case?

chunqingchen:
@SteveL-MSFT test is added

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1323,82 +1323,81 @@ internal void RemoveAssembly(string name)
...
+                error = null;
+
+                try
+                {
+                    loadedAssembly = Assembly.LoadFrom(filename);
+                    return loadedAssembly;
+                }
+                catch (FileNotFoundException fileNotFound)
+                {
+                    error = fileNotFound;
```
SteveL-MSFT: if we were given a filepath and it doesn't exist, it doesn't seem like we should try to load something that happens to have the same name but different path

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1323,82 +1323,81 @@ internal void RemoveAssembly(string name)
...
+                }
+                catch (FileLoadException fileLoadException)
+                {
+                    error = fileLoadException;
```
SteveL-MSFT: if we were given a filepath and it doesn't exist, it doesn't seem like we should try to load something that happens to have the same name but different path

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1323,82 +1323,81 @@ internal void RemoveAssembly(string name)
...
-                    return null;
-                }
-                catch (SecurityException securityException)
-                {
-                    error = securityException;
-                    return null;
-                }
-            }
-
 #if !CORECLR// Assembly.LoadWithPartialName is not in CoreCLR. In CoreCLR, 'LoadWithPartialName' can be replaced by Assembly.Load with the help of AssemblyLoadContext.
```
SteveL-MSFT: We are removing FullCLR code, so you should remove this since you're touching this file

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1323,82 +1323,81 @@ internal void RemoveAssembly(string name)
...
+                    return null;
+                }
+                catch (SecurityException securityException)
+                {
+                    error = securityException;
+                    return null;
+                }
+            }
+
+            if (loadedAssembly != null)
```
SteveL-MSFT: be consistent with line 1338 and `return loadedAssembly` on line 1375

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1323,82 +1323,81 @@ internal void RemoveAssembly(string name)
...
+                    return null;
+                }
+                catch (SecurityException securityException)
+                {
+                    error = securityException;
+                    return null;
+                }
+            }
+
+            string fixedName = null;
```
SteveL-MSFT: since the FullCLR code should be removed `fixedName` is only used within this `if` statement so move it inside the `if` statement

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1323,82 +1323,81 @@ internal void RemoveAssembly(string name)
...
+                    error = fileNotFound;
+                }
+                catch (FileLoadException fileLoadException)
+                {
+                    error = fileLoadException;
+                }
+                catch (BadImageFormatException badImage)
+                {
+                    error = badImage;
+                    return null;
```
SteveL-MSFT: if you didn't successfully load the assembly, `loadedAssembly` should still be null, so I think you can get rid of the `return null;` statements and just `return loadedAssembly` at the end

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1323,82 +1323,81 @@ internal void RemoveAssembly(string name)
...
+                }
+                catch (SecurityException securityException)
+                {
+                    error = securityException;
+                    return null;
+                }
+            }
+
+            string fixedName = null;
+            if (!String.IsNullOrEmpty(name))
```
SteveL-MSFT: make this `else if`

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -88,3 +88,26 @@ Describe "Import-Module for Binary Modules in GAC" -Tags 'CI' {
...
+
+    It "PS should try to load the assembly from file path first" {
+ $src = @"
+using System;
+namespace ModuleCmdlets
+{ 
+  public class TestBinaryModuleCmdlet1Command
+  {
+  }
+
```
chunqingchen: resolved

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -88,3 +88,26 @@ Describe "Import-Module for Binary Modules in GAC" -Tags 'CI' {
...
+"@
+
+    Add-Type -TypeDefinition $src -OutputAssembly $TESTDRIVE\System.Management.Automation.dll
+    Import-Module $TESTDRIVE\System.Management.Automation.dll -Force
+
+    #Ignore slash format difference under windows/Unix
+    $path = (Get-ChildItem $TESTDRIVE\System.Management.Automation.dll).FullName
+    [System.AppDomain]::CurrentDomain.GetAssemblies().Location.Contains($path) | Should be $true
+    }
+ }
```
chunqingchen: resolved

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1323,82 +1323,81 @@ internal void RemoveAssembly(string name)
...
-
-            if (loadedAssembly != null)
+            // First we try to load the assembly based on the filename
+
+            Assembly loadedAssembly = null;
+            error = null;
+
+            if (!String.IsNullOrEmpty(filename))
+            {
+                error = null;
```
chunqingchen: resolved

src/System.Management.Automation/engine/ExecutionContext.cs
chunqingchen: resolved

src/System.Management.Automation/engine/ExecutionContext.cs
chunqingchen: resolved

src/System.Management.Automation/engine/ExecutionContext.cs
chunqingchen: resolved.

src/System.Management.Automation/engine/ExecutionContext.cs
chunqingchen: resolved

src/System.Management.Automation/engine/ExecutionContext.cs
chunqingchen: resolved

src/System.Management.Automation/engine/ExecutionContext.cs
chunqingchen: resolved

src/System.Management.Automation/engine/ExecutionContext.cs
chunqingchen: resolved

chunqingchen:
@SteveL-MSFT your comments are resolved. thank you

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1323,110 +1323,73 @@ internal void RemoveAssembly(string name)
...
+                catch (BadImageFormatException badImage)
+                {
+                    error = badImage;
+                }
+                catch (SecurityException securityException)
+                {
+                    error = securityException;
+                }
+            }
+
```
SteveL-MSFT: Delete blank line

SteveL-MSFT (changes requested):

src/System.Management.Automation/engine/ExecutionContext.cs
chunqingchen: resolved

chunqingchen:
@SteveL-MSFT your comment is resolved thanks

SteveL-MSFT (approved):
LGTM

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -88,3 +88,25 @@ Describe "Import-Module for Binary Modules in GAC" -Tags 'CI' {
...
+  }
+}
+"@
+
+    Add-Type -TypeDefinition $src -OutputAssembly $TESTDRIVE\System.Management.Automation.dll
+    Import-Module $TESTDRIVE\System.Management.Automation.dll -Force
+
+    #Ignore slash format difference under windows/Unix
+    $path = (Get-ChildItem $TESTDRIVE\System.Management.Automation.dll).FullName
+    [System.AppDomain]::CurrentDomain.GetAssemblies().Location.Contains($path) | Should be $true
```
daxian-dbw: This works only because the actual assembly name of the assembly generated from `Add-Type` is not the same as the file name.

daxian-dbw:
This change seems risky to me.
1. If an assembly is already loaded, and you try to load another same assembly file of a different version, `Assembly.LoadFrom` will throw; if you want to load another assembly file with the same version, then the already loaded assembly will be returned. See an example here:
```
PS:6> [System.Reflection.AssemblyName]::GetAssemblyName("F:\tmp\System.Management.Automation.dll")

Version        Name
-------        ----
6.0.1.0        System.Management.Automation

PS:7> [psobject].Assembly.GetName()

Version        Name
-------        ----
6.0.0.0        System.Management.Automation

PS:8> [System.Reflection.Assembly]::LoadFrom("F:\tmp\System.Management.Automation.dll")
Exception calling "LoadFrom" with "1" argument(s): "Could not load file or assembly 'System.Management.Automation,
Version=6.0.1.0, Culture=neutral, PublicKeyToken=null'."
At line:1 char:1
+ [System.Reflection.Assembly]::LoadFrom("F:\tmp\System.Management.Auto ...
    + FullyQualifiedErrorId : FileLoadException
```
2. The original code prefers to assemblies in GAC, while the altered code prefers to file name. This is a behavior change, and since we now looking for GAC assemblies in powershell core for `Assembly.Load` call, this change could cause problems.

Add @lzybkr to reviewers to get more inputs on this change.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -88,3 +88,25 @@ Describe "Import-Module for Binary Modules in GAC" -Tags 'CI' {
...
 }
+
+Describe "Import-Module for Binary Modules" -Tags 'CI' {
+
+    It "PS should try to load the assembly from file path first" {
+ $src = @"
+using System;
+namespace ModuleCmdlets
+{ 
+  public class TestBinaryModuleCmdlet1Command
```
lzybkr: You should make this a complete cmdlet that can be called. That is the real scenario that needs to be tested.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
lzybkr: Instead of this, you can use `Import-Module -Passthru` and have the exact `Assembly` instance to validate.

We also want to avoid `Should Be $true` because the logs are less useful, so I'd like to see something like:

```powershell
$assem = Import-Module -PassThru ...
$assem.Location | Should Be $Path
```

chunqingchen: resolved

chunqingchen:
@daxian-dbw 
According to dongbo's concern about the exception. If user has specified the file path of the assembly, we should load the assembly according to user's willing. If an exception is thrown as assembly is already loaded, user should correct the file path or just give assembly name instead.

Loading assembly in GAC regardless user's given file path is exactly where the bug scenario comes from.

@lzybkr  your comments are resolved.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -88,3 +88,31 @@ Describe "Import-Module for Binary Modules in GAC" -Tags 'CI' {
         }
     }
 }
+
+Describe "Import-Module for Binary Modules" -Tags 'CI' {
+
+    It "PS should try to load the assembly from file path first" {
```
daxian-dbw: Are you sure this test validates the fix? It passes even without your fix.

daxian-dbw (changes requested):

daxian-dbw:
@chunqingchen please don't override existing commit history. Instead, push new commits to address comments.
Also, please make sure you add `[Feature]` label to your next commit message so that all tests can be validated with your fix.

daxian-dbw:
@chunqingchen if we decide to take this breaking change, then it'd better go in as soon as possible so that we have more time to validate the potential impact. Can you please address the comments as soon as you get a chance?

chunqingchen:
@daxian-dbw thanks for your comments. your comment is resolved.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
chunqingchen: previously when system.automation.management is used loading error will be occured. 
Now we have to check the actual cmdlet is loaded.
Resolved.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -88,3 +88,33 @@ Describe "Import-Module for Binary Modules in GAC" -Tags 'CI' {
...
+    protected override void BeginProcessing()
+    {
+      WriteObject("BinaryModuleCmdlet1 exported by the ModuleCmdlets module.");
+    }
+  }
+}
+"@
+
+    Add-Type -TypeDefinition $src -OutputAssembly $TESTDRIVE\System.dll
+    $assembly = Import-Module $TESTDRIVE\System.dll -Force -PassThru
```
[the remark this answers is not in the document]
daxian-dbw: Plus, `Import-Module -PassThru` returns a `PSModuleInfo` object, so the value name should be `$module` instead of  `$assembly`.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -88,3 +88,33 @@ Describe "Import-Module for Binary Modules in GAC" -Tags 'CI' {
...
+
+    #Ignore slash format difference under windows/Unix
+    $path = (Get-ChildItem $TESTDRIVE\System.dll).FullName
+    $assembly.Path | Should be $path
```
daxian-dbw: I think your actual intent is to use `$module.ImplementingAssembly.Location`.

daxian-dbw (changes requested):

daxian-dbw:
@chunqingchen Thanks for the update, I left a few more comments. Please push new commits to address the commits without overriding the commit history.

daxian-dbw (changes requested):

daxian-dbw:
@chunqingchen Please fix the test failure.

chunqingchen:
thanks @daxian-dbw , checks are passed.

daxian-dbw (approved):

PR #4198, opened 2017-07-07
Create symlinks to needed native libraries in publish path for RedHat variants

JamesWTruher:
This will keep the shell from exiting when running new-pssession (and tests which call it)

After a build on RedHat, the publish directory needs links to some native libraries otherwise when some commands are run PowerShell will fail with a null reference exception. While this issue may exist on other platforms, this PR addresses the issue only on the RedHat variants. 

When we create packages we create these links on the RedHat variants, this change does the same thing in the publish directory.

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -486,6 +486,19 @@ cmd.exe /C cd /d "$location" "&" "$($vcPath)\vcvarsall.bat" "$NativeHostArch" "&
         Copy-Item -Path "$PSScriptRoot/src/powershell-win-core/Microsoft.PowerShell_profile.ps1" -Destination $publishPath -Force
     }
 
+    if ($IsRedHatFamily) {
+        # add two symbolic links to system shared libraries that libmi.so is dependent on to handle
+        # platform specific changes. This is the only set of platforms needed for this currently
+        # as Ubuntu has these specific library files in the platform and OSX builds for itself
+        # against the correct versions.
+        if ( ! (test-path "$publishPath/libssl.so.1.0.0")) {
+            $null = New-Item -Force -ItemType SymbolicLink -Target "/lib64/libssl.so.10" -Path "$publishPath/libssl.so.1.0.0" -ErrorAction Stop
+        }
+        if ( ! (test-path "$publishPath/libcrypto.so.1.0.0")) {
+            $null = New-Item -Force -ItemType SymbolicLink -Target "/lib64/libcrypto.so.10" -Path "$publishPath/libcrypto.so.1.0.0" -ErrorAction Stop
+        }
+    }
+
     # download modules from powershell gallery.
     #   - PowerShellGet, PackageManagement, Microsoft.PowerShell.Archive
     if($PSModuleRestore)


build.psm1
```diff
@@ -468,6 +468,16 @@ cmd.exe /C cd /d "$location" "&" "$($vcVarsPath)\vcvarsall.bat" "$NativeHostArch
         }
         New-Item -Path $refDestFolder -ItemType Directory -Force -ErrorAction Stop > $null
         Copy-Item -Path $refAssemblies -Destination $refDestFolder -Force -ErrorAction Stop
+
+        # create the symlinks needed on Linux
+        if ($IsRedHatFamily) {
+            if ( ! (test-path "$publishPath/libssl.so.1.0.0")) {
```
adityapatwardhan: Should we also check if the target exists?

build.psm1
```diff
@@ -468,6 +468,16 @@ cmd.exe /C cd /d "$location" "&" "$($vcVarsPath)\vcvarsall.bat" "$NativeHostArch
...
+                $null = New-Item -Force -ItemType SymbolicLink -Target "/lib64/libssl.so.10" -Path "$publishPath/libssl.so.1.0.0"
+            }
+            if ( ! (test-path "$publishPath/libcrypto.so.1.0.0")) {
+                $null = New-Item -Force -ItemType SymbolicLink -Target "/lib64/libcrypto.so.10" -Path "$publishPath/libcrypto.so.1.0.0"
```
adityapatwardhan: Same as above.

build.psm1
```diff
@@ -468,6 +468,16 @@ cmd.exe /C cd /d "$location" "&" "$($vcVarsPath)\vcvarsall.bat" "$NativeHostArch
         }
         New-Item -Path $refDestFolder -ItemType Directory -Force -ErrorAction Stop > $null
         Copy-Item -Path $refAssemblies -Destination $refDestFolder -Force -ErrorAction Stop
+
+        # create the symlinks needed on Linux
+        if ($IsRedHatFamily) {
+            if ( ! (test-path "$publishPath/libssl.so.1.0.0")) {
+                $null = New-Item -Force -ItemType SymbolicLink -Target "/lib64/libssl.so.10" -Path "$publishPath/libssl.so.1.0.0"
```
adityapatwardhan: -ErrorAction Stop for both New-Item calls. That is the pattern used in the function.

adityapatwardhan (dismissed):

TravisEz13:
@JamesWTruher Please make your Microsoft org membership public.

build.psm1
JamesWTruher: a symlink can be created if the target doesn't exist, we can check but it's not really needed

build.psm1
JamesWTruher: fixed

build.psm1
JamesWTruher: fixed

JamesWTruher:
@adityapatwardhan should be good now

build.psm1
```diff
@@ -468,6 +468,16 @@ cmd.exe /C cd /d "$location" "&" "$($vcVarsPath)\vcvarsall.bat" "$NativeHostArch
         }
         New-Item -Path $refDestFolder -ItemType Directory -Force -ErrorAction Stop > $null
         Copy-Item -Path $refAssemblies -Destination $refDestFolder -Force -ErrorAction Stop
+
+        # create the symlinks needed on Linux
```
daxian-dbw: Could you please use the same comment from `New-UnixPackage` so that it's more clear that this is required by Linux remoting over OMI:
```
        # add two symbolic links to system shared libraries that libmi.so is dependent on to handle
        # platform specific changes. This is the only set of platforms needed for this currently
        # as Ubuntu has these specific library files in the platform and OSX builds for itself
        # against the correct versions.
```

JamesWTruher:
@adityapatwardhan should be ok now
@daxian-dbw since we're using `-force` in package creation it should be fine

build.psm1
```diff
@@ -476,6 +476,19 @@ cmd.exe /C cd /d "$location" "&" "$($vcPath)\vcvarsall.bat" "$NativeHostArch" "&
...
+            # platform specific changes. This is the only set of platforms needed for this currently
+            # as Ubuntu has these specific library files in the platform and OSX builds for itself
+            # against the correct versions.
+            if ( ! (test-path "$publishPath/libssl.so.1.0.0")) {
+                $null = New-Item -Force -ItemType SymbolicLink -Target "/lib64/libssl.so.10" -Path "$publishPath/libssl.so.1.0.0" -ErrorAction Stop
+            }
+            if ( ! (test-path "$publishPath/libcrypto.so.1.0.0")) {
+                $null = New-Item -Force -ItemType SymbolicLink -Target "/lib64/libcrypto.so.10" -Path "$publishPath/libcrypto.so.1.0.0" -ErrorAction Stop
+            }
+        }
```
daxian-dbw: I think this block should be moved out of this `try/finally` block, and to before or after the `if ($IsWindows) { # copy PowerShell host profile if Windows }` right below. This `try/finally` block is only for the purpose of `publish netcoreapp2.0 ref assemblies`.

JamesWTruher: does it really matter?

daxian-dbw (approved):
LGTM

build.psm1
daxian-dbw: Thanks! It's my personal opinion that it's probably cleaner to separate each after-build action in its own block.

Issue #4145, opened 2017-06-30
Behavior changed about `filesystemInfo.Attributes` in .NET Core preview2/3 causes 2 tests to fail in `Get-ChildItem.Tests.ps1`


daxian-dbw:
In `2.0.0-preview1` .NET Core, `filesystemInfo.Attributes` raises an `FileNotFoundException` exception when the file is deleted or renamed during the enumration.
In `2.0.0-preview2/3` .NET Core, `filesystemInfo.Attributes` returns -1 without raising any exception.
This behavior change was by design -- to align with the current behavior on Full .NET. For detailed information, please see https://github.com/dotnet/corefx/issues/20456.

This behavior change causes the following tests to fail in test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1 (they will be marked as pending for now):
- "Should continue enumerating a directory when a contained item is deleted"
- "Should continue enumerating a directory when a contained item is renamed"

These 2 tests were added in PR #3806, which assumes that the `FileNotFoundException` will be thrown in case the file is deleted or renamed during enumeration. Since the behavior has changed in .NET Core, the fix needs to be revisited.

/cc @jeffbi

daxian-dbw:
Since PR #3806 was to fix #2856, I set the properties of this issue similar to 2856.

jeffbi:
What I'm seeing is that the exception is no longer thrown, but `FilesystemInfo.Attributes` is _not_ returning -1. Instead it returns the attributes the file had before it was deleted, so there is no real indication that the file was deleted.

After the item is written to the pipeline, when it is displayed, the `Mode` property is null. `Mode` is not one of the regular properties of `FileSystemInfo` and is added after the provider writes the object.

I think it might be worth keeping the provider code as is (I don't think the `try/catch` is doing us any harm) and change the test so that it ensures that the deleted file is included in the result of `Get-ChildItem`.

iSazonov:
It seems we should refresh to get -1 https://github.com/dotnet/corefx/blob/801dde95a5eac06140d0ac633ac3f9bfdd25aca5/src/System.IO.FileSystem/src/System/IO/FileSystemInfo.Unix.cs#L264

jeffbi:
I had fetched the latest source (commit ece27ff68798f508c755ccb84be5ae78b386ac82), done a new `Start-PSBootStrap` and `Start-PSBuild -clean`. Is there something else I need to do?

[...]

PR #4200, opened 2017-07-07
Remove test hooks and tests for files being deleted or renamed during enumeration

jeffbi:
Addresses issue #4145

Previously with .NET Core, `FileNotFoundException` would be thrown if a file was deleted during an enumeration, and that issue was fixed by #3806. That behavior is changed in newer .NET Core to align with full .NET. So the tests hooks and tests added in #3806 is unneeded now and thus removed.


daxian-dbw (approved):
LGTM.

PR #4203, opened 2017-07-09
Checkbox to open PowerShell after the (Windows) installer has finished

bergmeister:
Added a checkbox (unchecked by default) to the last dialogue of the Windows installer to provide the option of opening PowerShell because that's what most people want to do if they installed PowerShell.

### DIFF
--- a/assets/Product.wxs
+++ b/assets/Product.wxs
@@ -33,7 +33,7 @@
     <Property Id="ARPHELPLINK" Value="$(var.InfoURL)" />
 
 	<!-- Checkbox to allow starting PowerShell after the installation (in UI mode only) -->
-	<Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="Open PowerShell" />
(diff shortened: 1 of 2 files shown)


assets/Product.wxs
```diff
@@ -31,6 +31,13 @@
 
     <!-- Set properties for add/remove programs -->
     <Property Id="ARPHELPLINK" Value="$(var.InfoURL)" />
+
+	<!-- Checkbox to allow starting PowerShell after the installation (in UI mode only) -->
+	<Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="Open PowerShell" />
```
SteveL-MSFT: Replace `PowerShell` with `$(env.ProductName)`

SteveL-MSFT (changes requested):

assets/Product.wxs
bergmeister: Good idea. Done and manually tested on the MSI artifact of CI to verify the CheckBox text is still 'Open PowerShell' after using a variable for it.

SteveL-MSFT (approved):
Thanks for the contribution!  LGTM

PR #4209, opened 2017-07-10
Run PowerShell class static methods in the correct Runspace/SessionState

### LINKED ISSUE #4001: PowerShell class static method has inconsistent behavior when the .ps1 file is dot sourced into multiple Runspace's
Root cause
------------
Here is the background information:
1. When running the same unchanged script file multiple times, the script file will not be re-parsed. Instead, the AST will be used.
1. When re-using the AST, we will reuse the same .NET type generated for a powershell class.
1. At compilation time, [`ScriptBlockMemberMethodWrapper.InitAtRuntime`](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/runtime/Operations/ClassOps.cs#L115) will be called to bound the static method of a powershell class to the [Runspace, SessionState] where the compilation is happening.

This causes a problem -- a static method of powershell class will always execute in the [Runspace, SessionState] where the script file that defines it lastly run.

Steps to reproduce
------------------
> class.ps1 -- defines a powershell class `Foo`

```powershell

class Foo
{
    static [string] GetName()
    {
        return (Get-Name)
    }
}
```
> test.ps1
Dot source `class.ps1`, create function `Get-Name` and run `[Foo]::GetName()`.
Then create a new runspace, define another `Get-Name` and dot source `class.ps1` in that runspace.
Then run `[Foo]::GetName()` again.

```powershell

. $PSScriptRoot\class.ps1

function Get-Name
{
    "Console Runspace"
}

[Foo]::GetName()

$ps = [powershell]::Create()
$ps.AddScript("function Get-Name { 'powershell instance Runspace' }").Invoke()
$ps.Commands.Clear()
$ps.AddScript(". $PSScriptRoot\class.ps1").Invoke()

[Foo]::GetName()
```
> Then just run `test.ps1` in powershell

```powershell
.\test.ps1
```

Expected behavior
-----------------
> Since I'm running both `[Foo]::GetName()` in the default console runspace, I expect the same result.

```none
Console Runspace
Console Runspace
```

Actual behavior
---------------

```none
Console Runspace
powershell instance Runspace
```

Environment data
----------------

```powershell
> $PSVersionTable
CLRVersion
GitCommitId                    v6.0.0-beta.2-22-g7a51b446e5cc428efd634e8a8a38271156cc182b
```

daxian-dbw:
Fix #4001 

Summary
----------
When the same script file gets executed in multiple Runspaces, the RuntimeType generated from the powershell class defined in the file will be shared among those Runspaces. For different Runspaces, different SessionState should be used to run the static methods.

Fix
---
We use the `SessionStateKeeper` to query for the correct `SessionState` to use for a static method call.
We already use `SessionStateKeeper` to track the `Runspace/SessionState` where a PowerShell class is defined, and thus we can leverage this for the static method invocation.

### DIFF
--- a/src/System.Management.Automation/engine/runtime/Operations/ClassOps.cs
+++ b/src/System.Management.Automation/engine/runtime/Operations/ClassOps.cs
@@ -206,7 +206,9 @@ namespace System.Management.Automation.Internal
                 // For static method, it's a little complex.
                 // - Check if the current default runspace is registered with the SessionStateKeeper. If so, use the registered SessionState.
                 // - Otherwise, check if default SessionState is still alive. If so, use the default SessionState.
-                // - Otherwise, the 'SessionStateInternal' property will be set to null, and the current default runspace will be used.
+                // - Otherwise, the 'SessionStateInternal' property will be set to null, and thus the default runspace of the current thread will be used.
+                //              If the current thread doesn't have a default Runspace, then an InvalidOperationException will be thrown when invoking the
+                //              script block, which is expected.
                 sessionStateToUse = (SessionStateInternal)_sessionStateKeeper.GetSessionState();
                 if (sessionStateToUse == null)
                 {
--- a/src/System.Management.Automation/engine/runtime/Operations/ClassOps.cs
+++ b/src/System.Management.Automation/engine/runtime/Operations/ClassOps.cs
@@ -32,8 +32,8 @@ namespace System.Management.Automation.Internal
     /// </summary>
     public class SessionStateKeeper
     {
-        // We use ConditionalWeakTable, because if GC already collect Runspace,
-        // then there is no way to call a ctor on the type in this Runspace.
+        // We use ConditionalWeakTable, because if GC already collect Runspace, then there
+        // is no way to call a ctor or a static method on the type in this Runspace.
         private readonly ConditionalWeakTable<Runspace, SessionStateInternal> _stateMap;
 
         internal SessionStateKeeper()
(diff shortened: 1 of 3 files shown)


src/System.Management.Automation/engine/runtime/Operations/ClassOps.cs
```diff
@@ -91,31 +101,122 @@ public class ScriptBlockMemberMethodWrapper
...
+        /// The default SessionState is used only if a static method is called from a Runspace where the class is never
+        /// defined, or is called on a thread without a default Runspace.
+        /// </remakr>
         internal void InitAtRuntime()
         {
-            var context = Runspace.DefaultRunspace.ExecutionContext;
-            _scriptBlock.Value.SessionStateInternal = context.EngineSessionState;
+            if (_isStatic)
+            {
+                // WeakReference<T>'s instance methods are not thread-safe, so we need the lock to gurantee
```
PaulHigin: spelling of "guarantee"

src/System.Management.Automation/engine/runtime/Operations/ClassOps.cs
```diff
@@ -91,31 +101,122 @@ public class ScriptBlockMemberMethodWrapper
...
+                    var context = Runspace.DefaultRunspace.ExecutionContext;
+                    _defaultSessionStateToUse.SetTarget(context.EngineSessionState);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Set the SessionState of the script block appropriately.
+        /// </summary>
+        private void PrepareScriptBlockToInvoke(object instance, object sessionStateInternal, object[] args)
```
PaulHigin: It doesn't look like "args" parameter is used here.

src/System.Management.Automation/engine/runtime/Operations/ClassOps.cs
```diff
@@ -91,31 +101,122 @@ public class ScriptBlockMemberMethodWrapper
...
+            {
+                // Use the SessionState passed in, which is the one associated with the instance.
+                sessionStateToUse = (SessionStateInternal)sessionStateInternal;
+            }
+            else
+            {
+                // For static method, it's a little complex.
+                // - Check if the current default runspace is registered with the SessionStateKeeper. If so, use the registered SessionState.
+                // - Otherwise, check if default SessionState is still alive. If so, use the default SessionState.
+                // - Otherwise, the 'SessionStateInternal' property will be set to null, and the current default runspace will be used.
```
PaulHigin: I don't understand this comment.  The _defaultSessionStateToUse is from the current runspace.  So if it is null it means the current default runspace is null.  I think this means the sessionstate of the current runspace in which this is being run ... not the current default runspace.

PaulHigin (changes requested):

src/System.Management.Automation/engine/runtime/Operations/ClassOps.cs
daxian-dbw: `_defaultSessionStateToUse` is not necessary from the current Runspace. It's from the Runspace/SessionState where the PowerShell class was lastly defined. The `RuntimeType` generated from the PowerShell class could be passed to a different Runspace or a thread that doesn't have a default Runspace, and then static methods get called. This should be a cornor case, and `_defaultSessionStateToUse` is only used in that situation. Take a look at the comment of `_defaultSessionStateToUse`, it explains when this field is used.

src/System.Management.Automation/engine/runtime/Operations/ClassOps.cs
daxian-dbw: Good catch. Fixed.

src/System.Management.Automation/engine/runtime/Operations/ClassOps.cs
daxian-dbw: Fixed.

src/System.Management.Automation/engine/runtime/Operations/ClassOps.cs
PaulHigin: I understand that, but I feel the comment is wrong.  I believe it should say: "Otherwise, the 'SessionStateInternal' property will be set to null and the current runspace will be used."

Not "the current default runspace will be used" since there is no default runspace...

daxian-dbw: I see the confusion point. I will change it to:
`- Otherwise, the 'SessionStateInternal' property will be set to null, and thus the default runspace of the current thread will be used.`

PaulHigin (approved):
LGTM

Issue #3232, opened 2017-02-28
Support sudo <PowerShell cmdlet>


joeyaiello:
Steps to reproduce
------------------
`sudo Remove-Item foo.txt` where foo.txt is owned by root

Expected behavior
-----------------
`foo.txt` is deleted

Actual behavior
---------------
`sudo: Remove-Item: command not found`

MaximoTrinidad:
Hum! Very interesting.  But, I think the actual behavior is as expected.  As in Windows we don't have a command like sudo.  
But, you could use 'sudo powershell' and then do the remove foo.txt (which was created using 'sudo powershell') then it works when using Remove-Item.  Of course, it won't work in a non-elevated permission. 

I would only use 'sudo' on Linux commands ("sudo nautilus") from within PowerShell.

:)

SteveL-MSFT:
Even on Windows, it would be good to have similar capability to run a cmdlet as another user or elevated and something that could relied on in portable scripts.  May need an RFC for this.

pcgeek86:
This functionality would be useful to work around #3506.

pixelrebirth:
This is going to become a blocker for adapting Linux for me if this is not resolved. We are not allowed to sudo bash and similarly will not be able to do sudo powershell.

I agree with Maximo, we should leave sudo to bash and make a new cmdlet for doing elevation in powershell using sudo as scaffolding on expected functionality.

SteveL-MSFT:
@dantraMSFT has been investigating this, Dan can you provide an update on some current thinking for this?

dantraMSFT:
@pixelrebirth Can you provide more detail as to what you expect?  Are you expecting simple execution with a wait, output redirection or pipeline redirection?

pixelrebirth:
@dantraMSFT For sake of the argument I am going to call the new sudo like cmdlet: Invoke-Elevated
This is how sudo works and would translate to Invoke-Elevated in a powershell way

```
$cred = Get-Credential
Invoke-Elevated -credential $cred -command {
        Get-ChildItem ./test | Remove-Item -force
} | Get-SomeCmdlet
```

In this example, everything in the scriptblock would be executed elevated and then piped to a NON elevated Get-SomeCmdlet 
Cred can be a different user or yourself as an elevated user (permissions needs to be kept somewhere like sudoers file does aka a whitelist)

Logging should be enabled, by default for instances where Invoke-Elevated is executed:
User | Command | DateTime or similar format logging

[...]

PR #4210, opened 2017-07-10
Add sudo built-in function to UNIX builds.

dantraMSFT:
Fixes https://github.com/PowerShell/PowerShell/issues/3232

Associated function help: https://github.com/PowerShell/PowerShell-Docs/pull/1437

This function wraps calls to the native sudo command to detect commands/expressions requiring PowerShell. In those cases, powershell is invoked with the expression.

### DIFF
--- a/src/System.Management.Automation/engine/InitialSessionState.cs
+++ b/src/System.Management.Automation/engine/InitialSessionState.cs
@@ -4827,19 +4827,8 @@ end
         {
             return @"
 Set-StrictMode -Version Latest
-$script:allargs = [System.Collections.Generic.List[string]]::new()
-[string] $arg = $null
-foreach ($arg in $args)
-{
-    $script:allargs.Add($arg)
-}
 
 $script:AmbiguousOperators = @('>', '<', '|', '&')
-$script:WhiteSpaceTokens = @(
-    [System.Management.Automation.PSTokenType]::NewLine,
-    [System.Management.Automation.PSTokenType]::Comment,
-    [System.Management.Automation.PSTokenType]::LineContinuation
-)
 $script:IgnoredTokens = @(
     [System.Management.Automation.PSTokenType]::CommandParameter,
     [System.Management.Automation.PSTokenType]::CommandArgument,
@@ -4891,6 +4880,7 @@ function Get-ScriptBlock
 
     [string] $sudoArgs = Get-SudoArgs -AllArgs $AllArgs
 
+    <# tokenize the remaining arguments to identify the type of expression; native or powershell #>
     $tokens = [System.Management.Automation.PSParser]::Tokenize($AllArgs.ToArray(), [ref] $null)
     foreach ($token in $Tokens)
     {
@@ -4945,7 +4935,10 @@ function Get-ScriptBlock
     {
         foreach ($arg in $AllArgs)
         {
-            $null = $sb.Append(' ')
+            if ($sb.Length -gt 0)
+            {
+                $null = $sb.Append(' ')
+            }
             $null = $sb.Append($arg)
         }
         $commandArgs = $sb.ToString()
@@ -4961,6 +4954,7 @@ function Get-ScriptBlock
 
     if (-not [string]::IsNullOrEmpty($commandArgs))
     {
+        $null = $sb.Append(' ')
         if (-not $usePowerShell)
         {
             $null = $sb.Append($commandArgs)
@@ -4969,19 +4963,37 @@ function Get-ScriptBlock
         {
             $bytes = [System.Text.Encoding]::Unicode.GetBytes($commandArgs)
             $arguments = [Convert]::ToBase64String($bytes)
-            $null = $sb.AppendFormat(' powershell -encodedCommand {0}', $arguments)
+            $null = $sb.AppendFormat([System.Globalization.CultureInfo]::InvariantCulture, 'powershell -encodedCommand {0}', $arguments)
         }
     }
     return  [ScriptBlock]::Create($sb.ToString())
 }
 
-$sudoCmd = Get-Command -Name 'sudo' -CommandType Application -ErrorAction Ignore
-if ($sudoCmd -eq $null)
+[string] $sudoCmd = $null
+$sudoCmds = Get-Command -Name 'sudo' -CommandType Application -ErrorAction Ignore
+if ($sudoCmds -eq $null)
 {
     throw 'Could not resolve the sudo command'
 }
 
-[ScriptBlock] $scriptBlock = Get-ScriptBlock -AllArgs $allargs -SudoCmd $sudoCmd.Source
+if ($sudoCmds.GetType().IsArray)
+{
+    $sudoCmd = $sudoCmds[0].Path
+}
+else
+{
+    $sudoCmd = $sudoCmds.Path
+}
+
+<# place all args into an updateable list #>
+$allargs = [System.Collections.Generic.List[string]]::new()
+[string] $arg = $null
+foreach ($arg in $args)
+{
+    $allargs.Add($arg)
+}
+
+[ScriptBlock] $scriptBlock = Get-ScriptBlock -AllArgs $allargs -SudoCmd $sudoCmd
 & $scriptBlock
 ";
         }
--- a/src/System.Management.Automation/engine/InitialSessionState.cs
+++ b/src/System.Management.Automation/engine/InitialSessionState.cs
@@ -4881,8 +4881,7 @@ function Get-ScriptBlock
     [string] $sudoArgs = Get-SudoArgs -AllArgs $AllArgs
 
     <# tokenize the remaining arguments to identify the type of expression; native or powershell
-       if errors are encountered; assuming a native command and pass the expression through to sudo
-    #>
+       if errors are encountered; assuming a native command and pass the expression through to sudo #>
     $tokens = [System.Management.Automation.PSParser]::Tokenize($AllArgs.ToArray(), [ref] $null)
     foreach ($token in $Tokens)
     {
--- a/src/System.Management.Automation/engine/InitialSessionState.cs
+++ b/src/System.Management.Automation/engine/InitialSessionState.cs
@@ -4880,7 +4880,9 @@ function Get-ScriptBlock
 
     [string] $sudoArgs = Get-SudoArgs -AllArgs $AllArgs
 
-    <# tokenize the remaining arguments to identify the type of expression; native or powershell #>
+    <# tokenize the remaining arguments to identify the type of expression; native or powershell
+       if errors are encountered; assuming a native command and pass the expression through to sudo
+    #>
     $tokens = [System.Management.Automation.PSParser]::Tokenize($AllArgs.ToArray(), [ref] $null)
     foreach ($token in $Tokens)
     {
--- a/src/System.Management.Automation/engine/InitialSessionState.cs
+++ b/src/System.Management.Automation/engine/InitialSessionState.cs
@@ -4822,6 +4822,171 @@ end
 ";
         }
 
+#if UNIX
+        internal static string GetSudoFunctionText()
+        {
+            return @"
+Set-StrictMode -Version Latest
+$script:allargs = [System.Collections.Generic.List[string]]::new()
+[string] $arg = $null
+foreach ($arg in $args)
+{
+    $script:allargs.Add($arg)
+}
+
+$script:AmbiguousOperators = @('>', '<', '|', '&')
+$script:WhiteSpaceTokens = @(
+    [System.Management.Automation.PSTokenType]::NewLine,
+    [System.Management.Automation.PSTokenType]::Comment,
+    [System.Management.Automation.PSTokenType]::LineContinuation
+)
+$script:IgnoredTokens = @(
+    [System.Management.Automation.PSTokenType]::CommandParameter,
+    [System.Management.Automation.PSTokenType]::CommandArgument,
+    [System.Management.Automation.PSTokenType]::NewLine,
+    [System.Management.Automation.PSTokenType]::Comment,
+    [System.Management.Automation.PSTokenType]::StatementSeparator,
+    [System.Management.Automation.PSTokenType]::LineContinuation
+    )
+
+function Get-SudoArgs
+{
+    param
+    (
+        [System.Collections.Generic.List[string]] $AllArgs
+    )
+    $sb = [System.Text.StringBuilder]::new()
+
+    [string] $arg = $AllArgs[0]
+    while ($arg.StartsWith('-'))
+    {
+        $AllArgs.RemoveAt(0)
+        $null = $sb.Append(' ')
+        $null = $sb.Append($arg)
+
+        if ($AllArgs.Count -gt 0)
+        {
+            $arg = $AllArgs[0]
+        }
+        else
+        {
+            break
+        }
+    }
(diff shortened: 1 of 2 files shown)


test/powershell/engine/DefaultCommands.Tests.ps1
```diff
@@ -528,4 +528,225 @@ Describe "Verify approved aliases list" -Tags "CI" {
...
+            'Sudoers file grammar version',
+            'Sudoers I/O plugin version'
+            )
+        $result = (sudo -V)
+        $result.Count | Should Be $expectedValues.Count
+        for ($index = 0; $index -lt $expectedValues.Count; $index++)
+        {
+            $expected = $expectedValues[$index]
+            $actual = $result[$index]
+            $actual.StartsWith($expected) | Should Be $true
```
JamesWTruher: i would be happier with
`$actual | should match "^$expected"`
the error message will be easier to interpret

test/powershell/engine/DefaultCommands.Tests.ps1
```diff
@@ -528,4 +528,225 @@ Describe "Verify approved aliases list" -Tags "CI" {
...
+            'Sudoers file grammar version',
+            'Sudoers I/O plugin version'
+            )
+        $result = (sudo --version)
+        $result.Count | Should Be $expectedValues.Count
+        for ($index = 0; $index -lt $expectedValues.Count; $index++)
+        {
+            $expected = $expectedValues[$index]
+            $actual = $result[$index]
+            $actual.StartsWith($expected) | Should Be $true
```
JamesWTruher: same as above

test/powershell/engine/DefaultCommands.Tests.ps1
```diff
@@ -528,4 +528,225 @@ Describe "Verify approved aliases list" -Tags "CI" {
...
+    }
+
+    It "Validates sudo -V is not interpreted as -Verbose" -Skip:$skipTest {
+        $expectedValues = @(
+            'Sudo version',
+            'Sudoers policy plugin version',
+            'Sudoers file grammar version',
+            'Sudoers I/O plugin version'
+            )
+        $result = (sudo -V)
```
JamesWTruher: if you want to be absolutely certain that you're getting your function, you can use the construct
`& ${function:sudo} --V`

test/powershell/engine/DefaultCommands.Tests.ps1
```diff
@@ -528,4 +528,225 @@ Describe "Verify approved aliases list" -Tags "CI" {
...
+    }
+
+    It "Validates native sudo is invoked" -Skip:$skipTest {
+        $expectedValues = @(
+            'Sudo version',
+            'Sudoers policy plugin version',
+            'Sudoers file grammar version',
+            'Sudoers I/O plugin version'
+            )
+        $result = (sudo --version)
```
JamesWTruher: same as above

test/powershell/engine/DefaultCommands.Tests.ps1
```diff
@@ -528,4 +528,225 @@ Describe "Verify approved aliases list" -Tags "CI" {
...
+            return @{ExpectedArguments = $sb.ToString(); ActualArguments = $output}
+        }
+
+        It "Validates sudo parameters formatting using the mock" -Skip:$skipTest {
+            $result = InvokeSudo -SudoArgs '--verify'
+            $result.ActualArguments | Should Be $result.ExpectedArguments
+        }
+
+        It "Validates sudo parameters and expression formatting using the mock" -Skip:$skipTest {
+            $result = InvokeSudo -SudoArgs '--verify' -Expression 'ls /etc'
```
JamesWTruher: should this be `/bin/ls`?

test/powershell/engine/DefaultCommands.Tests.ps1
```diff
@@ -528,4 +528,225 @@ Describe "Verify approved aliases list" -Tags "CI" {
...
+            {
+                $cmd = $items[0]
+            }
+            else
+            {
+                $cmd = $items
+            }
+            if ($cmd.Path -ne $script:sudoMockPath)
+            {
+                throw "Get-Command did not resolve to the mock sudo. Expected: $($script:sudoMockPath)  Actual: $($cmd.Path)"
```
JamesWTruher: `$($script:sudoMockPath)` can be simply `${script:sudoMockPath}`

test/powershell/engine/DefaultCommands.Tests.ps1
```diff
@@ -528,4 +528,225 @@ Describe "Verify approved aliases list" -Tags "CI" {
...
+            $script:sudoMockPath = Join-Path -Path $script:sudoTestPath -ChildPath 'sudo'
+            $sudoscript = $script:sudoMock.Replace("`r`n", "`n")
+            Set-Content -Path $script:sudoMockPath -Value $sudoscript
+            chmod 777 $script:sudoMockPath
+
+            # Verify sudo resolves to the mock
+            $items = Get-Command -Name 'sudo' -CommandType Application
+            if ($items -eq $null)
+            {
+                throw "Could not resolve $($script:sudoMockPath) command"
```
JamesWTruher: same as line 587

test/powershell/engine/DefaultCommands.Tests.ps1
```diff
@@ -528,4 +528,225 @@ Describe "Verify approved aliases list" -Tags "CI" {
...
+            if ($skipTest) {return}
+
+            # restore the environment
+            if (-not [string]::IsNullOrEmpty($script:sudoSaveEnv))
+            {
+                $env:PATH = $script:sudoSaveEnv
+            }
+
+            # clean up the mock files
+            if (Test-Path -Path $script:sudoTestPath)
```
JamesWTruher: strictly speaking this shouldn't be needed. the contents of TESTDRIVE are cleaned up by the harness

test/powershell/engine/DefaultCommands.Tests.ps1
```diff
@@ -528,4 +528,225 @@ Describe "Verify approved aliases list" -Tags "CI" {
...
+        Test-Path Function:sudo | Should Be $CoreUnix
+    }
+}
+
+if ($isCoreCLR -and !$IsWindows)
+{
+[string] $script:sudoTestPath = $null
+[string] $script:sudoMockPath = $null
+[string] $script:sudoSaveEnv = $null
+[string] $script:sudoMock = @"
```
JamesWTruher: if you use single quotes, you don't need to escape the `$`
```
[string]$script:sudoMock = @'
#!/bin/sh
echo $@
'@
```

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -4822,6 +4822,183 @@ .ForwardHelpCategory Cmdlet
...
+[string] $sudoCmd = $null
+$sudoCmds = Get-Command -Name 'sudo' -CommandType Application -ErrorAction Ignore
+if ($sudoCmds -eq $null)
+{
+    throw 'Could not resolve the sudo command'
+}
+
+if ($sudoCmds.GetType().IsArray)
+{
+    $sudoCmd = $sudoCmds[0].Path
```
dantraMSFT: If you're referring to the array versus single item logic, when I have the mock sudo command in the path, I end up with two entries.  The array aids in testing and avoids having to remove entries from env:PATH.  I suppose there could be real-world cases but I don't have the details to know if it is reasonable.

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -4822,6 +4822,183 @@ .ForwardHelpCategory Cmdlet
...
+        [Parameter(Mandatory)]
+        [ValidateNotNullOrEmpty()]
+        [string] $SudoCmd
+    )
+    [bool] $usePowerShell = $false
+
+    [string] $sudoArgs = Get-SudoArgs -AllArgs $AllArgs
+
+    <# tokenize the remaining arguments to identify the type of expression; native or powershell #>
+    $tokens = [System.Management.Automation.PSParser]::Tokenize($AllArgs.ToArray(), [ref] $null)
```
dantraMSFT: If I can't tokenize it, I'm assuming it's a native command string that doesn't tokenize and I'll pass it through to sudo.  I'll add a comment to that effect unless you have alternative ideas.

test/powershell/engine/DefaultCommands.Tests.ps1
dantraMSFT: I'm not testing for the function; that's done elsewhere. I'm verifying the common parameters are not being imposed.  I'll add a comment to clarify it.

test/powershell/engine/DefaultCommands.Tests.ps1
dantraMSFT: I'll remove it.

test/powershell/engine/DefaultCommands.Tests.ps1
dantraMSFT: I would expect the results to be the same; regardless.  I'm not actually executing the command; I'm verifying that the function detects that 'ls' is a native command and does not attempt to invoke PowerShell for it.
I'll add a comment to clarify.

test/powershell/engine/DefaultCommands.Tests.ps1
dantraMSFT: Will fix.

test/powershell/engine/DefaultCommands.Tests.ps1
dantraMSFT: Will fix.

test/powershell/engine/DefaultCommands.Tests.ps1
dantraMSFT: Will fix

src/System.Management.Automation/engine/InitialSessionState.cs
dantraMSFT: fixed

test/powershell/engine/DefaultCommands.Tests.ps1
dantraMSFT: Fixed

test/powershell/engine/DefaultCommands.Tests.ps1
dantraMSFT: Fixed

test/powershell/engine/DefaultCommands.Tests.ps1
dantraMSFT: Fixed

test/powershell/engine/DefaultCommands.Tests.ps1
dantraMSFT: Fixed

test/powershell/engine/DefaultCommands.Tests.ps1
dantraMSFT: Fixed

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -4822,6 +4822,184 @@ .ForwardHelpCategory Cmdlet
 ";
         }
 
+#if UNIX
+        internal static string GetSudoFunctionText()
+        {
+            return @"
+Set-StrictMode -Version Latest
+
+$script:AmbiguousOperators = @('>', '<', '|', '&')
```
lzybkr: Because built-in functions are not in a script file, `script:` implies global scope - which we don't want here.

Also - you don't need `@()` when the value is already an array (you use the `,` operator).

src/System.Management.Automation/engine/InitialSessionState.cs
lzybkr: This is more complicated than necessary. You can force the result to be an array with:

```powershell
$sudoCmds = @(Get-Command ...)
```
Or you can force it to be a singleton with:
```powershell
$sudoCmds = Get-Command ... | Select -First 1
```

I think I prefer the later.

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -4822,6 +4822,184 @@ .ForwardHelpCategory Cmdlet
...
+
+        [Parameter(Mandatory)]
+        [ValidateNotNullOrEmpty()]
+        [string] $SudoCmd
+    )
+    [bool] $usePowerShell = $false
+
+    [string] $sudoArgs = Get-SudoArgs -AllArgs $AllArgs
+
+    <# tokenize the remaining arguments to identify the type of expression; native or powershell
```
lzybkr: This feels very wrong - the values passed to the command have no relation to the original command line.

I *think* the intent here is to guess if the command line was a PowerShell command or external command, but you get into trouble with stuff like:

```powershell
$x = "echo"
sudo $x + $x
```

You'll have 3 arguments - 'echo','+','echo' and decide to call the external command, but really I asked for this expression to be evaluated.

I think this is a huge problem with this approach - and I think this means we need parser support much in the same way that @BrucePay did `&`.

lzybkr (changes requested):
Other than my big concern about the way the tokenizer is being used here, I don't understand why this is a big PowerShell function.

If I'm wrong about needing parser support, I'd much prefer seeing the majority of the implementation in C#, or all of it if we have an alias to a proper cmdlet like `Start-Sudo`.

src/System.Management.Automation/engine/InitialSessionState.cs
dantraMSFT: I understand but that presents a different question/problem and I'm not conversant with the parser to know how it can resolve it.
If I tokenize the string $x + $x, $x get's resolved to echo.  If I enclose the same expression in a scriptblock, $x is parsed as a variable reference.  My immediate thought is to resolve these variable references and use the resulting expression.  However, this presents a different problem, what if the expression is an assignment, such as the following:
$env:PATH += ":/foo/bar"
or 
$env:PATH = $env:PATH + ':/foo/bar'
Would you expect the user to type something like the following:
'$env:PATH' = '$env:PATH' + ':/foo/bar'

lzybkr: Based on your current implementation as a function - you **should not** utilize the parser at all - a function is passed values, not PowerShell.

If the intent is to invoke PowerShell as root, then a reasonable syntax would be:

```powershell
sudo { $x + $x }
```

This would be unambiguous - you receive a script block, it's meant to run PowerShell, if you receive anything else, you would simply forward on to the native command `sudo`.

It might be worth some @PowerShell/powershell-committee discussion.

dantraMSFT: Right, I considered that approach but found it a little cumbersome to use.  I can impose it but it means it will be error prone for the natural use case:

sudo get-item /etc/bin/foo
versus
sudo {get-item  /etc/bin/foo}
I'll revisit that path again and see how it impacts extracting sudo-specific arguments.

SteveL-MSFT: @PowerShell/powershell-committee discussed this and have concerns about having a `sudo` implementation that doesn't completely work the same as `sudo` in bash (for example).  Recommendation is to not pursue implementing this for 6.0.0.  We should consider instead something like `Invoke-AsUser` that works for both native cmds and cmdlets/script blocks and @joeyaiello and I to use the RFC process.

PR #4205, opened 2017-07-09
Fix PSScriptAnalyzer warnings

bergmeister:
To help fixing #3791 
Fixes a few hundred warnings of type PSPossibleIncorrectComparisonWithNull, PSAvoidUsingCmdletAliases (mostly for ForEach-Object, Where-Object, Select-Object) and some PSUseDeclaredVarsMoreThanAssignment warnings.
As part of this, it was also found that a test in ParameterBindingTests needed improvement because since the null comparison is now correct, the test failed initially.

PR #4213, opened 2017-07-10
Fixes of around 350 PSScriptAnalyzer warnings

bergmeister:
This is the clean and rebased version with the split into 4 separate commits as requested by the original PR #4205
Most of them are fixes of PSPossibleIncorrectComparisonWithNull, PSAvoidUsingCmdletAliases (mostly for ForEach-Object, Where-Object, Select-Object) and some PSUseDeclaredVarsMoreThanAssignment warnings.
As part of this, it was also found that a test in ParameterBindingTests needed improvement because since the null comparison is now correct, the test failed initially and needed correcting.

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -47,8 +47,6 @@ if ($IsLinux) {
     $IsCentOS = $LinuxInfo.ID -match 'centos' -and $LinuxInfo.VERSION_ID -match '7'
     $IsFedora = $LinuxInfo.ID -match 'fedora' -and $LinuxInfo.VERSION_ID -ge 24
     $IsOpenSUSE = $LinuxInfo.ID -match 'opensuse'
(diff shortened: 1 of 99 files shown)


iSazonov:
@bergmeister Thanks for your conttribution! 
Wait maintainer's reviews.

iSazonov:
@bergmeister Sorry, it seems @lzybkr meant that we should put each change _kind_ in separate commit: %, ?, OutputType, $null and so on. I have no idea how this could be done now without a complete re-work. 

@lzybkr Can we review as is?

lzybkr:
If you do a soft reset, then `git add -p`, git will interactively ask you which diffs (not files, but individual diffs) to add. It's still time consuming, but it makes it much easier to manage big changes, e.g. if we need to use `git bisect` to narrow down a failure.

bergmeister:
OK. I have created #4261, which contains the separation of the commits as requested and will close this one.

Issue #4211, opened 2017-07-10
SetModulePath throws a NullReference Exception if getting Program files returns null


TravisEz13:
Steps to reproduce
------------------

```powershell
start powershell from inside powershell in the latest nanoserver docker
```

Expected behavior
-----------------
it should work

Actual behavior
---------------

```none
throws a null reference exception
```

Environment data
----------------

```powershell
> $PSVersionTable

```

PR #4219, opened 2017-07-11
make sure we don't return null for share module path

TravisEz13:
fall back to environment if call to get special path doesn't work


daxian-dbw (approved):
LGTM.

TravisEz13:
fixes #4211

PR #4227, opened 2017-07-11
Get Pester fixes into PowerShell

JamesWTruher:
There are two fixes in psl-pester which we should have:

- Add Context information to the pester log which will reduce the name collisions with test names. This will improve the unique-ness of test names, which will enable us to be able to compare test runs across platforms.
- Update test teardown so exceptions thrown in AfterAll will not leave pester in a bad state.
When this happens, Pester fails all subsequent tests with the error 'Already in Describe'.


adityapatwardhan (approved):

TravisEz13:
LGTM,  Travis-CI is still running.  Ping me when it passes.

Issue #4221, opened 2017-07-11
Add tests for type accelerators defined in powershell


daxian-dbw:
Currently, there is no test for the type accelerators. We should add tests to cover all type accelerators.

daxian-dbw:
@adityapatwardhan Could you please help to add this issue to the 'Test Coverage' project?

PR #4230, opened 2017-07-12
Added tests for builtin type accelerators

dchristian3188:
First pass of tests for built in type accelerators.

Referenced in issue: https://github.com/PowerShell/PowerShell/issues/4221

### DIFF
--- a/test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
+++ b/test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
@@ -1,428 +1,437 @@
 
 Describe "Type accelerators" -Tags "CI" {
     BeforeAll {
-        
         $TypeAcceleratorsType = [psobject].Assembly.GetType("System.Management.Automation.TypeAccelerators")
-        
         $TypeAccelerators = $TypeAcceleratorsType::Get
-        $TypeAcceleratorsType::Add('userDefinedAcceleratorType', [int])
-        $TypeAcceleratorsType::Add('userDefinedAcceleratorTypeToRemove', [int])
+    }
+    Context 'BuiltIn Accelerators' {
+        BeforeAll {
 
-        $TypeAcceleratorTestCases = @(
-            @{
-                Accelerator = 'Alias'
-                Type        = [System.Management.Automation.AliasAttribute]
-            }
-            @{
-                Accelerator = 'AllowEmptyCollection'
-                Type        = [System.Management.Automation.AllowEmptyCollectionAttribute]
-            }
-            @{
-                Accelerator = 'AllowEmptyString'
-                Type        = [System.Management.Automation.AllowEmptyStringAttribute]
-            }
-            @{
-                Accelerator = 'AllowNull'
-                Type        = [System.Management.Automation.AllowNullAttribute]
-            }
-            @{
-                Accelerator = 'ArgumentCompleter'
-                Type        = [System.Management.Automation.ArgumentCompleterAttribute]
-            }
-            @{
-                Accelerator = 'array'
-                Type        = [System.Array]
-            }
-            @{
-                Accelerator = 'bool'
-                Type        = [System.Boolean]
-            }
-            @{
-                Accelerator = 'byte'
-                Type        = [System.Byte]
-            }
-            @{
-                Accelerator = 'char'
-                Type        = [System.Char]
-            }
-            @{
-                Accelerator = 'CmdletBinding'
-                Type        = [System.Management.Automation.CmdletBindingAttribute]
-            }
-            @{
-                Accelerator = 'datetime'
-                Type        = [System.DateTime]
-            }
-            @{
-                Accelerator = 'decimal'
-                Type        = [System.Decimal]
-            }
-            @{
-                Accelerator = 'double'
-                Type        = [System.Double]
-            }
-            @{
-                Accelerator = 'DscResource'
-                Type        = [System.Management.Automation.DscResourceAttribute]
-            }
-            @{
-                Accelerator = 'float'
-                Type        = [System.Single]
-            }
-            @{
-                Accelerator = 'single'
-                Type        = [System.Single]
-            }
-            @{
-                Accelerator = 'guid'
-                Type        = [System.Guid]
-            }
-            @{
-                Accelerator = 'hashtable'
-                Type        = [System.Collections.Hashtable]
-            }
-            @{
-                Accelerator = 'int'
-                Type        = [System.Int32]
-            }
-            @{
-                Accelerator = 'int32'
-                Type        = [System.Int32]
-            }
-            @{
-                Accelerator = 'int16'
-                Type        = [System.Int16]
-            }
-            @{
-                Accelerator = 'long'
-                Type        = [System.Int64]
-            }
-            @{
-                Accelerator = 'int64'
-                Type        = [System.Int64]
-            }
-            @{
-                Accelerator = 'ciminstance'
-                Type        = [Microsoft.Management.Infrastructure.CimInstance]
-            }
-            @{
-                Accelerator = 'cimclass'
-                Type        = [Microsoft.Management.Infrastructure.CimClass]
-            }
-            @{
-                Accelerator = 'cimtype'
-                Type        = [Microsoft.Management.Infrastructure.CimType]
-            }           
-            @{
-                Accelerator = 'cimconverter'
-                Type        = [Microsoft.Management.Infrastructure.CimConverter]
-            }
-            @{
-                Accelerator = 'IPEndpoint'
-                Type        = [System.Net.IPEndPoint]
-            }
-            @{
-                Accelerator = 'NullString'
-                Type        = [System.Management.Automation.Language.NullString]
-            }
-            @{
-                Accelerator = 'OutputType'
-                Type        = [System.Management.Automation.OutputTypeAttribute]
-            }
-            @{
-                Accelerator = 'ObjectSecurity'
-                Type        = [System.Security.AccessControl.ObjectSecurity]
-            }
-            @{
-                Accelerator = 'Parameter'
-                Type        = [System.Management.Automation.ParameterAttribute]
-            }
-            @{
-                Accelerator = 'PhysicalAddress'
-                Type        = [System.Net.NetworkInformation.PhysicalAddress]
-            }
-            @{
-                Accelerator = 'pscredential'
-                Type        = [System.Management.Automation.PSCredential]
-            }
-            @{
-                Accelerator = 'PSDefaultValue'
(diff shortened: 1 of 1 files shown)


test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
```diff
@@ -1,17 +1,411 @@
 
 Describe "Type accelerators" -Tags "CI" {
-    $TypeAcceleratorsType = [psobject].Assembly.GetType("System.Management.Automation.TypeAccelerators")
+    BeforeAll {
+        
+        $TypeAcceleratorsType = [psobject].Assembly.GetType("System.Management.Automation.TypeAccelerators")
+        
+        $TypeAccelerators = $TypeAcceleratorsType::Get
+        $TypeAcceleratorsType::Add('msft_2174855', [int])
+        $TypeAcceleratorsType::Add('msft_2174855_rm', [int])
```
iSazonov: it would good to decode 'msft_2174855' and 'msft_2174855_rm' - can we use more informative names and comments for it?

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
```diff
@@ -1,17 +1,411 @@
...
 
     It "Can query type accelerators" {
-        if ( $IsCoreCLR ) { $count = 80 } else { $count = 82 }
+        if ( $IsCoreCLR )
+        {
+            $count = 80 
+        } else
+        {
+            $count = 82 
+        }
```
iSazonov: I believe it is better use the pattern and formatting https://github.com/PowerShell/PowerShell/blob/master/test/powershell/engine/DefaultCommands.Tests.ps1

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
```diff
@@ -22,4 +416,14 @@ Describe "Type accelerators" -Tags "CI" {
         $TypeAcceleratorsType::Remove('msft_2174855_rm')
         $TypeAcceleratorsType::Get['msft_2174855_rm'] | Should Be $null
     }
+
+    It 'Should have a type acclerator for: <Accelerator>' -TestCases $TypeAcceleratorTestCases {
+        param($Accelerator, $Type)
+        $TypeAcceleratorsType::Get[$Accelerator] | Should be ($Type)
```
iSazonov: Can we use `Should BeOfType` ?

dchristian3188: This did not work as expected. The object returned by the type accelerator get is System.RuntimeType.

```powershell
$TypeAcceleratorsType = [psobject].Assembly.GetType("System.Management.Automation.TypeAccelerators")
$TypeAcceleratorsType::Get['String']

$TypeAcceleratorsType::Get['String'] | Get-Member
```

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
iSazonov: #Closed.

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
iSazonov: #Closed

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
iSazonov: Please address the comment.

dchristian3188: Resolved in latest commit

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
```diff
@@ -1,25 +1,428 @@
...
 
     It "Basic type accelerator usage" {
-        [msft_2174855] | Should Be ([int])
+        [userDefinedAcceleratorType] | Should Be ([int])
     }
 
     It "Can query type accelerators" {
-        if ( $IsCoreCLR ) { $count = 80 } else { $count = 82 }
-        $TypeAccelerators.Count -gt $count | Should Be $true
+        $TypeAccelerators.Count -gt $totalAccelerators | Should Be $true
```
daxian-dbw: I think it would be good to test against the exact number of built-in accelerators (89). And it's recommended to write as `$TypeAccelerators.Count | Should Be $totalAccelerators` because it would give more information compared to `Should Be $true` in case the test fails.

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
```diff
@@ -1,25 +1,428 @@
...
+            }
+        )
+        
+        if ( $IsCoreCLR )
+        {
+            $totalAccelerators = 80 
+        } 
+        else
+        {
+            $totalAccelerators = 82
```
daxian-dbw: I think it's better to test the exact number of built-in accelerators.

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
```diff
@@ -1,25 +1,428 @@
...
-        [msft_2174855] | Should Be ([int])
+        [userDefinedAcceleratorType] | Should Be ([int])
     }
 
     It "Can query type accelerators" {
-        if ( $IsCoreCLR ) { $count = 80 } else { $count = 82 }
-        $TypeAccelerators.Count -gt $count | Should Be $true
+        $TypeAccelerators.Count -gt $totalAccelerators | Should Be $true
         $TypeAccelerators['xml'] | Should Be ([System.Xml.XmlDocument])
         $TypeAccelerators['AllowNull'] | Should Be ([System.Management.Automation.AllowNullAttribute])
```
daxian-dbw: These two tests can be removed now as we are testing the complete set of accelerators 😄

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
```diff
@@ -1,25 +1,428 @@
...
     }
 
     It "Can remove type accelerator" {
```
daxian-dbw: I recommend to group this test and `It "Basic type accelerator usage"` together in a `Context`, and then move the initialization of `userDefinedAcceleratorType` and `userDefinedAcceleratorTypeToRemove` to the `BeforeAll` of the `Context`. And remember to clean them up in `AfterAll` block of the same `Context`.

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
dchristian3188: Understood, will make the update tonight.

daxian-dbw (commented):
Thanks for adding tests for accelerators. Some comments below.

dchristian3188:
Changes made, let me know how it looks.

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
```diff
@@ -1,25 +1,437 @@
...
+                        Accelerator = 'wmisearcher'
+                        Type        = [System.Management.ManagementObjectSearcher]
+                    }
+                )
+            }
+        }
+
+        It 'Should have all the type accelerators' {
+            $TypeAccelerators.Count | Should be $totalAccelerators
+        }
```
iSazonov: Let's clarify. I still believe that we should follow the [pattern](https://github.com/PowerShell/PowerShell/blob/master/test/powershell/engine/DefaultCommands.Tests.ps1) to make the tests deterministic and well formatting.

It makes no sense to check the number of because that one accelerator can be replaced with other and we won't see the difference.

Accelerator may not be supported on all protforms. So we should check this by " $($FullCLR -or $CoreWindows -or $CoreUnix)".

Formatting should be tabular. It is more compact and we can read easy:
```powershell
$TypeAcceleratorTestCases = @"
    "Accelerator", "Type",                                        "Present"
    "wmisearcher", "System.Management.ManagementObjectSearcher",  $($FullCLR -or $CoreWindows -or $CoreUnix)
"@
```

daxian-dbw: > It makes no sense to check the number of because that one accelerator can be replaced with other and we won't see the difference.

There are tests to check each of the accelerators, so we will see a failure if one is replaced with another. This "count" test makes sure no new accelerator is added accidentally.

> Accelerator may not be supported on all protforms. So we should check this by " $($FullCLR -or $CoreWindows -or $CoreUnix)".

Accelerators are essentially .NET types, so I think there is no difference between `CoreWindows` and `CoreUnix` as they both use .NET Core.

> I still believe that we should follow the pattern to make the tests deterministic and well formatting.

I agree that the `DefaultCommands.Tests.ps1` is well-formatted and I recommend to use a similar pattern. But I don't think a PR should be blocked if it's clear enough (not confusing in any ways).

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
```diff
@@ -1,25 +1,437 @@
...
+                    Accelerator = 'psvariableproperty'
+                    Type        = [System.Management.Automation.PSVariableProperty]
+                }
+            )
+        
+            if ( $IsCoreCLR )
+            {
+                $totalAccelerators = 89 
+            } else
+            {
```
daxian-dbw: Should be
```
else
{
```

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
```diff
@@ -1,25 +1,437 @@
...
+                $totalAccelerators = 94
 
-    It "Basic type accelerator usage" {
-        [msft_2174855] | Should Be ([int])
-    }
+                $nonCoreTypeAcceleratorTestCases = @(
```
daxian-dbw: Can you please change the name to `$extraFullPSAcceleratorTestCases`.

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
iSazonov: Thanks for clarify!

>There are tests to check each of the accelerators, so we will see a failure if one is replaced with another. This "count" test makes sure no new accelerator is added accidentally.

If we have explicit full accelerator list in the file we catch the problem - in  DefaultCommands.Tests.ps1  we compare two arrays for this.

daxian-dbw: Yes, I see your point. In `DefaultCommands.Tests.ps1` we guarantee no unexpected `alias` or `cmdlet` by comparing the full list and that's why we don't need the count check in that test.

daxian-dbw (approved):
LGTM

PR #4248, opened 2017-07-13
sort and add new words to dictionary

TravisEz13:


TravisEz13:
linux and windows have passed... merging..

PR #4250, opened 2017-07-14
Move Bootstrap to install for Travis-CI in order to make the logs quieter

TravisEz13:
Moving bootstrap to install causes it to "fold" or hide, unless it fails.  
This makes the logs appear quieter.
move ulimit after verifying it is sticky
https://travis-ci.org/TravisEz13/PowerShell/jobs/253434166


daxian-dbw (approved):
LGTM.

PR #4117, opened 2017-06-27
Fix spurious error generated when importing cmdlets from an in-memory assembly

BrucePay:
Fixes #4025 where a test importing cmdlets from an in-memory assembly produced an spurious error.

PR #4253, opened 2017-07-14
Fix C# tests for ValidateSetValuesGenerator

iSazonov:
Daily build failed: https://ci.appveyor.com/project/PowerShell/powershell-f975h

Issue description
---
We dynamically build and load a module with test 'wrong' cmdlets to test C# implementation of a valid values generator.
Later Get-Help tests catches the test 'wrong' cmdlets and throws.

Fix
---
Unload the test module.

/cc @daxian-dbw

### DIFF
--- a/test/powershell/Language/Classes/Scripting.Classes.Attributes.Tests.ps1
+++ b/test/powershell/Language/Classes/Scripting.Classes.Attributes.Tests.ps1
@@ -294,6 +294,10 @@ Describe 'ValidateSet support a dynamically generated set' -Tag "CI" {
             $testModule = Import-Module $cls.Assembly -PassThru
         }
 
+        AfterAll {
+            Remove-Module -ModuleInfo $testModule
+        }
+
         It 'Throw if IValidateSetValuesGenerator is not implemented' {
             { Get-TestValidateSet0 -Param1 "TestString" -ErrorAction Stop } | ShouldBeErrorId "Argument"
         }
--- a/test/powershell/Language/Classes/Scripting.Classes.Attributes.Tests.ps1
+++ b/test/powershell/Language/Classes/Scripting.Classes.Attributes.Tests.ps1
@@ -297,7 +297,6 @@ Describe 'ValidateSet support a dynamically generated set' -Tag "CI" {
 
         AfterAll {
             Remove-Module -ModuleInfo $testModule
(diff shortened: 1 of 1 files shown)


test/powershell/Language/Classes/Scripting.Classes.Attributes.Tests.ps1
```diff
@@ -292,7 +292,12 @@ Describe 'ValidateSet support a dynamically generated set' -Tag "CI" {
 
             $testAssemply = "$TestDrive\tst-$(New-Guid).dll"
             Add-Type -TypeDefinition $a -OutputAssembly $testAssemply
-            Import-Module $testAssemply
+            $testModule = Import-Module $testAssemply -PassThru
+        }
+
+        AfterAll {
+            Remove-Module -ModuleInfo $testModule
+            Remove-Item $testAssemply -Force -ErrorAction SilentlyContinue
```
iSazonov: We can remove the line - the Issue (get-help) should be resolved by unloading the test module

daxian-dbw: We already got the `Remove-Item` error and it's not related to the `Remove-Item` line added in this PR. It's because we create the test assembly in `$TestDrive`. Pester removes `$TestDrive` after a test run, and resulted in the error.

How about we don't create the test assembly, but instead do this:
```
$cls = Add-Type -TypeDefinition $a -PassThru | select -first 1
$testModule = Import-Module $cls.Assembly -PassThru
```
I didn't try it myself but think it should work.

iSazonov: Yes, we can - it is fixed in #4117

daxian-dbw:
Wait, we still need `Remove-Module -ModuleInfo $testModule` in the `AfterAll` block, right?

iSazonov:
:-) Ah, sorry - I removed the main fix.

Fixed.

daxian-dbw:
Sign .. 😄 I already added it, now there are 2 `AfterAll` blocks

daxian-dbw:
I will clean it up, just to let you know so that we don't conflict again.

daxian-dbw (approved):

iSazonov:
:-) Ah, I went to bed.

daxian-dbw:
@iSazonov can you please take a look at Gitter? I left a message for you.

PR #4258, opened 2017-07-14
Enable DataRow/DataRowView adapters in PowerShell Core

### LINKED ISSUE #4232: PowerShell Core - System.Data DataRow not providing the correct result.
I didn't pay too much attention in Beta 2 but after upgrading to Beta 3 I'm concern.

Steps to reproduce
------------------
Run Sample Code:
```powershell
## - Connect and run T-SQL query
$Global:getSvr = "mtrinidadlt2,1437";
$Global:con = "server=$Global:getSvr;database=Master;Integrated Security=false;User ID=sa;" + 'Password=$Adm1n!';
$SqlQry = @'
sp_who2
'@;

$sda = New-Object System.Data.SqlClient.SqlDataAdapter ($SqlQry, $global:con)
$sdt = New-Object System.Data.DataTable
$sda.fill($sdt) | Out-Null

$sdt.rows | Select -first 5 | Format-Table
```

Expected behavior
-----------------
DataRow data to provide proper objects results column with its data value. It works in Windows PowerShell.
(See Image)
pswindatarowok_2017-07-12_13-55-58

Actual behavior
---------------
Seems like a Python list array result: (See image)
pscoredatarowissue_2017-07-12_13-55-58

Environment data
----------------
Both OS's:
Windows 10 Build 16237
Ubuntu 16.04.02

```powershell
PS C:\Program Files\PowerShell\6.0.0-beta.3> $PSVersionTable

OS                             Microsoft Windows 10.0.16237

```

daxian-dbw:
Fix #4232 

Enable the DataRow/DataRowView adapters in PowerShell Core by moving the implementation from `ExtraAdapter.cs` to `CoreAdapter.cs`.
Add basic tests too.


src/System.Management.Automation/engine/CoreAdapter.cs
```diff
@@ -5007,6 +5008,221 @@ private static XmlNode[] FindNodes(object obj, string propertyName, StringCompar
         }
     }
 
+    /// <summary>
+    /// Deals with DataRow objects
+    /// </summary>
+    internal class DataRowAdapter : PropertyOnlyAdapter
```
TravisEz13: Why don't we follow the one class per file practice?

TravisEz13 (approved):

src/System.Management.Automation/engine/CoreAdapter.cs
daxian-dbw: Core adapters have always been in the same file. Some of them were moved to `ExtraAdapter.cs` previously due to missing the corresponding type in CoreCLR. Now we are gradually moving them back.

MaximoTrinidad:
Hi @daxian-dbw 

Any idea when can I test this change?  Looking forward to give it a shot.
:)

Thanks,

src/System.Management.Automation/engine/CoreAdapter.cs
TravisEz13: The question is more about the project in general.

daxian-dbw:
@MaximoTrinidad The release of beta.5 will on 8/3. But you can build powershell locally to verify the fix.

Issue #3946, opened 2017-06-06
Enable upgrade (not SxS) install from MSI


MaximoTrinidad:
Just for clarification. So, it's required to first remove the previous version before installing the latest.
Right!

I'm not sure if this could lead to some of the issues I'm experiencing with remoting (maybe).

So, are there any plan to automatically do the upgrade?

![poshinstall](https://cloud.githubusercontent.com/assets/19892435/26809999/d5b54660-4a36-11e7-8369-ef3b9aee0c5a.png)

SteveL-MSFT:
Currently each version has been a side-by-side install on Windows.  We've been discussing internally how to improve the install experience on Windows and hopefully we'll have an RFC posted in the near future.

cc @joeyaiello

SteveL-MSFT:
The leftover files on Linux are due to an older issue you opened https://github.com/PowerShell/PowerShell/issues/3758

joeyaiello:
Renaming this issue to only address the Windows aspect of the problem (given that the Linux part of it is duped).

PR #4259, opened 2017-07-14
Enable MSI upgrades rather than side-by-side install

SteveL-MSFT:
MSI only works against A.B.C type version strings.  So to enable beta.X to beta.X+1 upgrades, we have to use the `AllowSameVersionUpgrades` as they are all effectively 6.0.0 from MSI point of view.

Note that to get Wix toolset working on Win10, you may have to pass `-sval` to light.exe.

Fix https://github.com/PowerShell/PowerShell/issues/3946

### DIFF
--- a/assets/Product.wxs
+++ b/assets/Product.wxs
@@ -36,7 +36,7 @@
     <Condition Message="Supported only on Win8 and above"><![CDATA[ Installed OR $(var.MinOSVersionSupported) ]]></Condition>
 
     <!-- Information About When Older Versions Are Trying To Be Installed-->
-    <MajorUpgrade DowngradeErrorMessage="A newer version of PowerShell is already installed." />
+    <MajorUpgrade AllowSameVersionUpgrades="yes" DowngradeErrorMessage="A newer version of $(env.ProductName) is already installed." />
     
     <!-- Embed Cabinet Files in Product-->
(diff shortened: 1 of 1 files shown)


raghushantha:
@SteveL-MSFT Changes look good. The only thing is to watch out for is that the 'ProductGuid' needs to be the same if you need upgrade to be supported.

For major version changes, ProductGuid can be changed

raghushantha (approved):

SteveL-MSFT:
@raghushantha currently, `Start-PSPackage` has a default static `ProductGuid` so we should be good there.

ChristophB125:
@SteveL-MSFT or @raghushantha  It's great to have upgradability support but does this mean that we lose the ability of having multiple version installed side-by-side? We use older versions of PowerShell Core in our CI system (that is locked down and therefore not updated unless needed) and I always found it handy being able to repro on my local machine using the same version that is being used in CI in order to replicate the exact same behaviour.

SteveL-MSFT:
@ChristophB125 You can still have side-by-side by downloading the zip and extracting the contents in a secure location (this will always be supported).

ChristophB125:
Thanks for the info. The self contained version in a zip file should be sufficient for our use case.

Issue #3791, opened 2017-05-16
Fix PSScriptAnalyzer warnings and errors


SteveL-MSFT:
We should be PSScriptAnalyzer clean and any new PRs should pass PSScriptAnalyzer

```powershell
PS> $e = Get-ChildItem -Recurse -Include *.ps1,*.psm1 | ? { $_.fullname -notmatch "\\bin\\" } | % { Invoke-ScriptAnalyzer -Path $_ }
PS> $e | Group-Object severity

Count Name                      Group
----- ----                      -----
 1528 Warning                   {Microsoft.Windows.PowerShell.ScriptAnalyzer.Generic.DiagnosticRecord, Microsoft.Win...
   20 Error                     {Microsoft.Windows.PowerShell.ScriptAnalyzer.Generic.DiagnosticRecord, Microsoft.Win...
  129 Information               {Microsoft.Windows.PowerShell.ScriptAnalyzer.Generic.DiagnosticRecord, Microsoft.Win...
```

mi-hol:
Affected files identified below are the ones to be fixed, correct?
Count Name
----- ----	
    4 Microsoft.PowerShell.Archive.psm1
    1 Microsoft.PowerShell.Utility.psm1
    4 MSFT_PackageManagement.psm1
    3 MSFT_PackageManagementSource.psm1
    1 PackageManagementDscUtilities.psm1
    9 PackageProviderFunctions.psm1
  115 PSModule.psm1
   74 PSDesiredStateConfiguration.psm1
   14 PSDiagnostics.psm1
    4 Install-PowerShellRemoting.ps1
    3 RegisterManifest.ps1

SteveL-MSFT:
Running off latest master, I got 303 files:

$e | select scriptname -unique

SteveL-MSFT:
@mi-hol if you want to help out, I would suggest not fixing them all at once (which would make the PR review more difficult), but perhaps each folder can be a separate PR.  Thanks!

mi-hol:
@SteveL-MSFT  sure thats the right approach and reason why I've tried to locate the source file for 'Microsoft.PowerShell.Utility.psm1' in this repository but didn't succeed. 
What file would you suggest that I'd try first?

SteveL-MSFT:
Perhaps start with all the psm1 files

rjmholt:
I'll take on the warnings in Import-Module.Tests.ps1

PR #4261, opened 2017-07-14
PSScriptAnalyzer fixes by category

bergmeister:
This is the updated version of PR #4213, in which it was requested to split the commits on a per kind category.
The intend is to help with fixing #3791. The fixed PSScriptAnalyzer warnings are PSPossibleIncorrectComparisonWithNull and PSAvoidUsingCmdletAliases for ForEach-Object, Where-Object and Select-Object since those warning types were responsible for the majority of warnings.


daxian-dbw:
@bergmeister Thank you so much for taking extra effort splitting the original PR by the fix categories!

build.psm1
```diff
@@ -668,7 +668,7 @@ function Get-PesterTag {
     $alltags = @{}
     $warnings = @()
 
-    get-childitem -Recurse $testbase -File |?{$_.name -match "tests.ps1"}| ForEach-Object {
+    get-childitem -Recurse $testbase -File |Where-Object {$_.name -match "tests.ps1"}| ForEach-Object {
```
iSazonov: Please add a space before Where-Object and after `"tests.ps1"}`

demos/Apache/apache-demo.ps1
```diff
@@ -2,7 +2,7 @@ Import-Module $PSScriptRoot/Apache/Apache.psm1
 
 #list Apache Modules
 Write-Host -Foreground Blue "Get installed Apache Modules like *proxy* and Sort by name"
-Get-ApacheModule |Where {$_.ModuleName -like "*proxy*"}|Sort-Object ModuleName | Out-Host
+Get-ApacheModule | Where-Object {$_.ModuleName -like "*proxy*"}|Sort-Object ModuleName | Out-Host
```
iSazonov: Please add spaces in `"*proxy*"}|Sort-Object`.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ImplicitRemotingCommands.cs
```diff
@@ -2139,11 +2139,11 @@ private void GenerateHelperFunctionsWriteMessage(TextWriter writer)
         [Parameter(Mandatory = $false, Position = 1)]
         [bool] $createdByModule = $false)
 
-    if ($PSSession -ne $null)
+    if ($null -ne $PSSession)
     {{
         $script:PSSession = $PSSession
 
-        if ($createdByModule -and ($script:PSSession -ne $null))
+        if ($null -ne $createdByModule -and ($script:PSSession))
```
iSazonov: Please correct.

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -417,7 +417,7 @@ Describe "Proxy module is usable when the original runspace is no longer around"
 
         It "Verifies Remove-Module removed the runspace that was automatically created" -Pending {
             Remove-Module $module -Force
-            ((Get-PSSession -InstanceId $internalSession.InstanceId -ErrorAction SilentlyContinue) -eq $null) | Should Be $true
+            ($null -eq (Get-PSSession -InstanceId $internalSession.InstanceId -ErrorAction SilentlyContinue)) | Should Be $true
```
iSazonov: Please change to `Should BeNullOrEmpty`.

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -452,7 +452,7 @@ Describe "Proxy module is usable when the original runspace is no longer around"
         # removing the module should remove the implicitly/magically created runspace
         It "Verifies Remove-Module removes automatically created runspace" -Pending {
             Remove-Module $module -Force
-            ((Get-PSSession -InstanceId $internalSession.InstanceId -ErrorAction SilentlyContinue) -eq $null) | Should Be $true
+            ($null -eq (Get-PSSession -InstanceId $internalSession.InstanceId -ErrorAction SilentlyContinue)) | Should Be $true
```
iSazonov: Please change to `Should BeNullOrEmpty`.

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -767,7 +767,7 @@ Describe "Import-PSSession functional tests" -tags "Feature" {
     }
 
     It "Helper functions should not be imported" -Skip:$skipTest {
-        ((Get-Item function:*PSImplicitRemoting* -ErrorAction SilentlyContinue) -eq $null) | Should Be $true
+        ($null -eq (Get-Item function:*PSImplicitRemoting* -ErrorAction SilentlyContinue)) | Should Be $true
```
iSazonov: Please change to `Should BeNullOrEmpty`.

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -808,11 +808,11 @@ Describe "Import-PSSession functional tests" -tags "Feature" {
         }
 
         It "Temporary module should be automatically removed after runspace is closed" -Skip:$skipTest {
-            ((Get-Module | Where-Object { $_.Path -eq $module.Path }) -eq $null) | Should Be $true
+            ($null -eq (Get-Module | Where-Object { $_.Path -eq $module.Path })) | Should Be $true
```
iSazonov: Please change to `Should BeNullOrEmpty`.

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -808,11 +808,11 @@ Describe "Import-PSSession functional tests" -tags "Feature" {
...
         }
 
         It "Temporary psm1 file should be automatically removed after runspace is closed" -Skip:$skipTest {
-            ((Get-Item $module.Path -ErrorAction SilentlyContinue) -eq $null) | Should Be $true
+            ($null -eq (Get-Item $module.Path -ErrorAction SilentlyContinue)) | Should Be $true
```
iSazonov: Please change to `Should BeNullOrEmpty`.

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -1737,10 +1737,10 @@ Describe "Implicit remoting tests" -tags "Feature" {
             $module = Import-PSSession -Session $session -Name Get-Variable -Type cmdlet -Prefix My -AllowClobber
             (Get-MyVariable -Name pid).Value | Should Not Be $PID
         } finally {
-            if ($module -ne $null) { Remove-Module $module -Force -ErrorAction SilentlyContinue }
+            if ($null -ne $module) { Remove-Module $module -Force -ErrorAction SilentlyContinue }
         }
 
-        ((Get-Item function:Get-MyVariable -ErrorAction SilentlyContinue) -eq $null) | Should Be $true
+        ($null -eq (Get-Item function:Get-MyVariable -ErrorAction SilentlyContinue)) | Should Be $true
```
iSazonov: Please change to `Should BeNullOrEmpty`.

test/powershell/engine/SemanticVersion.Tests.ps1
```diff
@@ -117,9 +117,9 @@ Describe "SemanticVersion api tests" -Tags 'CI' {
             $operand -eq $operand | Should Be $true
             $operand -ne $operand | Should Be $false
             $null -eq $operand | Should Be $false
-            $operand -eq $null | Should Be $false
+            $null -eq $operand | Should Be $false
+            $null -ne $operand | Should Be $true
```
iSazonov: I believe it is special tests - please revert the changes in the file.

iSazonov:
LGTM.

daxian-dbw:
I don't think we should touch CSharp files and Markdown files. PSScriptAnalyzer warnings should be applicable only to powershell script files.

iSazonov:
Should we fix CSharp files and Markdown files in separate PRs?

daxian-dbw:
I don't think we need to make those changes in non-powershell-script files, since the goal is to avoid PSScriptAnalyzer warnings.

bergmeister:
@daxian-dbw If you say that the only goal is to avoid PSScriptAnalyzer warnings, then I'm not sure if you considered the reason why those rules exist. They are there best practices to avoid unintended/wrong behaviour and therefore apply to any file using PS syntax IMHO. PSPossibleIncorrectComparisonWithNull is the best example where it can result in e.g. an if condition that always evaluates to true. I can understand if you are concerned that in the case of PSPossibleIncorrectComparisonWithNull, changes in the code of the actual product could lead to a regression and should be excluded but in the case of PSAvoidUsingCmdletAliases, I do not see a reason as to why cs/md files should be excluded. WDYT?

daxian-dbw:
@bergmeister Thanks for the comment, and you have very good points. Yes, my main concern about changing the `.cs` files is the potential regression. Given the size of changes, it could be easy for us to overlook something in the review. That being said, I think you are right that the scripts declared in C# code should follow the rules as well, but I would suggest that we leave the comments unchanged in C# code, especially those that still use other aliases.

As for the PSScriptAnalyzer rules, it's very useful to write easy-to-maintain and less error-prone PS script files, but it's not necessarily always applicable to any PSH syntax, for example, the script that serves as a one-liner command for you to run interactively, in which case aliases may be more preferred because they make the command more concise and easy to type. The changes in [`KnownIssues.md`](https://github.com/PowerShell/PowerShell/pull/4261/files#diff-f25db63262b5a4a08e655687e9e2ec18R53) would be a good example of it -- `ls (gci *.txt | % name)` is a command for a user to run interactively, and thus it's OK to keep the aliases. It's a similar idea for scripts in c# comments -- those simple scripts are only for demonstrating the scenarios, and thus it's OK to use the aliases to keep them concise.

iSazonov:
Only we don't type samples from code and md files - we copy-paste 😄

daxian-dbw:
@iSazonov right :) my point is that for the one-liner scripts like those, it's OK to use aliases to keep them concise.
I will do a pass of review on the `.cs` files.

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -417,7 +417,7 @@ Describe "Proxy module is usable when the original runspace is no longer around"
 
         It "Verifies Remove-Module removed the runspace that was automatically created" -Pending {
             Remove-Module $module -Force
-            ((Get-PSSession -InstanceId $internalSession.InstanceId -ErrorAction SilentlyContinue) -eq $null) | Should Be $true
+            (Get-PSSession -InstanceId $internalSession.InstanceId -ErrorAction SilentlyContinue) | Should BeNullOrEmpty
```
daxian-dbw: Should be `Should Be $null`

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -452,7 +452,7 @@ Describe "Proxy module is usable when the original runspace is no longer around"
         # removing the module should remove the implicitly/magically created runspace
         It "Verifies Remove-Module removes automatically created runspace" -Pending {
             Remove-Module $module -Force
-            ((Get-PSSession -InstanceId $internalSession.InstanceId -ErrorAction SilentlyContinue) -eq $null) | Should Be $true
+            (Get-PSSession -InstanceId $internalSession.InstanceId -ErrorAction SilentlyContinue) | Should BeNullOrEmpty
```
daxian-dbw: ditto

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -668,7 +668,7 @@ Describe "Import-PSSession with FormatAndTypes" -tags "Feature" {
         # Should get 2 deserialized S.M.A.H.Coordinates objects
         $results.Count | Should Be 2
         # First object shouldn't have the additional ETS note property
-        $results[0].MyTestLabel -eq $null | Should Be $true
+        $null -eq $results[0].MyTestLabel | Should Be $true
```
daxian-dbw: Use `Should Be $null` ?

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -767,7 +767,7 @@ Describe "Import-PSSession functional tests" -tags "Feature" {
     }
 
     It "Helper functions should not be imported" -Skip:$skipTest {
-        ((Get-Item function:*PSImplicitRemoting* -ErrorAction SilentlyContinue) -eq $null) | Should Be $true
+        (Get-Item function:*PSImplicitRemoting* -ErrorAction SilentlyContinue) | Should BeNullOrEmpty
```
daxian-dbw: Ditto

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -800,19 +800,19 @@ Describe "Import-PSSession functional tests" -tags "Feature" {
...
+            while ( ($i -lt 20) -and ($null -ne (Get-Module | Where-Object { $_.Path -eq $module.Path })) )
             {
                 $i++
                 Start-Sleep -Milliseconds 50
             }
         }
 
         It "Temporary module should be automatically removed after runspace is closed" -Skip:$skipTest {
-            ((Get-Module | ? { $_.Path -eq $module.Path }) -eq $null) | Should Be $true
+            (Get-Module | Where-Object { $_.Path -eq $module.Path }) | Should BeNullOrEmpty
```
daxian-dbw: Ditto

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -800,19 +800,19 @@ Describe "Import-PSSession functional tests" -tags "Feature" {
...
         }
 
         It "Temporary psm1 file should be automatically removed after runspace is closed" -Skip:$skipTest {
-            ((Get-Item $module.Path -ErrorAction SilentlyContinue) -eq $null) | Should Be $true
+            (Get-Item $module.Path -ErrorAction SilentlyContinue) | Should BeNullOrEmpty
```
daxian-dbw: Ditto

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -1737,10 +1737,10 @@ Describe "Implicit remoting tests" -tags "Feature" {
             $module = Import-PSSession -Session $session -Name Get-Variable -Type cmdlet -Prefix My -AllowClobber
             (Get-MyVariable -Name pid).Value | Should Not Be $PID
         } finally {
-            if ($module -ne $null) { Remove-Module $module -Force -ErrorAction SilentlyContinue }
+            if ($null -ne $module) { Remove-Module $module -Force -ErrorAction SilentlyContinue }
         }
 
-        ((Get-Item function:Get-MyVariable -ErrorAction SilentlyContinue) -eq $null) | Should Be $true
+        (Get-Item function:Get-MyVariable -ErrorAction SilentlyContinue) | Should BeNullOrEmpty
```
daxian-dbw: Ditto

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -1779,17 +1779,17 @@ Describe "Implicit remoting tests" -tags "Feature" {
                 $getVariablePid | Should Be $remotePid
 
                 ## Get-Variable function should not be exported when importing a BadVerb-Variable function
-                ((Get-Item Function:\Get-Variable -ErrorAction SilentlyContinue) -eq $null) | Should Be $true
+                Get-Item Function:\Get-Variable -ErrorAction SilentlyContinue | Should BeNullOrEmpty
```
daxian-dbw: Ditto

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -1779,17 +1779,17 @@ Describe "Implicit remoting tests" -tags "Feature" {
...
 
                 ## BadVerb-Variable should be a function, not an alias (1)
-                ((Get-Item Function:\BadVerb-Variable -ErrorAction SilentlyContinue) -ne $null) | Should Be $true
+                Get-Item Function:\BadVerb-Variable -ErrorAction SilentlyContinue | Should BeNullOrEmpty
```
daxian-dbw: Ditto

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -1779,17 +1779,17 @@ Describe "Implicit remoting tests" -tags "Feature" {
...
 
                 ## BadVerb-Variable should be a function, not an alias (2)
-                ((Get-Item Alias:\BadVerb-Variable -ErrorAction SilentlyContinue) -eq $null) | Should Be $true
+                Get-Item Alias:\BadVerb-Variable -ErrorAction SilentlyContinue | Should BeNullOrEmpty
```
daxian-dbw: Ditto

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -1819,17 +1819,17 @@ Describe "Implicit remoting tests" -tags "Feature" {
...
```
daxian-dbw: Ditto

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -1870,14 +1870,14 @@ Describe "Implicit remoting tests" -tags "Feature" {
                 $getVariablePid | Should Be $remotePid
 
                 ## BadVerb-Variable should be an alias, not a function (1)
-                ((Get-Item Function:\BadVerb-Variable -ErrorAction SilentlyContinue) -eq $null) | Should Be $true
+                Get-Item Function:\BadVerb-Variable -ErrorAction SilentlyContinue | Should BeNullOrEmpty
 
                 ## BadVerb-Variable should be an alias, not a function (2)
-                ((Get-Item Alias:\BadVerb-Variable -ErrorAction SilentlyContinue) -ne $null) | Should Be $true
+                Get-Item Alias:\BadVerb-Variable -ErrorAction SilentlyContinue | Should BeNullOrEmpty
```
daxian-dbw: Ditto

test/powershell/engine/Remoting/InvokeCommandRemoteDebug.Tests.ps1
```diff
@@ -157,12 +157,12 @@ Describe "Invoke-Command remote debugging tests" -Tags 'Feature' {
...
         else
         {
-            if ($testDebugger -ne $null) { $testDebugger.Release() }
-            if ($ps -ne $null) { $ps.Dispose() }
-            if ($ps2 -ne $null) { $ps2.Dispose() }
-            if ($rs -ne $null) { $rs.Dispose() }
-            if ($rs2 -ne $null) { $rs2.Dispose() }
-            if ($remoteSession -ne $null) { Remove-PSSession $remoteSession -ErrorAction SilentlyContinue }
+            if ($null -ne $testDebugger) { $testDebugger.Release() }
+            if ($null -ne $psl) { $ps.Dispose() }
```
daxian-dbw: > ($null -ne $psl)

Should be `$ps`

daxian-dbw:
@bergmeister Please push new commits to address the comments, and please **DO NOT** rebase your branch. This way, we can keep the review history which is very important for large size changes like this PR.

daxian-dbw:
@bergmeister I just resolved the 2 conflicts. I think we need faster review iterations on this PR, because there are many other PRs changing the script files covered in this PR and thus there will be more conflicts very soon.

src/Microsoft.PowerShell.GraphicalHost/commandHelpers/ShowCommandHelper.cs
```diff
@@ -183,7 +183,7 @@ Function GetParameterInfo
                 $validateSetAttributes = $parameter.Attributes | Where {
                     [ValidateSet].IsAssignableFrom($_.GetType())
                 }
-                if (($validateSetAttributes -ne $null) -and ($validateSetAttributes.Count -gt 0))
+                if (($null -ne $validateSetAttributes) -and ($validateSetAttributes.Count -gt 0))
```
bergmeister: Good. Although if I had written the code myself I would have rather declared `$validateSetAttributes` as `[System.Array]` because then one would not only avoid the problem that the variable can be null but also prevents gotchas where in the case of array size 1 the type would not be an array and therefore addressing it later as $validateSetAttributes[0] would result in undesired behaviour (in most cases PowerShell would return the first character of the string representation). Just my two cents.

bergmeister:
@daxian-dbw I have applied your requested changes.

daxian-dbw (approved):
LGTM. Thanks @bergmeister!

iSazonov:
@bergmeister Thank you for your patience and hard work!

bergmeister:
@iSazonov @daxian-dbw Thank you both as well. I really had no idea it would turn out to be that complex. I am currently developing a `-AutoFix` switch for `Invoke-PSScriptAnalyzer` that will make it easier for everyone to fix some of the remaining warnings in the future.

Issue #4252, opened 2017-07-14
Invoke-Item -Path does not work 


jpsnover:
PS C:\PowerShell\6.0.0-beta.4> ii -Path c:\
ii : Access is denied
At line:1 char:1
+ ii -Path c:\
+ ~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Invoke-Item], Win32Exception
    + FullyQualifiedErrorId : System.ComponentModel.Win32Exception,Microsoft.PowerShell.Commands.InvokeItemCommand

PS C:\PowerShell\6.0.0-beta.4> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.4
OS                             Microsoft Windows 10.0.16226
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

mklement0:
Just to spell it out: any _directory_ path triggers the bug (as opposed to a _file_ path).

* On Windows, the terminating error cited above occurs.

* On Unix (macOS, Linux), the command is a  quiet no-op (no action, no output, `$?` is `$True`).

MaximoTrinidad:
Hum!  I'm not getting the error and I'm on Windows 10 Insider Build 16237. I can use the Tab key and it auto-completes the line.  I'm getting valid errors.

```
PS C:\Program Files\PowerShell\6.0.0-beta.4> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.4
OS                             Microsoft Windows 10.0.16237
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

Ah!  This is if you don't auto-complete what you're typing.  Press Enter after typing c: then the error happens.  I normally use the tab to auto filled the line.

Definitely the error is not the correct one!

![ps5ii_2017-07-14_10-04-53](https://user-images.githubusercontent.com/19892435/28215508-4c8d6cf0-687c-11e7-8b39-ca6c3125a816.jpg)

[...]

PR #4262, opened 2017-07-15
Support Invoke-Item -Path <folder>

SteveL-MSFT:
On CoreFx, UseShellExecute for Process start is false by default to be cross platform compatible.
In the case of a folder, Process.Start() returns Access Denied as it's not an executable.
On Windows we can use the ShellExecute path to have explorer open the folder.

Fix https://github.com/PowerShell/PowerShell/issues/4252
Fix https://github.com/PowerShell/PowerShell/issues/4282

### DIFF
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -1323,12 +1323,38 @@ namespace Microsoft.PowerShell.Commands
 
             string resource = StringUtil.Format(FileSystemProviderStrings.InvokeItemResourceFileTemplate, path);
 
+#if UNIX
+            void StartProcessWithOpen(string filename)
+            {
+                System.Diagnostics.Process process = new System.Diagnostics.Process();
+                // The file is possibly not an executable, so we try invoking the default program that handles this file.
+                const string quoteFormat = "\"{0}\"";
+                process.StartInfo.FileName = Platform.IsLinux ? "xdg-open" : /* OS X */ "open";
+                if (NativeCommandParameterBinder.NeedQuotes(filename))
+                {
+                    path = string.Format(CultureInfo.InvariantCulture, quoteFormat, filename);
+                }
+                process.StartInfo.Arguments = filename;
+                // xdg-open on Ubuntu outputs debug info to stderr, suppress it
+                process.StartInfo.RedirectStandardError = true;
+                process.Start();
+            }
+#endif
+
             if (ShouldProcess(resource, action))
             {
                 System.Diagnostics.Process invokeProcess = new System.Diagnostics.Process();
 
                 try
                 {
+                    // special case directories on UNIX so that we can open them in the window-manager
+                    // corefx doesn't return error if you try to start a directory
+#if UNIX
+                    if (Directory.Exists(path))
+                    {
+                        StartProcessWithOpen(path);
+                    }
+#endif
                     // Try Process.Start first.
                     //  - In FullCLR, this is all we need to do.
                     //  - In CoreCLR, this works for executables on Win/Unix platforms
@@ -1339,15 +1365,7 @@ namespace Microsoft.PowerShell.Commands
                 catch (Win32Exception ex) when (ex.NativeErrorCode == 13)
                 {
                     // Error code 13 -- Permission denied.
-                    // The file is possibly not an executable, so we try invoking the default program that handles this file.
-                    const string quoteFormat = "\"{0}\"";
-                    invokeProcess.StartInfo.FileName = Platform.IsLinux ? "xdg-open" : /* OS X */ "open";
-                    if (NativeCommandParameterBinder.NeedQuotes(path))
-                    {
-                        path = string.Format(CultureInfo.InvariantCulture, quoteFormat, path);
-                    }
-                    invokeProcess.StartInfo.Arguments = path;
-                    invokeProcess.Start();
+                    StartProcessWithOpen(path);
                 }
 #elif CORECLR
                 catch (Win32Exception ex) when (ex.NativeErrorCode == 193 || ex.NativeErrorCode == 5)
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -1313,9 +1313,11 @@ namespace Microsoft.PowerShell.Commands
         protected override void InvokeDefaultAction(string path)
         {
 #if UNIX
-            const int PERMISSION_DENIED_NOT_EXECUTABLE = 13;
+            // Error code 13 -- Permission denied
+            const int NOT_EXECUTABLE = 13;
 #else
-            const int PERMISSION_DENIED_NOT_EXECUTABLE = 193;
+            // Error code 193 -- BAD_EXE_FORMAT (not a valid Win32 application)
+            const int NOT_EXECUTABLE = 193;
 #endif
 
             if (String.IsNullOrEmpty(path))
@@ -1329,52 +1331,47 @@ namespace Microsoft.PowerShell.Commands
 
             string resource = StringUtil.Format(FileSystemProviderStrings.InvokeItemResourceFileTemplate, path);
 
-            void StartProcessUsingHandler(string filename)
-            {
-                System.Diagnostics.Process process = new System.Diagnostics.Process();
-#if UNIX
-                // The file is possibly not an executable, so we try invoking the default program that handles this file.
-                const string quoteFormat = "\"{0}\"";
-                process.StartInfo.FileName = Platform.IsLinux ? "xdg-open" : /* OS X */ "open";
-                if (NativeCommandParameterBinder.NeedQuotes(filename))
-                {
-                    path = string.Format(CultureInfo.InvariantCulture, quoteFormat, filename);
-                }
-                process.StartInfo.Arguments = filename;
-                // xdg-open on Ubuntu outputs debug info to stderr, suppress it
-                process.StartInfo.RedirectStandardError = true;
-                process.Start();
-#else
-                if (Platform.IsNanoServer || Platform.IsIoT) { throw new NotSupportedException(); }
-                process.StartInfo.FileName = filename;
-                ShellExecuteHelper.Start(process.StartInfo);
-#endif
-            }
-
             if (ShouldProcess(resource, action))
             {
+                var invokeProcess = new System.Diagnostics.Process();
+                invokeProcess.StartInfo.FileName = path;
+                bool invokeDefaultProgram = false;
 
-                try
+                if (Directory.Exists(path) && !Platform.IsNanoServer && !Platform.IsIoT)
                 {
-                    // special case directories on UNIX so that we can open them in the window-manager
-                    // corefx doesn't return error if you try to start a directory
-                    if (Directory.Exists(path))
+                    // Path points to a directory and it's not NanoServer or IoT, so we can opne the file explorer
+                    invokeDefaultProgram = true;
+                }
+                else
+                {
+                    try
                     {
-                        StartProcessUsingHandler(path);
-                    }
-                    else
-                    {
-                        System.Diagnostics.Process invokeProcess = new System.Diagnostics.Process();
-                        // Try Process.Start first.
-                        //  - In FullCLR, this is all we need to do
-                        //  - In CoreCLR, this works for executables on Win/Unix platforms
-                        invokeProcess.StartInfo.FileName = path;
+                        // Try Process.Start first. This works for executables on Win/Unix platforms
                         invokeProcess.Start();
                     }
+                    catch (Win32Exception ex) when (ex.NativeErrorCode == NOT_EXECUTABLE)
+                    {
+                        // The file is possibly not an executable. If it's headless SKUs, rethrow.
+                        if (Platform.IsNanoServer || Platform.IsIoT) { throw; }
+                        // Otherwise, try invoking the default program that handles this file.
+                        invokeDefaultProgram = true;
+                    }
                 }
-                catch (Win32Exception ex) when (ex.NativeErrorCode == PERMISSION_DENIED_NOT_EXECUTABLE)
+
+                if (invokeDefaultProgram)
                 {
-                    StartProcessUsingHandler(path);
+#if UNIX
+                    const string quoteFormat = "\"{0}\"";
+                    invokeProcess.StartInfo.FileName = Platform.IsLinux ? "xdg-open" : /* OS X */ "open";
+                    if (NativeCommandParameterBinder.NeedQuotes(path))
+                    {
+                        path = string.Format(CultureInfo.InvariantCulture, quoteFormat, path);
+                    }
+                    invokeProcess.StartInfo.Arguments = path;
+                    invokeProcess.Start();
+#else
+                    ShellExecuteHelper.Start(invokeProcess.StartInfo);
+#endif
                 }
             }
         } // InvokeDefaultAction
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -1350,9 +1350,10 @@ namespace Microsoft.PowerShell.Commands
                     invokeProcess.Start();
                 }
 #elif CORECLR
-                catch (Win32Exception ex) when (ex.NativeErrorCode == 193)
+                catch (Win32Exception ex) when (ex.NativeErrorCode == 193 || ex.NativeErrorCode == 5)
                 {
                     // Error code 193 -- BAD_EXE_FORMAT (not a valid Win32 application).
+                    // Error code 5 -- ACCESS_DENIED (a folder is not an app)
                     // If it's headless SKUs, rethrow.
                     if (Platform.IsNanoServer || Platform.IsIoT) { throw; }
                     // If it's full Windows, then try ShellExecute.
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -1312,6 +1312,12 @@ namespace Microsoft.PowerShell.Commands
         /// </exception>
         protected override void InvokeDefaultAction(string path)
         {
+#if UNIX
+            const int PERMISSION_DENIED_NOT_EXECUTABLE = 13;
+#else
+            const int PERMISSION_DENIED_NOT_EXECUTABLE = 193;
+#endif
+
             if (String.IsNullOrEmpty(path))
             {
                 throw PSTraceSource.NewArgumentException("path");
@@ -1323,10 +1329,10 @@ namespace Microsoft.PowerShell.Commands
 
             string resource = StringUtil.Format(FileSystemProviderStrings.InvokeItemResourceFileTemplate, path);
 
-#if UNIX
-            void StartProcessWithOpen(string filename)
+            void StartProcessUsingHandler(string filename)
             {
                 System.Diagnostics.Process process = new System.Diagnostics.Process();
+#if UNIX
                 // The file is possibly not an executable, so we try invoking the default program that handles this file.
                 const string quoteFormat = "\"{0}\"";
                 process.StartInfo.FileName = Platform.IsLinux ? "xdg-open" : /* OS X */ "open";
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1350,9 +1350,10 @@ protected override void InvokeDefaultAction(string path)
                     invokeProcess.Start();
                 }
 #elif CORECLR
-                catch (Win32Exception ex) when (ex.NativeErrorCode == 193)
+                catch (Win32Exception ex) when (ex.NativeErrorCode == 193 || ex.NativeErrorCode == 5)
```
daxian-dbw: Can you add `|| ex.NativeErrorCode == 5` to the catch in `Unix` block as well? `ii <folder-path>` doesn't work in Linux, and I believe it's because the `NativeErrorCode` in that case on Unix is also `5` (or a different value).

test/powershell/Modules/Microsoft.PowerShell.Utility/Invoke-Item.Tests.ps1
```diff
@@ -59,6 +59,11 @@ Describe "Invoke-Item basic tests" -Tags "CI" {
         }
         Get-Content $redirectFile -Raw | Should Match "usage: ping"
     }
+
+    It "Should invoke a folder without error" {
+        # can't validate that it actually opened, but no error should be returned
+        Invoke-Item -Path $PSHOME
```
daxian-dbw: On windows we can validate it by using our COM adapter:
```
$s = New-Object -ComObject "Shell.Application"
$w = $s.Windows()

$before = $w.Count
Invoke-Item -Path $PSHOME
$after = $w.Count

$before + 1 | Should Be $after
$item = $w.Item($after - 1)
$item.LocationURL | Should Match ($PSHOME -replace '\\', '/')
## close the windows explorer
$item.Quit()
```
But no easy way to do the same on non-windows platforms.

SteveL-MSFT: Cool!  I'll make the change

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: There's code above this for Unix that is supposed to catch Access Denied for the same situation.  However, `[System.Diagnostics.Process]::Start("/home/steve")` just forks powershell with the new folder.  Need to follow-up with corefx.

SteveL-MSFT: Opened issue in corefx https://github.com/dotnet/corefx/issues/22299

For now, we can special case directories as the existing Unix code path will work using `xdg-open` but it currently doesn't fall through to that code

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1339,20 +1365,13 @@ protected override void InvokeDefaultAction(string path)
...
-                    {
-                        path = string.Format(CultureInfo.InvariantCulture, quoteFormat, path);
-                    }
-                    invokeProcess.StartInfo.Arguments = path;
-                    invokeProcess.Start();
+                    StartProcessWithOpen(path);
                 }
 #elif CORECLR
-                catch (Win32Exception ex) when (ex.NativeErrorCode == 193)
+                catch (Win32Exception ex) when (ex.NativeErrorCode == 193 || ex.NativeErrorCode == 5)
```
iSazonov: Why can't we check `Directory.Exists(path)` on Windows (not Nano and IoT) in line 1353 and exclude the exception in the case?

test/powershell/Modules/Microsoft.PowerShell.Utility/Invoke-Item.Tests.ps1
```diff
@@ -59,6 +59,31 @@ Describe "Invoke-Item basic tests" -Tags "CI" {
...
+            $before + 1 | Should Be $after
+            $item = $windows.Item($after - 1)
+            $item.LocationURL | Should Match ($PSHOME -replace '\\', '/')
+            ## close the windows explorer
+            $item.Quit()
+        }
+        else
+        {
+            # can't validate that it actually opened, but no error should be returned
+            Invoke-Item -Path $PSHOME
```
iSazonov: Can we enhance `Invoke-Item -PassThru`  to return `ProcessInfo`?

iSazonov (commented):
Leave a comment

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: That might be cleaner so that the overall logic is in one place.  I'll change.

SteveL-MSFT: Actually, I think this will actually make the code a bit more complex as I have to check if it's a Directory & not Nano/IoT (headless) for Windows.  The exception is handling cases where you invoke a .gif or .doc so that also needs to check.

test/powershell/Modules/Microsoft.PowerShell.Utility/Invoke-Item.Tests.ps1
SteveL-MSFT: Not sure how well this will work in practice.  When you invoke a file that is not an executable, it gets handed off to something else to start it (xdg-open, open, or shellexecute on Windows) so if .gif starts mspaint.exe you would like to get mspaint.exe ProcessInfo back, but there's no way to do that as it's decoupled.

iSazonov: Closed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: Can we `Directory.Exists(path)` above for Windows and leave Nano/IoT here in the `catch`?

test/powershell/Modules/Microsoft.PowerShell.Utility/Invoke-Item.Tests.ps1
iSazonov: We could use `lsof` to detect that the directory is opened in a process.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: We'd have to do a check if Directory and Not Nano/IoT to exercise the Desktop Windows code path so you'd special case Nano/IoT twice unless I'm misunderstanding you

iSazonov: I guess if we miss the test (Nano/IoT) above and only check Directory , we'll get the exception for Nano/IoT.
Also we can cache the Nano/IoT check and reuse in the catch.
In fact, I don't see perf problems in double check - invoke-item is single operation vs `Dir`.

test/powershell/Modules/Microsoft.PowerShell.Utility/Invoke-Item.Tests.ps1
SteveL-MSFT: I'll add this.  Thanks!

SteveL-MSFT: Not able to figure out correct command to get this to work.  It looks like `lsof` only works with files and not directories.  If you specify a directory, it's looking for any files within that directory and not the directory itself.  Also, if I get Ubuntu GUI to open $pshome directory, it doesn't show up in `lsof` output at all.

```bash
lsof +d $pshome | grep -v "^powershel"
```

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: I think I figured out a way to make the code more clean and readable

test/powershell/Modules/Microsoft.PowerShell.Utility/Invoke-Item.Tests.ps1
iSazonov: lsof (w/o parameters) should show directories too - in Unix directory is file.

>It's probably because the GUI is only enumerating 

I think GUI open the directory, enumerate files and close directory. Then we run lsof...

Another idea is to change a directory default app, ex. script which create a flag file. I did this in ' "Describe "Invoke-Item tests on Windows" ' - maybe we can do the same on Unix.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1323,50 +1329,53 @@ protected override void InvokeDefaultAction(string path)
...
+            void StartProcessUsingHandler(string filename)
+            {
+                System.Diagnostics.Process process = new System.Diagnostics.Process();
+#if UNIX
+                // The file is possibly not an executable, so we try invoking the default program that handles this file.
+                const string quoteFormat = "\"{0}\"";
+                process.StartInfo.FileName = Platform.IsLinux ? "xdg-open" : /* OS X */ "open";
+                if (NativeCommandParameterBinder.NeedQuotes(filename))
+                {
+                    path = string.Format(CultureInfo.InvariantCulture, quoteFormat, filename);
```
iSazonov: Should we check NeedQuotes? Maybe do that for every filename?

daxian-dbw: `NeedQuotes` check is needed only if you are specifying arguments for ProcessInfo, for example, the argument string may have space in it.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1323,50 +1329,53 @@ protected override void InvokeDefaultAction(string path)
...
+                }
+                process.StartInfo.Arguments = filename;
+                // xdg-open on Ubuntu outputs debug info to stderr, suppress it
+                process.StartInfo.RedirectStandardError = true;
+                process.Start();
+#else
+                if (Platform.IsNanoServer || Platform.IsIoT) { throw new NotSupportedException(); }
+                process.StartInfo.FileName = filename;
+                ShellExecuteHelper.Start(process.StartInfo);
+#endif
```
daxian-dbw: So we are not differentiating FullCLR and CoreCLR on Windows. Are we supposed to ignore the FullCLR condition from now on? (which I'm totally fine with :))

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1323,50 +1329,53 @@ protected override void InvokeDefaultAction(string path)
...
+                // The file is possibly not an executable, so we try invoking the default program that handles this file.
+                const string quoteFormat = "\"{0}\"";
+                process.StartInfo.FileName = Platform.IsLinux ? "xdg-open" : /* OS X */ "open";
+                if (NativeCommandParameterBinder.NeedQuotes(filename))
+                {
+                    path = string.Format(CultureInfo.InvariantCulture, quoteFormat, filename);
+                }
+                process.StartInfo.Arguments = filename;
+                // xdg-open on Ubuntu outputs debug info to stderr, suppress it
+                process.StartInfo.RedirectStandardError = true;
```
daxian-dbw: What if `xdg-open` or `open` fails to open a file? In that case, user won't see anything after running `Invoke-Item` and will suspect that `Invoke-Item` doesn't work.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1323,50 +1329,53 @@ protected override void InvokeDefaultAction(string path)
...
+                process.Start();
+#else
+                if (Platform.IsNanoServer || Platform.IsIoT) { throw new NotSupportedException(); }
```
daxian-dbw: `NotSupportedException` should have a meaningful message with it.

daxian-dbw:
How about organizing the code in the following way, so that we can directly use the exception thrown from .NET Core to handle `NanoServer` and `IoT` cases.

```CSharp
#if UNIX
        // Error code 13 -- Permission denied.
        const int NOT_EXECUTABLE = 13;
#else
        // Error code 193 -- BAD_EXE_FORMAT (not a valid Win32 application).
        const int NOT_EXECUTABLE = 193;
#endif

        if (ShouldProcess(resource, action))
        {
            var invokeProcess = new System.Diagnostics.Process();
            bool invokeDefaultProgram = false

            if (Directory.Exists(path) && !Platform.IsNanoServer && !Platform.IsIoT)
            {
                // Path points to a directory and it's not NanoServer or IoT, so we can opne the file explorer
                invokeDefaultProgram = true;
            }
            else
            {
                try
                {
                    // Try Process.Start first. This works for executables on Win/Unix platforms
                    invokeProcess.StartInfo.FileName = path;
                    invokeProcess.Start();
                }
                catch (Win32Exception ex) when (ex.NativeErrorCode == NOT_EXECUTABLE)
                {
                    // The file is possibly not an executable. If it's headless SKUs, rethrow.
                    if (Platform.IsNanoServer || Platform.IsIoT) { throw; }
                    // Otherwise, try invoking the default program that handles this file.
                    invokeDefaultProgram = true
                }
            }

            if (invokeDefaultProgram)
            {
#if UNIX
                const string quoteFormat = "\"{0}\"";
                invokeProcess.StartInfo.FileName = Platform.IsLinux ? "xdg-open" : /* OS X */ "open";
                if (NativeCommandParameterBinder.NeedQuotes(path))
                {
                    path = string.Format(CultureInfo.InvariantCulture, quoteFormat, path);
                }
                invokeProcess.StartInfo.Arguments = path;
                invokeProcess.Start();
#else
                ShellExecuteHelper.Start(invokeProcess.StartInfo);
#endif
            }
        }
```

test/powershell/Modules/Microsoft.PowerShell.Utility/Invoke-Item.Tests.ps1
```diff
@@ -59,6 +59,79 @@ Describe "Invoke-Item basic tests" -Tags "CI" {
...
+                {
+                    Start-Sleep -Milliseconds 100
+                }
+
+                Get-Content $HOME/InvokeItemTest.Success | Should Be $PSHOME
+            }
+            else
+            {
+                # can't validate on MacOS, so just make sure no error
+                Invoke-Item -Path $PSHOME
```
SteveL-MSFT: iOS isn't same as MacOS, but found the way to do it on MacOS.  Will work on this.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: That's probably true.  I guess we'll just have to have the debug spew coming out.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: We're almost at the point where we will remove the FullCLR code...

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: I thought we could remove the 'if' and each name quoted - "filename.txt", "file name.txt".

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: What error an user see if run the ` xdg-open`  or `open`  from command line?

SteveL-MSFT: It's the same stderr if run directly

test/powershell/Modules/Microsoft.PowerShell.Utility/Invoke-Item.Tests.ps1
SteveL-MSFT: Will do this later, don't have Mac at work

SteveL-MSFT: It appears that lsregister is what one would use to work with existing registrations with Launch Services, however, it appears you need to build an app in Xcode to declare registrations with specific MIME types and there's no api to do it (let alone a command line tool).  The other problem is that `lsregister -dump` isn't user friendly.  I don't see this happening for Mac.

iSazonov: I see 😕 I suggest don't block the PR, 'pending' the test on MacOS and open new tracking Issue.

SteveL-MSFT: Created https://github.com/PowerShell/PowerShell/issues/4282

I don't think test should be `Pending` on Mac, we should still not get an exception on Mac, we just can't validate Finder actually opened it to the right location.

iSazonov: I believe the `pending` status is "We can't do it now, but we have to do it later."

daxian-dbw:
Travis CI froze at `Describing Web cmdlets tests using the cmdlet's aliases` for several runs, I'm afraid it could be a side effect from the `xdg-mime` ...

test/powershell/Modules/Microsoft.PowerShell.Utility/Invoke-Item.Tests.ps1
SteveL-MSFT: The problem with marking it `Pending` is that the test isn't run at all

iSazonov: I'm sorry I wasn't clear. I think we should split the test by platforms because every platform has its own way of testing in the case.

SteveL-MSFT:
Any other feedback?

daxian-dbw (approved):

daxian-dbw:
Very interesting approach of using `xdg-mime` and `AppleScript` to validate results!

test/powershell/Modules/Microsoft.PowerShell.Utility/Invoke-Item.Tests.ps1
```diff
@@ -59,6 +61,100 @@ Describe "Invoke-Item basic tests" -Tags "CI" {
...
+Icon=utilities-terminal
+Terminal=true
+Type=Application
+Categories=Application;
+"@
+                    xdg-mime default InvokeItemTest.desktop inode/directory
+                }
+                else
+                {
+                    $supportedEnvironment = $false
```
iSazonov: What environments we skip and why? Could you please add comment?

iSazonov:
Many thanks to @mklement0 for the idea to use  AppleScript!

test/powershell/Modules/Microsoft.PowerShell.Utility/Invoke-Item.Tests.ps1
daxian-dbw: Maybe I merged too fast :) I guess this can happen on a Linux without a desktop.

SteveL-MSFT: Correct

TravisEz13:
@SteveL-MSFT  `Should invoke text file '<TestFile>' without error on Mac` is failing intermittently after this change
https://travis-ci.org/PowerShell/PowerShell/jobs/258709720

mklement0:
I don't know for sure, but I presume that the issue arises from startup timing issues in the `'tell application "TextEdit" to count of windows'` AppleScript command, where the count returned may - unpredictably - either be `0` or `1` (or greater), if previously opened windows are automatically being restored (which macOS apps do by default).  

Though - unless a previous run failed - a previous test run shouldn't leave an open window behind (I don't know how these tests are being run).

Note that in a given macOS session this is not a concern with _Finder_, which can be assumed to be always running.

A more defensive approach that _may_ fix the issue is:

* to _start_ `TextEdit` with an initial, separate command that explicitly bypasses restoring previously opened windows: `open -F -a TextEdit`.

* after the test, to _quit_ `TextEdit` with `'tell application "TextEdit" to quit' | osascript`

SteveL-MSFT:
I've submitted a [PR](https://github.com/PowerShell/PowerShell/pull/4382) to address this.  First, I didn't realize the `invoke-item` tests were CI, they should have been Feature.  Second, I increased the timeout waiting for Finder since it seems slower on Travis-CI.

mklement0:
@SteveL-MSFT: If the increased timeout (for _TextEdit_, not Finder) fixes the issue, that's great.

Again, speaking in the abstract, not knowing the details of how Travis-CI runs, purely from a macOS / AppleScript perspective:

Note how the failing test expected _3_ open windows, suggesting that TextEdit either was already open with 2 windows, or, more likely, that it had been open in a _previous_ session with 2 windows, which, on restarting, TextEdit _implicitly_ tries to reopen, which is a standard macOS feature named [_Resume_](https://support.apple.com/en-us/HT204005).

It is the latter scenario that can cause unpredictable behavior, for 2 reasons:

* If TextEdit happens _not_ to be running at the time of executing AppleScript `tell application "TextEdit" to count windows`, the window count reported is _unreliable_, because there can be race conditions that result in a lower window count being reported than the actual count of windows ultimately restored.

* If TextEdit is already running and you're reopening an _already opened_ document, the window count will not change, because the existing window is simply _activated_.

For that reason, the defensive approaches I recommended earlier should generally make for a more robust solution (open TextEdit _without_ restoring previously opened windows, _quit_ TextEdit after the test is done).

mklement0:
@SteveL-MSFT: 

P.S.: Taking a step back: Arguably, in a CI environment, any macOS image should be configured with the Resume feature turned _off_ _system-wide_ (that feedback is probably for Travis CI):

* To turn Resume off persistently system-wide:

    * programmatically:

          defaults write NSGlobalDomain NSQuitAlwaysKeepsWindows -bool false

    * interactively: 
          
         Go to `System Preferences > General`  and check `Close windows when quitting an app`.

SteveL-MSFT:
@mklement0 I see what you're saying now.  I'll update it if it becomes an issue, currently have other issues to address.  I appreciate the information you're providing.

PR #4270, opened 2017-07-17
Add unhandled assemblyname characters

TimCurwick:
Fixes issue 4193, wherein it was discussed that the apostrophe, equals sign, and tilde are the last three visible en-us characters that are valid Windows file path characters but cause the assemblyname constructor to throw an error, thereby preventing scripts with class definitions from running when used in the script path. This effectively extends PR 4136 which did this for the comma.

### DIFF
--- a/src/System.Management.Automation/engine/parser/PSType.cs
+++ b/src/System.Management.Automation/engine/parser/PSType.cs
@@ -1119,6 +1119,9 @@ namespace System.Management.Automation.Language
                                       .Replace('\\', (char)0x29f9)
                                       .Replace('/', (char)0x29f9)
                                       .Replace(',', (char)0x201a)
+                                      .Replace('\'', (char)0x2019)
+                                      .Replace('=', (char)0xff1d)
+                                      .Replace('~', (char)0x223c)
                                       .Replace(':', (char)0x0589));
 
             var assembly = AssemblyBuilder.DefineDynamicAssembly(new AssemblyName(assemblyName),
--- a/src/System.Management.Automation/engine/parser/PSType.cs
+++ b/src/System.Management.Automation/engine/parser/PSType.cs
@@ -1120,6 +1120,7 @@ namespace System.Management.Automation.Language
                                       .Replace('/', (char)0x29f9)
                                       .Replace(',', (char)0x201a)
                                       .Replace('\'', (char)0x2019)
+                                      .Replace('"', (char)0x201d)
                                       .Replace('=', (char)0xff1d)
                                       .Replace('~', (char)0x223c)
                                       .Replace(':', (char)0x0589));
--- a/test/powershell/Language/Classes/Scripting.Classes.RunPath.Tests.ps1
+++ b/test/powershell/Language/Classes/Scripting.Classes.RunPath.Tests.ps1
@@ -1,8 +1,17 @@
 Describe "Script with a class definition run path" -Tags "CI" {
 
+    if ( $IsLinux -or $IsOSX )
+    {
+        $UnhandledName = "My,'`"=~Test.ps1"
+    }
+    else
+    {
+        $UnhandledName = "My,'=~Test.ps1"
+    }
+
     $TestCases = @(
-        @{ FileName =     'MyTest.ps1'; Name = 'typical path' }
-        @{ FileName = "My,'=~Test.ps1"; Name = 'path with unhandled assemblyname characters' }
+        @{ FileName = 'MyTest.ps1'  ; Name = 'typical path' }
+        @{ FileName = $UnhandledName; Name = 'path with unhandled assemblyname characters' }
     )
 
     It "Script with a class definition can run from a <Name>" -TestCases $TestCases {
--- a/test/powershell/Language/Classes/Scripting.Classes.RunPath.Tests.ps1
+++ b/test/powershell/Language/Classes/Scripting.Classes.RunPath.Tests.ps1
@@ -1,8 +1,8 @@
 Describe "Script with a class definition run path" -Tags "CI" {
 
     $TestCases = @(
-        @{ FileName =  'MyTest.ps1'; Name = 'path without a comma' }
-        @{ FileName = 'My,Test.ps1'; Name = 'path with a comma'    }
+        @{ FileName =     'MyTest.ps1'; Name = 'typical path' }
+        @{ FileName = "My,'=~Test.ps1"; Name = 'path with unhandled assemblyname characters' }
     )
 
     It "Script with a class definition can run from a <Name>" -TestCases $TestCases {


iSazonov:
Before fixing the Issue I would prefer to open Issue-Question in CoreFX repo to find root codes and root rules - we should understand that we should fix. What if a char disallowed on first place but allowed on next places? Are there disallowed char combinations? Otherwise, we can fix this code indefinitely.

TimCurwick:
I agree that the permanent fix is in the .Net code. But in the mean time, I need my PowerShell 6.0 scripts to run on all of my Windows machines whether or not they have some future, fixed version of .Net installed. We probably can't find and workaround every problem character. We have to draw the line somewhere. I would be content if the consensus is for the line to be where it is at the moment, as it was the comma that we fixed a couple weeks ago that was personally impacting me. But a less arbitrary line would be to fix (or workaround) all affected characters in the base set of ASCII characters, which is what this PR does.

daxian-dbw (dismissed):

daxian-dbw:
> Before fixing the Issue I would prefer to open Issue-Question in CoreFX repo to find root cause

Totally agreed. @TimCurwick Can you please open an issue in dotnet/corefx repo?

> We have to draw the line somewhere.

Yes, we should. Frankly speaking, it should be very rare to have chars like `,'=~` in a file path, but I agree that working around all affected chars in the base set of ASCII is a reasonable line. Even though adding `'=~` would cover the base set of ASCII chars on Windows, it's not the case on Unix platforms. On Unix platforms, you can have any characters in a file path, and I guess at least `"` will cause the same problem.

TimCurwick:
> Frankly speaking, it should be very rare to have chars like ,'=~ in a file path
I agree on '=~. The comma on the other hand (which we previously fixed and is not part of this PR), is becoming more common. As companies move to OneDrive for Business, they tend to embed their full legal company name in the path, often including commas.

> On Unix platforms, you can have any characters in a file path, and I guess at least " will cause the same problem.
Can someone test this? I don't have simple access to such a test box. Copy, modify and run the pester test from this PR to test a double quote.

daxian-dbw:
> The comma, on the other hand, is becoming more common. As companies move to OneDrive for Business, they tend to embed their full legal company name in the path, often including commas.

Thanks for this info! I didn't know of this 👍 

> Can someone test this? I don't have simple access to such a test box. Copy, modify and run the pester test from this PR to test a double quote.

double quote raises the same exception:
```
PS:49> Get-Content './te"st.ps1'                                                                                          
class MyClass { static [string]$MyProperty = 'Some value' }
 [MyClass]::MyProperty
PS:50> 
PS:50> . './te"st.ps1'                                                                                                    
The given assembly name or codebase was invalid. (Exception from HRESULT: 0x80131047)
At line:1 char:1
+ . './te"st.ps1'
+ ~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (:) [], FileLoadException
    + FullyQualifiedErrorId : System.IO.FileLoadException
```
But it works fine for these chars `*?<>|`.

iSazonov:
From MSDN docs for [AssemblyName Constructor (String)](https://msdn.microsoft.com/en-us/library/b6ccxba0(v=vs.110).aspx):
>The supplied assemblyName is **parsed**, and the appropriate fields of the new AssemblyName are initialized with values from the display name. This is the recommended way of parsing display names. Writing your own code to parse display names is not recommended. 

[BNF notation](https://docs.microsoft.com/en-us/dotnet/framework/reflection-and-codedom/specifying-fully-qualified-type-names).
Sample: `com.microsoft.crypto, Culture=en, PublicKeyToken=a5d015c7d5a0b012, Version=1.0.0.0`

I guess the CoreFX parser is here:
https://github.com/dotnet/coreclr/blob/32f0f9721afb584b4a14d69135bea7ddc129f755/src/binder/assemblyname.cpp#L183
https://github.com/dotnet/coreclr/blob/32f0f9721afb584b4a14d69135bea7ddc129f755/src/binder/textualidentityparser.cpp#L236

I have no full understanding why it disallow some characters.

Another thoughts. Can we refactor the code:
```c#
string assemblyName = ClrFacade.FIRST_CHAR_PSASSEMBLY_MARK + (string.IsNullOrWhiteSpace(rootAst.Extent.File)  
                                          ? "powershell"  
                                          : rootAst.Extent.File  

var assemblyObj = new AssemblyName(assemblyName);
assemblyObj.Name = assemblyName;

var assembly = AssemblyBuilder.DefineDynamicAssembly(assemblyObj,  
                                                                   AssemblyBuilderAccess.RunAndCollect, GetAssemblyAttributeBuilders()); 
var module = assembly.DefineDynamicModule(assemblyName); 

```

TimCurwick:
> Another thoughts. Can we refactor the code

I was thinking the same thing and working on a build to test the idea. I'm not sure your code snippet will quite work as it would still send the unhandled characters to the AssemblyName constructor which is what throws the error. But this should work:
`string assemblyName = ClrFacade.FIRST_CHAR_PSASSEMBLY_MARK + (string.IsNullOrWhiteSpace(rootAst.Extent.File)  
                                          ? "powershell"  
                                          : rootAst.Extent.File  

var assemblyNameObj = new AssemblyName("dummy");
assemblyNameObj .Name = assemblyName;
var assembly = AssemblyBuilder.DefineDynamicAssembly(assemblyNameObj ,  
                                                                   AssemblyBuilderAccess.RunAndCollect, GetAssemblyAttributeBuilders()); `

At least, this would prevent the error when creating the AssemblyName. My question is whether creating the AssemblyName is the only place in the code that throws errors when the unhandled characters are used.

iSazonov:
Sorry for my typo - we can remove "dummy":
```c#
var assemblyNameObj = new AssemblyName();
```

src/System.Management.Automation/engine/parser/PSType.cs
```diff
@@ -1119,6 +1119,10 @@ internal static Assembly DefineTypes(Parser parser, Ast rootAst, TypeDefinitionA
                                       .Replace('\\', (char)0x29f9)
```
lzybkr: I think it's time to convert this to a loop. This way we have 2 allocations - 1 StringBuilder and 1 string (the final string) instead of possibly 8 strings, and only 1 pass over the filename.

iSazonov:
@TimCurwick Could you please continue?

iSazonov:
@TimCurwick Please address two points:

1. Use `var assemblyNameObj = new AssemblyName();`
2. Use StringBuilder as Jason requested.

lzybkr:
I have a completely different approach to addressing this problem and retaining the ability of mapping assembly to script - I'll be submitting a PR soon, so I'm closing this one.

PR #4271, opened 2017-07-17
Register-PSSessionConfiguration fails if SesionConfig folder doesn't exist

SteveL-MSFT:
Create SessionConfig folder if it doesn't exist for Register-PSSessionConfiguration

### DIFF
--- a/src/System.Management.Automation/engine/remoting/commands/CustomShellCommands.cs
+++ b/src/System.Management.Automation/engine/remoting/commands/CustomShellCommands.cs
@@ -964,7 +964,13 @@ else
                     }
                 }
 
-                string destPath = System.IO.Path.Combine(Utils.DefaultPowerShellAppBase, "SessionConfig",
+                string destFolder = System.IO.Path.Combine(Utils.DefaultPowerShellAppBase, "SessionConfig");
+                if (!Directory.Exists(destFolder))
+                {
+                    Directory.CreateDirectory(destFolder);
+                }
+
+                string destPath = System.IO.Path.Combine(destFolder,
                     shellName + "_" + sessionGuid.ToString() + StringLiterals.PowerShellDISCFileExtension);
                 if (string.Equals(ProcessorArchitecture, "x86", StringComparison.OrdinalIgnoreCase))
                 {
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
@@ -2108,7 +2108,10 @@ Describe "Import-PSSession on Restricted Session" -tags "Feature","RequireAdminO
         }
     }
 
-    It "Verifies that Import-PSSession works on a restricted session" {
+    # Blocked by https://github.com/PowerShell/PowerShell/issues/4275
+    # Need a way to created restricted endpoint based on a different endpoint other than microsoft.powershell which points
+    # to Windows PowerShell 5.1
+    It "Verifies that Import-PSSession works on a restricted session" -Pending {
 
         $errorVariable = $null
         Import-PSSession -Session $session -AllowClobber -ErrorVariable $errorVariable


PaulHigin (dismissed):
LGTM

SteveL-MSFT:
@PaulHigin Windows PowerShell 5.1 needs this change as well before this test will pass, right (since the default remoting endpoint still goes to Windows PowerShell)?  If so, I'll change this test to `Pending`

PaulHigin:
@SteveL-MSFT I thought that we use a new plugin for PowerShell 6.0.  Not sure but I know a new plugin was created for Core.  @JamesWTruher can you comment?

Regarding PowerShell 5.1, the SessionConfig folder should always be created via Windows manifest.  I feel this change is only needed for the case where the folder is inadvertently deleted (through some previous test) or for PowerShell 6.0 when it is never created.

SteveL-MSFT:
Sync'd with @PaulHigin off line, this test will still fail because the default endpoint still points to 5.1 which doesn't have the fix to expose `-PowerShellVersion` yet.  Rather than making this test pending, I think it makes sense to have the default remoting to go to 6.0, I'll look into that.

daxian-dbw:
> Rather than making this test pending, I think it makes sense to have the default remoting to go to 6.0, I'll look into that.

@PaulHigin @SteveL-MSFT Will that be a separate PR? If so, I will merge this one.

SteveL-MSFT:
@daxian-dbw I'm going to make it part of this PR otherwise the test will still fail

SteveL-MSFT:
I'm going to make this test `Pending` as it's blocked by https://github.com/PowerShell/PowerShell/issues/4275

daxian-dbw (approved):

daxian-dbw:
@SteveL-MSFT one question before merging the PR: the test in question has been there for a long time, so why did it start to fail recently?

SteveL-MSFT:
@daxian-dbw that specific test was added recently by @PaulHigin https://github.com/SteveL-MSFT/PowerShell/commit/7fa0dd054992082787bd1ef7e99bc0f2c9e2a05c

PaulHigin:
It looks like similar implicit remoting tests in Implicit.Remoting.Tests.ps1 are disabled for core and we should do the same for this test.  Is there an Issue for this?

src/System.Management.Automation/engine/remoting/commands/CustomShellCommands.cs
```diff
@@ -964,7 +964,13 @@ private string ConstructPluginContent(out string srcConfigFilePath, out string d
                     }
                 }
 
-                string destPath = System.IO.Path.Combine(Utils.DefaultPowerShellAppBase, "SessionConfig",
+                string destFolder = System.IO.Path.Combine(Utils.DefaultPowerShellAppBase, "SessionConfig");
+                if (!Directory.Exists(destFolder))
+                {
+                    Directory.CreateDirectory(destFolder);
```
iSazonov: Sorry for being late - what is the user haven't permissions?

daxian-dbw: Good catch! For example, on Linux, powershell is installed at /opt/microsoft/powershell, and you don't have permission to create a folder unless using `sudo`.

PaulHigin: @iSazonov Thanks, we will have to think about how this will work and where configuration information goes for Linux cases, but for now I think letting an "access denied" error propagate is fine since these cmdlets cannot work correctly without access.

PaulHigin: I should mention that these cmdlets are currently implemented only on Windows platform.

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -2108,7 +2108,10 @@ Describe "Import-PSSession on Restricted Session" -tags "Feature","RequireAdminO
         }
     }
 
-    It "Verifies that Import-PSSession works on a restricted session" {
+    # Blocked by https://github.com/PowerShell/PowerShell/issues/4275
+    # Need a way to created restricted endpoint based on a different endpoint other than microsoft.powershell which points
+    # to Windows PowerShell 5.1
+    It "Verifies that Import-PSSession works on a restricted session" -Pending {
```
PaulHigin: I just noticed that similar implicit remoting tests are disabled using special checks:
```
# Skip tests for CoreCLR for now
# Skip tests on ARM
$skipTest = $IsCoreCLR -or $env:PROCESSOR_ARCHITECTURE -eq 'ARM'

if ($skipTest) { return }
```
We should make this test be the same since remoting really does not work yet with PowerShell 6.0 based endpoints.  Or maybe we should mark all tests that aren't supported as "pending" since we no longer support FullCLR.

SteveL-MSFT: @PaulHigin can you open a separate issue for this?

PaulHigin: Done.  Issue #4289

src/System.Management.Automation/engine/remoting/commands/CustomShellCommands.cs
iSazonov: Could you please open a tracking Issue?

PR #4273, opened 2017-07-17
Run CI and Feature tests by default with Start-PSPester

SteveL-MSFT:
Nightly keeps breaking because people forget to run Feature tests (I've done this myself).  Make CI, Feature default for local runs.

Reviewed the AppVeyor and Travis ps1 file and they explicitly include CI or exclude Feature so we should be good


TravisEz13 (approved):

JamesWTruher (approved):
this is fine - we may decide to remove the SLOW exclusion, but that need not be done now

Issue #4254, opened 2017-07-14
termination on ctrl + break


mwrock:
I'm working on an application that launches powershell scripts (using powershell core) to setup and launch processes. I want to be able to send ctrl+break or ctrl+c signals to these powershell processes and their child processes to cleanly shut them down.

When I create the powershell process I set `CREATE_NEW_PROCESS_GROUP` which implicitly calls `SetConsoleCtrlHandler(NULL,TRUE)` which effectively turns off ctrl+c for the created process and its children. So when stopping my launched powershell processes, I call `GenerateConsoleCtrlEvent` and send the powershell's process group a ctrl+break. While this effectively closes child processes of the powershell process, it puts the powershell process into debug mode which is undesirable.

I have tried 2 strategies to work around this both fail in different ways:

1. In the powershell process I launch, I call `SetConsoleCtrlHandler(NULL,FALSE)` to turn ctrl+c back on and have my application generate ctrl+c events. This works perfectly the first time. However, oddly, the parent console now no longer seems to propogate ctrl+c to child processes. For example, if I run my app and start some ps processes and succesfully stop them with ctrl+c events and then terminate my app and THEN run `ping 8.8.8.8 -t`, ctrl+c does nothing.

2. I register a new handler to capture BREAK events in the powershell process in hopes this will override the powershell behavior that enters debug mode. However, when this handler is called, even if I have it do absolutely nothing, via a ctrl+break signal, an access violation is thrown in the powershell process.

Is there a clean way to disable the debug behavior invoked from ctrl+break or is there any thoughts why ctrl+c signals do not get propogated if I turn that handler on?

mwrock:
I've done a bit more troubleshooting and research here.

I have nailed down why the first workaround above is failing. Watching process explorer while my ctrl+c was being swallowed, I found that conhost had some orphaned handles and closing those handles via process explorer magically "woke up" ctrl+c. Why those handles are not being closed is a mystery but I also discovered that firing `GenerateConsoleCtrlEvent` after a user invoked ctrl+c is what puts things into this state. Of course if a user enters ctrl+c, the signal is sent to all processes in the console so the call to `GenerateConsoleCtrlEvent` that my application makes directed to a specific process group is unnecesary. I can't find any documentation verifying this but it certainly appears to be the case.

[...]

PR #4283, opened 2017-07-18
if running noninteractively then do not break into debugger on ctrl +…

mwrock:
This addresses #4254 by mimicing ctrl+c behavior on ctrl+break signals when running under `-NonInteractive`. While the current ctrl+break behavior of starting a debugger is excellent for debugging scripts interactively, it may pose a dilemma when running scripts noninteractively. This is especially true, as discussed in #4254, if the scripts are managed by a parent application that wishes to signal spawned powershell processes to cease execution gracefully.

I have built this PR and run it in my [process supervisor application](https://github.com/habitat-sh/habitat) validating that it has the desired effect of gracefully terminating the pipeline when recieving a ctrl+break from `GenerateConsoleCtrlEvent` to its process group.

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
@@ -304,11 +304,19 @@ namespace Microsoft.PowerShell
             switch (args.SpecialKey)
             {
                 case ConsoleSpecialKey.ControlC:
-                    SpinUpBreakHandlerThread(false);
+                    SpinUpBreakHandlerThread(shouldEndSession: false);
                     return;
                 case ConsoleSpecialKey.ControlBreak:
-                    // Break into script debugger.
-                    BreakIntoDebugger();
+                    if (s_cpp.NonInteractive)
+                    {
+                        //ControlBreak mimics ControlC in Noninteractive shells
+                        SpinUpBreakHandlerThread(shouldEndSession: true);
+                    }
+                    else
+                    {
+                        // Break into script debugger.
+                        BreakIntoDebugger();
+                    }
                     return;
             }
         }
@@ -325,13 +333,21 @@ namespace Microsoft.PowerShell
             switch (signal)
             {
                 case ConsoleControl.ConsoleBreakSignal.CtrlBreak:
-                    // Break into script debugger.
-                    BreakIntoDebugger();
+                    if (s_cpp.NonInteractive)
+                    {
+                        //ControlBreak mimics ControlC in Noninteractive shells
+                        SpinUpBreakHandlerThread(shouldEndSession: true);
+                    }
+                    else
+                    {
+                        // Break into script debugger.
+                        BreakIntoDebugger();
+                    }
                     return true;
 
                 // Run the break handler...
                 case ConsoleControl.ConsoleBreakSignal.CtrlC:
-                    SpinUpBreakHandlerThread(false);
+                    SpinUpBreakHandlerThread(shouldEndSession: false);
                     return true;
 
                 case ConsoleControl.ConsoleBreakSignal.Logoff:
@@ -343,12 +359,12 @@ namespace Microsoft.PowerShell
 
                 case ConsoleControl.ConsoleBreakSignal.Close:
                 case ConsoleControl.ConsoleBreakSignal.Shutdown:
-                    SpinUpBreakHandlerThread(true);
+                    SpinUpBreakHandlerThread(shouldEndSession: true);
                     return false;
 
                 default:
                     // Log as much sqm data as possible before we exit.
-                    SpinUpBreakHandlerThread(true);
+                    SpinUpBreakHandlerThread(shouldEndSession: true);
                     return false;
             }
         }


src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -325,8 +333,16 @@ private static bool MyBreakHandler(ConsoleControl.ConsoleBreakSignal signal)
             switch (signal)
             {
                 case ConsoleControl.ConsoleBreakSignal.CtrlBreak:
-                    // Break into script debugger.
-                    BreakIntoDebugger();
+                    if (s_cpp.NonInteractive)
+                    {
+                        //ControlBreak mimics ControlC in Noninteractive shells
+                        SpinUpBreakHandlerThread(false);
```
PaulHigin: Since this performs the same function as Ctrl+C signal, why not just use that?  I thought we were going to implement Ctrl+Break as it was implemented before (stops running pipeline and exits process).

If so then it seems we should call the break handler like this:
```
SpinUpBreakHandlerThread(true);
```
BTW the previous implementation (before changing it to break into debugger) was to make absolutely sure that the process ends.  I don't think we need to be this draconian but this was how it was implemented before:
```
                    // The Big Red Button!

                    // check if we are in a pushed session
                    // if so exit out of it
                    if (ConsoleHost.SingletonInstance.IsRunspacePushed)
                    {
                        ConsoleHost.SingletonInstance.PopRunspace();
                        HandleBreak();
                        return true;
                    }

                    // Log all sqm data before we exit.
                    System.Management.Automation.Sqm.PSSQMAPI.LogAllDataSuppressExceptions();
                    ConsoleHost.SingletonInstance.shouldEndSession = true;

                    ConsoleHandle handle = ConsoleControl.GetActiveScreenBufferHandle();
                    ConsoleControl.WriteConsole(handle, "\n" + ConsoleHost.exitOnCtrlBreakMessage);

                    unchecked
                    {
                        Environment.Exit((int)ExitCodeCtrlBreak);
                    }
```

PaulHigin (changes requested):

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
mwrock: @PaulHigin because I launch powershell in its own process group, ctrl+C is implicitly disabled for the lauched process and its children (see [Remarks section](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682425(v=vs.85).aspx) of `CreateProcess`). This allows parent applications like my own to better control the propogation of `SIGINT` signals and prevent ctrl+c from broadcasting to all processes in the console.

I'll change `shouldEndSession` to `true`. I was blindly looking to imitate ctrl+c but `true` does make more sense.  Thanks for the above snippet of the former behavior. I was looking for the state before entering the debugger yesterday and couldn't find it in the history.

I think sticking with `SpinUpBreakHandlerThread` is likely the best pattern to follow now.

mwrock:
changed `shouldEndSession` to `true`

PaulHigin (approved):
LGTM

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -307,8 +307,16 @@ private static void MyBreakHandler(object sender, ConsoleCancelEventArgs args)
                     SpinUpBreakHandlerThread(false);
                     return;
                 case ConsoleSpecialKey.ControlBreak:
-                    // Break into script debugger.
-                    BreakIntoDebugger();
+                    if (s_cpp.NonInteractive)
+                    {
+                        //ControlBreak mimics ControlC in Noninteractive shells
+                        SpinUpBreakHandlerThread(true);
```
adityapatwardhan: Please use named parameters.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -325,8 +333,16 @@ private static bool MyBreakHandler(ConsoleControl.ConsoleBreakSignal signal)
             switch (signal)
             {
                 case ConsoleControl.ConsoleBreakSignal.CtrlBreak:
-                    // Break into script debugger.
-                    BreakIntoDebugger();
+                    if (s_cpp.NonInteractive)
+                    {
+                        //ControlBreak mimics ControlC in Noninteractive shells
+                        SpinUpBreakHandlerThread(true);
```
adityapatwardhan: Please use named parameters.

adityapatwardhan (changes requested):

mwrock:
@adityapatwardhan for consistency, I added the named params to all calls to `SpinUpBreakHandlerThread` and not just my additions.

PaulHigin:
@mwrock I like this with named parameters much better.  Thanks.

iSazonov:
Can we add tests?

mwrock:
I took a quick look @iSazonov at the pester tests covering ConsoleHost. Creating tests around sending  signals to the console is problematic because the spawned process needs to be created with the `CREATE_NEW_PROCESS_GROUP` flag. C#/Powershell's `[Process]::Start` API does not expose the passing of creation flags. Without its own process group, broadcasting a `ctrl+break` would be captured by pester and would terminate the tests.

I could certainly adjust the ConsoleHost tests to leverage `DllImport` and call directly into `CreateProcessW` but thats a rather heavy handed call and  requires setting up a bunch of structs to support the `CreateProcessW` args. So I'm gonna let that go here.

I'm totally open to other suggestions though!

PaulHigin:
I feel a test for this is unnecessary.

mwrock:
I don't at all mean to be pushy, but would love to see this merged and included in the next release.

adityapatwardhan (approved):

adityapatwardhan:
@mwrock  Thank you for your contribution!

Issue #4192, opened 2017-07-06
Could not load file or assembly error when loading a class library from a module in beta.3


barcexwiki:
I have a C# Powershell module compiled with netcoreapp2.0 that uses a C# class also library compiled with netcoreapp2.0.

Since beta.3 when trying to load the class library from the module I am seeing a 'Could not load file or assembly' error. The same code works just fine in beta.2.

I have created an example case in the following repo with an example of the issue: https://github.com/barcexwiki/cannotloadassembly

In the repo, project 'psmodule' is a module that has a single cmdlet: Get-HelloWorld. This cmdlet calls an static method in a class in the 'classlibrary' project and outputs "Hello World from classlibrary!".

Steps to reproduce
------------------

```
D:\>git clone https://github.com/barcexwiki/cannotloadassembly.git
Cloning into 'cannotloadassembly'...
remote: Counting objects: 8, done.
remote: Compressing objects: 100% (8/8), done.
remote: Total 8 (delta 0), reused 8 (delta 0), pack-reused 0
Unpacking objects: 100% (8/8), done.
Checking connectivity... done.

D:\>cd cannotloadassembly\psmodule

D:\cannotloadassembly\psmodule>dotnet restore
  Restoring packages for D:\cannotloadassembly\classlibrary\classlibrary.csproj...
  Restoring packages for D:\cannotloadassembly\psmodule\psmodule.csproj...
  Generating MSBuild file D:\cannotloadassembly\classlibrary\obj\classlibrary.csproj.nuget.g.props.
  Generating MSBuild file D:\cannotloadassembly\classlibrary\obj\classlibrary.csproj.nuget.g.targets.
  Restore completed in 296,18 ms for D:\cannotloadassembly\classlibrary\classlibrary.csproj.
  Generating MSBuild file D:\cannotloadassembly\psmodule\obj\psmodule.csproj.nuget.g.props.
  Generating MSBuild file D:\cannotloadassembly\psmodule\obj\psmodule.csproj.nuget.g.targets.
  Restore completed in 2,46 sec for D:\cannotloadassembly\psmodule\psmodule.csproj.

D:\cannotloadassembly\psmodule>dotnet build
Microsoft (R) Build Engine version 15.3.388.41745 for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  classlibrary -> D:\cannotloadassembly\classlibrary\bin\Debug\netcoreapp2.0\classlibrary.dll
[...]

PR #4285, opened 2017-07-18
Add test to validate 'Assembly.LoadFrom' behavior

daxian-dbw:
Add validation test for #4192


test/powershell/engine/Basic/Assembly.LoadFrom.ps1
```diff
@@ -0,0 +1,66 @@
...
+                {
+                    public static string GetProviderName()
+                    {
+                        return "Assembly.Bar.Provider";
+                    }
+                }
+            }
+'@
+
+        $TempPath = [System.IO.Path]::GetTempFileName()
```
SteveL-MSFT: Why not just use $testdrive?

SteveL-MSFT (changes requested):

test/powershell/engine/Basic/Assembly.LoadFrom.ps1
daxian-dbw: Once the assembly is loaded, the file cannot be removed. If I use `$testdrive`, Pester will try to remove it afterward and result in non-failure errors.

SteveL-MSFT: Got it, perhaps add a comment

daxian-dbw: A comment is added.

SteveL-MSFT (approved):
LGTM

iSazonov:
Should we test loading from different default paths? It seems we had an issue with CoreFX.

daxian-dbw:
@iSazonov can you please elaborate? And which CoreFx issue?

iSazonov:
I mean the comment https://github.com/PowerShell/PowerShell/issues/4192#issuecomment-313457476

daxian-dbw:
This test is to cover the corefx issue described in that comment. `Consumer.dll` and `Provider.dll` are in the same folder and the former implicit references to the latter.

PR #4296, opened 2017-07-19
Fix array expression to not return null or throw error

### LINKED ISSUE #4280: @() (array-subexpression operator) should not be optimized away when a cast to an array type is involved
<sup>This is a follow-up to the discussion involving @daxian-dbw and @PetSerAl in #4257</sup>

Context:

* PSv5.1 introduced a generally helpful optimization that optimizes `@()` away in cases where an array is _directly_ constructed inside of it.
   * I presume that the motivation for this optimization was the widespread, but redundant and inefficient practice of enclosing `,`-constructed arrays ("array literals") in `@()`, such as in `@(1, 2)`, where `1, 2` would not only suffice, but, in PSv5-, `@()` would actually (effectively) _clone_ the array.

* **Unfortunately, this optimization is also applied when a _cast_ to an array type (e.g, `[int[]]`) is involved, which can have unpleasant side effects** - see below.  
_Not_ optimizing in this scenario makes for conceptually clearer, predictable behavior.

* While a change would technically be a breaking one, it is hard to imagine anyone having relied on the optimized behavior _with casts_ - a typical [Bucket 3: Unlikely Grey Area](https://github.com/PowerShell/PowerShell/blob/master/docs/dev-process/breaking-change-contract.md#bucket-3-unlikely-grey-area) change.

   * Also note that when the optimization was _introduced_ in PSv5.1, it was technically a breaking change too - though I suspect it wasn't announced as such or even at all.

Steps to reproduce
------------------

```powershell
@([int[]] (1, 2)).GetType().Name

$arr = 1, 2; @([int[]] $arr).GetType().Name
```

Expected behavior
-----------------

```none
Object[]
Object[]
```

Actual behavior
---------------

```none
Int32[]
Int32[]
```

Note that versions up to and including v5 - before the optimization was introduced - behave as expected.

Why is the current behavior problematic?
-----------------------------------------

<sup>The examples in this section are courtesy of @PetSerAl.</sup>

All examples are undoubtedly edge cases, but the larger point is that this behavior should never have been introduced, and, given its "stealth status", this is an opportunity to get rid of it.

* It subverts the fundamental assumption that `@()` always creates regular PowerShell arrays (`System.Object[]`):

```powershell
> @([int[]] (1, 2))[-1] = 'foo'   # BREAKS, because the array is [int[]], not [object[]]
Cannot convert value "foo" to type "System.Int32".  ...
...
```

@PetSerAl points out that this fundamental assumption is even still reflected in other places in the source code: [`ArrayExpressionAst.StaticType`](https://github.com/PowerShell/PowerShell/blob/master@%7B2017-07-18%7D/src/System.Management.Automation/engine/parser/ast.cs#L9490) (a `@(...)` expression is of type `[System.Management.Automation.Language.ArrayExpressionAst]`).

* It subverts the expectation that `@()` always _clones_ an array:

```powershell
> $a = [int[]] (1, 2); $b = @([int[]] $a); [object]::ReferenceEquals($a, $b)
True    # $a and $b unexpectedly reference the same array.
```

Note that the accidental reference equality only happens ~if the source type and the cast type are identical~ if the cast type is either the same as the source type or a type that the source type is (directly or indirectly) derived from (covariance).

Thus, if the _cast_ type is `[object[]]` and the source type is a reference type too, the problem always surfaces.

Environment data
----------------

```powershell
PowerShell Core v6.0.0-beta.4 on macOS 10.12.5
PowerShell Core v6.0.0-beta.4 on Ubuntu 16.04.2 LTS
PowerShell Core v6.0.0-beta.4 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
Windows PowerShell v5.1.15063.413 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
```

daxian-dbw:
Fix #4280

Summary
----------
This PR focuses on 3 issues:
1. According to [PowerShell Language Specification Version 3.0](https://www.microsoft.com/en-us/download/details.aspx?id=36389), as quoted: "_The result is the (possibly empty) unconstrained 1-dimensional array_", `@(...)` should only return `object[]` array.
1. `@([object[]]$null).GetType()` fails with error `"You cannot call a method on a null-valued expression."`
1. `@([System.Collections.Generic.List[object]]$null)` fails with error `"Object reference not set to an instance of an object."`

Marked with label `Review-Committee` to review the first issue.

### DIFF
--- a/src/System.Management.Automation/engine/parser/Compiler.cs
+++ b/src/System.Management.Automation/engine/parser/Compiler.cs
@@ -37,8 +37,6 @@ namespace System.Management.Automation.Language
             typeof(List<object>).GetConstructor(PSTypeExtensions.EmptyTypes);
         internal static readonly MethodInfo ObjectList_ToArray =
             typeof(List<object>).GetMethod(nameof(List<object>.ToArray), PSTypeExtensions.EmptyTypes);
-        internal static readonly MethodInfo ObjectArray_Clone =
-            typeof(object[]).GetMethod(nameof(Array.Clone), PSTypeExtensions.EmptyTypes);
 
         internal static readonly MethodInfo ArrayOps_GetMDArrayValue =
             typeof(ArrayOps).GetMethod(nameof(ArrayOps.GetMDArrayValue), staticFlags);
@@ -5479,6 +5477,8 @@ namespace System.Management.Automation.Language
         public object VisitArrayExpression(ArrayExpressionAst arrayExpressionAst)
         {
             Expression values = null;
+            ExpressionAst pureExprAst = null;
+
             var subExpr = arrayExpressionAst.SubExpression;
             if (subExpr.Traps == null)
             {
@@ -5487,10 +5487,10 @@ namespace System.Management.Automation.Language
                     var pipelineBase = subExpr.Statements[0] as PipelineBaseAst;
                     if (pipelineBase != null)
                     {
-                        var exprAst = pipelineBase.GetPureExpression();
-                        if (exprAst != null)
+                        pureExprAst = pipelineBase.GetPureExpression();
+                        if (pureExprAst != null)
                         {
-                            values = Compile(exprAst);
+                            values = Compile(pureExprAst);
                         }
                     }
                 }
@@ -5502,26 +5502,10 @@ namespace System.Management.Automation.Language
             }
             values = values ?? CaptureAstResults(subExpr, CaptureAstContext.Enumerable);
 
-            if (values.Type == typeof(object[]) || values.Type == typeof(List<object>))
+            if (pureExprAst is ArrayLiteralAst)
             {
-                Expression toArrayExpr = null;
-                if (values.Type == typeof(object[]))
-                {
-                    toArrayExpr = Expression.Call(values, CachedReflectionInfo.ObjectArray_Clone).Cast(typeof(object[]));
-                }
-                else
-                {
-                    toArrayExpr = Expression.Call(values, CachedReflectionInfo.ObjectList_ToArray);
-                }
-
-                var temp = Expression.Variable(typeof(object[]));
-                var expr = Expression.Block(
-                    new[] { temp },
-                    Expression.IfThenElse(
-                        Expression.Equal(values, ExpressionCache.NullConstant),
-                        Expression.Assign(temp, Expression.NewArrayInit(typeof(object), ExpressionCache.NullConstant)),
-                        Expression.Assign(temp, toArrayExpr)),
-                    temp);
+                // If the pure expression is ArrayLiteralAst, just return the result.
+                return values;
             }
             if (values.Type.IsPrimitive || values.Type == typeof(string))
             {
--- a/src/System.Management.Automation/engine/parser/Compiler.cs
+++ b/src/System.Management.Automation/engine/parser/Compiler.cs
@@ -37,6 +37,8 @@ namespace System.Management.Automation.Language
             typeof(List<object>).GetConstructor(PSTypeExtensions.EmptyTypes);
         internal static readonly MethodInfo ObjectList_ToArray =
             typeof(List<object>).GetMethod(nameof(List<object>.ToArray), PSTypeExtensions.EmptyTypes);
+        internal static readonly MethodInfo ObjectArray_Clone =
+            typeof(object[]).GetMethod(nameof(Array.Clone), PSTypeExtensions.EmptyTypes);
 
         internal static readonly MethodInfo ArrayOps_GetMDArrayValue =
             typeof(ArrayOps).GetMethod(nameof(ArrayOps.GetMDArrayValue), staticFlags);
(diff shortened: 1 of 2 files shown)


test/powershell/Language/Scripting/Array.Tests.ps1
```diff
@@ -0,0 +1,54 @@
+Describe "ArrayExpression Tests" -Tags "CI" {
+    It "@([object[]](1,2,3)) should return a 3-element array of object[]" {
+        $result = @([object[]](1,2,3))
+        $result.GetType().FullName | Should Be "System.Object[]"
+        $result.Length | Should Be 3
```
daxian-dbw: What do you mean by `the $result are calculated`? Can you please elaborate a bit and give an example?

test/powershell/Language/Scripting/Array.Tests.ps1
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    It "@([int[]](1,2,3)) should return a 3-element array of object[]" {
+        $result = @([int[]](1,2,3))
```
daxian-dbw: For any array type other than `[object[]]`, it will fall in the code path of [`PSToObjectArrayBinder`](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/parser/Compiler.cs#L5524). I just use `[int[]]` as a simple validation.

test/powershell/Language/Scripting/Array.Tests.ps1
iSazonov: From Get-Content.Tests.ps1:
```powershell
  It "should Get-Content with a variety of -Tail and -ReadCount values" {#[DRT]
    set-content -path $testPath "Hello,World","Hello2,World2","Hello3,World3","Hello4,World4"
    $result=get-content -path $testPath -readcount:-1 -tail 5
    $result.Length | Should Be 4
```

iSazonov: Can we use Should BeOfType?

test/powershell/Language/Scripting/Array.Tests.ps1
iSazonov: Sorry, I mean member types.

daxian-dbw: Oh, the member type doesn't matter for the scenario of this fix. It's OK as long as the conversion works.

test/powershell/Language/Scripting/Array.Tests.ps1
daxian-dbw: To use `Should BeOfType` I need to wrap `$result` in a new array because it will be unraveled otherwise. I feel it's clearer to just use `GetType().FullName` in this case.

As for the bad pattern, I'm still confused. Why do you think putting `$result.Length | Should Be 3` after the type check would be a bad pattern? More technically speaking, the length check should happen after we know `$result` is an array, otherwise we might (actually impossible) end up checking the `Length` property of some random type :)

iSazonov:
```powershell
 function foo {$null}
 $res = foo
 $res.GetType()
You cannot call a method on a null-valued expression.
At line:1 char:1
+ $res.GetType()
+ ~~~~~~~~~~~~~~
    + FullyQualifiedErrorId : InvokeMethodOnNull
```

iSazonov:
Should we add tests for `@($var)` ? It seems absent.

test/powershell/Language/Scripting/Array.Tests.ps1
daxian-dbw: Ah, I see your point. A `Should Not BeNullOrEmpty` check should be done in that case.

src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -5500,16 +5502,28 @@ public object VisitArrayExpression(ArrayExpressionAst arrayExpressionAst)
             }
             values = values ?? CaptureAstResults(subExpr, CaptureAstContext.Enumerable);
 
-            if (values.Type.IsArray)
+            if (values.Type == typeof(object[]) || values.Type == typeof(List<object>))
```
lzybkr: The goal of the optimization was to cover `@(1,2)` - so maybe the original test was too general. This fix ends up adding unnecessary overhead to the optimization by cloning and adding an `if` test.

I think a better fix is to just constrain the optimization more - basically `if (values is ArrayLiteralAst)`.

I don't recall why I added the `List<object>` case.

daxian-dbw: The clone was added to address a comment raised by @PetSerAl about `$a = 1,2,3; @([object[]]$a)` doesn't return a new array. But if we only consider `if (subExpr is ArrayLiteralAst)` then the clone and the null check are both not necessary anymore.

The `List<object>` case seems like an optimization for `CaptureAstResults`, because a `List<object>` instance will be created to hold the object written to pipe in `CaptureAstResults`.
Are you suggesting to not cover the cases of `values.Type == typeof(object[])` and `values.Type == typeof(List<object>)` here?

daxian-dbw: Replace the cases of `values.Type.IsArray` and `values.Type == typeof(List<object>)` with `if (pureExprAst is ArrayLiteralAst)`. So the former cases will be handled by the dynamic site.

lzybkr (approved):
I think the change is fine.

I should add that the intent of `@()` is not to create a new array, but to ensure the result is an array, and not necessarily an `object[]`.

In other words, I think we could be open to changing what the language specification says to allow for better performance.

daxian-dbw:
> the intent of @() is not to create a new array, but to ensure the result is an array

The last commit removed the `[object]::ReferenceEquals` check from the test per Jason's comment.
But the check for `System.Object[]` is kept as we haven't changed the language specification yet.

mklement0:
Just for future consideration, I'd like to address @lzybkr's comments on the linked PR:

> I should add that the intent of @() is not to create a new array, but to ensure the result is an array, and not necessarily an object[].

While I agree re the overall intent of `@()`, I feel that it is better to not contravene two likely user expectations (which happen to be consistent with what the 3.0 spec says):

* that the array returned is unconstrained, i.e., of type `[object[]]`, i.e., a _regular_ PS array, as most cmdlets return. (Most PowerShell users probably don't even think about array-element typing.)

* that  `@()` always returns a _new_ array, without having to worry about reference equality (which is probably not even on the radar of most PowerShell users).

> In other words, I think we could be open to changing what the language specification says to allow for better performance.

Passing strongly typed arrays through strikes me as a rare use case that I don't think warrants optimization, given that it would contravene the expectations above.

By contrast, optimizing the _typical_ use case is well worth considering: passing the `[object[]]` output of a _single command_ through (optimizing the `@()` away if the enclosed single command happens to return a `[object[]]` array already).

I can't speak to the technical feasibility, but here's an example where the `@()` could be optimized away (without contravening the assumptions above), which makes a noticeable difference with large output sets:

```powershell
# Could be optimized to: Write-Output (1..10000000)
# Currently, the output is collected and a new array is constructed.
@(Write-Output (1..10000000))
```

PR #4301, opened 2017-07-19
Ignore case when binding PSReadline keyhandler functions

### LINKED ISSUE #4300: BUG: PSReadline Set-PSReadlineKeyHandler is case sensitive for Function names
Steps to reproduce
------------------

```powershell
set-psreadlinekeyhandler "ctrl+k" -Function yank # note: lower case "yank"
```

Expected behavior
-----------------

```none
Should set ctrl+k to function Yank
```

Actual behavior
---------------

```none
Causes null reference exception due to reflection using a case sensitive lookup for the method (default behaviour) - needs binding flags applied to call.
```

Environment data
----------------

```powershell
> $PSVersionTable
n/a -- issue is with PSReadline 1.2
```

oising:
set-psreadlinekeyhandler <binding> -function <builtin> was CS due to missing bindingflags for reflection  lookup. It's now CI.


oising:
fixes https://github.com/PowerShell/PowerShell/issues/4300

test/powershell/Modules/PSReadLine/PSReadLine.Tests.ps1
```diff
@@ -42,6 +42,12 @@ Describe "PSReadLine" -tags "CI" {
         (Get-PSReadLineKeyHandler | where { $_.Key -ceq "Alt+B" }).Function | Should Be SelectBackwardWord
     }
 
+    It "Should ignore case when using Function binding" {
+        Set-PSReadlineOption -EditMode Windows
+        Set-PSReadlineKeyHandler "Ctrl+k" -Function yank # NOTE: lower case function name
+        (Get-PSReadlineKeyHandler | Where { $_.Key -ceq "Ctrl+k"}).Function | Should Be Yank
```
iSazonov: Please use quotation for string constant - "Yank".
The test is case-sensitive so we should use `BeExactly` and make the comment clearer:
```powershell
# Attention! The test is case-sensitive! Don't change the cases!
$custom_name = "yank" ; $real_name = "Yank"

Set-PSReadlineOption -EditMode Windows
Set-PSReadlineKeyHandler "Ctrl+k" -Function $custom_name
(Get-PSReadlineKeyHandler | Where { $_.Key -ceq "Ctrl+k"}).Function | Should BeExactly $real_name
```

iSazonov (commented):
@oising Thanks for your contribution!

test/powershell/Modules/PSReadLine/PSReadLine.Tests.ps1
oising: Ok, so do you want me to change all the tests because I am only following their formatting. None of them uses quotes nor BeExactly.

iSazonov: No, in the PR change only your code. If you want to fix other test please open new PR.

oising: Gotcha.

oising: FYI, the BeExactly part is not neccessary. As long as Function is not empty, the test has passed.

test/powershell/Modules/PSReadLine/PSReadLine.Tests.ps1
```diff
@@ -44,8 +44,10 @@ Describe "PSReadLine" -tags "CI" {
 
     It "Should ignore case when using Function binding" {
         Set-PSReadlineOption -EditMode Windows
-        Set-PSReadlineKeyHandler "Ctrl+k" -Function yank # NOTE: lower case function name
-        (Get-PSReadlineKeyHandler | Where { $_.Key -ceq "Ctrl+k"}).Function | Should Be Yank
+        $lowerCaseFunctionName = "yank";
+        $pascalCaseFunctionName = "Yank"
+        Set-PSReadlineKeyHandler "Ctrl+k" -Function $lowerCaseFunctionName
+        (Get-PSReadlineKeyHandler | Where { $_.Key -ceq "Ctrl+k"}).Function | Should Be $pascalCaseFunctionName
```
oising: Looking at this now, having a variable named `$pascalCaseFunctionName` seems to be implying some fragility here when there is none. I think this really should be a "Yank" string literal. I looked at the code and whatever was used for the binding is assigned to the Function property, so in this case, the output is actually the same as the input, so the comparison should be against `$lowerCaseFunctionName` lol... I'll push another change. Sigh.

test/powershell/Modules/PSReadLine/PSReadLine.Tests.ps1
```diff
@@ -44,8 +44,10 @@ Describe "PSReadLine" -tags "CI" {
 
     It "Should ignore case when using Function binding" {
         Set-PSReadlineOption -EditMode Windows
-        Set-PSReadlineKeyHandler "Ctrl+k" -Function yank # NOTE: lower case function name
-        (Get-PSReadlineKeyHandler | Where { $_.Key -ceq "Ctrl+k"}).Function | Should Be Yank
+        $lowerCaseFunctionName = "yank";
```
iSazonov: Please remove ";" and add a comment with caution don't change case in next two lines.

test/powershell/Modules/PSReadLine/PSReadLine.Tests.ps1
iSazonov: I still believe we should use `BeExactly`. What do we expect `Function` return? Exactly "Yank". If we make `$pascalCaseFunctionName = "yank"` the test will be passed but it must fail. In other words, here we test a function "ConvertToPascalCase()".

oising: But it does NOT return "Yank" -- it returns whatever was used (in this case, "yank") to create the binding. The fix in this PR is in the code path that looks up the static method Yank. You're now asking me to make other unrelated changes, which would require a separate PR.

oising: OK, perhaps this is an opportunity to clean up other parts of the code. How does this sound:
- instead of using the input (e.g. "yank", "yAnk", "YANK"), we should use methodinfo.name
- briefdescription/longdescription seems not be copied to the relevant metadata fields on the binding when using set-psreadlinekeyhandler with a built-in function binding, I should ensure theses are propagated
- write tests for the above

I was just targeting the bug before. Screw it, let me address all the issues and amend the original issue #4300 to mention the metadata issues. Thoughts?

iSazonov: @oising Thanks for clarify! I skipped this. If we are already using a reflection I believe `BriefDescription = methodinfo.name` is good.

test/powershell/Modules/PSReadLine/PSReadLine.Tests.ps1
```diff
@@ -42,6 +42,13 @@ Describe "PSReadLine" -tags "CI" {
         (Get-PSReadLineKeyHandler | where { $_.Key -ceq "Alt+B" }).Function | Should Be SelectBackwardWord
     }
 
+    It "Should ignore case when using Function binding" {
+        Set-PSReadlineOption -EditMode Windows
+        $lowerCaseFunctionName = "yank";
+        Set-PSReadlineKeyHandler "Ctrl+k" -Function $lowerCaseFunctionName
```
lzybkr: Can you use a different key that is unbound, preferably one that most keyboards don't even have? That way if someone ran this test from their interactive sessions, it won't mess up the editing experience.

oising:
Certainly, Jason

Sent from my Windows 10 phone

test/powershell/Modules/PSReadLine/PSReadLine.Tests.ps1
oising: I guess setting the EditMode interferes in the same way? I should probably revert after.

lzybkr: Right, don't change the edit mode at all, it's not needed for this test.

iSazonov: Maybe a safest way is to save-restore the handler.

lzybkr: Save/restore is not always possible - if a key was bound to a script block, there is currently no way to get the script block.

I think it's reasonable to use some key like `F24`.

oising:
urgh, why is the merge commit appearing in here? Dammit, do I need to rebase and squash?

oising:
@lzybkr Looks like the binding `F24`  is not recognized under *nix -- the Travis CI build is failing the corresponding test to read it back. Windows is fine... sigh.

lzybkr:
The logic for converting string to key binding is different on *nix, so my guess is that code isn't working correctly. You wrote the Win32 version, so you know where to look. :)

And don't worry about the merge commit - we'll squash when merging anyway.

oising:
@lzybkr I was just coming to that realization when you posted... I was looking through the chord/key parsing code and thinking "hmmm, this seems familiar...oh ****"

src/Microsoft.PowerShell.PSReadLine/ConsoleKeyChordConverter.cs
```diff
@@ -79,8 +79,9 @@ private static ConsoleKeyInfo ConvertOneSequence(string sequence)
                 // key should be first token to be popped
                 if (key == 0)
                 {
-                    // the keyChar is this token
-                    keyChar = token[0];
+                    // the keyChar is this token, if not a special key like "Fxx" etc.
+                    if (token.Length == 1)
+                        keyChar = token[0];
```
iSazonov: Please use the pattern:
```c#
if (...)
{
    keyChar = token[0];
}
```

oising: Done. I have no idea why I omitted the braces. I never do that. I blame it on the hangover.

iSazonov: Weekend :-)

lzybkr (approved):

iSazonov:
@oising Thanks for your contribution!

oising:
@iSazonov It was an itch I scratched. There'll be more, I hope :)

PR #4302, opened 2017-07-19
 Due to the use of unsupported APIs, we must remove the LocalAccounts module until a better solution is found.

TravisEz13:
Due to the use of unsupported APIs, we must remove the LocalAccounts module until a better solution is found.


daxian-dbw (approved):

jeffbi:
I'd been wondering about that.

bahuwang:
Is there a better solution now？

ThomasNieto:
The Windows built-in local accounts module works with Powershell. Here is the module compability reference pages: https://aka.ms/PSModuleCompat

PR #4303, opened 2017-07-19
Due to the use of unsupported APIs, we must remove the Counter CmdLets in the Diagnostics Module until a better solution is found.

TravisEz13:
Due to the use of unsupported APIs, we must remove the Counter CmdLets in the Diagnostics Module until a better solution is found.


daxian-dbw (approved):

jcotton42:
Will these along with the LocalAccounts cmdlets (#4302) be added back before PowerShell 6 hits release?

iSazonov:
@jcotton42 Currently we have no solution but PG actively works to add this back.

TravisEz13:
@jcotton42 Is there some scenario where the Inbox (or WMF) module does not work?

PR #4313, opened 2017-07-20
Fixed passing common test modules path to unelevated Powershell

adityapatwardhan:
- Fixed the way common test modules are passed to elevated and unelevated powershell. Earlier, only elevated powershell got those through inheritance as a child process. Now we add them to the startup of the process.

- Fixed error reported by PSScriptAnalyzer about ? / Where-Object

- Converted all the parameters passed to powershell.exe to be a base64 encoded string to avoid complications with quotes.

- Removed code which was updated $env:PSModulePath as we do it in startup args for powershell process instead. 

- Added a way to disable -Quiet for Pester.

### DIFF
--- a/test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
+++ b/test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
@@ -187,7 +187,7 @@ try
 
     if($SuppressQuiet)
     {
-        $openCoverParams.Add('$SuppressQuiet', $true)
+        $openCoverParams.Add('SuppressQuiet', $true)
     }
 
     $openCoverParams | Out-String | Write-LogPassThru
--- a/test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
+++ b/test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
@@ -2,7 +2,7 @@
     [Parameter(Mandatory = $true, Position = 0)] $coverallsToken,
     [Parameter(Mandatory = $true, Position = 1)] $codecovToken,
     [Parameter(Position = 2)] $azureLogDrive = "L:\",
-    [Parameter] [switch] $SuppressQuiet
+    [switch] $SuppressQuiet
 )
 
 # Read the XML and create a dictionary for FileUID -> file full path.
--- a/test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
+++ b/test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
@@ -169,7 +169,7 @@ try
     Install-OpenCover -TargetDirectory $openCoverTargetDirectory -force
     Write-LogPassThru -Message "OpenCover installed."
 
-    Write-LogPassThru -Message "TestDirectory : $testPath"
+    Write-LogPassThru -Message "TestPath : $testPath"
     Write-LogPassThru -Message "openCoverPath : $openCoverTargetDirectory\OpenCover"
     Write-LogPassThru -Message "psbinpath : $psBinPath"
     Write-LogPassThru -Message "elevatedLog : $elevatedLogs"
@@ -177,7 +177,7 @@ try
     Write-LogPassThru -Message "TestToolsPath : $testToolsPath"
 
     $openCoverParams = @{outputlog = $outputLog;
-        TestDirectory = $testPath;
+        TestPath = $testPath;
         OpenCoverPath = "$openCoverTargetDirectory\OpenCover";
         PowerShellExeDirectory = "$psBinPath";
         PesterLogElevated = $elevatedLogs;
--- a/test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
+++ b/test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
@@ -1,7 +1,8 @@
 ﻿param(
     [Parameter(Mandatory = $true, Position = 0)] $coverallsToken,
     [Parameter(Mandatory = $true, Position = 1)] $codecovToken,
-    [Parameter(Position = 2)] $azureLogDrive = "L:\"
+    [Parameter(Position = 2)] $azureLogDrive = "L:\",
+    [Parameter] [switch] $Quiet = $true
 )
 
 # Read the XML and create a dictionary for FileUID -> file full path.
@@ -182,6 +183,11 @@ try
         PesterLogElevated = $elevatedLogs;
         PesterLogUnelevated = $unelevatedLogs;
         TestToolsModulesPath = "$testToolsPath\Modules";
+    }
+
+    if($Quiet)
+    {
+        $openCoverParams.Add('Quiet', $true)
     }
 
     $openCoverParams | Out-String | Write-LogPassThru
--- a/test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
+++ b/test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
@@ -2,7 +2,7 @@
     [Parameter(Mandatory = $true, Position = 0)] $coverallsToken,
     [Parameter(Mandatory = $true, Position = 1)] $codecovToken,
     [Parameter(Position = 2)] $azureLogDrive = "L:\",
-    [Parameter] [switch] $Quiet = $true
+    [Parameter] [switch] $SuppressQuiet
 )
 
 # Read the XML and create a dictionary for FileUID -> file full path.
@@ -185,9 +185,9 @@ try
         TestToolsModulesPath = "$testToolsPath\Modules";
     }
 
-    if($Quiet)
+    if($SuppressQuiet)
     {
-        $openCoverParams.Add('Quiet', $true)
+        $openCoverParams.Add('$SuppressQuiet', $true)
     }
 
     $openCoverParams | Out-String | Write-LogPassThru
--- a/test/tools/OpenCover/OpenCover.psm1
+++ b/test/tools/OpenCover/OpenCover.psm1
@@ -501,8 +501,6 @@ function Invoke-OpenCover
             # invoke OpenCover unelevated and poll for completion
             "$openCoverBin $cmdlineUnelevated" | Out-File -FilePath "$env:temp\unelevated.ps1" -Force
             runas.exe /trustlevel:0x20000 "powershell.exe -file $env:temp\unelevated.ps1"
-            # wait for process to start
-            Start-Sleep -Seconds 5
             # poll for process exit every 60 seconds
             # timeout of 6 hours
             # Runs currently take about 2.5 - 3 hours, we picked 6 hours to be substantially larger.
@@ -543,7 +541,18 @@ function CreateOpenCoverCmdline($target, $outputLog, $targetArgs)
     $base64targetArgs = [convert]::ToBase64String($bytes)
 
     # the order seems to be important. Always keep -targetargs as the last parameter.
-    $cmdline = "-target:$target","-register:user","-output:${outputLog}","-nodefaultfilters","-oldstyle","-hideskipped:all","-mergeoutput", "-filter:`"+[*]* -[Microsoft.PowerShell.PSReadLine]*`"", "-targetargs:`"-EncodedCommand $base64targetArgs`""
-    return $cmdline
+    $cmdline = "-target:$target",
+        "-register:user",
+        "-output:${outputLog}",
+        "-nodefaultfilters",
+        "-oldstyle",
+        "-hideskipped:all",
+        "-mergeoutput",
+        "-filter:`"+[*]* -[Microsoft.PowerShell.PSReadLine]*`"",
+        "-targetargs:`"-EncodedCommand $base64targetArgs`""
+
+    $cmdlineAsString = $cmdline -join " "
+
+    return $cmdlineAsString
 
 }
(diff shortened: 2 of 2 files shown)


test/tools/OpenCover/OpenCover.psm1
```diff
@@ -479,35 +482,33 @@ function Invoke-OpenCover
         $targetArgsUnelevated = $targetArgs + @("-excludeTag @('RequireAdminOnWindows')")
     }
 
-    $targetArgsElevated += @("-OutputFile $PesterLogElevated", "-OutputFormat $PesterLogFormat", "-Quiet")
-    $targetArgsUnelevated += @("-OutputFile $PesterLogUnelevated", "-OutputFormat $PesterLogFormat", "-Quiet")
+    $targetArgsElevated += @("-OutputFile $PesterLogElevated", "-OutputFormat $PesterLogFormat")
+    $targetArgsUnelevated += @("-OutputFile $PesterLogUnelevated", "-OutputFormat $PesterLogFormat")
```
SteveL-MSFT: Seems like there's lots of redundancy building the args, would it be better to have $targetArgs and then appending the unique parts of Elevated vs Unelevated closer to execution?

test/tools/OpenCover/OpenCover.psm1
```diff
@@ -479,35 +482,33 @@ function Invoke-OpenCover
...
+
     # the order seems to be important. Always keep -targetargs as the last parameter.
-    $openCoverArgsElevated = "-target:$target","-register:user","-output:${outputLog}","-nodefaultfilters","-oldstyle","-hideskipped:all","-mergeoutput", "-filter:`"+[*]* -[Microsoft.PowerShell.PSReadLine]*`"", "-targetargs:`"$targetArgStringElevated`""
-    $openCoverArgsUnelevated = "-target:$target","-register:user","-output:${outputLog}","-nodefaultfilters","-oldstyle","-hideskipped:all", "-mergeoutput", "-filter:`"+[*]* -[Microsoft.PowerShell.PSReadLine]*`"", "-targetargs:`"$targetArgStringUnelevated`""
+    $openCoverArgsElevated = "-target:$target","-register:user","-output:${outputLog}","-nodefaultfilters","-oldstyle","-hideskipped:all","-mergeoutput", "-filter:`"+[*]* -[Microsoft.PowerShell.PSReadLine]*`"", "-targetargs:`"-EncodedCommand $base64targetArgsElevated`""
+    $openCoverArgsUnelevated = "-target:$target","-register:user","-output:${outputLog}","-nodefaultfilters","-oldstyle","-hideskipped:all", "-mergeoutput", "-filter:`"+[*]* -[Microsoft.PowerShell.PSReadLine]*`"", "-targetargs:`"-EncodedCommand $base64targetArgsUnelevated`""
```
SteveL-MSFT: Seems like there's lots of redundancy building the args, would it be better to have $targetArgs and then appending the unique parts of Elevated vs Unelevated closer to execution?

test/tools/OpenCover/OpenCover.psm1
```diff
@@ -519,6 +520,9 @@ function Invoke-OpenCover
             # poll for process exit every 60 seconds
             # timeout of 6 hours
             $timeOut = ([datetime]::Now).AddHours(6)
```
SteveL-MSFT: How long does it typically take?  Not clear to me if 6 hours is a good value

test/tools/OpenCover/OpenCover.psm1
```diff
@@ -465,8 +466,10 @@ function Invoke-OpenCover
 
     # create the arguments for OpenCover
 
-    $startupArgs =  'Set-ExecutionPolicy Bypass -Force -Scope Process;'
-    $targetArgs = "-c","${startupArgs}", "Invoke-Pester","${TestDirectory}"
+    $updatedEnvPath = "${PowerShellExeDirectory}\Modules;$TestToolsModulesPath"
+
+    $startupArgs =  "Set-ExecutionPolicy Bypass -Force -Scope Process; `$env:PSModulePath = '${updatedEnvPath}';"
+    $targetArgs = "${startupArgs}", "Invoke-Pester","${TestDirectory}"
```
SteveL-MSFT: Should we depend on Start-PSPester instead as we've started building more specialization in our test runs in that cmdlet?

SteveL-MSFT (changes requested):

test/tools/OpenCover/OpenCover.psm1
adityapatwardhan: Start-PSPester launches powershell.exe, we need to launch opencover.console.exe which would launch powershell.exe to collect coverage numbers. Adding this logic to build.psm1 does not seem right and would complicate the implementation on Start-PSPester further.

test/tools/OpenCover/OpenCover.psm1
adityapatwardhan: Yes, having a look if this can be simplified.

test/tools/OpenCover/OpenCover.psm1
adityapatwardhan: Yes, having a look if this can be simplified.

test/tools/OpenCover/OpenCover.psm1
adityapatwardhan: It typically takes around 2.5 - 3 hours total, out of which maybe 2 hours for the unelevated one. '6 hours' was completely pulled out of thin air, to be substantially large enough.

test/tools/OpenCover/OpenCover.psm1
SteveL-MSFT: For example, start-pspester builds some test exes used by some of the tests.  Perhaps not in this PR, but eventually we should probably consider having a separate module for test execution (perhaps merge Start-PSPester and OpenCover module?)

test/tools/OpenCover/OpenCover.psm1
SteveL-MSFT: I think it would be worth putting that data in the comment to give context to 6 hours

test/tools/OpenCover/OpenCover.psm1
adityapatwardhan: Yes, I like the idea of have a separate module for test execution. It would have 2 modes, with code coverage and regular.

test/tools/OpenCover/OpenCover.psm1
adityapatwardhan: Removed some redundancy by extracting out a function.

test/tools/OpenCover/OpenCover.psm1
adityapatwardhan: Fixed.

adityapatwardhan:
@SteveL-MSFT Updated. Please have a look.

test/tools/OpenCover/OpenCover.psm1
SteveL-MSFT: If `-OutputFormat` is the same, why not have it in $targetArgs instead?

test/tools/OpenCover/OpenCover.psm1
SteveL-MSFT: I think the only difference here is `$base64targetArgs`.  You could have `$openCoverArgs` with all the common args, then on line 513, something like:

```powershell
& $OpenCoverBin ($openCoverArgsElevated + "-targetargs:`"-EncodedCommand $base64targetArgsElevated`"")
```

SteveL-MSFT (changes requested):

adityapatwardhan:
@SteveL-MSFT Had forgotten to push. Please have a look.

adityapatwardhan:
@JamesWTruher  Can you have a look too?

test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
```diff
@@ -168,22 +169,27 @@ try
...
+    Write-LogPassThru -Message "TestPath : $testPath"
     Write-LogPassThru -Message "openCoverPath : $openCoverTargetDirectory\OpenCover"
     Write-LogPassThru -Message "psbinpath : $psBinPath"
     Write-LogPassThru -Message "elevatedLog : $elevatedLogs"
     Write-LogPassThru -Message "unelevatedLog : $unelevatedLogs"
     Write-LogPassThru -Message "TestToolsPath : $testToolsPath"
 
     $openCoverParams = @{outputlog = $outputLog;
-        TestDirectory = $testPath;
+        TestPath = $testPath;
```
JamesWTruher: since these hashtable entries are on separate lines, the `;` is not really needed. I don't think it's a big deal, but something to watch in future

test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
```diff
@@ -168,22 +169,27 @@ try
...
         OpenCoverPath = "$openCoverTargetDirectory\OpenCover";
         PowerShellExeDirectory = "$psBinPath";
         PesterLogElevated = $elevatedLogs;
         PesterLogUnelevated = $unelevatedLogs;
         TestToolsModulesPath = "$testToolsPath\Modules";
     }
 
+    if($SuppressQuiet)
+    {
+        $openCoverParams.Add('$SuppressQuiet', $true)
```
JamesWTruher: should `$SuppressQuiet` be `SuppressQuiet`?

test/tools/OpenCover/OpenCover.psm1
```diff
@@ -524,15 +518,32 @@ function Invoke-OpenCover
...
+
+function CreateOpenCoverCmdline($target, $outputLog, $targetArgs)
+{
+    $targetArgString = $targetArgs -join " "
+
+    $bytes = [System.Text.Encoding]::Unicode.GetBytes($targetArgString)
+    $base64targetArgs = [convert]::ToBase64String($bytes)
+
+    # the order seems to be important. Always keep -targetargs as the last parameter.
+    $cmdline = "-target:$target","-register:user","-output:${outputLog}","-nodefaultfilters","-oldstyle","-hideskipped:all","-mergeoutput", "-filter:`"+[*]* -[Microsoft.PowerShell.PSReadLine]*`"", "-targetargs:`"-EncodedCommand $base64targetArgs`""
```
JamesWTruher: having these all on one line is a difficult to read, this would work as well:
```PowerShell
$cmdline = "-target:$target",
    "-register:user",
    "-output:${outputLog}",
    "-nodefaultfilters",
    "-oldstyle",
    "-hideskipped:all",
    "-mergeoutput",
    "-filter:`"+[*]* -[Microsoft.PowerShell.PSReadLine]*`"",
    "-targetargs:`"-EncodedCommand $base64targetArgs`""
```

test/tools/OpenCover/OpenCover.psm1
```diff
@@ -476,46 +479,37 @@ function Invoke-OpenCover
...
             # invoke OpenCover elevated
-            & $OpenCoverBin $openCoverArgsElevated
+            & $OpenCoverBin $cmdlineElevated
 
             # invoke OpenCover unelevated and poll for completion
-            "$openCoverBin $openCoverArgsUnelevatedString" | Out-File -FilePath "$env:temp\unelevated.ps1" -Force
+            "$openCoverBin $cmdlineUnelevated" | Out-File -FilePath "$env:temp\unelevated.ps1" -Force
             runas.exe /trustlevel:0x20000 "powershell.exe -file $env:temp\unelevated.ps1"
             # wait for process to start
             Start-Sleep -Seconds 5
```
JamesWTruher: this 5 second sleep is probably not needed since you have the sleep in line 515

JamesWTruher (approved):
except for the comment about `$SuppressQuiet` this is fine

test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
adityapatwardhan: Fixed.

test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
adityapatwardhan: Good catch. Fixed.

test/tools/OpenCover/OpenCover.psm1
adityapatwardhan: Removed the wait.

test/tools/OpenCover/OpenCover.psm1
adityapatwardhan: Fixed.

adityapatwardhan: Also converted it to a single string using -join " " before returning

adityapatwardhan:
@SteveL-MSFT @JamesWTruher I have addressed all feedback, please have a look.

SteveL-MSFT (approved):
LGTM

test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
```diff
@@ -1,7 +1,8 @@
 ﻿param(
     [Parameter(Mandatory = $true, Position = 0)] $coverallsToken,
     [Parameter(Mandatory = $true, Position = 1)] $codecovToken,
-    [Parameter(Position = 2)] $azureLogDrive = "L:\"
+    [Parameter(Position = 2)] $azureLogDrive = "L:\",
+    [Parameter] [switch] $SuppressQuiet
```
daxian-dbw: This will make the parameter type of `SuppressQuiet` be [parameter], for example:
```
function bar {
param(
  [Parameter] [switch] $SuppressQuiet
)
}

PS> gcm bar -Syntax
bar [[-SuppressQuiet] <Parameter>]
```

adityapatwardhan: Fixed.

adityapatwardhan:
@daxian-dbw @SteveL-MSFT  please have another look, I have addressed the feedback and added another bug fix.

SteveL-MSFT (approved):

PR #4304, opened 2017-07-20
Fix PRs not running Feature

TravisEz13:
[Update]Verified it's working with this PR in AppVeyor and Travis-CI (was already working for a push.)

PR #4323, opened 2017-07-22
Fix Implicit Remoting test failure on nightly run

SteveL-MSFT:
[Feature]
Systems running WS2012R2 w/o WMF5 has WinRM with restriction for MaxConcurrentUsers between 1-100

So the default needs to be 100 to allow PSCore6 to work w/o requiring WMF5+

This is technically a `Breaking Change` as we had explicitly changed the allowed max to Int.MaxValue, but for most users this shouldn't be an issue.

### DIFF
--- a/src/System.Management.Automation/engine/remoting/commands/CustomShellCommands.cs
+++ b/src/System.Management.Automation/engine/remoting/commands/CustomShellCommands.cs
@@ -113,8 +113,17 @@ function Register-PSSessionConfiguration
             }}
         }}
 
-        new-item -path WSMan:\localhost\Plugin -file ""$filepath"" -name ""$pluginName""
-        # $? is to make sure the last operation is succeeded
+        try
+        {{
+            new-item -path WSMan:\localhost\Plugin -file ""$filepath"" -name ""$pluginName""
+        }}
+        catch [System.InvalidOperationException] # WS2012/R2 WinRM w/o WMF has limitation where MaxConcurrentUsers can't be greater than 100
+        {{
+            $xml = [xml](get-content ""$filepath"")
+            $xml.PlugInConfiguration.Quotas.MaxConcurrentUsers = 100
+            Set-Content -path ""$filepath"" -Value $xml.OuterXml
+            new-item -path WSMan:\localhost\Plugin -file ""$filepath"" -name ""$pluginName""
+        }}
 
         if ($? -and $runAsUserName)
         {{


SteveL-MSFT:
@TravisEz13 having trouble getting Feature tests to run

iSazonov:
> WS2012R2 w/o WMF5 has WinRM with restriction for MaxConcurrentUsers between 1-100

I never knew that restriction! I tried and I can confirm.
It's also a very fantastic scenario - to a server should connect more than 100 users by WinRS!

I believe the fix does not need because the limit (1) can be changed in future, (2) can vary between systems and versions
If anybody catch the issue he get a standard error message.
So I believe the best fix is to enhance documentation  https://msdn.microsoft.com/en-us/library/ee309367%28v=vs.85%29.aspx

iSazonov:
@SteveL-MSFT
>[Feature]

"If in the first line of the last (most recent) _**commit**_ description you add `[Feature]`,  
we will ensure that we will also run the tests tagged with `Feature`."

TravisEz13:
#4304 was causing [Feature] not to work for PRs

SteveL-MSFT:
@iSazonov the fix is only needed because any customer (and affects our AppVeyor CI) using WS2012/R2 without WMF5 installed will have this restriction.  This means that by default, if they do `new-pssessionconfigurationfile` and `register-pssessionconfigurationfile` it will return this error.  Unfortunately, we don't expose a way to set this value, they would have to understand WinRM XML configuration to make it work.

daxian-dbw:
The 3 failures in implicit remoting will be resolved by #4326

iSazonov:
@SteveL-MSFT Why do we rely on the default value? What if I want 200? I can install WMF5 and get 200, but this fix is blocking me. This error is rised on the server but is sent to the remote user. 
This user does not know the configuration of the remote server, which can also be changed dynamically. If we want to inform the user, we must catch this exception and create our own.

SteveL-MSFT:
@iSazonov we need a default as that value is required by WinRM.  I guess the question is whether we want to support downlevel systems w/o WMF5 or not.  Personally, I don't like making PSCore6 dependent on WMF5.  cc @joeyaiello

iSazonov:
I don't like making PSCore6 dependent on WMF5 too.

SteveL-MSFT:
Based on some internal discussion, I think where we ended up is to revert the default as it was changed due to customer impact (100 was too low) and in the cmdlet, if the registration fails, we try again with MaxConcurrentUsers=100.

src/System.Management.Automation/engine/remoting/commands/CustomShellCommands.cs
```diff
@@ -113,8 +113,17 @@ public sealed class RegisterPSSessionConfigurationCommand : PSSessionConfigurati
             }}
         }}
 
-        new-item -path WSMan:\localhost\Plugin -file ""$filepath"" -name ""$pluginName""
-        # $? is to make sure the last operation is succeeded
+        try
+        {{
+            new-item -path WSMan:\localhost\Plugin -file ""$filepath"" -name ""$pluginName""
+        }}
+        catch [System.InvalidOperationException] # WS2012/R2 WinRM has limitation where MaxConcurrentUsers can't be greater than 100
```
iSazonov: The comment should  mention WFM 5.

SteveL-MSFT: WS2012R2 has PSv3 by default and I'm not sure if the Max limit was changed in WMF4 or WMF5, so I'll just mention WMF

SteveL-MSFT:
@mirichmo @PaulHigin since this is causing our nightly runs to fail, can you prioritize this review higher?  thanks

src/System.Management.Automation/engine/remoting/commands/NewPSSessionConfigurationOptionCommand.cs
```diff
@@ -36,7 +36,7 @@ public class WSManConfigurationOption : PSTransportOption
         private int? _idleTimeoutSec = null;
 
         private const string AttribMaxConcurrentUsers = "MaxConcurrentUsers";
-        internal static readonly int? DefaultMaxConcurrentUsers = int.MaxValue;
+        internal static readonly int? DefaultMaxConcurrentUsers = 100; // WS2012R2 limits this value to be 1-100
```
mirichmo: Given the decision to retry, shouldn't this be reverted?

SteveL-MSFT: Yes, thought I did.  Will fix.

daxian-dbw (approved):

mirichmo (approved):

PaulHigin (approved):
LGTM

PR #4326, opened 2017-07-24
Fix regressions that cause implicit remoting tests to fail

daxian-dbw:
Three implicit remoting tests failed in daily test runs. They reflect regressions introduced in #4261, and this PR is to fix them.

The original test for those 3 are checking `-ne $null | Should Be $true`:
```
## BadVerb-Variable should be a function, not an alias (1)
((Get-Item Function:\BadVerb-Variable -ErrorAction SilentlyContinue) -ne $null) | Should Be $true
...
## BadVerb-Variable should be a function, not an alias (1)
((Get-Item Function:\BadVerb-Variable -ErrorAction SilentlyContinue) -ne $null) | Should Be $true
...
## BadVerb-Variable should be an alias, not a function (2)
((Get-Item Alias:\BadVerb-Variable -ErrorAction SilentlyContinue) -ne $null) | Should Be $true
```
So we should change them to `| Should Not Be $null`.


daxian-dbw:
The failure in AppVeyor will be resolved by #4323

iSazonov (commented):
LGTM.

PR #4333, opened 2017-07-24
move packaging string to psd1 file

TravisEz13:
move here strings in packaging module to psd1 file for easier maintenance of strings and code


lzybkr (commented):
We should only use locale specific resources for strings that might require localization.

TravisEz13:
@lzybkr I moved the psd1 out of a local specific folder, does that address your concern?

lzybkr:
You still use `Import-LocalizedData` - remove that and my concerns are addressed.

tools/packaging/packaging.psm1
```diff
@@ -1,4 +1,23 @@
...
+        [string]$Path
+    )
+
+    $ErrorActionPreference = 'Stop'
+
+    $psd1 = Get-Content -Raw -Path $Path
+    [ScriptBlock]$scriptBlock = [ScriptBlock]::Create($psd1)
+    [string[]] $allowedCommands = 'ConvertFrom-StringData'
+    $scriptBlock.CheckRestrictedLanguage($allowedCommands,$null,$false)
+    return $scriptBlock.InvokeReturnAsIs()
```
JamesWTruher: you could add `[pscustomobject]` here

JamesWTruher (approved):

TravisEz13:
I filed #4337 for a fix to PowerShell, which won't help here as this script needs to run down level as well.
@lzybkr I added `Import-Data` which does almost the same thing as `Import-LocalizedData` but with out the localized file look-up

tools/packaging/packaging.psm1
lzybkr: Why not just use the data statement in a ps1 file and run it?

TravisEz13: I'd like to have a file that is restricted to data. In a similar pattern as `Import-LocalizedData`.

lzybkr: You can use `Import-PowerShellDataFile` - it's more restrictive than it needs to be for pure data, but I believe already covers your use case.

daxian-dbw:
All comments have been addressed, so I'm about to merge this PR.
@TravisEz13 Since `Import-PowerShellDataFile` is the command to import a .psd1 file, I think #4337 probably should be closed. Please re-open the issue if you think otherwise.

Issue #4331, opened 2017-07-24
Empty string in HelpMessage parameter causes stopping error with unhelpful message


ViperTG:
Description
------------------
I found this error today when having problems getting a newly created script to work, and the error message was not helping. It refers to the #requires statement being formatted incorrectly but the script didn't even have the word "requires" anywhere in it. Turned out to be the HelpMessage parameter in one of the parameters that was empty as i had forgot to write something there.

Seems like it is a parsing error of some kind, not expecting the HelpMessage string to be empty.

In the documentation for HelpMessage it does not state that it cannot be empty.

Error was found on Win7 x64 running Windows Powershell 5.1
Tested on Windows 10 (15063) with Windows Powershell 5.1 and Powershell Core 6.0.0-Beta.4

Steps to reproduce
------------------
Take a simple script like this, call it Test-HelpMessage.ps1

[CmdletBinding()]
Param (
    [Parameter(HelpMessage="")]
    [String]$Parameter1
)
Write-Output "Hello"

Run the script .\Test-HelpMessage

Expected behavior
-----------------
Expect to see "Hello" text as output.

Actual behavior
---------------

PS C:\temp> .\Test-HelpMessage.ps1
.\Test-HelpMessage.ps1 : Cannot process the #requires statement because it is not in the correct format.
The #requires statement must be in one of the following formats:
 "#requires -shellid <shellID>"
 "#requires -version <major.minor>"
 "#requires -psedition <edition>"
 "#requires -pssnapin <psSnapInName> [-version <major.minor>]"
 "#requires -modules <ModuleSpecification>"
 "#requires -runasadministrator"
At line:1 char:1
+ .\Test-HelpMessage.ps1
+ ~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (Test-HelpMessage.ps1:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : ScriptRequiresInvalidFormat

Suggestion [3,General]: The command Test-HelpMessage.ps1 was not found, but does exist in the current location. Windows PowerShell does not load commands from the current location by default. If you trust this command, instead type: ".\Test-HelpMessage.ps1". See "get-help about_Command_Precedence" for more details.

Content of $error[0].Exception.InnerException:
Cannot process argument because the value of argument "value" is not valid. Change the value of the "value" argument and run the operation again.

Environment data
----------------
$PSVersionTable

Name                           Value
----                           -----
[...]

PR #4334, opened 2017-07-24
Useful error message when HelpMessage attributes are empty string

SteveL-MSFT:
Fix invalid value error message on some cmdlet attributes to point to the specific attribute.

Removed wrapping of argument exception into not useful requires exception

Before:
```
.\helpmessage.ps1 : Cannot process the #requires statement because it is not in the correct format.
The #requires statement must be in one of the following formats:
 "#requires -shellid <shellID>"
 "#requires -version <major.minor>"
 "#requires -psedition <edition>"
 "#requires -pssnapin <psSnapInName> [-version <major.minor>]"
 "#requires -modules <ModuleSpecification>"
 "#requires -runasadministrator"
At line:1 char:1
+ .\helpmessage.ps1
+ ~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (helpmessage.ps1:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : ScriptRequiresInvalidFormat
```

After:
```
Cannot process argument because the value of argument "HelpMessage" is not valid. Change the value of the
"HelpMessage" argument and run the operation again.
At line:1 char:1
+ .\helpmessage.ps1
+ ~~~~~~~~~~~~~~~~~
    + FullyQualifiedErrorId : Argument
```

Fix https://github.com/PowerShell/PowerShell/issues/4331

### DIFF
--- a/src/System.Management.Automation/engine/Attributes.cs
+++ b/src/System.Management.Automation/engine/Attributes.cs
@@ -686,7 +686,7 @@ namespace System.Management.Automation
             {
                 if (string.IsNullOrEmpty(value))
                 {
-                    throw PSTraceSource.NewArgumentException("value");
+                    throw PSTraceSource.NewArgumentException("HelpMessage");
                 }
                 _helpMessage = value;
             }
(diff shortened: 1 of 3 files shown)


lzybkr (approved):
I sign off assuming the `#requires` error message has not degraded from this change.

SteveL-MSFT:
@lzybkr `ScriptRequiresException` still gets thrown in all other cases

daxian-dbw (approved):

PR #4335, opened 2017-07-24
Separate Windows PSRP binary build out of Start-PSBuild

### LINKED ISSUE #3014: Move Windows Native Components to a Sub-project or Separate Project

daxian-dbw:
Fix #3014

There are following major changes:
1. `Start-PSBootstrap -BuildWindowsNative` installs the native dependencies required for building PSRP binary. Without `-BuildWindowsNative`, it only installs the dotnet-SDK on Windows platform.
2. `Start-PSBuild` doesn't build Windows PSRP binary. Instead, `Start-BuildNativeWindowsBinaries` is added to build it. After the build, 3 files (`'pwrshplugin.dll'`, `'pwrshplugin.pdb'`, `'Install-PowerShellRemoting.ps1'`) will be bin-placed at `src\powershell-win-core` like before.
3. The NuGet package `'psrp.windows'` is added to `powershell-core` feed, and we reference it in `powershell-win-core.csproj` to get the Windows PSRP related files. The NuGet package is [here](https://powershell.myget.org/feed/powershell-core/package/nuget/psrp.windows), and the content of the package is as follows: (.dll and .pdb files are from the beta.4 release build)

image

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -1850,7 +1850,7 @@ function script:logerror([string]$message) {
 function script:precheck([string]$command, [string]$missedMessage) {
     $c = Get-Command $command -ErrorAction SilentlyContinue
     if (-not $c) {
-        if ($null -ne $missedMessage)
+        if (-not [string]::IsNullOrEmpty($missedMessage))
         {
             Write-Warning $missedMessage
         }
--- a/build.psm1
+++ b/build.psm1
@@ -105,14 +105,14 @@ function Start-BuildNativeWindowsBinaries {
 
     # cmake is needed to build powershell.exe
     if (-not (precheck 'cmake' $null)) {
-        throw 'cmake not found. Run "Start-PSBootstrap -Package". You can also install it from https://chocolatey.org/packages/cmake'
+        throw 'cmake not found. Run "Start-PSBootstrap -BuildWindowsNative". You can also install it from https://chocolatey.org/packages/cmake'
     }
 
     Use-MSBuild
 
     # mc.exe is Message Compiler for native resources
     if (-Not (Test-Win10SDK)) {
-        throw 'Win 10 SDK not found. Run "Start-PSBootstrap -Package" or install Microsoft Windows 10 SDK from https://developer.microsoft.com/en-US/windows/downloads/windows-10-sdk'
+        throw 'Win 10 SDK not found. Run "Start-PSBootstrap -BuildWindowsNative" or install Microsoft Windows 10 SDK from https://developer.microsoft.com/en-US/windows/downloads/windows-10-sdk'
     }
 
     $vcPath = (Get-Item(Join-Path -Path "$env:VS140COMNTOOLS" -ChildPath '../../vc')).FullName
@@ -1093,6 +1093,7 @@ function Start-PSBootstrap {
         [string]$Version = $dotnetCLIRequiredVersion,
         [switch]$Package,
         [switch]$NoSudo,
+        [switch]$BuildWindowsNative,
         [switch]$Force
     )
 
@@ -1100,12 +1101,12 @@ function Start-PSBootstrap {
 
     Push-Location $PSScriptRoot/tools
 
-    # This allows sudo install to be optional; needed when running in containers / as root
-    # Note that when it is null, Invoke-Expression (but not &) must be used to interpolate properly
-    $sudo = if (!$NoSudo) { "sudo" }
-
     try {
         if ($Environment.IsLinux -or $Environment.IsOSX) {
+            # This allows sudo install to be optional; needed when running in containers / as root
+            # Note that when it is null, Invoke-Expression (but not &) must be used to interpolate properly
+            $sudo = if (!$NoSudo) { "sudo" }
+
             try {
                 # Update googletest submodule for linux native cmake
                 Push-Location $PSScriptRoot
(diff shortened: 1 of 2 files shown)


build.psm1
```diff
@@ -89,6 +89,95 @@ function Test-Win10SDK {
...
+function Start-BuildNativeWindowsBinaries {
+    param(
+        [ValidateSet('Debug', 'Release')]
+        [string]$Configuration = 'Release',
+
+        [ValidateSet('x64', 'x86')]
+        [string]$Arch = 'x64'
+    )
+
+    if (-not $Environment.IsWindows) { return }
```
SteveL-MSFT: Would it be better to return error?

build.psm1
```diff
@@ -89,6 +89,95 @@ function Test-Win10SDK {
...
+
+    # cmake is needed to build powershell.exe
+    if (-not (precheck 'cmake' $null)) {
+        throw 'cmake not found. Run "Start-PSBootstrap -Package". You can also install it from https://chocolatey.org/packages/cmake'
```
SteveL-MSFT: Why not introduce a new `-BuildNativeWindows` switch instead of reusing `-Package`?  That way I can privately build a package w/o needing the SDK or building pwshplugin locally.

SteveL-MSFT (changes requested):

build.psm1
daxian-dbw: We support building Appx package for Windows platform (`New-AppxPackage`) and it requires Win10 SDK, that's why I keep using `-Package` here. For other package formats, the windows native dependencies are not needed.

build.psm1
daxian-dbw: Yes, I think a warning would be better.

build.psm1
```diff
@@ -1135,88 +1105,89 @@ function Start-PSBootstrap {
     $sudo = if (!$NoSudo) { "sudo" }
 
     try {
-        # Update googletest submodule for linux native cmake
         if ($Environment.IsLinux -or $Environment.IsOSX) {
             try {
+                # Update googletest submodule for linux native cmake
```
TravisEz13: we should add a switch to bootstrap the components needed for native components so that we don't add them in systems where they are not needed.

build.psm1
```diff
@@ -89,6 +89,98 @@ function Test-Win10SDK {
...
+
+    # cmake is needed to build powershell.exe
+    if (-not (precheck 'cmake' $null)) {
+        throw 'cmake not found. Run "Start-PSBootstrap -Package". You can also install it from https://chocolatey.org/packages/cmake'
+    }
+
+    Use-MSBuild
+
+    # mc.exe is Message Compiler for native resources
+    if (-Not (Test-Win10SDK)) {
```
TravisEz13: Do we need this if we are not building native?

build.psm1
```diff
@@ -1237,13 +1208,12 @@ function Start-PSBootstrap {
...
+        else {
             log "dotnet is already installed.  Skipping installation."
         }
 
-        # Install for Windows
-        if ($Environment.IsWindows) {
+        # Install Windows dependencies for building native binaries or creating Appx package
+        if ($Environment.IsWindows -and $Package) {
             $machinePath = [Environment]::GetEnvironmentVariable('Path', 'MACHINE')
             $newMachineEnvironmentPath = $machinePath
```
TravisEz13: for example line 1221 checks for win 10 sdk

TravisEz13: This is the biggest component that can reduce the docker image size.

daxian-dbw: @TravisEz13 and I talked offline. The `-package` parameter is required only for creating Appx package, and before this change `-package` takes effect only on Unix platforms. If we don't care about Appx, we can just run `Start-PSBootstrap`.
We agreed that @TravisEz13 will verify with @SteveL-MSFT about if we can remove Appx package script, and then update the packaging script and `Start-PSBootstrap` function.

build.psm1
SteveL-MSFT: Got it

build.psm1
SteveL-MSFT: @joeyaiello are you ok if we remove Appx support?  Seems like this is something we can bring back later.

build.psm1
daxian-dbw: @TravisEz13 and I talked offline. The `-package` parameter is required only for creating Appx package, and before this change `-package` only takes effect on Unix platforms. If we don't care about Appx, for example in our release build, we can just run `Start-PSBootstrap` for windows.

If we can remove the script about Appx package, then we need to make the following subsequent changes:
1. Add switch parameter `-BuildWindowsNative` to install dependencies for building PSRP on Windows.
2. Install `Wix` when running `Start-PSPackage -Package` on Windows.

@SteveL-MSFT can you please confirm if we can remove `New-AppxPackage`? If so, I will take care of (1) in this PR and @TravisEz13 will take care of (2) in his packaging changes.

build.psm1
daxian-dbw: Yes, just talked to @TravisEz13. Appx is for NanoServer, and we never published one.

TravisEz13: I also I believe it never worked quite right.

TravisEz13 (approved):

daxian-dbw:
I think it should be OK to remove `New-AppxPackage`, so add `-BuildWindowsNative` to `Start-PSBootstrap` to indicate installing dependencies for building PSRP. It's more clear to review from CodeFlow.

TravisEz13 (approved):

daxian-dbw:
@SteveL-MSFT could you please take another look?

SteveL-MSFT (approved):
LGTM

Issue #3353, opened 2017-03-17
Single Thread Apartment required for Enable-WSManCredSSP and Disable-WSManCredSSP


jeffbi:
When given `-Role Client`, these cmdlets attempt to make use of the `IGroupPolicyObject` COM interface. `IGroupPolicyObject` requires that it be called from a Single Thread Apartment. STA is not supported in CoreCLR, so these cmdlets will fail when using the Client role.

This is the underlying cause of issue #2599.

The core folk have an [existing issue](https://github.com/dotnet/corefx/issues/522) for support of STA.

Steps to reproduce
------------------
`Enable-WSManCredSSP -DelegateComputer ServerA.foo.xyz.com -Role Client -Force`
or
`Disable-WSManCredSSP -Role Client`

Expected behavior
-----------------
Perform the operation without error.

Actual behavior
---------------
Throws `System.InvalidCastException` when trying to cast an object to `IGroupPolicyObject`.

Environment data
----------------

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSRemotingProtocolVersion      2.3
GitCommitId                    v6.0.0-alpha.16-49-g97be759cc7bd33abbe220dd87...
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSEdition                      Core
CLRVersion
PSVersion                      6.0.0-alpha
BuildVersion                   3.0.0.0
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

iSazonov:
Related #2969

iSazonov:
CoreFX issue https://github.com/dotnet/corefx/issues/522 milestone is 1.0.0-rtm so I set "Waiting - NetStandard20"

karelz:
@iSazonov what is exactly the expectation for .NET Standard 2.0? Are there any missing APIs? Or just missing capability of the runtime? -- in that case it would be more about .NET Core 2.0 than .NET Standard 2.0.
Either way, I am not sure if we can make that happen in .NET Core 2.0 timeframe -- let's have that discussion in the new CoreFX issue. Is it something blocking PowerShell Core release? Is it something that PowerShell team (or community) is interested in contributing to CoreFX repo?

iSazonov:
@karelz We have some Issues here about this (2 or even more). 
 I believe @daxian-dbw can comment in depth. It seem @daxian-dbw already open such Issue in CoreFX repo.

daxian-dbw:
[...]

Issue #4158, opened 2017-07-01
Add tests for WSMan provider, cmdlets and CredSSP


adityapatwardhan:
Add tests for WSMan provider, cmdlets and CredSSP
Related code coverage report:

https://codecov.io/gh/PowerShell/PowerShell/tree/9e41c647a9407e05d86f4971a6bd25719a7130d0/src/Microsoft.WSMan.Management

Environment data
----------------

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.3
OS                             Microsoft Windows 10.0.16204
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

```

SteveL-MSFT:
This is a big area, so I'm doing this is stages:

- [ ] ConfigProvider
- [X] CredSSP
- [ ] CurrentConfigurations
- [ ] InvokeWSManAction
- [ ] New-WSManSession
- [ ] PingWSMan
- [ ] Set-QuickConfig
- [ ] WSManConnections
- [ ] WSManInstance

SteveL-MSFT:
The remaining files may be covered already by https://github.com/PowerShell/PowerShell/pull/4756, will wait on results before doing more here.

PR #4336, opened 2017-07-24
Enable WSManCredSSP cmdlets

SteveL-MSFT:
Fixed error message using a null resourcemanager
With most recent CoreFx supporting STA, removed special CoreClr code paths
Added test coverage
Removed some dead commented out code that was never used

Part of https://github.com/PowerShell/PowerShell/issues/4158

Fix https://github.com/PowerShell/PowerShell/issues/3353


SteveL-MSFT:
AppVeyor failure waiting on https://github.com/PowerShell/PowerShell/pull/4323

src/Microsoft.WSMan.Management/WsManHelper.cs
```diff
@@ -172,7 +171,7 @@ internal string GetResourceMsgFromResourcetext(string rscname)
         internal string FormatResourceMsgFromResourcetext(string resourceName,
             params object[] args)
         {
-            return FormatResourceMsgFromResourcetextS(this._resourceMgr, resourceName, args);
+            return FormatResourceMsgFromResourcetextS(g_resourceMgr, resourceName, args);
```
mirichmo: Ca you rename this variable to remove the initial "g"?

src/Microsoft.WSMan.Management/CredSSP.cs
```diff
@@ -941,8 +898,6 @@ protected override void BeginProcessing()
                 throw new InvalidOperationException(message);
             }
 #endif
-            //If not running elevated, then throw an "elevation required" error message.
-            WSManHelper.ThrowIfNotAdministrator();
```
mirichmo: Why were these two checks removed?

test/powershell/Modules/Microsoft.WSMan.Management/CredSSP-nonAdmin.Tests.ps1
```diff
@@ -0,0 +1,27 @@
+Describe "CredSSP cmdlet tests non-admin" -Tags 'Feature' {
+
+    $isAdmin = (New-Object Security.Principal.WindowsPrincipal ([Security.Principal.WindowsIdentity]::GetCurrent())).IsInRole([Security.Principal.WindowsBuiltinRole]::Administrator)
```
mirichmo: 1. Does this line work on non-Windows platforms?
2. Code should not be [outside of a Pester block](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/WritingPesterTests.md#free-code-in-a-describe-block). This should be moved within a BeforeAll
3. Check out the guidelines on [Admin for Windows](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/WritingPesterTests.md#admin-privileges-in-tests)
4. Consider using the [pattern for skipping tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/WritingPesterTests.md#skipping-tests-in-bulk) as well.

test/powershell/Modules/Microsoft.WSMan.Management/CredSSP.Tests.ps1
```diff
@@ -0,0 +1,48 @@
+Describe "CredSSP cmdlet tests" -Tags 'Feature','RequireAdminOnWindows' {
```
mirichmo: Why are these tests separated into separate files? We have been grouping tests for each cmdlet into a single file.

test/powershell/Modules/Microsoft.WSMan.Management/CredSSP.Tests.ps1
```diff
@@ -0,0 +1,48 @@
+Describe "CredSSP cmdlet tests" -Tags 'Feature','RequireAdminOnWindows' {
+
+    It "Error returned if invalid parameters: <description>" -Skip:(!$IsWindows) -TestCases @(
```
mirichmo: The same comments apply to this file as well

mirichmo (changes requested):

src/Microsoft.WSMan.Management/CredSSP.cs
SteveL-MSFT: line 890 already does the check

src/Microsoft.WSMan.Management/WsManHelper.cs
SteveL-MSFT: can do

src/Microsoft.WSMan.Management/CredSSP.cs
SteveL-MSFT: check happens twice

test/powershell/Modules/Microsoft.WSMan.Management/CredSSP-nonAdmin.Tests.ps1
SteveL-MSFT: 1. will wrap in $IsWindows
2. will wrap in BeforeAll
3. this is the opposite of requiring admin, this test requires NOT admin
4. two different test cases here where the first is only applicable when NOT admin, the second one works either way

test/powershell/Modules/Microsoft.WSMan.Management/CredSSP.Tests.ps1
SteveL-MSFT: I suppose I can combine them.  I thought to separate them because at least one test case requires non-admin.

SteveL-MSFT: Just tried it combined and `start-pspester` only ran the tests as elevated so that one test gets skipped.  I'll have to have the non-admin tests as separate.

SteveL-MSFT: Actually it appears you have explicitly run `start-pspester` with `-unelevated`.  I'll have to take care of this myself in the script then.

src/Microsoft.WSMan.Management/CredSSP.cs
```diff
@@ -493,8 +461,6 @@ protected override void BeginProcessing()
...
```
SteveL-MSFT: this is redundant from line 486

SteveL-MSFT:
CI failure is due to one of the same failures in the nightly run unrelated to this PR.  I'll fix that first.

test/powershell/Modules/Microsoft.WSMan.Management/CredSSP.Tests.ps1
```diff
@@ -0,0 +1,48 @@
...
+        @{params=@{Role="Server"};description="server"}
+    ) {
+        param ($params)
+        $c = Enable-WSManCredSSP @params -Force
+        $c.CredSSP | Should Be $true
+
+        $c = Get-WSManCredSSP
+        if ($params.Role -eq "Client")
+        {
+            $c[0] | Should Match "The machine is configured to allow delegating fresh credentials to the following target\(s\):wsman/\*"
```
adityapatwardhan: Please check for en-US before comparing.

test/powershell/Modules/Microsoft.WSMan.Management/CredSSP.Tests.ps1
```diff
@@ -0,0 +1,48 @@
...
+        }
+        else
+        {
+            $c[1] | Should Match "This computer is configured to receive credentials from a remote client computer"
```
adityapatwardhan: Same as above

test/powershell/Modules/Microsoft.WSMan.Management/CredSSP.Tests.ps1
```diff
@@ -0,0 +1,48 @@
...
+        @{Role="Client"},
+        @{Role="Server"}
+    ) {
+        param ($role)
+        Disable-WSManCredSSP -Role $role | Should BeNullOrEmpty
+
+        $c = Get-WSManCredSSP
+        if ($role -eq "Client")
+        {
+            $c[0] | Should Match "The machine is not configured to allow delegating fresh credentials."
```
adityapatwardhan: Same as above.

test/powershell/Modules/Microsoft.WSMan.Management/CredSSP.Tests.ps1
```diff
@@ -0,0 +1,48 @@
...
+        }
+        else
+        {
+            $c[1] | Should Match "This computer is not configured to receive credentials from a remote client computer"
```
adityapatwardhan: Same as above

adityapatwardhan (changes requested):

test/powershell/Modules/Microsoft.WSMan.Management/CredSSP.Tests.ps1
SteveL-MSFT: will fix

test/powershell/Modules/Microsoft.WSMan.Management/CredSSP.Tests.ps1
SteveL-MSFT: will fix

test/powershell/Modules/Microsoft.WSMan.Management/CredSSP.Tests.ps1
SteveL-MSFT: will fix

test/powershell/Modules/Microsoft.WSMan.Management/CredSSP.Tests.ps1
SteveL-MSFT: will fix

SteveL-MSFT:
@mirichmo @adityapatwardhan feedback addressed

adityapatwardhan (approved):

SteveL-MSFT:
@mirichmo can you complete your review?

test/powershell/Modules/Microsoft.WSMan.Management/CredSSP.Tests.ps1
```diff
@@ -0,0 +1,105 @@
...
+
+    BeforeAll {
+        $powershell = Join-Path $PSHOME "powershell"
+
+        $originalDefaultParameterValues = $PSDefaultParameterValues.Clone()
+        if ( ! $IsWindows ) {
+            $PSDefaultParameterValues["it:skip"] = $true
+        }
+
+        $notEnglish = $false
```
mirichmo: As a general rule when using this pattern, a test that is skipped in an environment should not execute any code.

Lines 11 through 15 should not be executed if the test is skipped. Putting them in an else {} would solve this problem.

mirichmo (changes requested):

mirichmo (approved):

adityapatwardhan:
@mirichmo Can you rebase, it should fix the test failures.

adityapatwardhan:
Tests are failing on Linux.
image

mirichmo:
All the CredSSP tests are now passing in AppVeyor and Travis.

ltamaster:
Hi All,
I was looking to use the WSManCredSSP functionalities.
has this PR enabled in a release?
I checked the v6.0.0-beta.6 because it mentions that it was added, but it didn't work for me.

screenshot 2017-11-22 17 16 30


Thanks
Luis

SteveL-MSFT:
@ltamaster CredSSP is only supported on Windows

iSazonov:
@SteveL-MSFT It seems PowerShell docs haven't  "Supported platforms: Windows, Linux, MacOS" in cmdlet's descriptions. 
Also we could make the error message more clear: "The term `Cmdlet` is _not supported_ on the platform or is not recognized ..."

SteveL-MSFT:
@iSazonov can you point me to the docs that say it's supported on Linux/macOS?  Don't see that line [here](https://github.com/PowerShell/PowerShell-Docs/blob/717336cea2cc4a3c7f6c100a4330256a27823331/reference/6/Microsoft.WSMan.Management/Enable-WSManCredSSP.md)

I think the error message is technically correct as that cmdlet doesn't exist on non-Windows.  It's a generic error message if a cmdlet is not found, so implying it's `not supported` may cause confusion particularly for typos.

iSazonov:
@SteveL-MSFT I agree that we can leave the generic message "as is" if we enhance our docs.
>can you point me to the docs that say it's supported on Linux/macOS?

It is just my request to enhance our docs - add new section "Supported platforms".

Issue #2867, opened 2016-12-09
Invoke-Webrequest is missing some properties, like .ParsedHtml and .AllElements


ilgrank:
Hello Everyone
PsCore 6.x's Invoke-Webrequest is missing on the most 'loved' Ps 'classic' features, ie the ability to parse web pages and turn them into explorable objects.

Steps to reproduce
------------------
$test=invoke-webrequest -url http://www.github.com

Expected behavior
-----------------
$test.ParsedHtml 
$test.AllElements
(amont others) are missing:

Microsoft Powershell :
![psclassic](https://cloud.githubusercontent.com/assets/4583412/21061793/200fed48-be4e-11e6-88ac-98a06d79be34.png)

Core Poweshell:
![pscore](https://cloud.githubusercontent.com/assets/4583412/21061794/20177a72-be4e-11e6-991b-63e36dd15583.png)

Actual behavior
---------------
these properties are missing

Environment data
----------------

```
CLRVersion
SerializationVersion 1.1.0.1
PSRemotingProtocolVersion 2.3
PSEdition Core
WSManStackVersion 3.0
BuildVersion 3.0.0.0
GitCommitId v6.0.0-alpha.13
PSVersion 6.0.0-alpha
PSCompatibleVersions {1.0, 2.0, 3.0, 4.0...}

```

I have hope to have web parsing capabilities in PSCore!
Thanks everyone!

BrucePay:
@joeyaiello  FYI: Full PowerShell essentially uses Internet Explorer to do the HTML parsing. This is likely to be problematic to replicate on .NET core.

joeyaiello:
@BrucePay I'd still like to investigate after we look at whatever ends up shipping .NET Standard 2.0

jakerobinson:
What's the best alternative if this is low priority? Any low effort workarounds for parsing HTML in core?

SteveL-MSFT:
.Net Std 2.0 doesn't solve this.  There is another issue where we discussed the need to leverage an OSS html parser rather than rely on a web browser being available (https://github.com/PowerShell/PowerShell/issues/2867), doing that work should also resolve this one.

SRGOM:
May I propose using https://github.com/servo/html5ever?  Loosely speaking, Servo is the next generation of Firefox written in Rust. You can writing your bindings against https://github.com/utkarshkukreti/select.rs which is written on top of html5ever. 

[...]

PR #4338, opened 2017-07-25
Use HTML meta charset attribute value, if present, when the Context-Type header does not specify it.

dantraMSFT:
Fix https://github.com/PowerShell/PowerShell/issues/3267

The change enables checking for the charset meta attribute when the Content-Type header does not define the charset. When found, the text content is re-encoded to the target charset.

A new property has been added to BasicHtmlWebResponseObject and HtmlWebResponseObject to support verifying the encoding. Attempting to do so from the returned content is non-deterministic.
See https://github.com/PowerShell/PowerShell-Docs/pull/1543 for example usage.

NOTE: Three tests are pending the resolution for issue https://github.com/PowerShell/PowerShell/issues/2867

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/HtmlWebResponseObject.CoreClr.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/HtmlWebResponseObject.CoreClr.cs
@@ -49,7 +49,40 @@ namespace Microsoft.PowerShell.Commands
 
         #endregion Constructors
 
+       #region Properties
+
+        /// <summary>
+        /// Gets the Encoding that was used to decode the Content
+        /// </summary>
+        /// <value>
+        /// The Encoding used to decode the Content; otherwise, a null reference if the content is not text.
+        /// </value>
+        public Encoding Encoding { get; private set; }
+
+        #endregion Properties
+
         #region Methods
+
+        // NOTE: Currently this code path is not enabled.
+        // See FillRequestStream in WebRequestPSCmdlet.CoreClr.cs and
+        // GetResponseObject in WebResponseObjectFactory.CoreClr.cs for details.
+        private void InitializeContent()
+        {
+            string contentType = ContentHelper.GetContentType(BaseResponse);
+            string content = null;
+            if (ContentHelper.IsText(contentType))
+            {
+                Encoding encoding = null;
+                // fill the Content buffer
+                string characterSet = WebResponseHelper.GetCharacterSet(BaseResponse);
+                this.Content = StreamHelper.DecodeStream(RawContentStream, characterSet, out encoding);
+                this.Encoding = encoding;
+            }
+            else
+            {
+                this.Content = string.Empty;
+            }
+        }
 
         private void InitializeRawContent(HttpResponseMessage baseResponse)
         {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/StreamHelper.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/StreamHelper.cs
@@ -391,20 +391,8 @@ namespace Microsoft.PowerShell.Commands
             }
         }
 
-        internal static string DecodeStream(Stream stream, string characterSet)
+        private static string StreamToString(Stream stream, Encoding encoding)
         {
-            Encoding encoding = ContentHelper.GetEncodingOrDefault(characterSet);
-            return DecodeStream(stream, encoding);
-        }
-
-        internal static string DecodeStream(Stream stream, Encoding encoding)
-        {
-            if (null == encoding)
-            {
-                // just use the default encoding if one wasn't provided
-                encoding = ContentHelper.GetDefaultEncoding();
-            }
-
             StringBuilder result = new StringBuilder(capacity: ChunkSize);
             Decoder decoder = encoding.GetDecoder();
 
@@ -413,9 +401,8 @@ namespace Microsoft.PowerShell.Commands
             {
                 useBufferSize = encoding.GetMaxCharCount(10);
             }
+
             char[] chars = new char[useBufferSize];
-
-
             byte[] bytes = new byte[useBufferSize * 4];
             int bytesRead = 0;
             do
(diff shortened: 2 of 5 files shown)


SteveL-MSFT (dismissed):

TravisEz13 (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -734,6 +760,84 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
     #endregion SkipHeaderVerification Tests
 
+    #region charset encoding tests
+
+    Context  "BasicHtmlWebResponseObject Encoding tests" {
+        It "Verifies Invoke-WebRequest detects charset meta value when the ContentType header does not define it." {
+            $output = '<html><head><meta charset="gb2312"></head></html>'
+            $expectedEncoding = [System.Text.Encoding]::GetEncoding('gb2312')
```
JamesWTruher: I'm worried about platforms where these encodings don't exist (linux/mac), which will cause the tests to fail. Do additional steps need to be taken to register new encodings, or does that happen elsewhere?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/StreamHelper.cs
```diff
@@ -444,12 +431,65 @@ internal static string DecodeStream(Stream stream, Encoding encoding)
...
+                encoding = ContentHelper.GetDefaultEncoding();
+                isDefaultEncoding = true;
+            }
+
+            string content = StreamToString (stream, encoding);
+
+            if (isDefaultEncoding) do
+            {
+                // check for a charset meta element to override the default.
+                string searchString = "<meta charset=\"";
```
SteveL-MSFT: I think a regular expression might be better here

dantraMSFT:
Moving to regex for detecting charset. Will reopen when done.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/HtmlWebResponseObject.CoreClr.cs
```diff
@@ -49,8 +49,41 @@ internal HtmlWebResponseObject(HttpResponseMessage response, Stream contentStrea
...
 
+       #region Properties
+
+        /// <summary>
+        /// Gets the Encoding that was used to decode the Content
+        /// </summary>
+        /// <value>
+        /// The Encoding used to decode the Content; otherwise, a null reference if the content is not text.
+        /// </value>
+        public Encoding Encoding { get; private set; }
```
dantraMSFT: Done

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/StreamHelper.cs
dantraMSFT: Fixed

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: fixed

dantraMSFT:
@TravisEz13 and @JamesWTruher, please take a look again. I've completed the changes to use regex.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/StreamHelper.cs
```diff
@@ -444,12 +432,74 @@ internal static string DecodeStream(Stream stream, Encoding encoding)
...
+                // Use the default encoding if one wasn't provided
+                encoding = ContentHelper.GetDefaultEncoding();
+                isDefaultEncoding = true;
+            }
+
+            string content = StreamToString (stream, encoding);
+            if (isDefaultEncoding) do
+            {
+                // check for a charset attribute on the meta element to override the default.
+                Match match = s_metaexp.Match(content);
```
TravisEz13: In order to cache the regex, you need to use the static regex methods.  See: https://msdn.microsoft.com/en-us/library/system.text.regularexpressions.regex(v=vs.110).aspx#static_vs_instance

dantraMSFT: @TravisEz13 - According to my reading of the page, I'm doing exactly what it says...

"To prevent recompilation, you should instantiate a single Regex object that is accessible to all code that requires it, as shown in the following rewritten example."

Since s_metaexp is static/readonly, there is only one compilation. That appears to be the point of the of the caching; avoiding recompilation.  Am I missing something?

TravisEz13: `In the .NET Framework versions 1.0 and 1.1, all compiled regular expressions, whether they were used in instance or static method calls, were cached. Starting with the .NET Framework 2.0, only regular expressions used in static method calls are cached.`

dantraMSFT: My review of the corefx shows that constructing and reusing an instance (regex.Match) is equivalent to calling the static Match method.  The constructor manages cache lookups/updates. The static Match method constructs a new instance and calls through the instance Match method.
The relevant static method is:
public static Match Match(string input, string pattern, RegexOptions options, TimeSpan matchTimeout)
{
    return new Regex(pattern, options, matchTimeout, true).Match(input);
}

TravisEz13: Sound good, thanks for looking into it.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -805,6 +831,245 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
     #endregion SkipHeaderVerification Tests
 
+    #region charset encoding tests
+
+    Context  "BasicHtmlWebResponseObject Encoding tests" {
+        It "Verifies Invoke-WebRequest detects charset meta value when the ContentType header does not define it." {
```
TravisEz13: Do you need to add test cases for any of the issues that caused the change to regex?

TravisEz13 (approved):

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: I did add test cases covering variations of <meta charset...?> and <meta http-equiv...>, including various newline variations and unquoted attribute value.

dantraMSFT:
@TravisEz13: would you merge this.  Thanks.

chuanjiao10:
Do not rush to close,Please help look, there seems to be a problem, i updated issues 3267.

SteveL-MSFT:
@chuanjiao10 since this is already merged, @dantraMSFT will submit new PR to address your issue

dantraMSFT:
@chuanjiao10, see https://github.com/PowerShell/PowerShell/pull/4692

chuanjiao10:
good 3q

Issue #4148, opened 2017-06-30
Add tests for *-Item and Get-ChildItem cmdlets


adityapatwardhan:
Add tests for *-Item cmdlets. Missing areas are: 

- [x] Literal paths
- [x] Error cases
- [ ] AccessDenied Error
- [ ] SessionStateItem and SessionStateContainer error cases
- [x] Dynamic parameters
- [x] *-Item for alternate streams
- [x] Get-ChildItem for file
- [x] Rename-Item for directory
- [x] Copy-Item over remote session
- [x] *-Item for Registry provider
- [x] *-Item for Function provider
- [ ] Mapped network drive

Environment data
----------------

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.3
OS                             Microsoft Windows 10.0.16204
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

```

jeffbi:
@adityapatwardhan What are you looking for regarding tests for "Get-ChildItem for file"? Just that it will return information about the file specified?

What is the expected behavior if, for example, `Get-ChildItem` is given a specific, non-globbed file name and also the `-Directory` parameter? Should it return information about the file anyway or return nothing because the file is not a directory?

Also, regarding "Rename-Item for directory", do you just mean that the cmdlet can rename a directory, or something more involved than that?

adityapatwardhan:
@jeffbi We are interested covering the missing code paths at: https://codecov.io/gh/PowerShell/PowerShell/src/master/src/System.Management.Automation/namespaces/FileSystemProvider.cs

The list above is a summary of what is missing. So some tests might not look very useful, but there is untested code that exists and hence we need to exercise those code paths.

jeffbi:
Thanks, @adityapatwardhan, I'll look that over.

SteveL-MSFT:
I think the remaining items can be pushed to MQ

TravisEz13:
@kalgiz I recommend looking at the function provider first.  Work with @adityapatwardhan  to figure out what tests are needed.
After that, you could work on the remoting tests.  @PaulHigin  would be a good resource for showing you how to use remoting in a test.

PixleProgrambetaxl:
+1

PR #4342, opened 2017-07-25
Increase code coverage of Get-ChildItem on file system.

jeffbi:
Addresses part of #4148.

Using the `-Name` parameter on `Get-ChildItem` exercises a part of the file system provider code not previously covered by tests.

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Management/FileSystem.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Management/FileSystem.Tests.ps1
@@ -84,6 +84,11 @@ Describe "Basic FileSystem Provider Tests" -Tags "CI" {
             $dirContents.Count | Should Be 2
         }
 
+        It "Verity Get-ChildItem can get the name of a specified item." {
+            $fileName = Get-ChildItem $testFile -Name
+            $fileName | Should BeExactly $fileName
+        }
+
         It "Set-Content to a file" {
             $content =  Set-Content -Value $testContent -Path $testFile -PassThru
             $content | Should BeExactly $testContent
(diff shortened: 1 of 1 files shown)


test/powershell/Modules/Microsoft.PowerShell.Management/FileSystem.Tests.ps1
```diff
@@ -84,6 +84,11 @@ Describe "Basic FileSystem Provider Tests" -Tags "CI" {
             $dirContents.Count | Should Be 2
         }
 
+        It "Verity Get-ChildItem can get the name of a specified item." {
+            $fileName = Get-ChildItem $testFile -Name
+            $fileName | Should BeExactly $fileName
```
dantraMSFT: This compares to itself.
I suggest calling Get-ChildItem twice, once with -Name to get the name string and once without to get the fileinfo. The test would look something like the following:
$filename | Should BeExactly $fileInfo.Name

dantraMSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Management/FileSystem.Tests.ps1
jeffbi: Yikes! Bad typo on my part. That should have been `$filename | Should BeExactly $testFile`. Is making that change sufficient, or would you prefer the dual-call method?

dantraMSFT:
You're better off comparing the returned value against the name property since, I believe, that's the intent.

jeffbi:
Well, the intent was to check that I got the item I asked for, which is why I was comparing against `$testFile`. But I'll go ahead with your recommendation.

dantraMSFT (approved):

dantraMSFT:
Sorry jeffbi; I should have been clearer.  When I said the intent of -Name, I mean the intent is to return the Name property of the FileInfo object.
The current implementation of FileInfo 'does' preserve the file name as it was passed in; however, since it's not documented explicitly, I avoid depending on it.
Additionally, I would expect the test to succeed with other values, such as .\filename.txt, ./filename.txt, and /etc/filename.txt
I hope that clarifies my feedback.

Issue #1265, opened 2016-07-01
Tab completion confusion with common parameters


lzybkr:
## Steps to reproduce

`function foo { param($DebugBuild) }`
1. `Receive-Job -wa<TAB>`
2. `foo -D<TAB>`
3. `foo -Debug<TAB>`
## Expected behavior
1. Cycle through -Wait and -WarningAction
2. Cycle through -Debug and -DebugBuild
3. Cycle through -Debug and -DebugBuild
## Actual behavior
1. Only completes -WarningAction
2. Only completes -DebugBuild
3. No additional completion

Peter-Schneider:
The actual behavior of case 2 should be ok unless you are using the cmdletbinding attribute.

SteveL-MSFT:
@louistio Thanks for the investigation details!

In the your first issue, even if an exact match is found, I would expect it to cycle through other possible matches.  For the second, it should also all possible matches including common parameters.

Issue #3373, opened 2017-03-20
Tab completion don't work for $error


iSazonov:
Steps to reproduce
------------------
```powershell
5/0
$error[0].<tab>
```
Expected behavior
-----------------
Tab completion substitutes property names (CategoryInfo, Exception ...)

Actual behavior
---------------
Tab completion don't work.

Environment data
----------------

All (Windows and Windows Core) PowerShell versions.

SteveL-MSFT:
Workaround is:

```powershell
$global:error[0].<tab>
```

Did some debugging, command completion finds another variable named `Error` in a lower scope which is empty so nothing to tab complete.  Still investigating.

SteveL-MSFT:
On master branch, tab completion appears to work.  @iSazonov I agree we should add some tests.

vexx32:
Does this also resolve #7828, @SteveL-MSFT ?

SteveL-MSFT:
@vexx32 no, that one isn't resolved

PR #4356, opened 2017-07-27
Add more tab completion tests

### LINKED ISSUE #4160: Add tests for TabCompletion for various input types
Add tests for TabCompletion for various input types

Related code coverage report:

https://codecov.io/gh/PowerShell/PowerShell/tree/9e41c647a9407e05d86f4971a6bd25719a7130d0/src/System.Management.Automation/engine/CommandCompletion

Environment data
----------------

```powershell
> $PSVersionTable
OS                             Microsoft Windows 10.0.16204
```

daxian-dbw:
The first round of adding tab completion tests - issue: #4160
---------------------------------------------------
246 tests are added in this iteration.
Summary of addressed areas:

CommandCompletion.cs
- [x] negatives cases/error handling (if some of these can't be hit, consider converting to asserts)
- [x] nested runspace/remote runspace
- [x] getting next result including wrap around
- [x] user overridden tabexpansion
- [x] legacy tabexpansion
- [x] ~~remove PSv2 code~~ `PSv2CompletionCompleter` is for tab completion in a remote session that is from win8+ to win7, so it needs to be kept.
- [x] adding/handling quotes

CompletionAnalysis.cs
- [x] filename after redirection
- [x] DSC resource

CompletionCompleters.cs
- [x] fully qualified cmdlet
- [x] complete module name
- [x] positional parameter

### DIFF
--- a/test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
+++ b/test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
@@ -19,6 +19,11 @@ Describe "TabCompletion" -Tags CI {
         $res.CompletionMatches[0].CompletionText | Should be 'Foreach('
     }
 
+    It "Should complete Magic where" {
+        $res = TabExpansion2 -inputScript '(1..10).wh' -cursorColumn '(1..10).wh'.Length
+        $res.CompletionMatches[0].CompletionText | Should be 'Where('
+    }
+
     It 'Should complete types' {
         $res = TabExpansion2 -inputScript '[pscu' -cursorColumn '[pscu'.Length
         $res.CompletionMatches[0].CompletionText | Should be 'pscustomobject'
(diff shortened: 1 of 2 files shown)


iSazonov:
We have two non-resolved Issues #1265 and #3373 - maybe add pending tests for them for tracking?

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -200,4 +205,331 @@ Describe "TabCompletion" -Tags CI {
...
         $completionText -join ' '| Should Be 'blg csv tsv'
     }
+
+    Context "File name completion" {
+        BeforeAll {
+            $tempDir = Join-Path -Path $TestDrive -ChildPath "baseDir"
+            $oneSubDir = Join-Path -Path $tempDir -ChildPath "oneSubDir"
+            $oneSubDirPrime = Join-Path -Path $tempDir -ChildPath "prime"
+            $twoSubDir = Join-Path -Path $oneSubDir -ChildPath "twoSubDir"
+            $separator = [System.IO.Path]::DirectorySeparatorChar
```
SteveL-MSFT: Shouldn't we also test where it's not the default separator?

SteveL-MSFT:
Did you rerun code coverage?

SteveL-MSFT:
Don't see any negative cases for the unhit error handling code.

daxian-dbw:
@iSazonov for those 2 issues, tests should be added along with the fix.
@SteveL-MSFT I didn't rerun code coverage, but I will add more tests to cover those areas.

SteveL-MSFT:
The tests you have seem fine, but as I'm looking at the code coverage report some other areas missing (I'll summarize it here including prior feedback):

CommandCompletion.cs
- [ ] negatives cases/error handling (if some of these can't be hit, consider converting to asserts)
- [ ] in debugger
- [ ] nested runspace/remote runspace
- [ ] getting next result including wrap around
- [ ] user overridden tabexpansion
- [ ] legacy tabexpansion
- [ ] remove PSv2 code
- [ ] adding/handling quotes
CompletionAnalysis.cs
- [ ] cursor position (you have some of these covered, not sure if all)
- [ ] constrained language mode
- [ ] filename after redirection
- [ ] line continuation
- [ ] DSC resource
CompletionCompleters.cs
- [ ] fully qualified cmdlet
- [ ] remove workflow code
- [ ] complete module name
- [ ] import-dscresource
- [ ] positional parameter

etc...

It might be worth doing separate PRs per source file to keep them reasonable to review

SteveL-MSFT (changes requested):
Suggest breaking up as seperate PRs per source file

daxian-dbw:
@SteveL-MSFT I added more tests in this first iteration, and the covered areas are listed in the PR description. Please take another look when you have time.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -200,4 +205,629 @@ Describe "TabCompletion" -Tags CI {
...
+        It "Input '<inputStr>' should successfully complete" -TestCases $testCases {
+            param ($inputStr, $localExpected)
+
+            try {
+                Push-Location -Path $tempDir
+                $res = TabExpansion2 -inputScript $inputStr -cursorColumn $inputStr.Length
+                $res.CompletionMatches.Count | Should BeGreaterThan 0
+                $res.CompletionMatches[0].CompletionText | Should Be $localExpected
+            } finally {
+                Pop-Location
```
adityapatwardhan: May be we can do Pop-Location in AfterEach so that it is not repeated in every It.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -200,4 +205,629 @@ Describe "TabCompletion" -Tags CI {
...
+            Remove-Item -Path $tempDir -Recurse -Force
+        }
+
+        It "Input '<inputStr>' should successfully complete" -TestCases $testCases {
+            param ($inputStr, $localExpected)
+
+            try {
+                Push-Location -Path $tempDir
+                $res = TabExpansion2 -inputScript $inputStr -cursorColumn $inputStr.Length
+                $res.CompletionMatches.Count | Should BeGreaterThan 0
```
adityapatwardhan: I believe this is unnecessary as we are checking for CompletionText in the next line. 
```Powershell
$res.CompletionMatches[0].completiontext 
```
will be null if it does not exist and the next should would still fail when tab completion does not work.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -200,4 +205,629 @@ Describe "TabCompletion" -Tags CI {
...
+        AfterAll {
+            Remove-Item -Path $tempDir -Recurse -Force
+        }
+
+        It "Input '<inputStr>' should successfully complete" -TestCases $testCases {
+            param ($inputStr, $localExpected)
+
+            try {
+                Push-Location -Path $tempDir
+                $res = TabExpansion2 -inputScript $inputStr -cursorColumn $inputStr.Length
```
adityapatwardhan: Add ```ErrorAction -Stop ```

adityapatwardhan (changes requested):

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
daxian-dbw: In case `CompletionMatches` is null, the indexing operation will result in an exception:
```
PS:134> $null[0]
Cannot index into a null array.
At line:1 char:1
+ $null[0]
+ ~~~~~~~~
    + FullyQualifiedErrorId : NullArray
```

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
daxian-dbw: Could you please help me understand why `-ErrorAction stop` is needed here?

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
daxian-dbw: Fixed.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
adityapatwardhan:
```Powershell
PS D:\> $inputStr = "d:\doesnoexis"
PS D:\> $res = TabExpansion2 -inputScript $inputStr -cursorColumn $inputStr.Length
PS D:\> $res.CompletionMatches[0].CompletionText
PS D:\> $res.CompletionMatches[0].CompletionText -eq $null
True
```

daxian-dbw: This is assuming that `$res.CompletionMatches` will never be null.
In case it's null (which should not happen), `$res.CompletionMatches[0]` will throw the `index into null array` error.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
adityapatwardhan: Sorry, not needed here.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
SteveL-MSFT: @daxian-dbw don't see which test cases address this?

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -200,4 +205,614 @@ Describe "TabCompletion" -Tags CI {
...
+        AfterEach {
+            Pop-Location
+        }
+
+        It "Input '<inputStr>' should successfully complete" -TestCases $testCases {
+            param ($inputStr, $localExpected)
+
+            Push-Location -Path $tempDir
+            $res = TabExpansion2 -inputScript $inputStr -cursorColumn $inputStr.Length
+            $res.CompletionMatches.Count | Should BeGreaterThan 0
```
SteveL-MSFT: Since you're controlling the directory structure, should this be an exact match rather than just -gt?  You can have it as another parameter.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -200,4 +205,614 @@ Describe "TabCompletion" -Tags CI {
...
+
+        It "Tab completion for wsman provider" -Skip:(!$IsWindows) {
+            $beforeTab = 'wsman::localh'
+            $afterTab = 'wsman::localhost'
+            $res = TabExpansion2 -inputScript $beforeTab -cursorColumn $beforeTab.Length
+            $res.CompletionMatches.Count | Should BeGreaterThan 0
+            $res.CompletionMatches[0].CompletionText | Should Be $afterTab
+        }
+
+        It "Tab completion for filesystem provider qualified path" -Skip:(!$IsWindows) {
```
SteveL-MSFT: This test is not Windows specific

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -200,4 +205,614 @@ Describe "TabCompletion" -Tags CI {
...
+
+        It "Tab completion for functions takes precedence over file system" {
+            $myf = Join-Path -Path $TestDrive -ChildPath myf
+            try {
+                New-Item -Path $myf -ItemType File -Force
+                function MyFunction { "Hi there" }
+
+                $inputStr = "myf"
+                $res = TabExpansion2 -inputScript $inputStr -cursorColumn $inputStr.Length
+                $res.CompletionMatches[0].CompletionText | Should be 'MyFunction'
```
SteveL-MSFT: you should check that it cycles to the file system

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -200,4 +205,614 @@ Describe "TabCompletion" -Tags CI {
...
+        It "Input '<inputStr>' should throw in tab completion" -TestCases $testCases {
+            param($inputStr, $expected)
+            $inputStr | ShouldBeErrorId $expected
+        }
+    }
+
+    Context "DSC tab completion tests" {
+        BeforeAll {
+            $testCases = @(
+                @{ inputStr = 'Configura'; expected = 'Configuration' }
```
SteveL-MSFT: Need tests against `import-dscresource` and (I think) the imported resource itself

SteveL-MSFT (changes requested):
Almost there!

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
daxian-dbw: tab completion not only find the files in the folder, but also all applications startings with `$inputStr`, for example, when `$inputStr = "b"`, there are 40 completion results. This test makes sure that the first one is the corresponding file in the same folder.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
daxian-dbw: Fixed.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
daxian-dbw: Fixed.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
daxian-dbw: Fixed.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
daxian-dbw: Fixed. Make the tests use both `/` and `\` as directory separators.

daxian-dbw:
@SteveL-MSFT @adityapatwardhan Your comments have been addressed, please take another look when you have time.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
SteveL-MSFT: ok

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
SteveL-MSFT: I see tests for `/`, but you should have a single test where the path contains both `\` and `/` and is normalized to the OS seperator

SteveL-MSFT (changes requested):

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
daxian-dbw: Update the tests to include both `\` and `/` in paths to be completed.

daxian-dbw:
@SteveL-MSFT your comment is addressed, please take another look when you have time. Thanks!

SteveL-MSFT (approved):

adityapatwardhan (approved):

PR #4360, opened 2017-07-27
add amazonlinux compatibility in install-powershell.sh

DarwinJS:
Amazon linux is a fine mix of CentOS versions ;)  Needed to make a few adaptations to get the installer working for it.  It might also pickup older versions of CentOS that use /etc/system-release instead of /etc/redhat-release

### DIFF
--- a/tools/install-powershell.sh
+++ b/tools/install-powershell.sh
@@ -74,6 +74,11 @@ else
             DIST=`cat /etc/redhat-release |sed s/\ release.*//`
             PSUEDONAME=`cat /etc/redhat-release | sed s/.*\(// | sed s/\)//`
             REV=`cat /etc/redhat-release | sed s/.*release\ // | sed s/\ .*//`
+        elif [ -f /etc/system-release ] ; then
+            DistroBasedOn='redhat'
+            DIST=`cat /etc/system-release |sed s/\ release.*//`
+            PSUEDONAME=`cat /etc/system-release | sed s/.*\(// | sed s/\)//`
+            REV=`cat /etc/system-release | sed s/.*release\ // | sed s/\ .*//`
         elif [ -f /etc/SuSE-release ] ; then
             DistroBasedOn='suse'
             PSUEDONAME=`cat /etc/SuSE-release | tr "\n" ' '| sed s/VERSION.*//`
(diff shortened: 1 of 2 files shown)


DarwinJS:
Tested and working on AMI:
ami-c58c1dd3 (amzn-ami-hvm-2017.03.0.20170417-x86_64-gp2)
Had to download install-powershell.sh and installpsh-redhat.sh and chmod 755.

TravisEz13:
It would be good if we have a way of testing this.  Do you want to create a docker file with amazonlinux https://hub.docker.com/_/amazonlinux/  which we can use to verify this works during a release?

DarwinJS:
@TravisEz13 - not sure - the changes were quite moderate - probably just changes needed to support earlier versions of CentOS.

Do we test all the other supported distros and versions for the installer?

TravisEz13:
If you don't think it's needed, please file an issue and mention this PR.  It's not much work if the installer is working.

Also, Please use the present tense, imperative style for commit messages and title per [Contribution Guide](https://github.com/PowerShell/PowerShell/blob/7a8589a8c516c3048b4555415b9ac13a24b6d6ed/.github/CONTRIBUTING.md#pull-request-submission)

DarwinJS:
@TravisEz13 - I'm not sure what I should be putting in the issue?  This PR is against issue 4359.  If you mean that I file one for the docker file - do you mean I ask for one to be made or mention that one is not needed?

DarwinJS:
@TravisEz13 - I might be interested in doing the docker file.  Is there an existing pattern or guidance document?  I looked around for docker files in the testing files of the repository - I found one - but can't tell exactly how to wire a new dockerfile into the automated testing.

TravisEz13:
@DarwinJS  Just write a docker file, similar to the ones in https://github.com/PowerShell/PowerShell/tree/master/docker/release and then add it to the list of tests cases here:  https://github.com/PowerShell/PowerShell/blob/master/docker/tests/containerTestCommon.psm1#L68
There is a md on how to run the tests in the same folder.

The issue would say that the above needs to be done.

TravisEz13 (approved):

TravisEz13:
docker file submitted on #4393

PR #4353, opened 2017-07-26
Fix DeviceGuard data being returned from Get-ComputerInfo and Start-Process test failure

SteveL-MSFT:
fix Get-ComputerInfo return of status for DeviceGuard
fix corresponding tests
this is causing one of the failures in our nightly run

PR #4362, opened 2017-07-27
Make ShellExecuteEx use STA

### LINKED ISSUE #2969: powershell should run `ShellExecuteEx` on STA thread
[Edited by @daxian-dbw]

PR #3281 adds support to ShellExecute in powershell core when it's running on Windows full SKUs. With that fix, `Start-Process` is able to run MSI in powershell core.

However, it's not a complete solution. It calls ShellExecuteEx directly in MTA thread instead of from STA thread [as suggested in MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/bb762154(v=vs.85).aspx). Therefore, the current solution may not work for some shell extensions that require COM.

In .NET Core, managed threads are all eagerly initialized with MTA mode, so to call ShellExecuteEx from a STA thread, we need to create a native thread using CreateThread function and initialize COM with STA on that thread. In #3281, we are calling ShellExecuteEx directly on MTA thread, and it works for things like opening a folder in File Explorer, opening a file with the application that is associated with its extension in shell, opening URL in web browser and etc, but it's not guaranteed to work in all ShellExecution scenarios.

Will keep this issue open to track the "invoke-on-STA-thread" work.

-------------------------------------------------------------------------------------------------

```powershell
PS:4> Start-Process .\PowerShell_6.0.0-alpha.17-win10-win2016-x64.msi -PassThru

 NPM(K)    PM(M)      WS(M)     CPU(s)     Id  SI ProcessName
 ------    -----      -----     ------     --  -- -----------
      9     6.15       7.55       0.03  20092   3 msiexec

PS:5> $PSVersionTable

GitCommitId                    v6.0.0-alpha.16-54-gc86a287726fe366895723dfcc7739f93038615ea
CLRVersion
```

SteveL-MSFT:
fix shellexecuteex call to use STA as recommended by MSDN
also removed some FullCLR code (left the TODO as it's a bigger item)

Fix https://github.com/PowerShell/PowerShell/issues/2969

### DIFF
--- a/src/System.Management.Automation/engine/Utils.cs
+++ b/src/System.Management.Automation/engine/Utils.cs
@@ -23,13 +23,8 @@ using System.Text;
 
 using TypeTable = System.Management.Automation.Runspaces.TypeTable;
 
-#if CORECLR
 using System.Diagnostics;
 using Microsoft.Win32.SafeHandles;
-#else
-using System.Security.Principal;
-using PSUtils = System.Management.Automation.PsUtils;
-#endif
 
 namespace System.Management.Automation
 {
@@ -242,43 +237,9 @@ namespace System.Management.Automation
 
         internal static string GetApplicationBase(string shellId)
         {
-#if CORECLR
             // Use the location of SMA.dll as the application base.
             Assembly assembly = typeof(PSObject).GetTypeInfo().Assembly;
             return Path.GetDirectoryName(assembly.Location);
-#else
-            // This code path applies to Windows FullCLR inbox deployments. All CoreCLR
-            // implementations should use the location of SMA.dll since it must reside in PSHOME.
-            //
-            // try to get the path from the registry first
-            string result = GetApplicationBaseFromRegistry(shellId);
-            if (result != null)
-            {
-                return result;
-            }
-
-            // The default keys aren't installed, so try and use the entry assembly to
-            // get the application base. This works for managed apps like minishells...
-            Assembly assem = Assembly.GetEntryAssembly();
-            if (assem != null)
-            {
-                // For minishells, we just return the executable path.
-                return Path.GetDirectoryName(assem.Location);
-            }
-
-            // For unmanaged host apps, look for the SMA dll, if it's not GAC'ed then
-            // use it's location as the application base...
-            assem = typeof(PSObject).GetTypeInfo().Assembly;
-            string gacRootPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.Windows), "Microsoft.Net\\assembly");
-            if (!assem.Location.StartsWith(gacRootPath, StringComparison.OrdinalIgnoreCase))
-            {
-                // For other hosts.
-                return Path.GetDirectoryName(assem.Location);
-            }
-
-            // otherwise, just give up...
-            return "";
-#endif
         }
 
         private static string[] s_productFolderDirectories;
@@ -324,11 +285,7 @@ namespace System.Management.Automation
                 // TODO: #1184 will resolve this work-around
                 // Side-by-side versions of PowerShell use modules from their application base, not
                 // the system installation path.
-#if CORECLR
                 progFileDir = Path.Combine(appBase, "Modules");
-#else
-                progFileDir = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles), "WindowsPowerShell", "Modules");
-#endif
 
                 if (!string.IsNullOrEmpty(progFileDir))
                 {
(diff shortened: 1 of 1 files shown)


test/powershell/Modules/Microsoft.PowerShell.Management/Start-Process.Tests.ps1
```diff
@@ -106,6 +106,7 @@ Describe "Start-Process" -Tags @("CI","SLOW") {
     }
 
     It "Should open the application that associates with extension '.txt'" -Skip:(!$isFullWin) {
+        "c:\windows\system32\notepad.exe" | Should Exist
```
iSazonov: I suggest use `%WINDIR%`.

SteveL-MSFT: Ignore this for now, trying to figure out why the test succeeds on my machine but fails in AppVeyor

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1666,35 +1546,10 @@ internal static Process Start(ProcessStartInfo startInfo, ProcessWindowStyle win
...
-
-                    if(errorCode == NativeMethods.ERROR_BAD_EXE_FORMAT || errorCode == NativeMethods.ERROR_EXE_MACHINE_TYPE_MISMATCH)
-                    {
-                        throw new Win32Exception(errorCode, "InvalidApplication");
-                    }
-                    
-                    throw new Win32Exception(errorCode);
-                }
+                Thread thread = new Thread(InvokeShellExecuteEx);
+                thread.SetApartmentState(System.Threading.ApartmentState.STA);
```
daxian-dbw: We should check if the current thread is STA already. If so, invoke the method without creating a new thread.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1719,6 +1574,39 @@ internal static Process Start(ProcessStartInfo startInfo, ProcessWindowStyle win
             return processToReturn;
         }
 
+        private static void InvokeShellExecuteEx(object param)
+        {
+            NativeMethods.ShellExecuteInfo shellExecuteInfo = (NativeMethods.ShellExecuteInfo)param;
+            if (!NativeMethods.ShellExecuteEx(shellExecuteInfo))
+            {
+                int errorCode = Marshal.GetLastWin32Error();
+                if (errorCode == 0)
```
daxian-dbw: The error handling code needs to happen in the main thread, otherwise, the Win32Exception thrown from here will be ignored.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1666,35 +1546,10 @@ internal static Process Start(ProcessStartInfo startInfo, ProcessWindowStyle win
...
-                    {
-                        throw new Win32Exception(errorCode, "InvalidApplication");
-                    }
-                    
-                    throw new Win32Exception(errorCode);
-                }
+                Thread thread = new Thread(InvokeShellExecuteEx);
+                thread.SetApartmentState(System.Threading.ApartmentState.STA);
+                thread.Start(shellExecuteInfo);
+                thread.Join();
```
daxian-dbw: The Win32Exception thrown from the new thread will probably be ignored. We should handle the error in the main thread.
Please take a look at the .NET implementation at https://referencesource.microsoft.com/#System/services/monitoring/system/diagnosticts/Process.cs,2239
The `ShellExecuteOnSTAThread` method is defined at https://referencesource.microsoft.com/#System/services/monitoring/system/diagnosticts/Process.cs,2856

test/powershell/Modules/Microsoft.PowerShell.Management/Start-Process.Tests.ps1
SteveL-MSFT: Fixed as part of https://github.com/PowerShell/PowerShell/pull/4353

src/System.Management.Automation/engine/Utils.cs
SteveL-MSFT: will fix

src/System.Management.Automation/engine/Utils.cs
SteveL-MSFT: will fix

src/System.Management.Automation/engine/Utils.cs
SteveL-MSFT: will fix

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1600,12 +1495,15 @@ public static void SetTestHook(string property, bool value)
...
 
-#if CORECLR && !UNIX
+#if !UNIX
     /// <summary>
     /// Helper to start process using ShellExecuteEx. This is used only in PowerShell Core on Full Windows.
     /// </summary>
     internal class ShellExecuteHelper
     {
+        private static bool _succeeded;
+        private static int _errorCode;
```
daxian-dbw: These 2 static fields will cause race condition when multiple Runsapces are running `ShellExecuteHelper.Start` at the same time. Using a [helper class](https://referencesource.microsoft.com/#System/services/monitoring/system/diagnosticts/Process.cs,2839) like in .NET implementation may be a better approach.

SteveL-MSFT: will fix

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1719,6 +1591,56 @@ internal static Process Start(ProcessStartInfo startInfo, ProcessWindowStyle win
             return processToReturn;
         }
 
+        private void ShellExecuteFunction() {
```
daxian-dbw: Please use the existing style -- put the open parenthesis on a new line.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1719,6 +1591,56 @@ internal static Process Start(ProcessStartInfo startInfo, ProcessWindowStyle win
...
+                Thread thread = new Thread(threadStart);
+                thread.SetApartmentState(System.Threading.ApartmentState.STA);
+                thread.Start();
+                thread.Join();
+            }
+            else
+            {
+                ShellExecuteFunction();
+            }
+            return _succeeded;
```
daxian-dbw: `_succeeded` is never set to true. Maybe you can set its default value to true, or you can use similar way as in https://referencesource.microsoft.com/#System/services/monitoring/system/diagnosticts/Process.cs,2850

SteveL-MSFT: missed that, will fix

daxian-dbw (approved):

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -324,11 +285,7 @@ private static string[] GetProductFolderDirectories()
                 // TODO: #1184 will resolve this work-around
                 // Side-by-side versions of PowerShell use modules from their application base, not
                 // the system installation path.
-#if CORECLR
                 progFileDir = Path.Combine(appBase, "Modules");
-#else
-                progFileDir = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ProgramFiles), "WindowsPowerShell", "Modules");
```
iSazonov: Don't we want load Windows PowerShell Modules from %ProgramFiles%\WindowsPowerShell? Currently we use `PSModulePath` as first step solution.

SteveL-MSFT: We'll have a different solution for supporting Windows PowerShell `PSModulePath` in PSCore6 that replaces the current interim solution

Issue #3565, opened 2017-04-14
Review all if/def and Unix specific code after moving to .NET Core 2.0


daxian-dbw:
PowerShell Core is moving to .NET Core 2.0 with PR #3556. Once that PR is merged, we need to review all if/def and Unix specific code to:
- Clean up the code
- Make sure the code make sense and is not a temporary hack

#### An Example (Please remove when fixed)
`Platform.InternalGetFolderPath` in `src/System.Management.Automation/CoreCLR/CorePsPlatform.cs` has the following code:
```
                case System.Environment.SpecialFolder.ProgramFiles:
                    folderPath = "/bin";
                    if (!System.IO.Directory.Exists(folderPath)) { folderPath = null; }
                    break;
                case System.Environment.SpecialFolder.ProgramFilesX86:
                    folderPath = "/usr/bin";
                    if (!System.IO.Directory.Exists(folderPath)) { folderPath = null; }
                    break;
                case System.Environment.SpecialFolder.System:
                case System.Environment.SpecialFolder.SystemX86:
                    folderPath = "/sbin";
                    if (!System.IO.Directory.Exists(folderPath)) { folderPath = null; }
                    break;
```
It use `/bin` for `ProgramFiles`, `/usr/bin` for `ProgramFilesX86` and `/sbin` for `System` and `SystemX86`. It's not clear if this is appropriate, so code like this should be reviewed.

iSazonov:
@daxian-dbw If this does not require a lot of time please create a sln file - without it IntelliSense does not work in VS Code (after json files was removed) properly.

daxian-dbw:
@iSazonov IntelliSense doesn't work for you? It works for me in VS Code after moving to `.csproj` file. Just run `Start-PSBuild -publish`, then open the `.csproj` file in VS Code, and if you see a prompt on top of VS Code asking to restore, just click `Close`, and things like reference resolution and IntelliSense should just work.
I thought creating a `sln` file would be just as simple as opening `powershell-win-core.csproj` in VS2017 and then saving a solution file, but it turned out the `.csproj` doesn't work properly within VS2017. I haven't got the chance to figure out why and fix it.

jeffbi:
@daxian-dbw, @iSazonov IntelliSense doesn't work for me either. I tried following the steps above, but no joy.

iSazonov:
As mentioned in #3356 there is a limitation https://github.com/OmniSharp/omnisharp-vscode/issues/1156 
[...]

PR #4366, opened 2017-07-28
Clean up '#if CORE' in ClrFacade.cs

iSazonov:
Related #4357 and #3565.

### DIFF
--- a/src/System.Management.Automation/utils/ClrFacade.cs
+++ b/src/System.Management.Automation/utils/ClrFacade.cs
@@ -63,11 +63,7 @@ namespace System.Management.Automation
         /// <returns>SafeHandle</returns>
         internal static SafeHandle GetSafeProcessHandle(Process process)
         {
-#if CORECLR
             return process.SafeHandle;
-#else
-            return new SafeProcessHandle(process.Handle);
-#endif
         }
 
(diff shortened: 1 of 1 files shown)


src/System.Management.Automation/utils/ClrFacade.cs
```diff
@@ -121,11 +105,7 @@ internal static T PtrToStructure<T>(IntPtr ptr)
             IntPtr ptr,
             bool deleteOld)
         {
-#if CORECLR
             Marshal.StructureToPtr<T>(structure, ptr, deleteOld);
-#else
-            Marshal.StructureToPtr(structure, ptr, deleteOld);
-#endif
```
daxian-dbw: For instances like the above ones, we should make changes in the call sites of those methods to directly use the API available in .NET Core, so that we can completely remove those methods from ClrFacade.cs.

daxian-dbw: You can leave them out, and I will clean them up later.

iSazonov: 😕 I don't understand this in depth so leave them to you.

daxian-dbw (approved):

PR #4370, opened 2017-07-28
Clean up '#if CORE' in AddType.cs

iSazonov:
Related #3565 and #4357.


iSazonov:
@SteveL-MSFT Again Invoke-Item.Tests.ps1 was failed on MacOS 😕 Because it seems very-very slow.

daxian-dbw (approved):

PR #4371, opened 2017-07-28
Clean up '#if CORE' in OrderObjectBase.cs

iSazonov:
Related #3565 and #4357.

The PR enable using a hex and int string for Culture parameter in order object cmdlets - Group-Object, Sort-Object, Compare-Object.

Second commit _only reformat_ the test file.

Tests is added in third commit.

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/Group-Object.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/Group-Object.Tests.ps1
@@ -118,8 +118,7 @@ Describe "Check 'Culture' parameter in order object cmdlets (Group-Object, Sort-
         if ( (Get-Culture).Name -eq "ru-Ru" ) {
             $testCulture = "ru-UA"
         }
-        else
-        {
+        else {
             $testCulture = "ru-RU"
         }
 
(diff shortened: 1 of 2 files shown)


test/powershell/Modules/Microsoft.PowerShell.Utility/Group-Object.Tests.ps1
```diff
@@ -92,16 +91,65 @@ Describe "Group-Object" -Tags "CI" {
...
+
+    It "Should accept a culture by name" {
+
+        if ( (Get-Culture).Name -eq "ru-Ru" ) {
+            $testCulture = "ru-UA"
+        }
+        else
+        {
+            $testCulture = "ru-RU"
+        }
```
daxian-dbw: nitpick: you are mixing the open parenthesis styles (here and also the following 2 It block) :) maybe unify them as:
```
if (...) {
   ...
} else {
   ...
}
```

But this doesn't block merging the PR.

daxian-dbw (approved):

test/powershell/Modules/Microsoft.PowerShell.Utility/Group-Object.Tests.ps1
iSazonov: Fixed.

daxian-dbw (approved):

PR #4375, opened 2017-07-28
Remove ClrFacade.GetUninitializedObject and 'if CORECLR' cleanup

daxian-dbw:
Related #3565 and #4357.
Break up the cleanup work for ClrFacade.cs into multiple PRs.
In this PR, `[System.Runtime.Serialization.FormatterServices]::GetUninitializedObject` is back in .NET Core, so we replace ` ClrFacade.GetUninitializedObject` with it.

### DIFF
--- a/src/System.Management.Automation/utils/ClrFacade.cs
+++ b/src/System.Management.Automation/utils/ClrFacade.cs
@@ -8,24 +8,18 @@ using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Management.Automation.Internal;
 using System.Management.Automation.Language;
 using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+using System.Runtime.Loader;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Security;
 using Microsoft.Win32.SafeHandles;
 using System.Runtime.InteropServices.ComTypes;
-
-#if CORECLR
-using System.Runtime.Loader; /* used in facade APIs related to assembly operations */
-using System.Management.Automation.Host;          /* used in facade API 'GetUninitializedObject' */
-using System.Management.Automation.Internal;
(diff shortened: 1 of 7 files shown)


src/System.Management.Automation/utils/ClrFacade.cs
```diff
@@ -497,55 +482,12 @@ internal static string ToDmtfDateTime(DateTime date)
...
-        }
-
-        /// <summary>
         /// Facade for ProfileOptimization.SetProfileRoot
         /// </summary>
         /// <param name="directoryPath">The full path to the folder where profile files are stored for the current application domain.</param>
         internal static void SetProfileOptimizationRoot(string directoryPath)
         {
-#if CORECLR
             PSAssemblyLoadContext.SetProfileOptimizationRootImpl(directoryPath);
```
iSazonov: Should we remove the method in follow PR?

src/System.Management.Automation/utils/ClrFacade.cs
```diff
@@ -554,11 +496,7 @@ internal static void SetProfileOptimizationRoot(string directoryPath)
         /// <param name="profile">The file name of the profile to use.</param>
         internal static void StartProfileOptimization(string profile)
         {
-#if CORECLR
             PSAssemblyLoadContext.StartProfileOptimizationImpl(profile);
```
iSazonov: Should we remove the method in follow PR?

iSazonov (approved):

src/System.Management.Automation/utils/ClrFacade.cs
daxian-dbw: Yes we should, and it requires corresponding changes in `CorePsAssemblyLoadContext.cs` as `SetProfileOptimizationRootImpl` is no longer needed. I will do it in subsequent PRs.

src/System.Management.Automation/utils/ClrFacade.cs
daxian-dbw: Yes, same as above.

Issue #4383, opened 2017-07-28
PowerShell crash when Invoke-Command and WinRM isn't available


SteveL-MSFT:
New NanoServer container images don't have WinRM

```
docker run -it microsoft/nanoserver-insider-powershell
invoke-command -computername . -command { get-command }
```

Expected:
an error that winrm is required

Actual:
```
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Management.Automation.Remoting.PrioritySendDataCollection.Clear()
   at System.Management.Automation.Remoting.Client.WSManClientSessionTransportManager.CloseAsync()
   at System.Management.Automation.Remoting.Client.BaseClientTransportManager.Finalize()
Invoke-Command : Unable to load DLL 'WsmSvc.dll': The specified module could not be found. (Exception from HRESULT:
0x8007007E)
At line:1 char:1
+ Invoke-Command -ComputerName . -Command { get-command }
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Invoke-Command], DllNotFoundException
    + FullyQualifiedErrorId : System.DllNotFoundException,Microsoft.PowerShell.Commands.InvokeCommandCommand
```

PR #4387, opened 2017-07-28
Give error instead of crashing if WSMan client lib not available

SteveL-MSFT:
Affects NanoServer Insider build where WinRM is not available.  Currently if user tries `invoke-command`, powershell crashes.

Validated using:
```powershell
PS> docker run -it -v C:\users\slee\repos\PowerShell\src\powershell-wincore\bin\Debug\netcoreapp2.0\win10-x64\publish:c:\powershell microsoft/nanoserver-insider-powershell
PS> c:\powershell\powershell
PS> invoke-command -computer . -command { "hello" }

invoke-command : WSMan client library is not installed
At line:1 char:1
+ invoke-command -ComputerName . -Command { "hello" }
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ResourceUnavailable: (:) [Invoke-Command], PSRemotingTransportException
    + FullyQualifiedErrorId : System.Management.Automation.Remoting.PSRemotingDataStructureException,Microsoft.PowerSh
   ell.Commands.InvokeCommandCommand
```

Fix https://github.com/PowerShell/PowerShell/issues/4383

### DIFF
--- a/src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
+++ b/src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
@@ -2597,7 +2597,16 @@ namespace System.Management.Automation.Remoting.Client
 #endif
 
                 _handle = IntPtr.Zero;
-                ErrorCode = WSManNativeApi.WSManInitialize(WSManNativeApi.WSMAN_FLAG_REQUESTED_API_VERSION_1_1, ref _handle);
+
+                try
+                {
+                    ErrorCode = WSManNativeApi.WSManInitialize(WSManNativeApi.WSMAN_FLAG_REQUESTED_API_VERSION_1_1, ref _handle);
+                }
+                catch (DllNotFoundException)
+                {
+                    throw new PSRemotingTransportException(
+                        StringUtil.Format(RemotingErrorIdStrings.WSManClientDllNotAvailable));
+                }
 
                 // input / output streams common to all connections
                 _inputStreamSet = new WSManNativeApi.WSManStreamIDSet_ManToUn(
--- a/src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
+++ b/src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
@@ -2604,8 +2604,7 @@ namespace System.Management.Automation.Remoting.Client
                 }
                 catch (DllNotFoundException)
                 {
-                    throw new PSRemotingTransportException(
-                        StringUtil.Format(RemotingErrorIdStrings.WSManClientDllNotAvailable));
+                    throw new PSRemotingTransportException(RemotingErrorIdStrings.WSManClientDllNotAvailable);
                 }
 
                 // input / output streams common to all connections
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/resources/RemotingErrorIdStrings.resx
```diff
@@ -1636,7 +1636,10 @@ All WinRM sessions connected to Windows PowerShell session configurations, such
   <data name="InvalidRoleCapabilityFileExtension" xml:space="preserve">
     <value>The provided role capability file {0} does not have the required .psrc extension.</value>
   </data>
-  <data name="SSHAbruptlyTerminated" >
+  <data name="SSHAbruptlyTerminated" xml:space="preserve">
     <value>The SSH transport process has abruptly terminated causing this remote session to break.</value>
   </data>
+  <data name="WSManClientDllNotAvailable" xml:space="preserve">
+    <value>WSMan client library is not installed</value>
```
iSazonov: Typo - skipped a point at end. 
And may be add short recomendation to user how resolve the problem?

PaulHigin: Agreed.  It sounds like this only affects Nano.  So maybe something like "WSMan client library is not installed.  This parameter set requires WSMan for PowerShell remoting commands such as Invoke-Command, New-PSSession, Enter-PSSession."  The problem is that SSH remoting may be available if SSH client is installed.  But I don't know if that is a likely scenario.

PaulHigin (changes requested):

src/System.Management.Automation/resources/RemotingErrorIdStrings.resx
PaulHigin: I see.  Yes I agree that the error message is fine as is.

PaulHigin (approved):
LGTM

src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
```diff
@@ -2597,7 +2597,16 @@ internal WSManAPIDataCommon()
...
-                ErrorCode = WSManNativeApi.WSManInitialize(WSManNativeApi.WSMAN_FLAG_REQUESTED_API_VERSION_1_1, ref _handle);
+
+                try
+                {
+                    ErrorCode = WSManNativeApi.WSManInitialize(WSManNativeApi.WSMAN_FLAG_REQUESTED_API_VERSION_1_1, ref _handle);
+                }
+                catch (DllNotFoundException)
+                {
+                    throw new PSRemotingTransportException(
+                        StringUtil.Format(RemotingErrorIdStrings.WSManClientDllNotAvailable));
```
daxian-dbw: `StringUtil.Format(..)` is unnecessary as no arguments is required for the error message.

src/System.Management.Automation/resources/RemotingErrorIdStrings.resx
joeyaiello: How about "This parameter set requires WSMan, and no WSMan client library is available. WSMan is either not installed or unavailable for this platform. It might also be possible to use SSH-based remoting instead."

SteveL-MSFT: @joeyaiello I'm fine with everything except the last sentence as SSH may or may not be installed.  I don't think we need to go to that much detail at this point.  Ideally, we should just tell the user `run Enable-PSRemoting` and magic happens, but that's not anytime soon.

src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
SteveL-MSFT: will fix.  lazy cut and paste :)

joeyaiello (approved):
Fine on the SSH pushback, good enough for me.

PR #4395, opened 2017-07-31
Fix unblock read only file

### LINKED ISSUE #4390: Unblock-File provides no message if the file is ReadOnly (and thus fails)
Steps to reproduce
------------------
Using Pester;
```powershell
        $TestPath = "TestDrive:\test.ps1"
        Set-Content -Path $TestPath -value 'Break'
        $ZoneIdentifier = {
            [ZoneTransfer]
            ZoneId=3
        }
        Set-Content -Path $Testpath -Value $ZoneIdentifier -Stream 'Zone.Identifier'
        Set-ItemProperty -Path $TestPath -Name IsReadOnly -Value $True
        .\Set-UnblockPS1.ps1 -Path $Testpath
```

Expected behavior
-----------------
- Warning/Error/Message (anything?) that the file has failed to be unblocked. Even a generic error would be useful - for example:
```powershell
Write-Warning -Message "Skipping file $($file) as it is ReadOnly"
```
Alternative would be:
```powershell
Write-Warning -Message: "Failed to unblock $($file)"
```

Actual behavior
---------------
No output (also not via -Verbose or -Debug) is given.

Environment data
----------------

```powershell
> $PSVersionTable
GitCommitId                    v6.0.0-beta.4
OS                             Microsoft Windows 10.0.16232
```

```powershell
> $PSVersionTable
PSVersion                      5.1.16232.1000
BuildVersion                   10.0.16232.1000
```

iSazonov:
Adds a new non-terminating error 

Fix #4390.

### Problem ###
If a file is read only it is skipping silently.

### Fix ###
Analize GetLastError() and throw if a file stream hasn't been removed.

First commit only refactor tests to use `ShouldBeErrorId`.

### DIFF
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -8770,7 +8770,11 @@ namespace System.Management.Automation.Internal
             }
             string resultPath = path + adjustedStreamName;
 
-            NativeMethods.DeleteFile(resultPath);
+            if (!NativeMethods.DeleteFile(resultPath))
+            {
+                int error = Marshal.GetLastWin32Error();
+                throw new Win32Exception(error);
+            }
         }
 
         internal static void SetZoneOfOrigin(string path, SecurityZone securityZone)
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
@@ -7,6 +7,7 @@ Copyright (c) Microsoft Corporation.  All rights reserved.
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.ComponentModel;
 using System.Diagnostics.CodeAnalysis;
 using System.Management.Automation;
 using System.Management.Automation.Internal;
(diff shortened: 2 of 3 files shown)


SteveL-MSFT:
@iSazonov what is the error message the user sees now?

iSazonov:
[MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363915(v=vs.85).aspx) says that `GetLastError` can return either "Access denied" or "File not found".
```powershell
Unblock-File : Access is denied
At line:1 char:1
+ Unblock-File -Path $Testpath
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Unblock-File], Win32Exception
    + FullyQualifiedErrorId : System.ComponentModel.Win32Exception,Microsoft.PowerShell.Commands.UnblockFileCommand
```

SteveL-MSFT (dismissed):
LGTM

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -8770,7 +8770,11 @@ internal static void DeleteFileStream(string path, string streamName)
             }
             string resultPath = path + adjustedStreamName;
 
-            NativeMethods.DeleteFile(resultPath);
+            if (!NativeMethods.DeleteFile(resultPath))
+            {
+                int error = Marshal.GetLastWin32Error();
+                throw new Win32Exception(error);
```
lzybkr: Shouldn't this be a non-terminating error? The following should behave similarly:
```powershell
dir | Unblock-File
dir | Remove-Item
```
In the `Remove-Item` case, a read only file reports an error, but all other files are deleted.

iSazonov: Good catch! 
`AlternateDataStreamUtilities` uses the pattern. So maybe add `try-catch` in Unblock-File cmdlet file?

iSazonov:
@lzybkr @SteveL-MSFT Please continue the code review.

lzybkr (approved):
Make sure the final commit message points out this is a new non-terminating error.

SteveL-MSFT (approved):

PR #4398, opened 2017-07-31
Clean up '#if CORE' in CimIndicationWatcher.cs

iSazonov:
Related #3565 and #4357.


src/Microsoft.Management.Infrastructure.CimCmdlets/CimIndicationWatcher.cs
```diff
@@ -258,9 +258,6 @@ private void NewSubscriptionResultHandler(object src, CimSubscriptionEventArgs a
         /// If set EnableRaisingEvents to false, which will be ignored
         /// </para>
         /// </summary>
-#if !CORECLR
-        [BrowsableAttribute(false)]
-#endif
```
daxian-dbw: `BrowsableAttribute` is back in .NET Core, so we should leave `[BrowsableAttribute(false)]` there.

iSazonov: Fixed.

daxian-dbw (approved):

PR #4402, opened 2017-07-31
Clean up '#if CORE' in MshHostTraceListener.cs

iSazonov:
Related #3565 and #4357.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/trace/MshHostTraceListener.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/trace/MshHostTraceListener.cs
@@ -62,35 +62,8 @@ namespace Microsoft.PowerShell.Commands
         [SecurityPermission(SecurityAction.LinkDemand)]
         protected override void Dispose(bool disposing)
         {
-#if CORECLR
             base.Dispose(disposing);
-#else
-            try
-            {
-                if (disposing)
-                {
-                    this.Close();
-                }
-            }
-            finally
-            {
-                base.Dispose(disposing);
-            }
-#endif
         }
-
-#if !CORECLR
-        /// <summary>
-        /// Closes the dialog and then calls the base class Close
-        /// </summary>
-        [SecurityPermission(SecurityAction.LinkDemand)]
-        public override void Close()
-        {
-            // Call the base class close
-
-            base.Close();
-        }
-#endif
 
         #endregion TraceListener constructors and disposer
 
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/trace/MshHostTraceListener.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/trace/MshHostTraceListener.cs
@@ -62,7 +62,17 @@ namespace Microsoft.PowerShell.Commands
         [SecurityPermission(SecurityAction.LinkDemand)]
         protected override void Dispose(bool disposing)
         {
-            base.Dispose(disposing);
+            try
+            {
+                if (disposing)
+                {
+                    this.Close();
+                }
+            }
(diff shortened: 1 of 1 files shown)


daxian-dbw (dismissed):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/trace/MshHostTraceListener.cs
```diff
@@ -62,35 +62,8 @@ internal PSHostTraceListener(PSCmdlet cmdlet)
...
-#if !CORECLR
-        /// <summary>
-        /// Closes the dialog and then calls the base class Close
-        /// </summary>
-        [SecurityPermission(SecurityAction.LinkDemand)]
-        public override void Close()
-        {
-            // Call the base class close
-
-            base.Close();
```
daxian-dbw: The `Close` virtual method is back, so maybe we should restore the full clr implementation.

iSazonov: In CoreFX it is dummy method https://github.com/dotnet/corefx/blob/7da531794fe1f77668b0705502cd0658eba75889/src/System.Diagnostics.TraceSource/src/System/Diagnostics/TraceListener.cs#L173
In .Net too https://referencesource.microsoft.com/#System/compmod/system/diagnostics/TraceListener.cs,2facc6077fc08453
What is a behavior we want get?

daxian-dbw: How the method is implemented in the base class is implementation details. Think about `base.Dispose(disposing);`, it currently also does nothing in .NET Core : https://github.com/dotnet/corefx/blob/7da531794fe1f77668b0705502cd0658eba75889/src/System.Diagnostics.TraceSource/src/System/Diagnostics/TraceListener.cs#L79, but does that mean we can avoid calling it as well?

But I think we can remove the method `public override void Close()`. It doesn't nothing other than calling `base.Close()`, which is the same as directly calling `Close()` from the base type.

iSazonov: The method comment confuse too.

iSazonov: Is something I have to do?

daxian-dbw: I think we should keep full clr code for `Dispose(bool disposing)` and remove the method `public override void Close()`. Thoughts?

iSazonov: [MSDN](https://msdn.microsoft.com/en-us/library/system.diagnostics.tracelistener.close(v=vs.110).aspx) says that Close() should be overridden. But we don't need override it. So we could remove both our methods (Close() and Dispose()) as dummy and call base class virtual methods.

iSazonov: @daxian-dbw Could you please continue?

daxian-dbw: I think the current change is good. The `Dispose(bool disposing)` implementation in this class is better to be kept so that if a sub class of `PSHostTraceListener` can depend on `Dispose(true)` to call `close()`. It's not a big deal, but this `Dispose` implementation indeed does a bit more than the base type `Dispose` implementation.

daxian-dbw (approved):

PR #4409, opened 2017-07-31
Clean up '#if CORE' in CurrentConfigurations.cs

iSazonov:
Related #3565 and #4357.


daxian-dbw (approved):

daxian-dbw:
For simple changes like: removing unnecessary using directives, or purely removing enclosing `if/def`s, please group them in one PR in future.

PR #4417, opened 2017-07-31
Clean up '#if CORE' in Attributes.cs

iSazonov:
Related #3565 and #4357.

Now `System.Transactions` is in CoreFX 2.0 but all transaction cmdlets is only in FullCLR and we haven't tests at all.

### DIFF
--- a/src/System.Management.Automation/engine/Attributes.cs
+++ b/src/System.Management.Automation/engine/Attributes.cs
@@ -370,13 +370,7 @@ namespace System.Management.Automation
             get { return _supportsTransactions; }
             set
             {
-#if !CORECLR
                 _supportsTransactions = value;
-#else
-                // Disable 'SupportsTransactions' in CoreCLR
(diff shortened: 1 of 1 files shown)


src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -370,13 +370,7 @@ public bool SupportsTransactions
             get { return _supportsTransactions; }
             set
             {
-#if !CORECLR
                 _supportsTransactions = value;
-#else
-                // Disable 'SupportsTransactions' in CoreCLR
-                // No transaction supported on CSS due to the lack of System.Transactions namespace
-                _supportsTransactions = false;
-#endif
```
daxian-dbw: Let's keep the transaction related if/defs for now. More work is needed to enable transaction support.

iSazonov:
From @daxian-dbw 
>Let's keep the transaction related if/defs for now. More work is needed to enable transaction support.

PR #4422, opened 2017-07-31
Clean up '#if CORE' in ConnectionFactory.cs

iSazonov:
Related #3565 and #4357.

### DIFF
--- a/src/System.Management.Automation/engine/hostifaces/ConnectionFactory.cs
+++ b/src/System.Management.Automation/engine/hostifaces/ConnectionFactory.cs
@@ -82,11 +82,7 @@ namespace System.Management.Automation.Runspaces
         /// <exception cref="ArgumentNullException">
         /// Thrown when runspaceConfiguration is null
         /// </exception>
-#if CORECLR
         internal
-#else
-        public
-#endif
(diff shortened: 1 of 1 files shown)


src/System.Management.Automation/engine/hostifaces/ConnectionFactory.cs
```diff
@@ -117,11 +113,7 @@ static Runspace CreateRunspace(RunspaceConfiguration runspaceConfiguration)
         /// <exception cref="ArgumentNullException">
         /// Thrown when runspaceConfiguration is null
         /// </exception>
-#if CORECLR
         internal
-#else
-        public
-#endif
         static Runspace CreateRunspace(PSHost host, RunspaceConfiguration runspaceConfiguration)
```
daxian-dbw: For these 2 changes, please move `internal` to the same line of the rest of the method signature. Like:
`internal static Runspace CreateRunspace(RunspaceConfiguration runspaceConfiguration)`

iSazonov: Fixed.

daxian-dbw (approved):

PR #4423, opened 2017-07-31
Clean up '#if CORE' in DynamicInstructionN.cs

iSazonov:
Related #3565 and #4357.


daxian-dbw (approved):

daxian-dbw:
For simple changes like this one -- removing unnecessary using directives -- you should use one PR for all of them in future.

PR #4425, opened 2017-07-31
Clean up '#if CORE' in Utilities.cs

iSazonov:
Related #3565 and #4357.

### DIFF
--- a/src/System.Management.Automation/engine/interpreter/Utilities.cs
+++ b/src/System.Management.Automation/engine/interpreter/Utilities.cs
@@ -836,10 +836,6 @@ namespace System.Management.Automation.Interpreter
         /// </summary>
         private StorageInfo CreateStorageInfo()
         {
-            // we do our own locking, tell hosts this is a bad time to interrupt us.
-#if !SILVERLIGHT && !CORECLR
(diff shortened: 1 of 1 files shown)


src/System.Management.Automation/engine/interpreter/Utilities.cs
```diff
@@ -886,9 +882,6 @@ private StorageInfo CreateStorageInfo()
                     // let others access the storage again
                     Interlocked.Exchange(ref _stores, curStorage);
                 }
-#if !SILVERLIGHT && !CORECLR
-                Thread.EndCriticalRegion();
-#endif
```
daxian-dbw: `Thread.EndCriticalRegion()` is back in .NET Core, so let's enable them here.

iSazonov: Fixed.

iSazonov: it seems we should return to ClrFacade.cs `internal static SafeHandle GetSafeProcessHandle(Process process)`

daxian-dbw (approved):

PR #4428, opened 2017-07-31
Clean up '#if CORE' in ProcessCodeMethods.cs

iSazonov:
Related #3565 and #4357.

### DIFF
--- a/src/System.Management.Automation/engine/ProcessCodeMethods.cs
+++ b/src/System.Management.Automation/engine/ProcessCodeMethods.cs
@@ -58,11 +58,8 @@ namespace Microsoft.PowerShell {
             Diagnostics.Assert(process != null, "Ensure process is not null before calling");
             PROCESS_BASIC_INFORMATION pbi;
             int size;
-#if CORECLR
             var res = NtQueryInformationProcess(process.SafeHandle.DangerousGetHandle(), 0, out pbi, Marshal.SizeOf<PROCESS_BASIC_INFORMATION>(), out size);
-#else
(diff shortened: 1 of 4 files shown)


src/System.Management.Automation/engine/ProcessCodeMethods.cs
```diff
@@ -58,11 +58,8 @@ internal static int GetParentPid(Process process)
             Diagnostics.Assert(process != null, "Ensure process is not null before calling");
             PROCESS_BASIC_INFORMATION pbi;
             int size;
-#if CORECLR
             var res = NtQueryInformationProcess(process.SafeHandle.DangerousGetHandle(), 0, out pbi, Marshal.SizeOf<PROCESS_BASIC_INFORMATION>(), out size);
-#else
-            var res = NtQueryInformationProcess(process.Handle, 0, out pbi, Marshal.SizeOf<PROCESS_BASIC_INFORMATION>(), out size);
```
daxian-dbw: `Process.Handle` is back, so probably we should use `NtQueryInformationProcess(process.Handle, ...` here.

iSazonov: It seems we should return in ClrFacade.cs `internal static SafeHandle GetSafeProcessHandle(Process process)` - we removed '#if CORE' there too.

daxian-dbw: That method in ClrFacade.cs should go away now I think.

iSazonov: Fixed.

daxian-dbw (approved):
LGTM

iSazonov:
@daxian-dbw Could you please merge?

PR #4429, opened 2017-07-31
Clean up '#if CORE' in PropertyAccessor.cs

iSazonov:
Related #3565 and #4357.


src/System.Management.Automation/engine/PropertyAccessor.cs
```diff
@@ -491,7 +485,6 @@ private void RemoveValueFromFile<T>(string fileName, string key)
         }
     }
 
-#endif // CORECLR
 
     internal class RegistryAccessor : ConfigPropertyAccessor
```
daxian-dbw: Maybe the `RegistryAccessor` should be removed now to increase the code coverage of this file. @mirichmo what do you think?

src/System.Management.Automation/engine/PropertyAccessor.cs
```diff
@@ -31,13 +30,9 @@ internal abstract class ConfigPropertyAccessor
         /// </summary>
         static ConfigPropertyAccessor()
         {
-#if CORECLR
             Instance = Platform.IsInbox
```
mirichmo: This can be simplified to:
Instance = new JsonConfigFileAccessor(); 

Platform.IsInbox() should be removed from the code as well

src/System.Management.Automation/engine/PropertyAccessor.cs
mirichmo: It is safe to remove since RegistryAccessor is only used for Windows PowerShell

mirichmo (changes requested):

src/System.Management.Automation/engine/PropertyAccessor.cs
iSazonov: "Instance = " is fixed.

For `Platform.IsInbox()` I'll open new PR (We have still 4 reference on it).

mirichmo (approved):

PR #4431, opened 2017-07-31
Clean up '#if CORE' in remoterunspace.cs

iSazonov:
Related #3565 and #4357.

### DIFF
--- a/src/System.Management.Automation/engine/remoting/client/remoterunspace.cs
+++ b/src/System.Management.Automation/engine/remoting/client/remoterunspace.cs
@@ -220,8 +220,7 @@ namespace System.Management.Automation
         /// <summary>
         /// runspaceConfiguration information for this runspace
         /// </summary>
-        internal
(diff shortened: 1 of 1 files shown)


src/System.Management.Automation/engine/remoting/client/remoterunspace.cs
```diff
@@ -220,11 +220,7 @@ private void SetEventHandlers()
         /// <summary>
         /// runspaceConfiguration information for this runspace
         /// </summary>
-#if CORECLR
         internal
-#else
-        public
-#endif
         override RunspaceConfiguration RunspaceConfiguration
```
daxian-dbw: Please move the method signature to the same line.

iSazonov: Fixed.

daxian-dbw (approved):

PR #4432, opened 2017-07-31
Clean up '#if CORE' in RemoteSessionCapability.cs

iSazonov:
Related #3565 and #4357.

### DIFF
--- a/src/System.Management.Automation/engine/remoting/common/WireDataFormat/RemoteSessionCapability.cs
+++ b/src/System.Management.Automation/engine/remoting/common/WireDataFormat/RemoteSessionCapability.cs
@@ -2,16 +2,14 @@
 Copyright (c) Microsoft Corporation.  All rights reserved.
 --********************************************************************/
 
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Management.Automation.Host;
 using System.Management.Automation.Internal.Host;
-
-#if !CORECLR
-using BinaryFormatter = System.Runtime.Serialization.Formatters.Binary.BinaryFormatter;
-#endif
-
+using System.Runtime.Serialization.Formatters.Binary;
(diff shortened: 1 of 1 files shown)


src/System.Management.Automation/engine/remoting/common/WireDataFormat/RemoteSessionCapability.cs
```diff
@@ -121,19 +112,15 @@ internal static byte[] GetCurrentTimeZoneInByteFormat()
...
             }
 
             return s_timeZoneInByteFormat;
         }
 
-#if !CORECLR // TimeZone Not In CoreCLR
         /// <summary>
         /// Gets the TimeZone of the destination machine. This may be null
         /// </summary>
         internal TimeZone TimeZone { get; set; }
```
PaulHigin: I see that BinaryFormatter type is now available in Core, good news!  Since this property is now available in Core can you also remove the "#if !CORECLR" at the two places it is referenced (serverremotesession.cs and EncodeAndDecode.cs)?  
https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/remoting/common/WireDataFormat/EncodeAndDecode.cs#L2405
https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/remoting/server/serverremotesession.cs#L747

PaulHigin (changes requested):

src/System.Management.Automation/engine/remoting/common/WireDataFormat/RemoteSessionCapability.cs
iSazonov: @PaulHigin Please see and comment #4433 - we have a problem with TimeZone.

iSazonov:
Some files depend on depredecated TimeZone type - so I close the PR and combine the files in #4433.

PR #4436, opened 2017-07-31
Fix deviceguard test failure

SteveL-MSFT:
investigating deviceguard failure.  only failing on AppVeyor.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/GetComputerInfoCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/GetComputerInfoCommand.cs
@@ -3683,9 +3683,13 @@ namespace Microsoft.PowerShell.Commands
     /// <summary>
     /// Device Guard hardware security properties
     /// </summary>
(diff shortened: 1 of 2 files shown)


daxian-dbw:
The DeviceGuard test is still failing:
```
[-] Test for DeviceGuard properties 107ms
   Expected {0} to be greater than {0}
   at line: 1392 in C:\projects\powershell\test\powershell\Modules\Microsoft.PowerShell.Management\Get-ComputerInfo.Tests.ps1
   1392:                     $observed.DeviceGuardRequiredSecurityProperties.Count | Should BeGreaterThan 0
```

adityapatwardhan:
If this is the only failure blocking daily builds, should we mark this test as pending on Windows, till we figure out the root cause. We have not had a successful daily build in about 19 days.

SteveL-MSFT:
@adityapatwardhan good idea, I'll submit that as a different PR so I can keep investigating this

iSazonov:
@SteveL-MSFT Maybe close the PR?

PR #4444, opened 2017-08-01
Clean up '#if CORE' in UpdatableHelpCommandBase.cs

iSazonov:
Related #3565 and #4357.


src/System.Management.Automation/help/UpdatableHelpCommandBase.cs
```diff
@@ -514,7 +514,6 @@ private void ProcessModule(UpdatableHelpModuleInfo module)
                     ProcessException(module.ModuleName, culture, new UpdatableHelpSystemException("AccessIsDenied",
                         e.Message, ErrorCategory.PermissionDenied, null, e));
                 }
-#if !CORECLR
```
daxian-dbw: Changes in this file need to be careful. The implementation was changed from using `WebClient` to `HttpClient`, and thus some `'if !CoreCLR'` blocks don't really make sense anymore. Let's leave this file aside for now.

iSazonov:
I close the PR because we need in-depth review.

PR #4445, opened 2017-08-01
Clean up '#if CORE' in RunspaceConfiguration.cs

iSazonov:
Related #3565 and #4357.

Similar  #4447, #4446, #4448


iSazonov:
Moved to #4448 .

PR #4449, opened 2017-08-01
Clean up '#if CORE' in FileSystemProvider.cs

iSazonov:
Related #3565 and #4357.

### DIFF
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -7966,26 +7966,6 @@ namespace Microsoft.PowerShell.Commands
             IntPtr OutBuffer, int nOutBufferSize,
             out int pBytesReturned, IntPtr lpOverlapped);
 
-#if !CORECLR
-
-        [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
-        private static extern IntPtr FindFirstFileName(
-                string lpFileName,
-                uint flags,
-                ref UInt32 StringLength,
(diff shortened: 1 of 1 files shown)


src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -7966,26 +7966,6 @@ private struct REPARSE_GUID_DATA_BUFFER
             IntPtr OutBuffer, int nOutBufferSize,
             out int pBytesReturned, IntPtr lpOverlapped);
 
-#if !CORECLR
-
-        [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
-        private static extern IntPtr FindFirstFileName(
```
daxian-dbw: Again, need to understand what those APIs are for. Maybe they are for features that need to be brought back to powershell core.

iSazonov: Currently it is not used. The API is specific for Windows file systems. It works with hard links - it is already in PowerShell Core.

daxian-dbw: Then `private static List<string> InternalGetTarget(string filePath)`  needs to be revisited as it basically only handles Unix platforms. Maybe just remove it, and change `public static IEnumerable<string> GetTarget(PSObject instance)` to directly handle `Windows/Unix` with `if UNIX/else`.

iSazonov:
I close th PR - we need in-depth review the API - perhaps it is not used but it is public.

PR #4450, opened 2017-08-01
Clean up '#if CORE' - remove Win32Native.cs

iSazonov:
Related #3565 and #4357.


src/System.Management.Automation/namespaces/Win32Native.cs
```diff
@@ -175,720 +175,5 @@ internal enum SECURITY_IMPERSONATION_LEVEL : short
         // Security Quality of Service flags
         internal const int SECURITY_ANONYMOUS = ((int)SECURITY_IMPERSONATION_LEVEL.Anonymous << 16);
         internal const int SECURITY_SQOS_PRESENT = 0x00100000;
-
-#if !CORECLR // Only enable/port what is needed by CORE CLR.
```
daxian-dbw: Not so sure that they should be removed. There are transaction related APIs here. We will need them if we want to enable transaction support in powershell core.

iSazonov: This is not used now. Any of this may be useful in the future. We can always add this if necessary. Or, by that time, it will be implemented in CoreFX.

daxian-dbw: If we are going to bring back the transaction support on windows, then I think it would be useful to keep these codes as references for that work. It can be cleaned later when working on the transaction support.

SteveL-MSFT: perhaps we can use a different #ifdef that we can find easily in the future

iSazonov: After we have cleaned up old codes under #if CORE the tag doesn't make any other sense as to find what we want to port.

iSazonov:
1. Top comment (// Only enable/port what is needed by CORE CLR.) says that the codes were brought in batch "just in case".
2. The codes is not compiled 

-- so I close the PR.

PR #4468, opened 2017-08-02
Use fpm 1.8.1 to build packages

daxian-dbw:


build.psm1
```diff
@@ -1265,7 +1265,8 @@ function Start-PSBootstrap {
             if ($Package) {
                 try {
                     # We cannot guess if the user wants to run gem install as root
-                    Start-NativeExecution { gem install fpm ronn }
+                    Start-NativeExecution { gem install fpm -v 1.8.1 }
```
adityapatwardhan: Can we verify that it does not error out if this or higher version is already installed. Or do we need to use ```-​-conservative``` or ```-f```

daxian-dbw: We can do that later, please open an issue for it. For now, we need to unblock our release 😄

adityapatwardhan: Maybe add a comment on why this is added.

Issue #4194, opened 2017-07-06
Calling powershell.exe in PSCore6 on Windows starts Windows PowerShell


SteveL-MSFT:
On Windows, `$env:Path` has Windows PowerShell in the path before `$PSHOME` which means that if you're in PowerShell Core 6 and type `powershell`, it starts Windows PowerShell 5.1 instead of a nested instance of PowerShell Core 6.  This may be unexpected for end users as it seems that one would expect `powershell` to refer to the currently running version of PowerShell.

The immediate impact is that partner teams running existing PowerShell script tests in PowerShell Core 6 that spawn `powershell` will get incorrect results as that process is running Windows PowerShell.

cc @joeyaiello @HemantMahawar 

Steps to reproduce
------------------

```powershell
# within PowerShell Core 6
powershell
$psversiontable
```

Expected behavior
-----------------

```none
PowerShell Core 6
```

Actual behavior
---------------

```none
Windows PowerShell 5.1
```

Environment data
----------------

```powershell
Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.3
OS                             Microsoft Windows 10.0.16225
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

iSazonov:
@SteveL-MSFT Maybe Area-SideBySide?

iSazonov:
If we have some PowerShell versions on system we should think about "default PowerShell" and a way to run specific PowerShell version by `-Version` parameter.

SteveL-MSFT:
Based on some internal discussions, I think we agree that on Windows, we should put $PSHOME first in $env:PATH rather than at the end.

iSazonov:
What about to enhance `-Version`?
```powershell
powershell -Version 5.1
powershell -Version 6.0.0
powershell -Version 6.0.0-Beta.4
powershell -Version UserDefault
powershell -Version SystemDefault
```

SteveL-MSFT:
@iSazonov I think you should open a separate issue to discuss that

iSazonov:
Done #4199

chuanjiao10:
In my opinion, this is the default value question.
In the shortcut, start the directory settings.

joeyaiello:
[...]

PR #4481, opened 2017-08-03
ensure running `powershell` within PowerShell starts instance of currently running PowerShell

SteveL-MSFT:
modify PATH env var at startup so that $PSHOME is in front

Fix https://github.com/PowerShell/PowerShell/issues/4194

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
@@ -135,16 +135,12 @@ namespace Microsoft.PowerShell
 #endif
 
             // put PSHOME in front of PATH so that calling `powershell` within `powershell` always starts the same running version
-            StringBuilder path = new StringBuilder(Environment.GetEnvironmentVariable("PATH"));
-            string pathSeparator = ";";
-            if (!Platform.IsWindows)
-            {
-                pathSeparator = ":";
-            }
+            string path = Environment.GetEnvironmentVariable("PATH");
             string pshome = Utils.DefaultPowerShellAppBase;
-            path.Replace(pathSeparator + pshome, "");
-            path.Insert(0, pshome + pathSeparator);
-            Environment.SetEnvironmentVariable("PATH", path.ToString());
+            if (!path.Contains(pshome))
+            {
+                Environment.SetEnvironmentVariable("PATH", pshome + Path.PathSeparator + path);
+            }
 
             try
             {
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
@@ -134,6 +134,18 @@ namespace Microsoft.PowerShell
             }
 #endif
 
+            // put PSHOME in front of PATH so that calling `powershell` within `powershell` always starts the same running version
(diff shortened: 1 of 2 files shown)


src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -134,6 +134,18 @@ internal sealed partial class ConsoleHost
             }
 #endif
 
+            // put PSHOME in front of PATH so that calling `powershell` within `powershell` always starts the same running version
+            StringBuilder path = new StringBuilder(Environment.GetEnvironmentVariable("PATH"));
+            string pathSeparator = ";";
+            if (!Platform.IsWindows)
+            {
+                pathSeparator = ":";
+            }
```
daxian-dbw: You can use `Path.PathSeparator`.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -134,6 +134,18 @@ internal sealed partial class ConsoleHost
...
+            string pshome = Utils.DefaultPowerShellAppBase;
+            path.Replace(pathSeparator + pshome, "");
```
daxian-dbw: Why is this needed?

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -134,6 +134,18 @@ internal sealed partial class ConsoleHost
...
+            path.Insert(0, pshome + pathSeparator);
```
daxian-dbw: We should check if the first path in `$env:PATH` is `pshome`, and only insert if it's not. Otherwise, the nested powershell instance will have 2 `pshome` paths in `$env:PATH`

daxian-dbw (changes requested):

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
SteveL-MSFT: ok

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
SteveL-MSFT: I thought that the PWD got appended to the end of PATH so I didn't want it on both ends, looks like that isn't the case.  I'll remove.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
SteveL-MSFT: ok

daxian-dbw (approved):
LGTM

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -134,6 +134,14 @@ internal sealed partial class ConsoleHost
             }
 #endif
 
+            // put PSHOME in front of PATH so that calling `powershell` within `powershell` always starts the same running version
```
iSazonov: 1. Is this right for other applications?
2. I would prefer more deterministic `[System.Diagnostics.Process]::GetCurrentProcess().MainModule.FileName`

PR #4482, opened 2017-08-03
Change test to not set PSDefaultParameterValues

JamesWTruher:
Create a local variable for determining whether test should be skipped
Also, change depth of export-clixml to speed up tests

PSDefaultParameterValues did not get set back properly, so we had about 1500 tests which did not get run as expected. Since there's only a single test which needed to be skipped, I changed the logic to use `-skip` instead

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/clixml.tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/clixml.tests.ps1
@@ -58,7 +58,7 @@
 
                 try
                 {
-                    Export-Clixml -LiteralPath $test.testFile -InputObject $test.inputObject -Force
+                    Export-Clixml -Depth 1 -LiteralPath $test.testFile -InputObject $test.inputObject -Force
                 }
                 catch
                 {
@@ -72,7 +72,7 @@
         It "can be created with literal path" {
 
             $filePath = Join-Path $subFilePath 'gps.xml'
-            Export-Clixml -LiteralPath $filePath -InputObject ($gpsList | Select-Object -First 1)
+            Export-Clixml -Depth 1 -LiteralPath $filePath -InputObject ($gpsList | Select-Object -First 1)
 
             $filePath | Should Exist
 
@@ -99,7 +99,7 @@
 
 
             $filePath = Join-Path $subFilePath 'gps.xml'
-            ($gpsList | Select-Object -First 1) | Export-Clixml -LiteralPath $filePath
+            ($gpsList | Select-Object -First 1) | Export-Clixml -Depth 1 -LiteralPath $filePath
 
             $filePath | Should Exist
 
@@ -154,7 +154,7 @@
         }
 
         It "can import from a literal path" {
-            Export-Clixml -LiteralPath $filePath -InputObject $gps
+            Export-Clixml -Depth 1 -LiteralPath $filePath -InputObject $gps
             $filePath | Should Exist
 
             $fileContent = Get-Content $filePath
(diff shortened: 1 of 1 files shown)


test/powershell/Modules/Microsoft.PowerShell.Utility/clixml.tests.ps1
```diff
@@ -195,30 +195,22 @@ Describe "Deserializing corrupted Cim classes should not instantiate non-Cim typ
 
         # Only run on Windows platform.
         # Ensure calc.exe is avaiable for test.
-        if ( !$IsWindows -or ((Get-Command calc.exe 2>$null) -eq $null) )
-        {
-            $orginalDefaultParameters = $PSDefaultParameterValues.Clone()
-            $PSDefaultParameterValues["it:skip"] = $true
-        }
-        else
+        $shouldRunTest = $IsWindows -or ((Get-Command calc.exe -ea SilentlyContinue) -ne $null)
```
PaulHigin: Should be $IsWindows **-and** ((Get-Command ...))

test/powershell/Modules/Microsoft.PowerShell.Utility/clixml.tests.ps1
```diff
@@ -195,30 +195,22 @@ Describe "Deserializing corrupted Cim classes should not instantiate non-Cim typ
...
+        $skipNotWindows = ! $shouldRunTest
```
PaulHigin: I personally like this a lot better.  But with the previous pattern the issue was variable scoping.  It should have been $global:PSDefaultParameters, correct?

test/powershell/Modules/Microsoft.PowerShell.Utility/clixml.tests.ps1
adityapatwardhan: This should be an ```-and```. Otherwise, on windows ```$shouldRunTest``` is always true.

PaulHigin (changes requested):

adityapatwardhan (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/clixml.tests.ps1
daxian-dbw: Yes, it should be $global:PSDefaultParameters

JamesWTruher: i changed it to this way since there was only a single test. There's a couple of patterns, if you have a large number of tests you want to mark skipped, then PSDefaultParameterValues make some sense, but with a small number this can be easier

test/powershell/Modules/Microsoft.PowerShell.Utility/clixml.tests.ps1
JamesWTruher: fixed

JamesWTruher: fixed

PaulHigin (approved):
LGTM

adityapatwardhan (approved):

daxian-dbw (approved):
The PR is submitted from the fork `JamesWTruher`, however the branch `jameswtruher/clixmltestfix` was pushed to powershell/powershell repo (I think by accident), so it triggered 4 build CIs (two PRs and two pushes).
The failed CI build is because AppImage failure, which is already fixed in the release branch so it's OK to ignore this CI failure.

Issue #3931, opened 2017-06-03
globbing for native commands is too agressive


vors:
Steps to reproduce
------------------

Intuitively globbing should not kick-in inside the single-quoted strings.

```
echo '11:1' | grep '.*:.'
```

Expected behavior
-----------------
Works, output is `11:1`, like in bash.

Actual behavior
---------------

```
Cannot find drive. A drive with the name '.*' does not exist.
At line:1 char:1
+ echo '11:1' | grep '.*:.'
+ ~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (.*:String) [], DriveNotFoundException
    + FullyQualifiedErrorId : DriveNotFound 
```

The error is pretty confusing for a unix user.

Workaround
------------

Escape `*` by a backtick in the regex.

Environment data
----------------

```powershell
> $PSVersionTable

Name                           Value                                                                                      
----                           -----                                                                                      
PSVersion                      6.0.0-beta                                                                                 
PSEdition                      Core                                                                                       
BuildVersion                   3.0.0.0                                                                                    
CLRVersion                                                                                                                
GitCommitId                    v6.0.0-beta.2                                                                              
OS                             Darwin 16.6.0 Darwin Kernel Version 16.6.0: Fri Apr 14 16:21:16 PDT 2017; root:xnu-3789....
Platform                       Unix                                                                                       
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                    
PSRemotingProtocolVersion      2.3                                                                                        
SerializationVersion           1.1.0.1                                                                                    
WSManStackVersion              3.0  
```

mklement0:
Just to state it explicitly: globs shouldn't be expanded inside `"..."` (double-quoted strings) either, which currently happens too:

    printf '%s\n' '*'    # should print literal *
    printf '%s\n' "*"    # ditto

Only _unquoted_ tokens should ever be subject to globbing, as in POSIX-like shells.

latkin:
Just noticed this after moving to Beta on OSX. Maybe a recent regression?

It makes using `curl` impossible if you have URL query parameters.

```
> curl 'https://google.com'                                                                                                                                                                                                                                         
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
[...]

Issue #4476, opened 2017-08-03
Unable to access contents in mtp Device


saurav-sahu:
Steps to reproduce
------------------
- Plug in mobile device in `Media Device` mode
- `Set-Location` to `"/run/user/$(bash -c 'echo $UID')/gvfs"`
- Try to navigate to mtp device:
A `Get-ChildItem` shows my device:
```powershell
PS /run/user/1000/gvfs> Get-ChildItem                                                                           


    Directory: /run/user/1000/gvfs


Mode                LastWriteTime         Length Name                                                          
----                -------------         ------ ----                                                          
d-r---         01/01/70   5:30 AM                mtp:host=%5Busb%3A002%2C003%5D
```

Expected behavior
-----------------
Read only access to all files and folders on the device 

Actual behavior
---------------

```powershell
PS /run/user/1000/gvfs> cd ./mtp:host=%5Busb%3A002%2C003%5D/                                                    
cd : Cannot find drive. A drive with the name 'run/user/1000/gvfs/mtp' does not exist.
At line:1 char:1
+ cd ./mtp:host=%5Busb%3A002%2C003%5D/
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (run/user/1000/gvfs/mtp:String) [Set-Location], DriveNotFoundExc 
   eption
    + FullyQualifiedErrorId : DriveNotFound,Microsoft.PowerShell.Commands.SetLocationCommand
```

Environment data
----------------

```powershell
PS /home/user> lsb_release -a                                                                                   
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 17.04
Release:	17.04
Codename:	zesty


PS /home/user> $PSVersionTable
Name                           Value                                                                           
----                           -----                                                                           
PSVersion                      6.0.0-alpha                                                                     
PSEdition                      Core                                                                            
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                         
BuildVersion                   3.0.0.0                                                                         
GitCommitId                    v6.0.0-alpha.16                                                                 
CLRVersion                                                                                                     
WSManStackVersion              3.0                                                                             
PSRemotingProtocolVersion      2.3                                                                             
SerializationVersion           1.1.0.1

PS /home/user> screenfetch                                                                             
                          ./+o+-       sahu@Dell-Inspiron
[...]

PR #4489, opened 2017-08-04
Fix globbing for native commands

BrucePay:
Fix for #3931. With this change, globbing for native commands will not longer be done if the argument string is quoted (single or double quotes).

### DIFF
--- a/src/System.Management.Automation/engine/CommandParameter.cs
+++ b/src/System.Management.Automation/engine/CommandParameter.cs
@@ -15,12 +15,14 @@ namespace System.Management.Automation
     {
         private class Parameter
         {
+            internal Ast ast;
             internal IScriptExtent extent;
             internal string parameterName;
             internal string parameterText;
         }
         private class Argument
         {
+            internal Ast ast;
             internal IScriptExtent extent;
             internal object value;
             internal bool splatted;
@@ -73,12 +75,21 @@ namespace System.Management.Automation
             get { return _parameter != null ? _parameter.extent : PositionUtilities.EmptyExtent; }
         }
 
+        internal Ast ParameterAst {
+            get { return _parameter != null ? _parameter.ast : null; }
+        }
+
         /// <summary>
         /// The extent of the optional argument, if one was specified.
         /// </summary>
         internal IScriptExtent ArgumentExtent
         {
             get { return _argument != null ? _argument.extent : PositionUtilities.EmptyExtent; }
+        }
+
+        internal Ast ArgumentAst
+        {
+            get { return _argument != null ? _argument.ast : null; }
         }
 
         /// <summary>
@@ -141,10 +152,12 @@ namespace System.Management.Automation
         /// <summary>
         /// Create a parameter when no argument has been specified.
         /// </summary>
+        /// <param name="ast">The ast for this parameter</param>
         /// <param name="extent">The extent in script of the parameter.</param>
         /// <param name="parameterName">The parameter name (with no leading dash).</param>
         /// <param name="parameterText">The text of the parameter, as it did, or would, appear in script.</param>
-        internal static CommandParameterInternal CreateParameter(
+        internal static CommandParameterInternal CreateParameterWithAst(
+            Ast ast,
             IScriptExtent extent,
             string parameterName,
             string parameterText)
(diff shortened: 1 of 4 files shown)


test/powershell/Language/Scripting/NativeExecution/NativeUnixGlobbing.Tests.ps1
```diff
@@ -38,6 +38,12 @@ Describe 'Native UNIX globbing tests' -tags "CI" {
     It 'Should return the original pattern if there are no matches' {
         /bin/echo $TESTDRIVE/*.nosuchfile | Should Match "\*\.nosuchfile$"
     }
+    It 'Should not expand double-quoted strings' {
+        /bin/echo "*" | Should BeExactly "*"
+    }
+    It 'Should not expand single-quoted strings' {
+        /bin/echo '*' | Should BeExactly "*"
+    }
```
vors: Please add a test for something along the lines of `echo '11:1' | grep '.*:.'`. The `:` part is important too to avoid interpreting it as a drive latter

BrucePay: This change only impacts expansion of wildcard characters. ':' isn't a wildcard character so I don't see a reason to add this test.

SteveL-MSFT: I think the colon is an [issue](https://github.com/PowerShell/PowerShell/issues/4476) in the FileSystem Provider and should probably be in those tests

src/System.Management.Automation/engine/CommandParameter.cs
```diff
@@ -15,12 +15,14 @@ internal sealed class CommandParameterInternal
     {
         private class Parameter
         {
+            internal Ast ast;
             internal IScriptExtent extent;
```
lzybkr: `ast` makes storing the `extent` no longer necessary, right?

lzybkr: I mean - `extent` should always be `ast.Extent`, so we could remove the field and use a property instead.

src/System.Management.Automation/engine/CommandParameter.cs
```diff
@@ -15,12 +15,14 @@ internal sealed class CommandParameterInternal
...
             internal string parameterName;
             internal string parameterText;
         }
         private class Argument
         {
+            internal Ast ast;
             internal IScriptExtent extent;
```
lzybkr: The same applies here regarding `ast` and `extent`.

src/System.Management.Automation/engine/CommandParameter.cs
BrucePay: This API is used in many places where the AST is not available e.g. the PowerShell API, F&O. In those cases dummy extents are explicitly passed in. So both the Ast and extent members are required.

src/System.Management.Automation/engine/CommandParameter.cs
BrucePay: Same response as above - the API is used in places where there is no `ast`.

test/powershell/Language/Scripting/NativeExecution/NativeUnixGlobbing.Tests.ps1
BrucePay: Steve - I'm not sure i understand your comment. I'm not testing the file provider here. I'm just making sure that quoting suppresses globbing. Specific issues with the filesystem provider are orthogonal.

src/System.Management.Automation/engine/CommandParameter.cs
lzybkr: We could introduce a dummy ast for the dummy position.

My thinking - if the normal case is a script, we optimize for that case. If the field is typically redundant, we use less memory, generate less code (the push of the extra parameter), and reduce the number of fields gc needs to visit.

test/powershell/Language/Scripting/NativeExecution/NativeUnixGlobbing.Tests.ps1
SteveL-MSFT: Didn't notice that the test didn't involve cmdlets

src/System.Management.Automation/engine/CommandParameter.cs
BrucePay: I think a dummy AST is overkill. Reworking the API to use dummy extents if the AST is null, etc. is probably the best solution but that's out of scope for this issue.

lzybkr: Adding a new parameter will bloat the **generated** code significantly - this is an extra parameter for every argument on every command invocation.  We should avoid that if there are alternatives.

lzybkr:
@BrucePay - please rebase.

lzybkr:
@BrucePay - please rebase, and look at the log before force pushing to ensure only your commits appear between master and your branch.

SteveL-MSFT:
@lzybkr looks like the commits got cleaned up and the CI is passing, any other concerns?

src/System.Management.Automation/engine/CommandParameter.cs
lzybkr: I'm still not happy with storing 2 references a typical script has a ton of these instances.

What if we use something like:
```c#
    private object context;  // Instead of storing Extent and Ast
    // ...
    internal Ast ParameterAst { return _parameter?.context as Ast; }
    internal IScriptExtent ParameterExtent {
        return _parameter?.context as IScriptExtent ?? (_parameter?.context as Ast)?.Extent;
    }
    // etc.
```

rkeithhill:
Folks, we could **REALLY** use this fix!  We have scripts that broke when we recently updated from an old alpha to beta.8.  ATM we've found no workaround other than to fall back to Bash.

lzybkr:
Closing in favor of #5188.

Issue #4272, opened 2017-07-17
Remove use of unsupported APIs


TravisEz13:
Remove use of unsupported APIs.  I'll file issues to add the functionality back using supported APIs.

TravisEz13:
PR #4303 and #4302  are related to this

PR #4496, opened 2017-08-04
Use supported API to set CAPID in the SACL

TravisEz13:
Addresses #4272

### DIFF
--- a/src/System.Management.Automation/security/nativeMethods.cs
+++ b/src/System.Management.Automation/security/nativeMethods.cs
@@ -1786,7 +1786,7 @@ namespace System.Management.Automation.Security
             uint nAclLength,
             uint dwAclRevision);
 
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/security/nativeMethods.cs
```diff
@@ -1786,8 +1786,8 @@ internal struct CENTRAL_ACCESS_POLICY
             uint nAclLength,
             uint dwAclRevision);
 
-        [DllImport("ntdll.dll", CharSet = CharSet.Unicode)]
-        internal static extern uint RtlAddScopedPolicyIDAce(
+        [DllImport("kernel32.dll", CharSet = CharSet.Unicode)]
```
daxian-dbw: Shall we use the API set DLL?
See http://onecoresdk/APISets/api-ms-win-security-base-l1-2-0.htm

daxian-dbw (approved):

PR #4500, opened 2017-08-04
fix shortcut to start in %userprofile% instead of the install folder

SteveL-MSFT:


SteveL-MSFT:
Realized this doesn't have the intended effect I want.  It'll hardcode the shortcut to the profile of whomever ran the installer.  Closing until I can find solution.

PR #4502, opened 2017-08-04
Add Missing Assemblies to TPA List for Pwrshplugin.dll

mirichmo:
This PR updates the TPA list for pwrshplugin.dll to reflect the recent additions to the DLL. It also adds support for VS 2017 compilation by adding a work-around for and include path that gets corrupted by VS 2017.


daxian-dbw (approved):

src/powershell-native/nativemsh/pwrshcommon/CMakeLists.txt
```diff
@@ -10,6 +10,18 @@ add_library(pwrshcommon
...
+# VS 2017 corrupts the INCLUDE environment variable when processing the csproj
+# file generated for this template. The netfxsdk path is shortened from
+# C:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\Include\um
+# to
+# Include\um
+# This prevents msbuild from locating corerror.h and throws a resolution error
+# during compilation. We work around the problem by force-including the
+# contents of the environment variable before it gets corrupted.
+#
+target_include_directories(pwrshcommon PUBLIC $ENV{INCLUDE})
```
daxian-dbw: I assume adding this would still work on a machine with VS15 installed only, right?

mirichmo: That is the scenario where I ran it. My system has VS 2015 and VS 2017 installed side-by-side. I have not verified it on a system that only has VS 2017. As long as vcvarsall.bat generates the correct INCLUDE environment variable, it should work though.

daxian-dbw: > I have not verified it on a system that only has VS 2017.

I'm not worried about a system that only has VS 2017, but a system that only has VS 2015. Our current release build pipeline in VSTS only installs VS 2015. I want to make sure we don't break there.

daxian-dbw: OK. I checked out this PR and verified that it builds fine with VS 2015 installed only and the generated plugin binary works.

daxian-dbw (approved):

PR #4506, opened 2017-08-05
update brew before installing powershell

### LINKED ISSUE #4505: Install-powershell.sh on macOS can install very old version of powershell
in investigating #4471, I found that `install-powershell.sh` was installing an alpha version of PowerShell because it was not updating brew before installing.
Examples:
https://travis-ci.org/richardszalay/pshosts/jobs/260027319#L94
https://travis-ci.org/PowerShell/PowerShell/jobs/261189298#L98

TravisEz13:
fixes #4505 
Thanks, @richardszalay for spotting the issue

### DIFF
--- a/tools/installpsh-osx.sh
+++ b/tools/installpsh-osx.sh
@@ -134,6 +134,12 @@ if [[ ! -d $(brew --prefix cask) ]]; then
     fi
 fi
 
+# Suppress output, it's very noisy on travis-ci
+echo "Refreshing Homebrew cache..."
+if ! brew update >/dev/null; then
+    echo "ERROR: Refreshing Homebrew cache failed..." >&2
+fi
+
 if ! hash powershell 2>/dev/null; then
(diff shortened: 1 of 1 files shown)


tools/installpsh-osx.sh
```diff
@@ -134,6 +134,12 @@ if [[ ! -d $(brew --prefix cask) ]]; then
     fi
 fi
 
+# Suppress output, it's very noisy on travis-ci
+echo "Refreshing Homebrew cache..."
+if ! brew update >/dev/null; then
+    echo "ERROR: Refreshing Homebrew cache failed..." >&2
```
daxian-dbw: Shall we stop the script at if the update fails?

adityapatwardhan (approved):
LGTM

tools/installpsh-osx.sh
TravisEz13: Resolved

PR #4517, opened 2017-08-07
Only push to NuGet for Full Build (cron)

TravisEz13:
Currently, every merge is also publishing.

### DIFF
--- a/tools/travis.ps1
+++ b/tools/travis.ps1
@@ -184,10 +184,10 @@ else
         foreach($package in $packages)
         {
             # Publish the packages to the nuget feed if:
-            # 1 - It's a fullBuild (already checked, for not a PR)
+            # 1 - It's a Daily build (already checked, for not a PR)
             # 2 - We have the info to publish (NUGET_KEY and NUGET_URL)
             # 3 - it's a nupkg file
-            if($isFullBuild -and $env:NUGET_KEY -and $env:NUGET_URL -and [system.io.path]::GetExtension($package) -ieq '.nupkg')
+            if($isDailyBuild -and $env:NUGET_KEY -and $env:NUGET_URL -and [system.io.path]::GetExtension($package) -ieq '.nupkg')
             {
                 log "pushing $package to $env:NUGET_URL"
                 Start-NativeExecution -sb {dotnet nuget push $package --api-key $env:NUGET_KEY --source "$env:NUGET_URL/api/v2/package"} -IgnoreExitcode
--- a/tools/travis.ps1
+++ b/tools/travis.ps1
@@ -184,7 +184,7 @@ else
         foreach($package in $packages)
         {
             # Publish the packages to the nuget feed if:
-            # 1 - It's a fullBuild (allready checked, for not a PR)
+            # 1 - It's a fullBuild (already checked, for not a PR)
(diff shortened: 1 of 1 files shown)


tools/travis.ps1
```diff
@@ -181,7 +181,11 @@ else
         $packages += Start-PSPackage  @packageParams -Type AppImage
         foreach($package in $packages)
         {
-            if($env:NUGET_KEY -and $env:NUGET_URL -and [system.io.path]::GetExtension($package) -ieq '.nupkg')
+            # Publish the packages to the nuget feed if:
+            # 1 - It's a fullBuild (allready checked, for not a PR)
```
adityapatwardhan: Typo in allready -> already

adityapatwardhan (changes requested):
I found another issue, we seem to be updating the badge on full build, including the one with [Feature]. We need a way to identify a CRON build and then do nuget publish and badge update.

tools/travis.ps1
```diff
@@ -181,7 +181,11 @@ else
...
+            # 2 - We have the info to publish (NUGET_KEY and NUGET_URL)
+            # 3 - it's a nupkg file
+            if($isFullBuild -and $env:NUGET_KEY -and $env:NUGET_URL -and [system.io.path]::GetExtension($package) -ieq '.nupkg')
```
adityapatwardhan: This also published if it is [Feature] run?

adityapatwardhan (changes requested):

tools/travis.ps1
TravisEz13: Only if we merge a feature... that is equivalent to a daily build, I'm fine with it.  If it's a PR, it won't publish.   And if you squash merge, you will likely, remove the `[Feature]` from the commit message.

TravisEz13: actually, a true merge would change the description of the last commit message too.  Only merging to master using the rebase option would have this possible effect.

TravisEz13:
The badge issue is being fixed in #4536

tools/travis.ps1
```diff
@@ -181,7 +181,11 @@ else
         $packages += Start-PSPackage  @packageParams -Type AppImage
         foreach($package in $packages)
         {
-            if($env:NUGET_KEY -and $env:NUGET_URL -and [system.io.path]::GetExtension($package) -ieq '.nupkg')
+            # Publish the packages to the nuget feed if:
+            # 1 - It's a fullBuild (already checked, for not a PR)
+            # 2 - We have the info to publish (NUGET_KEY and NUGET_URL)
+            # 3 - it's a nupkg file
+            if($isFullBuild -and $env:NUGET_KEY -and $env:NUGET_URL -and [system.io.path]::GetExtension($package) -ieq '.nupkg')
```
daxian-dbw: After merging a PR to master, if the commit message happen to have `[Feature]` tag, we will publish the nuget package as well. Is that desired? Maybe we should use `$isDailyBuild` instead.

TravisEz13: Rebased, and updated.

daxian-dbw (approved):

adityapatwardhan (approved):
LGTM

PR #4522, opened 2017-08-07
Remove last-modified from headers from travis CI script

JamesWTruher:
It causes a no such resource error when used

When invoke-webrequest is invoked, the call fails with a resource not found error and the badge is not updated. The documentation is pretty spotty with regard to what that resource (last-modified) could have done. It should be enough to set the cache-control header value to let browsers know to refresh the badge value rather than pulling it from cache.

this change should enable the PowerShell main page to report the proper state of the daily build on travis.

### DIFF
--- a/tools/travis.ps1
+++ b/tools/travis.ps1
@@ -50,7 +50,6 @@ function Set-DailyBuildBadge
     $headers = @{
         "x-ms-date"      = $now
         "cache-control"  = "no-cache"
-        "last-modified"  = $now
         "x-ms-blob-type" = "BlockBlob"
         "x-ms-version"   = "$headerDate"
     }


tools/travis.ps1
```diff
@@ -50,7 +50,6 @@ function Set-DailyBuildBadge
     $headers = @{
```
adityapatwardhan: $now is not used any more. We should remove it.

adityapatwardhan (changes requested):

adityapatwardhan:
Can we add more description in the PR as to how this change fixes the issue.

tools/travis.ps1
TravisEz13: It is still being used in `x-ms-date`

TravisEz13 (approved):

tools/travis.ps1
JamesWTruher: it's still used - two lines higher

TravisEz13:
Also, AppVeyor failed in filesystem provider tests.  Can we investigate?

adityapatwardhan:
@JamesWTruher Please also add the issue it fixes.

JamesWTruher:
it looks like new test code went in on 8/2 from jeffbi

JamesWTruher:
@adityapatwardhan updated description

JamesWTruher:
is there a way to restart the job in appveyor?

adityapatwardhan:
@JamesWTruher Restarted the AppVeyor build.

adityapatwardhan (approved):

Issue #4475, opened 2017-08-03
SSH KeyFilePath parameter not working 


luzlab:
When I try running New-PSSession or Enter-PSSession using a public key, powershell either:
1. When passed a valid KeyFilePath, complains that it doesn't exist and continues gracefully to password-based authentication.
2. When passed a bogus KeyFilePath, throws an ItemNotFoundException

Steps to reproduce
------------------
```powershell
> New-PSSession -UserName "xx" -KeyFilePath "/fakepath" -Hostname " xx.xx.xx.xx"
New-PSSession : Cannot find path 'doesntexist' because it does not exist.
At line:1 char:1
+ New-PSSession -UserName "Administrator" -KeyFilePath "doesntexist" -H ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (doesntexist:String) [New-PSSession], ItemNotFoundException
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.NewPSSessionCommand

> New-PSSession -UserName "xx" -KeyFilePath "~/.ssh/id_rsa" -Hostname " xx.xx.xx.xx"
Warning: Identity file "/Users/cquinonez/.ssh/workgangkey" not accessible: No such file or directory.

```

Expected behavior
-----------------
```powershell
>  New-PSSession -UserName "xx" -KeyFilePath "/fakepath" -Hostname " xx.xx.xx.xx"
Warning: Identity file "/Users/cquinonez/.ssh/workgangkey" not accessible: No such file or directory.
>  New-PSSession -UserName "xx" -KeyFilePath "<validKeyFilePath>" -Hostname " xx.xx.xx.xx"

 Id Name            ComputerName    ComputerType    State         ConfigurationName     Availability
 -- ----            ------------    ------------    -----         -----------------     ------------
  1 SSH1            xx.xx.xx.xx     RemoteMachine   Opened        DefaultShell             Available
```

Actual behavior
---------------
![untitled](https://user-images.githubusercontent.com/9924835/28905960-ee6d962a-77c8-11e7-9cb5-e8b306345d03.png)

I've verified that `Test-Path` also correctly detects the key file as well.

Environment data
----------------
```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.0.0-beta
[...]

Issue #3904, opened 2017-06-01
Create SSH remoting tests


PaulHigin:
We currently don't have SSH remoting tests because there is no way to automatically set up SSH.  Setting up SSH will be addressed by Enable-SSHRemoting (Issue #3616).  And once this is working we need to begin adding SSH remoting tests.

PaulHigin:
Un-assigning since I am not actively working on this.

PR #4529, opened 2017-08-08
Fixes SSH Remoting KeyFilePath Parameter

PaulHigin:
This change addresses issue #4475.

When creating the SSH child process for remoting the KeyFilePath parameter path was enclosed in double quotes to handle potential space characters in the file path.  The problem was that the new SSH child process creation code was leaving the quote characters in the key filepath causing SSH to fail to find the key file.  In addition KeyFilePath paths with space characters were not being handled correctly.

With this change the KeyFilePath parameter now works as expected:
```powershell
PS > New-PSSession -HostName Machine1 -UserName User1 -KeyFilePath c:\users\User1\Keys\User1Key

PS > New-PSSession -HostName Machine2 -UserName User2 -KeyFilePath "c:\users\User2\My Keys\User2Key"
```
This worked correctly on Windows and only failed on Linux systems using the new SSH process creation code.

---
I also noticed and fixed a problem where trying to back out of a SSH remoting password prompt resulted in a hang (until the 1 minute protocol timeout occurred).
```powershell
PS > New-PSSession -HostName Machine1 -UserName User1
User1@Machine1's password:  
Ctrl+C
```
The problem was that the client side connection did not end.  Fix is to add CloseAsync override to close the connection.


src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -2217,7 +2217,35 @@ private static string[] ParseArgv(ProcessStartInfo psi)
...
+            for (int i=0; i<argsLength; )
+            {
+                var iStart = i;
+         
+                switch (argsToParse[i])
+                {
+                    case '"':
+                        // Special case for arguments within quotes
+                        // Just return argument value within the quotes
+                        while ((++i < argsLength) && argsToParse[i] != '"') { };
```
mirichmo: Is there enforcement for the string ending in a '"'? Does that happen in a different layer?

PaulHigin: No enforcement here.  This expects the argument string to be correctly formatted.  If it is not then the argument list is wrong and any errors are generated when the process is created.

This is currently scoped internally so we currently have complete control in how it is used.  However, I did intend this to be robust against malformed strings (empty, single quote, unmatched quotes, single space, only spaces, etc.).  If you see something I missed please let me know.

mirichmo: My initial concern regarded unmatched double quotes. Thinking about it some more, two other things came to mind:
1. Strings that use single quotes
2. Argument lists that include other commands. Do we do validation on it? I'm wondering about something like this: `arg1 arg2; rm -rf *`

PaulHigin: Single quotes are not supported and not needed.  It is something we could add if we need it in the future.  Argument validation is not intended to be performed here.  This just parses space delimited strings with a special case for double quoted arguments.

I just tried to make sure it doesn't blow up with malformed strings and cause an access violation or null reference exception.

mirichmo (approved):

PaulHigin:
@adityapatwardhan 
Can this be merged?  Thanks.

adityapatwardhan:
@PaulHigin Can we add tests for this?

PaulHigin:
#3904

PaulHigin:
Thanks!

Hema-dell:
HI ,
I have installed PoSH6 and 7 on Windows 10 box, and PoSh6 on linux box, trying to do remote throush PsSession using -keyfilepath, followed instructions for ssh as provided in  https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse, after following the instructions and providing this command : New-PSSession -HostName root@x.x.x.x:22 -KeyFilePath C:\Users\dell\.ssh\authorized_keys  i get prompt for password , i have copied authorized_keys into ./ssh folder of linux box too!!! why is it prompting for password?

PaulHigin:
@Hema-dell  It sounds like your sshd_config configuration file still allows password authentication and should be disabled.  FYI, I have a module with 'Enable-SSHRemomting' cmdlet that I have been working on.  It is not been published by you can experiment with it if you like:
https://gist.github.com/PaulHigin/b12b552b12dc4e8c940efe4f402e5362

We decided not to touch general SSH settings, and assume that the user has it set up as needed.  The 'Enable-SSHRemoting' cmdlet merely updates the sshd_config file with subsytem entry that becomes a PowerShell SSH remoting end point, so that the machine can receive connections.

Hema-dell:
Hi Paul!!,
I tried the above steps as shared in the link by you ,however i am still unable to do pssession using -keyfilepath parameter!!!!, Need your help on this!!!!.

Thanks,
Hema-dell

PR #4536, opened 2017-08-09
Fix the hang issue in Travis CI build triggered by the [Feature] tag

daxian-dbw:
Some class related parser tests may cause Travis build to hang if they are running with CrossGen'ed assemblies. This is an issue to the CrossGen utility, and it's tracked by https://github.com/dotnet/coreclr/issues/9745.

To work around the issue, we made the following changes to our build:
1. For pull request and push commit, build without '-CrossGen' and run the parsing tests
2. For daily build, build with '-CrossGen' but don't run the parsing tests

The recent `[Feature]` tag support allows us to run the full build in PR or Commit-Push CI builds. The parser tests should run in those `[Feature]` tag builds, so we need to make sure the `[Feature]` tag builds don't use `-CrossGen` before running the tests.

Two other changes:
1. Skip inheritance tests during daily build.
2. Update badge only in daily build.


TravisEz13 (approved):

adityapatwardhan (approved):
LGTM

daxian-dbw:
The failures in AppVeyor are all `*-PSSessionConfiguration` related, not caused by this PR.

Issue #2504, opened 2016-10-19
Tests should not rely on external websites which lead to random failures


vors:
Today https://httpbin.org/ is down the whole day.
It caused two daily builds to fail with 43 tests for `Invoke-WebRequest` and `Invoke-RestMethod`.

https://ci.appveyor.com/project/PowerShell/powershell-f975h/build/6.0.0-alpha.10-26
https://ci.appveyor.com/project/PowerShell/powershell-f975h/build/6.0.0-alpha.11-27

We should replace httpbin with eaver our own hosted version or localhost server, i.e.
https://github.com/Runscope/httpbin

localhost server will introduce unnecessary testing dependencies, not sure which one would be a better solution.

cc @daxian-dbw @Francisco-Gamino

daxian-dbw:
I prefer to host the service in cloud ourselves. I got it hosted at Heroku, but still not clear if the free plan would be sufficient.

SteveL-MSFT:
Move tests to use httplistener where we have more control and also suppress progress for the tests

SteveL-MSFT:
Also review for other external websites being used beyond httpbin.org

markekraus:
@TravisEz13 I plan to move all of the HTTPS tests to WebListener (#4719).  

For HTTP (no-S) tests that are currently pointed to httpbin I plan to pick the one (possibly /get), create the same functionality in WebListener, and move all the tests that use it to WebListener. Before I begin, though, I'd like to just confirm that this is the right path to take. I figured I would start moving one httpbin target URL at a time and then work on moving the HttpListener tests one URL at a time. The end goal being that all HTTP/HTTPS tests will run against WebListener and HttpListener can be retired.

SteveL-MSFT:
@markekraus sounds right to me.  Thanks!

markekraus:
@SteveL-MSFT How should I do the PR's for these? Should I create a One PR for adding the WebListener functionality and another for switching the tests it replaces? Or just lump the WebListener functionality in with the test changes in a single PR?

SteveL-MSFT:
@markekraus I think it would be better to have both the WebListener changes and test code changes together for each URL replacement so the reviewers can understand what's happening on both the client side and server side

markekraus:
after #5665 is merged, there is only 4 tests which rely on [http://http.lee.io/method](http://http.lee.io/method),  all other tests rely on WebListener or HttpListener.

TravisEz13:
[...]

PR #4546, opened 2017-08-10
[Feature] Add Certificate Authentication Support for WebCmdlets

### LINKED ISSUE #4544: Invoke-RestMethod and Invoke-WebRequest don't support client certificate authentication with certificate files
`Invoke-RestMethod` and `Invoke-WebRequest` requires the certificate specified by _-Certificate_ to be imported in the certificate store. Both methods do not allow using certificates from files. This breaks client certificate authentication on non-Windows platforms.

**Proposal**: Support the use of certificate files for the client authentication by adding a new parameter _-CertificateFile_ or by using the x509 object from `Get-PfxCertificate` directly.

This was tested on Ubuntu 16.04 with PowerShell 6.0.0-beta.5 and on Windows with PowerShell 5.0.

Ref: https://github.com/PowerShell/PowerShell/issues/473 and https://github.com/PowerShell/PowerShell/issues/1865

markekraus:
Fixes #4544

`HttpClientHandler` now supports client certificates. Enabling client certificates in Core for `Invoke-WebRequest` and `Invoke-RestMethod` makes it possible to do certificate authentication on Linux.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
@@ -171,7 +171,8 @@ namespace Microsoft.PowerShell.Commands
 
             if (SkipCertificateCheck)
             {
-                handler.ServerCertificateCustomValidationCallback = HttpClientHandler.DangerousAcceptAnyServerCertificateValidator; 
+                handler.ServerCertificateCustomValidationCallback = HttpClientHandler.DangerousAcceptAnyServerCertificateValidator;
+                handler.ClientCertificateOptions = ClientCertificateOption.Manual;
             }
 
             // This indicates GetResponse will handle redirects.
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
@@ -164,13 +164,10 @@ namespace Microsoft.PowerShell.Commands
                 handler.Proxy = WebSession.Proxy;
             }
 
-            /*
-            TODO: HttpClientHandler will support client certificate in RTM
-            See https://github.com/dotnet/corefx/issues/7623 for more details.
             if (null != WebSession.Certificates)
             {
-                handler.ClientCertificates = WebSession.Certificates;
-            }*/
+                handler.ClientCertificates.AddRange(WebSession.Certificates);
+            }
 
             if (SkipCertificateCheck)
             {
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -296,7 +296,6 @@ function ExecuteWebRequest
     return $result
 }
 
-
 function GetSelfSignedCert {
     <#    
         .NOTES
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -917,7 +917,13 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         $response.Content.UserAgent | Should Match $Pattern
     }
 
-    It "Verifies Invoke-WebRequest Certificate Authentication Fails without -Certificate"{
+    #endregion SkipHeaderVerification Tests
+
+    #region Certificate Authentication Tests
+
+    # Test pending creation of native test solution 
+    # https://github.com/PowerShell/PowerShell/issues/4609
+    It "Verifies Invoke-WebRequest Certificate Authentication Fails without -Certificate" -Pending {
         $command = 'Invoke-WebRequest https://prod.idrix.eu/secure/'
         $result = ExecuteWebCommand -command $command
         ValidateResponse -response $result
@@ -925,7 +931,9 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         $result.Output | Should Match ([regex]::Escape('Error: No SSL client certificate presented'))
     }
 
-    It "Verifies Invoke-WebRequest  Certificate Authentication Successful with -Certificate" {
+    # Test pending creation of native test solution 
+    # https://github.com/PowerShell/PowerShell/issues/4609
+    It "Verifies Invoke-WebRequest  Certificate Authentication Successful with -Certificate" -Pending {
         $Certificate = GetSelfSignedCert
         $command = 'Invoke-WebRequest https://prod.idrix.eu/secure/ -Certificate $Certificate'
         $result = ExecuteWebCommand -command $command
@@ -934,7 +942,7 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         $result.Output.Content | Should Match ([regex]::Escape('SSL Authentication OK!'))
     }
 
-    #endregion SkipHeaderVerification Tests
+    #endregion Certificate Authentication Tests
 
     #region charset encoding tests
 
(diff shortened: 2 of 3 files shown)


test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -271,6 +271,89 @@ function ExecuteRequestWithCustomUserAgent {
     return $result
 }
 
+function GetSelfSignedCert {
```
TravisEz13: We already have a common module to add certificates here:
https://github.com/PowerShell/PowerShell/blob/master/test/powershell/Modules/Microsoft.PowerShell.Security/certificateCommon.psm1

We shouldn't duplicate the code.

markekraus: @TravisEz13 I tested with that. But the cert in `New-GoodCertificate` does not contain the proper key usage to be used for SSL/TLS Auth. It is issued for only Document Encryption. I was afraid if I change the cert in `certificateCommon.psm1` I might be break tests that depend on that restriction. Thus, I added a new function in  `WebCmdlets.Tests.ps1` instead. What do you suggest?

TravisEz13: Move the module to a common location and add your function to it, with descriptions of each cert (at least what you learned.)  At the minimum, file an issue that this work needs to be done to avoid future duplication of effort.  Even then, I would suggest adding the comments to each now.

markekraus:
@TravisEz13 I think what may need to be done is issue a self-signed cert that has all the key usage and enhanced key usages required by all the tests, a kind of one-cert-to-rule-them-all and then also move the module to a common area etc. 

For now I have added comments.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -803,6 +886,23 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         $response.Content.UserAgent | Should Match $Pattern
     }
 
+    It "Verifies Invoke-WebRequest Certificate Authentication Fails without -Certificate"{
+        $command = 'Invoke-WebRequest https://prod.idrix.eu/secure/'
```
dantraMSFT: I suggest you check to see if you can mock using WebRequest.RegisterPrefix instead of relying on an external site.  We find external sites make the tests unreliable.  See https://github.com/PowerShell/PowerShell/issues/2504

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -164,13 +164,10 @@ internal virtual HttpClient GetHttpClient(bool handleRedirect)
...
             }
 
-            /*
-            TODO: HttpClientHandler will support client certificate in RTM
-            See https://github.com/dotnet/corefx/issues/7623 for more details.
             if (null != WebSession.Certificates)
             {
-                handler.ClientCertificates = WebSession.Certificates;
-            }*/
+                handler.ClientCertificates.AddRange(WebSession.Certificates);
```
dantraMSFT: Suggest explicitly setting ClientCertificateOptions to ClientCertificateOption.Manual.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
TravisEz13: Especially introducing a dependency on a new site.

markekraus: I agree. But I lack the skill to implement this. I would need guidance or assistance.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
markekraus: added.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: It seems @SteveL-MSFT have plans to add HTTPS support in our test `HttpListener.psm1`

markekraus: @iSazonov awesome. that would be better. It will probably also need  some way to ensure the client authentication cert is received by the listener too or to enforce certificate authentication on the listener so it can fail if one is not provided.

@dantraMSFT  I'm not entirely sure `WebRequest.RegisterPrefix` will work on Core. Unless I'm missing something, I don't think it would affect anything being done with the new classes under `System.Net.Http` in use in Core. I was able to [make a mockup](https://gist.github.com/markekraus/624c7a4969c013e66c357a2d8adeeafc) with it, but it only works on Windows PowerShell (tested on 5.1). Unless I'm going totally in the wrong direction?

dantraMSFT: You're right, that alone won't solve it.  I also looked at doing a quick change to the the httplistener side but, as written, the client cert is not visible to the listener.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
adityapatwardhan: @markekraus Can you address this feedback please?

markekraus: I did [here](https://github.com/PowerShell/PowerShell/pull/4546#issuecomment-321892193) and with [305c048](https://github.com/PowerShell/PowerShell/pull/4546/commits/305c048399d7a2119c63a717eb05bb312fe78a97). Was that insufficient?

adityapatwardhan: Sorry missed that.

adityapatwardhan:
@TravisEz13 @dantraMSFT Can you have another look please?

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: @markekraus - I don't have a quick solution to mocking the call; however, I still question the use of the link (https://prod.idrix.eu/secure/).  I found a reference to it on stackoverflow but I cannot find any information on the parent site indicating its for public use.  Have I missed something?

markekraus: I couldn't find anything written on the site. But, it appears that @idrassi is the owner of https://prod.idrix.eu/secure/ and was the same one who replied on that stackoverflow post with those links. I can try reaching out to them for written consent if needed.

iSazonov: Can we use local CI web server? https://www.appveyor.com/docs/deployment/agent/

markekraus: @iSazonov  I think you meant [This](https://www.appveyor.com/docs/services-databases/#internet-information-services). There would probably have to be a separate solution for AppVeyor and [Travis CI](https://docs.travis-ci.com/user/gui-and-headless-browsers/#Starting-a-Web-Server). It would probably be best if there was some kind of cross-platform solution that worked regardless of CI environment.

TravisEz13: @markekraus Have you looked to see if there is some Linux (so it would work on both platforms) docker container that would work?  Perhaps ASP.NET core in Linux should be familiar to most people using this platform.

TravisEz13: I think we need some explicit statement linked to the site that we can use the site in our tests.

TravisEz13 (changes requested):
We need to find some solution to deal with https://prod.idrix.eu/secure/

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @TravisEz13 unfortunately, I'm a novice with Docker and have no experience with ASP.NET. I could look into it, but It will take a considerable amount of time to figure out what all needs to be done. I'm not opposed to looking into it, but, perhaps there is some way to engage someone else for help on this? I'm not really sure how to proceed.

iSazonov: Maybe open a tracking Issue and don't block the PR?

TravisEz13: I'm open to that, but I don't think we should be regularly 
 running the tests without doing one one the other either:

1. Get explicit permission to use the site for automated tested
1. find an alternative
1. mark the tests a pending until the issue is resolved.

markekraus: @TravisEz13 Ok. Let me look into docker/aspnetcore. If I can't make heads or tails of it, I will mark the test pending and create an issue. 

On the docker topic, would it be best to include a subfolder with a dockerfile and whatever assets are needed, have the test code run the docker build, run the container, do the relevant test, and tear down the container? Or or there some better way to do that? I imagine that could be a rather slow test.

TravisEz13: It would definitely make these slow tests and I think feature tests.  There are some tests that use docker here:  https://github.com/PowerShell/PowerShell/tree/master/test/docker/networktest   But those tests, are meant to be a different suite.

I think @iSazonov  might be right.  I think option 3 might be the best option.  Mark the tests as pending, and file an issue.  If you are still willing to work on trying to fix the tests, I'd love to work with you.  Assuming @adityapatwardhan (the maintainer for the PR) agrees.

TravisEz13: Oh, and add a comment with the issue number saying why the tests are marked pending.

markekraus: @TravisEz13 Issue #4609 created, tests set to pending, and comments added. I'm still interested in figuring this out, I'm just unsure of the time it will take me to get up to speed on aspnetcore enough to do this.

TravisEz13 (approved):
I'm ok with this assuming Aditya agrees

adityapatwardhan (approved):
LGTM

adityapatwardhan:
Had a chat with @TravisEz13 about the testing challenges. Now that #4609 is opened, I can merge this.

PR #4550, opened 2017-08-11
OpenCover and Code coverage improvements

adityapatwardhan:
- Delete the code coverage output file before the run instead of after the run. So we can remote desktop and see the last runs results,
- Use -noprofile for powershell.exe so that we restrict the testing the PSv6.

### DIFF
--- a/test/tools/OpenCover/OpenCover.psm1
+++ b/test/tools/OpenCover/OpenCover.psm1
@@ -555,7 +555,7 @@ function CreateOpenCoverCmdline($target, $outputLog, $targetArgs)
         "-hideskipped:all",
         "-mergeoutput",
         "-filter:`"+[*]* -[Microsoft.PowerShell.PSReadLine]*`"",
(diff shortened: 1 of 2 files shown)


test/tools/OpenCover/OpenCover.psm1
```diff
@@ -555,7 +555,7 @@ function CreateOpenCoverCmdline($target, $outputLog, $targetArgs)
         "-hideskipped:all",
         "-mergeoutput",
         "-filter:`"+[*]* -[Microsoft.PowerShell.PSReadLine]*`"",
-        "-targetargs:`"-EncodedCommand $base64targetArgs`""
+        "-targetargs:`"-NoProfile -EncodedCommand $base64targetArgs`""
```
JamesWTruher: this is fine, and another possible parameter to use is `-noninteractive` which will keep PSReadLine from being loaded.

JamesWTruher (approved):

PR #2450, opened 2016-10-11
Native pipe

vors:
Fix #559 
Improve pipeline for native commands
- Start native process in Prepare() instead of Complete()
  in NativeCommandProcessor.
- Remove unneeded input, output, error threads. Replaced by
  simpler primitives.

Motivation example:

`ping | %{powershell-code} | grep` now works in real-time.
See the [test](https://github.com/PowerShell/PowerShell/pull/2450/commits/e34a81a30fda95685084b09d4a0d5f60006e4d49#diff-e2e61256a6481fbfcebf5757b220793aR35)

TODO before merging:

- [x] replace spin-lock by something like `notify/wait`
- [x] resurrect and test minishell with custom host serialization
- [x] Changelog entry
- [x] Fix feature tests for https://github.com/PowerShell/PowerShell/blob/ae7b857c235203d1fd3708fe05c076e5a376859a/test/powershell/Host/ConsoleHost.Tests.ps1#L57
- [x] Investigate, is this test flaky or not https://ci.appveyor.com/project/PowerShell/powershell/build/6.0.0-alpha.12-1287/tests

PR #4553, opened 2017-08-11
Make PowerShell Core enumerate COM collections

### LINKED ISSUE #3775: COM objects are not enumerating properly
Repro
------
Open a File Explorer window first if you find $w.Count is 0
```
PSv6> $s = New-Object -ComObject "Shell.Application"
PSv6> $w = $s.Windows()
PSv6> $w.Count
1
PSv6> $w  ## dump $w, and powershell should enumerate $w to unwrap the collection
```
Expected Behavior
--------------------
$w is unwrapped and the contained items are dumpped out (PSv5 behavior)
```
Application          : System.__ComObject
Parent               : System.__ComObject
Container            :
Document             : System.__ComObject
TopLevelContainer    : True
Type                 :
Left                 : -32000
Top                  : -32000
Width                : 160
Height               : 28
LocationName         : 6.0.0-beta1
LocationURL          : file:///D:/PowerShell/6.0.0-beta1
Busy                 : False
Name                 : File Explorer
HWND                 : 923136
FullName             : C:\WINDOWS\Explorer.EXE
Path                 : C:\WINDOWS\
Visible              : True
StatusBar            : False
StatusText           :
ToolBar              : 1
MenuBar              : False
FullScreen           : False
ReadyState           : 4
Offline              : False
Silent               : False
RegisterAsBrowser    : False
RegisterAsDropTarget : True
TheaterMode          : False
AddressBar           : True
Resizable            : True
```

Actual Behavior
-----------------
```
Count
-----
    2
```

Environment
--------------
```
PSv6> $PSVersionTable

CLRVersion
GitCommitId                    v6.0.0-beta.1
```

Additional Observation
-------------------------
The problem appears to be this:
```
PSv6>  $w._NewEnum()
An error occurred while enumerating through a collection: Could not load type 'System.Runtime.InteropServices.ComTypes.IEnumerator' from assembly 'System.Private.CoreLib, Version=4.0.0.0, Culture=neutral,
PublicKeyToken=7cec85d7bea7798e'..
At line:1 char:2
+  $w._NewEnum()
+  ~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.__ComObject:__ComObject) [], RuntimeException
    + FullyQualifiedErrorId : BadEnumeration
```

daxian-dbw:
Fix #3775

'GetEnumerator()' is not supported on COM collections in .NET Core (see https://github.com/dotnet/corefx/issues/19731), so we need to have our own implementation to enumerate COM collections.


lzybkr (dismissed):

src/System.Management.Automation/engine/COM/ComUtil.cs
```diff
@@ -355,4 +356,74 @@ internal static ComMethodInformation[] GetMethodInformationArray(COM.ITypeInfo t
...
+                    catch (Exception)
+                    {
+                        /* Catch all exception. */
+                    }
+                }
+            }
+
+            return enumVariant != null ? new ComEnumerator(enumVariant) : null;
+        }
+    }
```
adityapatwardhan: Add extra line at the end.

adityapatwardhan (commented):
LGTM with minor comment

adityapatwardhan (dismissed):

src/System.Management.Automation/engine/COM/ComUtil.cs
daxian-dbw: Added a new line.

daxian-dbw:
I should use `Measure-Object` in the test but instead used `Measure-Command`, which pops for mandatory parameter input and thus caused the build to hang. It's fixed now.

src/System.Management.Automation/engine/COM/ComUtil.cs
```diff
@@ -355,4 +356,74 @@ internal static ComMethodInformation[] GetMethodInformationArray(COM.ITypeInfo t
...
+
+        public void Reset()
+        {
+            _element[0] = null;
+            _enumVariant.Reset();
+        }
+
+        internal static ComEnumerator Create(object comObject)
+        {
+            if (!comObject.GetType().IsCOMObject) { return null; }
```
iSazonov: Should we check `comObject != null` ?

src/System.Management.Automation/engine/COM/ComUtil.cs
```diff
@@ -355,4 +356,74 @@ internal static ComMethodInformation[] GetMethodInformationArray(COM.ITypeInfo t
...
+                        }
+                    }
+                    catch (Exception)
+                    {
+                        /* Catch all exception. */
+                    }
+                }
+            }
+
+            return enumVariant != null ? new ComEnumerator(enumVariant) : null;
```
iSazonov: Maybe use `else` to exclude duplicate check with line 399?

test/powershell/engine/COM/COM.Basic.Tests.ps1
```diff
@@ -0,0 +1,35 @@
+
+try {
+    $defaultParamValues = $PSdefaultParameterValues.Clone()
+    $PSDefaultParameterValues["it:skip"] = !$IsWindows
+
+    Describe 'Basic COM Tests' -Tags "CI" {
+        It "Should enumerate ShellWindows" {
+            $shell = New-Object -ComObject "Shell.Application"
+            $windows = $shell.Windows()
```
iSazonov: Is it work on Core and IoT?

src/System.Management.Automation/engine/COM/ComUtil.cs
daxian-dbw: It's not a duplicate check, line 416 could make `enumVariant` not null or continue to be null.

src/System.Management.Automation/engine/COM/ComUtil.cs
daxian-dbw: Will add a null check here.

test/powershell/engine/COM/COM.Basic.Tests.ps1
daxian-dbw: Good catch. NanoServer and IoT don't have the shell, so these tests don't apply. Will update.

daxian-dbw:
@iSazonov I have addressed your comments. Please take another look. Thanks!

iSazonov:
@lzybkr  @adityapatwardhan We have new commits - please take another look.

src/System.Management.Automation/engine/COM/ComUtil.cs
iSazonov: Closed.

adityapatwardhan (approved):
LGTM

src/System.Management.Automation/engine/COM/ComUtil.cs
iSazonov: I meant that here slightly hidden logic. Maybe:
```c#
// if IEnumVARIANT interface
var enumVariant = comObject as COM.IEnumVARIANT;
if (enumVariant != null)
{
    return new ComEnumerator(enumVariant);
}

//  if  a collection
var enumerable = comObject as IEnumerable; 
...
    enumVariant = result as COM.IEnumVARIANT; 
    if (enumVariant != null)
    {
        return new ComEnumerator(enumVariant);
    }
...

return null;
```

src/System.Management.Automation/engine/COM/ComUtil.cs
```diff
@@ -355,4 +356,74 @@ internal static ComMethodInformation[] GetMethodInformationArray(COM.ITypeInfo t
...
+                            // IEnumVARIANT interface out of it by invoking its '_NewEnum (DispId: -4)' function.
+                            var result = ComInvoker.Invoke(target, ComTypeInfo.DISPID_NEWENUM,
+                                                           args: Utils.EmptyArray<object>(), byRef: null,
+                                                           invokeKind: comTypeInfo.NewEnumInvokeKind.Value);
+                            enumVariant = result as COM.IEnumVARIANT;   
+                        }
+                    }
+                    catch (Exception)
+                    {
+                        /* Catch all exception. */
```
iSazonov: Obvious comment and typo (exception -> exception**s**). Catch all exceptions - why? Please enhance the comment or remove.

src/System.Management.Automation/engine/COM/ComUtil.cs
daxian-dbw: This is more of a coding style instead of an issue that needs to be corrected.

src/System.Management.Automation/engine/COM/ComUtil.cs
daxian-dbw: Fixed.

src/System.Management.Automation/engine/COM/ComUtil.cs
daxian-dbw: Wait, I see your point now -- there will be one more null check if the COM object is already an enum variant. Will update the code.

daxian-dbw: Code updated.

iSazonov: Thanks for understanding and sorry that I was unclear.

iSazonov (approved):

iSazonov:
@lzybkr Can you approve?

daxian-dbw:
@iSazonov Jason probably doesn't have time to come back and approve it again, so it's up to the maintainer's judgment to see if it's necessary for reviewers who had approved to approve again. There are some guidelines you can use, for example:
1. are the changes after the previous approval fundamental or just touch-ups
2. are the changes to address comments left by the reviewer who previously approved
Certainly, you will get your own guidelines as you become more experienced in this role.

iSazonov:
@daxian-dbw Thank you for the helpful comment!

In this case, it's more of a courtesy - I have to show respect and ask him after I formally dismissed his approval.

test/powershell/engine/COM/COM.Basic.Tests.ps1
```diff
@@ -0,0 +1,48 @@
+
+try {
+    $defaultParamValues = $PSdefaultParameterValues.Clone()
+    $PSDefaultParameterValues["it:skip"] = ![System.Management.Automation.Platform]::IsWindowsDesktop
+
+    Describe 'Basic COM Tests' -Tags "CI" {
```
JamesWTruher: the following might be a more direct test:
```
Describe a {
        BeforeAll {
                $null = new-item -type file $TESTDRIVE/file1
                $null = new-item -type file $TESTDRIVE/file2
                $null = new-item -type file $TESTDRIVE/file3
                $null = new-item -type file $TESTDRIVE/file4
        }
        it b {
                $shell = New-Object -com shell.application
                $folder = $shell.Namespace("$TESTDRIVE")
                $folder.Items().Count | should be 4
                $names = $folder.Items() | %{$_.name}
                $names -join "," | should be "file1,file2,file3,file4"
        }
}
```
you have complete control over what you expect and can validate it

daxian-dbw: I like this one. Will add it to the test.

daxian-dbw:
@iSazonov Another related comment. I usually don't dismiss a reviewer's approval unless the PR needs some fundamental changes (e.g. design flaw, logic changes and etc.). If the following-up changes are touch-ups or minor fixes (like typo, a null condition check, test fix-up), I think it's fine to keep the previous approval, and once the minor issues are addressed, you can approve the PR and merge it.

test/powershell/engine/COM/COM.Basic.Tests.ps1
daxian-dbw: Test updated.

iSazonov:
@daxian-dbw I wonder why CI AppVeyor failed?

daxian-dbw:
I have investigated the failure, it actually a bug in Binder, more specifically at [here](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/runtime/Binding/Binders.cs#L4978), the restriction generated in the call to `this.DeferForPSObject` doesn't respect the fact that the baseobject needs to be a COM object.

A repro would be:
```
$shell = New-Object -ComObject "Shell.Application"
$folder = $shell.Namespace("F:\tmp")
#$item = $folder.Items().Item(0)
$item = $folder.Items() | select -Last 1
$item.Name


$members = new-object System.Collections.ObjectModel.Collection[System.Management.Automation.PSMemberInfo]
$n=new-object Management.Automation.PSNoteProperty a,1
$members.Add($n)
$r=add-member -InputObject a -MemberType MemberSet -Name Name -Value $members -passthru
$r.Name.a
```

The `$r.Name.a` is supposed to return 1. If you run the lower part first then upper part, then all work fine.
This happens when a COM object is wrapped into a PSObject, which is the case when the COM object is returned from a pipeline, like the above repro code.

I cannot open an issue until this PR is merged because the repro requires enumerating a COM collection.
I have updated the test to not access `Name` member, so it would be good now. When I resolve the bug, I will update the COM basic test to keep accessing the `Name` member.

iSazonov:
@daxian-dbw Maybe made two commits for code and tests before merge?

daxian-dbw:
@iSazonov Please just choose 'Squash and merge'. It's a simple fix and there is no need to keep history commits.

daxian-dbw:
A follow-up of the binder bug:
The issue repros without COM enumeration. #4607 opened to track it.

iSazonov:
@daxian-dbw Merged.
Great work! Thanks!

daxian-dbw:
@iSazonov Thank you for the review and good work as a maintainer 😄 

Here are some of my experiences to choose among `Squash and merge`, `Rebase and merge` and `Create a merge commit`:
1. number of meaningful commits (commits that are part of the PR work, or commits to address major comments that require non-trivial design/logic changes) is one thing to consider
1. the complexity of the PR work is also something to consider.

- Some PR may involve many commits, but the fix or feature work is not complex, in those cases, just `Squash and merge`. 
- Some PR may have clean history commits, but again, the fix/work is simple, in those cases, `Squash and merge`
- A PR may be complex work, and the history is not clean -- complex work usually means a lot major changes during the review process -- in such case, choose `create a merge commit` to preserve the history even though it's not clean. Example: #2450
- A PR is non-trivial and the history is clean, then it's worthy to keep the history. In this case, consider using `rebase and merge` to keep the main branch history clean if the number of commits is not too many (~5/6). Otherwise, use `create a merge commit`.
- A PR may contain more than 1 fix, in this case we usually should ask the author to break it into two PRs.

As you can see from the main branch history, most of time we use (and prefer to) `squash and merge`.

As for what is a simple/complex/non-trivial fix/work, it's hard to describe but you will have your judgement call as you gain more expeirences.

iSazonov:
@daxian-dbw Thanks for sharing your experience!

Issue #4310, opened 2017-07-20
Update docker files to use http://label-schema.org/rc1/


TravisEz13:

daxian-dbw:
Close via #4737

PR #4555, opened 2017-08-11
Update nano server insider for new bulid

TravisEz13:
Also, Address #4310

### DIFF
--- a/docker/release/nanoserver-insider/Dockerfile
+++ b/docker/release/nanoserver-insider/Dockerfile
@@ -29,7 +29,20 @@ ARG IMAGE_NAME=microsoft/nanoserver-insider-powershell
 LABEL maintainer="PowerShell Team <powershellteam@hotmail.com>"
 LABEL readme.md="https://github.com/PowerShell/PowerShell/blob/master/docker/README.md"
 LABEL description="This Dockerfile will install the latest release of PS."
-
+# Usage on URL to Usage
+LABEL org.label-schema.usage="https://github.com/PowerShell/PowerShell/tree/master/docker#run-the-docker-image-you-built"
+LABEL org.label-schema.url="https://github.com/PowerShell/PowerShell/blob/master/docker/README.md"
+LABEL org.label-schema.vcs-url="https://github.com/PowerShell/PowerShell"
+# Image friendly name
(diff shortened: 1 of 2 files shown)


docker/release/nanoserver-insider/Dockerfile
```diff
@@ -21,18 +21,36 @@ RUN Expand-Archive powershell.zip -DestinationPath \PowerShell
...
 # Copy Powershell from the installer containter
 ENV ProgramFiles C:\Program Files
 COPY --from=installer-env ["\\PowerShell\\", "$ProgramFiles\\PowerShell"]
 
 # Persist %PSCORE% ENV variable for user convenience
 ENV PSCORE="$ProgramFiles\PowerShell\PowerShell.exe"
 
-# setx /M fails on nanoserver-insider, setting the user path
+# Set the path
 RUN setx PATH "%PATH%;%ProgramFiles%\PowerShell"
```
daxian-dbw: Just curious, `setx /M` is not necessary on nanoserver because it's a single user OS, correct?

docker/release/nanoserver-insider/Dockerfile
```diff
@@ -21,18 +21,36 @@ RUN Expand-Archive powershell.zip -DestinationPath \PowerShell
...
 
-CMD ["C:\\Program Files\\PowerShell\\PowerShell.exe"]
```
daxian-dbw: In docker/README.md:
> Due to an issue with the container not picking up the path, you must specify the path when running a command on the command line.

Looks like the PATH issue has been fixed, so the document should be updated.

adityapatwardhan (approved):
LGTM

docker/release/nanoserver-insider/Dockerfile
TravisEz13: correct, they fixed the container to use the per user path.

docker/release/nanoserver-insider/Dockerfile
TravisEz13: I added a note that it will be fixed soon.  I have to setup a build machine matching the exact build of the container, which might take me some time.

daxian-dbw (approved):

PR #4556, opened 2017-08-11
Add File Coverage to coverage data

JamesWTruher:
It is now possible to see the coverage based on the file and then format
the data so a report generator is not needed. You can map the files from
one location to another, so if you have a copy of the repo, but in a different
location, you can still format the output in a way so it may be studied

Also, help is available for Format-FileCoverage!


JamesWTruher:
the test failures seem to be flaky tests

TravisEz13:
I've triggered AppVeyor to run again

test/tools/OpenCover/OpenCover.psm1
```diff
@@ -40,6 +40,172 @@ function Get-ClassCoverageData([xml.xmlelement]$element)
...
+   0800 -                 catch (System.ServiceProcess.TimeoutException) // still waiting
+   0801 -                 {
+   0802 -                     if (serviceController.Status != pendingStatus
+   ...
+#>
+function Format-FileCoverage
+{
+    param ( 
+        [Parameter(Position=0,Mandatory=$true,ValueFromPipeline=$true)]$FileCoverageData, 
+        [Parameter()][string]$oldBase = "",
```
adityapatwardhan: Should be have a default value for this?

test/tools/OpenCover/OpenCover.psm1
```diff
@@ -40,6 +40,172 @@ function Get-ClassCoverageData([xml.xmlelement]$element)
...
+   0801 -                 {
+   0802 -                     if (serviceController.Status != pendingStatus
+   ...
+#>
+function Format-FileCoverage
+{
+    param ( 
+        [Parameter(Position=0,Mandatory=$true,ValueFromPipeline=$true)]$FileCoverageData, 
+        [Parameter()][string]$oldBase = "", 
+        [Parameter()][string]$newBase = "",
```
adityapatwardhan: Maybe use git to find project root for the default value?

test/tools/OpenCover/OpenCover.psm1
```diff
@@ -40,6 +40,172 @@ function Get-ClassCoverageData([xml.xmlelement]$element)
...
+{
+    param ( 
+        [Parameter(Position=0,Mandatory=$true,ValueFromPipeline=$true)]$FileCoverageData, 
+        [Parameter()][string]$oldBase = "", 
+        [Parameter()][string]$newBase = "", 
+        [Parameter()][string]$filter = ".*" 
+    )
+
+    $files = @($fileCoverageData.Keys) | Where-Object { $_ -match $filter }
+    write-verbose -verbose $files
```
adityapatwardhan: Do we always want a verbose message or a Write-Host -Fore Yellow?

adityapatwardhan (changes requested):

test/tools/OpenCover/OpenCover.psm1
JamesWTruher: under most circumstances (for example, when running on the system where the data was collected), it's not needed at all, so I don't want to do any substitution. By setting it to default of "" and "" (for old and new) no substitution takes place and everything works as expected.

test/tools/OpenCover/OpenCover.psm1
JamesWTruher: see above - generally don't want to change the path, but sometimes I do. setting this to "" makes it work nearly all the time, but when I'm working on my mac which has different path structure, I can still look at the coverage data. In this case I may not be in a repo at all.

test/tools/OpenCover/OpenCover.psm1
JamesWTruher: probably not - removed

JamesWTruher:
@adityapatwardhan updated and pushed

TravisEz13:
restarted appveyor due to a myget failure

test/tools/OpenCover/OpenCover.psm1
adityapatwardhan: There is precedence here with Invoke-OpenCover, we try to find powershell root directory and build path.

test/tools/OpenCover/OpenCover.psm1
adityapatwardhan: See above about precedence in Invoke-OpenCover.

test/tools/OpenCover/OpenCover.psm1
JamesWTruher: I believe it's the wrong thing to do. This may be used completely outside of a repo unlike invoke-opencover. consider the scenario inspecting coverage on a remote machine where the paths are not the same as what is in the file. so I can say `-oldbase f:/github/blah -newwbase \\system\share\blah` If you're on the system where you collected the data, you don't to substitute anything, doing a replacement is unneeded.

adityapatwardhan: Closed.

test/tools/OpenCover/OpenCover.psm1
adityapatwardhan: Closed.

adityapatwardhan (approved):
LGTM

adityapatwardhan:
@TravisEz13 would you be having a look at this?

TravisEz13 (approved):

test/tools/OpenCover/OpenCover.psm1
```diff
@@ -134,7 +299,8 @@ function Get-CoverageData($xmlPath)
         CoverageLogFile = $xmlPath
         CoverageSummary = (Get-CoverageSummary -element $CoverageXml.CoverageSession.Summary)
         Assembly = $assemblies
-    }
+        FileCoverage = Get-FileCoverageData $CoverageXml
+        }
```
daxian-dbw: Can you please fix this indentation?

JamesWTruher: certainly - commit amended and pushed

adityapatwardhan: @daxian-dbw Can you take a quick look?

daxian-dbw: Thanks. #Closed

daxian-dbw (approved):

Issue #4542, opened 2017-08-10
Using Copy-Item from Linux (Ubuntu 16.04.2) To Windows (2016 Server Core)


dnataraj:
I'm wondering, since SSH remoting from Linux to Windows is available in some form now, if it will be possible eventually to copy files from Linux to Windows.

I've got the remoting to work very nicely following this excellent guide https://github.com/PowerShell/PowerShell/tree/master/demos/SSHRemoting

Steps to reproduce
------------------

```powershell
$session = New-PSSession -HostName remotewinservercore2016.azure.com -UserName foo

Copy-Item -Path ./foo.text -Destination C:\users\foo\ -Verbose -ToSession $session

```

Expected behavior
-----------------

```none

```

Actual behavior
---------------

```none
Copy-Item : Unable to load DLL 'api-ms-win-core-file-l1-2-2.dll': The specified module or one of its dependencies could not be found.
(Exception from HRESULT: 0x8007007E)
At line:1 char:1
+ Copy-Item -Path ./Downloads/com.cer -Destination C:\users\op ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Copy-Item], DllNotFoundException
    + FullyQualifiedErrorId : System.DllNotFoundException,Microsoft.PowerShell.Commands.CopyItemCommand
```

Environment data
----------------

```powershell
> $PSVersionTable

Name                           Value                                                                                                                                                      
----                           -----                                                                                                                                                      
PSVersion                      6.0.0-beta                                                                                                                                                 
PSEdition                      Core                                                                                                                                                       
GitCommitId                    v6.0.0-beta.5                                                                                                                                              
OS                             Linux 4.4.0-89-generic #112-Ubuntu SMP Mon Jul 31 19:38:41 UTC 2017                                                                                        
Platform                       Unix                                                                                                                                                       
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                                                                    
PSRemotingProtocolVersion      2.3                                                                                                                                                        
SerializationVersion           1.1.0.1                                                                                                                                                    
WSManStackVersion              3.0           

```

dnataraj:
Actually, it appears that the file is actually transferred! (I didn't realize this at first, sorry). But the error still occurs.

iSazonov:
[...]

PR #4567, opened 2017-08-14
Remove alternate file stream code from non-Windows

SteveL-MSFT:
AlternateStreams support relies on pinvoke to win32 apis which don't work

on non-Windows and produces errors about not loading a dll in the case of copy-item

Fix https://github.com/PowerShell/PowerShell/issues/4542

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
@@ -14,6 +14,7 @@ using System.Management.Automation.Internal;
 
 #endregion
 
+#if !UNIX
 namespace Microsoft.PowerShell.Commands
 {
     /// <summary>Removes the Zone.Identifier stream from a file.</summary>
@@ -162,3 +163,4 @@ namespace Microsoft.PowerShell.Commands
         }
     }
 }
+#endif
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
@@ -1,3 +1,4 @@
+#if !UNIX
 /********************************************************************++
 Copyright (c) Microsoft Corporation.  All rights reserved.
 --********************************************************************/
@@ -14,7 +15,6 @@ using System.Management.Automation.Internal;
 
 #endregion
 
-#if !UNIX
 namespace Microsoft.PowerShell.Commands
 {
     /// <summary>Removes the Zone.Identifier stream from a file.</summary>
--- a/src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs
+++ b/src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs
@@ -1304,7 +1304,9 @@ namespace Microsoft.PowerShell.Commands
                     fullPath,
                     file.RawFileData);
 
+#if !UNIX
                 AlternateDataStreamUtilities.SetZoneOfOrigin(fullPath, SecurityZone.Intranet);
+#endif
             }
 
             return relativePathsToCreatedFiles;
--- a/src/System.Management.Automation/namespaces/FileSystemContentStream.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemContentStream.cs
@@ -843,7 +843,9 @@ namespace Microsoft.PowerShell.Commands
             {
                 if (!String.IsNullOrEmpty(streamName))
                 {
+#if !UNIX
                     _stream = AlternateDataStreamUtilities.CreateFileStream(filePath, streamName, fileMode, fileAccess, fileShare);
+#endif
                 }
                 else
                 {
@@ -854,7 +856,9 @@ namespace Microsoft.PowerShell.Commands
             {
                 if (!String.IsNullOrEmpty(streamName))
                 {
+#if !UNIX
                     _stream = AlternateDataStreamUtilities.CreateFileStream(filePath, streamName, fileMode, requestedAccess, fileShare);
+#endif
                 }
                 else
                 {
--- a/src/System.Management.Automation/namespaces/FileSystemContentStream.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemContentStream.cs
@@ -845,6 +845,8 @@ namespace Microsoft.PowerShell.Commands
                 {
 #if !UNIX
                     _stream = AlternateDataStreamUtilities.CreateFileStream(filePath, streamName, fileMode, fileAccess, fileShare);
+#else
+                    throw new PlatformNotSupportedException();
 #endif
                 }
                 else
@@ -858,6 +860,8 @@ namespace Microsoft.PowerShell.Commands
                 {
 #if !UNIX
                     _stream = AlternateDataStreamUtilities.CreateFileStream(filePath, streamName, fileMode, requestedAccess, fileShare);
+#else
+                    throw new PlatformNotSupportedException();
 #endif
                 }
                 else
--- a/src/System.Management.Automation/namespaces/FileSystemContentStream.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemContentStream.cs
@@ -841,30 +841,26 @@ namespace Microsoft.PowerShell.Commands
 
             try
             {
+#if !UNIX
                 if (!String.IsNullOrEmpty(streamName))
                 {
-#if !UNIX
                     _stream = AlternateDataStreamUtilities.CreateFileStream(filePath, streamName, fileMode, fileAccess, fileShare);
-#else
-                    throw new PlatformNotSupportedException();
-#endif
                 }
                 else
+#endif
                 {
                     _stream = new FileStream(filePath, fileMode, fileAccess, fileShare);
                 }
             }
             catch (IOException)
             {
+#if !UNIX
                 if (!String.IsNullOrEmpty(streamName))
                 {
-#if !UNIX
                     _stream = AlternateDataStreamUtilities.CreateFileStream(filePath, streamName, fileMode, requestedAccess, fileShare);
-#else
-                    throw new PlatformNotSupportedException();
-#endif
                 }
                 else
+#endif
                 {
                     _stream = new FileStream(filePath, fileMode, requestedAccess, fileShare);
                 }
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -1046,6 +1046,7 @@ namespace Microsoft.PowerShell.Commands
             path = NormalizePath(path);
             path = EnsureDriveIsRooted(path);
 
+#if !UNIX
             // Remove alternate data stream references
             // See if they've used the inline stream syntax. They have more than one colon.
             int firstColon = path.IndexOf(':');
@@ -1054,6 +1055,7 @@ namespace Microsoft.PowerShell.Commands
             {
                 path = path.Substring(0, secondColon);
             }
+#endif
 
             //Make sure the path is either drive rooted or UNC Path
             if (!IsAbsolutePath(path) && !IsUNCPath(path))
@@ -4686,9 +4688,9 @@ namespace Microsoft.PowerShell.Commands
 
             bool result = CopyFileStreamToRemoteSession(file, remoteFilePath, ps, false, null);
 
+#if !UNIX
             bool targetSupportsAlternateStreams = RemoteTargetSupportsAlternateStreams(ps, remoteFilePath);
 
-#if !UNIX
             // Once the file is copied successfully, check if the file has any alternate data streams
             if (result && targetSupportsAlternateStreams)
             {
@@ -5165,6 +5167,7 @@ namespace Microsoft.PowerShell.Commands
 
             s_tracer.WriteLine("basePath = {0}", basePath);
 
+#if !UNIX
             // Remove alternate data stream references
             // See if they've used the inline stream syntax. They have more than one colon.
             string alternateDataStream = String.Empty;
@@ -5176,6 +5179,7 @@ namespace Microsoft.PowerShell.Commands
                 alternateDataStream = path.Replace(newPath, "");
                 path = newPath;
             }
+#endif
 
             string result = path;
 
@@ -5299,10 +5303,12 @@ namespace Microsoft.PowerShell.Commands
                 }
             } while (false);
 
+#if !UNIX
             if (!String.IsNullOrEmpty(alternateDataStream))
             {
                 result = result + alternateDataStream;
             }
+#endif
 
             return result;
         } // NormalizeRelativePathHelper
@@ -6619,6 +6625,7 @@ namespace Microsoft.PowerShell.Commands
                 } // dynParams != null
             } // DynamicParameters != null
 
+#if !UNIX
             // See if they've used the inline stream syntax. They have more than one colon.
             int firstColon = path.IndexOf(':');
             int secondColon = path.IndexOf(':', firstColon + 1);
@@ -6627,6 +6634,7 @@ namespace Microsoft.PowerShell.Commands
                 streamName = path.Substring(secondColon + 1);
                 path = path.Remove(secondColon);
             }
+#endif
 
             FileSystemContentReaderWriter stream = null;
 
@@ -6765,6 +6773,7 @@ namespace Microsoft.PowerShell.Commands
                 } // dynParams != null
             }
 
+#if !UNIX
             // See if they've used the inline stream syntax. They have more than one colon.
             int firstColon = path.IndexOf(':');
             int secondColon = path.IndexOf(':', firstColon + 1);
@@ -6773,6 +6782,7 @@ namespace Microsoft.PowerShell.Commands
                 streamName = path.Substring(secondColon + 1);
                 path = path.Remove(secondColon);
             }
+#endif
 
             FileSystemContentReaderWriter stream = null;
 
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -2804,17 +2804,22 @@ namespace Microsoft.PowerShell.Commands
                     return;
                 }
 
-#if !UNIX
-                if ((!removeStreams) && iscontainer)
-#else
+#if UNIX
                 if (iscontainer)
-#endif
                 {
                     RemoveDirectoryInfoItem((DirectoryInfo)fsinfo, recurse, Force, true);
                 }
                 else
                 {
-#if !UNIX
+                    RemoveFileInfoItem((FileInfo)fsinfo, Force);
+                }
+#else
+                if ((!removeStreams) && iscontainer)
+                {
+                    RemoveDirectoryInfoItem((DirectoryInfo)fsinfo, recurse, Force, true);
+                }
+                else
+                {
                     // If we want to remove the file streams, retrieve them and remove them.
                     if (removeStreams)
                     {
@@ -2854,11 +2859,11 @@ namespace Microsoft.PowerShell.Commands
                         }
                     }
                     else
-#endif
                     {
                         RemoveFileInfoItem((FileInfo)fsinfo, Force);
                     }
                 }
+#endif
             }
             catch (IOException exception)
             {
@@ -4267,11 +4272,13 @@ namespace Microsoft.PowerShell.Commands
                         ps.AddParameter("force", true);
                     }
 
+#if !UNIX
                     if (isAlternateDataStream)
                     {
                         ps.AddParameter("isAlternateStream", true);
                         ps.AddParameter("streamName", streamName);
                     }
+#endif
 
                     Hashtable op = SafeInvokeCommand.Invoke(ps, this, null);
 
--- a/src/System.Management.Automation/namespaces/FileSystemProvider.cs
+++ b/src/System.Management.Automation/namespaces/FileSystemProvider.cs
@@ -6612,8 +6612,10 @@ namespace Microsoft.PowerShell.Commands
                     // Get the wait value
                     waitForChanges = dynParams.Wait;
 
+#if !UNIX
                     // Get the stream name
                     streamName = dynParams.Stream;
+#endif
                 } // dynParams != null
             } // DynamicParameters != null
 
@@ -6756,7 +6758,9 @@ namespace Microsoft.PowerShell.Commands
                         encoding = dynParams.EncodingType;
                     }
 
+#if !UNIX
                     streamName = dynParams.Stream;
+#endif
                     suppressNewline = dynParams.NoNewline.IsPresent;
                 } // dynParams != null
             }
@@ -7628,12 +7632,13 @@ namespace Microsoft.PowerShell.Commands
         [Parameter]
         public FileSystemCmdletProviderEncoding Encoding { get; set; } = FileSystemCmdletProviderEncoding.String;
 
+#if !UNIX
         /// <summary>
         /// A parameter to return a stream of an item.
         /// </summary>
         [Parameter]
         public String Stream { get; set; }
-
+#endif
 
         /// <summary>
         /// Gets the encoding from the specified StreamType parameter.
@@ -7677,11 +7682,13 @@ namespace Microsoft.PowerShell.Commands
     /// </summary>
     public class FileSystemClearContentDynamicParameters
     {
+#if !UNIX
         /// <summary>
         /// A parameter to return a stream of an item.
         /// </summary>
         [Parameter]
         public String Stream { get; set; }
+#endif
     } //FileSystemContentWriterDynamicParameters
 
     /// <summary>
@@ -7805,6 +7812,7 @@ namespace Microsoft.PowerShell.Commands
     /// </summary>
     public class FileSystemProviderGetItemDynamicParameters
     {
+#if !UNIX
         /// <summary>
         /// A parameter to return the streams of an item.
         /// </summary>
@@ -7812,6 +7820,7 @@ namespace Microsoft.PowerShell.Commands
         [ValidateNotNullOrEmpty()]
         [SuppressMessage("Microsoft.Performance", "CA1819:PropertiesShouldNotReturnArrays")]
         public string[] Stream { get; set; }
+#endif
     } // class FileSystemItemProviderDynamicParameters
 
     /// <summary>
@@ -7819,6 +7828,7 @@ namespace Microsoft.PowerShell.Commands
     /// </summary>
     public class FileSystemProviderRemoveItemDynamicParameters
     {
+#if !UNIX
         /// <summary>
         /// A parameter to return the streams of an item.
         /// </summary>
@@ -7826,6 +7836,7 @@ namespace Microsoft.PowerShell.Commands
         [ValidateNotNullOrEmpty()]
         [SuppressMessage("Microsoft.Performance", "CA1819:PropertiesShouldNotReturnArrays")]
         public string[] Stream { get; set; }
+#endif
     } // class FileSystemItemProviderDynamicParameters
 
     #endregion
@@ -8285,7 +8296,7 @@ namespace Microsoft.PowerShell.Commands
         }
 
 #if !UNIX
-        internal static bool WinIsSameFileSystemItem(string pathOne, string pathTwo)
+        private static bool WinIsSameFileSystemItem(string pathOne, string pathTwo)
         {
             var access = FileAccess.Read;
             var share = FileShare.Read;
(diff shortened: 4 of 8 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
```diff
@@ -13,6 +13,7 @@
 
 #endregion
 
+#if !UNIX
```
iSazonov: `Unblock-File` is excluded from Unix at all so we don't need to exclude the code here.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
```diff
@@ -126,4 +126,17 @@ Describe "Get-Content" -Tags "CI" {
...
+  It "Should support NTFS streams using colon syntax" -Skip:(!$IsWindows) {
+    Set-Content "${testPath}:Stream" -Value "Foo"
+    { Test_path "${testPath}:Stream" | ShouldBeErrorId "ItemExistsNotSupportedError,Microsoft.PowerShell.Commands,TestPathCommand" }
+    Get-Content "${testPath}:Stream" | Should BeExactly "Foo"
+    Get-Content $testPath | Should BeExactly $testString
+  }
+
+  It "Should support colons in filename on Linux/Mac" -Skip:($IsWindows) {
+    Set-Content "${testPath}:Stream" -Value "Hello"
+    Get-Content "${testPath}:Stream" | Should BeExactly "Hello"
```
iSazonov: Please add a check that `${testPath}:Stream` is a file name.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
```diff
@@ -126,4 +126,17 @@ Describe "Get-Content" -Tags "CI" {
...
+  }
+
```
iSazonov: Extra line.

iSazonov (commented):
Leave a comment

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -6828,10 +6843,11 @@ public void ClearContent(string path)
 
             try
             {
+#if !UNIX
                 bool clearStream = false;
+                string streamName = null;
```
adityapatwardhan: Unnecessary change of line.

src/System.Management.Automation/utils/ClrFacade.cs
```diff
@@ -147,20 +147,15 @@ internal static SecurityZone GetFileSecurityZone(string filePath)
         {
             Diagnostics.Assert(Path.IsPathRooted(filePath), "Caller makes sure the path is rooted.");
             Diagnostics.Assert(Utils.NativeFileExists(filePath), "Caller makes sure the file exists.");
-#if CORECLR
             string sysRoot = System.Environment.GetEnvironmentVariable("SystemRoot");
```
adityapatwardhan: Should this block be surrounded by #if !UNIX?

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
```diff
@@ -126,4 +126,17 @@ Describe "Get-Content" -Tags "CI" {
     } else {$expected = "Hell","o,ll","ll","Worll","d`nHell","o2,ll","ll","Worll","d2`n"}
     for ($i = 0; $i -lt $result.Length ; $i++) { $result[$i]  | Should BeExactly $expected[$i]}
   }
+
+  It "Should support NTFS streams using colon syntax" -Skip:(!$IsWindows) {
+    Set-Content "${testPath}:Stream" -Value "Foo"
+    { Test_path "${testPath}:Stream" | ShouldBeErrorId "ItemExistsNotSupportedError,Microsoft.PowerShell.Commands,TestPathCommand" }
```
adityapatwardhan: Should this be ```Test-Path``` instead of ```Test_path```

adityapatwardhan (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
SteveL-MSFT: That file is still being compiled on Unix

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
SteveL-MSFT: Doesn't `Get-Content` implicitly cover that?

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: The change of grouping of the variables makes it a bit more readable to me

src/System.Management.Automation/utils/ClrFacade.cs
SteveL-MSFT: Some of this code is unnecessary on UNIX so I'll wrap it

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
SteveL-MSFT: Yes, odd that this test passed

src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
iSazonov: I think we should exclude them in csproj.

SteveL-MSFT: I'm fine with that but then we have effectively conditional compilation in both the source and the csproj files

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
SteveL-MSFT: Will remove

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
SteveL-MSFT: I guess if the Content cmdlets were doing something funny it could mask an issue, I'll add a check

iSazonov (dismissed):

src/Microsoft.PowerShell.Commands.Utility/Microsoft.PowerShell.Commands.Utility.csproj
```diff
@@ -35,6 +35,7 @@
     <Compile Remove="commands\utility\ShowCommand\ShowCommandParameterType.cs" />
     <Compile Remove="commands\utility\ShowCommand\ShowCommandProxy.cs" />
     <Compile Remove="commands\utility\update-list.cs" />
+    <Compile Remove="commands\utility\Unblockfile.cs" Condition=" '$(Configuration)' == 'Linux' " />
```
adityapatwardhan: Should we add ```#if !UNIX``` to the start of the file like: https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/utils/tracing/EtwActivity.cs#L1

That seems more discoverable to me when doing code reviews or debugging.

adityapatwardhan (dismissed):

src/Microsoft.PowerShell.Commands.Utility/Microsoft.PowerShell.Commands.Utility.csproj
SteveL-MSFT: That's what I had originally.  Did a search and I don't see other cases of conditional compilation, so I think it's more consistent to use a whole file `#ifdef`.  @iSazonov ok?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
dantraMSFT: Should we rename this to UnblockFile.Windows.cs? 

---
In reply to: [133130873](https://github.com/PowerShell/PowerShell/pull/4567#discussion_r133130873) [](ancestors = 133130873)

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1146,10 +1147,12 @@ protected override void GetItem(string path)
                         }
                     }
                 }
+#endif
 
                 FileSystemInfo result = GetFileSystemItem(path, ref isContainer, false);
                 if (result != null)
                 {
+#if !UNIX
```
dantraMSFT: This #if/#endif block is cumbersome to read.  Suggest instead:
if (result != null)
{
#if UNIX
     WriteItemObject(...)
#else
    original code
 }

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -2798,12 +2804,17 @@ protected override void RemoveItem(string path, bool recurse)
                     return;
                 }
 
+#if !UNIX
```
dantraMSFT: As above, the code is easier to read if you use a #if UNIX and contain the UNIX lines completely.  Its a simple if/else with a single statement for UNIX but  looks unwieldy with the current #if/#else placement.

dantraMSFT:
>                     if (isAlternateDataStream)

Should this also be conditioned on !UNIX?

---
Refers to: src/System.Management.Automation/namespaces/FileSystemProvider.cs:4270 in 7209973. [](commit_id = 720997351e1de1cf26593cfe0e95699d8b3b9eb3, deletion_comment = False)

dantraMSFT:
>             return Platform.NonWindowsIsSameFileSystemItem(pathOne, pathTwo);

Shouldn't there be a Platform.IsSameFileSystemItem?

---
Refers to: src/System.Management.Automation/namespaces/FileSystemProvider.cs:8274 in 7209973. [](commit_id = 720997351e1de1cf26593cfe0e95699d8b3b9eb3, deletion_comment = False)

src/System.Management.Automation/utils/ClrFacade.cs
```diff
@@ -147,20 +147,17 @@ internal static SecurityZone GetFileSecurityZone(string filePath)
...
             string urlmonPath = Path.Combine(sysRoot, @"System32\urlmon.dll");
             if (Utils.NativeFileExists(urlmonPath))
             {
                 return MapSecurityZoneWithUrlmon(filePath);
             }
-            return MapSecurityZoneWithoutUrlmon(filePath);
-#else
-            return MapSecurityZoneWithUrlmon(filePath);
 #endif
+            return MapSecurityZoneWithoutUrlmon(filePath);
```
dantraMSFT: it appears you're still hitting alternate data streams on unix.
ClrFacade:  MapSecurityZoneWithoutUrlmon -> ReadFromZoneIdentifierDataStream

src/System.Management.Automation/namespaces/FileSystemContentStream.cs
```diff
@@ -843,7 +843,9 @@ private void CreateStreams(string filePath, string streamName, FileMode fileMode
             {
                 if (!String.IsNullOrEmpty(streamName))
                 {
+#if !UNIX
```
dantraMSFT: On UNIX, _stream will be null causing a ArgumentNullException from new StreamReader at line 874.  At best, it will be misleading. Consider throwing NotSupportException on UNIX in the associated constructors

src/Microsoft.PowerShell.Commands.Utility/Microsoft.PowerShell.Commands.Utility.csproj
iSazonov: I think it's technically better to exclude files in csproj files. But currently we don't use a lot of opportunities of MSBuild and also we are cleaning up FullCLR files and respectively remove "Compile Remove" from csproj files. So it is Ok to use  #ifdef .

/cc @daxian-dbw  Could you comment too?

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: The `WriteItemObject` code applies to both UNIX and !UNIX.  If I move that bit of code up, I'd need more `#ifdefs` making it less readable

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: I can move some code around so the UNIX code is in one place and the !UNIX code is also together.

src/System.Management.Automation/namespaces/FileSystemContentStream.cs
SteveL-MSFT: makes sense, will change

SteveL-MSFT:
@dantraMSFT I addressed your feedback.  Regarding `Platform.IsSameFileSystem` we should probably have a single API that is platform agnostic, but that's out of scope of this PR.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
dantraMSFT: Actually, if you don't mind duplicating the WriteItemObject line, it can be a pure #if #else block.  Your call.

src/System.Management.Automation/namespaces/FileSystemContentStream.cs
```diff
@@ -845,6 +845,8 @@ private void CreateStreams(string filePath, string streamName, FileMode fileMode
                 {
 #if !UNIX
                     _stream = AlternateDataStreamUtilities.CreateFileStream(filePath, streamName, fileMode, fileAccess, fileShare);
+#else
+                    throw new PlatformNotSupportedException();
```
dantraMSFT: I think these exceptions need to be explicit in indicating alternate data streams are not supported. That's why I suggested an exception at ctor time; an exception that indicates streamName is not supported on UNIX.

src/System.Management.Automation/utils/ClrFacade.cs
```diff
@@ -147,20 +147,17 @@ internal static SecurityZone GetFileSecurityZone(string filePath)
         {
```
daxian-dbw: SecurityZone is a windows concept which I think shouldn't be applied when running on Unix plats. We shouldn't have `#if UNIX/endif` in this method or the methods exclusively used by this method, instead, we probably should analyze the call sites of this method to make sure this method doesn't run on unix plats, or better, enclose this method within `#if !UNIX/endif`.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1117,6 +1117,7 @@ protected override void GetItem(string path)
 
             try
             {
+#if !UNIX
                 bool retrieveStreams = false;
                 FileSystemProviderGetItemDynamicParameters dynamicParameters = null;
```
daxian-dbw: I think the fix should be to make the dynamic parameters related to `Stream` operation unavailable in FileSystemProvider. Otherwise, it's broken user experience as I can use the `-Stream` parameter but it doesn't behave as expected.

daxian-dbw (dismissed):

src/System.Management.Automation/namespaces/FileSystemProvider.cs
JamesWTruher: yes, it would be better to not have the `-stream` parameter if it's not supported. It may also be cleaner to do it this way

src/System.Management.Automation/namespaces/FileSystemContentStream.cs
```diff
@@ -843,7 +843,11 @@ private void CreateStreams(string filePath, string streamName, FileMode fileMode
             {
                 if (!String.IsNullOrEmpty(streamName))
                 {
+#if !UNIX
                     _stream = AlternateDataStreamUtilities.CreateFileStream(filePath, streamName, fileMode, fileAccess, fileShare);
+#else
+                    throw new PlatformNotSupportedException();
```
daxian-dbw: I think it's better to make sure `CreateStreams` is never used for an alternate stream instead of inserting PlatformNotSupportedException.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
daxian-dbw: Once we stop providing the `-Stream` dynamic parameter for FileSystemProvider on Unix, we don't need to add so many `#if UNIX/endif` blocks in this file because those codes will not be hit. I think that will make the code more readable and easy to maintain.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
```diff
@@ -126,4 +126,18 @@ Describe "Get-Content" -Tags "CI" {
     } else {$expected = "Hell","o,ll","ll","Worll","d`nHell","o2,ll","ll","Worll","d2`n"}
     for ($i = 0; $i -lt $result.Length ; $i++) { $result[$i]  | Should BeExactly $expected[$i]}
   }
+
```
JamesWTruher: if you approach this by removing the dynamic parameter on ! windows, we should probably have an explicit test for the parameter - windows => parameter present ; ! windows => parameter absent

src/Microsoft.PowerShell.Commands.Utility/Microsoft.PowerShell.Commands.Utility.csproj
daxian-dbw: I don't really have a preference although `csproj` approach seems more discoverable. However in this case, I don't think `Unblockfile.cs` should be completely excluded from Unix build as it will increase the number of `#if unix/endif` blocks in places where the code from this file is used, which will impact readability.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
adityapatwardhan: Closed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -4667,6 +4688,7 @@ private bool PerformCopyFileToRemoteSession(FileInfo file, string destinationPat
 
             bool targetSupportsAlternateStreams = RemoteTargetSupportsAlternateStreams(ps, remoteFilePath);
 
+#if !UNIX
```
PaulHigin: The PSCopyToSessionHelper script and PSCopyFromRemoteSession script for remote file copy contains support for alternate streams.  Ideally we would have versions of this script for Linux platforms that remove that support.  But I am not sure it should be part of this change.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -8260,6 +8284,7 @@ internal static bool IsSameFileSystemItem(string pathOne, string pathTwo)
 #endif
         }
 
+#if !UNIX
         internal static bool WinIsSameFileSystemItem(string pathOne, string pathTwo)
```
PaulHigin: This method can be made private access since it is only used here.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -8297,6 +8323,7 @@ internal static bool GetInodeData(string path, out System.ValueTuple<UInt64, UIn
             return rv;
         }
 
+#if !UNIX
         internal static bool WinGetInodeData(string path, out System.ValueTuple<UInt64, UInt64> inodeData)
```
PaulHigin: Likewise, should be private access.

PaulHigin (changes requested):

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: I think the other case I was fine with a bit of duplication, but prefer to leave this one as-is.

src/System.Management.Automation/namespaces/FileSystemContentStream.cs
SteveL-MSFT: Good idea, will change.

src/System.Management.Automation/namespaces/FileSystemContentStream.cs
```diff
@@ -854,7 +858,11 @@ private void CreateStreams(string filePath, string streamName, FileMode fileMode
             {
                 if (!String.IsNullOrEmpty(streamName))
                 {
+#if !UNIX
                     _stream = AlternateDataStreamUtilities.CreateFileStream(filePath, streamName, fileMode, requestedAccess, fileShare);
+#else
+                    throw new PlatformNotSupportedException();
```
SteveL-MSFT: ditto

src/System.Management.Automation/namespaces/FileSystemContentStream.cs
SteveL-MSFT: see @daxian-dbw comment below as better way to address this

src/System.Management.Automation/utils/ClrFacade.cs
SteveL-MSFT: There are some other checks this code does, however, upon closer inspection, I think they also don't apply to Unix so I'll `#ifdef` this code out on UNIX

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: So we still want the code compiled on UNIX?

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
SteveL-MSFT: will add

src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs
```diff
@@ -1304,7 +1304,9 @@ private static bool IsCmdletizationFile(RemoteDiscoveryHelper.CimModuleFile cimM
                     fullPath,
                     file.RawFileData);
 
+#if !UNIX
```
SteveL-MSFT: I believe CDXML cmdlets can work on Linux via OMI

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: I think that code cleanup should be part of a different PR.  Created https://github.com/PowerShell/PowerShell/issues/4601

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: will do

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: will do

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
SteveL-MSFT: Need to check if there are already tests using -stream otherwise I'll add

src/System.Management.Automation/namespaces/FileSystemProvider.cs
daxian-dbw: If that makes the code cleaner overall, then I'm fine to have the code compiled on Unix.

SteveL-MSFT: Looking at the existing code, the FileSystemProvider has a ton of `#if UNIX` and `#if !UNIX` already because the filesystems are inherently different (beyond just NTFS streams).  I think refactoring this code is an exercise best left for the future.

SteveL-MSFT:
I believe all feedback has been addressed

PaulHigin (approved):
LGTM

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -4667,6 +4688,7 @@ private bool PerformCopyFileToRemoteSession(FileInfo file, string destinationPat
 
             bool targetSupportsAlternateStreams = RemoteTargetSupportsAlternateStreams(ps, remoteFilePath);
```
dantraMSFT: Shouldn't this be in the #if !UNIX block?

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1146,10 +1147,12 @@ protected override void GetItem(string path)
...
                     }
                 }
+#endif
 
                 FileSystemInfo result = GetFileSystemItem(path, ref isContainer, false);
                 if (result != null)
                 {
+#if !UNIX
                     // If we want to retrieve the file streams, retrieve them.
                     if (retrieveStreams)
```
daxian-dbw: The alternate streams can be specified using `-Stream` or an inline stream syntax `"file::stream"`, the check of the inline stream syntax is done in 7 places in `FileSystemProvider.cs` but some of them are not within `#if !UNIX` block.

daxian-dbw: And when running `Get-Content -Path file::stream` on Linux, it probably should throw an error message about "alternate stream is not supported" instead of working unexpectedly or fail with other errors.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
daxian-dbw: I thought about it more and now have this question: which the following experience is better -- copy `Get-Content -Path <file> -Stream <stream>` to Linux powershell and see the error `"A parameter cannot be found"`, or `"Alternate stream is not supported on this platform"`? I think the latter message is better. To achieve this, maybe we can make the setter of the `Stream` property throw `PlatformNotSupportedException` on non-windows.

As for `#if UNIX/endif` blocks, IMHO, we don't necessarily need to have all codes that only apply to a specific platform guarded by the compile definition. I think it's OK to keep such codes around even for other platforms, and the guideline should be readability and maintainability. I still feel it's better to avoid most of the `#if UNIX` blocks by controlling the entry points:
1. Make sure `-Stream` not available on Unix platforms
2. Throw error when the inline stream syntax is used for `-Path` on Unix platforms
3. Skip the alternate stream discovery codes during remote copy (both `PerformCopyFileFromRemoteSession` and `PerformCopyFileToRemoteSession`) on Unix platforms.

After controlling those entry points (and put some comments for sure), it's safe to know that the alternate stream related code paths will not be hit on Unix platforms.

assets/Product.wxs
```diff
@@ -60,6 +60,30 @@
 
     <!--We need to show EULA, and provide option to customize download location-->        
     <Property Id="WIXUI_INSTALLDIR" Value="INSTALLFOLDER" />
```
TravisEz13: something isn't correct in your PR.  Already committed changes are in the diff.

TravisEz13 (dismissed):
Please work on correctly resolving merge/rebase issue.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: With the current changes, you don't get any error using `copy-item` going from Linux->Windows nor Windows->Linux

I created a new issue https://github.com/PowerShell/PowerShell/issues/4633 to clean up the code later since filesystemprovider.cs already has 21 `#if` statements before I made any changes

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: yes, will fix

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: On Unix, a colon is completely allowed in a filename, so we want that to continue working as just a filename.  Tests were added to make sure this works.

I'll fix other cases of using the `file:stream` syntax.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
SteveL-MSFT: will fix

SteveL-MSFT: will fix

TravisEz13:
It looks like the email on your latest commit is not associated with your github account.

SteveL-MSFT:
@TravisEz13 fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
lzybkr: Which is more likely - somebody copying a script from Windows to Linux that uses `-Stream`, or somebody on Linux using parameter completion, sees `-Stream` as a possible option, and wastes time trying to learn what it does?

I'm thinking it's better overall to not suggest parameters on platform if they aren't actually supported.

SteveL-MSFT: `Copy-Item` doesn't expose a `-Stream` param at all.  In the case of copying a NTFS file that contains alternate streams, those streams don't exist on the Linux destination.

SteveL-MSFT:
@TravisEz13 @daxian-dbw feedback addressed

daxian-dbw (approved):

Issue #4351, opened 2017-07-26
CLI: Ambiguous parameter names such as -no are misinterpreted as positional parameter values


mklement0:
Steps to reproduce
------------------

```powershell
# -no is ambiguous: could be -NoProfile, -NoLogo, ...
powershell -no -command get-date
```

Expected behavior
-----------------

An error complaining about the ambiguity of the parameter name and refusal to continue.

Actual behavior
---------------

PS Core and Windows PS behave differently, due to the recent Core change to default to `-File` instead of `-Command`:

* PowerShell Core:

```none
The argument '-no' to the -File parameter does not exist. Provide the path to an existing '.ps1' file as an argument to the -File parameter.
PowerShell v6.0.0-beta.4
Copyright (C) Microsoft Corporation. All rights reserved.
```

* Windows PowerShell:

```none
-no : The term '-no' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was
included, verify that the path is correct and try again.
At line:1 char:1
+ -no -command get-date
+ ~~~
    + CategoryInfo          : ObjectNotFound: (-no:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

Environment data
----------------

```powershell
PowerShell Core v6.0.0-beta.4 on macOS 10.12.5
PowerShell Core v6.0.0-beta.4 on Ubuntu 16.04.2 LTS
PowerShell Core v6.0.0-beta.4 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
Windows PowerShell v5.1.15063.413 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
```

SteveL-MSFT:
`-no` might be a valid script file, so I don't think we want check against ambiguous intended parameters (`-Command` didn't do this either).  However, we can improve the error message.  How about:

```
-no : The term `-no` is not recognized as a valid script file.  Check the spelling of the name, or if a path was
included, verify that the path is correct and try again.
```

mklement0:
@SteveL-MSFT:

For consistency and predictability, you should never assume that `-`-prefixed tokens are parameter _values_, _unless_ they follow a recognized, unambiguous parameter name (such as `-NoProfile`):

* If a _positional_ token _looks like_ a parameter name - i.e., if it starts with `-` followed by a _letter_ -  it should be interpreted as such.
[...]

PR #4573, opened 2017-08-14
Make error message consistent when invalid script is passed to -File, better error when passed ambiguous arg

SteveL-MSFT:
Since -File is now the positional parameter for powershell.exe, made the error message consistent with 
-Command when passed an invalid file.

If ambiguous arg is passed, give a better error message:

```powershell
PS C:\users\poker\repos\PowerShell> powershell -no -file foo
Invalid argument '-no', did you mean:

        -nologo
        -noexit
        -noprofile
        -noninteractive
PowerShell v6.0.0-beta.5-42-g3ed9a816b2452551aaad1ac1245a1cbb35e87612-dirty
Copyright (C) Microsoft Corporation. All rights reserved.
```

Enable -WindowStyle to work

Fix https://github.com/PowerShell/PowerShell/issues/4351

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
@@ -617,7 +617,7 @@ namespace Microsoft.PowerShell
                         break;
                     }
                 }
-#if !CORECLR  // windowstyle parameter not supported on NanoServer because ProcessWindowStyle does Not exist on CoreCLR
+#if !UNIX
                 else if (MatchSwitch(switchKey, "windowstyle", "w"))
                 {
                     ++i;
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
@@ -169,6 +169,30 @@ namespace Microsoft.PowerShell
 
     internal class CommandLineParameterParser
     {
+        internal static string[] validParameters = {
+            "psconsoleFile",
+            "version",
+            "nologo",
+            "noexit",
+#if !CORECLR
+            "sta",
+            "mta",
+#endif
+            "noprofile",
+            "noninteractive",
+            "inputformat",
+            "outputformat",
+#if !UNIX
+            "windowstyle",
+#endif
+            "encodedcommand",
+            "configurationname",
+            "file",
+            "executionpolicy",
+            "command",
+            "help"
+        };
+
         internal CommandLineParameterParser(PSHostUserInterface hostUI, string bannerText, string helpText)
         {
             if (hostUI == null) { throw new PSArgumentNullException("hostUI"); }
@@ -657,7 +681,7 @@ namespace Microsoft.PowerShell
                         WriteCommandLineError(
                             "The -module option can only be specified with the -iss option.",
                             showHelp: true,
-                            showBanner: true);
+                            showBanner: false);
                         break;
                     }
 
@@ -880,7 +904,7 @@ namespace Microsoft.PowerShell
                 WriteCommandLineError(
                     CommandLineParameterParserStrings.MissingFileArgument,
                     showHelp: true,
-                    showBanner: true);
+                    showBanner: false);
                 return false;
             }
 
@@ -924,15 +948,36 @@ namespace Microsoft.PowerShell
                 {
                     WriteCommandLineError(
                         string.Format(CultureInfo.CurrentCulture, CommandLineParameterParserStrings.InvalidFileArgument, args[i], exceptionMessage),
-                        showBanner: true);
+                        showBanner: false);
                     return false;
                 }
 
                 if (!System.IO.File.Exists(_file))
                 {
+                    if (args[i].StartsWith("-") && args[i].Length > 1)
+                    {
+                        string param = args[i].Substring(1, args[i].Length - 1).ToLower();
+                        StringBuilder possibleParameters = new StringBuilder();
+                        foreach (string validParameter in validParameters)
+                        {
+                            if (validParameter.Contains(param))
+                            {
+                                possibleParameters.Append("\n  -");
+                                possibleParameters.Append(validParameter);
+                            }
+                        }
+                        if (possibleParameters.Length > 0)
+                        {
+                            WriteCommandLineError(
+                                string.Format(CultureInfo.CurrentCulture, CommandLineParameterParserStrings.InvalidArgument, args[i]),
+                                showBanner: false);
+                            WriteCommandLineError(possibleParameters.ToString(), showBanner: false);
+                            return false;
+                        }
+                    }
                     WriteCommandLineError(
                         string.Format(CultureInfo.CurrentCulture, CommandLineParameterParserStrings.ArgumentFileDoesNotExist, args[i]),
-                        showBanner: true);
+                        showBanner: false);
                     return false;
                 }
 
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleControl.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleControl.cs
@@ -445,8 +445,7 @@ namespace Microsoft.PowerShell
         internal const int SW_FORCEMINIMIZE = 11;
         internal const int SW_MAX = 11;
 
-
-#if !CORECLR  // ProcessWindowStyle does Not exist on CoreCLR
+#if !UNIX
         /// <summary>
         /// Code to control the display properties of the a window...
         /// </summary>
--- a/src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
+++ b/src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
@@ -252,7 +252,7 @@ EXAMPLES
     <value>Processing -File '{0}' failed because the file does not have a '.ps1' extension. Specify a valid Windows PowerShell script file name, and then try again.</value>
   </data>
   <data name="ArgumentFileDoesNotExist" xml:space="preserve">
-    <value>The argument '{0}' to the -File parameter does not exist. Provide the path to an existing '.ps1' file as an argument to the -File parameter.</value>
+    <value>The argument '{0}' is not recognized as the name of a script file. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.</value>
   </data>
   <data name="BadArgsValue" xml:space="preserve">
     <value>Cannot process the command because the value specified with -EncodedArguments is not properly encoded. The value must be Base64 encoded.</value>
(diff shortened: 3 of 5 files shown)


src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
```diff
@@ -252,7 +252,7 @@ EXAMPLES
     <value>Processing -File '{0}' failed because the file does not have a '.ps1' extension. Specify a valid Windows PowerShell script file name, and then try again.</value>
   </data>
   <data name="ArgumentFileDoesNotExist" xml:space="preserve">
-    <value>The argument '{0}' to the -File parameter does not exist. Provide the path to an existing '.ps1' file as an argument to the -File parameter.</value>
+    <value>The term '{0}' is not recognized as the name of a script file. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.</value>
```
iSazonov: I believe `argument` is better `term` and we can be shorter:
```
The argument '{0}' is not recognized as the name of a script file.
```

SteveL-MSFT: Will change

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -930,6 +954,27 @@ bool TryGetBoolValue(string arg, out bool boolValue)
 
                 if (!System.IO.File.Exists(_file))
                 {
+                    if (args[i].StartsWith("-") && args[i].Length > 1)
+                    {
+                        string param = args[i].Substring(1, args[i].Length - 1);
+                        StringBuilder possibleParameters = new StringBuilder();
+                        foreach (string validParameter in validParameters)
+                        {
+                            if (validParameter.Contains(param))
```
lzybkr: Should this be `validParameter.StartsWith(param, StringComparer.OrdinalIgnoreCase)`?

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -930,6 +954,27 @@ bool TryGetBoolValue(string arg, out bool boolValue)
...
+                            {
+                                possibleParameters.Append("\n\t-");
```
lzybkr: I'd indent with spaces - we don't indent with tabs anywhere else.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -930,6 +954,27 @@ bool TryGetBoolValue(string arg, out bool boolValue)
...
+                                possibleParameters.Append(validParameter);
+                            }
+                        }
+                        if (possibleParameters.Length > 0)
+                        {
+                            WriteCommandLineError(
+                                string.Format(CultureInfo.CurrentCulture, CommandLineParameterParserStrings.InvalidArgument, args[i]),
+                                showBanner: false);
+                            WriteCommandLineError(possibleParameters.ToString(), showBanner: true);
```
lzybkr: Showing the banner after the error is weird and not like the other errors I reviewed.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: I was thinking to cover cases like: `-format`

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: will change

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: I'll add a test case for this

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: I thought that was weird in general, but it seems consistent with other uses of `WriteCommandLineError`.  I'm assuming it's to show that powershell still started if you started it within powershell since the prompt would look same/similar.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -169,6 +169,30 @@ public override void WriteWarningLine(string message)
...
     {
+        internal static string[] validParameters = {
+            "psconsoleFile",
+            "version",
+            "nologo",
+            "noexit",
+#if !CORECLR
+            "sta",
+            "mta",
+#endif
```
iSazonov: `#if !CORECLR ` is no longer relevant. Should we remove the parameters at all?

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleControl.cs
```diff
@@ -446,7 +446,6 @@ internal enum KeyboardFlag : uint
         internal const int SW_MAX = 11;
 
 
-#if !CORECLR  // ProcessWindowStyle does Not exist on CoreCLR
```
iSazonov: Maybe 
`#if !UNIX`

src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
```diff
@@ -269,4 +269,7 @@ EXAMPLES
   <data name="MissingConfigurationNameArgument" xml:space="preserve">
     <value>Cannot process the command because -Configuration requires an argument that is a remote endpoint configuration name. Specify this argument and try again.</value>
   </data>
+  <data name="InvalidArgument" xml:space="preserve">
+    <value>Invalid argument '{0}', did you mean:</value>
```
iSazonov: "did you mean:" - Do you want add something else?

test/powershell/Host/ConsoleHost.Tests.ps1
```diff
@@ -492,6 +492,85 @@ foo
...
+                $outString | Should Match $expectedMatch
+            }
+        }
+    }
+}
+
+Describe "WindowStyle argument" -Tag Feature {
+    BeforeAll {
+        $ExitCodeBadCommandLineParameter = 0xFFFD0000
+        Add-Type -Name User32 -Namespace Test -MemberDefinition @"
```
iSazonov: We should exclude this on Unix.

test/powershell/Host/ConsoleHost.Tests.ps1
```diff
@@ -492,6 +492,85 @@ foo
...
+{
+    Hidden = 0,
+    Normal = 1,
+    Minimized = 2,
+    Maximized = 3,
+}
+"@
+    }
+
+    It "-WindowStyle <WindowStyle> should work on Windows" -Skip:(!$IsWindows) -TestCases @(
```
iSazonov: All tests is excluded on Unix so we could use our pattern "$PSDefaultParameterValues["it:skip"] = ..."

iSazonov (commented):
Leave a comment

src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
SteveL-MSFT: The possible matches show up after this.  See example above in the PR description.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
SteveL-MSFT: I'm fixing that as part of enabling ApartmentState PR and not part of this PR

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleControl.cs
SteveL-MSFT: makes sense

test/powershell/Host/ConsoleHost.Tests.ps1
SteveL-MSFT: will do

test/powershell/Host/ConsoleHost.Tests.ps1
SteveL-MSFT: agree

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
iSazonov: Closed.

src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
iSazonov: Closed.

test/powershell/Host/ConsoleHost.Tests.ps1
```diff
@@ -492,6 +492,94 @@ foo
...
+                $outString | Should Match $expectedMatch
+            }
+        }
+    }
+}
+
+Describe "WindowStyle argument" -Tag Feature {
+    BeforeAll {
+        $defaultParamValues = $PSDefaultParameterValues.Clone()
+        $PSDefaultParameterValues["it:skip"] = (!$IsWindows)
```
iSazonov: Parentheses can be removed.

SteveL-MSFT: Will remove

SteveL-MSFT:
Not sure why some of the new tests fail in AppVeyor but not on my desktop

iSazonov:
No host?

SteveL-MSFT:
Figured out why it's failing on non-Windows.  It appears that exit codes from programs have to be in the [range 0-255 and anything greater than 255 is mod 256](http://www.tldp.org/LDP/abs/html/exitcodes.html).  I verified this with a simple console c# app.  In this case, the value for ExitCodeBadCommandLineParameter mod 256 == 0.  Hmmm.

Although a breaking change, I think we should conform to [libc standards](https://opensource.apple.com/source/Libc/Libc-320/include/sysexits.h) which means this error code should be 64 "command line usage error".  We should also update all the other error codes that powershell console returns.

In general, I don't think this should be a problem as typically 0 means success and non-0 means error so although a breaking change, I suspect it have limited real world impact.

cc @PowerShell/powershell-committee

lzybkr (approved):
I approve the changes, but there are 2 distinct fixes in the PR, so I think there should be 2 distinct commits.

Interactive add (`git add -p`) is an easy way to do that.

SteveL-MSFT:
@lzybkr are you referring to -WindowStyle and everything else as two commits?

SteveL-MSFT:
@lzybkr sure, I can make that two commits in the same PR.

SteveL-MSFT:
Split into 3 commits (Thanks @lzybkr for the tip on `git add -p`!)

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and is ok with changing the exit codes

iSazonov (approved):

TravisEz13:
@iSazonov and @SteveL-MSFT  thanks as always!

PR #4581, opened 2017-08-15
Remove workflow related modules

daxian-dbw:
Related to #4357 

Remove workflow related modules (`PSWorkflow` and `PSWorkflowUtility`) from `Modules\windows-full`.

Note that folders `Modules\windows-full` and `Modules\windows-core-full` are kept because:
1. Some modules in it may still be ported to powershell core in future, such as `PSScheduledJob`
1. It's easy to compare and see which cmdlets are missing from powershell core.


adityapatwardhan (approved):
:shipit:

SteveL-MSFT:
Don't merge yet along with all the activities ones until @PowerShell/powershell-committee has closed on workflow

PR #4589, opened 2017-08-15
Remove src\Microsoft.PowerShell.Workflow.ServiceCore

daxian-dbw:
Related to #4357


adityapatwardhan (approved):

iSazonov:
Is Workflow depredicated?

SteveL-MSFT:
Don't remove Workflow code until @PowerShell/powershell-committee has closed on this

SteveL-MSFT:
@PowerShell/powershell-committee is fine with proceeding to remove the workflow code.  PowerShell workflow is deprecated.  @joeyaiello probably needs another blog post...  Although CoreCLR has support for workflow, it doesn't support XAML which is what all the PowerShell workflow code depends upon.  Our data indicates that most customers use PowerShell workflow specifically for parallel execution.  We have an [RFC](https://github.com/PowerShell/PowerShell-RFC/blob/master/1-Draft/RFC0023-Native-Support-for-Concurrent-Programming-in-PowerShell.md) to make it easier with PowerShell script so we are ok removing support entirely.

Issue #4458, opened 2017-08-01
MSI package should perform pre-req checks


SteveL-MSFT:
Windows Package has some [pre-reqs](https://github.com/PowerShell/PowerShell/blob/master/docs/installation/windows.md#prerequisites) required.  MSI package should give a good error message directing user to download and install pre-reqs rather than failing at runtime with an unactionable error message.

bergmeister:
OK, so to summarize:

- [Universal C Runtime](https://www.microsoft.com/en-us/download/details.aspx?id=50410) prior to Windows 10
- Visual C++ Redistributable for VS2015 (the [link ](https://my.visualstudio.com/Downloads?pid=2082)to it in the [pre-reqs](https://my.visualstudio.com/Downloads?pid=2082) does not work for people without an MSDN subscription by the way)

@SteveL-MSFT @adityapatwardhan  I think I might be able to help with the 2nd item by following [this](http://wixtoolset.org/documentation/manual/v3/howtos/redistributables_and_install_checks/install_vcredist.html) approach from the official WiX guideline, which is to include and install the Visual++ as part of the installer. WDYT?

adityapatwardhan:
@bergmeister I think that might work. Thanks!

bergmeister:
Due to the fact that the Visual C++ team has made significant refactorings in 2015 (see [here](https://blogs.msdn.microsoft.com/vcblog/2015/03/03/introducing-the-universal-crt/)), the initial approach of including them as a merge module in the MSI has been deprecated. The alternatives are complex and therefore I think it is best to go back to just giving an error message as initially proposed by Steve.

bergmeister:
@adityapatwardhan : I created the pull request #4602 for it.

SteveL-MSFT:
There's an open issue https://github.com/PowerShell/PowerShell/issues/4665 for this, but we can consider this one closed.

PR #4602, opened 2017-08-16
MSI installer to perform pre-requisite checks

bergmeister:
Implemented #4458: Checks that the following [pre-requisites](https://github.com/PowerShell/PowerShell/blob/master/docs/installation/windows.md#prerequisites) of PowerShell Core on Windows are met by displaying an error message that includes a download link:
- [Universal C Runtime](https://www.microsoft.com/en-us/download/details.aspx?id=50410) prior to Windows 10: A check for the existence of ucrtbase.dll in the system32 folder is performed (it would be possible to additionally assert against a minimum file version if you wish so)
- [Visual C++ Redistributable for VS2015](https://www.microsoft.com/en-gb/download/details.aspx?id=48145): Registry entries are checked. The 64 bit installer checks for the 32 and 64 bit version of it whereas the 32 bit installer only asserts against the 32 bit version of the redistributables.

The ideas against what to check comes from research and especially this example [here](https://github.com/dokan-dev/dokany/blob/master/dokan_wix/Dokan_x64.wxs) to which I was redirected in a WiX discussion.

Includes tests to check that the WiX file contains the download URLs and those URLs are not dead.

I'll have to leave the extensive testing on different systems up to you. Maybe it is also best if you double check the used assertions with the right experts at MSFT.

NB: Due to recent refactorings of the VC++ team (see blog post [here](https://blogs.msdn.microsoft.com/vcblog/2015/03/03/introducing-the-universal-crt/)) it has become very complicated to include those binaries and therefore a check against them should be the minimum viable solution as originally outlined by @SteveL-MSFT


adityapatwardhan:
@bergmeister Thanks!

assets/Product.wxs
```diff
@@ -60,6 +60,30 @@
...
+    </Property>	
+    <Property Id="VCREDISTINSTALLEDX64" Secure="yes">
+      <RegistrySearch Id="VCRedist_RegSearch_x64" Root="HKLM" Name="Version" Type="raw" Win64="yes"
+                      Key="SOFTWARE\Microsoft\DevDiv\VC\Servicing\14.0\RuntimeMinimum"/>
+    </Property>
+    <Property Id="VCREDISTINSTALLEDX86" Secure="yes">
+      <RegistrySearch Id="VCRedist_RegSearch_x86" Root="HKLM" Name="Version" Type="raw" Win64="no"
+                      Key="SOFTWARE\Microsoft\DevDiv\VC\Servicing\14.0\RuntimeMinimum"/>
+    </Property>
+    <Condition Message="$(env.ProductName) requires because the Universal C Runtime to be installed. You can download it here: https://www.microsoft.com/en-us/download/details.aspx?id=50410">
```
mirichmo: Typo in the error string. Please remove "because."

mirichmo (changes requested):

assets/Product.wxs
bergmeister: Well spotted. I fixed it.

bergmeister:
Should I also update the link to the Visual Studio C++ redistributes in the [pre-requisites](https://my.visualstudio.com/Downloads?pid=2082) documentation for consistency and also because the link in the documentation requires an MSDN subscription?

TravisEz13:
If you have a good alternative, that would be great.

TravisEz13:
@mirichmo can you update your review?

assets/Product.wxs
```diff
@@ -60,6 +60,30 @@
...
+                      Key="SOFTWARE\Microsoft\DevDiv\VC\Servicing\14.0\RuntimeMinimum"/>
+    </Property>
+    <Property Id="VCREDISTINSTALLEDX86" Secure="yes">
+      <RegistrySearch Id="VCRedist_RegSearch_x86" Root="HKLM" Name="Version" Type="raw" Win64="no"
+                      Key="SOFTWARE\Microsoft\DevDiv\VC\Servicing\14.0\RuntimeMinimum"/>
+    </Property>
+    <Condition Message="$(env.ProductName) requires the Universal C Runtime to be installed. You can download it here: https://www.microsoft.com/en-us/download/details.aspx?id=50410">
+      <![CDATA[Installed OR UCRTINSTALLED]]>
+    </Condition>
+    <Condition Message="$(env.ProductName) requires the Visual Studio C++ 2015 x64 redistributables to be installed. You can download it here: https://www.microsoft.com/en-gb/download/details.aspx?id=48145">
```
mirichmo: It's close. I overlooked this the first time. Use a link without the language specifier so that it uses the browser's default language. For example:
https://www.microsoft.com/download/details.aspx?id=48145 automatically redirects to en-us for me.

This should also be fixed in the other files as well.

mirichmo (changes requested):

mirichmo (approved):
Thanks

assets/Product.wxs
bergmeister: OK, makes sense, especially for non-English speaker. But interestingly when I omit the language specifier, I get redirected to 'en/us' as well and not 'en/gb'. I fixed that in the files that I changed as part of this PR. However, note that the whole source code contains around 52 occurrences of '/en-us' in case you want to raise a separate issue for that.

mirichmo: I created issue #4610 to cover it.

Issue #4215, opened 2017-07-11
Add ability to create a temporary file with a specific extension to New-TemporaryFile


mklement0:
<sup>Related: #3442</sup>

Currently, `New-TemporaryFile` invariably creates temp. files with extension `.tmp`.

Sometimes, however, it is helpful to create temp. files with a specific filename extension.

Adding an `-Extension` parameter could provide that:

Desired behavior
------------------

```powershell
> New-TemporaryFile -Extension .ps1

    Directory: C:\Users\jdoe\AppData\Local\Temp

Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        7/10/2017   9:53 PM              0 tmpE3B8.ps1

```

Environment data
----------------

```powershell
PowerShell Core v6.0.0-beta.3
```

iSazonov:
What practical scenarios exist when do we need to change the extention?

mklement0:
Whenever you need to create a temporary file to be processed by a utility that requires a specific filename extension.

Using PowerShell itself as an example, say you want to create a throw-away script dynamically, whose path you want to pass to another PowerShell instance via the CLI (`powershell -File ...`).  
Such a throw-away script only works if it has extension `*.ps1`.

SteveL-MSFT:
@mklement0 with a recent PR, if you use -File, it doesn't have to have a .ps1 extension :)

mklement0:
@SteveL-MSFT: Good point; I should have used a different example, such as `*.csv` for files to be processed by Excel, for instance.

bergmeister:
I opened the pull request above for it. Please ignore the commits above except for the last one as I squashed the commits in my branch into one commit at the end.

mklement0:
Thanks, @bergmeister!

SimonWahlin:
PR #4612 was closed, waiting for functionality in CoreFX.
Will this cover both extension and total number of files cap?
Where can I vote for the API mentioned in [https://github.com/PowerShell/PowerShell/pull/4612#issuecomment-367381354](https://github.com/PowerShell/PowerShell/pull/4612#issuecomment-367381354)?

bergmeister:
The corefx issue is here:
https://github.com/dotnet/corefx/issues/24001
The file cap limit is due to the limitation of the Windows API being used (and the short file name) for legacy reasons to avoid breaking changes. One would hope so that the new overload does not make use of that API any more, but please discuss this in the corefx repo/issue.

[...]

PR #4612, opened 2017-08-18
Implement -Extension parameter in New-TemporaryFile cmdlet

bergmeister:
#4215: Add ability to create a temporary file with a specific extension to `New-TemporaryFile`
Implemented exactly as described by adding an optional `-Extension` parameter to specify the extension e.g. as `.csv` or just `csv`.
Because of the required change to call different .NET APIs, the cmdlet had to be slightly rewritten to cater for the special case that the file to be created already exists in the temp folder.


src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -12,18 +12,57 @@ namespace Microsoft.PowerShell.Commands
...
+                WriteError(
+                    new ErrorRecord(
+                            new IOException("Could not find an available temporary file name in {tempPath}."),
+                            "NewTemporaryFileResourceUnavailable",
+                            ErrorCategory.ResourceUnavailable,
+                            tempPath));
+                    return;
+            }
+
+            if (ShouldProcess(filePath))
```
mklement0: While this implicitly fixes the bug in the original code that referred to environment variable `TEMP`, which can only be assumed to exist in Windows, note that you've changed the logic from reporting the _directory_ path to the _file_ path.

In other words: to remain consistent with how thing worked before, it should remain `if (ShouldProcess(tempPath))`

I guess the logic is: `New-TemporaryFile` operates on the temp. _directory_ as its target, and creates a new file _inside it_.

```
What if: Performing the operation "New-TemporaryFile" on target "C:\Users\jdoe\AppData\Local\Temp".
```

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -12,18 +12,57 @@ namespace Microsoft.PowerShell.Commands
...
-            if (ShouldProcess(tempPath))
+            string tempPath = Path.GetTempPath();
+            int attempts = 0;
+            bool temporaryFilePathFound = false;
+            while(attempts++ < 10 && !temporaryFilePathFound)
+            {
+                string fileName = Path.GetRandomFileName();
+                fileName = Path.ChangeExtension(fileName, _extension);
+                filePath = Path.Combine(tempPath, fileName);
+                if(!File.Exists(filePath))
```
mklement0: Unfortunately, you can run into race conditions with this approach: Between the time you check for the existence of a file with a given name and a subsequent attempt to create it, someone else could have created such a file, at least hypothetically.

The more robust approach is to attempt to create the file and let that attempt fail if a file with that name happens to exist already (which is what file mode `CreateNew` does), catch the resulting `IOException` exception, and try again.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
bergmeister: Ok. Makes sense.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -32,48 +32,53 @@ protected override void EndProcessing()
...
+                                ErrorCategory.WriteError,
+                                tempPath));
+                        return;
+                    }
                 }
-                catch (Exception e)
+
+                if (!creationOfFileSuccessful)
                 {
                     WriteError(
```
mklement0: Thanks for the quick turnaround. 

The following, which predates your PR, may not be worth doing, because it is technically a breaking change, but it's worth pointing out: 

Based on the official cmdlet error-reporting guidelines, this should be a (statement)-_terminating_ error: "Cmdlets that can accept only 0-1 input objects and generate only 0-1 output objects should treat errors as terminating errors and generate terminating exceptions." - https://msdn.microsoft.com/en-us/library/ms714412%28v=vs.85%29.aspx  

cc @SteveL-MSFT

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
bergmeister: I was aware of this when writing it but did not consider handling it specifically since the global catch would have still handled it and the probability of having a clash of a random file name in such a short time period is really low. However, I liked your idea and as your approach does not seem to have a downside, I implemented your suggestion.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
bergmeister: Thanks for the quick review as well. 👍  You have an interesting point. My thinking is: If one has got a script in which a temporary file gets created, then in order to use it, one would need need to use the returned object because otherwise one does not know where the temporary file is. Therefore it is up to the user to check that that their variable is not `$null` and handle it or otherwise the next operation will probably throw a NullReference exception and terminate anyway. But I'll leave this decision of course up to you guys.

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
```diff
@@ -21,11 +21,25 @@ Describe "NewTemporaryFile" -Tags "CI" {
 
         Test-Path $tempFile | Should be $true
         $tempFile | Should BeOfType System.IO.FileInfo
+        $tempFile.Extension | Should be '.tmp'
 
-        if(Test-Path $tempFile)
-        {
-            Remove-Item $tempFile -ErrorAction SilentlyContinue -Force
-        }
+        Remove-Item $tempFile -ErrorAction SilentlyContinue -Force
```
dantraMSFT: Suggestion: To make cleanup reliable, use 'try' around the Should statements and place 'Remove-Item' in the finally block.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -12,34 +12,73 @@ namespace Microsoft.PowerShell.Commands
     [OutputType(typeof(System.IO.FileInfo))]
     public class NewTemporaryFileCommand : Cmdlet
     {
+        private string _extension = ".tmp";
```
PaulHigin: Please remove the _extension field and just use the property:
```
[Parameter(Position = 0)]
[ValidateNotNullOrEmpty]
public string Extension
{
    get;
    set;
} = ".tmp";
```

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -12,34 +12,73 @@ namespace Microsoft.PowerShell.Commands
...
-            string tempPath = System.Environment.GetEnvironmentVariable("TEMP");
+            string tempPath = Path.GetTempPath();
+
             if (ShouldProcess(tempPath))
             {
-                try
+                int attempts = 0;
+                bool creationOfFileSuccessful = false;
+
+                // In case the random temporary file already exists, retry
```
PaulHigin: I am not convinced that this retry loop is necessary.  What are the odds that the random file name already exists?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
PaulHigin: I agree that with @mklement0 a terminating error should be thrown.  But it would be a breaking change and so we should continue with the existing pattern.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -12,34 +12,73 @@ namespace Microsoft.PowerShell.Commands
...
+
+                if (!creationOfFileSuccessful)
                 {
                     WriteError(
                         new ErrorRecord(
-                            e,
-                            "NewTemporaryFileWriteError",
-                            ErrorCategory.WriteError,
-                            tempPath));
+                            new IOException("Could not find an available temporary file name in {tempPath}."),
```
PaulHigin: This exception message string should go into a resx file so it can be localized.  I think this would be the right file: Microsoft.PowerShell.Commands.Utility\resources\UtilityCommonStrings.resx

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -12,34 +12,73 @@ namespace Microsoft.PowerShell.Commands
     [OutputType(typeof(System.IO.FileInfo))]
     public class NewTemporaryFileCommand : Cmdlet
```
PaulHigin: Please add a comment section here that describes this public property.

PaulHigin (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
iSazonov: It is CoreFX default - we can exclude this and simply check that the parameter presents.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
iSazonov: We can use approved pattern https://github.com/PowerShell/PowerShell/blob/02b5f357a20e6dee9f8e60e3adb9025be3c94490/src/System.Management.Automation/utils/PathUtils.cs#L420

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -12,34 +12,73 @@ namespace Microsoft.PowerShell.Commands
...
     {
+        private string _extension = ".tmp";
+
+        /// <summary>
+        /// Specify a different file extension other than the default one, which is '.tmp'. The period in this parameter is optional
+        /// </summary>
+        [Parameter(Position = 0)]
+        [ValidateNotNullOrEmpty]
+        public string Extension
```
dantraMSFT: Consider validating the value in the setter for valid characters instead of rely only on an exception in your EndProcessing implementation.  Something like...

if (!string.IsNullOrEmpty() && value.IndexOfAny(Path.GetInvalidFileNameChars() != -1)
{
 throw...
}

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
bergmeister: I was thinking the same but rather followed the existing style. What about a more elegant one-liner?
```
public string Extension { get; set; } = ".tmp";
```

iSazonov: We can directly use `Path.GetRandomFileName()` if the parameter absent.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
mklement0: Understood, @PaulHigin. 

@bergmeister: Just to clarify:  by default, the script would continue to run even if a terminating error were reported with `ThrowTerminatingError()`, because only the _statement_ is terminated; the difference between reporting a non-terminating vs. a terminating error only matters with respect to `-ErrorAction` use and `try` / `catch` use.

Since the conceptual fog around error handling lifted for me personally only recently, perhaps you'll find this useful too: [Our Error Handling, Ourselves - time to fully understand and properly document PowerShell's error handling](https://github.com/PowerShell/PowerShell-Docs/issues/1583)

iSazonov: Another point is to follow the standard, make the breaking change and "fix" it in our interop module.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
bergmeister: The odds are low but we could make them really low by using e.g. Guids for the file name instead of `Path.GetRandomFileName()` because the latter is only 8 characters long...

iSazonov: I think we should follow CoreFX while we can and fix issues there.

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
bergmeister: Ok. Will do.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
bergmeister: I disagree with using `Path.GetRandomFileName()` in the special case the Extension parameter not being present because this function does not only give you a random file name but already creates the file for you. Therefore it would highly increase the complexity of the code because then one ends up with 2 different ways of generating the file and having to handle the errors differently in 2 places. I strongly believe in having only one common code path. Can you please re-consider?

src/Microsoft.PowerShell.Commands.Utility/resources/UtilityCommonStrings.resx
```diff
@@ -174,4 +174,10 @@
   <data name="TypeNotSupported" xml:space="preserve">
     <value>'{0}' is not supported in this system.</value>
   </data>
+  <data name="InvalidCharacterInParameter" xml:space="preserve">
+    <value>Parameter '{0}' contains at least one invalid character. It's value is: '{1}'</value>
```
mklement0: Please change "it's" to "its".

bergmeister: Oops. Sure.

bergmeister:
I have addressed the main suggestions but would like to wait until we have reached a consensus on the preferred approaches with the issues of the possibility of not being able to create a temp file and how to deal with it

- We could use a `Guid` for the filename to minimise the probability of a file name clash instead of `Path.GetRandomFileName()`. @iSazonov has suggested to use [this](https://github.com/PowerShell/PowerShell/blob/02b5f357a20e6dee9f8e60e3adb9025be3c94490/src/System.Management.Automation/utils/PathUtils.cs#L420) existing pattern as an alternative but I struggle to see why as it suffers from the same problem and also does not deal with other exceptions due to e.g. insufficient permissions, disk space. Maybe you can elaborate more on it?
- We could simply let it fail and not have a retry mechanism

I think in general it is good that we move away from `Get-TempFileName` due to it's limitations (see e.g.
 [here](https://github.com/dotnet/cli/issues/6995)) but I am wondering if this is not a also a case for the coreclr to improve the current implementation and possible offer an extension overload...

PaulHigin:
A GUID based filename would certainly solve the problem but after thinking about it some more I am Ok with the retry loop.  It feels like a reasonable compromise to write an error and fail after 10 tries.  Also I am fine with initializing the new property with ".tmp".  Using GUIDs feels like overkill.  But maybe we could add a switch to allow it as an option.  Something like:
```powershell
New-TemporaryFile -GuidBasedName
```

bergmeister:
Yes, we could easily add a `GuidBasedName` without making the code complicated. But I am wondering if we are not over-engineering this cmdlet and trying to solve problems that have not been reported yet? One should also keep in mind that users still have the full set of .Net methods available to handcraft something that meets their special needs in a few lines of code. I was also thinking about the problem on a higher level. Should someone not be able to create a temporary file and would need a Guid based approach, then to me this is either:
- Bad luck -> We have the retry logic for that
- Their temp folder is so full of temp files that there aren't many possible file names left -> There seems to be a more serious problem that needs to be addressed and therefore it might be good to see an error before files worth of TB are written.
Just let me know what you want me to do at the end.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
iSazonov: @lzybkr Could you please comment - should we make the breaking change or exclude it?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
iSazonov: Yes, I meant that
```c#
if (Extension parameter absent)
{
    return  Path.GetFileName()
}
else
{
   Here new code to create a temp file with explicit extension.
}
```
We should follow CoreFX if we can. Here we can. `Extension` parameter will be used rarely so it is good two code paths.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
iSazonov: Core FX generates unique file name but not extension. If we replace extension we're going to get an exception with the same probability. So I don't see the need to create a new algorithm.

iSazonov:
@PaulHigin 
>A GUID based filename would certainly solve the problem but after thinking about it some more I am Ok with the retry loop.

Why do you think this problem exists?
Why hasn't this been corrected in .Net Framework in the last 15 years and has not been corrected in CoreFX now?
Do we actually have scripts that create more than 50000 temporary files without removing them?
Are there real support cases on this problem?
Does anyone really need temporary file names based on GUIDs?

If so, right way is - we have to open Issue in CoreFX and fix the problem there without complicating PowerShell Core code.

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
```diff
@@ -15,19 +15,100 @@
...
+        try
+        {
+            
+            Test-Path $tempFile | Should be $true
+            $tempFile | Should BeOfType System.IO.FileInfo
+            $tempFile.Extension | Should be $defaultTemporaryFileExtension
+        }
+        finally
+        {
+            Remove-Item $tempFile -ErrorAction SilentlyContinue -Force
```
iSazonov: We use this multiple times - it is good candidate for `AfterEach` .

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
```diff
@@ -15,19 +15,100 @@
...
+        foreach($invalidFileNameChar in $invalidFileNameChars)
+        {
+            try
+            {
+                New-TemporaryFile -Extension $invalidFileNameChar -ErrorAction Stop
+                throw "No Exception!"
+            }
+            catch
+            {
+                $_.FullyQualifiedErrorId | Should Be "NewTemporaryInvalidArgument,Microsoft.PowerShell.Commands.NewTemporaryFileCommand"
```
iSazonov: Currently we use our pattern:
```powershell
{ New-TemporaryFile -Extension $invalidFileNameChar -ErrorAction Stop } | ShouldBeErrorId "NewTemporaryInvalidArgument,Microsoft.PowerShell.Commands.NewTemporaryFileCommand"
```

iSazonov:
After further thinking, I believe that if the CoreFX algorithm is bad, it must be corrected in CoreFX, not here.
So our code should be simple:
```c#
try
{
    var tempFile = GetTempFileName();
    if (parameter Extension present)
    {
         replace extension in tempFile
    }

    WriteObject(tempFile);
}
catch
{
    WriteError(...) or write terminating error;
}
```

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -2,44 +2,107 @@
...
+        [Parameter()]
+        public SwitchParameter GuidBasedName  { get; set; }
+
+        /// <summary>
+        /// Returns a TemporaryFile.
         /// </summary>
         protected override void EndProcessing()
         {
+            // Check for invalid characters in extension
+            if (!string.IsNullOrEmpty(Extension) && Extension.IndexOfAny(Path.GetInvalidFileNameChars()) != -1)
```
iSazonov: I guess it makes sense if we want to fix the problem somehow, but we just write a error message. So it's best to just follow CoreFX, where this check is done and just handle the exception. As a result, we will have only one block `try-catch` - see [my post](https://github.com/PowerShell/PowerShell/pull/4612#issuecomment-323577938)

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
bergmeister: Ok, can do. I was following the PowerShell Pester guideline [here](https://github.com/PowerShell/PowerShell/blob/68bcd4b5284bb169c52a570afbfdd1bb70d0de10/docs/testing-guidelines/WritingPesterTests.md). Should the guideline then be updated as well?

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
bergmeister: Ok. But I will wait with updating the tests for the moment as it is more important to know the final decision first whether the implementation would get accepted or if it requires a fix in corefx first.

bergmeister:
@iSazonov In the beginning I was thinking that a fix in corefx would be nicer as well since this is a common problem that also others have (see e.g. [here](https://github.com/dotnet/cli/issues/6995)).
If you look at my links above to the actual implementations, then you can see that it would be trivial to improve the corefx algorithms (Windows & Unix) to have an overload with a non-default extension but it would require more work to improve the file name uniqueness limitation to 65,535 files due to the Windows API that it calls at the end, etc.
The reason why I was hesitating to ask for a change in corefx is because corefx is probably not very happy with adding an overload that is not part of .net standard (which would then also require an implementation of the new overload in Xamarin and full .Net). What do others think?

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
iSazonov: Yes. Could you please open new Issue for that?

bergmeister: I'd prefer if you open the issue because the person opening an issue or PR is implicitly responsible for arguing and leading the discussion whether the guideline is out of date and/or should be changed you are definitely the better person for that.

iSazonov: >person opening an issue or PR is implicitly responsible for arguing and leading the discussion 

No, we welcome any error report - you don't have to "leading the discussion". Every time you see a bug - feel free to open new Issue (and forget it :smile:) MSFT experts will help to open a Issue if the issue is difficult to understand.

iSazonov:
Sorry for being unclear - I meant that it is better to fix "the file name uniqueness limitation to 65,535 files" algorithm in CoreFX not extension - adding new overload for extension require .Net Standard change - it is very long process. Also it will be not approved without real business case - mentioned issue is unuseful because don't contains any arguments.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
lzybkr: I think a terminating error is appropriate and an unlikely breaking change, but we can make that change in a different PR.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -2,44 +2,107 @@
...
+                        {
+                            fileName = Guid.NewGuid().ToString();
+                        }
+                        else
+                        {
+                            fileName = Path.GetRandomFileName();
+                        }
+                        fileName = Path.ChangeExtension(fileName, Extension);
+                        filePath = Path.Combine(tempPath, fileName);
+                        using (new FileStream(filePath, FileMode.CreateNew)) { }
```
lzybkr: A comment would be good here - explaining why we create the file and do not remove it, similar to `Path.GetTempFileName`.

src/Microsoft.PowerShell.Commands.Utility/resources/UtilityCommonStrings.resx
```diff
@@ -174,4 +174,10 @@
   <data name="TypeNotSupported" xml:space="preserve">
     <value>'{0}' is not supported in this system.</value>
   </data>
+  <data name="InvalidCharacterInParameter" xml:space="preserve">
+    <value>Parameter '{0}' contains at least one invalid character. Its value is: '{1}'</value>
+  </data>
+  <data name="CouldNotFindTemporaryFilename" xml:space="preserve">
```
lzybkr: `Find` doesn't seem quite right, I would use `Create`, and change the message to `Could not create a temporary file name in {0}.`

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
lzybkr: `Extension` can't be null or empty because of `ValidatenotNorOrEmpty`, so this test is not necessary.

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
```diff
@@ -15,19 +15,100 @@
...
+        finally
+        {
+            Remove-Item $tempFile -ErrorAction SilentlyContinue -Force
+        }
 
-        if(Test-Path $tempFile)
+        $tempFile = New-TemporaryFile 'csv' # check that one can also omit the period and parameter name
+        try
+        {
+            Test-Path $tempFile | Should be $true
```
lzybkr: Shouldn't this be `$tempFile | Should Exist`?

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
```diff
@@ -15,19 +15,100 @@
...
-        Test-Path $tempFile | Should be $true
-        $tempFile | Should BeOfType System.IO.FileInfo
+    It "creates a new temporary file with a specific extension using the -Extension parameter" {
+        $expectedExtension = '.csv'
+        
+        $tempFile = New-TemporaryFile -Extension $expectedExtension
+        try
+        {
+            
+            Test-Path $tempFile | Should be $true
```
lzybkr: Shouldn't this be `$tempFile | Should Exist`?

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
```diff
@@ -15,19 +15,100 @@
...
+        }
+        finally
+        {
+            Remove-Item $tempFile -ErrorAction SilentlyContinue -Force
+        }
+    }
 
-        Test-Path $tempFile | Should be $true
-        $tempFile | Should BeOfType System.IO.FileInfo
+    It "creates a new temporary file with a specific extension using the -Extension parameter" {
```
lzybkr: This test has repeated code, you should use the `-Testcases` feature of Pester instead.

lzybkr (changes requested):
If this is merged, the documentation of `New-TemporaryFile` will need to be updated as it describes in some detail the current algorithm.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -2,44 +2,107 @@
...
+                        creationOfFileSuccessful = true;
+                        WriteVerbose("Created temporary file {filePath}.");
```
bergmeister: Yes. Thanks

src/Microsoft.PowerShell.Commands.Utility/resources/UtilityCommonStrings.resx
bergmeister: OK, can do. I wasn't 100% happy with the error message myself either.

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
bergmeister: I was following the pattern of an existing test but you are right that `Should Exist` is better.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
SteveL-MSFT: Is there already an issue for the terminating error change?

bergmeister: @SteveL-MSFT Now there is: [here](https://github.com/PowerShell/PowerShell/issues/4634)

iSazonov:
@lzybkr Could you please clarify about used algorithm (my comments [here](https://github.com/PowerShell/PowerShell/pull/4612#issuecomment-323546991) and [here](https://github.com/PowerShell/PowerShell/pull/4612#issuecomment-323577938))? I still think we're making an unreasonable complication trying to "fix" CoreFX without justifying and business case and adding ` -UseGuid`.

lzybkr:
Oh, I think `-UseGuid` is overkill, but the retry logic is necessary at some layer, and in the case of wanting an extension that differs from what .Net uses, we can't simply replace the extension, .Net guarantees the original filename is safe, but not the basename + any other extension.

bergmeister:
@PaulHigin @lzybkr I have made all the requested changes for the test syntax, error message and removed the switch parameter for the Guid file name. The only thing that I noticed at the end was the fact that the cmdlet has always always (before this PR) overridden the `EndProcessing()` method and I was wondering if it should not rather be `ProcessRecord()`?

lzybkr:
For cmdlets that do not accept any pipeline input, `EndProcessing` is the correct place for the core logic.

bergmeister:
Ok. Thanks for explaining it.

PaulHigin (approved):
LGTM

iSazonov:
Current algorithm is bad and have security hole on Unix.

PaulHigin:
@iSazonov 
Please explain the security hole.

iSazonov:
@PaulHigin Sorry for being short - I was busy.

Disposition. We use CoreFX method to get a temp file. CoreFX uses system calls to get one (links above) - both Windows and Unix have the functions implemented many years ago and they is well-tested and used everyone without claims. Now we want make the thing _better than_ Windows kernel team and glibc team. We haven't Windows sources (we have a comment from MSDN, see above) but we can review[ glibc implementation](https://github.com/gcc-mirror/gcc/blob/master/libiberty/mkstemps.c) and read [mkstemps](http://man7.org/linux/man-pages/man3/mkstemp.3.html) docs.
With the PR's algorithm we get two classic problem for both platforms - for Unix they is critical.
1. Race condition. We create a file name then create a file with ten attempts - glibc do >16000 attempts. It is critical for Unix because _all_ processes creates temp files in shared folder `/temp`.
2. Security. We create file with [0666 permissions](https://github.com/dotnet/coreclr/blob/ /src/mscorlib/shared/System/IO/FileStream.Unix.cs#L50) by default. On Unix all temp files is created in shared folder `/temp` and security is critical.  In glibc implementation on Unix all temp files is created in shared folder `/temp` with 0600 permissions to block access to the file from other accounts. 
3. It also explains why our attempt to use Guids is bad too.

So my conclusion - we can not get better solution than CoreFX. We only can reimplement glibc `mstemps()` and customize it to support any suffix. I believe that our efforts are not worth the result - I don't see the urgent need to create temp files with a different extension (business case?). We should close the PR and Issue with Won't fix.

lzybkr:
Thanks @iSazonov for the analysis - I had your intuition in the back of my mind, but no specifics.

I do believe there is a real customer need and that the workarounds are probably worse than what is currently implemented in this PR.

If we addressed the permissions and the race condition, then I think it's an improvement over the status quo even if it isn't perfect.

mklement0:
@iSazonov: 

* As for the business case: Remember that GNU `mktemp` already allows you to specify a custom extension (suffix), via its far more flexible `XXX...` templating feature;  e.g, something like  
 `mktemp -t XXX.csv` yields something like `/tmp/kLJ.csv`. (BSD `mktemp` works differently and is less flexible - it only allows you to specify a custom _prefix_.)

* All great technical points.

A simple way of rewriting this PR in order to keep `Path.GetTempFileName()` in charge is to:

* _Always_ start out with `filePath = Path.GetTempFileName()`  - this will ensure the proper permissions.

* Only if `-Extension` was specified:
  * Use `filePathWithCustomExt = Path.ChangeExtension(filePath, Extension)` and then  _rename_ that file (`File.Move(filePath, filePathWithCustomExt)`) _in a retry loop_.

A retry loop is then only necessary in the `-Extension` case, which means _we_ have to pick a retry count; picking a high count should be fine, as collisions are unlikely.

iSazonov:
@lzybkr @mklement0 Thanks for the discussion!
Without MSFT CRM 😄  I can not disprove "I do believe there is a real customer need" - so we want `-Extention`.
About implementation. We could implement simple workaround as @mklement0 suggested to only "close the hole". I prefer "perfect" solution 😄 
I have no doubt that the right place is CoreFX. This requires an exhancement of the .Net Standard and subsequent implementation - it is a long process - we should open a Issue in CoreFX.
Today we could implement this in PowerShell Core but we have an Issue to create temp directories either in the cmdlet or in new one. And we see a symmetrical situation - Unix has a richer API than Windows. We again should ask CoreFX to enhance API to create temp directories.
Hence we should create the proposed API and implement it in our Core Facade block to easily migrate to CoreFX implementations later - nothing new - we used such approach in .Net Core 1.0 time and continue currently with 2.0.
Thus we suggest to get Unix mktemp-like functions and re-implement them for both platforms.

mklement0:
@iSazonov: 👍 

> I cannot disprove "I do believe there is a real customer need" 

Well, I'm a customer, and I have that need. 
Q.E.D. 
😜

PaulHigin:
@iSazonov 
Sorry but I am still unclear on the issues.  The latest algorithm in this PR seems Ok to me.  It makes ten attempts to a) generate a random file name then b) create the file.  Each attempt either succeeds or fails.

But it sounds like the dotNet implementation of file creation is susceptible to race and file access permission errors where a) two processes end up with access to the same temp file and b) file access permission is incorrect and possibly detrimental to one of the processes owning the file causing a security hole.

Is this correct?

So the conclusion is that currently there is no way to safely create a new file until these dotNet layer bugs are fixed.  You are suggesting that we create our own native layer file creation for both Windows and Linux platforms that do not have these weaknesses, and then try and get these changes into dotNet.  Correct?

Thanks!

iSazonov:
@PaulHigin 
>The latest algorithm in this PR seems Ok to me

Formally it is Ok. Really it has race and security hole. 

>until these dotNet layer bugs are fixed.

It is not CoreFX bugs. This is due to the misuse of these functions. With glibc we can get the same - that's why `mkstemps()` was added.

My suggestion for CoreFX is add something like:
```c#
public static string GetTempFileName(string Suffix)
public static string GetTempFileName(string Prefix)
public static string GetTempFileName(string InTempDirectory)
public static string GetTempFileName(string Prefix, string Suffix)
public static string GetTempFileName(string Prefix, string Suffix, int VariableLength)
public static string GetTempFileName(string Prefix, string Suffix, int VariableLength, string InTempDirectory)

// And API to create temp directory
```

Yes, my suggestion is implement the APIs temporary in PowerShell.

bergmeister:
I think it would be best  to modify the existing algorithm using the proposal of @mklement0 , which sounds like an overal improvement and closes the  'hole' (assuming that the rename method does not change  the file permissions). I am not sure if this 'hole' is a bit superficial because on Windows any other process can edit the generated temporary file and it would only affect security if a developer directly injected content of the temporary file into other methods in their program or script(please correct me if i'm wrong). Any decent developer should know that any input has the possibility of being a vulnerability for injection methods and therefore needs to assess the risk and potentially defend against it. We should probably have a chat with corefx (maybe we don't have enough understanding of the security model that underlines corefx) but trying to do all the low level stuff ourselves makes it look like to me that we are either overegging the pudding or that corefx is unsafe.

iSazonov:
@bergmeister Sorry for the delay.
Based on @lzybkr and @mklement0 conlusion and we can continue to follow the @mklement0  proposal:
>A simple way of rewriting this PR in order to keep `Path.GetTempFileName()` in charge is to:
* _Always_ start out with `filePath = Path.GetTempFileName()`  - this will ensure the proper permissions.

* Only if `-Extension` was specified:
  * Use `filePathWithCustomExt = Path.ChangeExtension(filePath, Extension)` and then  _rename_ that file (`File.Move(filePath, filePathWithCustomExt)`) _in a retry loop_.

bergmeister:
@iSazonov @lzybkr I have committed the change in the algorithm as agreed in the last comment, the only subtle difference was that the retry logic always applies because it was 'free' and allowed to have less and more elegant code this way.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -2,44 +2,103 @@
...
+                        temporaryFilePath = Path.GetTempFileName(); // this already creates the temporary file
+                        if (Extension != defaultExtension)
+                        {
+                            // rename file
+                            var temporaryFileWithCustomExtension = Path.ChangeExtension(temporaryFilePath, Extension);
+                            File.Move(temporaryFilePath, temporaryFileWithCustomExtension);
+                            temporaryFilePath = temporaryFileWithCustomExtension;
+                        }
+                        creationOfFileSuccessful = true;
+                        WriteVerbose($"Created temporary file {temporaryFilePath}.");
```
SteveL-MSFT: This should probably be a WriteDebug as that tempfile may be deleted

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -2,44 +2,103 @@
...
+                bool creationOfFileSuccessful = false;
+
+                // In case the random temporary file already exists, retry 
+                while (attempts++ < 10 && !creationOfFileSuccessful)
                 {
-                    filePath = Path.GetTempFileName();
+                    try
+                    {
+                        temporaryFilePath = Path.GetTempFileName(); // this already creates the temporary file
+                        if (Extension != defaultExtension)
```
iSazonov: That is why we must follow the @mklement0  proposal and implement two code routes. Here we do unneeded checks in loop. This loop only makes sense for a very loaded scenarios. If we want to guarantee anything in this cycle, the number of attempts should be greater than 10000 - see Unix implementations.
Also we should exclude `creationOfFileSuccessful` - if we get file name we should return immediately without extra checks.
```c#
if ( !Extension.IsPresent() )
{
    return  Path.GetTempFileName();
}
else
{
   return GetFileNameWithExtension(Extention);
}
```

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -2,44 +2,103 @@
...
+                            var temporaryFileWithCustomExtension = Path.ChangeExtension(temporaryFilePath, Extension);
+                            File.Move(temporaryFilePath, temporaryFileWithCustomExtension);
+                            temporaryFilePath = temporaryFileWithCustomExtension;
+                        }
+                        creationOfFileSuccessful = true;
+                        WriteDebug($"Created temporary file {temporaryFilePath}.");
+                    }
+                    catch (IOException) // file already exists -> retry
+                    {
+                        attempts++;
```
iSazonov: Please remove - we do this in cycle header.

src/Microsoft.PowerShell.Commands.Utility/resources/UtilityCommonStrings.resx
iSazonov: If we change the text we should rename `CouldNotFindTemporaryFilename` to `CouldNotCreateTemporaryFilename`.

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
```diff
@@ -6,25 +6,57 @@
...
     Expected Result:
-        A FileInfo object for the temporary file is returned.
+        A FileInfo object for the temporary file is returned and the temporary file gets created correctly.
 #>
 
 Describe "NewTemporaryFile" -Tags "CI" {
 
-    It "creates a new temporary file" {
-        $tempFile = New-TemporaryFile
+    $defaultExtension = '.tmp'
```
iSazonov: Please move the line to BeforeAll.

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
```diff
@@ -6,25 +6,57 @@
...
+            $tempFile.FullName.EndsWith($extension) | Should be $true
+    }
 
-        if(Test-Path $tempFile)
+    It "New-TemporaryItem with an an invalid character in the -Extension parameter should throw NewTemporaryInvalidArgument error" {
+        $invalidFileNameChars = [System.IO.Path]::GetInvalidFileNameChars()
+        foreach($invalidFileNameChar in $invalidFileNameChars)
         {
-            Remove-Item $tempFile -ErrorAction SilentlyContinue -Force
+            { New-TemporaryFile -Extension $invalidFileNameChar -ErrorAction Stop } | ShouldBeErrorId "NewTemporaryInvalidArgument,Microsoft.PowerShell.Commands.NewTemporaryFileCommand"
```
iSazonov: Doesn't make sense to test for all invalid chars - enough of one - we should check only the error id.

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
```diff
@@ -6,25 +6,57 @@
...
+    It "creates a new temporary file with a specific extension" -TestCases $extensionParameterTestCases -Test {
+        Param ([string]$extension)
+        
+        $script:tempFile = New-TemporaryFile -Extension $extension
+            $tempFile | Should Exist
+            # Because the internal algorithm does renaming it is worthwhile checking that the original file does not get left behin
+            [System.IO.Path]::ChangeExtension($tempFile, $defaultExtension) | Should Not Exist
+            $tempFile | Should BeOfType System.IO.FileInfo
+            $tempFile.Extension | Should be ".csv"
+            $tempFile.FullName.EndsWith($extension) | Should be $true
```
iSazonov: Please remove the line - previous line already make the check.

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
```diff
@@ -6,25 +6,57 @@
...
+        $tempFile.Extension | Should be $defaultExtension
+    }
+
+    $extensionParameterTestCases = @(
+        @{ extension = 'csv' }
+        @{ extension = '.csv' }
+        @{ extension = '..csv' }
+        @{ extension = 'txt.csv' }
+        @{ extension = '.txt.csv' }
+    )
```
iSazonov: Please move to `It` block.
```powershell
It "creates a new temporary file with a specific extension" -TestCases $extensionParameterTestCases -Test @(  
        @{ extension = 'csv' }  
        @{ extension = '.csv' }  
        @{ extension = '..csv' }  
        @{ extension = 'txt.csv' }  
        @{ extension = '.txt.csv' }  
     ) { 
```

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
```diff
@@ -6,25 +6,57 @@
...
 Describe "NewTemporaryFile" -Tags "CI" {
 
-    It "creates a new temporary file" {
-        $tempFile = New-TemporaryFile
+    $defaultExtension = '.tmp'
+
+    AfterEach {
+        if ($null -ne $script:tempFile)
+        {
+            Remove-Item $script:tempFile -ErrorAction SilentlyContinue -Force # variable needs script scope because it gets defined in It block
```
iSazonov: Please move the comment on separate line.

iSazonov (commented):
Leave a comment

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
bergmeister: The last test is for test cases like `.txt.csv` because the `Extension` property returns only  `.csv` in this case

iSazonov: Thanks for clarify. 
We use CoreFX method `Path.ChangeExtension()` to change the extention so we shouldn't have many tests which test CoreFX - one test is enough for our code..

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -2,44 +2,103 @@
...
+
+                // In case the random temporary file already exists, retry 
+                while (attempts++ < 10 && !creationOfFileSuccessful)
                 {
-                    filePath = Path.GetTempFileName();
+                    try
+                    {
+                        temporaryFilePath = Path.GetTempFileName(); // this already creates the temporary file
+                        if (Extension != defaultExtension)
+                        {
```
lzybkr: We avoid string comparisons using an operator, instead preferring `string.Equals` and explicitly specify the comparison with either Ordinal or OrdinalIgnoreCase. In this case, Ordinal is probably fine, but we would want a test to make sure File.Move works fine when only the case of the extension differs (e.g. `-Extension .TMP`.)

One other thought - if we compare the extension of `temporaryFilePath` with `Extension`, we can avoid a file move if `GetTempFileName` changed it's algorithm and the new algorithm matches what the user specifies for `Extension` - an unlikely scenario I suppose, but I don't see a downside to this code change.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -2,44 +2,103 @@
...
+                        creationOfFileSuccessful = true;
+                        WriteDebug($"Created temporary file {temporaryFilePath}.");
+                    }
+                    catch (IOException) // file already exists -> retry
+                    {
+                        attempts++;
+                        if (temporaryFilePath != null)
+                        {
+                            File.Delete(temporaryFilePath);
+                        }
```
lzybkr: What if this throws? We won't retry.

bergmeister: Should we then simply try File.Delete with an empty catch block instead?

lzybkr: If the retry count is 10, that seems fine. If it was a huge number, then I'd say probably not.  A short sleep in the catch might not hurt either, like 10ms or something.

SteveL-MSFT: Would it make sense to write a warning if we're not able to delete it?

bergmeister: At least a warning seems reasonable. But I am slightly more leading to actually throwing a (terminating) error because if deleting the temporary file that just got created failed then I suspect that in most cases there is rather a fundamental problem with permissions. WDYT?

bergmeister:
I addressed the main points made in the review:
 - If the extension is '.tmp', then it goes through the same codepath as before this PR and only otherwise use the new algorithm
 - cater for failure if file deletion fails in catch block
 - improve tests and add Unix specific test for '.TMP' where we expect the file ending in '.tmp' to not exist. Unfortunately this test fails on the Mac build and I am not sure why. Could this be a bug in Pester/PowerShell? Any ideas?

bergmeister:
I addressed the main points made in the review:
 - If the extension is '.tmp', then it goes through the same codepath as before this PR and only otherwise use the new algorithm
 - cater for failure if file deletion fails in catch block
 - Usage of `Ordinal` string comparison
 - improve tests and add Unix specific test for '.TMP' where we expect the file ending in '.tmp' to not exist. Unfortunately this test fails on the Mac build and I am not sure why. Could this be a bug in Pester/PowerShell? Any ideas?

bergmeister:
I have updated the test to pass since one test did not make sense on Mac due to the the fact that Mac sees files with different casing as the same file similar to Windows.

iSazonov:
Currently we have new API request in CoreFX and CoreFX team has post 2.1 plans to enhance the API. Feel free to vote for the API.
I believe we can close the PR as non-critical and wait CoreFX implementation.

Issue #4609, opened 2017-08-17
Create Native Test Solution for Certificate Authentication


markekraus:
## Problem

There is currently no native cross-platform, CI independent mechanism to test Certificate Authentication in `Invoke-WebRequest` and `Invoke-RestMethod`.

## Proposal

Create a docker container to house a minimal aspnetcore app to validate Client Certificate presence in calls made by `Invoke-WebRequest` and `Invoke-RestMethod`. The docker container will need a self-signed certificate, SSL/TLS configuration, and be Linux based (debian jessie?). The aspnetcore app will need to, at minimum, echo whether a Client Certificate was presented or not and possibly include details about the Client Certificate. The entire solution will need to work in both AppVeyor and Travis CI and be as CI agnostic as possible.

see #4546

markekraus:
@TravisEz13 We can work on the solution through this issue.

markekraus:
@TravisEz13 I think I'm on to something. the built-in Kestrel server for ASP.NET core supports client certificates. I'm working on an small app that returns the client certificate details. The idea is that the test calls `dotnet run` on the project in a `Start-Job`, does the tests, then tears it down the job. I actually have a prototype of this working in a web browser, but `Invoke-WebRequest` and `Invoke-RestMethod` can't find a common algorithm with Kestrel. It looks like I need to enable TLS in Kestrel, but the examples I have found for doing so no longer work in 2.0.

iSazonov:
I believe we should use natively supported web server - Apache on CI Travis and IIS on CI AppVeyor. It requires more effort, but it makes our tests more predictable. I do not see the pressing need to make sure to use a port server.

markekraus:
As an update on this, it was discovered in #4620 that Docker is not supported on macOS in Travis CI. The current approach in #4622 is to have the listener app published by `Publish-PSTestTools` like `TextExe` and then run the listener on demand using the baked in Kestrel server. This has the advantage of being much quicker.

Also, the possibility of moving all tests form HttpListener to this app is open for discussion.

TravisEz13:
@markekraus It seems worth investigating.  I would start with feature tests and if there are any issues with the approach the impact would not be blocking.  Then if it proves a good solution, we can try broadening the approach to other tests.

markekraus:
@TravisEz13 How should I approach PR's for this? There are a bunch of feature `it`'s for both cmdlets, how should I break them up, if at all?

TravisEz13:
[...]

PR #4620, opened 2017-08-19
[Feature] Add Tests for Web Cmdlet Certificate Authentication

markekraus:
Addresss #4609 and corrects `Pending` tests from #4546

This test uses an ASP.NET Core 2.0 app in a docker container to run a simple service to echo back the client certificate details if one was provided. If a client cert is not provided, the details are empty and the status reports `FAILED`. Since this test will be slow in the CI's I have use a separate `Describe` block and included the `Feature` and `Slow` tags.


test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1854,3 +1717,64 @@ Describe "Web cmdlets tests using the cmdlet's aliases" -Tags "CI" {
...
+        $result = Invoke-RestMethod -Uri $uri -SkipCertificateCheck
+        $result | Should Match ([regex]::Escape($failedMessage))
+    }
+
+    It "Verifies Invoke-RestMethod Certificate Authentication Successful with -Certificate" {
+        $certificate = Get-PfxCertificate -FilePath $pfxPath
+        $result = Invoke-RestMethod -uri $uri -Certificate $Certificate -SkipCertificateCheck
+        $result | Should Match ([regex]::Escape($successMessage))
+    }
+}
```
iSazonov: Please add new line at EOF.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1854,3 +1717,64 @@ Describe "Web cmdlets tests using the cmdlet's aliases" -Tags "CI" {
...
+        $containerPath       = Join-Path $PSScriptRoot 'assets\docker\ClientCertificateCheck'
+        $pfxPath             = Join-Path  $containerPath 'ClientCert.pfx'
+        $initCompleteMessage = 'Now listening on'
+        $successMessage      = 'Status: OK'
+        $failedMessage       = 'Status: Failed'
+        
+        Push-Location $containerPath
+        $null = docker build -t $containerName .
+        $timeOut = (get-date).AddSeconds($initTimeoutSeconds)
+        $null = docker run -d -p ${portNumber}:${portNumber} --name $containerName $containerName $portNumber
```
iSazonov: Can we compile the container at CI startup like our `TestExe` and run? If it is possible we could remove "Slow" and "Feature".
Also maybe move the project to Tools folder.

test/powershell/Modules/Microsoft.PowerShell.Utility/assets/docker/ClientCertificateCheck/Pages/_Layout.cshtml
```diff
@@ -0,0 +1 @@
+@RenderBody()
```
iSazonov: Please check all files and add new line at EOF - some files have it and some files haven't it - we should use one pattern for all files.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: If we compile the container at CI start, that couples the test with an entire build scenario instead of a test scenario.

test/powershell/Modules/Microsoft.PowerShell.Utility/assets/docker/ClientCertificateCheck/Pages/_Layout.cshtml
markekraus: @iSazonov  `_Layout.cshtml` and `index.cshtml` intentional as those determine the response stream and superfluous new lines cause it to response less like a raw replay and more like an HTML page. I will correct the others.

markekraus:
I think  different approach will be needed. After looking at the Travis CI failures and reading the documentation, [Docker is not supported on OS X](https://docs.travis-ci.com/user/docker/).

I'm going to close this pull request and think of a new plan of attack.

iSazonov:
Thank you so much for your efforts!

Perhaps we should still use servers that are supported by CIs internally.

markekraus:
@iSazonov I'm considering using the aspnetcore project without docker as the baked in Kestrel server is available cross-platform. Making the tests CI specific couples them with the CI and then breaks when the CI environment changes. The only problem with the aspnetcore approach is that the test will now be coupled with the dotnet SDK, probably not a huge deal as moving to the next major version would entail other code and testing reworks. I welcome your input on it in #4609.

Issue #4598, opened 2017-08-16
Move certificateCommon.psm1 to a Common Area and Replace Self-Signed Cert Valid for All tests


markekraus:
[certificateCommon.psm1](https://github.com/PowerShell/PowerShell/blob/master/test/powershell/Modules/Microsoft.PowerShell.Security/certificateCommon.psm1) contains a self signed certificate used for tests in the `Microsoft.PowerShell.Security`. Other modules will require certificates so this test module should be moved to a common area accessible to all tests.

Additionally, the certificate  in `New-GoodCertificate` is issued for only the `Key Encipherment`, `Data Encipherment`, and `Document Encryption` key usages. This key cannot be used for tests such as certificate authentication with `Invoke-WebRequest`/`Invoke-RestMethod`. A new self-signed certificate should be issued that includes key usage for all tests that may require it to prevent the need for separate certificates and functions for each key requirement.

For more details see #4546

iSazonov:
@adityapatwardhan Could you please comment the suggestion?

iSazonov:
If we want one (two) reusable certificate(s) we should document their properties (usages) by means of comments in the module.

markekraus:
The comments should probably include the following (at minimum):

* Subject
* Thumbprint
* NotBefore
* NotAfter 
* Key Usages (including Extended Key Usages)

markekraus:
Closed by #7875

PR #4622, opened 2017-08-20
Add test WebListener module and tests for Web Cmdlet Certificate Authentication

### LINKED ISSUE #4609: Create Native Test Solution for Certificate Authentication

markekraus:
Closes #4609 and replaces tests for #4546.

This test uses an ASP.NET Core 2.0 app to run a simple service to echo back the client certificate details if one was  provided. If a client cert is not provided, the details are empty and the status reports FAILED. The app runs in a Job.


iSazonov:
@markekraus See as we prepare TestExe.csproj https://github.com/PowerShell/PowerShell/blob/b3e7fd374f2898f7e84b28229aada18a3fcbe23e/build.psm1#L780

markekraus:
@iSazonov Thanks, I'll see how I can fit this one in into `Publish-PSTestTools`

on another note, the `mdspell` is falsely calling out `ASP.NET` in my `README.md`. Is there place to add that to be ignored? Or is there another way to handle it?

iSazonov:
Yes, it is `.spelling` - add `Asp.Net` (case-sensitive).

iSazonov:
@markekraus Can we replace our `ListenerHttp` script with the application (not in the PR)?

markekraus:
@iSazonov In theory we could move all of the we tests to this, but, it would mean rewriting the existing features of HttpListener into ASP.NET. Or doing something that allowed PowerShell to manage the business logic.

I was already considering switching the `Validate Invoke-WebRequest -SkipCertificateCheck` tests over to this new listener since the ServerCert fails as a self-signed cert.

iSazonov:
@markekraus Please keep the PR as small as possibly - it is better to make cleanups later.

If we can move to the new application from `HttpListener` we should ask the test area experts to  this /cc @adityapatwardhan - if it will be approved we'll review the code with the point in mind.

markekraus:
@iSazonov Should I roll back the baddssl changes?  Aside from those, the rest of the changes on this PR are interdependent.

Can you restart the failed Travis CI test? it failed an unrelated test due to httpbin not being unavailable.

iSazonov:
@markekraus Main rule is to keep PR as small as possible and don't add unneeded changes.
So I believe, yes, we should rollback - only keep the commit for follow PR. Also you can wait conclusion from @adityapatwardhan I asked to lead this change as the test area expert.
Main question now to continue the review is - do we want move from `HttpListener` in future or not.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -296,6 +296,46 @@ function ExecuteWebRequest
     return $result
 }
 
+# Starts the ClientCertificateCheck listener
+function StartClientCertificateCheck 
+{
+    param([int]$Port)
+    $initTimeoutSeconds  = 15
+    $projectRootPath     = Resolve-Path (Join-Path $PSScriptRoot "\..\..\..\..\")
```
iSazonov: `function Publish-PSTestTools` already added the path into PATH environment variable - so we don't need build the full path.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -296,6 +296,46 @@ function ExecuteWebRequest
     return $result
 }
 
+# Starts the ClientCertificateCheck listener
+function StartClientCertificateCheck
```
iSazonov: Maybe `Start-ClientCertificateCheck`

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -296,6 +296,46 @@ function ExecuteWebRequest
...
+    $projectRootPath     = Resolve-Path (Join-Path $PSScriptRoot "\..\..\..\..\")
+    $appSrcPath          = Join-Path $projectRootPath  'test\tools\ClientCertificateCheck'
+    $appPublishPath      = Join-Path $appSrcPath 'bin'
+    $appDllPath          = Join-Path $appPublishPath 'ClientCertificateCheck.dll'
+    $serverPfxPath       = Join-Path $appSrcPath 'ServerCert.pfx'
+    $serverPfxPassword   = 'password'
+    $initCompleteMessage = 'Now listening on'
+    
+    $timeOut = (get-date).AddSeconds($initTimeoutSeconds)
+    $Job = Start-Job {
```
iSazonov: If we'll use the application in multiple test files we should save time - run it once and check if it already is running.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -296,6 +296,46 @@ function ExecuteWebRequest
...
+    while (-not $isRunning -and (get-date) -lt $timeOut)
+
+    if (-not $isRunning) {
+        throw 'Container did not start before the timeout was reached.'
+    }
+    return $job
+}
+
+# Stops the ClientCertificateCheck listener
+function StopClientCertificateCheck
```
iSazonov: Maybe `Stop-ClientCertificateCheck`.
If we'll use the application in multiple test files we should it once and shouldn't stop it.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1183,6 +1200,26 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
     #endregion charset encoding tests
 
+    #region Client Certificate Authentication
```
iSazonov: It is better use `Context {}`. I suggest don't mix HTTP and HTTPS tests and put HTTPS test in separate `Context {}` while we use different tools for them.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1183,6 +1200,26 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
...
+
+    It "Verifies Invoke-WebRequest Certificate Authentication Fails without -Certificate"  {
```
iSazonov: Maybe `Invoke-WebRequest certificate authentication without -Certificate`.
What IIS or Apache returns in the case? I believe the test shouldn't depend from underlying web server.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1183,6 +1200,26 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
...
+        $uri = 'https://localhost:8443/'
+        $result = Invoke-WebRequest -Uri $uri -SkipCertificateCheck
+        $failedMessage = 'Status: Failed'
+        $result.RawContent  | Should Match ([regex]::Escape($failedMessage ))
+    }
+
+    It "Verifies Invoke-WebRequest Certificate Authentication Successful with -Certificate" {
```
iSazonov: Maybe `Invoke-WebRequest certificate authentication with -Certificate`.
What IIS or Apache returns in the case? I believe the test shouldn't depend from underlying web server.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1701,28 +1740,25 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
 
     #endregion SkipHeaderVerification tests
 
-    #region Certificate Authentication Tests
+    #region Client Certificate Authentication
```
iSazonov: Please use `Context`

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1701,28 +1740,25 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
...
 
-    # Test pending creation of native test solution 
-    # https://github.com/PowerShell/PowerShell/issues/4609
-    It "Verifies Invoke-RestMethod Certificate Authentication Fails without -Certificate" -Pending {
-        $command = 'Invoke-RestMethod https://prod.idrix.eu/secure/'
-        $result = ExecuteWebCommand -command $command
-
-        $result.Output | Should Match ([regex]::Escape('Error: No SSL client certificate presented'))
+    It "Verifies Invoke-RestMethod Certificate Authentication Fails without -Certificate" {
```
iSazonov: See my comment above for `Invoke-WebRequest'.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1701,28 +1740,25 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
...
 
-    # Test pending creation of native test solution 
-    # https://github.com/PowerShell/PowerShell/issues/4609
-    It "Verifies Invoke-RestMethod Certificate Authentication Successful with -Certificate" -Pending {
-        $Certificate = GetSelfSignedCert
-        $command = 'Invoke-RestMethod https://prod.idrix.eu/secure/ -Certificate $Certificate'
-        $result = ExecuteWebCommand -command $command
-
-        $result.Output | Should Match ([regex]::Escape('SSL Authentication OK!'))
+    It "Verifies Invoke-RestMethod Certificate Authentication Successful with -Certificate" {
```
iSazonov: See my comment above for `Invoke-WebRequest'.

test/tools/ClientCertificateCheck/Pages/Index.cshtml
```diff
@@ -0,0 +1,10 @@
+﻿@page
+@model IndexModel
+Status: @Model.Status
+Thumbprint: @Model.CertThumbprint
+Subject: @Model.CertSubject
+Subject Name: @Model.CertSubjectName
+Issuer: @Model.CertIssuer
+Issuer Name: @Model.CertIssuerName
+Not After: @Model.CertNotAfter
+Not Before: @Model.CertNotBefore
```
iSazonov: GitHub mark absent new line at EOF as error and suggest GitHub community to fix it. We already fixed most of our files in the repo. So please add new line at EOF in all files you change until it is really need empty file.

test/tools/ClientCertificateCheck/Pages/_Layout.cshtml
```diff
@@ -0,0 +1 @@
+@RenderBody()
```
iSazonov: New line.

test/tools/ClientCertificateCheck/Pages/_ValidationScriptsPartial.cshtml
```diff
@@ -0,0 +1 @@
+﻿
```
iSazonov: New line?

test/tools/ClientCertificateCheck/README.md
```diff
@@ -0,0 +1,54 @@
...
+dotnet dotnet publish --output app --configuration Release
+cd app
+dotnet ClientCertificateCheck.dll ServerCert.pfx password 8443
+```
+
+The test site can then be accessed via `https://localhost:8443/`.  
+
+The `ClientCertificateCheck.dll` takes 3 arguments: the path to the Server Certificate, the Server Certificate Password, and the TCP Port to bind on.
+
+# Run As a Docker Container
```
iSazonov: Is it actual?

test/tools/ClientCertificateCheck/dockerfile
```diff
@@ -0,0 +1,18 @@
+# Stage 1: Build the Application
```
iSazonov: Do we still need the file?

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @iSazonov I tried to match the other private functions in the test already. I originally wanted to use `Start-ClientCertificateCheck`, but though I would get a note for not matching the others. I can't win  :D

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @iSazonov That is used to pass to the Job scope where the path changes made by the `Publish-PSTestTools` are not available.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @iSazonov  I'm open to suggestions on how to implement that. As it stands now, it starts in less than a second. I was just following the current paradigm for how `Start-HttpListener` is used in the tests. Currently, `StartClientCertificateCheck` is called once at the beginning of the `Invoke-WebReqeust` and once at the begining of the `Invoke-RestMethod` tests. it is stopped at the end of the tests for those describe blocks.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @iSazonov same as with `Start-ClientCertificateCheck`

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @iSazonov  I think we should maybe do that as a separate PR to clean up all of the current HTTPS tests too. Or, I could build the contexts now and move the other HTTPS into them as we update the tests (i.e. tests referencing badssl.com).

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: My thoughts was about HttpListener style. Maybe somebody else comment this too.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: not available? I wonder. Do you check? Usually env variables is inherited in sub process.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @iSazonov  The problem with this test is what the intended result is. IMO, the intended result is that the end server is able to see details about the certificate that is passed. That is why the Kestrel server is set to `ClientCertificateMode.AllowCertificate` instead of `ClientCertificateMode.RequireCertificate`. In the real world, this can be implemented in any number of ways. I wanted the method of testing success or fail of this to be as close as possible to each other with the only different being the presence or lack there of a client certificate.

test/tools/ClientCertificateCheck/README.md
markekraus: I'm not sure what you asking here.

test/tools/ClientCertificateCheck/dockerfile
markekraus: @iSazonov  If we don't want to try this as a docker container, then no. I can remove the docker parts.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: CI VMs is very-very slow. If locally you see no start delay then on CI it is possible the delay will be large. That's why I think it's best to use one tool (remove HttpListener) and run it only once.

>how to implement that

Call Invoke-WebRequest and analyze the response.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: You have to edit all HTTPS tests so I believe it is ok to add Context in the PR.
But if you want I agree to follow the file's pattern now and make cleanups later.

test/tools/ClientCertificateCheck/Pages/_ValidationScriptsPartial.cshtml
markekraus: @iSazonov it lookslike I don't need this file. i will remove it.

test/tools/ClientCertificateCheck/Pages/Index.cshtml
markekraus: @iSazonov will fix

test/tools/ClientCertificateCheck/Pages/_Layout.cshtml
markekraus: @iSazonov will fix

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: [see my comment here](https://github.com/PowerShell/PowerShell/pull/4622#discussion_r134122470)

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: Thanks for clarify! I agree with test.

Closed.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @iSazonov  [see my other comment](https://github.com/PowerShell/PowerShell/pull/4622#discussion_r134122290)

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: Closed.

test/tools/ClientCertificateCheck/README.md
iSazonov: Sorry. If it is not supported on Mac it make sense to remove.

test/tools/ClientCertificateCheck/dockerfile
iSazonov: If it is not supported on Mac it make sense to remove.

markekraus: @iSazonov I will remove it and .dockerignore as well as the docker info in the readme.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: I'll do a separate followup PR to this one under the guise of HTTPS test adjustments. moving all the tests (including those not touched in this PR) that depend on HTTPS to a single context in each cmdlet section and switching the URL to the new app... assuming this PR get's approved and that is the way we want to go with this.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @iSazonov As for the less than 1 second start time, I was referring to the start time within the CI VMs. The majority of the overhead is in the build stage. I just have the `wait` in there to prevent a race condition between the job delay and the first test that requires the resource.  

I think the current method for ensuring the process is running is fine. It makes sense to use `Invoke-WebRequest` for `Start-HttpListener` as there is no other means for ensuring the server is live. But with this I have access to the console output from the server itself saying it is live.

What I was asking about was how to implement it to be running for multiple tests. What mechanism would start it? `Start-PSPester`?

test/tools/ClientCertificateCheck/README.md
markekraus: @iSazonov removed

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @iSazonov it turns out you were right. I was able to adjust this and pull the parent path from the command.

markekraus:
Hmm... The macOS Build is "passing" but in actuality it is hitting an unhandled exception and exiting without property saying it has failed.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: We want run tests without ` Start-PSPester` as `.\file.test.ps1`.
I think we should move the code (1-function to start the app and 2-function check that it is running) in `HelpersCommon` - in first file we detect that the app is not running and start it, in second file we detect that the app is already running and use it.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -299,32 +299,32 @@ function ExecuteWebRequest
...
         $containerStatus = $Job.ChildJobs[0].Output | Out-String
         $isRunning = $containerStatus -match $initCompleteMessage
     }
     while (-not $isRunning -and (get-date) -lt $timeOut)
 
     if (-not $isRunning) {
-        throw 'Container did not start before the timeout was reached.'
+        $Job | Stop-Job -PassThru | Receive-Job
+        $Job | Remove-Job
+        throw 'Job did not start before the timeout was reached.'
```
iSazonov: 1. We should make the message more clear - maybe "ClientCertificateCheck is not started."
2. Since we're still throw maybe move the throw into do-while and make the cycle unlimited:
```powershell
do
    Start-Sleep -Milliseconds 100
    $containerStatus = $Job.ChildJobs[0].Output | Out-String

    if ($containerStatus -match $initCompleteMessage)
    {
        # Application started.
        break;
    }

    if (get-date) -gt $timeOut) 
    {
         throw "ClientCertificateCheck is not started."
    }
while ($true)
```

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -299,32 +299,32 @@ function ExecuteWebRequest
...
+        $path = Split-Path -parent (get-command ClientCertificateCheck).Path
+        Push-Location $path
+        dotnet $using:appDll $using:serverPfx $using:serverPfxPassword $using:Port
     }
     # Wait until the app is running or until the initTimeoutSeconds have been reached
     do
     {
-        Start-Sleep -Seconds 1
+        Start-Sleep -Milliseconds 100
         $containerStatus = $Job.ChildJobs[0].Output | Out-String
```
iSazonov: We don't use container - please rename the variable.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: If we move the function in ` HelpersCommon.psm1` it is ok rename it to ` Start-ClientCertificateCheck`.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1701,28 +1740,25 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
...
+        $uri = 'https://localhost:8443/'
+        $certificate = GetSelfSignedCert
+        $result = Invoke-RestMethod -uri $uri -Certificate $certificate -SkipCertificateCheck
+        $successMessage = 'Status: OK'
+        $result | Should Match ([regex]::Escape($successMessage))
+        $result | Should Match $certificate.Thumbprint
     }
 
-    #endregion Certificate Authentication Tests
+    #edregion Client Certificate Authentication
```
GavinEke: @markekraus slight typo on endregion

markekraus: @GavinEke Thanks. I will correct that.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @iSazonov I will update the throw message. 

On the other suggestion, I'm not sure it makes sense to introduce flow logic inside a `do` block when the `while` is intended to manage the `do`'s exit. I have never been a fan of `while ($true)` Is there something wrong with the correct logic?

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @iSazonov I will change it to `$initStatus`

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @iSazonov  Ah. I see. Since it may be decided to move all tests to this, would it make sense to create a module of its own, similar to `HttpLitsener`? If we want to have logic detecting if it is running or not and with possibility of running multiple of these on different ports, it might warrant a `class` definition to contain the port and job for a specific listener and a module variable to track started listeners. It might also warrant a rename of the app to something more generic than `ClientCertificateCheck`

For this PR, to keep it small, Moving the start and stop functions to HelpersCommon and renaming them proper `Verb-Noun` would be a start.

markekraus:
It looks like the [AppVeyor build beta.5-4786](https://ci.appveyor.com/project/PowerShell/powershell/build/6.0.0-beta.5-4786#L97) failed due to temporary issue with `powershell.myget.org`. if someone could restart the build, that would be helpful.

iSazonov:
@markekraus It seems I haven't permissions. So you can "rename" last commit and push - it will be restart CI jobs.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: No, only my sample looks more clear for me 😄 because we replace two code blocks with single logical block.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: We could convert the app to binary module (and rename it) and get benefits from autoloading. It may require more effort, but it seems worth.

markekraus: I'm not sure how that would work. This is an executable binary. I've never seen that kind of thing included in a binary module. I'm not saying it's impossible, just that it is beyond me how to do that.

SteveL-MSFT:
@iSazonov I'm going to assign this to you as you've been involved while @adityapatwardhan is taking time off

iSazonov:
@SteveL-MSFT We need get approve from test area experts because of significant changes in the PR - please ping anybody if @adityapatwardhan is busy.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
TravisEz13: Scripts can still autoload.  I think the optimization of converting it to an assembly module can be filed as an issue and let @markekraus  prove the concept.  Unless someone has time to work with him now on writing an assembly based module.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
TravisEz13: Maybe add comments to the intent of the test if it's not clear.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
TravisEz13: Again, perhaps add comments to the expected behavior of the server.

TravisEz13:
Thanks, @markekraus  and @iSazonov for all your work on this so far.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @TravisEz13 Should I add comments in the tests? I have the behavior of the server documented in README.md for the `ClientCertificateCheck` app.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @TravisEz13 and @iSazonov Ok, since it appears we may head in the direction of using this for more tests than just the certificate authentication, I have renamed the app `HttpsListener` and added an `HttpsListener` module to house the `Start-`, `Stop`, and `Get-` commands. There is a module script variable to track the running listeners so if it is already listening on the requested port the running listener is returned. To keep it always running, I have removed the stop commands form the `AfterAll` blocks. I think that incorporates everything discussed here except making it a binary module.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: If the test fail the comments should help us quickly discover a root cause. README.md is useful for enhancement of `ClientCertificateCheck` app.

markekraus: @iSazonov I'm not opposed to adding the comments, but, shouldn't this be consistent with the other tests which are not documenting external behavior? If I add the comments here, the other tests should require the same. If there are no plans to document every test accessing either HttpListener or httpbin, why should my test be different? When I'm troubleshooting one of the current test failures I have to go to either httpbin or the source for HttpListener to figure out what should be returned. All of these tests are checking the response from a server, to me it makes sense for the documentation of that server to rest with the server and not in the test. We shouldn't document the behavior in 2 places, IMO

As a compromise, what about a comment suggesting to consult the README.md for behavior of the listener?

iSazonov: You are right. I believe Travis meant that it is not clear that is `'https://localhost:8083/'` (Google translate? :smile:) so we need short comment that we expect here and where investigate further. I think we could do one comment for all test but we agreed to add `Context` later.

markekraus: Ah. ok. That will work. Do I need to take a specific action in this PR?

markekraus:
Something is amiss with macOS and the client certificate. It was working with `Get-PfxCertificate` and the `ClientCert.pfx` then it started failing without any change to the cert or the `Get-PfxCertificate`. so I switched to `GetSelfSignedCert` and it was working a few commits ago. I made change to another portion of `WebCmdlets.Tests.ps1` but didn't touch this function and now it is failing to import the certificate with this function. I'll have to find some macOS environment to troubleshoot this in.

markekraus:
After moving past the password protected pfx issue, I discovered that this is still not supported on macOS

`The handler does not support client authentication certificates with this combination of libcurl (7.55.1)`

dotnet/corefx#9728
dotnet/corefx#19718

It was missed as passing on mac because of the issue reported in #4641. The test was actually failing
https://travis-ci.org/PowerShell/PowerShell/jobs/265494069#L8693 but the CI reported passing. Then we ultimately marked the test pending so that we could address this through #4609 

@TravisEz13 @iSazonov , I'm not sure where to go from here. It looks like this is definitely still unsupported in macOS. It does work in windows and Linux. I could make this test skipped on macOS, but the bigger issue of this functionality being enable but not available in macOS, I'm unsure what to do.

markekraus:
@TravisEz13 The test for this are currently set to pending in master. But, without these replacement tests, there is no need to add the `HttpsListener`. Essentially, this entire PR is probably not worth merging if the tests will remain pending unless we want this PR to add the `HttpsListner` in preparation for the future with the new tests pending. Would you like to go forward with adding the `HttpsListener` or should we close this PR and save the HttpsListener for when it is needed?

TravisEz13:
Sorry, I should clarify.  You should detect if the tests are running on macOS and set them to skip.  We have this pattern for windows only features in many tests.

TravisEz13:
Here is an example that skips for everything that is not Windows:  https://github.com/PowerShell/PowerShell/blob/e83fcca277a9dd81d5cb2ff56002b6179c34ec3b/test/powershell/engine/Api/BasicEngine.Tests.ps1#L7

markekraus:
@TravisEz13  Thanks for clarifying. I have adjusted the tests. I still kept the changes to client cert. It at least removes that hideous base64 block from the test file.

iSazonov:
@markekraus Great investigations! I think we should open new tracking Issue for the MacOs issue (not 4609) and cross reference it to 3648 (and change comments in test files).

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -303,86 +303,9 @@ function GetSelfSignedCert {
...
-YJX20neZhIGnsT36bd8e38Mj+7hrVhvV/G2x0aS+lB2lD0HIvRNW02+UxRsZ+S+TtBXnlTHFLAm5
-+IBnXcKWBVnaEvBjwyMIo/bI8C0fhFOt+W88XyoIuPeRYSKVRmg2vjyqMSUwIwYJKoZIhvcNAQkV
-MRYEFC3s8TSP8ht4D0XTFqA5tetMYxL3MDEwITAJBgUrDgMCGgUABBS39FfrA3N6RIvd2k2XO1rY
-hqPP3QQIlSXpfTECuB4CAggA
-'@
-    $Bytes = [System.Convert]::FromBase64String($PfxBase64)
-    [System.Security.Cryptography.X509Certificates.X509Certificate2]::new($Bytes)
+    $parentPath = Split-Path -parent (get-command HttpsListener).Path
+    $pfxPath = Join-Path $parentPath 'ClientCert.pfx'
+    [System.Security.Cryptography.X509Certificates.X509Certificate2]::new($pfxPath,'password')
```
iSazonov: If we plan reuse the certificate in multiple test files it is better to move the function (as Get-TestCertificate?) in new HttpsListener module.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1222,6 +1121,28 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
     #endregion Content Header Inclusion
 
+    #region Client Certificate Authentication
+
+    It "Verifies Invoke-WebRequest Certificate Authentication Fails without -Certificate"  {
+        $uri = 'https://localhost:8083/'
```
iSazonov: I believe we should create a new functions (Get-HttpsTestUrl/Get-HttpTestUrl?) in new HttpsListener module to get the test server Urls .

test/tools/HttpsListener/Pages/Index.cshtml.cs
```diff
@@ -0,0 +1,41 @@
...
+        public string CertNotBefore;
+        public string CertIssuer;
+        public string CertIssuerName;
+        public string CertSubjectName;
+        public string CertThumbprint;
+        public string Status = "FAILED";
+        
+        public void OnGet()
+        {
+            if(null == HttpContext.Connection.ClientCertificate){
```
iSazonov: Please move the brace on new line.

test/tools/HttpsListener/Pages/Index.cshtml.cs
```diff
@@ -0,0 +1,41 @@
...
+    public class IndexModel : PageModel
+    {
+        public string CertSubject;
+        public string CertNotAfter;
+        public string CertNotBefore;
+        public string CertIssuer;
+        public string CertIssuerName;
+        public string CertSubjectName;
+        public string CertThumbprint;
+        public string Status = "FAILED";
```
iSazonov: Does all have to be public?

markekraus:
@iSazonov #4650 created and comments updated.

test/tools/HttpsListener/Program.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace HttpsListener
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            if (args.Count() != 3)
+            {
+                System.Console.WriteLine("Required: <CertificatePath> <CertificatePassword> <PortNumber>");
```
iSazonov: Do we really want the password being mandatory?

test/tools/HttpsListener/Pages/Index.cshtml.cs
```diff
@@ -0,0 +1,41 @@
...
+            CertSubject = HttpContext.Connection.ClientCertificate.Subject;
+            CertNotAfter = HttpContext.Connection.ClientCertificate.NotAfter.ToString();
+            CertNotBefore = HttpContext.Connection.ClientCertificate.NotBefore.ToString();
+            CertIssuer = HttpContext.Connection.ClientCertificate.Issuer;
+            CertIssuerName = HttpContext.Connection.ClientCertificate.IssuerName.Name;
+            CertSubjectName = HttpContext.Connection.ClientCertificate.SubjectName.Name;
+            CertThumbprint = HttpContext.Connection.ClientCertificate.Thumbprint;
+            if (!string.IsNullOrEmpty(CertThumbprint))
+            {
+                Status = "OK";
```
iSazonov: All above assigments we could move in the "if" block.

test/tools/Modules/HttpsListener/HttpsListener.psd1
```diff
@@ -0,0 +1,10 @@
+@{
+    ModuleVersion = '1.0.0'
+    GUID = '90572e25-3f15-49b0-8f25-fb717d3ef46a'
+    Author = 'Mark Kraus'
+    CompanyName = ''
+    Copyright = ''
+    Description = 'Creates a new HTTPS Listener for testing purposes'
+    RootModule = 'HttpsListener.psm1'
```
iSazonov: If we plan move here `HttpListener.psm1` it make sense rename the module - `WebListener`?

test/tools/Modules/HttpsListener/HttpsListener.psm1
```diff
@@ -0,0 +1,126 @@
...
+    (
+        [ValidateRange(1,65535)]
+        [int[]]$Port
+    )
+
+    process 
+    {
+        if ($Port)
+        {
+            return $Script:HttpsListeners[$port]
```
iSazonov: Why we want multiple test listeners? I think one is enough.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: #4598 addresses the larger issue of certificates in tests. Ideally, there would be one certificate for all tests and it would be part of the common code for certificates.

test/tools/HttpsListener/Pages/Index.cshtml.cs
markekraus: My understanding is that it does in order to be referenced in the view. I can test with private though.

test/tools/HttpsListener/Pages/Index.cshtml.cs
markekraus: I'll also add a space between `if` and `(`

test/tools/HttpsListener/Pages/Index.cshtml.cs
markekraus: The `if` block is probably not even needed. I'm unaware of any situation where a `ClientCertificate` would not be null and have no thumbprint. I think my test here is may be over-paranoid. The ` if(null == HttpContext.Connection.ClientCertificate)` is probably a sufficient test on its own.

test/tools/HttpsListener/Program.cs
markekraus: @iSazonov  The password is either supplied at invocation, or is hard coded. In theory, We may need to run multiple listeners with multiple certificates that might not have the password set to `password`. I thought supplying it at listener invocation was better than hard coding it in the listener.

test/tools/Modules/HttpsListener/HttpsListener.psd1
markekraus: Web listener works for me. Its more general and covers both HTTP and HTTPS. I will rename it.

test/tools/Modules/HttpsListener/HttpsListener.psm1
markekraus: @iSazonov  Assuming we may need multiple HTTPS based listeners with different certificates. For example, we may want a listener with an expired certificate to test expired cert scenarios vs self-signed vs invalid chain, etc. For an HTTP (no S) listener, 1 port and 1 listener works, but for HTTPS it is probably not.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @iSazonov  It's a possibility. Should I add that now or include it in future PRs when I start adding features to the new listener and moving tests over? HTTPS might be tricky if we need to run multiple.

test/tools/Modules/HttpsListener/HttpsListener.psm1
iSazonov: We only need to do this if we run tests in parallel. I think we're never going to do this (no Pester support is and CI VMs is very slow). So to cache config and to reinitialize the listener if the config changed is enough.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: If you already create the module why defer? It is better to make main important things now and further fix (migrate) only tests and minor bugs in the module.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: If you already create the module we should make it as completed solution.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: Because the client certificate is not part of the listener. It's included now only because of issue #4598. The `ClientCert.pfx` would be removed and replaced with the certificate used for all certificate based tests (like the ones in M.P.Security module). That would likely be a separate module from this listener. I can certainly add it now, but  `GetSelfSignedCert` function already exists in master. Since it hasn't yet been determined how the common certificate functions will be handled, it would change in the future anyway once #4598 is resolved.

test/tools/Modules/HttpsListener/HttpsListener.psm1
markekraus: @iSazonov Ok, that certainly makes things easier. If that's the direction to go the `Get-WebListenerUrl` idea will work.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @iSazonov  Ok I will add them and update the tests. On that note, I should probably extend this to included the HTTP (no s) Listener.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: I suggest to think only about _web test_ and move common thing to the WebListener module.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: I suggest make this formally - implement but not touch HttpListener and HTTP test in the PR to keep it small.

test/tools/HttpsListener/Program.cs
iSazonov: Ah, sorry - it is a server certificate.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1124,21 +1112,27 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
...
-        $result = Invoke-WebRequest -Uri $uri -SkipCertificateCheck
-        $failedMessage = 'Status: Failed'
-        $result.RawContent  | Should Match ([regex]::Escape($failedMessage ))
+        $uri = '{0}/Cert/' -f (Get-WebListenerUrl -Https)
+        $result = Invoke-WebRequest -Uri $uri -SkipCertificateCheck | 
+            Select-Object -ExcludeProperty Content |
+            ConvertFrom-Json
+        
+        $result.Status  | Should Be 'FAILED'
+        $result.Status  | Should Not Be 'OK'
```
iSazonov: Please remove one line - it is dup. Below too.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1124,21 +1112,27 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
...
     # https://github.com/PowerShell/PowerShell/issues/4650
     It "Verifies Invoke-WebRequest Certificate Authentication Successful with -Certificate" -skip:$IsOSX {
-        $uri = 'https://localhost:8083/'
-        $certificate = GetSelfSignedCert
-        $result = Invoke-WebRequest -Uri $uri -Certificate $certificate -SkipCertificateCheck
-        $successMessage = 'Status: OK'
-        $result.RawContent | Should Match ([regex]::Escape($successMessage))
-        $result.RawContent | Should Match $certificate.Thumbprint
+        $uri = '{0}/Cert/' -f (Get-WebListenerUrl -Https)
+        $certificate = Get-TestClientCertificate
```
iSazonov: I think we should have the same prefix for all cmdlets from `WebListener` module.

test/tools/Modules/WebListener/WebListener.psd1
```diff
@@ -0,0 +1,10 @@
+@{
+    ModuleVersion = '1.0.0'
+    GUID = '90572e25-3f15-49b0-8f25-fb717d3ef46a'
+    Author = 'Mark Kraus'
+    CompanyName = ''
+    Copyright = ''
```
iSazonov: Please remove the two dummy property.

test/tools/Modules/WebListener/WebListener.psd1
```diff
@@ -0,0 +1,10 @@
...
+    Description = 'An HTTP and HTTPS Listener for testing purposes'
+    RootModule = 'WebListener.psm1'
+    FunctionsToExport = @('Start-WebListener','Stop-WebListener', 'Get-WebListener', 'Get-WebListenerUrl','Get-TestClientCertificate')
```
iSazonov: From my experience - It is better to put every name on separate line and sort - easy search, easy add new.

test/tools/Modules/WebListener/WebListener.psm1
```diff
@@ -0,0 +1,147 @@
...
+        return $Script:WebListener
+    }
+}
+
+function Stop-WebListener 
+{
+    [CmdletBinding(ConfirmImpact = 'Low')]
+    [OutputType([Void])]
+    param 
+    ( )
```
iSazonov: Formatting -> `param()`

test/tools/Modules/WebListener/WebListener.psm1
```diff
@@ -0,0 +1,147 @@
...
+    {
+        $Script:WebListener.job | Stop-Job -PassThru | Remove-Job
+        $Script:WebListener = $null
+    }
+}
+
+function Get-TestClientCertificate {
+    [CmdletBinding(ConfirmImpact = 'Low')]
+    [OutputType([System.Security.Cryptography.X509Certificates.X509Certificate2])]
+    param ( )
```
iSazonov: Formatting -> `param()`

test/tools/Modules/WebListener/WebListener.psm1
```diff
@@ -0,0 +1,147 @@
...
+    process {
+        $parentPath = Split-Path -parent (get-command WebListener).Path
```
iSazonov: The same about path.

test/tools/Modules/WebListener/WebListener.psm1
```diff
@@ -0,0 +1,147 @@
...
+    [CmdletBinding()]
+    [OutputType([Uri])]
+    param (
+        [switch]$Https
+    )
+    process {
+        $runningListener = Get-WebListener
+        if ($null -eq $runningListener -or $runningListener.GetStatus() -ne 'Running')
+        {
+            return
```
iSazonov: Return `$null` ?

test/tools/WebListener/Controllers/CertController.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    public class CertController : Controller
+    {
+        public ActionResult Index()
+        {
+            // X509Certificate2 objects do not serialize as JSON. Create a HashTable instead
+            Hashtable output = new Hashtable 
+            {
+                {"Status", "FAILED"}
+            };
```
iSazonov: Please remove ';'. Below too.

test/tools/WebListener/Controllers/CertController.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using mvc.Models;
+using System.Collections;
```
iSazonov: Please sort the list. In other files too.

iSazonov (commented):
Great work!

test/tools/Modules/WebListener/WebListener.psm1
markekraus: shouldn't it be `param ()` with a space to match formatting with everything else?

test/tools/Modules/WebListener/WebListener.psm1
markekraus: I'm sorry, I'm sure what you are suggesting on this one.

test/tools/WebListener/Controllers/CertController.cs
markekraus: Since I'm fairly new to C#, what is recommended?

```csharp
using Microsoft.AspNetCore.Mvc;
using mvc.Models;
using System;
using System.Collections;
using System.Collections.Generic;
using System.Diagnostics;
using System.Linq;
using System.Threading.Tasks;
```

or
```csharp
using System;
using System.Collections;
using System.Collections.Generic;
using System.Diagnostics;
using System.Linq;
using System.Threading.Tasks;
using Microsoft.AspNetCore.Mvc;
using mvc.Models;
```

or something else?

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: will change to `Get-WebListenerClientCertificate`

test/tools/Modules/WebListener/WebListener.psd1
markekraus: will fix

test/tools/Modules/WebListener/WebListener.psd1
markekraus: me too, I was just following conventions in HttpListener. I will fix it.

test/tools/Modules/WebListener/WebListener.psm1
markekraus: They are equivalent, but I will make it explicit.

test/tools/WebListener/Controllers/CertController.cs
markekraus: will fix

test/tools/Modules/WebListener/WebListener.psm1
```diff
@@ -0,0 +1,147 @@
...
+        $serverPfx           = 'ServerCert.pfx'
+        $serverPfxPassword   = 'password'
+        $initCompleteMessage = 'Now listening on'
+        
+        
+        $timeOut = (get-date).AddSeconds($initTimeoutSeconds)
+        $Job = Start-Job {
+            $path = Split-Path -parent (get-command WebListener).Path
+            Push-Location $path
+            dotnet $using:appDll $using:serverPfx $using:serverPfxPassword $using:HtttpPort $using:HtttpsPort
```
markekraus: `$using:HtttpsPort` -> `$using:HttpsPort` and `$using:HtttpPort` -> `$using:HtttpPort` should fix the build errors.

test/tools/Modules/WebListener/WebListener.psm1
iSazonov: I don't found `param ()` pattern in the repo.

test/tools/WebListener/Controllers/CertController.cs
iSazonov: Second looks better.
Idea is to easy search a string - so we can order and split on logical blocks.

test/tools/Modules/WebListener/WebListener.psd1
```diff
@@ -2,9 +2,13 @@
...
     Author = 'Mark Kraus'
-    CompanyName = ''
-    Copyright = ''
     Description = 'An HTTP and HTTPS Listener for testing purposes'
     RootModule = 'WebListener.psm1'
-    FunctionsToExport = @('Start-WebListener','Stop-WebListener', 'Get-WebListener', 'Get-WebListenerUrl','Get-TestClientCertificate')
+    FunctionsToExport = @(
+        'Start-WebListener'
+        'Stop-WebListener'
+        'Get-WebListener'
```
iSazonov: 'G' after 'S' ? 😄

markekraus: They are "S" class cmdlets... higher priority 😆. I will fix it.

test/tools/Modules/WebListener/WebListener.psm1
markekraus: resolved

test/tools/WebListener/Controllers/CertController.cs
markekraus: resolved

test/tools/WebListener/Controllers/CertController.cs
markekraus: resolved

test/tools/Modules/WebListener/WebListener.psm1
markekraus: resolved

test/tools/Modules/WebListener/WebListener.psm1
markekraus: resolved

test/tools/Modules/WebListener/WebListener.psm1
markekraus: I will make them `param()` in next commit

test/tools/Modules/WebListener/WebListener.psd1
markekraus: resolved

test/tools/Modules/WebListener/WebListener.psd1
markekraus: resolved

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: resolved

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: resolved

test/tools/Modules/WebListener/WebListener.psm1
```diff
@@ -0,0 +1,146 @@
...
+        $Script:WebListener = $null
+    }
+}
+
+function Get-WebListenerClientCertificate {
+    [CmdletBinding(ConfirmImpact = 'Low')]
+    [OutputType([System.Security.Cryptography.X509Certificates.X509Certificate2])]
+    param ()
+    process {
+        $parentPath = Split-Path -parent (get-command WebListener).Path
```
iSazonov: I meant we can `$MyInvocation.MyCommand.Module.ModuleBase `

markekraus: ClientCert.pfx is in the `WebListener` published directory. If I switch to `$MyInvocation.MyCommand.Module.ModuleBase` I would need to move it to the `WebListener` module folder. Is that what you would suggest?

test/tools/Modules/WebListener/WebListener.psd1
markekraus: resolved

test/tools/Modules/WebListener/WebListener.psm1
markekraus: resolved

test/tools/Modules/HttpsListener/HttpsListener.psm1
markekraus: reolved

test/tools/Modules/HttpsListener/HttpsListener.psd1
markekraus: reolved

test/tools/HttpsListener/Program.cs
markekraus: resolved

test/tools/HttpsListener/Pages/Index.cshtml.cs
markekraus: resolved

test/tools/HttpsListener/Pages/Index.cshtml.cs
markekraus: resolved

test/tools/HttpsListener/Pages/Index.cshtml.cs
markekraus: resolved by switchng from razor pages to MVC.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: resolved

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: resolved

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: resolved

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: resolved

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: reolved

test/tools/Modules/WebListener/WebListener.psm1
iSazonov: Looks good to me.

markekraus: @iSazonov for clarity, are you asking me to move it and update the code or leave it as is?

iSazonov: I agree to move ClientCert.pfx to module folder.

markekraus: resolved

test/tools/Modules/WebListener/WebListener.psm1
```diff
@@ -0,0 +1,136 @@
...
+    [int]$HttpsPort
+    [System.Management.Automation.Job]$Job
+
+    WebListener () { }
+
+    [String] GetStatus() 
+    {
+        return $This.Job.JobStateInfo.State
+    }
+
```
iSazonov: Extra line.

test/tools/WebListener/Controllers/CertController.cs
iSazonov: I still see ';'.

test/tools/WebListener/Controllers/CertController.cs
```diff
@@ -0,0 +1,43 @@
...
+                {
+                    {"Status"      , "OK"},
+                    {"Thumbprint"  , HttpContext.Connection.ClientCertificate.Thumbprint},
+                    {"Subject"     , HttpContext.Connection.ClientCertificate.Subject},
+                    {"SubjectName" , HttpContext.Connection.ClientCertificate.SubjectName.Name},
+                    {"Issuer"      , HttpContext.Connection.ClientCertificate.Issuer},
+                    {"IssuerName"  , HttpContext.Connection.ClientCertificate.IssuerName.Name},
+                    {"NotAfter"    , HttpContext.Connection.ClientCertificate.NotAfter},
+                    {"NotBefore"   , HttpContext.Connection.ClientCertificate.NotBefore}
+                };
```
iSazonov: I still see ';'.

test/tools/WebListener/WebListener.csproj
```diff
@@ -0,0 +1,20 @@
...
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <DotNetCliToolReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Tools" Version="2.0.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+     <None Include="ServerCert.pfx" CopyToPublishDirectory="PreserveNewest" />
```
iSazonov: Maybe put it near client certificate and remove the group?

test/tools/WebListener/Controllers/CertController.cs
markekraus: The Semicolon has to be there. When I tried to remove it I get compile errors saying it is missing.

test/tools/WebListener/Controllers/CertController.cs
markekraus: The Semicolon has to be there. When I tried to remove it I get compile errors saying it is missing.

test/tools/WebListener/Controllers/CertController.cs
iSazonov: Sorry. Closed.

test/tools/Modules/WebListener/WebListener.psm1
markekraus: resolved

test/tools/WebListener/WebListener.csproj
markekraus: resolved

iSazonov:
@markekraus Thank you for your patience and your speedy work.

LGTM.

markekraus:
@iSazonov Thanks for being patient with me too and for all your suggestions!

SteveL-MSFT:
@iSazonov can you formally approve?

iSazonov (approved):

iSazonov:
It will be good to get a review from test area experts.

TravisEz13 (approved):

iSazonov:
Thanks! I'll merge on next week.

iSazonov:
@markekraus Many thanks for great work!

The PR was merged and  way is opened for follow work:
- [ ] - Cleanup HTTPS tests, replace #region with "Context".
- [ ] - Cleanup HTTP test to use WebListener.

Please move the points to a tracking Issue.

markekraus:
@iSazonov  I have opened #4179 to track the HTTPS tests. 

Should I continue the work for the second item under the already open #2504 or should I create a new issue?

iSazonov:
@markekraus #2504 is good. Please leave a message about WebListener there for cross reference. And if you want work on this you should ask Travis there.

PR #4627, opened 2017-08-20
Correct linux.md macOS brew reinstall command

TheNewStellW:
Updated macOS reinstall command. Using the originally written "brew reinstall powershell" fails, as the reinstall needs to be done using "brew cask reinstall powershell".


TravisEz13 (approved):

TravisEz13:
nice catch

Issue #4639, opened 2017-08-22
Add support for multiple response headers with the same name from HttpListener


markekraus:
Multiple response header support was added in #4494 but the HttpListener concatenates multiple headers with the same name in Linux/macOS instead of returning multiple lines with the same header name.

In Windows:
```
HTTP/1.1 200 OK
Content-Length: 2
Content-Type: text/plain
Server: Microsoft-HTTPAPI/2.0
X-Fake-Header: testvalue01
X-Fake-Header: testvalue02
Date: Tue, 22 Aug 2017 09:49:37 GMT
```

In Linux/macOS:
```
HTTP/1.1 200 OK
X-Fake-Header: testvalue01, testvalue02
Server: Microsoft-NetCore/2.0
Date: Tue, 22 Aug 2017 09:18:18 GMT
Content-Type: text/plain
Content-Length: 2
```

This results in a false failure of the `Verifies Invoke-WebRequest Supports Multiple response headers with same name` test on Linux/macOS.

We should find a cross-platform method of returning a multiple header response. (Possibly with the new listener inf #4622)

markekraus:
@iSazonov this issue is still outstanding. We still do not have a test in place for Linux or macOS to ensure the multiple headers feature is working. The merged PR referenced is to address the nightly build errors the tests were causing.

iSazonov:
Ah, sorry - I reopen the Issue.

markekraus:
I think that WebListener supports this properly on all platforms. since we are moving all the tests to WebListener and off of HttpListener, I will come back to this when that transition is complete.

PR #4494, opened 2017-08-04
Add support for Content Headers to BasicHtmlWebResponseObject and HtmlWebResponseObject

markekraus:
Fixes #4467

`System.Net.Http.HttpResponseMessage` has split the Content headers from the Response headers. This split appears to be intentional and permanent. This split results in the `Content-Type` and `Content-Length` headers to be missing from the `WebResponseObject.Headers` dictionary and from the `RawContent` property of `BasicHtmlWebResponseObject` and `HtmlWebResponseObject`.

This adds the Content headers back to those locations to make it similar to 5.1.

This also adds support to `RawContent` for multiple response headers with the same field name. `System.Net.Http.Headers.HttpContentHeaders` and `System.Net.Http.Headers.HttpResponseHeaders` implement values as arrays.

PR #4640, opened 2017-08-22
[feature] Mark Multiple Response Header Test Pending

markekraus:
The `Verifies Invoke-WebRequest Supports Multiple response headers with same name` test from #4494 is failing on Linux and macOS due to an implementation difference of `HttpListener` on Linux and macOS where multiple response headers are concatenated instead of sent as separate headers.

In Windows:
```
HTTP/1.1 200 OK
Content-Length: 2
Content-Type: text/plain
Server: Microsoft-HTTPAPI/2.0
X-Fake-Header: testvalue01
X-Fake-Header: testvalue02
Date: Tue, 22 Aug 2017 09:49:37 GMT
```

In Linux/macOS:
```
HTTP/1.1 200 OK
X-Fake-Header: testvalue01, testvalue02
Server: Microsoft-NetCore/2.0
Date: Tue, 22 Aug 2017 09:18:18 GMT
Content-Type: text/plain
Content-Length: 2
```

This is not a failure of the test but an issue with the current method used for serving test content with `HttpListener`.

Issue  #4639 has been created to track creating a cross-platform multiple header response solution for this test. In the mean time, mark this test as pending to prevent nightly failures.

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -1204,7 +1204,9 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         $result.Output.RawContent | Should Match ([regex]::Escape('Content-Length: 2'))
     }
 
-    It "Verifies Invoke-WebRequest Supports Multiple response headers with same name" {
+    # Test pending due to HttpListener limitation on Linux/macOS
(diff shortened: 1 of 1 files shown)


anmenaga:
@TravisEz13 @SteveL-MSFT This needs to be merged asap as it fixes a failure in night test pass.

TravisEz13 (approved):

TravisEz13:
merging before 24 hours, because this fixes a test failure.  @anmenaga Please in the future, be clear if you reviewed the change.

PR #4653, opened 2017-08-23
add cross-compile options to build script

SteveDesmond-ca:
This PR adds cross-compile options to the Linux build script `build.sh`. See #2463.

By calling `./build.sh --arch=arm --rid=linux-arm`, both the native and managed components are built for Linux on ARM. Other architectures and RIDs can easily be added as they are brought up.

Calling `./build.sh` without args should continue to work identically. The argument parsing is done in the same style as [all](https://github.com/dotnet/coreclr/blob/master/build.sh) [the](https://github.com/dotnet/corefx/blob/master/src/Native/build-native.sh) [dotnet](https://github.com/dotnet/core-setup/blob/master/src/corehost/build.sh) [build](https://github.com/dotnet/sdk/blob/master/build.sh) [scripts](https://github.com/dotnet/cli/blob/master/build.sh).

3 things of note, any/all of which can be done as part of this PR if desired, or a separate PR, or not at all:
1. `psl-native-test` is currently skipped for cross-compiled ARM builds because a QEMU ARM chroot/rootfs is not available to run the tests. [At first glance](https://www.tomaz.me/2013/12/02/running-travis-ci-tests-on-arm.html) it looks like Travis can handle it fine.

2. While the build currently succeeds, running Powershell on an ARM device currently segfaults. I can look into this either on this PR or a separate one -- let me know which is preferred.

3. Should I add documentation, similar to [CoreCLR's](https://github.com/dotnet/coreclr/blob/master/Documentation/building/cross-building.md)? Not sure how "supported" we want this to be at this time.


stevedesmond-ca:
cc @SteveL-MSFT

stevedesmond-ca:
Just ran gdb on it, segfault dump:
```
0xb674aecc in VirtualCallStubManager::predictStubKind(unsigned int) () from /home/vtsv/powershell/libcoreclr.so
(gdb) bt
#0  0xb674aecc in VirtualCallStubManager::predictStubKind(unsigned int) () from /home/vtsv/powershell/libcoreclr.so
#1  0xb67501d6 in VirtualCallStubManager::getStubKind(unsigned int) () from /home/vtsv/powershell/libcoreclr.so
#2  0xb674d1b4 in VirtualCallStubManager::FindStubManager(unsigned int, VirtualCallStubManager::StubKind*) ()
   from /home/vtsv/powershell/libcoreclr.so
#3  0xb674e98e in VSD_ResolveWorker () from /home/vtsv/powershell/libcoreclr.so
#4  0xb67f4b30 in ResolveWorkerAsmStub () from /home/vtsv/powershell/libcoreclr.so
#5  0xaa205a6e in ?? ()
```
Sorry, things this deep are beyond my pay grade ;)

SteveL-MSFT:
@stevedesmond-ca very cool!  I'll try to work on this tomorrow or Friday

TravisEz13:
I'm not sure if this script is being kept up to date.  It might be better to update build.psm1 and get rid of this.

stevedesmond-ca:
Segfault issue is being tracked at dotnet/coreclr#13667

SteveL-MSFT:
@TravisEz13 I have changes to build.psm1 to enable cross-compile in my fork however it's pending the coreclr issue.  If we don't depend on build.sh, perhaps we should consider getting rid of it since it's not up to date with build.psm1

stevedesmond-ca:
My only hesitance with removing the bash script is the old chicken and egg problem: how do you build Powershell on a system that can't already run Powershell? And maybe that's fine / not a supported use case.

TravisEz13:
You can cross-compile in the build.psm1.  You cannot cross-package currently, but I usually just manually copy the files to the target platform and get packaging working there.

SteveL-MSFT:
@TravisEz13 perhaps the @PowerShell/powershell-maintainers should own the decision on whether we still want build.sh?

iSazonov:
Who last run the build.sh and why?

TravisEz13:
It looks like @stevedesmond-ca has run it and he says why.   I don't use it because it's not been maintained and building properly take more/different steps than what is in build.sh.

iSazonov:
So if we add the cross-compile scenario in our installation guide we can safely remove `build.sh`?

TravisEz13:
I think so.

iSazonov:
@stevedesmond-ca Could you please check that the cross-compile scenario works for you (for arm)?

iSazonov:
@stevedesmond-ca Could you please check that the cross-compile scenario works for you (for arm)?

stevedesmond-ca:
Just took a look at @SteveL-MSFT's [branch](https://github.com/SteveL-MSFT/PowerShell/tree/raspberry-pi), and I think it's a much better implementation. Since there is an Ubuntu release for Powershell 6b6, we can rely on use Powershell to build Powershell on Ubuntu, and it's made clear in the updated `build.psm1` that Ubuntu is the only supported distro for cross-compiling.

I made [one quick amendment](https://github.com/PowerShell/PowerShell/pull/4653/commits/b9ee4acc46cff6652cebcb92237222c3dc65c062) to Steve's branch  to ensure that the ARM toolchain file is getting called correctly when `Start-PSBuild $Runtime linux-arm` is run, and pulled that all into this PR. Hopefully that's OK, Steve! Otherwise, feel free to cherry-pick [that single commit](https://github.com/PowerShell/PowerShell/pull/4653/commits/b9ee4acc46cff6652cebcb92237222c3dc65c062) into your branch and create a new PR that way.

SteveL-MSFT:
@stevedesmond-ca I've incorporated your change into my PR (which incidentally will remove build.sh since it's not being maintained).  You can close this PR.

stevedesmond-ca:
Sounds good, thanks!

iSazonov:
@stevedesmond-ca Thanks! Come back with new contributions!

PR #4657, opened 2017-08-24
Remove parsing script file names as assembly names

### LINKED ISSUE #4193: Apostrophe in filepath breaks Import-Module when the module has a class in it
When there is an apostrophe in the filepath to a module Import-Module will work with modules that only have functions in however will fail if there is a class in the module.  The exact same code will work when the module is located on a different path.  

This situation was encountered due to the default OneDrive for Business path including an apostrophe.  When using the user windowspowershell module with folder redirection all use modules now have an apostrophe in the path.  

This affects both Version 5.1 and Version 6beta3

Steps to reproduce
------------------
Create Folder C:\Te'mp
Save the below code to C:\Te'mp\TypeName.psm1
```powershell
# Define a class
class TypeName
{
   # Property with validate set
   [ValidateSet("val1", "Val2")]
   [string] $P1

   # Constructor
   TypeName ([string] $s)
   {
       $this.P1 = $s       
   }
}

```

```powershell
Import-Module 'C:\te''mp\TypeName.psm1'

```

Expected behavior
-----------------

```none
Module Loaded Successfully / No Output
```

Actual behavior
---------------

```none
Import-Module : The given assembly name or codebase was invalid. (Exception from HRESULT: 0x80131047)
At line:1 char:1
+ Import-Module 'C:\Te''mp\TypeName.psm1'
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Import-Module], FileLoadException
    + FullyQualifiedErrorId : System.IO.FileLoadException,Microsoft.PowerShell.Commands.ImportModuleCommand
```

Environment data
----------------

```powershell
> $PSVersionTable
OS                             Microsoft Windows 10.0.14393

PSVersion                      5.1.15063.413                                                                                                
BuildVersion                   10.0.15063.413                                                                                               
```

iSazonov:
Fix #4193.

### Problem

Before the fix we pass a script file name into 'AssemblyName(assemblyName)' constructor. The constructor is trying to parse the name as an assembly name:
```
'Name <,Culture = CultureInfo> <,Version = Major.Minor.Build.Revision> <, StrongName> <,PublicKeyToken> '\0''
```
See notes in https://msdn.microsoft.com/en-us/library/system.reflection.assemblyname(v=vs.110).aspx

### Fix
Replace 'new AssemblyName(assemblyName)' constructor with 'new
AssemblyName()' to exclude parsing the script file name.

### Additional cosiderations
The PR replaces the frozen PR #4270.

### DIFF
--- a/src/System.Management.Automation/engine/parser/PSType.cs
+++ b/src/System.Management.Automation/engine/parser/PSType.cs
@@ -1118,13 +1118,17 @@ namespace System.Management.Automation.Language
                                       : rootAst.Extent.File
                                       .Replace('\\', (char)0x29f9)
                                       .Replace('/', (char)0x29f9)
-                                      .Replace(',', (char)0x201a)
                                       .Replace(':', (char)0x0589));
 
-            var assembly = AssemblyBuilder.DefineDynamicAssembly(new AssemblyName(assemblyName),
+            // We can not use 'AssemblyName(string assemblyName)' constructor because it is trying to parse the assembly name
+            // as 'Name <,Culture = CultureInfo> <,Version = Major.Minor.Build.Revision> <, StrongName> <,PublicKeyToken> '\0''
+            // See notes in https://msdn.microsoft.com/en-us/library/system.reflection.assemblyname(v=vs.110).aspx
+            var assemblyObj = new AssemblyName();
+            assemblyObj.Name = assemblyName;
+
+            var assembly = AssemblyBuilder.DefineDynamicAssembly(assemblyObj,
                                                                  AssemblyBuilderAccess.RunAndCollect, GetAssemblyAttributeBuilders());
             var module = assembly.DefineDynamicModule(assemblyName);
-
             var defineTypeHelpers = new List<DefineTypeHelper>();
             var defineEnumHelpers = new List<DefineEnumHelper>();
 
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/engine/parser/PSType.cs
```diff
@@ -1118,13 +1118,17 @@ internal static Assembly DefineTypes(Parser parser, Ast rootAst, TypeDefinitionA
...
                                       .Replace('/', (char)0x29f9)
-                                      .Replace(',', (char)0x201a)
                                       .Replace(':', (char)0x0589));
 
-            var assembly = AssemblyBuilder.DefineDynamicAssembly(new AssemblyName(assemblyName),
+            // We can not use 'AssemblyName(string assemblyName)' constructor because it is trying to parse the assembly name
+            // as 'Name <,Culture = CultureInfo> <,Version = Major.Minor.Build.Revision> <, StrongName> <,PublicKeyToken> '\0''
+            // See notes in https://msdn.microsoft.com/en-us/library/system.reflection.assemblyname(v=vs.110).aspx
+            var assemblyObj = new AssemblyName();
+            assemblyObj.Name = assemblyName;
```
lzybkr: This seems like a bypass that we shouldn't use. I'm worried that some random non-CLR code might parse the assembly name (e.g. PowerShell) in a manner similar to how the CLR does when clearly.

The goal here was to make it easy to map the assembly to the original source, but in some ways this already fails because of the path.

A better fix is probably to use a simple name like `PowerShell Dynamic Class` possibly with a counter, but to **add** a constructor to `DynamicClassImplementationAttribute` and pass the filename as an argument to that attribute.

This way we have a non-hacky way of mapping assemblies to script files.

lzybkr: Here is my approach - it could use a test but it works: https://github.com/lzybkr/PowerShell/tree/fix_dyn_class_assem

iSazonov: Thanks for clarify! Your approach looks good. Feel free to close the PR and push new one.

PR #4684, opened 2017-08-27
clean up coreclr/fullclr statements in tracesource

SteveL-MSFT:
removed `#if` statements for CORECLR/!CORECLR where it no longer applies

there is one remaining as System.Configuration.ConfigurationException doesn't exist in corefx

### DIFF
--- a/src/System.Management.Automation/utils/StructuredTraceSource.cs
+++ b/src/System.Management.Automation/utils/StructuredTraceSource.cs
@@ -290,6 +290,7 @@ namespace System.Management.Automation
                 FullName = fullName;
                 _name = name;
 
+                // TODO: move this to startup json file instead of using env var
                 string tracingEnvVar = Environment.GetEnvironmentVariable("MshEnableTrace");
 
                 if (String.Equals(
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/utils/StructuredTraceSource.cs
```diff
@@ -290,7 +290,6 @@ internal PSTraceSource(string fullName, string name, string description, bool tr
                 FullName = fullName;
                 _name = name;
 
-#if !CORECLR // TraceSource.Attributes is not in CoreCLR
                 string tracingEnvVar = Environment.GetEnvironmentVariable("MshEnableTrace");
```
iSazonov: Do we want to depend on an environment variable?

SteveL-MSFT: In general, I don't want PowerShell to depend on environment variables.  We should move this to the startup json file.  I'll add a TODO comment.

daxian-dbw (approved):

iSazonov (approved):

PR #4689, opened 2017-08-28
Modify travis build to upload a daily build badge based on platform

JamesWTruher:
change README.md to view a platform specific badge
This will give us better visibility into our platform specific builds, and improve efficiency when investigating a build failure

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,9 +108,9 @@ If you have any problems building, please consult the developer [FAQ][].
 
 ### Build status of nightly builds
 
-| AppVeyor (Windows)       | Travis CI (Linux / macOS) | Code Coverage Status |
-|--------------------------|---------------------------|----------------------|
-| [![av-nightly-image][]][av-nightly-site] | [![tv-nightly-image][]][tv-site] | [![cc-image][]][cc-site] |
+| AppVeyor (Windows)       | Travis CI (Linux) | Travis CI (macOS) | Code Coverage Status |
+|--------------------------|-------------------|-------------------|----------------------|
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -108,9 +108,9 @@ If you have any problems building, please consult the developer [FAQ][].
 
 ### Build status of nightly builds
 
-| AppVeyor (Windows)       | Travis CI (Linux / macOS) | Code Coverage Status |
-|--------------------------|---------------------------|----------------------|
-| [![av-nightly-image][]][av-nightly-site] | [![tv-nightly-image][]][tv-site] | [![cc-image][]][cc-site] |
+| AppVeyor (Windows)       | Travis CI (Linux) | Travis CI (OSX)  | Code Coverage Status |
```
SteveL-MSFT: I believe Apple has standardized on macOS vs OSX, so we should be consistent with macOS (ignore the fact we have $isOSX, unfortunately it's a breaking change now)

SteveL-MSFT (changes requested):

README.md
JamesWTruher: shouldn't we be rather self consistent? I believe that it will be more confusing for our users unless we also make the change in our variables

SteveL-MSFT: Agree we should be consistent with the variables, but that's outside the scope of this PR.  I think at least within the README.md, we can be consistent on using macOS (which it is currently)

JamesWTruher: @SteveL-MSFT updated to include this change

SteveL-MSFT: Still shows OSX, did you push?

JamesWTruher: @SteveL-MSFT yup

SteveL-MSFT (approved):
LGTM

JamesWTruher:
@TravisEz13 can we get this in?

PR #4692, opened 2017-08-28
Add explicit ContentType detection to Invoke-RestMethod

### LINKED ISSUE #3267: Webcmdlets should parse the <html><head><meta charset="foo"> attribute for the correct encoding if not in http header

dantraMSFT:
fix https://github.com/PowerShell/PowerShell/issues/3267

This change removes the call to ContentHelper.GetEncoding since it's logic for returning a default encoding when a ContentType header is not found disables subsequent checks for meta charset definitions.

All variations of the Invoke-WebRequest tests have been mirrored for Invoke-RestMethod.  Verbose output is used to verify the encoding since Invoke-RestMethod returns the json content directly instead of a response object. (See ExecuteRestMethod)


test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -296,8 +296,71 @@ function ExecuteWebRequest
...
+{
+    param
+    (
+        [Parameter(Mandatory)]
+        [string]
+        $Uri,
+
+        [switch] $UseBasicParsing
+    )
+    $result = [PSObject]@{Output = $null; Error = $null; Encoding = $null; Content = $null}
```
JamesWTruher: i think that '[PSObject]' should probably be '[PSCustomObject]'  - the `[PSObject]` doesn't actually convert the type away from a hashtable. So, `[PSObject]` can be removed, or changed to `[PSCustomObject]`

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -296,8 +296,71 @@ function ExecuteWebRequest
...
+                if ($line.StartsWith($verboseEncodingPrefix))
+                {
+                    $encodingName = $item.SubString($verboseEncodingPrefix.Length).Trim()
+                    $result.Encoding = [System.Text.Encoding]::GetEncoding($encodingName)
+                    break
+                }
+            }
+            if ($result.Encoding -eq $null)
+            {
+                throw "Encoding not found in verbose output. Lines: $($result).Verbose.Count Content:$($result.Verbose)"
```
JamesWTruher: `$($result).Verbose.Count` should be `$($result.Verbose.Count)`

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -296,8 +296,71 @@ function ExecuteWebRequest
...
+        if ($result.Verbose -eq $null)
+        {
+            throw "No verbose output was found"
+        }
+    }
+    catch
+    {
+        $result.Error = $_ | select-object * | Out-String
+    }
+    finally
```
JamesWTruher: since the $TESTDRIVE is removed by pester, you only need set the verbosePreference back.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1765,6 +1828,224 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
...
+            $output = '<html><head><meta charset="Unicode"></head></html>'
+            $expectedEncoding = [System.Text.Encoding]::GetEncoding('Unicode')
+            $response = ExecuteRestMethod -Uri "http://localhost:8081/PowerShell?test=response&output=$output" -UseBasicParsing
+
+            $response.Error | Should BeNullOrEmpty
+            $response.Encoding.EncodingName | Should Be $expectedEncoding.EncodingName
+        }
+
+        It "Verifies Invoke-WebRequest detects charset meta value when newlines are encountered in the element." {
+            $output = @'
```
JamesWTruher: just a question, would: 
```
"<html>`n<head>`n<meta`n   charset=""Unicode""  `n>`n</head>`n</html>`n"
```
be the same?

JamesWTruher (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: The test drive isn't cleared until the end of the context block. line 322 uses a test for the file's existence. With the current configuration, subsequent calls to `ExecuteRestMethod` could result in false positives for the test on line 322 if the file is not removed at the end at the end of the function. This cleanup action would need to remain or the filename generated on each call to  `ExecuteRestMethod` would need to be unique.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -296,8 +296,71 @@ function ExecuteWebRequest
...
+    catch
+    {
+        $result.Error = $_ | select-object * | Out-String
+    }
+    finally
+    {
+        $VerbosePreference = $verbosePreferenceSave
+        if (Test-Path -Path $verboseFile)
+        {
+            Remove-Item -Path $verboseFile -ErrorAction Ignore
```
iSazonov: Usualy we use `SilentlyContinue` not `Ignore`.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: As @markekraus indicates, this is by design and simpler than generating file names.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: Thanks, I'll fix it.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: Will do.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: Gotcha, I'll change it.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: Yes, but I find the HERE string much easier to read and it avoids me making silly escaping errors.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: Closed.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: Fixed

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: Fixed

markekraus (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -296,8 +296,71 @@ function ExecuteWebRequest
     return $result
 }
 
+[string] $verboseEncodingPrefix = 'Content encoding: '
```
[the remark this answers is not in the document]
dantraMSFT: I don't think I want a locally scoped variable that has to be initialized each call.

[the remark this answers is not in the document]
markekraus: fair enough

JamesWTruher (approved):

daxian-dbw:
The `WebCmdlets.Tests.ps1` has conflicts due to the merge of #4622 🤕

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -296,6 +296,158 @@ function ExecuteWebRequest
...
+        if (Test-Path -Path $verboseFile)
+        {
+            Remove-Item -Path $verboseFile -ErrorAction SilentlyContinue
+        }
+    }
+
+    return $result
+}
+
+function GetSelfSignedCert {
```
markekraus: This function is no longer needed it was removed in #4622

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -831,6 +983,31 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
     #endregion SkipHeaderVerification Tests
 
+    #region Certificate Authentication Tests
```
markekraus: Everything in this region was removed in #4622

markekraus (changes requested):
Artifacts from merge should be removed.

daxian-dbw:
@markekraus thanks! @dantraMSFT can you go through the file `WebCmdlets.Tests.ps1` and see if things are good after resolving conflicts?

dantraMSFT:
I'll verify the merge.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
dantraMSFT: it looks like Dongbo removed it when he merged.

markekraus: yes. the commit came in after I hit submit. both of these comments in this review are fixed now

Issue #4686, opened 2017-08-27
PowerShell Core Enter-PSSession not exiting gracefully in Windows when connected to Linux.


MaximoTrinidad:
I'm happy to say that I love connectivity in both directions 'Linux-to-Windows' and Windows-to-Linux' using OpenSSH-Win64.  Now, in beta-6, I notice when exiting from Linux using Enter-PSSession  in Windows, it doesn't exit gracefully.  Meaning, that when typing 'Exit', its giving an error when it shouldn't.

Steps to reproduce
------------------
In Windows -> connect to Linux:
PS> Enter-PSSession -hostname orion -username maxt
: 
[orion]: PS /home/maxt> exit
(gives error message and then disconnect)

```powershell
Enter-PSSession -hostname orion -username maxt
```

Expected behavior
-----------------
Typing 'exit' should terminate session without any errors

Actual behavior
---------------
Error message when exiting session.
```
PowerShell v6.0.0-beta.6
Copyright (C) Microsoft Corporation. All rights reserved.

WARNING: Appended Windows PowerShell PSModulePath
Loading personal and system profiles took 2520ms.
PS C:\Program Files\PowerShell\6.0.0-beta.6> enter-pssession -HostName orion -UserName maxt
maxt@orion's password:
[orion]: PS /home/maxt> exit
A task was canceled.
The SSH client session has ended with error message: A task was canceled.
    + CategoryInfo          : ResourceUnavailable: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : System.Management.Automation.Remoting.PSRemotingDataStructureException

PS C:\Program Files\PowerShell\6.0.0-beta.6>
```

Environment data
----------------
Windows 10 Insider Build 16257
Ubuntu 16.04.3 LTS
openssh 0.0.19.0

```powershell
PS C:\Program Files\PowerShell\6.0.0-beta.6> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.6
OS                             Microsoft Windows 10.0.16275
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

```
[...]

PR #4693, opened 2017-08-28
[Regression] Fix for error on Enter-PSSession exit

PaulHigin:
This is a fix for issue #4686.

I inadvertently created this regression with PR #4475.  The problem is that connection state is terminated before the session close sequence completes.  The fix is to only clean up connection state on session close if the session was never established.

I also added code to Enter-PSSession to handle a similar case where the Cmdlet operation is stopped before the connection or session is created.  This is important for SSH remoting because PowerShell interacts with SSH and it is possible to cancel before a connection is even started.

```powershell
PS > Enter-PSSession -hostname <hostname> -User <username>
<username>@<hostname>'s password:
Ctrl+C
PS >
```

### DIFF
--- a/src/System.Management.Automation/engine/remoting/commands/PushRunspaceCommand.cs
+++ b/src/System.Management.Automation/engine/remoting/commands/PushRunspaceCommand.cs
@@ -1284,6 +1284,10 @@ namespace Microsoft.PowerShell.Commands
         {
             var sshConnectionInfo = new SSHConnectionInfo(this.UserName, ResolveComputerName(HostName), this.KeyFilePath, this.Port);
             var typeTable = TypeTable.LoadDefaultTypeFiles();
+
+            // Use the class _tempRunspace field while the runspace is being opened so that StopProcessing can be handled at that time.
+            // This is only needed for SSH sessions where a Ctrl+C during an SSH password prompt can abort the session before a connection
+            // is established.
             _tempRunspace = RunspaceFactory.CreateRunspace(sshConnectionInfo, this.Host, typeTable) as RemoteRunspace;
             _tempRunspace.Open();
             _tempRunspace.ShouldCloseOnPop = true;
--- a/src/System.Management.Automation/engine/remoting/commands/PushRunspaceCommand.cs
+++ b/src/System.Management.Automation/engine/remoting/commands/PushRunspaceCommand.cs
@@ -39,6 +39,7 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         public new Int32 ThrottleLimit { set { } get { return 0; } }
         private ObjectStream _stream;
+        private RemoteRunspace _tempRunspace;
 
         #endregion
 
@@ -534,6 +535,17 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         protected override void StopProcessing()
         {
+            var remoteRunspace = _tempRunspace;
+            if (remoteRunspace != null)
+            {
+                try
+                {
+                    remoteRunspace.CloseAsync();
+                }
+                catch (InvalidRunspaceStateException) { }
+                return;
+            }
+
             IHostSupportsInteractiveSession host = this.Host as IHostSupportsInteractiveSession;
             if (host == null)
             {
@@ -1272,10 +1284,12 @@ namespace Microsoft.PowerShell.Commands
         {
             var sshConnectionInfo = new SSHConnectionInfo(this.UserName, ResolveComputerName(HostName), this.KeyFilePath, this.Port);
             var typeTable = TypeTable.LoadDefaultTypeFiles();
-            var remoteRunspace = RunspaceFactory.CreateRunspace(sshConnectionInfo, this.Host, typeTable) as RemoteRunspace;
-            remoteRunspace.Open();
-            remoteRunspace.ShouldCloseOnPop = true;
-
+            _tempRunspace = RunspaceFactory.CreateRunspace(sshConnectionInfo, this.Host, typeTable) as RemoteRunspace;
+            _tempRunspace.Open();
+            _tempRunspace.ShouldCloseOnPop = true;
+            var remoteRunspace = _tempRunspace;
+            _tempRunspace = null;
+            
             return remoteRunspace;
         }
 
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
```diff
@@ -1403,6 +1403,7 @@ internal sealed class SSHClientSessionTransportManager : OutOfProcessClientSessi
         private StreamWriter _stdInWriter;
         private StreamReader _stdOutReader;
         private StreamReader _stdErrReader;
+        private bool _connectionEstablished;
```
JamesWTruher: should this have an explicit initialization to false?

src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
```diff
@@ -1642,6 +1648,7 @@ private void ProcessReaderThread(object state)
                     else
                     {
                         // Normal output data.
+                        if (!_connectionEstablished) { _connectionEstablished = true; }
```
JamesWTruher: should this just be simply `_connectionEstablished = true`

src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
JamesWTruher: should this be an explicit initialization to false?

JamesWTruher (approved):

src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
PaulHigin: No need as CLR guarantees all type fields are initialized to zero.

PaulHigin: No need as CLR guarantees all type fields are initialized to zero.

src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
PaulHigin: I prefer to do the assignment only once in the loop.

PaulHigin:
Thanks!

src/System.Management.Automation/engine/remoting/commands/PushRunspaceCommand.cs
```diff
@@ -1272,10 +1284,12 @@ private RemoteRunspace GetRunspaceForSSHSession()
         {
             var sshConnectionInfo = new SSHConnectionInfo(this.UserName, ResolveComputerName(HostName), this.KeyFilePath, this.Port);
             var typeTable = TypeTable.LoadDefaultTypeFiles();
-            var remoteRunspace = RunspaceFactory.CreateRunspace(sshConnectionInfo, this.Host, typeTable) as RemoteRunspace;
-            remoteRunspace.Open();
-            remoteRunspace.ShouldCloseOnPop = true;
-
+            _tempRunspace = RunspaceFactory.CreateRunspace(sshConnectionInfo, this.Host, typeTable) as RemoteRunspace;
```
dantraMSFT: You should comment why _tempRunspace is being used.  On the surface, the usage either implies reentrancy or threading issue; if so, StopProcessing's use of _tempRunspace is problematic.  Comments explaining it's expected usage would help.

dantraMSFT:
>                         string errorData = data.Substring(System.Management.Automation.Remoting.Server.NamedPipeErrorTextWriter.ErrorPrepend.Length);

You continue processing data so why isn't this also considered a _connectionEstablished state?  As it stands, it appears that there is a nebulas state where errors are processed but the connection cannot be closed via CloseAsync.
If I'm misreading it, a comment would help.

---
Refers to: src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs:1645 in 5924f77. [](commit_id = 5924f773bffa6d096a5473bc1e8449b858020b7d, deletion_comment = False)

dantraMSFT (changes requested):

src/System.Management.Automation/engine/remoting/commands/PushRunspaceCommand.cs
PaulHigin: This is a common pattern we use for Cmdlet async StopProcessing.  I'll add a comment.

PaulHigin: Regarding above comment about connection established.  I think it is possible for an error to occur before the first PSRP message is processed.

PaulHigin: However, we want to do immediate connection clean up only if PSRP messages are being processed because then we can rely on the protocol to clean up the client connection state.  This is just for the case where the session is abruptly severed and we can't count on PSRP for clean up.  I'll add comment to clarify.

SteveL-MSFT (commented):
The second source file isn't showing a diff and usually this is caused by one of them having LF and the other CRLF endings.  Usually what I do is explicitly save it with LF endings and push it and the diff will show up correctly.

PaulHigin:
Thanks!  Will fix.

PaulHigin:
@mirichmo 
I have responded to all CR comments.  Can this be merged?

PaulHigin:
@dantraMSFT 
Do you have any more comments or questions?

dantraMSFT (approved):

src/System.Management.Automation/engine/remoting/commands/PushRunspaceCommand.cs
```diff
@@ -1272,10 +1284,16 @@ private RemoteRunspace GetRunspaceForSSHSession()
...
-            var remoteRunspace = RunspaceFactory.CreateRunspace(sshConnectionInfo, this.Host, typeTable) as RemoteRunspace;
-            remoteRunspace.Open();
-            remoteRunspace.ShouldCloseOnPop = true;
 
+            // Use the class _tempRunspace field while the runspace is being opened so that StopProcessing can be handled at that time.
+            // This is only needed for SSH sessions where a Ctrl+C during an SSH password prompt can abort the session before a connection
+            // is established.
+            _tempRunspace = RunspaceFactory.CreateRunspace(sshConnectionInfo, this.Host, typeTable) as RemoteRunspace;
+            _tempRunspace.Open();
+            _tempRunspace.ShouldCloseOnPop = true;
```
mirichmo: Isn't this an order of operations issue? It seems like the flag should be set before the call to Open().

PaulHigin: Do you mean the ShouldCloseOnPop property?  No.  A runspace pop cannot occur before this method completes.

PR #4697, opened 2017-08-29
GetNetworkCredential() returns null if PSCredential has null or empty user name

### LINKED ISSUE #4696: PSCredential.Empty.GetNetworkCredential() throws NRE instead of returning null per docs
Steps to reproduce
------------------

```powershell
> [pscredential]::Empty.GetNetworkCredential()
```

Expected behavior
-----------------

```none
$null
```

Per [the docs](https://github.com/PowerShell/PowerShell/blob/fd047a82506ac6a38c20052f45683cf5407c9c3a/src/System.Management.Automation/engine/Credential.cs#L248), an "uninitialized" `PSCredential` object should return `null`

MSDN also documents it as such, though only on [the page of the explicit cast operator](https://msdn.microsoft.com/en-us/library/system.management.automation.pscredential.op_explicit) which just calls `GetNetworkCredential`

Actual behavior
---------------

```none
Exception calling "GetNetworkCredential" with "0" argument(s): "Object reference not set to an instance of an object."
At line:1 char:1
+ [pscredential]::Empty.GetNetworkCredential()
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + FullyQualifiedErrorId : NullReferenceException
```

Environment data
----------------

```powershell
> $PSVersionTable

PSVersion                      5.0.10586.117
BuildVersion                   10.0.10586.117
```

---

[`_userName` is `null`](https://github.com/PowerShell/PowerShell/blob/fd047a82506ac6a38c20052f45683cf5407c9c3a/src/System.Management.Automation/engine/Credential.cs#L266) so [`SplitUserDomain` throws at the `IndexOf()`](https://github.com/PowerShell/PowerShell/blob/fd047a82506ac6a38c20052f45683cf5407c9c3a/src/System.Management.Automation/engine/Credential.cs#L369)

```
> $Error[0].Exception.InnerException.StackTrace

   at System.Management.Automation.PSCredential.SplitUserDomain(String input, String& user, String& domain)
   at System.Management.Automation.PSCredential.IsValidUserName(String input, String& user, String& domain)
   at System.Management.Automation.PSCredential.GetNetworkCredential()
   at CallSite.Target(Closure , CallSite , PSCredential )
```

iSazonov:
Fix #4696

### Problem descriptions

[System.Management.Automation.PSCredential](https://msdn.microsoft.com/en-us/library/system.management.automation.pscredential.empty) allow user name being empty but explicit cast [pscredential]::Empty.GetNetworkCredential() throws.

### Fix

GetNetworkCredential() returns null if PSCredential has null or empty user name

### DIFF
--- a/test/powershell/engine/Basic/Credential.Tests.ps1
+++ b/test/powershell/engine/Basic/Credential.Tests.ps1
@@ -1,6 +1,6 @@
 Describe "Credential tests" -Tags "CI" {
     It "Explicit cast for an empty credential returns null" {
          # We should explicitly check that the expression returns $null
-         [PSCredential]::Empty.GetNetworkCredential() -eq $null | Should Be $true
+         [PSCredential]::Empty.GetNetworkCredential() | Should Be $null
     }
 }
--- a/test/powershell/engine/Basic/Credential.Tests.ps1
+++ b/test/powershell/engine/Basic/Credential.Tests.ps1
@@ -1,0 +1,6 @@
+Describe "Credential tests" -Tags "CI" {
+    It "Explicit cast for an empty credential returns null" {
+         # We should explicitly check that the expression returns $null
+         [PSCredential]::Empty.GetNetworkCredential() -eq $null | Should Be $true
+    }
+}
--- a/src/System.Management.Automation/engine/Credential.cs
+++ b/src/System.Management.Automation/engine/Credential.cs
@@ -330,6 +330,12 @@ namespace System.Management.Automation
                                             out string user,
                                             out string domain)
         {
+            if (String.IsNullOrEmpty(input))
+            {
+                user = domain = null;
+                return false;
+            }
+
             SplitUserDomain(input, out user, out domain);
 
             if ((user == null) ||


test/powershell/engine/Basic/Credential.Tests.ps1
```diff
@@ -0,0 +1,6 @@
+Describe "Credential tests" -Tags "CI" {
+    It "Explicit cast for an empty credential returns null" {
+         # We should explicitly check that the expression returns $null
+         [PSCredential]::Empty.GetNetworkCredential() -eq $null | Should Be $true
```
PaulHigin: Shouldn't the test be: "[PSCredential]::Empty.GetNetworkCredential() | Should Be $Null"?

PaulHigin (changes requested):
I agree that this should return $null rather than throw a NR exception.  I have just one minor test comment.

TravisEz13:
@Indhukrishna FYI

test/powershell/engine/Basic/Credential.Tests.ps1
TravisEz13: A agree this would make any error, have more information to debug the issue.  Therefore making a rerun less likely to be needed.

test/powershell/engine/Basic/Credential.Tests.ps1
```diff
@@ -0,0 +1,6 @@
+Describe "Credential tests" -Tags "CI" {
+    It "Explicit cast for an empty credential returns null" {
```
TravisEz13: Should we add a case for an empty username?  How do you get the password from the object if `GetNetworkCredential` returns `$null`

TravisEz13 (dismissed):
Left a comment.  I think at least one other test case needs to be added.

test/powershell/engine/Basic/Credential.Tests.ps1
iSazonov: PSCredential constructor fails if username is empty. The only way to get an empty credentials is [PSCredential]::Empty. And if we can [PSCredential]::Empty any methods shouldn't throw for the empty credential - we're just fix the issue.

test/powershell/engine/Basic/Credential.Tests.ps1
iSazonov: Fixed.

PaulHigin (approved):
LGTM

test/powershell/engine/Basic/Credential.Tests.ps1
daxian-dbw: Actually, you are able to create empty PSCredential instances with a public constructor `PSCredential(psobject pso)`
```powershell
$psobj = [pscustomobject] @{}
$cred = [pscredential]::new($psobj)
```
The current implementation of `GetNetworkCredential()` doesn't allow returning a null value -- it throws if the user name is invalid. I think we should keep it that way and throw exception when user name is null or empty.

iSazonov: Why we allow the sample? It seems constructor should throw in the case.
On the other hand, [PSCredential]::Empty is a valid object - why any its method should throw?

daxian-dbw: > Why we allow the sample?

Do you mean the XML comment? I missed the comment, and it looks totally fine to return null then :)

iSazonov: I meant your sample:
```
$psobj = [pscustomobject] @{}
$cred = [pscredential]::new($psobj)
```
Why the constructor don't throw?  It seems the constructor should throw.

daxian-dbw: I don't know the original intention, but changing this behavior would be a breaking change for sure -- it makes something that works stop working. It feels insignificant to me, so I don't think it's worth to be fixed. But you can go both ways -- it's insignificant so a breaking change might not matter.

iSazonov: For the PR I think we can leave the constructor as is.

Please clarify what is your conclusion about the PR fix?

daxian-dbw: Based on the existing XML comments for `GetNetworkCredential()`, I withdraw my comment about making `GetNetworkCredential` throw. Returning null is totally fine.

daxian-dbw (approved):

PR #4699, opened 2017-08-29
Make TabCompletion case-insensitive for file names on Unix

### LINKED ISSUE #1273: Location completion should be case-insensitive
## Steps to reproduce

On Linux or OS X
- create a folder `Foo`
- type `cd foo` and hit `Tab`
## Expected behavior

Completion will correct `cd foo` into `cd ./Foo`
## Actual behavior

Nothing happens
## Environment data

``` powershell
> $PSVersionTable
PSVersion                      5.1.10032.0                                                                                                  
PSEdition                      Linux                                                                                                        
GitCommitId                    v0.5.0                                                                                                       
CLRVersion                                                                                                                                  
WSManStackVersion              1.0                                                                                                          
```
## Reasoning
- we already do that with `/` and `\` on all systems.
- fish shell does it and it's very convinient.

iSazonov:
## PR Summary

Attempt to fix #1273 

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
+++ b/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
@@ -4073,21 +4073,41 @@ namespace System.Management.Automation
                     executionContext.SessionState.Drive.GetAtScope(wordToComplete.Substring(0, 1), "global");
                 }
 #if UNIX
-                StringBuilder sb = new StringBuilder();
-
-                foreach (var ch in wordToComplete.ToCharArray())
+                // We use globbing to get completions.
+                // Globbing is based on system functions which is case-sensitive on Unix.
+                // To make globbing on Unix case-insensitive we transform every char from the input string:
+                // "alias:dir" -> "alias:[dD][iI][rR]"
+                if (context.GetOption("UnixCaseInsensitiveGlobbing", @default: true) == true)
                 {
-                    if (Char.IsLetter(ch))
-                    {
-                        sb.Append('[').Append(Char.ToLower(ch)).Append(Char.ToUpper(ch)).Append(']');
-                    }
-                    else
-                    {
-                        sb.Append(ch);
-                    }
-                }
+                    StringBuilder sb = new StringBuilder(wordToComplete.Length * 4);
 
-                wordToComplete = sb.ToString();
+                    //wordToComplete = WildcardPattern.Escape(wordToComplete, Utils.Separators.Brackets);
+
+                    // Copy a provider name "as is" without transformation.
+                    var providerPrefixIndex = wordToComplete.IndexOf(':') + 1;
+                    int i;
+                    for (i=0; i < providerPrefixIndex; i++)
+                    {
+                        sb.Append(wordToComplete[i]);
+                    }
+
+                    // Transform rest of input string after a provider prefix.
+                    for (var j=i; j < wordToComplete.Length; j++)
+                    {
+                        var lch = Char.ToLower(wordToComplete[j]);
+                        var uch = Char.ToUpper(wordToComplete[j]);
+                        if (lch != uch)
+                        {
+                            sb.Append('[').Append(lch).Append(uch).Append(']');
+                        }
+                        else
+                        {
+                            sb.Append(wordToComplete[j]);
+                        }
+                    }
+
+                    wordToComplete = sb.ToString();
+                }
 #endif
                 var powerShellExecutionHelper = context.Helper;
                 powerShellExecutionHelper
--- a/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
+++ b/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
@@ -4061,10 +4061,8 @@ namespace System.Management.Automation
                 var relativePaths = context.GetOption("RelativePaths", @default: defaultRelative);
                 var useLiteralPath = context.GetOption("LiteralPaths", @default: false);
 
-                if (useLiteralPath && LocationGlobber.StringContainsGlobCharacters(wordToComplete))
-                {
-                    wordToComplete = WildcardPattern.Escape(wordToComplete, Utils.Separators.StarOrQuestion);
-                }
+                // We should always escape '[' and ']' if present.
+                wordToComplete = WildcardPattern.Escape(wordToComplete, Utils.Separators.StarOrQuestion);
 
                 if (!defaultRelative && wordToComplete.Length >= 2 && wordToComplete[1] == ':' && char.IsLetter(wordToComplete[0]) && executionContext != null)
                 {
@@ -4072,6 +4070,7 @@ namespace System.Management.Automation
                     // can succeed.  This call will mount the drive if it wasn't already.
                     executionContext.SessionState.Drive.GetAtScope(wordToComplete.Substring(0, 1), "global");
                 }
+
 #if UNIX
                 // We use globbing to get completions.
                 // Globbing is based on system functions which is case-sensitive on Unix.
--- a/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
+++ b/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
@@ -4250,7 +4250,7 @@ namespace System.Management.Automation
                     }
 
                     // Sorting the results by the path
-                    var sortedPsobjs = psobjs.OrderBy(a => a, new ItemPathComparer());
+                    var sortedPsobjs = psobjs.OrderBy(a => a, new ItemPathComparer(wordToComplete));
 
                     foreach (PSObject psobj in sortedPsobjs)
                     {
@@ -6672,6 +6672,13 @@ namespace System.Management.Automation
 
         private class ItemPathComparer : IComparer<PSObject>
         {
+            private String _baseWord;
+
+            public ItemPathComparer(String baseWord)
+            {
+                _baseWord = baseWord + ".*";
+            }
+
             public int Compare(PSObject x, PSObject y)
             {
                 var xPathInfo = PSObject.Base(x) as PathInfo;
@@ -6701,7 +6708,21 @@ namespace System.Management.Automation
                 if (string.IsNullOrEmpty(xPath) || string.IsNullOrEmpty(yPath))
                     Diagnostics.Assert(false, "Base object of item PSObject should be either PathInfo or FileSystemInfo");
 
-                return String.Compare(xPath, yPath, StringComparison.CurrentCultureIgnoreCase);
+                var result = String.Compare(xPath, yPath, StringComparison.CurrentCultureIgnoreCase);
+#if UNIX
+                if (result == 0)
+                {
+                    if (Regex.IsMatch(xPath, _baseWord))
+                    {
+                        return -1;
+                    }
+                    else
+                    {
+                        return 1;
+                    }
+                }
+#endif
+                return result;
             }
         }
 
--- a/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
+++ b/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
@@ -4072,7 +4072,23 @@ namespace System.Management.Automation
                     // can succeed.  This call will mount the drive if it wasn't already.
                     executionContext.SessionState.Drive.GetAtScope(wordToComplete.Substring(0, 1), "global");
                 }
+#if UNIX
+                StringBuilder sb = new StringBuilder();
 
+                foreach (var ch in wordToComplete.ToCharArray())
+                {
+                    if (Char.IsLetter(ch))
+                    {
+                        sb.Append('[').Append(Char.ToLower(ch)).Append(Char.ToUpper(ch)).Append(']');
+                    }
+                    else
+                    {
+                        sb.Append(ch);
+                    }
+                }
+
+                wordToComplete = sb.ToString();
+#endif
                 var powerShellExecutionHelper = context.Helper;
                 powerShellExecutionHelper
                     .AddCommandWithPreferenceSetting("Microsoft.PowerShell.Management\\Resolve-Path")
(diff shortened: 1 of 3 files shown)


test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -256,6 +260,17 @@ Describe "TabCompletion" -Tags CI {
             Pop-Location
         }
 
+        It "TabCompletion should be case-insensitive for file names" {
```
vors: I'd also love to see tests for the order of the completed options. The first one should be the one with a correct casing, if accessible. Otherwise, the UX would be pretty bad in the multiple casing cases.

iSazonov: We use alphabetical [sort](https://github.com/iSazonov/PowerShell/blob/d5673cb661bed2d6406cd4e07757db7b927d5528/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs#L6735) with ignoring casing. I don't know how put an object with correct casing on first place. It looks you request creating very complex sorting.
Do anybody _really_ create multiple files/directories with different casing on Unix?

lzybkr: It is very common to have both `Makefile` and `makefile`.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -4122,7 +4122,23 @@ internal static IEnumerable<CompletionResult> CompleteFilename(CompletionContext
                     // can succeed.  This call will mount the drive if it wasn't already.
                     executionContext.SessionState.Drive.GetAtScope(wordToComplete.Substring(0, 1), "global");
                 }
+#if UNIX
+                StringBuilder sb = new StringBuilder();
+
+                foreach (var ch in wordToComplete.ToCharArray())
```
lzybkr: You don't need to call `ToCharArray` - that will unnecessarily create an array.

In C#, strings are enumerable, so you would just use `foreach (var ch in wordToComplete) ...`.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -4122,7 +4122,23 @@ internal static IEnumerable<CompletionResult> CompleteFilename(CompletionContext
                     // can succeed.  This call will mount the drive if it wasn't already.
                     executionContext.SessionState.Drive.GetAtScope(wordToComplete.Substring(0, 1), "global");
                 }
+#if UNIX
+                StringBuilder sb = new StringBuilder();
```
lzybkr: You can reasonably assume all characters are letters and pass `wordToComplete.Length * 4` when constructing the string builder to avoid growing and copying the buffer.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -4122,7 +4122,23 @@ internal static IEnumerable<CompletionResult> CompleteFilename(CompletionContext
...
+
+                foreach (var ch in wordToComplete.ToCharArray())
+                {
+                    if (Char.IsLetter(ch))
```
lzybkr: I'm no expert in Unicode - but I can imagine there are letters that have no upper/lowercase variants, so I would avoid the wildcard if ToUpper == ToLower.

lzybkr (changes requested):
I think we'll want this as an option (on by default).

Options are not really documented or formalized, but see uses of https://github.com/PowerShell/PowerShell/blob/dc399a121b3fa956887c066d2f22776540f7e676/src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs#L38 for examples.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
vors: German `ß` for example doesn't have upper in dotnet. (The correct capitalization is probably `SS`, but dotnet doesn't seem to do it, same as python 2.)
Also the The Turkish `I` is a classical example: if default locale is used, then lower-cased `I` will be different. We probably can skip these details as a first implementation.

Note that `[aa]` is a valid regex.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
iSazonov: Thabks! Fixed.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
iSazonov: Thanks! Fixed.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
iSazonov: First I made a prototype on PowerShell. :-)
Fixed.

iSazonov:
Options was added.
Also I had to fix an Issue with brackes in paths (tests failed for "Get-Help '.\My ``[Path``]'\").

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -4123,6 +4120,43 @@ internal static IEnumerable<CompletionResult> CompleteFilename(CompletionContext
...
+                // We use globbing to get completions.
+                // Globbing is based on system functions which is case-sensitive on Unix.
+                // To make globbing on Unix case-insensitive we transform every char from the input string:
+                // "alias:dir" -> "alias:[dD][iI][rR]"
+                if (context.GetOption("UnixCaseInsensitiveGlobbing", @default: true) == true)
+                {
+                    StringBuilder sb = new StringBuilder(wordToComplete.Length * 4);
+
+                    //wordToComplete = WildcardPattern.Escape(wordToComplete, Utils.Separators.Brackets);
+
```
lzybkr: Remove, but it might be worth a comment saying we assume `wordToComplete` is already a wildcard pattern and do not want to escape anything.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1459,6 +1459,7 @@ internal static class Separators
             internal static readonly char[] SpaceOrTab = new char[] { ' ', '\t' };
             internal static readonly char[] Newline = new char[] { '\n' };
             internal static readonly char[] CrLf = new char[] { '\r', '\n' };
+            internal static readonly char[] Brackets = new char[] { '[', ']' };
```
lzybkr: Not used now, right?

iSazonov: Removed.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
iSazonov: Removed.
Comment added above in line 4114. 

Previous I removed there `if useLiteralPath` -  Please clarify the `if` - it seems we should 
(1) always escape '[' and ']'
(2) escape `StarOrQuestion` for `useLiteralPath` but we did not this previous and don't  now.

iSazonov:
One MacOs test failed - I have no opportunity to debug 😕 @vors could you help?

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
iSazonov: It is only one extra <tab> press.
I have no idea how to make this weighted sort by simple way 😕

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
lzybkr: The word to complete is usually a wildcard pattern (unless the filename specifies `-LiteralPath`, but I don't think we do that today).

So I think it's probably best to not escape anything. It is probably uncommon to use tab completion and square brackets together, but I'm sure I've done it before, so it might feel broken.

This does mean you need to detect letters inside square brackets and ignore those.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
lzybkr: Not exactly simple, but sort insensitive, then compare each element to it's successor in a case sensitive manner, swapping to get case correct first.

iSazonov: Yes, my concern is that we will significantly increase the number of comparisons - is it ok for interactive scenarios?

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
iSazonov: So we leave as is now and add escape  StarOrQuestion for useLiteralPath when we'll need?

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
vors: > It is only one extra press.

It's not hard to save this extra press for many many people over years to come, is it?

> Yes, my concern is that we will significantly increase the number of comparisons

PowerShell does sooo much stuff behind the scene, it's peanuts compare to it.

vors (changes requested):
- [ ] prefer exact case match in the completion order.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
iSazonov: Sample 1:
wordToComplete = "aB"

Input:
abcd
aBcd
abde
aBde

Result:
aBcd
aBde
abcd
abde

Sample 2:
wordToComplete = "aB*D"

Input:
abcdef
aBCDef
abcdgh
aBcDgh

Result:
aBCDef
aBcDgh
abcdef
abcdgh

-- If we want the results the sort is not simple as seems - we must take into account wildcards.

iSazonov: Maybe:
1. Get a case-sensitive array and sort it
2. Get a case-insensitive array and sort it
3. Remove elements of (1) array remove from (2) array
4. Add elements of (2) array after end of (1) array

lzybkr: I would:

1. Case-insensitive array sort
2. For each element, compare case **insensitive** with the next element, if they are equal (should be rare), do a case **sensitive** wildcard match with the **original** word to complete to determine the ordering of these 2 elements, swapping if the first fails to match and the second does.

This should be O(n) with an upper bound of 2n, so it's cheaper than 2 sorts.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -4111,10 +4111,8 @@ internal static IEnumerable<CompletionResult> CompleteFilename(CompletionContext
                 var relativePaths = context.GetOption("RelativePaths", @default: defaultRelative);
                 var useLiteralPath = context.GetOption("LiteralPaths", @default: false);
 
-                if (useLiteralPath && LocationGlobber.StringContainsGlobCharacters(wordToComplete))
-                {
-                    wordToComplete = WildcardPattern.Escape(wordToComplete, Utils.Separators.StarOrQuestion);
-                }
+                // We should always escape '[' and ']' if present.
+                wordToComplete = WildcardPattern.Escape(wordToComplete, Utils.Separators.StarOrQuestion);
```
lzybkr: It seems like the code was correct before - only when we know we have -LiteralPath.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
lzybkr: Yeah, except I missed the code above that already does that, except you changed it.

---
In reply to: [136893930](https://github.com/PowerShell/PowerShell/pull/4699#discussion_r136893930) [](ancestors = 136893930)

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
iSazonov: Is my sample 2 right? If so your suggestion don't work for it. We get:
```
Result:
aBcd
abcd
aBde
abde

```

lzybkr: I'm not sure there is a right and wrong result here - only a least surprising, and I think my suggestion prefers case insensitive over the wildcard match which is less surprising, but still orders with some case sensitivity.

At any rate, real world experience will guide us best - try and keep it simple and handle the `Makefile`/`makefile` case nicely, and we'll see if it's sufficient.

iSazonov:
@vors Could you please build and check that we have case match.

vors:
@iSazonov order doesn't prefer the original casing from my experiments on macOS

iSazonov:
@vors @lzybkr 
Today I make some tests and now I suggest either close the PR or remove last commit (remove case-sensivire sorting for Unix) because I don't see a solution - we resolve `wordToComplete` to _absolute_ filesystem paths then we should case-sensitively compare the paths with `wordToComplete` (as regex pattern) but in common case the `wordToComplete` can be _relative_ "path" with wildcards or _absolute_ "path" with wildcards. Ex.: `/usr/*/*nn*/M*file` and `./*/*nn*/M*file`. Also we should convert (escape)  `wordToComplete` to regex pattern. I have no ideas how make that in short code and I have no Unix development environment to create large solition.

vors:
couple thoughts: order really matters only for a common case (no wildcards). If that is still hard, we can merge it without reordering and improve it later, if this behavior is guarded by an option flag.

iSazonov:
I wonder can we use `Resolve-Path -CaseInsensitive`. It seems the new ` -CaseInsensitive` could be useful.

PR #4709, opened 2017-08-30
Push locals of automatic variables to 'DottedScopes' when dotting script cmdlets

### LINKED ISSUE #4688: $PSScriptRoot is not populated when used in a default value for a non-mandatory script parameter
Steps to reproduce
------------------
Run this script (via PowerShell -File ....)

```powershell
param(
    [Parameter(Mandatory = $false)]
    [string] $Foo = 'hi\{0}\hi' -f $PSScriptRoot
)

Write-Host $Foo

Function Bar {
    param(
        [Parameter(Mandatory = $false)]
        [string] $Bar = $PSScriptRoot
    )
    Write-Host $Bar
}

Function Baz {
    [CmdletBinding()]
    param(
        [Parameter(Mandatory = $false)]
        [string] $Baz = $PSScriptRoot
    )
    Write-Host $Baz
}

Bar
Baz
```

Expected behavior
-----------------
```
hi\C:\tmp\hi
C:\tmp
C:\tmp
```

Actual behavior
---------------
```
hi\\hi
C:\tmp
C:\tmp
```

Environment data
----------------
```
PSVersion                      5.1.16353.1000
BuildVersion                   10.0.16353.1000
```

_[edited by @daxian-dbw to put the repro in code blocks]_

daxian-dbw:
Fix #4688

When dotting a script cmdlet, the locals of automatic variables from the `PSScriptCmdlet` is not set up in the current scope before parameter binding. The fix is to push the locals in `CommandProcessor.OnSetCurrentScope` and pop them in `CommandProcessor.OnRestorePreviousScope`, which will be called from `SetCurrentScopeToExecutionScope` and `RestorePreviousScope` respectively.

Summary of changes:
1. When a new local scope is used, currently we set the locals for `CommandProcessor` right before parameter binding (in `BindCommandLineParametersNoValidation`); we set the locals for `ScriptCommandProcessor` in `Prepare`. I moved both to the constructor, right after the new scope is created so that the code is more consistent.

2. In `CmdletParameterBinderController.cs`, we set up the `PSBoundParameters` and `MyInvocation` variables in `HandleCommandLineDynamicParameters` again, which I think is unnecessary because this method is only called from `BindCommandLineParametersNoValidation`, where the setup is done for the first time.

3. Currently, the locals will be set for dotted script cmdlet in `EnterScope()` and `ExitScope()`. Now, that logic is moved to `OnSetCurrentScope` and `OnRestorePreviousScope` of `CommandProcessor`. This not only makes sure that locals are set before parameter binding, but also is consistent with the `ScriptCommandProcessor`.

### DIFF
--- a/src/System.Management.Automation/engine/CmdletParameterBinderController.cs
+++ b/src/System.Management.Automation/engine/CmdletParameterBinderController.cs
@@ -215,9 +215,7 @@ namespace System.Management.Automation
             var psCompiledScriptCmdlet = this.Command as PSScriptCmdlet;
             if (psCompiledScriptCmdlet != null)
             {
-                psCompiledScriptCmdlet.PrepareForBinding(
-                    ((ScriptParameterBinder)this.DefaultParameterBinder).LocalScope,
-                    this.CommandLineParameters);
+                psCompiledScriptCmdlet.PrepareForBinding(this.CommandLineParameters);
             }
 
             // Add the passed in arguments to the unboundArguments collection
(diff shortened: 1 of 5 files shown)


vors (approved):

daxian-dbw:
@vors Thanks!
@lzybkr can you please take a look when you have time?

ViktorDronov (approved):

PR #4718, opened 2017-08-31
add absolute url for issue template and pull request template

chucklu:
Correct the url for FAQ and known issues in issue template
Add url for contributing.md in pull request template

Both of them work fine in preview model of github.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -4,8 +4,8 @@ If it is a bug report:
 - make sure you are able to repro it on the latest released version. 
 You can install the latest version from https://github.com/PowerShell/PowerShell/releases
 - Search the existing issues.
-- Refer to the [FAQ](../docs/FAQ.md).
-- Refer to the [known issues](../docs/KNOWNISSUES.md).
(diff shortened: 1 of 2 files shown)


markekraus:
As far as I can tell, these links are never clickable or even visible as hyperlinks. I wonder if it would make more sense to use absolute URLs so the can at least be copy/pasted from the PR/Issue TextBox.

daxian-dbw:
I agree with @markekraus that we probably should use absolute URLs in those places.

chucklu:
@markekraus  When you remove the comment mark in markdown,then preview it,you can click the url.

The template with comment mark is weird,since it was written in markdown. But it‘s invisible in preview model.
I suggest you check the issue template of git for windows.Here is the [link](https://github.com/git-for-windows/git/blob/master/.github/ISSUE_TEMPLATE.md).

chucklu:
@markekraus  Hi,I have add absolute url for it,and force push to my fork repository.
Now,this pull request fetches the commit with absolute url.
You can check it,thanks.

markekraus:
Looks good to me! Thanks!

daxian-dbw (approved):
LGTM

PR #4721, opened 2017-08-31
Enable use 'Singleline,Multiline' option in split operator

### LINKED ISSUE #4712: -split operator doesn't let you use options SingleLine and MultiLine at the same time (and it should)
Steps to reproduce
---------------------

```powershell
PS C:\> '' -split '', 0, 'singleline,multiline'
The combination of options with the -split operator is not valid.
    + CategoryInfo          : ParserError: (:) [], ParseException
    + FullyQualifiedErrorId : InvalidSplitOptionCombination
```

But it is not an invalid combination, they do different things, and it is common and useful to use them together.

The PowerShell spec v 3.0, section 7.8.4.5 "Binary split operator" says:

> [an options string] must not contain both Multiline and Singleline.
> Multiline: This mode recognizes the start and end of lines and strings. The default mode is Singleline.

Singleline is not the default of Multiline, they are not two sides of the same option. Multiline changes how the regex characters `^` and `$` work with newlines in a string, and Singleline changes how the regex character `.` works with newlines in a string. It is valid to use both options to change both `^$` and `.` at the same time.

e.g.

```powershell
$text = @'
a12a
b34b
c56c
d78d
'@

# The following split fails, because '^' only matches the start of the entire string, 
# and 'b' is not at the start of the string:
$text -split '^b'

# Enable 'Multiline' mode and '^' now matches the start of each line in the string
# and this now succeeds:
$text -split '^b', $numSplits, 'multiline'

# The following split fails, because '.+' only matches to the end of the line, and never gets to 'c'
$text -split 'b.+c'

# Enable 'Singleline' mode and '.' now matches over newlines
# and this now succeeds:
$text -split 'b.+c', $numSplits, 'singleline'

# To combine them and match from "the b at the start of a line,
# over newlines and up to the last c" needs both singleline and multiline options.
# This fails and throws an exception:
$text -split '^b.+c', $numSplits, 'singleline, multiline'

# but you can turn both on at the same time using regex syntax:
# so it can and does work, if you work around PowerShell:
$text -split '(?ms)^b.+c'
# splits on the text b34b`nc56c

# and you can also do it with the options string passed to the [regex] accelerator:
# This works, too:
PS D:\> [regex]::Split($text, '^b.+c', 'singleline,multiline')
a12a


d78d
```

The documentation in `Get-Help about_split` is also incorrect:

>          -- Multiline: Multiline mode recognizes the start and end of lines
>             and strings. Valid only with RegexMatch. Singleline is the default.
>
>          -- Singleline: Singleline mode recognizes only the start and end of
>             strings. Valid only with RegexMatch. Singleline is the default.

 - [Link to source code of parserutils.cs where the error is thrown](https://github.com/PowerShell/PowerShell/blob/7a55bf98b2370ab4817ab2533cb67673053ee446/src/System.Management.Automation/engine/lang/parserutils.cs#L726)

 - [Link to source code of parserutils.cs where the comment in the SplitOptionsEnum has the same inaccurate suggestion that one is the default setting of the other](https://github.com/PowerShell/PowerShell/blob/7a55bf98b2370ab4817ab2533cb67673053ee446/src/System.Management.Automation/engine/lang/parserutils.cs#L295)

Expected behavior
-----------------

Specifying an options string `'singleline,multiline'` is valid, and enables both Singleline and Multiline mode.

The documentation is accurate.

Actual behavior
---------------

It throws an exception and misleadingly says they are invalid together.

The documentation is inaccurate in three senses - implying there is one option being set instead of two, implying Singleline is the opposite of Multiline, and stating Singleline "is the default" when Singleline itself is off by default.

Environment data
----------------

```powershell
> $PSVersionTable
GitCommitId                    v6.0.0-beta.6
OS                             Linux 2.6.32-042stab120.6 #1 SMP Thu Oct 27 16:59:03 MSK 2016
```

iSazonov:
Fix #4712 

RegEx Singleline and Multiline options are not mutually exclusive and can be used at the same time.

Tests for -split operator is added to get real code coverage and exclude regressions.

### DIFF
--- a/test/powershell/Language/Operators/SplitOperator.Tests.ps1
+++ b/test/powershell/Language/Operators/SplitOperator.Tests.ps1
@@ -107,12 +107,15 @@ Describe "Split Operator" -Tags CI {
 
     Context "Binary split operator options" {
         BeforeAll {
-            $testText = "a12a`nb34b`nc56c`nd78d"
-            # Add '%' - now second line doesn't start with 'b'.
-            $testText2 = "a12a`n%b34b`nc56c`nd78d"
+            # Add '%' in testText2 in order to second line doesn't start with 'b'.
+            $testCases = @(
+                @{ Name = '`n';   testText = "a12a`nb34b`nc56c`nd78d";       testText2 = "a12a`n%b34b`nc56c`nd78d";       newLine = "`n" }
+                @{ Name = '`r`n'; testText = "a12a`r`nb34b`r`nc56c`r`nd78d"; testText2 = "a12a`r`n%b34b`r`nc56c`r`nd78d"; newLine = "`r`n" }
+                )
         }
 
-        It "Binary split operator has no Singleline and no Multiline by default" {
+        It "Binary split operator has no Singleline and no Multiline by default (new line = '<Name>')" -TestCases $testCases {
+            param($testText, $testText2, $newLine)
             # Multiline isn't default
             $res = $testText -split '^b'
             $res.count | Should Be 1
@@ -122,40 +125,43 @@ Describe "Split Operator" -Tags CI {
             $res.count | Should Be 1
         }
 
-        It "Binary split operator works with Singleline" {
+        It "Binary split operator works with Singleline (new line = '<Name>')" -TestCases $testCases {
+            param($testText, $testText2, $newLine)
             $res = $testText -split 'b.+c', 0, 'Singleline'
             $res.count | Should Be 2
-            $res[0] | Should Be "a12a`n"
-            $res[1] | Should Be "`nd78d"
+            $res[0] | Should Be "a12a$($newLine)"
+            $res[1] | Should Be "$($newLine)d78d"
 
             $res = $testText2 -split 'b.+c', 0, 'Singleline'
             $res.count | Should Be 2
-            $res[0] | Should Be "a12a`n%"
-            $res[1] | Should Be "`nd78d"
+            $res[0] | Should Be "a12a$($newLine)%"
+            $res[1] | Should Be "$($newLine)d78d"
         }
 
-        It "Binary split operator works with Multiline" {
+        It "Binary split operator works with Multiline (new line = '<Name>')" -TestCases $testCases {
+            param($testText, $testText2, $newLine)
             $res = $testText -split '^b', 0, 'Multiline'
             $res.count | Should Be 2
-            $res[0] | Should Be "a12a`n"
-            $res[1] | Should Be "34b`nc56c`nd78d"
+            $res[0] | Should Be "a12a$($newLine)"
+            $res[1] | Should Be "34b$($newLine)c56c$($newLine)d78d"
         }
 
-        It "Binary split operator works with Singleline,Multiline" {
+        It "Binary split operator works with Singleline,Multiline (new line = '<Name>')" -TestCases $testCases {
+            param($testText, $testText2, $newLine)
             $res = $testText -split 'b.+c', 0, 'Singleline,Multiline'
             $res.count | Should Be 2
-            $res[0] | Should Be "a12a`n"
-            $res[1] | Should Be "`nd78d"
+            $res[0] | Should Be "a12a$($newLine)"
+            $res[1] | Should Be "$($newLine)d78d"
 
             $res = $testText2 -split 'b.+c', 0, 'Singleline,Multiline'
             $res.count | Should Be 2
-            $res[0] | Should Be "a12a`n%"
-            $res[1] | Should Be "`nd78d"
+            $res[0] | Should Be "a12a$($newLine)%"
+            $res[1] | Should Be "$($newLine)d78d"
 
             $res = $testText -split '^b.+c', 0, 'Singleline,Multiline'
             $res.count | Should Be 2
-            $res[0] | Should Be "a12a`n"
-            $res[1] | Should Be "`nd78d"
+            $res[0] | Should Be "a12a$($newLine)"
+            $res[1] | Should Be "$($newLine)d78d"
 
             $res = $testText2 -split '^b.+c', 0, 'Singleline,Multiline'
             $res.count | Should Be 1
--- a/test/powershell/Language/Operators/SplitOperator.Tests.ps1
+++ b/test/powershell/Language/Operators/SplitOperator.Tests.ps1
@@ -108,8 +108,8 @@ Describe "Split Operator" -Tags CI {
     Context "Binary split operator options" {
         BeforeAll {
             $testText = "a12a`nb34b`nc56c`nd78d"
-            # Add '#' - now second line do not start with 'b'.
-            $testText2 = "a12a`n#b34b`nc56c`nd78d"
+            # Add '%' - now second line doesn't start with 'b'.
+            $testText2 = "a12a`n%b34b`nc56c`nd78d"
         }
 
         It "Binary split operator has no Singleline and no Multiline by default" {
@@ -130,7 +130,7 @@ Describe "Split Operator" -Tags CI {
 
             $res = $testText2 -split 'b.+c', 0, 'Singleline'
             $res.count | Should Be 2
-            $res[0] | Should Be "a12a`n#"
+            $res[0] | Should Be "a12a`n%"
             $res[1] | Should Be "`nd78d"
         }
 
@@ -149,7 +149,7 @@ Describe "Split Operator" -Tags CI {
 
             $res = $testText2 -split 'b.+c', 0, 'Singleline,Multiline'
             $res.count | Should Be 2
-            $res[0] | Should Be "a12a`n#"
+            $res[0] | Should Be "a12a`n%"
             $res[1] | Should Be "`nd78d"
 
             $res = $testText -split '^b.+c', 0, 'Singleline,Multiline'
--- a/test/powershell/Language/Operators/SplitOperator.Tests.ps1
+++ b/test/powershell/Language/Operators/SplitOperator.Tests.ps1
@@ -1,0 +1,236 @@
+Describe "Split Operator" -Tags CI {
+    Context "Binary split operator" {
+        It "Binary split operator can split array of value" {
+            $res = "a b", "c d" -split " "
+            $res.count | Should Be 4
+            $res[0] | Should Be "a"
+            $res[1] | Should Be "b"
+            $res[2] | Should Be "c"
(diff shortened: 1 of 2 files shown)


test/powershell/Language/Operators/SplitOperator.Tests.ps1
```diff
@@ -0,0 +1,236 @@
...
+            $res[1] | Should Be "::"
+            $res[2] | Should Be "::d"
+        }
+
+    }
+
+    Context "Binary split operator options" {
+        BeforeAll {
+            $testText = "a12a`nb34b`nc56c`nd78d"
+            # Add '#' - now second line do not start with 'b'.
```
vors: let's use some other char instead of `#` - it's easy to confuse with comments, but comments are irrelevant here. 

also nit: `doesn't start`

test/powershell/Language/Operators/SplitOperator.Tests.ps1
```diff
@@ -0,0 +1,236 @@
...
+        It "Unary split operator has higher precedence than a comma" {
+            $res = -split "a b", "c d"
+            $res.count | Should Be 2
+            $res[0][0] | Should Be "a"
+            $res[0][1] | Should Be "b"
+            $res[1] | Should Be "c d"
+        }
+
+        It "Unary split operator can split array of values" {
+            $res = -split ("a b", "c d")
```
vors: wow didn't know you can do that!

vors (approved):

test/powershell/Language/Operators/SplitOperator.Tests.ps1
iSazonov: Fixed.

test/powershell/Language/Operators/SplitOperator.Tests.ps1
```diff
@@ -0,0 +1,236 @@
...
+            $res[1] | Should Be "b c d"
+
+            $res = "a b c d" -split " ", 0
+            $res.count | Should Be 4
+            $res[0] | Should Be "a"
+            $res[1] | Should Be "b"
+            $res[2] | Should Be "c"
+            $res[3] | Should Be "d"
+
+            $res = "a b c d" -split " ", -1
```
lzybkr: Is this a valid scenario, or should PowerShell report an error when the count is less than 1?

test/powershell/Language/Operators/SplitOperator.Tests.ps1
```diff
@@ -0,0 +1,236 @@
...
+            $res[0] | Should Be "a::"
+            $res[1] | Should Be "::"
+            $res[2] | Should Be "::d"
+        }
+
+    }
+
+    Context "Binary split operator options" {
+        BeforeAll {
+            $testText = "a12a`nb34b`nc56c`nd78d"
```
lzybkr: Multi-line tests should probably all test the three linefeed variants: `` `n ``, `` `r ``, and `` `r`n ``.

lzybkr (approved):
Approved, but it would be nice to see more tests with the various valid variants of line feeds.

test/powershell/Language/Operators/SplitOperator.Tests.ps1
iSazonov: Here we is full based on RegEx implementation in our code. We don't parse the string, we only pass it to RegEx as is.  Do we really want to test CoreFX implementation?

test/powershell/Language/Operators/SplitOperator.Tests.ps1
iSazonov: I get this from our [documentation](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_split?view=powershell-5.1) - see <Max-substrings>.
So it will be a breaking change if we add a exception here. I don't see the need for that.

@mklement0 Could you please comment too?

test/powershell/Language/Operators/SplitOperator.Tests.ps1
lzybkr: True today and probably true forever, but if somebody decided they could be smarter than regex (maybe in limited scenarios), it's useful (if it's easy) to cover those additional scenarios.

iSazonov: I see your point. I added tests for '`n' and '`r`n'. SingleLine option catch only '`n` so I believe no tests needed for '`r'.

test/powershell/Language/Operators/SplitOperator.Tests.ps1
mklement0: @iSazonov: Yes, the documentation states "A value of 0 and negative values return all the substrings". 

Supporting `0` definitely makes sense, because in order to be able to specify _options_ you must also specify a `<max-substrings>` value for syntactical reasons, so you need a way to signal the default of "return all".

Negative values serve no real purpose - and could lead people to mistakenly assume that they have special meaning - but given that they're documented and that we could break existing code, I don't think a change is warranted.

test/powershell/Language/Operators/SplitOperator.Tests.ps1
mklement0: @vors: Regarding the _array_-as-operand aspect (not the unary-operator aspect):

Since the only operand unary `-split` accepts is of type `[string]`, the array is simply stringified.

In other words, assuming the default `$OFS` value:

       -split ("a b", "c d")

is the same as:

       -split "a b c d"

What puzzles me, however is that the implicit array-to-string conversion in this case does NOT respect `$OFS`:

```powershell
       > & { $OFS='@'; $a = 'a b', 'c d'; (-split $a).Count; (-split "$a").Count }
       4 # !! $OFS not respected.
       3  # OK: the equivalent of (-split 'a b@c d').Count
```

Is this by design?

test/powershell/Language/Operators/SplitOperator.Tests.ps1
iSazonov: @mklement0 Thanks for good comment!

test/powershell/Language/Operators/SplitOperator.Tests.ps1
mklement0: @vors: 
Turns out I didn't know either, mostly because [the documentation](https://github.com/PowerShell/PowerShell-Docs/blob/staging/reference/6/Microsoft.PowerShell.Core/About/about_Split.md):

*  in part suggests that only string _scalars_ are supported (`<String>` in the syntax diagram) 

* where it does mention arrays (multiple input strings), mistakenly claims that only the _binary_ form supports them ("To split more than one string, use the binary split operator ").

I've created a [doc issue](https://github.com/PowerShell/PowerShell-Docs/issues/1640) to address that.

(I originally thought that when you pass an array it simply gets stringified based on `$OFS`, but that's not the case.)

test/powershell/Language/Operators/SplitOperator.Tests.ps1
lzybkr: There is an argument in favor of a breaking change - is a negative count a likely bug? For example, if I compute the count and swapped my operands, this feels like a bug PowerShell should catch.

I do think it's an unlikely breaking change, mostly because the count it most commonly a constant.

mklement0: > if I compute the count and swapped my operands

I don't understand. You mean accidentally swapping the RHS operands?

Regardless, if restricting the "return all" meaning to `0` only is not considered too risky, I welcome it.

It would leave the door open for negative values serving a distinct (yet to be implemented) purpose, such as:

```powershell
> 'a b c d' -split ' ', -1  # only split 1 token _from the end_
a b c
d
```

lzybkr: I meant something like:
```powershell
$cnt = $b - $a # Should have been $a - $b
'a b c d' -split ' ', $cnt
```

mklement0: Got it, thanks. Preventing negative values categorically is certainly at odds with what I just proposed, but I'd still consider it an improvement.

iSazonov: It is interesting improvement:
```powershell
PS C:\temp> 'a b c d' -split ' ', 2
a
b c d
PS C:\temp> 'a b c d' -split ' ', 1
a b c d
PS C:\temp> 'a b c d' -split ' ', 0
a
b
c
d
PS C:\temp> 'a b c d' -split ' ', -1
a b c d
PS C:\temp> 'a b c d' -split ' ', -2
a b c
d
```

mklement0: @iSazonov: Yes, though your `-2` behavior is what I suggest for `-1` - analogous to index `[-1]` retrieving the last element of a collection.
Not sure if providing a way to not split at all is necessary.

```powershell
> 'a b c d' -split ' ', -1  # last 1
a b c
d

> 'a b c d' -split ' ', -2  # last 2
a b
c
d
```

Also, it's probably worth always returning a first item that represents the remaining, unsplit part - even if that part is empty:

```powershell
> 'a b' -split ' ', -2  # last 2, with nothing remaining
   # empty string  to signal that there's no remaining part
a
b
```

That enables the following idiom, irrespective of whether there's a remaining part or not.

```powershell
> $rest, $tokens = 'a b' -split ' ', -2
```

iSazonov: @mklement0  I like the "breaking change". If @lzybkr approve we should open Issue Enhancement.

lzybkr:
I approve of the breaking change, but we'll need an ack from the @PowerShell/powershell-committee before proceeding.

SteveL-MSFT:
@PowerShell/powershell-committee can see how this can be useful and the current documented behavior is an edge case that isn't likely used so accepting this breaking change is approved

lzybkr:
The suggested behavior for negative count should be a different PR.

iSazonov:
@mklement0 Feel free to open Issue for the negative count enhancement.

mklement0:
@iSazonov: Done - see #4765.

iSazonov:
@lzybkr Is the PR ready to merge?

iSazonov:
@lzybkr @mklement0 Thanks for great discussion!

daxian-dbw:
@iSazonov Great test additions! Thanks!

Issue #3346, opened 2017-03-16
Add DevOps relevant verbs to approved verb list


HemantMahawar:
https://windowsserver.uservoice.com/forums/301869-powershell/suggestions/12303924-add-a-compile-verb

NEW Verbs:
- Build

joeyaiello:
I think we also talked about `Deploy`, right?

SteveL-MSFT:
@joeyaiello @HemantMahawar These would be under Lifecycle, correct?  Give me a description of Build and Deploy

dragonwolf83:
Deploy seems to already have a comparable Lifecycle verb, Install. Based on the existing description below, I would think Install would be the more appropriate verb.

> Install (is)
Places a resource in a location, and optionally initializes it. This verb is paired with Uninstall.
For this action, do not a use verb such as Setup.

If you have a Deploy, you also need a Rollback verb. Install and Uninstall both already cover those concepts quite well.

On the other hand, you could change the Install language to say it is intended for **local** resources while Deploy is used for remote resources. 

For Build, I think in software dev it makes more sense than any other verb. The guidance for New though should be updated to remove Build from the list of "do not use" verbs.

SteveL-MSFT:
@joeyaiello will write up descriptions of the verbs: build and deploy

joeyaiello:
Here's a first pass, can I get a review from @SteveL-MSFT, @HemantMahawar, and @BrucePay ? 

## Build

Creates an artifact (usually a binary or ) out of some set of input files (usually source code or declarative documents). For PowerShell's purposes this may include the "linking" step of a compilation process. 
> Comments: do not use `Compile`

## Deploy

Sends an application, website, or solution to a remote location . This differs from `Install` because it sends this application to a remote target (or targets) to be installed whereas `Install` targets a local instance.

RichardSiddaway:
I think you also need Rollback to complete the set - Similar argument that Uninstall is local while Rollback is remote - also rollback has the connotation of setting the state back to the point before the deployment.
Deploy/Rollback then tend to function almost as a transaction

ArieHein:
Although PS likes verbs to have a "positive" and "negative" effects ,I dont think RollBack is necessary the way forwards, in quite a few DevOps pipelines, Rollback is simple a Deploy -Version "last working one", which you can also say that Uninstall is basically Install -U or Install -Action U so while I can understand Install/Uninstall being the English language, RollBack isn't necessary the opposite of Deploy.
[...]

PR #4725, opened 2017-08-31
added new approved Build and Deploy verbs

SteveL-MSFT:
Fix https://github.com/PowerShell/PowerShell/issues/3346

removed suggestions that these were not valid

organized verbs to be in alphabetical order to make it easier to compare against
(unfortunately, the diff viewer on GitHub isn't so smart about lines that were moved, easier to review the individual commits)
https://msdn.microsoft.com/en-us/library/ms714428(VS.85).aspx

Created doc bug to update above link https://github.com/PowerShell/PowerShell-Docs/issues/1630

### DIFF
--- a/src/System.Management.Automation/utils/Verbs.cs
+++ b/src/System.Management.Automation/utils/Verbs.cs
@@ -348,6 +348,11 @@ namespace System.Management.Automation
         public const string Assert = "Assert";
 
         /// <summary>
+        /// Creates an artifact (usually a binary or ) out of some set of input files (usually source code or declarative documents)
+        /// </summary>
+        public const string Build = "Build";
+
+        /// <summary>
         /// Finalize an interruptable activity. Makes pending changes permanent.
         /// </summary>
         public const string Complete = "Complete";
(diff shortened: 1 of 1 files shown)


src/System.Management.Automation/utils/Verbs.cs
```diff
@@ -348,6 +348,11 @@ public static class VerbsLifecycle
         public const string Assert = "Assert";
 
         /// <summary>
+        /// Creates an artifact (usually a binary or ) out of some set of input files (usually source code or declarative documents)
```
markekraus: `(usually a binary or )` Missing info after 'or'.

markekraus (commented):
Missing some info

src/System.Management.Automation/utils/Verbs.cs
SteveL-MSFT: will fix

SteveL-MSFT:
@markekraus can you `Approve` if you're ok with the changes?  Thanks!

markekraus (approved):

daxian-dbw (approved):

PR #4728, opened 2017-08-31
get tests to display the same number of tests regardless of platform

JamesWTruher:
We should be skipping tests on platforms where they not applicable rather than not reporting at all. These are changes needed to have Get-ComputerInfo and the Certificate provider tests report the same number of tests regardless of platform

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Security/CertificateProvider.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Security/CertificateProvider.Tests.ps1
@@ -2,6 +2,7 @@
 if($IsWindows)
 {
     Import-Module (Join-Path -Path $PSScriptRoot 'certificateCommon.psm1') -Force
+}
 
     $currentUserMyLocations = @(
         @{path = 'Cert:\CurrentUser\my'}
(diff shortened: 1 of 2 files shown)


JamesWTruher:
@TravisEz13 @SteveL-MSFT comments?

test/powershell/Modules/Microsoft.PowerShell.Security/CertificateProvider.Tests.ps1
```diff
@@ -2,6 +2,7 @@
 if($IsWindows)
 {
     Import-Module (Join-Path -Path $PSScriptRoot 'certificateCommon.psm1') -Force
+}
```
SteveL-MSFT: I don't understand why this brace needed to be moved.  With the original code, the two variables below would be $null and the foreach below them doesn't execute, but the tests get skipped anyways on non-Windows?

SteveL-MSFT (approved):
LGTM, just one question on a change whether it's necessary

test/powershell/Modules/Microsoft.PowerShell.Security/CertificateProvider.Tests.ps1
JamesWTruher: the brace needed to be moved so the right number of tests could be created. Since these are used as test cases, it represents 7 tests which weren't being reported earlier. You're right about the skipping, but it was the wrong result count (1 vs 7), since each test in a set of test cases is reported as an individual test.

---
In reply to: [136633885](https://github.com/PowerShell/PowerShell/pull/4728#discussion_r136633885) [](ancestors = 136633885)

PR #4729, opened 2017-08-31
Set Redist test pending to unblock daily build failure

### LINKED ISSUE #4720: WindowsInstaller.Tests.ps1 failed on windows
Similar to issue #4663

Only this time it repros on windows and affects daily appveyor.

TEST FAILURES
Description: Should have download link about Universal C runtime that is reachable
Name:        Windows Installer.Universal C Runtime Download Link.Should have download link about Universal C runtime that is reachable
message:
Error 404 - Not FoundWarning: This site requires the use of scripts, which your browser does not currently allow.See how to enable scripts

Should we set the tests to be pending until we can stabilize the result?

chunqingchen:
This is to resolve #4720 in case test owner need more time to stabilize the test under windows.

If we get a better solution, feel free to close this.


daxian-dbw:
@TravisEz13 could this be related to #4701?

TravisEz13 (approved):

TravisEz13:
@daxian-dbw  Could be... writing that code will be much more difficult if we truly want to have one package.

PR #4731, opened 2017-08-31
Report the same number of basic parsing tests when run in Travis via cron

JamesWTruher:
update the logic when tests are skipped on travis for our daily build so it better mirrors what we report when we actually run the tests. This means that when we compare runs across platforms we report on the same tests (even though they're skipped)


daxian-dbw (approved):
LGTM

JamesWTruher:
thanks!

PR #4735, opened 2017-09-01
Add -WhatIf switch to Start-Process cmdlet

### LINKED ISSUE #4702: Add -WhatIf support to Start-Process.
Had instance were an OpenSSL process was failing with little help from OpenSSL as to why. Having an output of the exact command being executed would helped me solve my problem alot sooner.

sarithsutha:
Closes #4702 

Add -WhatIf switch to Start-Process cmdlet

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
@@ -2051,8 +2051,7 @@ namespace Microsoft.PowerShell.Commands
                 startInfo.WindowStyle = _windowstyle;
             }
 
-            string targetMessage =  StringUtil.Format(ProcessResources.ProcessStartInfo, startInfo.FileName, startInfo.Arguments.Trim());
-            
+            string targetMessage = StringUtil.Format(ProcessResources.ProcessStartInfo, startInfo.FileName, startInfo.Arguments.Trim());
             if (!ShouldProcess(targetMessage)) { return; }
 
             //Starts the Process
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
@@ -2051,7 +2051,7 @@ namespace Microsoft.PowerShell.Commands
                 startInfo.WindowStyle = _windowstyle;
             }
 
-            string targetMessage = StringUtil.Format(ProcessResources.ProcessStartInfo, startInfo.FileName, startInfo.Arguments.Trim());
+            string targetMessage = StringUtil.Format(ProcessResources.StartProcessTarget, startInfo.FileName, startInfo.Arguments.Trim());
             if (!ShouldProcess(targetMessage)) { return; }
 
             //Starts the Process
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
@@ -1658,7 +1658,7 @@ namespace Microsoft.PowerShell.Commands
     /// <summary>
     /// This class implements the Start-process command
     /// </summary>
-    [Cmdlet(VerbsLifecycle.Start, "Process", DefaultParameterSetName = "Default", HelpUri = "https://go.microsoft.com/fwlink/?LinkID=135261")]
+    [Cmdlet(VerbsLifecycle.Start, "Process", DefaultParameterSetName = "Default", SupportsShouldProcess = true, HelpUri = "https://go.microsoft.com/fwlink/?LinkID=135261")]
     [OutputType(typeof(Process))]
     public sealed class StartProcessCommand : PSCmdlet, IDisposable
     {
@@ -2050,6 +2050,10 @@ namespace Microsoft.PowerShell.Commands
                 //WindowStyle
                 startInfo.WindowStyle = _windowstyle;
             }
+
+            string targetMessage =  StringUtil.Format(ProcessResources.ProcessStartInfo, startInfo.FileName, startInfo.Arguments.Trim());
+            
+            if (!ShouldProcess(targetMessage)) { return; }
 
             //Starts the Process
             Process process = Start(startInfo);
--- a/src/Microsoft.PowerShell.Commands.Management/resources/ProcessResources.resx
+++ b/src/Microsoft.PowerShell.Commands.Management/resources/ProcessResources.resx
@@ -132,6 +132,9 @@
   <data name="ProcessNameForConfirmation" xml:space="preserve">
     <value>{0} ({1})</value>
   </data>
+  <data name="ProcessStartInfo" xml:space="preserve">
+    <value>{0} {1}</value>
+  </data>
   <data name="CouldNotEnumerateModules" xml:space="preserve">
     <value>Cannot enumerate the modules of the "{0}" process.</value>
   </data>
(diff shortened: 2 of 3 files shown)


test/powershell/Modules/Microsoft.PowerShell.Management/Start-Process.Tests.ps1
```diff
@@ -104,6 +104,10 @@ Describe "Start-Process" -Tags @("Feature") {
         $process.Name | Should Be "notepad"
         $process | Stop-Process
     }
+
+    It "Should be able to use the whatif switch without error" {
+        { Start-Process $pingCommand -ArgumentList $pingParam -WhatIf } | Should Not Throw
+    }
```
markekraus: Perhaps add a test that would normally perform an action and ensure that `-WhatIf` prevented the action? 
For example, have a `Start-Process` command to write a string to a file and then test to ensure the file was not written to with `-WhatIf` is supplied.

sarithsutha:
Thanks for your suggestion @markekraus . I've added that test.

test/powershell/Modules/Microsoft.PowerShell.Management/Start-Process.Tests.ps1
bergmeister: Maybe also a test that using the `PassThru` switch, no object gets returned (which is also a good additional test that the command was not executed)?
```
It "returns null when using -WhatIf switch with `PassThru`" {
     { Start-Process $pingCommand -ArgumentList $pingParam -PassThru -WhatIf } | Should Be $null
}
```

src/Microsoft.PowerShell.Commands.Management/resources/ProcessResources.resx
```diff
@@ -132,6 +132,9 @@
   <data name="ProcessNameForConfirmation" xml:space="preserve">
     <value>{0} ({1})</value>
   </data>
+  <data name="ProcessStartInfo" xml:space="preserve">
+    <value>{0} {1}</value>
+  </data>
```
iSazonov: We re-use `ProcessNameForConfirmation` in Stop-Process and Debug-Process so I believe we can use the same in Start-process.

test/powershell/Modules/Microsoft.PowerShell.Management/Start-Process.Tests.ps1
sarithsutha: Thanks for your suggestion @bergmeister . I've added the test that you described.

src/Microsoft.PowerShell.Commands.Management/resources/ProcessResources.resx
sarithsutha: The original request was to output the exact command being executed. By re-using the ProcessNameForConfirmation, the format of the output will be changed slightly, the arguments would appear in the parentheses. i.e. `C:\WINDOWS\system32\PING.EXE (-n 2 localhost)` instead of `C:\WINDOWS\system32\PING.EXE -n 2 localhost`.
If that is acceptable, I'm happy to make the change.

test/powershell/Modules/Microsoft.PowerShell.Management/Start-Process.Tests.ps1
```diff
@@ -104,6 +104,16 @@ Describe "Start-Process" -Tags @("Feature") {
...
```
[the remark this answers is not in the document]
sarithsutha: I'm sorry, I did not get which `-` were you referring to.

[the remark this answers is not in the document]
iSazonov: I meant next `It` block.

src/Microsoft.PowerShell.Commands.Management/resources/ProcessResources.resx
iSazonov: Thanks for clarify.
Closed.

test/powershell/Modules/Microsoft.PowerShell.Management/Start-Process.Tests.ps1
```diff
@@ -104,6 +104,20 @@ Describe "Start-Process" -Tags @("Feature") {
...
+
+    It "Should be able to use the whatif switch without performing the actual action" {
+        $pingOutput = Join-Path $TestDrive "pingOutput.txt"
+        Start-Process -Wait $pingCommand -ArgumentList $pingParam -RedirectStandardOutput $pingOutput -WhatIf
```
iSazonov: We can remove previous `It` and do check here.

test/powershell/Modules/Microsoft.PowerShell.Management/Start-Process.Tests.ps1
```diff
@@ -105,17 +105,13 @@ Describe "Start-Process" -Tags @("Feature") {
...
 
-    It "Should be able to use the whatif switch without error" {
-        { Start-Process $pingCommand -ArgumentList $pingParam -WhatIf } | Should Not Throw
-    }
-
-    It "Should be able to use the whatif switch without performing the actual action" {
+    It "Should be able to use the -WhatIf switch without performing the actual action" {
         $pingOutput = Join-Path $TestDrive "pingOutput.txt"
-        Start-Process -Wait $pingCommand -ArgumentList $pingParam -RedirectStandardOutput $pingOutput -WhatIf
+        { Start-Process -Wait $pingCommand -ArgumentList $pingParam -RedirectStandardOutput $pingOutput -WhatIf} | Should Not Throw
```
markekraus: I believe `-ErrorAction Stop` is required or it will never actually throw.

iSazonov (approved):
AppVeyor failed on other tests so the PR LGTM with one minor comment..

test/powershell/Modules/Microsoft.PowerShell.Management/Start-Process.Tests.ps1
iSazonov: I agree.

sarithsutha: Thanks @markekraus for pointing this out. Not sure how I missed it.

sarithsutha:
Thanks @iSazonov for reviewing and approving my changes. The minor comment is now incorporated.

markekraus (approved):
LGTM

iSazonov:
@daxian-dbw Could you please review and approve?

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
```diff
@@ -2051,6 +2051,9 @@ protected override void BeginProcessing()
                 startInfo.WindowStyle = _windowstyle;
             }
 
+            string targetMessage = StringUtil.Format(ProcessResources.ProcessStartInfo, startInfo.FileName, startInfo.Arguments.Trim());
```
daxian-dbw: The resource string id `ProcessStartInfo` should be changed to `StartProcessTarget`.

daxian-dbw (changes requested):

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
sarithsutha: @daxian-dbw Renamed the resource string as suggested.

daxian-dbw (approved):
@sarithsutha Thanks!

iSazonov:
@sarithsutha Thanks for your contribution!

PR #4737, opened 2017-09-01
update opensuse docker image to 42.2

TravisEz13:
add a missing dependency to  `amazonlinux`
Address #4310, make images conform with label-schema.org


daxian-dbw (approved):
The RIDs `opensuse.42.1` and `opensuse.13.2` may be removed from `.csproj` files now. But that can be done in a separate PR.

PR #4738, opened 2017-09-01
Replace httpbin.org/get tests With WebListener

markekraus:
* Adds the /Get/ functionality to Weblistener
* Replaces the tests that rely on httpbin.org/get with WebListener

Some tests which use a loop still call httpbin.org. those will be moved when the other functions they rely on have been added. Also, A few of the tests has disabled `should`'s that can now be enabled.

reference #2504

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -669,7 +669,7 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
             # Validate response content
             $jsonContent = $result.Output.Content | ConvertFrom-Json
-            $jsonContent.headers.Host | Should Match "localhost"
+            $jsonContent.headers.Host | Should Be $uri.Authority
             $jsonContent.headers.'User-Agent' | Should Match $expectedAgent
         }
     }
@@ -679,7 +679,7 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         $uri = Get-WebListenerUrl -Test 'Get'
         $result = ExecuteRequestWithOutFile -cmdletName "Invoke-WebRequest" -uri $uri
         $jsonContent = $result.Output | ConvertFrom-Json
-        $jsonContent.headers.Host | Should Match "localhost"
+        $jsonContent.headers.Host | Should Be $uri.Authority
         $jsonContent.headers.'User-Agent' | Should Match "WindowsPowerShell"
     }
 
@@ -1269,7 +1269,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
         $result = ExecuteWebCommand -command $command
 
         # Validate response
-        $result.Output.headers.Host | Should Match "localhost"
+        $result.Output.headers.Host | Should Be $uri.Authority
         $result.Output.headers.'User-Agent' | Should Match "WindowsPowerShell"
         $result.Output.Headers.Connection | Should Be "Close"
     }
@@ -1474,7 +1474,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
             $result = ExecuteWebCommand -command $command
 
             # Validate response
-            $result.Output.headers.Host | Should Match "localhost"
+            $result.Output.headers.Host | Should Be $uri.Authority
             $result.Output.headers.'User-Agent' | Should Match $expectedAgent
         }
     }
@@ -1484,7 +1484,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
         $uri = Get-WebListenerUrl -Test 'Get'
         $result = ExecuteRequestWithOutFile -cmdletName "Invoke-RestMethod" -uri $uri
         $jsonContent = $result.Output | ConvertFrom-Json
-        $jsonContent.headers.Host | Should Match "localhost"
+        $jsonContent.headers.Host | Should Be $uri.Authority
         $jsonContent.headers.'User-Agent' | Should Match "WindowsPowerShell"
     }
 
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -41,7 +41,7 @@ function ExecuteRequestWithOutFile
         [string]
         $cmdletName,
         [string]
-        $uri = "http://httpbin.org/get"
+        $uri = (Get-WebListenerUrl -Test 'Get')
     )
 
     $result = [PSObject]@{Output = $null; Error = $null}
@@ -82,7 +82,7 @@ function ExecuteRequestWithHeaders
         [string]
         $cmdletName,
         [string]
-        $uri = "http://httpbin.org/get"
+        $uri = (Get-WebListenerUrl -Test 'Get')
     )
 
     $result = [PSObject]@{Output = $null; Error = $null}
@@ -442,13 +442,13 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
     It "Validate Invoke-WebRequest -DisableKeepAlive" {
 
         # Operation options
-        $uri = "http://httpbin.org/get"
+        $uri = Get-WebListenerUrl -Test 'Get'
         $command = "Invoke-WebRequest -Uri $uri -TimeoutSec 5 -DisableKeepAlive"
 
         $result = ExecuteWebCommand -command $command
         ValidateResponse -response $result
 
-        $result.Output.Headers["Connection"] | Should Be "Close"
+        $result.Output.Headers.Connection | Should Be "Close"
     }
 
     It "Validate Invoke-WebRequest -MaximumRedirection" {
@@ -641,10 +641,10 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
     It "Validate Invoke-WebRequest -Headers --> Set KeepAlive to false via headers" {
 
-        $uri = "http://httpbin.org/get"
+        $uri = Get-WebListenerUrl -Test 'Get'
         $result = ExecuteRequestWithHeaders -cmdletName Invoke-WebRequest -uri $uri
         ValidateResponse -response $result
-        $result.Output.Headers["Connection"] | Should Be "Close"
+        $result.Output.Headers.Connection | Should Be "Close"
     }
 
     # Validate all available user agents for Invoke-WebRequest
@@ -658,7 +658,7 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
     foreach ($agentName in $agents.Keys)
     {
         $expectedAgent = $agents[$agentName]
-        $uri = "http://httpbin.org/get"
+        $uri = Get-WebListenerUrl -Test 'Get'
         $userAgent = "[Microsoft.PowerShell.Commands.PSUserAgent]::$agentName"
         $command = "Invoke-WebRequest -Uri $uri -UserAgent ($userAgent)  -TimeoutSec 5"
 
@@ -669,17 +669,17 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
             # Validate response content
             $jsonContent = $result.Output.Content | ConvertFrom-Json
-            $jsonContent.headers.Host | Should Match "httpbin.org"
+            $jsonContent.headers.Host | Should Match "localhost"
             $jsonContent.headers.'User-Agent' | Should Match $expectedAgent
         }
     }
 
     It "Validate Invoke-WebRequest -OutFile" {
 
-        $uri = "http://httpbin.org/get"
+        $uri = Get-WebListenerUrl -Test 'Get'
         $result = ExecuteRequestWithOutFile -cmdletName "Invoke-WebRequest" -uri $uri
         $jsonContent = $result.Output | ConvertFrom-Json
-        $jsonContent.headers.Host | Should Match "httpbin.org"
+        $jsonContent.headers.Host | Should Match "localhost"
         $jsonContent.headers.'User-Agent' | Should Match "WindowsPowerShell"
     }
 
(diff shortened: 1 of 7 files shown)


markekraus:
AppVeyor Failure is unrelated

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -41,7 +41,7 @@ function ExecuteRequestWithOutFile
         [string]
         $cmdletName,
         [string]
-        $uri = "http://httpbin.org/get"
+        $uri = (Get-WebListenerUrl -Test 'Get')
```
iSazonov: Please remove parentheses.
We use this many times - can we move the line to BeforeAll?

It seems the time has come to add `ValidateSet()` to `Get-WebListenerUrl` parameters.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -82,7 +82,7 @@ function ExecuteRequestWithHeaders
...
```
iSazonov: Please remove parentheses.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -715,7 +715,8 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
     It "Validate Invoke-WebRequest body is converted to query params for CustomMethod GET" {
 
-        $command = "Invoke-WebRequest -Uri 'http://httpbin.org/get' -CustomMethod GET -Body @{'testparam'='testvalue'}"
+        $uri = Get-WebListenerUrl -Test 'Get'
+        $command = "Invoke-WebRequest -Uri '$uri' -CustomMethod GET -Body @{'testparam'='testvalue'}"
```
iSazonov: Typo  '$uri' ->  $uri

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1262,16 +1263,15 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
     It "Validate Invoke-RestMethod -DisableKeepAlive" {
 
         # Operation options
-        $command = "Invoke-RestMethod -Uri 'http://httpbin.org/get' -TimeoutSec 5 -DisableKeepAlive"
+        $uri = Get-WebListenerUrl -Test 'Get'
+        $command = "Invoke-RestMethod -Uri '$uri' -TimeoutSec 5 -DisableKeepAlive"
```
iSazonov: Typo '$uri' -> $uri

test/tools/WebListener/Controllers/GetController.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            Hashtable args = new Hashtable();
+            foreach (var key in Request.Query.Keys)
+            {
+                args.Add(key, String.Join(" ,", Request.Query[key]));
+            }
+            Hashtable headers = new Hashtable();
+            foreach (var key in Request.Headers.Keys)
+            {
+                headers.Add(key, String.Join(" ,", Request.Headers[key]));
```
iSazonov: We use " ," twice - it is good to use a const.
Do we really have space?

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: Unless I'm mistaken, the Parens are required when assign a default value with a statement in a Param block. At your suggestion I tried removing them but I get all kinds of errors.

As for moving this to the BeforeAll block, I don't think that makes sense in this context since this is part of the `ExecuteRequestWithOutFile` function. 

For the more general concern, we could create a variable, but, as we move the other tests over you will see that it will get less possible to do so in the `BeforeAll` Block without losing meaning in the `It` blocks (in the case where there will be a query string required as well). I would prefer we keep the URL generation to each `It` block for consistency.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: Same pase issue when trying to remove.

test/tools/WebListener/Controllers/GetController.cs
markekraus: I will create a constant, as for the space, it should be the other way around `", "` and yes, the space is there when multiple values are supplied for a given header.

```
X-Fake-Header: Value1
X-Fake-Header: Value2
```

and 

```
X-Fake-Header: Value1, Value2
```

Are equivalent.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: I am not a big fan of unencapsulated strings in evaluated commands. the command would become

```powershell
Invoke-WebRequest -Uri http://localhost:8083/Get -CustomMethod GET -Body @{'testparam'='testvalue'}
```

Which is improper IMO. Does this have to be done? I feel like this change was actually an improvement and I made it everywhere purposefully.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: Ah, sorry - My thoughts was about many `Get-WebListenerUrl -Test 'Get'` in the file - I see only 10 times so I think we can keep it as is.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: Closed.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: See [this comment](https://github.com/PowerShell/PowerShell/pull/4738#discussion_r136797418)

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: Binding is not preprocessor, we don't parse the variable value, so we shouldn't quote variables. `Uri` get right string.
My thoughts was that we should use the file pattern.

markekraus: In this instance the `$Command` string is later evaluated by `Invoke-Expression`. `$uri` is expanded before that evaluation. In instances where the url may contain a `$`, it would be expanded first during the `$command` assignment and then when the `Invoke-Expression` is run on `$command` the `$`  in the url would attempt to be expanded again.

```powershell
# $Field1 is a literal field name, not a variable
$uri = 'http://localhost:8083/Get?$Field1=Value1'
$Command = "Write-Output $uri"
Invoke-Expression $Command
```

Result:
```
http://localhost:8083/Get?=Value1
```

vs:

```powershell
$uri = 'http://localhost:8083/Get?$Field1=Value1'
$Command = "Write-Output '$uri'"
Invoke-Expression $Command
```

Result:

```
http://localhost:8083/Get?$Field1=Value1
```

iSazonov: Sorry again - I lost the context.
Closed.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: Closed.

test/tools/Modules/WebListener/WebListener.psm1
```diff
@@ -112,6 +112,11 @@ function Get-WebListenerUrl {
     [OutputType([Uri])]
     param (
         [switch]$Https,
+        [ValidateSet(
+            'Cert',
+            'Get',
+            '/'
```
iSazonov: It semms we skip a description for '/' in Readme.md.

markekraus: `The default page will return a list of available tests.`

iSazonov: Please clarify. Previously we used `Get-WebListenerUrl -Https` in tests - do we use by default '/'?

markekraus: we do return the `/` by default.

This is the scenario I included it for:

```powershell
$Testcases = @(
    @{
        Name           = 'Get Test'
        Test           = 'Get'
        ExpectedResult = 'Foo'
        Https          = $False
    }
    @{
        Name           = 'Cert test'
        Test           = 'Cert'
        ExpectedResult = 'Bar'
        Https          = $False
    }
    @{
        Name           = 'Default Test'
        Test           = '/'
        ExpectedResult = 'Baz'
        Https          = $False
    }
)

It "<Name>" -TestCases $TestCases {
    Param($Name, $Test, $ExpectedResult, $Https)
    $Url = Get-WebListenerUrl -Test $Test -Https:$Https
    Invoke-RestMethod | Should Match $ExpectedResult
}
```

Test is always passed, but a named test might not always be needed.

iSazonov: Thanks! I think we need the explicit section.

markekraus: The `ValidetSet` already blocks them:

```powershell
Get-WebListenerUrl -Test ""
```
```none
Get-WebListenerUrl : Cannot validate argument on parameter 'Test'. The argument "" does not
belong to the set "Cert,Get,Home,/" specified by the ValidateSet attribute. Supply an argument
that is in the set and then try the command again.
At line:1 char:26
+ Get-WebListenerUrl -Test ""
+                          ~~
    + CategoryInfo          : InvalidData: (:) [Get-WebListenerUrl], ParameterBindingValidationEx
   ception
    + FullyQualifiedErrorId : ParameterArgumentValidationError,Get-WebListenerUrl
```
```powershell
Get-WebListenerUrl -Test $Null
```
```none
Get-WebListenerUrl : Cannot validate argument on parameter 'Test'. The argument "" does not
belong to the set "Cert,Get,Home,/" specified by the ValidateSet attribute. Supply an argument
that is in the set and then try the command again.
At line:1 char:26
+ Get-WebListenerUrl -Test $Null
+                          ~~~~~
    + CategoryInfo          : InvalidData: (:) [Get-WebListenerUrl], ParameterBindingValidationEx
   ception
    + FullyQualifiedErrorId : ParameterArgumentValidationError,Get-WebListenerUrl
```

test/tools/WebListener/README.md
```diff
@@ -30,6 +30,10 @@ $Listener = Start-WebListener -HttpPort 8083 -HttpsPort 8084
 
 # Tests
 
+## / or /Home/
+
+Returns a static HTML page containing links and descriptions of the available tests in WebListener.
```
iSazonov: I think we could add that the page can be used as default for some test as we do in HTTPS tests already.

test/tools/Modules/WebListener/WebListener.psm1
iSazonov: Good!

test/tools/WebListener/README.md
markekraus: heh. I had a blurb about that but removed it thinking it was too wordy (I have a tendency to be overly verbose). I'll add it back.

iSazonov (approved):

iSazonov:
@markekraus Sorry for my inconsiderateness today. Thanks for the great work!

markekraus:
@iSazonov  No problem! I was making a lot of tiny changes in a huge context. I got lost a few times making these changes. I suspected it might make for a tough review.

iSazonov:
It seems WebListener.psm1 is stable now and follow PRs will be small. 😄

markekraus:
@iSazonov It actually will change again. I plan on adding a `Query` parameter to build the query string from a dictionary for the tests that require it. I have held off until it is needed. I'm using it in my own project [here](https://github.com/markekraus/PSRAW/blob/IterativeWork/Tests/tools/Modules/WebListener/WebListener.psm1#L93). After that, though, it will just be a matter of adding additional tests to the `ValidateSet`.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1262,16 +1263,15 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
...
 
         $result = ExecuteWebCommand -command $command
 
         # Validate response
-        $result.Output.headers.Host | Should Match "httpbin.org"
+        $result.Output.headers.Host | Should Match "localhost"
```
TravisEz13: shouldn't the be localhost... not just match?

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1476,17 +1474,17 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
             $result = ExecuteWebCommand -command $command
 
             # Validate response
-            $result.Output.headers.Host | Should Match "httpbin.org"
+            $result.Output.headers.Host | Should Match "localhost"
```
TravisEz13: same, should be

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
TravisEz13: This is in a few locations

markekraus: it actually returns `localhost:<portnumber>` but, I didn't want to couple the port number which may change with the individual tests. If 8083 and 8084 will be the permanent ports, then this can be changes.

markekraus: I guess it could be

```powershell
$result.Output.headers.Host | Should Be $uri.Authority
```

markekraus:
AppVeyor fail is unrelated

iSazonov:
@markekraus The test is now pending - so please rebase to pass CI.

markekraus:
@iSazonov rebased and passing.

TravisEz13 (approved):

iSazonov:
@markekraus Thanks for very useful work!

Issue #4740, opened 2017-09-02
Several messages have 2 spaces between words


korygill:
Steps to reproduce
------------------

Run a script from non-elevated that requires admin, for example.
```
...Start Windows PowerShell by  using the Run as Administrator option...
..............................^^
```

Or just look at the change and see the issue.

Expected behavior
-----------------

Messages should have 1 space between words.
(I did not fix all the `"\w+  \w+"` errors in all the other, non resx files...)

Actual behavior
---------------

Some messages have 2 space between words.

Environment data
----------------

PSVersion                      5.1.15063.502
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.15063.502
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1

iSazonov:
@korygill Thanks for your contribution!

Could you please clarify (samples):
> (I did not fix all the  "\w+ \w+"  errors in all the other, non resx files...)

korygill:
@iSazonov what I mean is there are more instances of 2 spaces between words (~401 hits in 143 files) but most, if not all, appear in comments on non-user-facing strings, so no need to fix those necessarily.  If someone wanted to do this, they could search for "\w+  \w+" in the files using any regex/editor like notepad++ under src and see them.

iSazonov:
@korygill Thanks for clarify!

@TravisEz13 Should we fix such typos and add the check to meta tests?

PR #4741, opened 2017-09-02
Remove double spaces between words in resx files (#4740)

korygill:
Fix for issue #4740.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Diagnostics/resources/GetEventResources.resx
+++ b/src/Microsoft.PowerShell.Commands.Diagnostics/resources/GetEventResources.resx
@@ -185,7 +185,7 @@
     <value>The following export destination path is ambiguous: {0}.</value>
   </data>
   <data name="CounterPathIsInvalid" xml:space="preserve">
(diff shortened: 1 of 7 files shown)


iSazonov (approved):
@korygill Thanks for fixing the typos!

markekraus (approved):
lgtm

SteveL-MSFT (approved):
LGTM

markekraus:
@SteveL-MSFT  Thinking about this further, couldn't some of these possibly be Bucket 2 or 3 breaking changes since they modify errors that some may be relying on?

SteveL-MSFT:
@markekraus users should not rely on parsing error messages as they can be localized.  the error messages are meant for humans to read.  automation should rely on things like the FullyQualifiedErrorId and Exception types.

markekraus:
@SteveL-MSFT Thanks! I was pretty sure that would be the answer but wanted to verify.

iSazonov:
We even have examples of where we changed (removed and added)  FullyQualifiedErrorId and Exception types - this was due to error correction or the addition of new features and was therefore not considered an invalid or destructive change.

PR #4743, opened 2017-09-02
Removed double spaces from .cs and .ps1 files

SteveL-MSFT:
Did a regular expression search "\b  \b" in VSCode while checking if https://github.com/PowerShell/PowerShell/pull/4741/files missed any, so fixing the ones in .cs and .ps1 files.

A few left in that help align comments making them easier to read.

Removed one commented out code that doesn't provide value.

The file SessionStateDriveAPIs.cs has a section that I didn't change, but git complained it had LF line endings and resaving as CRLF, so maybe only that section had mixed line endings.


src/System.Management.Automation/engine/CoreAdapter.cs
```diff
@@ -3122,7 +3122,7 @@ internal static bool DisallowPrivateReflection(TypeInfo typeInfo)
             }
 
             var productAttribute = assembly.GetCustomAttribute<AssemblyProductAttribute>();
-            if (productAttribute != null && string.Equals(productAttribute.Product, "Microsoft� .NET Framework", StringComparison.OrdinalIgnoreCase))
+            if (productAttribute != null && string.Equals(productAttribute.Product, "Microsoft� .NET Framework", StringComparison.OrdinalIgnoreCase))
```
iSazonov: Please revert the change.

src/System.Management.Automation/engine/SessionStateDriveAPIs.cs
```diff
@@ -508,13 +508,13 @@ private PSDriveInfo GetDrive(string name, bool automount)
...
-                {
-                    result = AutomountBuiltInDrive(name); // internally this calls GetDrive(name, false)
-                }
+                // first try to automount as a file system drive
+                result = AutomountFileSystemDrive(name);
+                // if it didn't work, then try automounting as a BuiltIn drive (e.g. "Cert"/"Certificate"/"WSMan")
+                if (result == null)
+                {
+                    result = AutomountBuiltInDrive(name); // internally this calls GetDrive(name, false)
+                }
```
iSazonov: I wonder what is changed here? EOL?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/XmlCommands.cs
```diff
@@ -525,7 +525,7 @@ protected override void StopProcessing()
         private CustomSerialization _serializer;
 
         /// <summary>
-        ///Memory Stream used for  serialization
+        ///Memory Stream used for serialization
```
iSazonov: Please add space before `Memory`.
It seems we should set final point.

src/System.Management.Automation/engine/CoreAdapter.cs
SteveL-MSFT: Will revert this

src/System.Management.Automation/engine/SessionStateDriveAPIs.cs
SteveL-MSFT: Git gave some warnings about line endings (and performing some LF<->CRLF conversions on my behalf since I was doing this on Windows), so I think that's what this is.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/XmlCommands.cs
SteveL-MSFT: Ok

src/System.Management.Automation/engine/SessionStateDriveAPIs.cs
iSazonov: Thanks for clarify. 
Closed.

iSazonov (approved):
So many files - we need an approvement from someone else.

src/System.Management.Automation/commands/utility/FormatAndOutput/common/FormatViewManager.cs
```diff
@@ -95,7 +95,7 @@ private static string PSObjectTypeName(PSObject so)
                                     FormattingCommandLineParameters parameters)
         {
             ViewDefinition view = null;
-            const string findViewType = "FINDING VIEW  TYPE: {0}";
+            const string findViewType = "FINDING VIEW TYPE: {0}";
             const string findViewShapeType = "FINDING VIEW {0} TYPE: {1}";
             const string findViewNameType = "FINDING VIEW NAME: {0}  TYPE: {1}";
```
lzybkr: You missed one.

src/System.Management.Automation/engine/CoreAdapter.cs
```diff
@@ -3122,7 +3122,7 @@ internal static bool DisallowPrivateReflection(TypeInfo typeInfo)
             }
 
             var productAttribute = assembly.GetCustomAttribute<AssemblyProductAttribute>();
-            if (productAttribute != null && string.Equals(productAttribute.Product, "Microsoft� .NET Framework", StringComparison.OrdinalIgnoreCase))
+            if (productAttribute != null && string.Equals(productAttribute.Product, "Microsoft® .NET Framework", StringComparison.OrdinalIgnoreCase))
```
lzybkr: Can we change this to `Microsoft\u00a9` (assuming that is the correct character). When we run automated tools, they always mess this character up.

src/System.Management.Automation/engine/SessionStateDriveAPIs.cs
lzybkr: The original file had inconsistent line endings, `LF` everywhere but this region. Your script changed the file to have consistent line endings, so git sees this region differently I guess.

src/System.Management.Automation/engine/CoreAdapter.cs
iSazonov: We have the character in some files - maybe fix this all in follow PR?

src/System.Management.Automation/commands/utility/FormatAndOutput/common/FormatViewManager.cs
SteveL-MSFT: will fix

src/System.Management.Automation/engine/CoreAdapter.cs
SteveL-MSFT: Created issue https://github.com/PowerShell/PowerShell/issues/4752 as we should just fix that (and potentially Trademark as well)

lzybkr (approved):

PR #4745, opened 2017-09-03
Fix of #4665 to have a simple file based check for the VC++ 2015 redistributables

### LINKED ISSUE #4665: Incorrect Windows Installer Check for VC++ Redistributable Dependency
The root cause of this issue is that the check is specific to a registry key that gets created when the VC++ redistributable package is directly installed via the suggested link. It overlooks scenarios where those files are installed via Windows Update.

A secondary issue is that the test should be tagged to only execute on Windows systems.

Steps to reproduce
------------------

Build the MSI for a Windows SKU using commit 54e892ab06fc0b705a329be8061df9f77cbd54f8 and run it on a system that does not have Visual Studio 2015 or the VC++ redistributable package installed via MSI. The dependency is required, but it can be installed via Windows Update.

Expected behavior
-----------------

The MSI installs without error.

Actual behavior
---------------

The failure message is, "PowerShell requires the Visual Studio C++ 2015 x64 redistributables to be installed..." This occurs on a fully patched system that has the dependency installed via Windows Update.

Environment data
----------------

Beta.5 + commit 54e892ab06fc0b705a329be8061df9f77cbd54f8

bergmeister:
To fix #4665 
The previous check registry check of `SOFTWARE\Microsoft\DevDiv\VC\Servicing\14.0\RuntimeMinimum` turned out to not work in the scenario whereby the redistributables were installed via Windows update (which does not use MSI) because the registry entry did not get populated in this case.
In the issue discussion @SteveL-MSFT suggested to have only a simple file based check, therefore this PR simply checks for the existence of `vcruntime140.dll` in the `Windows/System32` folder

The 'Pending' attribute was removed from existing tests since the download links are now present again and the tests were improved using Pester TestCases.

### DIFF
--- a/assets/Product.wxs
+++ b/assets/Product.wxs
@@ -61,7 +61,7 @@
     <!--We need to show EULA, and provide option to customize download location-->        
     <Property Id="WIXUI_INSTALLDIR" Value="INSTALLFOLDER" />
 
-    <!-- Prerequisite check for Windows Universal C time -->
+    <!-- Prerequisite check for Windows Universal C runtime -->
     <Property Id="UCRTINSTALLED" Secure="yes">
       <DirectorySearch Id="Windows_System32_search1" Path="[WindowsFolder]System32" Depth="0">
         <FileSearch Name="ucrtbase.dll"/>
--- a/assets/Product.wxs
+++ b/assets/Product.wxs
@@ -62,7 +62,7 @@
     <Property Id="WIXUI_INSTALLDIR" Value="INSTALLFOLDER" />
 
     <!-- Prerequisite check for Windows Universal C runtime -->
-    <Property Id="Universal_CRuntime_Installed" Secure="yes">
+    <Property Id="UNIVERSAL_C_RUNTIME_INSTALLED" Secure="yes">
       <DirectorySearch Id="WindowsDirectory" Path="[WindowsFolder]">
         <DirectorySearch Id="System32" Path="System32">
           <FileSearch Id="ucrtbase" Name="ucrtbase.dll"/>
@@ -70,17 +70,17 @@
       </DirectorySearch>
     </Property>	
     <Condition Message="$(env.ProductName) requires the Universal C Runtime to be installed. You can find a download link to it here: https://github.com/PowerShell/PowerShell/blob/master/docs/installation/windows.md#prerequisites">
-      <![CDATA[Installed OR Universal_CRuntime_Installed]]>
+      <![CDATA[Installed OR UNIVERSAL_C_RUNTIME_INSTALLED]]>
     </Condition>
 
      <!-- Prerequisite check for Visual Studio 2015 C++ redistributables -->
-     <Property Id="Visual_CPP_Runtime_Installed" Secure="yes">
+     <Property Id="VISUAL_CPP_RUNTIME_INSTALLED" Secure="yes">
       <DirectorySearchRef Id="System32" Parent="WindowsDirectory" Path="System32">
         <FileSearch Id="vcruntime140" Name="vcruntime140.dll"/>
       </DirectorySearchRef>
     </Property>	
     <Condition Message="$(env.ProductName) requires the Visual Studio 2015 C++ redistributables to be installed. You can find a download link to it here: https://github.com/PowerShell/PowerShell/blob/master/docs/installation/windows.md#prerequisites">
-      <![CDATA[Installed OR Visual_CPP_Runtime_Installed]]>
+      <![CDATA[Installed OR VISUAL_CPP_RUNTIME_INSTALLED]]>
     </Condition>
 
     <Directory Id="TARGETDIR" Name="SourceDir">
--- a/assets/Product.wxs
+++ b/assets/Product.wxs
@@ -60,15 +60,25 @@
 
     <!--We need to show EULA, and provide option to customize download location-->        
     <Property Id="WIXUI_INSTALLDIR" Value="INSTALLFOLDER" />
-	
-    <!-- Prerequisites check for Windows Universal C time and Visual Studio 2015 C++ redistributables -->
+
+    <!-- Prerequisite check for Windows Universal C time -->
     <Property Id="UCRTINSTALLED" Secure="yes">
-      <DirectorySearch Id="Windows_System32" Path="[WindowsFolder]System32" Depth="0">
+      <DirectorySearch Id="Windows_System32_search1" Path="[WindowsFolder]System32" Depth="0">
         <FileSearch Name="ucrtbase.dll"/>
       </DirectorySearch>
     </Property>	
     <Condition Message="$(env.ProductName) requires the Universal C Runtime to be installed. You can download it here: https://www.microsoft.com/download/details.aspx?id=50410">
       <![CDATA[Installed OR UCRTINSTALLED]]>
+    </Condition>
+
+     <!-- Prerequisite check for Visual Studio 2015 C++ redistributables -->
+     <Property Id="VCRUNTIMEINSTALLED" Secure="yes">
+      <DirectorySearch Id="Windows_System32_search2" Path="[WindowsFolder]System32" Depth="0">
+        <FileSearch Name="vcruntime140.dll"/>
+      </DirectorySearch>
+    </Property>	
+    <Condition Message="$(env.ProductName) requires the Visual Studio 2015 C++ redistributables to be installed. You can download it here: https://www.microsoft.com/download/details.aspx?id=48145">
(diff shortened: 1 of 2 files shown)


assets/Product.wxs
```diff
@@ -60,17 +60,27 @@
 
     <!--We need to show EULA, and provide option to customize download location-->        
     <Property Id="WIXUI_INSTALLDIR" Value="INSTALLFOLDER" />
-	
-    <!-- Prerequisites check for Windows Universal C time and Visual Studio 2015 C++ redistributables -->
+
+    <!-- Prerequisite check for Windows Universal C time -->
     <Property Id="UCRTINSTALLED" Secure="yes">
-      <DirectorySearch Id="Windows_System32" Path="[WindowsFolder]System32" Depth="0">
+      <DirectorySearch Id="Windows_System32_search1" Path="[WindowsFolder]System32" Depth="0">
```
SteveL-MSFT: Would it be better to call this `Universal_Runtime_Search`?

assets/Product.wxs
```diff
@@ -60,17 +60,27 @@
...
         <FileSearch Name="ucrtbase.dll"/>
       </DirectorySearch>
     </Property>	
     <Condition Message="$(env.ProductName) requires the Universal C Runtime to be installed. You can download it here: https://www.microsoft.com/download/details.aspx?id=50410">
       <![CDATA[Installed OR UCRTINSTALLED]]>
     </Condition>
 
+     <!-- Prerequisite check for Visual Studio 2015 C++ redistributables -->
+     <Property Id="VCRUNTIMEINSTALLED" Secure="yes">
+      <DirectorySearch Id="Windows_System32_search2" Path="[WindowsFolder]System32" Depth="0">
```
SteveL-MSFT: I think it's better to calls this `Visual_CPP_Runtime_Search`

assets/Product.wxs
```diff
@@ -60,17 +60,27 @@
 
     <!--We need to show EULA, and provide option to customize download location-->        
     <Property Id="WIXUI_INSTALLDIR" Value="INSTALLFOLDER" />
-	
-    <!-- Prerequisites check for Windows Universal C time and Visual Studio 2015 C++ redistributables -->
+
+    <!-- Prerequisite check for Windows Universal C time -->
```
iSazonov: Typo time -> runtime.

assets/Product.wxs
```diff
@@ -60,17 +60,27 @@
...
       <![CDATA[Installed OR UCRTINSTALLED]]>
     </Condition>
 
+     <!-- Prerequisite check for Visual Studio 2015 C++ redistributables -->
+     <Property Id="VCRUNTIMEINSTALLED" Secure="yes">
+      <DirectorySearch Id="Windows_System32_search2" Path="[WindowsFolder]System32" Depth="0">
+        <FileSearch Name="vcruntime140.dll"/>
+      </DirectorySearch>
+    </Property>	
+    <Condition Message="$(env.ProductName) requires the Visual Studio 2015 C++ redistributables to be installed. You can download it here: https://www.microsoft.com/download/details.aspx?id=48145">
```
iSazonov: The link is not persistent and may be changed in any time. I think it is better remove it. Anybody can find it by means of Bing/Google.

assets/Product.wxs
bergmeister: I can make 'search' uppercase but the numbers at the end are needed now because otherwise WiX gives a compilation error that the DirectorySearch Id is duplicated. This seems to be a WiX peculiarity in the case of file search properties and I have seen other people do the same as well but if you know a better way of doing it, feel free to propose.

SteveL-MSFT: I see.  Not a blocking issue for me.

bergmeister: Sorry, I see only now that I misread your 2 comment value suggestions as `Windows_System32_Search`, which would have resulted in the duplicated DirectorySearch Id compilation error (which I actually had in the first place when I just copy pasted the existing block as a template with some modifications). So, yes, you could name it `Universal_Runtime_Search` but I am not sure if there is much value because the property name already contains this information and this Id is not used anywhere (but needed for compilation). Maybe it is better to rename the property value(s) instead?

assets/Product.wxs
bergmeister: Thanks. I fixed it.

assets/Product.wxs
bergmeister: This is why there are tests to check that the links are still active. Also the original issue #4458 requested that it `'should give a good error message directing user to download'`

assets/Product.wxs
SteveL-MSFT: My thinking is that if we need to extend this in the future, `Windows_System32_searchN` doesn't really provide much information.  If the requirement is simply it has to be unique, I think something more descriptive to what the search does is preferred.

assets/Product.wxs
SteveL-MSFT: I don't know how often those links get recycled, but it seems another solution would be to link back to our pre-requisites page: https://github.com/PowerShell/PowerShell/blob/master/docs/installation/windows.md#prerequisites

bergmeister: Ok. Do you want me to link to the pre-requisites page then or do you want some time to think about it first? But if those links get recycled then the pre-requisites page will suffer from the same issue (except for that it can be updated even after release, which is not an insignificant advantage)

SteveL-MSFT: @bergmeister the fact that we can update the pre-reqs page fairly easily and quickly is the advantage I was thinking

assets/Product.wxs
bergmeister: OK, I did more research and found a documented solution, which allows to define it once and then reference it in order to get rid of this problem.

test/powershell/Installer/WindowsInstaller.Tests.ps1
```diff
@@ -14,24 +14,22 @@ Describe "Windows Installer" -Tags "Scenario" {
...
+    It "WiX (Windows Installer XML) file contains download link '<downloadLink>' for '<Name>'" -TestCases $downloadLinks -Test {
+        Param ([string]$downloadLink)
+        (Get-Content $wixProductFile -Raw).Contains($downloadLink) | Should Be $true
     }
 
-}
\ No newline at end of file
+    It "Download link '<downloadLink>' for '<Name>' is reachable" -TestCases $downloadLinks -Test {
+        Param ([string]$downloadLink)
+        (Invoke-WebRequest $universalCRuntimeDownloadLink.Replace("https://", 'http://')) | Should Not Be $null
```
SteveL-MSFT: Why are you replacing HTTPS with HTTP?

SteveL-MSFT: Also, perhaps if we just want to validate the download links haven't been removed, we should just check that you get a 200 status?

SteveL-MSFT (changes requested):

test/powershell/Installer/WindowsInstaller.Tests.ps1
bergmeister: The replacement of https with http was because I sometimes got errors, I can take it out if you don't like it.
If you take an invalid URL like e.g. `https://www.microsoft.com/download/details.aspx?id=5041000000`, then you still get a 200 response because I get redirected to an error page but the returned object is `$null`

SteveL-MSFT: I think it's better to leave as HTTPS.  Depending on why it's failing, you could try it a few times before failing the test since this is `Scenario`.

For the $null vs 200 status, you should add a comment so other people understand why you made the choice.

assets/Product.wxs
iSazonov: @SteveL-MSFT Can PG request a permanent link for Universal C runtime?

SteveL-MSFT (approved):

bergmeister:
Please do not merge. My last commit, which replaced the Urls had a WiX compilation error (maybe due to the hash in the url). The Appveyor build should be improved to go red if it does not produce an MSI.

SteveL-MSFT:
@bergmeister can you open an issue to test the MSI packaging?

bergmeister:
@SteveL-MSFT I opened issue #4760 for it and fixed the compilation error (Property Id had to be public due to the DirectorySearch and therefore did not allow lowercase characters). I installed the MSI produced by the build on my machine as a smoke test.

mirichmo:
I restarted the Travis build to see if it would fix itself.

mirichmo (approved):

PR #4746, opened 2017-09-03
Get-Verb: add verb descriptions and alias prefixes

### LINKED ISSUE #4372: Get-Verb should list alias prefixes too (extend System.Management.Automation.VerbInfo)
<sup>Related documentation issue: https://github.com/PowerShell/PowerShell-Docs/issues/1480</sup>

`Get-Help` helpfully lists the verb names and the group they fall into.

However, it doesn't list the - also standardized - **alias prefix** associated with each verb, such as `g` for `Get` and `ip` for `Import`, as defined in [Approved Verbs for Windows PowerShell Commands](https://technet.microsoft.com/en-us/library/ms714428(v=vs.85).aspx).

Having these aliases readily available for reference would help establish a common, consistent language for _aliases_ too, which in the long run would hopefully make people move away from the hotly debated _legacy_ aliases.

It looks like all that is needed is to an `.AliasPrefix` property to type `[System.Management.Automation.VerbInfo]`.

Current behavior
------------------

```powershell
> Get-Verb | Select-Object -First 3

Verb  Group 
----  ----- 
Add   Common
Clear Common
Close Common
```

Desired behavior
-----------------

```powershell
> Get-Verb | Select-Object -First 3

Verb  Group Alias Prefix
----  ----- ---------
Add   Common a
Clear Common cl
Close Common cs
```

For easy reference, here's the current mapping (without group) information:

```none
Add (a)
Approve (ap)
Assert (as)
Backup (ba)
Block (bl)
Checkpoint (ch)
Clear (cl)
Close (cs)
Compare (cr)
Complete (cp)
Compress (cm)
Confirm (cn)
Connect (cc)
Convert (cv)
ConvertFrom (cf)
ConvertTo (ct)
Copy (cp)
Debug (db)
Deny (dn)
Disable (d)
Disconnect (dc)
Dismount (dm)
Edit (ed)
Enable (e)
Enter (et)
Exit (ex)
Expand (en)
Export (ep)
Find (fd)
Format (f)
Get (g)
Grant (gr)
Group (gp)
Hide (h)
Import (ip)
Initialize (in)
Install (is)
Invoke (i)
Join (j)
Limit (l)
Lock (lk)
Measure (ms)
Merge (mg)
Mount (mt)
Move (m)
New (n)
Open (op)
Out (o)
Ping (pi)
Protect (pt)
Publish (pb)
Push (pu)
Read (rd)
Receive (rc)
Redo (re)
Register (rg)
Remove (r)
Rename (rn)
Repair (rp)
Request (rq)
Reset (rs)
Resolve (rv)
Restart (rt)
Restore (rr)
Resume (ru)
Revoke (rk)
Save (sv)
Search (sr)
Select (sc)
Send (sd)
Set (s)
Show (sh)
Skip (sk)
Split (sl)
Start (sa)
Step (st)
Stop (sp)
Submit (sb)
Suspend (ss)
Switch (sw)
Sync (sy)
Test (t)
Trace (tr)
Unblock (ul)
Undo (un)
Uninstall (us)
Unlock (uk)
Unprotect (up)
Unpublish (ub)
Unregister (ur)
Update (ud)
Use (u)
Wait (w)
Watch (wc)
Write (wr)
```

Environment data
----------------

```powershell
PowerShell Core v6.0.0-beta.4
```

### LINKED ISSUE #4732: VerbInfo should have a Description member
VerbInfo only has Group and Verb members, would be convenient to have a `Description` member

Steps to reproduce
------------------

```powershell
PS> get-verb
```

Expected behavior
-----------------

```none
Verb        Group        Description
----        -----        -----
Add         Common       Foo
Clear       Common       Bar
```

Actual behavior
---------------

```none
Verb        Group
----        -----
Add         Common
Clear       Common
```

Environment data
----------------

```powershell
GitCommitId                    v6.0.0-beta.6
OS                             Microsoft Windows 10.0.16350
```

Tadas:
Fixes #4732 - VerbInfo should have a Description member
Fixes #4372 - Get-Verb should list alias prefixes too (extend System.Management.Automation.VerbInfo)

### Summary
Added `VerbDescriptions` and `VerbAliasPrefixes` static classes with description and alias strings. Not actually sure if this is the best way to store the descriptions.

To keep the descriptions short I only took the first sentence from https://msdn.microsoft.com/en-us/library/ms714428(v=vs.85).aspx "Action" field. Aliases come from the same page

Need document [approved prefixes](https://github.com/PowerShell/PowerShell/pull/4746#discussion_r137717675) for some new verbs:
- [x] Optimize - om
- [x] Resize - rz
- [x] Build - bd
- [x] Deploy - dp
- [x] Complete - cmp


iSazonov:
@Tadas Thanks for your contribution!

@lzybkr Could you please comment. Currently we use reflection and dynamically re-create `verbinfo` objects in `Get-Verb` cmdlet and also use reflection and build `s_validVerbs` and `s_recommendedAlternateVerbs` in Verbs.cs - perhaps the time has come to optimize performance of the code? Can we use someting like we use in `Types_Ps1Xml.cs`?

src/System.Management.Automation/utils/Verbs.cs
```diff
@@ -592,6 +592,1018 @@ public static class VerbsOther
     }
 
     /// <summary>
+    /// Verb descriptions.
+    /// </summary>
+    public static class VerbDescriptions
```
lzybkr: Descriptions needs to be in a resx file so they can be localized.

src/System.Management.Automation/utils/Verbs.cs
```diff
@@ -592,6 +592,1018 @@ public static class VerbsOther
...
+
+        /// <summary>
+        /// "Block" verb alias prefix
+        /// </summary>
+        public const string Block = "bl";
+
+        /// <summary>
+        /// "Build" verb alias prefix
+        /// </summary>
+        public const string Build = "";
```
lzybkr: @HemantMahawar should probably be defining a prefix for the verbs which have none: `Build`, `Deploy`, `Optimize`, and `Resize`.

lzybkr:
@Tadas @iSazonov - we don't need a cache for these new properties and reflection is fine - `Get-Verb` isn't important for performance.

`s_recommendedAlternateVerbs` should probably be `Lazy` - it shouldn't be normal to need it.

`s_validVerbs` could avoid reflection, but use reflection in debug builds to ensure that nothing was missed, but it never showed up in profiles during startup, so the difference might not be measurable.

src/System.Management.Automation/utils/Verbs.cs
iSazonov: Should it be internal?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetVerbCommand.cs
```diff
@@ -71,7 +71,20 @@ protected override void ProcessRecord()
 
                         VerbInfo verb = new VerbInfo();
                         verb.Verb = field.Name;
+
+                        FieldInfo aliasField = typeof(VerbAliasPrefixes).GetField(field.Name);
```
dantraMSFT: Suggestion: Refactor this as static method in the associated class. It will provide more flexibility with refactoring, should it be necessary.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetVerbCommand.cs
```diff
@@ -71,7 +71,20 @@ protected override void ProcessRecord()
...
+                        if(aliasField != null)
+                        {
+                            verb.AliasPrefix = (string)aliasField.GetValue(null);
+                        }
+
                         verb.Group = groupName;
+
+                        FieldInfo descriptionField = typeof(VerbDescriptions).GetField(field.Name);
```
dantraMSFT: The description lookup should be a static method in the VerbDescription class; especially since the strings should be in a resx file to enable localization.

dantraMSFT (changes requested):

src/System.Management.Automation/utils/Verbs.cs
HemantMahawar: - Build --> bd
- Deploy --> dp
- Optimize --> om
- Resize --> rs

/cc @joeyaiello

src/System.Management.Automation/utils/Verbs.cs
```diff
@@ -594,513 +594,21 @@ public static class VerbsOther
...
-        /// </summary>
-        public const string Watch = "Continually inspects or monitors a resource for changes";
-
-        /// <summary>
-        /// Adds information to a target
-        /// </summary>
-        public const string Write = "Adds information to a target";
+        public static string GetVerbDescription(string verb)
+        {
+            return VerbDescriptionStrings.ResourceManager.GetString(verb + "VerbDescription");
```
iSazonov: I do not see the benefits of "verbdescription". I think we could remove it.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Verb.Tests.ps1
```diff
@@ -50,4 +50,14 @@
             {Get-Verb -Group $group} | Should Not Throw
         }
     }
+
+    It "Should have descriptions for all verbs" {
+        $noDesc = Get-Verb | Where-Object { [string]::IsNullOrEmpty($_.Description) }
+        $noDesc | Should be $null
```
iSazonov: For "all verbs" we should _count_.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Verb.Tests.ps1
```diff
@@ -50,4 +50,14 @@
...
+    }
+
+    It "Should have alias prefixes for all verbs" {
+        $noPrefix = Get-Verb | Where-Object { [string]::IsNullOrEmpty($_.AliasPrefix) }
+        $noPrefix | Should be $null
```
iSazonov: The same. We should count verbs and verbs with AliasPrefix.

src/System.Management.Automation/utils/Verbs.cs
```diff
@@ -1601,6 +1109,22 @@ public static class VerbAliasPrefixes
         /// "Write" verb alias prefix
         /// </summary>
         public const string Write = "wr";
+
+        /// <summary>
+        /// Gets verb prefix
+        /// </summary>
+        public static string GetVerbPrefix(string verb)
```
iSazonov: Maybe `GetVerbAliasPrefix` ?

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Verb.Tests.ps1
Tadas: You mean I should test for `$verbsWithDescription.count | Should be $allVerbs.count` or do you mean something else?

iSazonov: Yes. We should count verbs and verbs with Description.

Tadas:
I've just noticed that Copy and Complete verbs have the same alias "cp". Should those be unique?

Tadas:
@HemantMahawar @joeyaiello what do you think of the duplicate verb prefixes for Complete and Copy?

src/System.Management.Automation/utils/Verbs.cs
```diff
@@ -592,6 +592,542 @@ public static class VerbsOther
...
+        public static string Get(string verb)
+        {
+            return VerbDescriptionStrings.ResourceManager.GetString(verb + "VerbDescription");
+        }
+    }//VerbDescriptions
+
+    /// <summary>
+    /// Verb Alias prefixes.
+    /// </summary>
+    internal static class VerbAliasPrefixes
```
lzybkr: Maybe these should be attributes, like:

```c#
    public static class VerbsCommon
    {
        /// <summary>
        /// Synonyms: Add to, append or attach.
        /// </summary>
        [AliasPrefix("a")]
        public const string Add = "Add";
}
```

iSazonov: Looks good as common pattern.
Should we do this in the _special_ case?

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and we should change alias for `Complete` to `cmp`

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetVerbCommand.cs
```diff
@@ -71,7 +71,9 @@ protected override void ProcessRecord()
 
                         VerbInfo verb = new VerbInfo();
                         verb.Verb = field.Name;
+                        verb.AliasPrefix = VerbAliasPrefixes.GetVerbAliasPrefix(field.Name);
                         verb.Group = groupName;
+                        verb.Description = VerbDescriptions.Get(field.Name);
```
iSazonov: Above we use `VerbAliasPrefixes.GetVerbAliasPrefix`. I believe we should use the same pattern and use `VerbDescriptions.GetVerbDescription`.

iSazonov (commented):
Leave a comment

iSazonov (approved):

iSazonov:
@Tadas I edit the PR description for future documentation.

mirichmo:
@dantraMSFT Please re-review

dantraMSFT (changes requested):
I suggest you remove the VerbDescription part of the resource name in VerbDescriptionStrings.resx and avoid the string allocation in VerbDescriptions.GetVerbDescription (verb + "VerbDescription").  By definition, there should not be a name collision so the allocation is unnecessary.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Verb.Tests.ps1
```diff
@@ -50,4 +50,24 @@
             {Get-Verb -Group $group} | Should Not Throw
         }
     }
+
+    It "Should have descriptions for all verbs" {
```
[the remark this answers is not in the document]
iSazonov: I'd prefer do this in place without a helper function

SteveL-MSFT:
@dantraMSFT can you review again?

dantraMSFT (approved):

dantraMSFT:
LGTM!

iSazonov (approved):

iSazonov:
@Tadas Thanks for your contribution! Welcome with new ones!

PR #4748, opened 2017-09-04
VS 2017 sln

lzybkr:
I created a VS2017 solution file for the powershell-win-core csproj and everything it references.

It "upgraded" each csproj - but I reverted all changes but the one to suppress future upgrades.

The solution doesn't build and I haven't investigated why, but it's useful for editing across the entire project.

### DIFF
--- a/src/powershell-win-core/powershell-win-core.csproj
+++ b/src/powershell-win-core/powershell-win-core.csproj
@@ -1,7 +1,5 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
-
-  <Import Project="..\..\PowerShell.Common.props"/>
-
+﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
+  <Import Project="..\..\PowerShell.Common.props" />
   <PropertyGroup>
     <Description>PowerShell Core on Windows top-level project</Description>
     <AssemblyName>powershell</AssemblyName>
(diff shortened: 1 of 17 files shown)


iSazonov:
Maybe put the file in root folder with PowerShell.Common.props file? Currently I use @powercode sln file such way.

lzybkr:
I'll defer to @daxian-dbw on taking the sln from @powercode or this one. I can move it to the root if we decide to take this.

src/powershell-win-core/powershell-win-core.csproj
```diff
@@ -55,5 +53,7 @@
   <PropertyGroup Condition=" '$(Configuration)' == 'CodeCoverage' ">
     <DebugType>full</DebugType>
   </PropertyGroup>
-
+  <PropertyGroup>
+    <RootNamespace>Microsoft.PowerShell</RootNamespace>
```
daxian-dbw: Is the <RootNamespace> necessary to be added?
If so, can you please move it to the `PropertyGroup` section at top of the file?

daxian-dbw (changes requested):

daxian-dbw:
I will take this `.sln` file since we already have a PR.
@lzybkr please move the `.sln` to the root and rename it to `PowerShell-Win.sln`. VS2017 for Mac is available, so maybe someone will create a `PowerShell-Mac.sln` for that.

daxian-dbw:
@lzybkr Could you please address the comments when you have time? Thanks!

src/powershell-win-core/powershell-win-core.csproj
lzybkr: It avoids a warning when opening the solution. I moved it to the top.

lzybkr:
Comments addressed and I rebased on master to fix the conflict.

daxian-dbw (approved):
LGTM. Thanks!

Issue #4700, opened 2017-08-29
Rename $IsOSX to $IsmacOS


SteveL-MSFT:
Apple calls it macOS, we should be consistent (our documentation is mostly consistent I think, but we have $IsOSX which makes it inconsistent.  I propose we make this breaking change now otherwise never.

thezim:
$IsDarwin? It would cover any Apple operating system current and future.

SteveL-MSFT:
@mklement0 perhaps if we're willing to make this breaking change, perhaps we should consider adding the prefix at the same time

markekraus:
I agree with the name change, but I'm thinking `IsMacOS` is better. I see other languages have implemented `isMacOS` in camelCase. IMO, `IsMacOS` is the Pascal Case equivalent. Lower cased Branding doesn't have to remain.

I also agree with @mklement0 that creating the `ps:` variable scope and moving it there would be best.

SteveL-MSFT:
@PowerShell/powershell-committee discussed this weighing different pros and cons.  I think where we ended up is that for these specific variables, we'll stick with $IsWindows, $IsLinux, and $IsMacOS (staying with Pascal casing for consistency).  Authoring a PSProvider is not a small effort and doesn't seem to provide enough value for discovery purposes and adding a new variable scope would break existing variable scope semantics.

TL;DR we'll just rename to $IsMacOS

For the general interop issue, we'll defer that to the RFC and don't believe these variables conflict with that resolution.

mklement0:
@SteveL-MSFT: 

Understood.

> the general interop issue

To be clear, the idea of a `$ps:` namespace/scope is about _more_ than just interop: it's about _all PowerShell-controlled variables_, which fall into the following categories:

* interop variables: platform abstractions, such as the variable discussed here and a variable pointing to the platform-appropriate folder for temporary files.

* automatic, typically read-only variables: e.g., `$PID`, `$Host`, and `$Matches`

* preference variables: e.g., `$ErrorActionPreference` and `$OFS`

Representing each of these categories in its own namespace/scope is an option, but may be overkill (and more cumbersome to type).

I know we've discussed such an RFC, but I don't think it exists yet, does it?

SteveL-MSFT:
I believe @joeyaiello was supposed to author that RFC we keep referencing...

[...]

PR #4757, opened 2017-09-06
Rename $IsOSX to $IsMacOS

SteveL-MSFT:
PS-Committee decided we should be consistent with our naming and conform to Apple's use of macOS instead of OSX, however, for readability and consistently we are staying with Pascal casing.

Also renamed instances of variations of `OSX` to `macOS` for consistency in comments/text (separate commit to make it easier to review)

Fix https://github.com/PowerShell/PowerShell/issues/4700

### DIFF
--- a/tools/travis.ps1
+++ b/tools/travis.ps1
@@ -42,7 +42,7 @@ function Set-DailyBuildBadge
     $storageAccountKey = $Env:TestResultAccountKey
 
     # this is the url referenced in README.MD which displays the badge
-    $platform = if ( $IsOSX ) { "OSX" } else { "Linux" }
+    $platform = if ( $IsLinux ) { "Linux" } else { "OSX" }
     $Url = "https://jimtru1979.blob.core.windows.net/badges/DailyBuildStatus.${platform}.svg"
 
     $body = $content
--- a/tools/travis.ps1
+++ b/tools/travis.ps1
@@ -29,7 +29,7 @@ function Get-DailyBadge
 #
 # the best way to do this would be if travis-ci supported a webcall to get
 # the status of cron_job builds, but it doesn't, so we have this
-# also, since we can have a build on Linux which succeeds and one on OSX which
+# also, since we can have a build on Linux which succeeds and one on macOS which
 # doesn't we'll set the appropriate badge so the the README can pick it up
 function Set-DailyBuildBadge
 {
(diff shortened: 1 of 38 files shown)


daxian-dbw:
I don't see changes to `build.psm1`, is that file missed from this PR?

SteveL-MSFT:
@daxian-dbw not sure how I missed that, will fix

iSazonov:
[Apple Inc. ](https://www.apple.com/lae/macos/sierra/) uses `macOS` not `MacOS` - maybe we should use the same too? - `$IsmacOS`

SteveL-MSFT:
@iSazonov we had that discussion in the issue, the consensus is that $IsMacOS is preferred over $IsmacOS for readability and other languages (like Java) use isMacOS, so it's already somewhat common.

tools/travis.ps1
```diff
@@ -42,7 +42,7 @@ function Set-DailyBuildBadge
     $storageAccountKey = $Env:TestResultAccountKey
 
     # this is the url referenced in README.MD which displays the badge
-    $platform = if ( $IsOSX ) { "OSX" } else { "Linux" }
+    $platform = if ( $IsLinux ) { "Linux" } else { "OSX" }
```
iSazonov: Minor comment - maybe replace "OSX" too?

iSazonov (commented):
LGTM

tools/travis.ps1
SteveL-MSFT: The file stored in Azure for the badge has OSX in the filename.  I'm fine leaving that one.

daxian-dbw (approved):
The package name for macOS should also be changed, but that can be a separate PR.

markekraus:
Should this run through the CI with [feature] to ensure none of the feature tests broke with the change?

SteveL-MSFT:
@markekraus good point, I'll resubmit

SteveL-MSFT:
AppVeyor failure is due to the current issue causing nightly to fail and @dantraMSFT is looking into it

dantraMSFT:
Yes, the failure is due to https://github.com/PowerShell/PowerShell/pull/4763. You can ignore it.

SteveL-MSFT:
Rebased to pick up fix

iSazonov (approved):

TravisEz13 (approved):

Issue #4750, opened 2017-09-04
Get-Date -UFormat inconsistencies compared to Linux strftime


timparkinson:
With reference to #4264 first reported by @chuanjiao10 - on further examination there are a number of UFormats outputs which diverge from the output under Linux. Because the UFormat parameter is supposed to be UNIX format, the output should match that of the date command (i.e. strftime) on Linux.

It's clear that some of the divergences are simple format issues, for instance leading zeros. Others are down to following the ISO 8601 week date specification (https://en.wikipedia.org/wiki/ISO_week_date). And others appear to be differences between locale data on Linux and .Net. In these latter cases it looks like the C locale is followed, which may actually be the best that can be done given the lack of Linux locale data.

# %c

- [x] (PR #4805) The preferred date and time representation for the current locale.

#### PowerShell
```powershell
[System.Threading.Thread]::CurrentThread.CurrentCulture = [System.Globalization.CultureInfo]::GetCultureInfo('en-US') 
get-date -Year 2005 -Month 1 -Day 1 -Hour 1 -Minute 0 -Second 0  -UFormat %c                                                   
```

```none                                                       
Sat Jan  1 01:00:00 2005
```

#### Linux
```bash
LC_TIME="en_US.UTF-8" date -d 2005-01-01T01:00:00 +%c
```
```none                                                       
Sat 01 Jan 2005 01:00:00 AM GMT
```

#### PowerShell
```powershell
[System.Threading.Thread]::CurrentThread.CurrentCulture = [System.Globalization.CultureInfo]::GetCultureInfo('en-US') 
get-date -Year 2005 -Month 1 -Day 1 -Hour 1 -Minute 0 -Second 0  -UFormat %c                                                   
```

```none                                                       
Sat Jan  1 01:00:00 2005
```

#### Linux
```bash
LC_TIME="C.UTF-8" date -d 2005-01-01T01:00:00 +%c
```

```none
Sat Jan  1 01:00:00 2005
```

---
# %g

- [x] (PR #14555) Like %G, but without century, that is, with a 2-digit year (00-99).

#### PowerShell
```powershell
[System.Threading.Thread]::CurrentThread.CurrentCulture = [System.Globalization.CultureInfo]::GetCultureInfo('en-US') 
get-date -Year 2005 -Month 1 -Day 1 -Hour 1 -Minute 0 -Second 0  -UFormat %g                                                     
```

```none                                                       
05
```

#### Linux
```bash
[...]

PR #4769, opened 2017-09-07
Correct %c format in Get-Date -UFormat

iSazonov:
Related #4750.

Add leading zero: now day of the month is from 01 through 31.

Before fix:
```powershell
>Get-Date -UFormat %c
Thu Sep 7 17:49:48 2017
```
After the fix:
```powershell
>Get-Date -UFormat %c
Thu Sep 07 17:49:48 2017
```

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
@@ -369,9 +369,7 @@ namespace Microsoft.PowerShell.Commands
                             break;
 
                         case 'c':
-                            sb.Append("{0:ddd} {0:MMM} ");
-                            sb.Append(StringUtil.Format("{0,2} ", dateTime.Day));
-                            sb.Append("{0:HH}:{0:mm}:{0:ss} {0:yyyy}");
+                            sb.Append("{0:ddd} {0:MMM} {0:dd} {0:HH}:{0:mm}:{0:ss} {0:yyyy}");
                             break;
 
                         case 'D':
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Date.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Date.Tests.ps1
@@ -30,7 +30,7 @@ Describe "Get-Date DRT Unit Tests" -Tags "CI" {
     }
 
     It "using -uformat 'aAbBcCdDehHIjmMpr' produces the correct output" {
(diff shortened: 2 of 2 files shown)


iSazonov:
@timparkinson @mklement0 Could you please review too?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
```diff
@@ -369,9 +369,7 @@ private string UFormatDateString(DateTime dateTime)
                             break;
 
                         case 'c':
-                            sb.Append("{0:ddd} {0:MMM} ");
-                            sb.Append(StringUtil.Format("{0,2} ", dateTime.Day));
-                            sb.Append("{0:HH}:{0:mm}:{0:ss} {0:yyyy}");
+                            sb.Append("{0:ddd} {0:MMM} {0:dd} {0:HH}:{0:mm}:{0:ss} {0:yyyy}");
```
mklement0: Apparently, there are platform differences (not specific to the `date` _utility_ implementation): on macOS 10.12.6 I get space-padded representation (e.g., `<space>7`), on Ubuntu 16.04 I get the `0`-padded representation (e.g., `07`).

iSazonov: On WSL I get space too. And it seems it can be culture-depended (see related Issue). I think we can not be full-compliant with Unix here. So we should follow "common" rule.

mklement0: > we should follow "common" rule.

I'm asking innocently: what do you mean by that?

I just checked the [spec for the POSIX `date` utility](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/date.html), and the definition of `%c` is (a) definitely locale-(culture)-specific and (b) pretty open-ended:

> %c
> Locale's appropriate date and time representation.
> ...
> The %x and %c conversion specifications, however, are intended for local representation; these may be based on a different, non-Gregorian calendar.

iSazonov: Sorry, I meant that "common" - used in most Unix clones.
Thank you for confirmation this is a locale-(culture)-specific. The problem may lie in the fact that we are forced to rely on CoreFX rather than Unix locales. I think that's a good compromise because our goal is to _emulate_ Unix-like format. If we want more we should address this in CoreFX.

iSazonov:
Moved to #4805

PR #4756, opened 2017-09-05
Added WSMan Config provider tests

SteveL-MSFT:
added tests for get-item, get-childitem, set-item, new-item remove-item, clear-item, dynamic parameters
merged CoreClr and FullClr code
did not add tests for clientcert handling

Part of https://github.com/PowerShell/PowerShell/issues/4158

Should get line coverage from 50.39% to 72.2%

Lots of error handling code not hit

PR #4772, opened 2017-09-07
Enable ApartmentState support

### LINKED ISSUE #4350: CLI: -STA and -MTA are mistaken for a -File parameter value
Note: `-STA` and `-MTA` are not supported in Core, but for consistency with Windows PowerShell, they should still be _recognized_, and either an error or a warning should be reported.

[Here](https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs#L715) you can see that checking for `-STA` and `-MTA` is simply _bypassed_ in Core.

Steps to reproduce
------------------

```powershell
powershell -sta
```

Expected behavior
-----------------

An error or a warning about how `-STA` and `-MTA` are not supported in PowerShell Core.

Actual behavior
---------------

`-sta` is misinterpreted as an (implied) `-File` parameter value; an error message is reported, and the session is exited.

```none
The argument '-sta' to the -File parameter does not exist. Provide the path to an existing '.ps1' file as an argument to the -File parameter.
```

Environment data
----------------

```powershell
PowerShell Core v6.0.0-beta.4 on macOS 10.12.5
PowerShell Core v6.0.0-beta.4 on Ubuntu 16.04.2 LTS
PowerShell Core v6.0.0-beta.4 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
```

SteveL-MSFT:
WSManConfigProvider doesn't work once this is enabled if starting powershell with `-sta`, it does work if using `-mta`.  The same/similar WSManConfigProvider code works fine under Windows PowerShell 5.1 with `sta` or `mta`.  We shouldn't have to change the WSManConfigProvider code which relies on underlying non-thread safe COM objects, but need some help debugging this.

Fix #4350

### DIFF
--- a/src/Microsoft.WSMan.Management/CredSSP.cs
+++ b/src/Microsoft.WSMan.Management/CredSSP.cs
@@ -18,9 +18,7 @@ using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using System.Security;
 using System.Threading;
-#if CORECLR
 using System.Xml.XPath;
-#endif
 
 using Dbg = System.Management.Automation;
 
@@ -335,15 +333,6 @@ namespace Microsoft.WSMan.Management
         /// </summary>
         protected override void BeginProcessing()
         {
-#if !CORECLR
-            if (Environment.OSVersion.Version.Major < 6)
-            {
-                //OS is XP/Win2k3. Throw error.
-                WSManHelper helper = new WSManHelper(this);
-                string message = helper.FormatResourceMsgFromResourcetext("CmdletNotAvailable");
-                throw new InvalidOperationException(message);
-            }
-#endif
             //If not running elevated, then throw an "elevation required" error message.
             WSManHelper.ThrowIfNotAdministrator();
 
@@ -453,14 +442,6 @@ namespace Microsoft.WSMan.Management
             //If not running elevated, then throw an "elevation required" error message.
             WSManHelper.ThrowIfNotAdministrator();
             helper = new WSManHelper(this);
-#if !CORECLR
-            if (Environment.OSVersion.Version.Major < 6)
-            {
-                //OS is XP/Win2k3. Throw error.
-                string message = helper.FormatResourceMsgFromResourcetext("CmdletNotAvailable");
-                throw new InvalidOperationException(message);
-            }
-#endif
 
             // DelegateComputer cannot be specified when Role is other than client
             if ((delegatecomputer != null) && !Role.Equals(Client, StringComparison.OrdinalIgnoreCase))
@@ -725,11 +706,7 @@ namespace Microsoft.WSMan.Management
                 //open the registry key.If key is not present,create a new one
                 Credential_Delegation_Key = rootKey.OpenSubKey(Registry_Path_Credentials_Delegation, true);
                 if (Credential_Delegation_Key == null)
-                    Credential_Delegation_Key = rootKey.CreateSubKey(Registry_Path_Credentials_Delegation
-#if !CORECLR
-                        , RegistryKeyPermissionCheck.ReadWriteSubTree
-#endif
-                        );
+                    Credential_Delegation_Key = rootKey.CreateSubKey(Registry_Path_Credentials_Delegation);
 
                 Credential_Delegation_Key.SetValue(helper.Key_Allow_Fresh_Credentials, 1, RegistryValueKind.DWord);
                 Credential_Delegation_Key.SetValue(helper.Key_Concatenate_Defaults_AllowFresh, 1, RegistryValueKind.DWord);
@@ -737,11 +714,7 @@ namespace Microsoft.WSMan.Management
                 // add the delegate value
                 Allow_Fresh_Credential_Key = rootKey.OpenSubKey(Registry_Path_Credentials_Delegation + @"\" + helper.Key_Allow_Fresh_Credentials, true);
                 if (Allow_Fresh_Credential_Key == null)
-                    Allow_Fresh_Credential_Key = rootKey.CreateSubKey(Registry_Path_Credentials_Delegation + @"\" + helper.Key_Allow_Fresh_Credentials
-#if !CORECLR
-                        , RegistryKeyPermissionCheck.ReadWriteSubTree
-#endif
-                        );
+                    Allow_Fresh_Credential_Key = rootKey.CreateSubKey(Registry_Path_Credentials_Delegation + @"\" + helper.Key_Allow_Fresh_Credentials);
 
                 if (Allow_Fresh_Credential_Key != null)
                 {
@@ -890,14 +863,6 @@ namespace Microsoft.WSMan.Management
             WSManHelper.ThrowIfNotAdministrator();
 
             helper = new WSManHelper(this);
-#if !CORECLR
-            if (Environment.OSVersion.Version.Major < 6)
-            {
-                //OS is XP/Win2k3. Throw error.
-                string message = helper.FormatResourceMsgFromResourcetext("CmdletNotAvailable");
-                throw new InvalidOperationException(message);
-            }
-#endif
 
             IWSManSession m_SessionObj = null;
             try
--- a/src/Microsoft.WSMan.Management/Microsoft.WSMan.Management.csproj
+++ b/src/Microsoft.WSMan.Management/Microsoft.WSMan.Management.csproj
@@ -10,6 +10,7 @@
   <ItemGroup>
     <ProjectReference Include="..\System.Management.Automation\System.Management.Automation.csproj" />
     <ProjectReference Include="..\Microsoft.WSMan.Runtime\Microsoft.WSMan.Runtime.csproj" />
+    <PackageReference Include="System.ServiceProcess.ServiceController" Version="4.4.0" />
   </ItemGroup>
 
   <PropertyGroup>
--- a/src/Microsoft.WSMan.Management/WsManHelper.cs
+++ b/src/Microsoft.WSMan.Management/WsManHelper.cs
@@ -17,9 +17,7 @@ using Microsoft.Win32;
 using System.Management.Automation;
 using System.Management.Automation.Provider;
 using System.Threading;
-#if CORECLR
 using System.Xml.XPath;
-#endif
 
 namespace Microsoft.WSMan.Management
 {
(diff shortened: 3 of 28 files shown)


src/System.Management.Automation/engine/remoting/common/WireDataFormat/EncodeAndDecode.cs
```diff
@@ -2401,7 +2387,7 @@ internal static RemoteSessionCapability GetSessionCapability(object data)
                 //
                 // So just return the CurrentTimeZone.
 
-#if !CORECLR // TimeZone Not In CoreCLR
+#if !CORECLR // TimeZone deprecated, need to change to TimeZoneInfo
```
iSazonov: Maybe add that this change affects serialization, remoting and backward compatibility. We can go back to that when we'll have tests for the areas.

SteveL-MSFT: Good suggestion, will add

src/System.Management.Automation/engine/remoting/common/WireDataFormat/EncodeAndDecode.cs
```diff
@@ -2401,7 +2387,7 @@ internal static RemoteSessionCapability GetSessionCapability(object data)
                 //
                 // So just return the CurrentTimeZone.
 
-#if !CORECLR // TimeZone Not In CoreCLR
+#if !CORECLR // TODO: TimeZone deprecated, need to change to TimeZoneInfo.  Affects serialization, remoting, and backwards compatibility.
```
adityapatwardhan: According to documentation here: https://docs.microsoft.com/en-us/dotnet/api/system.timezone?view=netcore-2.0
TimeZone class can still be used to get CurrentTimeZone. A comment about deprecation might not be accurate.

adityapatwardhan (changes requested):

src/System.Management.Automation/engine/remoting/common/WireDataFormat/EncodeAndDecode.cs
SteveL-MSFT: At compilation time, dotnet reported it as deprecated

iSazonov: Confirm.

PaulHigin: @SteveL-MSFT 
What do you mean when you say the WSManConfigProvider doesn't work?  Do you have repro steps and errors?

SteveL-MSFT: @PaulHigin using `sta`, you would get this error:

```none
dir : COM object that has been separated from its underlying RCW cannot be used.
At line:1 char:1
+ dir
+ ~~~
    + CategoryInfo          : NotSpecified: (:) [Get-ChildItem], InvalidComObjectException
    + FullyQualifiedErrorId : System.Runtime.InteropServices.InvalidComObjectException,Microsoft.PowerShell.Commands.GetChildItemCommand
```

src/Microsoft.WSMan.Management/CredSSP.cs
```diff
@@ -725,23 +706,15 @@ private void UpdateGPORegistrySettings(string applicationname, string[] delegate
                 //open the registry key.If key is not present,create a new one
                 Credential_Delegation_Key = rootKey.OpenSubKey(Registry_Path_Credentials_Delegation, true);
                 if (Credential_Delegation_Key == null)
-                    Credential_Delegation_Key = rootKey.CreateSubKey(Registry_Path_Credentials_Delegation
-#if !CORECLR
-                        , RegistryKeyPermissionCheck.ReadWriteSubTree
```
daxian-dbw: `Microsoft.Win32.RegistryKeyPermissionCheck` is available in .NET Core. Maybe we should keep it?

src/Microsoft.WSMan.Management/CredSSP.cs
```diff
@@ -725,23 +706,15 @@ private void UpdateGPORegistrySettings(string applicationname, string[] delegate
...
 
                 Credential_Delegation_Key.SetValue(helper.Key_Allow_Fresh_Credentials, 1, RegistryValueKind.DWord);
                 Credential_Delegation_Key.SetValue(helper.Key_Concatenate_Defaults_AllowFresh, 1, RegistryValueKind.DWord);
 
                 // add the delegate value
                 Allow_Fresh_Credential_Key = rootKey.OpenSubKey(Registry_Path_Credentials_Delegation + @"\" + helper.Key_Allow_Fresh_Credentials, true);
                 if (Allow_Fresh_Credential_Key == null)
-                    Allow_Fresh_Credential_Key = rootKey.CreateSubKey(Registry_Path_Credentials_Delegation + @"\" + helper.Key_Allow_Fresh_Credentials
-#if !CORECLR
-                        , RegistryKeyPermissionCheck.ReadWriteSubTree
```
daxian-dbw: Same here.

src/Microsoft.WSMan.Management/CredSSP.cs
```diff
@@ -890,14 +863,6 @@ protected override void BeginProcessing()
             WSManHelper.ThrowIfNotAdministrator();
 
             helper = new WSManHelper(this);
-#if !CORECLR
-            if (Environment.OSVersion.Version.Major < 6)
-            {
-                //OS is XP/Win2k3. Throw error.
-                string message = helper.FormatResourceMsgFromResourcetext("CmdletNotAvailable");
```
daxian-dbw: Maybe the resource string `CmdletNotAvailable` can be removed?

src/Microsoft.WSMan.Management/Microsoft.WSMan.Management.csproj
```diff
@@ -10,6 +10,7 @@
   <ItemGroup>
     <ProjectReference Include="..\System.Management.Automation\System.Management.Automation.csproj" />
     <ProjectReference Include="..\Microsoft.WSMan.Runtime\Microsoft.WSMan.Runtime.csproj" />
+    <PackageReference Include="System.ServiceProcess.ServiceController" Version="4.4.0" />
```
daxian-dbw: Did you rebase? Some changes in this PR is also in #4756

src/Microsoft.WSMan.Management/WsManHelper.cs
```diff
@@ -1009,17 +1007,11 @@ internal bool ValidateCredSSPRegistry(bool AllowFreshCredentialsValueShouldBePre
             {
                 RegistryKey rGPOLocalMachineKey = Registry.LocalMachine.OpenSubKey(
                     Registry_Path_Credentials_Delegation + @"\CredentialsDelegation",
-#if !CORECLR
-                    RegistryKeyPermissionCheck.ReadWriteSubTree,
```
daxian-dbw: Same here.

src/Microsoft.WSMan.Management/WsManHelper.cs
```diff
@@ -1009,17 +1007,11 @@ internal bool ValidateCredSSPRegistry(bool AllowFreshCredentialsValueShouldBePre
...
-#endif
-                    System.Security.AccessControl.RegistryRights.FullControl);
+                        System.Security.AccessControl.RegistryRights.FullControl);
 
                 if (rGPOLocalMachineKey != null)
                 {
                     rGPOLocalMachineKey = rGPOLocalMachineKey.OpenSubKey(Key_Allow_Fresh_Credentials,
-#if !CORECLR
-                        RegistryKeyPermissionCheck.ReadWriteSubTree,
```
daxian-dbw: ditto

src/Microsoft.WSMan.Management/WsManHelper.cs
```diff
@@ -1078,11 +1070,7 @@ internal static void LoadResourceData()
             try
             {
                 string filepath = System.Environment.ExpandEnvironmentVariables("%Windir%") + "\\System32\\Winrm\\" +
-#if CORECLR
                     "0409" /* TODO: don't assume it is always English on CSS? */
-#else
-                    String.Concat("0", String.Format(CultureInfo.CurrentCulture, "{0:x2}", checked((uint)CultureInfo.CurrentUICulture.LCID)))
```
daxian-dbw: Shall we use the `#else` block instead?

src/System.Management.Automation/engine/hostifaces/LocalPipeline.cs
```diff
@@ -247,55 +242,18 @@ private void SetupInvokeThread(Thread invokeThread, bool changeName)
...
         {
             get
             {
-                int i = ReadRegistryInt("PipelineMaxStackSizeMB", 10);
-                if (i < 10)
-                    i = 10; // minimum 10MB
-                else if (i > 100)
-                    i = 100; // maximum 100MB
-                return i * 1000000;
+                // TODO: move this to startupconfig is needed.  This number is in bytes.
```
daxian-dbw: Can you add one more comment indicating we use 10MB by default?

src/System.Management.Automation/engine/hostifaces/LocalPipeline.cs
```diff
@@ -1233,15 +1189,6 @@ internal class PipelineThread : IDisposable
...
-#if CORECLR
-        internal PipelineThread()
-        {
-            _worker = new Thread(WorkerProc);
-            _workItem = null;
-            _workItemReady = new AutoResetEvent(false);
-            _closed = false;
-        }
-#else
         internal PipelineThread(ApartmentState apartmentState)
```
daxian-dbw: Is it possible that this method can be called on Unix plats? If so, `_worker.SetApartmentState(apartmentState)` might throw.

daxian-dbw (changes requested):

SteveL-MSFT:
@daxian-dbw I've been thinking that I've been looking at this problem the wrong way.  ApartmentState is only used for COM interop and dotnet will marshal between STA and MTA so the only loss is some perf impact.  I'm thinking that the right thing to do here is to remove support for ApartmentState.  Thoughts?

daxian-dbw:
> and dotnet will marshal between STA and MTA

@SteveL-MSFT I'm not very clear about the quoted sentence. Can you please elaborate a bit on it?

My understanding is that STA is for proper calls to COM object with the single-thread apartment (mainly about COM objects related to the user interface, for example, windows forms). Windows form may not be a problem as it's not supported in .NET Core. However, it may still be possible that user gets a STA COM object by `New-Object` and operate on it, so it might be affected when removing the `Apartment` support from powershell core.

A question, if we remove `Apartment` support, what would be the default thread `apartment` for powershell?

I want to point out that on Windows PowerShell, users are able to specify the `ApartmentState` in `PSInvocationSettings` and pass the setting to `powershell.Invoke(...)`. Removing `ApartmentState` support in PS Core is technically a breaking change, so it probably needs to be reviewed by the Committee.

SteveL-MSFT:
@daxian-dbw see [Interop Marshaling](https://docs.microsoft.com/en-us/dotnet/framework/interop/interop-marshaling#marshaling_and_com_apartments)

I'll close this for now and open a new issue regarding removing ApartmentState support and will discuss with Committee

PR #4775, opened 2017-09-07
Add a generic file watcher function in HelpersCommon.psm1

### LINKED ISSUE #4524: Add a generic file watcher function in HelpersCommon.psm1
See https://github.com/PowerShell/PowerShell/pull/4523#discussion_r131796738

Polling for a file can be a common pattern. Consider implementing a generic file watcher function in HelpersCommon.psm1.
```
while (((Get-Date) - $startTime).TotalSeconds -lt 10 -and (-not (Test-Path "$HOME/nativeCommandProcessor.Success"))) {
    Start-Sleep -Milliseconds 100
}
```

daxian-dbw:
Fix #4524 

Add a generic file watcher function in HelpersCommon.psm1 and update some tests to use the `Wait-FileToBePresent`.


test/tools/Modules/HelpersCommon/HelpersCommon.psm1
```diff
@@ -11,14 +11,35 @@ function Wait-UntilTrue
 
     # Loop until the script block evaluates to true
     while (-not ($sb.Invoke())) {
+        # If the timeout period has passed, return false
```
iSazonov: Interesting, make sense to enhance `Invoke()` to support timeout?

test/tools/Modules/HelpersCommon/HelpersCommon.psm1
```diff
@@ -11,14 +11,35 @@ function Wait-UntilTrue
...
+    param (
+        [string]$File,
+        [int]$TimeoutInSeconds = 10,
+        [int]$IntervalInMilliseconds = 100
+    )
+    # Get the current time
+    $startTime = [DateTime]::Now
+
+    # Loop until the file is present
+    while (-not (Test-Path $File)) {
```
iSazonov: Maybe use `Wait-UntilTrue`? We duplicate code here.

test/powershell/Language/Scripting/NativeExecution/NativeCommandProcessor.Tests.ps1
```diff
@@ -196,19 +196,13 @@ Categories=Application;
         elseif ($IsLinux) {
             # Validate on Linux by reassociating default app for text file
             & $TestFile
-            $startTime = Get-Date
             # It may take time for handler to start
-            while (((Get-Date) - $startTime).TotalSeconds -lt 10 -and (-not (Test-Path "$HOME/nativeCommandProcessor.Success"))) {
-                Start-Sleep -Milliseconds 100
-            }
+            Wait-FileToBePresent -File "$HOME/nativeCommandProcessor.Success" -TimeoutInSeconds 10 -IntervalInMilliseconds 100 > $null
```
iSazonov: Every use we add ">$null" - so maybe do this internally in Wait-FileToBePresent ? If we need we could add `-Passthrou`
And why not use Out-Null? It is optimized for pipelines.

test/tools/Modules/HelpersCommon/HelpersCommon.psm1
daxian-dbw: You can do async invocation via powershell API, but the experience is not optimal.

test/tools/Modules/HelpersCommon/HelpersCommon.psm1
daxian-dbw: Good point. I will change it.

test/powershell/Language/Scripting/NativeExecution/NativeCommandProcessor.Tests.ps1
daxian-dbw: Will change to ignore output in `Wait-FileToBePresent`. `-PassThru` can be added when there is a need coming up.

As for Out-Null, the optimization is to make `Out-Null` and same as `> $null` when it's the last command in a pipeline, namely write to null pipe directly without going through the cmdlet.

daxian-dbw:
The failure in AppVeyor is a known issue. It's fixed by #4806

iSazonov (approved):

Issue #4154, opened 2017-07-01
Add tests for New-Object -ArgumentList and -ComObject


adityapatwardhan:
Add tests for New-Object 
- [x] -ArgumentList and 
- [x] -ComObject
    - [ ] - STA vs MTA

Related code coverage report: 
https://codecov.io/gh/PowerShell/PowerShell/src/9e41c647a9407e05d86f4971a6bd25719a7130d0/src/Microsoft.PowerShell.Commands.Utility/commands/utility/new-object.cs

Environment data
----------------

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.3
OS                             Microsoft Windows 10.0.16204
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

```

chunqingchen:
@adityapatwardhan I think this issue can be closed as all the scenarios are covered.

PR #4776, opened 2017-09-07
Add tests for New-Object -com

TravisEz13:
addresses #4154


daxian-dbw:
`new-object -ComObject` has been tested in [COM.Basic.Tests.ps1](https://github.com/PowerShell/PowerShell/blob/master/test/powershell/engine/COM/COM.Basic.Tests.ps1). Maybe it makes sense to move these 2 tests to that file (in `basic com tests`) as well?

One more thing, the COM test should be skipped for nano server and IoT as well. Please see how it's done in COM.basic.Tests.ps1.

TravisEz13:
Moving the New-Object tests to the engine tests makes it hard to run all the new-object tests together.  We have this structure for a reason.  I think we should stick with it.

daxian-dbw (approved):

Issue #2112, opened 2016-08-28
Simplify usage of Multipart Uploads with WebRequestPSCmdlet


ffeldhaus:
Doing multipart requests with PowerShell is quite complicated and requires several extra steps for large file uploads. [An easy example for a small file can be found on StackOverflow](http://stackoverflow.com/questions/36268925/powershell-invoke-restmethod-multipart-form-data). [An in depth discussion can be found in this blog post](http://blog.majcica.com/2016/01/13/powershell-tips-and-tricks-multipartform-data-requests/). 

It would be a huge improvement if the WebRequestPSCmdlets (Invoke-RestMethod and Invoke-WebRequest) could be enhanced so that they support Multipart messages directly.

For an implementation I would expect the following parameters:
- MultipartFile (Path)
- MultipartName (Name to be used in Multipart message)

[.NET Core seems to have support for MultipartContent](https://github.com/dotnet/corefx/blob/d0dc5fc099946adc1035b34a8b1f6042eddb0c75/src/System.Net.Http/src/System/Net/Http/MultipartContent.cs) which may simplify the implementation.

swinster:
This is horrendous in Powershell and hasn't moved in an age. I personally would love to see some movement here, as uploading large file is very, very difficult. 

As some pointer around the topic see my comments here (albeit for PS 5.1) which like to some other sites, namely http://blog.majcica.com/2016/01/13/powershell-tips-and-tricks-multipartform-data-requests/

iSazonov:
@markekraus Could you please look this if you have free time?

markekraus:
@iSazonov sure. I will take a look at it. I have seen other requests in the past few years for multipart forms in general as several IoT APIs seem to only work with multipart forms.

iSazonov:
Feel free to ping MSFT experts here if there is compatibility issues.

swinster:
FWIW, we mainly Python, Django and Apache for our Web interfaace stuff. This is all RESTfull APIs. I am not developer, just attempting to interact with the API as we are seeing more sysadmins wanting to do this.

markekraus:
@swinster if you could provide me an exact scenario that would be very useful. Right now, all I have to go off of is to build a out a test API that takes multipart forms and files and hope that fits most scenarios.

swinster:
[...]

PR #4782, opened 2017-09-08
Add Multipart Support to Web Cmdlets

markekraus:
Summary
----------
Partially implements  #2112

- Adds `System.Net.Http.MultipartFormDataContent` as a possible type for `-Body`
- Adds `/Multipart/` test to WebListener 

This allows for the user to create their own `Http.MultipartFormDataContent` object and submit it. Since `multipart/form-data` submissions are highly flexible, adding direct support for it to the CmdLets may over-complicate the command parameters and a limited implementation would not address the broad scope of use cases. This at least allows the user to submit multipart forms using the Web Cmdlets and not have to manage their own `HttpClient`. Once this is introduced, limited multipart implementations can be expanded to use the code in this PR.

Documentation Needed
--------------------------
Note that documentation will probably need to make clear that `-ContentType` will be ignored when a `MultipartFormDataContent` object is supplied to `-Body`. Similarly, content headers specified by `-Headers` and `-WebSession` will be ignored too.

`ContentType` must be set by the `MultipartFormDataContent` object as it controls the `boundary` in the `Content-Type` header used to distinguish between the various fields supplied. It doesn't make sense to use any other `Content-Type` header when doing `multipart/form-data` anyway.  If this isn't done here, the Foreach on line 421 throws. when `Content` is set with a `MultipartFormDataContent` it overwrites all the content headers, so trying to add content headers again causes an exception about `ContentType` already exists.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
@@ -389,10 +389,9 @@ namespace Microsoft.PowerShell.Commands
                     byte[] bytes = content as byte[];
                     SetRequestContent(request, bytes);
                 }
-                else if (content is MultipartFormDataContent)
+                else if (content is MultipartFormDataContent multipartFormDataContent)
                 {
                     WebSession.ContentHeaders.Clear();
-                    MultipartFormDataContent multipartFormDataContent = content as MultipartFormDataContent;
                     SetRequestContent(request, multipartFormDataContent);
                 }
                 else
@@ -806,9 +805,13 @@ namespace Microsoft.PowerShell.Commands
         internal long SetRequestContent(HttpRequestMessage request, MultipartFormDataContent multipartContent)
         {
             if (request == null)
+            {
                 throw new ArgumentNullException("request");
+            }
             if (multipartContent == null)
+            {
                 throw new ArgumentNullException("multipartContent");
+            }
 
             request.Content = multipartContent;
 
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
@@ -389,6 +389,12 @@ namespace Microsoft.PowerShell.Commands
                     byte[] bytes = content as byte[];
                     SetRequestContent(request, bytes);
                 }
+                else if (content is MultipartFormDataContent)
+                {
+                    WebSession.ContentHeaders.Clear();
+                    MultipartFormDataContent multipartFormDataContent = content as MultipartFormDataContent;
+                    SetRequestContent(request, multipartFormDataContent);
+                }
                 else
                 {
                     SetRequestContent(request,
@@ -785,6 +791,28 @@ namespace Microsoft.PowerShell.Commands
             request.Content = streamContent;
 
             return streamContent.Headers.ContentLength.Value;
+        }
+
+        /// <summary>
+        /// Sets the ContentLength property of the request and writes the specified content to the request's RequestStream.
+        /// </summary>
+        /// <param name="request">The WebRequest who's content is to be set</param>
+        /// <param name="multipartContent">A MultipartFormDataContent object containing multipart/form-data content.</param>
+        /// <returns>The number of bytes written to the requests RequestStream (and the new value of the request's ContentLength property</returns>
+        /// <remarks>
+        /// Because this function sets the request's ContentLength property and writes content data into the requests's stream,
+        /// it should be called one time maximum on a given request.
+        /// </remarks>
+        internal long SetRequestContent(HttpRequestMessage request, MultipartFormDataContent multipartContent)
+        {
+            if (request == null)
+                throw new ArgumentNullException("request");
+            if (multipartContent == null)
+                throw new ArgumentNullException("multipartContent");
+
+            request.Content = multipartContent;
+
+            return multipartContent.Headers.ContentLength.Value;
         }
 
         internal long SetRequestContent(HttpRequestMessage request, IDictionary content)
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -388,7 +388,7 @@ function GetMultipartBody
         $fileContent.Headers.ContentType = [System.Net.Http.Headers.MediaTypeHeaderValue]::Parse("text/plain")
         $multipartContent.Add($fileContent)
     }
-    # unary comma required
+    # unary comma required to prevent $multipartContent from being unwrapped/enumerated
     return ,$multipartContent 
 }
 
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -359,6 +359,39 @@ function ExecuteRestMethod
     return $result
 }
 
+function GetMultipartBody 
+{
+    param
+    (
+        [Switch]$String,
+        [Switch]$File
+    )
+    $multipartContent = [System.Net.Http.MultipartFormDataContent]::new()
+    if ($String.IsPresent)
+    {
+        $stringHeader = [System.Net.Http.Headers.ContentDispositionHeaderValue]::new("form-data")
+        $stringHeader.Name = "TestString"
+        $StringContent = [System.Net.Http.StringContent]::new("TestValue")
+        $StringContent.Headers.ContentDisposition = $stringHeader
+        $multipartContent.Add($stringContent)
+    }
+    if ($File.IsPresent)
+    {
+        $multipartFile = Join-Path $TestDrive 'multipart.txt'
+        "TestContent" | Set-Content $multipartFile
+        $FileStream = [System.IO.FileStream]::new($multipartFile, [System.IO.FileMode]::Open)
+        $fileHeader = [System.Net.Http.Headers.ContentDispositionHeaderValue]::new("form-data")
+        $fileHeader.Name = "TestFile"
+        $fileHeader.FileName = 'multipart.txt'
+        $fileContent = [System.Net.Http.StreamContent]::new($FileStream)
+        $fileContent.Headers.ContentDisposition = $fileHeader
+        $fileContent.Headers.ContentType = [System.Net.Http.Headers.MediaTypeHeaderValue]::Parse("text/plain")
+        $multipartContent.Add($fileContent)
+    }
+    # unary comma required
+    return ,$multipartContent 
+}
+
 <#
     Defines the list of redirect codes to test as well as the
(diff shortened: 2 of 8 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -389,6 +389,12 @@ internal virtual void FillRequestStream(HttpRequestMessage request)
                     byte[] bytes = content as byte[];
                     SetRequestContent(request, bytes);
                 }
+                else if (content is MultipartFormDataContent)
+                {
+                    WebSession.ContentHeaders.Clear();
+                    MultipartFormDataContent multipartFormDataContent = content as MultipartFormDataContent;
```
iSazonov: We should exclude double casting. Here it is better use C# 7.0 pattern:
```c#
else if (content is MultipartFormDataContent multipartFormDataContent ) 
```

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -787,6 +793,28 @@ internal long SetRequestContent(HttpRequestMessage request, Stream contentStream
...
+        /// <remarks>
+        /// Because this function sets the request's ContentLength property and writes content data into the requests's stream,
+        /// it should be called one time maximum on a given request.
+        /// </remarks>
+        internal long SetRequestContent(HttpRequestMessage request, MultipartFormDataContent multipartContent)
+        {
+            if (request == null)
+                throw new ArgumentNullException("request");
+            if (multipartContent == null)
+                throw new ArgumentNullException("multipartContent");
```
iSazonov: We should use curly braces:
```c#
if ( ... )
{
    ...
}
```
or single line operator.

test/tools/WebListener/Controllers/MultipartController.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        private IHostingEnvironment _environment;
+
+        public MultipartController(IHostingEnvironment environment)
+        {
+            _environment = environment;
+        }
+        public ActionResult Index()
+        {
+
```
iSazonov: Please remove extra line.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
markekraus: I can do this, but it will not match the surrounding code.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
markekraus: Awesome. I will do that.

test/tools/WebListener/Controllers/MultipartController.cs
markekraus: will do

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
iSazonov: I believe we should use approved patterns for new code.
You can wait a view of other maintainers.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1200,6 +1233,41 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         }
     }
 
+    Context "Multipart/form-data Tests" {
+        It "Verifies Invoke-WebRequest Supports Multipart String Values" {
```
iSazonov: We use the same Context header 😕 
I believe we should add cmdlet name `Context "Invoke-WebRequest Multipart/form-data Tests"`

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1737,6 +1805,38 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
         }
     }
 
+    Context "Multipart/form-data Tests" {
+        It "Verifies Invoke-RestMethod Supports Multipart String Values" {
```
iSazonov: Context "Invoke-RestMethod Multipart/form-data Tests"

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: Yes, because they are under different describe blocks. We did the same thing with HTTPS

iSazonov: I think it's somewhat confusing, but we can leave it for Context.

markekraus: For clarity, would you like me to change it or leave it as is?

iSazonov: Leave it as is.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
daxian-dbw: I agreed with @iSazonov. The new code is somewhat isolated as it's in its own method, so I think it's fine to use the more preferred style.

markekraus: @daxian-dbw Awesome! format is already corrected in the most recent commit for this PR.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -389,6 +389,11 @@ internal virtual void FillRequestStream(HttpRequestMessage request)
                     byte[] bytes = content as byte[];
                     SetRequestContent(request, bytes);
                 }
+                else if (content is MultipartFormDataContent multipartFormDataContent)
+                {
+                    WebSession.ContentHeaders.Clear();
```
daxian-dbw: `WebSession.ContentHeader` is set at line 325 when `ContentType` is specified. Is it OK to ignore that as well?

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -359,6 +359,39 @@ function ExecuteRestMethod
...
+        $FileStream = [System.IO.FileStream]::new($multipartFile, [System.IO.FileMode]::Open)
+        $fileHeader = [System.Net.Http.Headers.ContentDispositionHeaderValue]::new("form-data")
+        $fileHeader.Name = "TestFile"
+        $fileHeader.FileName = 'multipart.txt'
+        $fileContent = [System.Net.Http.StreamContent]::new($FileStream)
+        $fileContent.Headers.ContentDisposition = $fileHeader
+        $fileContent.Headers.ContentType = [System.Net.Http.Headers.MediaTypeHeaderValue]::Parse("text/plain")
+        $multipartContent.Add($fileContent)
+    }
+    # unary comma required
```
daxian-dbw: nit: can you please also mention why it's required -- `$multipartContent` will be unwrapped/enumerated otherwise.

daxian-dbw (commented):
@markekraus Thank you for the good work! Could you please take a look at the 2 comments I left?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
markekraus: Yes, `ContentType` must be set by the `MultipartFormDataContent` object as it controls the `boundary` in the `Content-Type` header used to distinguish between the various fields supplied. Documentation will probably need to make clear that `-ContentType` will be ignored when a `MultipartFormDataContent` object is supplied. It doesn't make sense to use any other `Content-Type` header when doing `multipart/form-data` anyway.  If this isn't done here, the Foreach on line 421 throws. when `Content` is set with a `MultipartFormDataContent` it overwrites all the content headers, so trying to add content headers again causes an exception about `ContentType` already existing.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
daxian-dbw: Thanks for the clarification. I updated the PR description to capture your words.

Then there is potential problem at line 239:
```
if (WebSession.Headers.Count > 0)
{
    WebSession.ContentHeaders.Clear();
    foreach (var entry in WebSession.Headers)
    {
        if (HttpKnownHeaderNames.ContentHeaders.Contains(entry.Key))
        {
            WebSession.ContentHeaders.Add(entry.Key, entry.Value);
        }
...
```
If `-Headers` or `-WebSession` is specified, then the `foreach` on line 421 would throw.

daxian-dbw (approved):
LGTM

iSazonov (approved):

daxian-dbw:
@markekraus Thanks again for the good work!

markekraus:
@daxian-dbw For documentation, when is that handled? Should I submit a PR/Issue to PowerShell-Docs now or does it need to wait for the next beta release?

daxian-dbw:
Yes, you are welcome to submit a PR to [PowerShell-Docs](https://github.com/PowerShell/PowerShell-Docs) now :)
Here is the guideline for maintainers about PRs that require doc changes (see [CONTRIBUTING.md](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)):
> If your change warrants an update to user-facing documentation, a Maintainer will add the Documentation Needed label to your PR and add an issue to the PowerShell-Docs repository, so that we make sure to update the official documentation to reflect your contribution. As an example, this requirement includes any changes to cmdlets (including cmdlet parameters) and features which have associated about_* topics. While not required, we appreciate any contributors who add this label and create the issue themselves. Even better, all contributors are free to contribute the documentation themselves. (See Contributing to documentation related to PowerShell for more info.)

Please see https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#contributing-to-documentation for more details

PR #4788, opened 2017-09-08
Mitigate intermittent failures in access denied tests.

### LINKED ISSUE #4561: Basic filesystem tests fail intermittently
Example:
https://ci.appveyor.com/project/PowerShell/powershell/build/6.0.0-beta.5-4620/tests

```
Basic FileSystem Provider Tests.Validate behavior when access is denied.Access-denied test for 'Move-Item -Path C:\Windows\appcompat\Programs -Destination bar
Pester 442 ms
Error message:
Expected string length 70 but was 65. Strings differ at index 0.
Expected: {MoveDirectoryItemIOError,Microsoft.PowerShell.Commands.MoveItemCommand}
But was:  {RenameItemIOError,Microsoft.PowerShell.Commands.RenameItemCommand}
-----------^
Error stack trace:
at line: 241 in C:\projects\powershell\test\powershell\Modules\Microsoft.PowerShell.Management\FileSystem.Tests.ps1
241:             $err | Should Be $expectedError
```

dantraMSFT:
Fix https://github.com/PowerShell/PowerShell/issues/4561

This change mitigates intermittent timeout failures in the access denied tests as well as fixes unintended cross-test dependencies.  If PowerShell takes longer that 10 seconds to complete a race can cause the subsequent test to pick up the output file of the previous test and fail (i.e., the expected output error file already exists from the previous test because it was created by the long running PowerShell process after the second test started.)  Worse case, the failure cascades causing the race to propogate through the remaining tests in the set.

The fix is two fold:
1: Increase the timeout for waiting for the launched powershell process. (mitigation)
2: Ensure each uses a unique name for the error and done files to avoid polluting the next test.

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Management/FileSystem.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Management/FileSystem.Tests.ps1
@@ -213,20 +213,20 @@ Describe "Basic FileSystem Provider Tests" -Tags "CI" {
         }
 
         It "Access-denied test for '<cmdline>" -Skip:(-not $IsWindows) -TestCases @(
-            # NOTE: ensure the testId parameter is unique for each test case; it is used to generate a unique error and done file name.
-            @{cmdline = "Get-Item $protectedPath2"; expectedError = "ItemExistsUnauthorizedAccessError,Microsoft.PowerShell.Commands.GetItemCommand"; testId = 'GetItem'}
-            @{cmdline = "Get-ChildItem $protectedPath"; expectedError = "DirUnauthorizedAccessError,Microsoft.PowerShell.Commands.GetChildItemCommand"; testId = 'GetChildItem'}
-            @{cmdline = "New-Item -Type File -Path $newItemPath"; expectedError = "NewItemUnauthorizedAccessError,Microsoft.PowerShell.Commands.NewItemCommand"; testId = 'NewItem'}
-            @{cmdline = "Rename-Item -Path $protectedPath -NewName bar"; expectedError = "RenameItemIOError,Microsoft.PowerShell.Commands.RenameItemCommand"; testId = 'RenameItem'},
-            @{cmdline = "Move-Item -Path $protectedPath -Destination bar"; expectedError = "MoveDirectoryItemIOError,Microsoft.PowerShell.Commands.MoveItemCommand";  testId = 'MoveItem'},
-            @{cmdline = "Remove-Item -Path $protectedPath"; expectedError = "RemoveItemUnauthorizedAccessError,Microsoft.PowerShell.Commands.RemoveItemCommand"; testId = 'RemoveItem'}
+            # NOTE: ensure the fileNameBase parameter is unique for each test case; it is used to generate a unique error and done file name.
+            @{cmdline = "Get-Item $protectedPath2"; expectedError = "ItemExistsUnauthorizedAccessError,Microsoft.PowerShell.Commands.GetItemCommand"; fileNameBase = 'GetItem'}
+            @{cmdline = "Get-ChildItem $protectedPath"; expectedError = "DirUnauthorizedAccessError,Microsoft.PowerShell.Commands.GetChildItemCommand"; fileNameBase = 'GetChildItem'}
+            @{cmdline = "New-Item -Type File -Path $newItemPath"; expectedError = "NewItemUnauthorizedAccessError,Microsoft.PowerShell.Commands.NewItemCommand"; fileNameBase = 'NewItem'}
+            @{cmdline = "Rename-Item -Path $protectedPath -NewName bar"; expectedError = "RenameItemIOError,Microsoft.PowerShell.Commands.RenameItemCommand"; fileNameBase = 'RenameItem'},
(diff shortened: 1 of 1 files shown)


test/powershell/Modules/Microsoft.PowerShell.Management/FileSystem.Tests.ps1
```diff
@@ -210,27 +210,27 @@ Describe "Basic FileSystem Provider Tests" -Tags "CI" {
...
+            @{cmdline = "Get-ChildItem $protectedPath"; expectedError = "DirUnauthorizedAccessError,Microsoft.PowerShell.Commands.GetChildItemCommand"; testId = 'GetChildItem'}
+            @{cmdline = "New-Item -Type File -Path $newItemPath"; expectedError = "NewItemUnauthorizedAccessError,Microsoft.PowerShell.Commands.NewItemCommand"; testId = 'NewItem'}
+            @{cmdline = "Rename-Item -Path $protectedPath -NewName bar"; expectedError = "RenameItemIOError,Microsoft.PowerShell.Commands.RenameItemCommand"; testId = 'RenameItem'},
+            @{cmdline = "Move-Item -Path $protectedPath -Destination bar"; expectedError = "MoveDirectoryItemIOError,Microsoft.PowerShell.Commands.MoveItemCommand";  testId = 'MoveItem'},
+            @{cmdline = "Remove-Item -Path $protectedPath"; expectedError = "RemoveItemUnauthorizedAccessError,Microsoft.PowerShell.Commands.RemoveItemCommand"; testId = 'RemoveItem'}
         ) {
-            param ($cmdline, $expectedError)
+            param ($cmdline, $expectedError, $testId)
```
SteveL-MSFT: Rather than adding a TestId param, you can either just use `cmdline` or use $cmdline.split(' ')[0] to get the same as what you have.  Using the cmdline may be better if new test cases for the same command get added in the future.

SteveL-MSFT: Also, line 238 to check existence of $errfile should be removed since it'll always exist since you're creating it before the test.

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Management/FileSystem.Tests.ps1
dantraMSFT: I considered using split but it won't solve the problem; splitting on space gives the cmdlet name. I also considered using the expected error id but that could end up colliding as well.  The test id was the most straight forward solution. I'll add a comment to the test cases to call the unique requirement out specifically.

As to the errFile test, I'll remove it.

dantraMSFT: @SteveL-MSFT Does that work for you?

SteveL-MSFT: I think it would be simpler to just use the entirety of $cmdline in place of $testid (just put it in quotes on line 229).

dantraMSFT: That's not reliable; there could be characters in the command-line that are not valid for a file name. For example, the '|' is not a valid filename character.

dantraMSFT:
@SteveL-MSFT How does this look to you?

test/powershell/Modules/Microsoft.PowerShell.Management/FileSystem.Tests.ps1
dantraMSFT: The testid parameter is used to generate the test-specific error and done file names. This is being done to address the issue of the launched PowerShell process taking longer than the timeout to shutdown. As observed in the nightly build, this can cause a race between the current test and next test if the PowerShell process doesn't exit before the next test starts.

I've changed the parameter name to fileNameBase for clarity.

SteveL-MSFT:
@dantraMSFT I think you misunderstood my feedback.  You don't need a different errfile for each variation of the test.  The important part is having the context in the contents so when it fails, the engineer knows which variation it failed on and why.  My suggestion is to get rid of $fileBaseName, it's unncessary.  Just put the contents of $cmdline into $errFile.  Since the validation checks the contents of $errFile and expects $expectedError, it won't match and you'll see the context message in $errFile

dantraMSFT:
I'm closing this to use a different approach to handling the timeout issue.

dantraMSFT:
Discussed offline with Steve; will leave the test-specific error filename but generate it from a hash of $cmdline instead of requiring it's definition in the test case data.

SteveL-MSFT (approved):

daxian-dbw:
@dantraMSFT Does the PR description needs to be updated based on your offline discussion with Steve?

dantraMSFT:
@daxian-dbw: no, the gist was to avoid the extra parameter.  We consider attempting to use start-process but that proved a dead end.

daxian-dbw:
@dantraMSFT thanks for the clarification.

Issue #4784, opened 2017-09-08
CredSSP.cs - Unhandled exception in DeleteUserDelegateSettings calling GPO.OpenLocalMachineGPO(1)


dantraMSFT:
Steps to reproduce
------------------
running full CI/FEATURE tests on my server datacenter Selfhost OS.

Expected behavior
-----------------
All tests pass

Actual behavior
---------------
'Disable-WSManCredSSP works' test fails with unhandled COM exception. PowerShell process exits with -532462766 (0xe0434352 EXCEPTION_COMPLUS)

This repro's consistently interactively running Disable-WSManCredSSP -Role Client. The below callstack is reported, a delay occurs, and the process exits.

Unhandled Exception: System.Runtime.InteropServices.COMException: Error HRESULT E_FAIL has been returned from a call to a COM component.
   at Microsoft.WSMan.Management.IGroupPolicyObject.OpenLocalMachineGPO(UInt32 dwFlags)
   at Microsoft.WSMan.Management.DisableWSManCredSSPCommand.DeleteUserDelegateSettings() in C:\git\Microsoft\powershell\src\Microsoft.WSMan.Management\CredSSP.cs:line 235
   at System.Threading.Thread.ThreadMain_ThreadStart()
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)

Environment data
----------------
Server DataCenter 2016

> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      5.1.16357.1000
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.16357.1000
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1

SteveL-MSFT:
This isn't repro'ing on
```powershell
Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.6
OS                             Microsoft Windows 10.0.16350
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

PR #4789, opened 2017-09-08
Disable Enable-WSManCredSSP and Disable-WSManCredSSP tests

dantraMSFT:
Disabling due to Issue https://github.com/powershell/powershell/issues/4784

### DIFF
--- a/test/powershell/Modules/Microsoft.WSMan.Management/CredSSP.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.WSMan.Management/CredSSP.Tests.ps1
@@ -10,7 +10,7 @@ Describe "CredSSP cmdlet tests" -Tags 'Feature','RequireAdminOnWindows' {
         {
             $PSDefaultParameterValues["it:skip"] = $true
         }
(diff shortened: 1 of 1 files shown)


SteveL-MSFT:
You have a bad link to issue https://github.com/powershell/powershell/issues/4784

This isn't repro'ing for me

dantraMSFT:
It's possible its a server + domain joined issue. I haven't investigated it but it fails on my system consistently.
The issue is not that it is failing on my system but that we are not handling the COM exception and the process is getting shut down.

SteveL-MSFT:
Agree that the real issue is that we aren't handling the COM exception.  However, this PR doesn't address that.

SteveL-MSFT:
@dantraMSFT as we discussed, you should close this PR

SteveL-MSFT (changes requested):
Doesn't resolve the actual issue and problem seems to only repro on your laptop

dantraMSFT:
Closing this PR; will work on a fix for the underlying issue, instead.

PR #4798, opened 2017-09-10
Replace httpbin.org/user-agent with WebListener

markekraus:
Replaces all calls to httpbin.org/user-agent with WebListener Get test.

reference #2504

PR #4799, opened 2017-09-10
Replace httpbin.org/headers with WebListener Get

markekraus:
Replaces calls to httpbin.org/headers with WebListener Get.

reference #2504

(side note: I'm not sure what the point is of these vs the ones in #4798. I think a separate issue needs to be worked to streamline these tests as there appears to be many that perform the same exact checks against slightly different URLs without much difference in behavior or expected outcome).


markekraus:
Test fails due to #4800

iSazonov (approved):

TravisEz13 (approved):

Issue #4800, opened 2017-09-10
Set-Service.Tests.ps1 tests Failing builds


markekraus:
#4785 Introduced tests which are failing builds.

On Windows:

```
Description: New-Service with bad parameters will fail for 'badstarttype' where 'StartupType' = 'System'
Name:        Set/New-Service cmdlet tests..New-Service with bad parameters will fail for 'badstarttype' where 'StartupType' = 'System'
message:
Expected string length 66 but was 13. Strings differ at index 0.
Expected: {CouldNotNewService,Microsoft.PowerShell.Commands.NewServiceCommand}
But was:  {No Exception!}
-----------^
stack-trace:
at line: 65 in C:\projects\powershell\test\tools\Modules\HelpersCommon\HelpersCommon.psm1
65:             $_.FullyQualifiedErrorId | Should Be $FullyQualifiedErrorId | Out-Null
```

On Linux and macOS:

```
[-] Error occurred in Describe block 79ms

   The term 'Get-Service' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

   At /home/travis/build/PowerShell/PowerShell/test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1:25 char:48
```

`System` is a valid `System.ServiceProcess.ServiceStartMode` so either [the test](https://github.com/PowerShell/PowerShell/blob/2761c3ff210caf7a9c18968c03c18c6921046251/test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1#L165) is flawed or there is an issue with `New-Service`.

The Linux/macOS failures are due to `Get-Service` being called in the TestCases [HashTable literal](https://github.com/PowerShell/PowerShell/blob/2761c3ff210caf7a9c18968c03c18c6921046251/test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1#L25). That literal gets parsed regardless of the OS skip resulting in calls to `Get-Service`  on OSE's where it is not present.

iSazonov:
@markekraus Thanks for your report - we skipped `[Feature]` in last commits 😕

markekraus:
@iSazonov There are some other PRs "in the pipe" that look like they affect feature tests but don't have any CI runs with `[feature]`

[...]

PR #4802, opened 2017-09-10
fix set-service failing test

SteveL-MSFT:
missed -ErrorAction Stop
changed new-service to return error when given unsupported startuptype

Fix https://github.com/PowerShell/PowerShell/issues/4800
Fix https://github.com/PowerShell/PowerShell/issues/4803

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
@@ -2015,10 +2015,11 @@ namespace Microsoft.PowerShell.Commands
                         dwStartType = NativeMethods.SERVICE_DISABLED;
                         break;
                     default:
-                        Diagnostics.Assert(
-                            ((ServiceStartMode)(-1)) == StartupType,
-                            "bad StartupType");
-                        break;
+                        WriteNonTerminatingError(StartupType.ToString(), DisplayName, Name,
+                            new ArgumentException(), "CouldNotNewService",
+                            ServiceResources.UnsupportedStartupType,
+                            ErrorCategory.InvalidArgument);
+                        return;
                 }
 
                 // set up the double-null-terminated lpDependencies parameter
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
@@ -1511,8 +1511,6 @@ namespace Microsoft.PowerShell.Commands
         internal ServiceStartMode startupType = (ServiceStartMode)(-1);
 
 
-
-
         /// <summary>
         /// The following is the definition of the input parameter "Status".
         /// This specifies what state the service should be in (e.g. Running, Stopped,
@@ -1684,22 +1682,14 @@ namespace Microsoft.PowerShell.Commands
                             || (ServiceStartMode)(-1) != StartupType)
                         {
                             DWORD dwStartType = NativeMethods.SERVICE_NO_CHANGE;
-                            switch (StartupType)
+                            if ((ServiceStartMode)(-1) != StartupType &&
+                                !NativeMethods.GetNativeStartupType(StartupType, out dwStartType))
                             {
-                                case ServiceStartMode.Automatic:
-                                    dwStartType = NativeMethods.SERVICE_AUTO_START;
-                                    break;
-                                case ServiceStartMode.Manual:
-                                    dwStartType = NativeMethods.SERVICE_DEMAND_START;
-                                    break;
-                                case ServiceStartMode.Disabled:
-                                    dwStartType = NativeMethods.SERVICE_DISABLED;
-                                    break;
-                                default:
-                                    Diagnostics.Assert(
-                                        ((ServiceStartMode)(-1)) == StartupType,
-                                        "bad StartupType");
-                                    break;
+                                WriteNonTerminatingError(StartupType.ToString(), "Set-Service", Name,
+                                    new ArgumentException(), "CouldNotSetService",
+                                    ServiceResources.UnsupportedStartupType,
+                                    ErrorCategory.InvalidArgument);
+                                return;
                             }
                             bool succeeded = NativeMethods.ChangeServiceConfigW(
                                 hService,
@@ -2002,26 +1992,15 @@ namespace Microsoft.PowerShell.Commands
                         ErrorCategory.PermissionDenied);
                     return;
                 }
-                DWORD dwStartType = NativeMethods.SERVICE_AUTO_START;
-                switch (StartupType)
+                DWORD dwStartType = NativeMethods.SERVICE_NO_CHANGE;
+                if (! NativeMethods.GetNativeStartupType(StartupType, out dwStartType))
                 {
-                    case ServiceStartMode.Automatic:
-                        dwStartType = NativeMethods.SERVICE_AUTO_START;
-                        break;
-                    case ServiceStartMode.Manual:
-                        dwStartType = NativeMethods.SERVICE_DEMAND_START;
-                        break;
-                    case ServiceStartMode.Disabled:
-                        dwStartType = NativeMethods.SERVICE_DISABLED;
-                        break;
-                    default:
-                        WriteNonTerminatingError(StartupType.ToString(), DisplayName, Name,
-                            new ArgumentException(), "CouldNotNewService",
-                            ServiceResources.UnsupportedStartupType,
-                            ErrorCategory.InvalidArgument);
-                        return;
+                    WriteNonTerminatingError(StartupType.ToString(), "New-Service", Name,
+                        new ArgumentException(), "CouldNotNewService",
+                        ServiceResources.UnsupportedStartupType,
+                        ErrorCategory.InvalidArgument);
+                    return;
                 }
-
                 // set up the double-null-terminated lpDependencies parameter
                 IntPtr lpDependencies = IntPtr.Zero;
                 if (null != DependsOn)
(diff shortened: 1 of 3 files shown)


iSazonov:
Blocked by #4803.

markekraus:
@SteveL-MSFT something should probably be done about [Set-Service.Tests.ps1#L25](https://github.com/PowerShell/PowerShell/blob/fb60c47a1ecbca2a6f287e0626a372497112f0a3/test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1#L25). That will continue to fail on Linux and macOS as the literal will still be parsed even though the `IT` is being skipped, and `Get-Service` is unavailable on Linux/macOS.

SteveL-MSFT:
@markekraus I see.  will fix.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2015,10 +2015,11 @@ protected override void BeginProcessing()
                         dwStartType = NativeMethods.SERVICE_DISABLED;
                         break;
                     default:
-                        Diagnostics.Assert(
-                            ((ServiceStartMode)(-1)) == StartupType,
-                            "bad StartupType");
-                        break;
+                        WriteNonTerminatingError(StartupType.ToString(), DisplayName, Name,
+                            new ArgumentException(), "CouldNotNewService",
+                            ServiceResources.UnsupportedStartupType,
```
daxian-dbw: There is another instance of `Assert(((ServiceStartMode)(-1)) == StartupType, "bad StartupType"` in `Set-Service`. It would be the best if all can be fixed here.

src/Microsoft.PowerShell.Commands.Management/resources/ServiceResources.resx
```diff
@@ -201,4 +201,7 @@
   <data name="ComputerAccessDenied" xml:space="preserve">
     <value>The command cannot be used to configure the service '{0}' because access to computer '{1}' is denied. Run PowerShell as admin and run your command again.</value>
   </data>
+  <data name="UnsupportedStartupType" xml:space="preserve">
+    <value>The startup type '{0}' is not supported by New-Service.</value>
```
daxian-dbw: `by New-Service` needs to be changed since the same applies to `set-service` as well.

daxian-dbw (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -163,8 +166,7 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
             [System.Management.Automation.PSCredential]::new("username", 
             (ConvertTo-SecureString "PlainTextPassword" -AsPlainText -Force)))
         },
-        # This test case fails due to #4803. Disabled for now.
-        # @{name = 'badstarttype'; parameter = "StartupType"; value = "System"},
+        @{name = 'badstarttype'; parameter = "StartupType"; value = "System"},
```
iSazonov: Should we add "Boot" too? 
Should we check valid and invalid values from `System.ServiceProcess.ServiceStartMode`? If yes it is better refactor the tests. And for `Set-Service` too.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
SteveL-MSFT: will fix

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
SteveL-MSFT: will change

SteveL-MSFT:
@iSazonov @daxian-dbw feedback addressed

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -1684,22 +1682,14 @@ protected override void ProcessRecord()
                             || (ServiceStartMode)(-1) != StartupType)
                         {
                             DWORD dwStartType = NativeMethods.SERVICE_NO_CHANGE;
```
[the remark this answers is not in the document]
SteveL-MSFT: I can make that change for New-Service, but for Set-Service, I need to use dwStartType on line 1697 so it has to be initialized.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2411,6 +2390,40 @@ internal static extern
...
+        /// <param name="StartupType">
+        /// StartupType provided by the user.
+        /// </param>
+        /// <param name="dwStartType">
+        /// Out parameter of the native win32 StartupType
+        /// </param>
+        /// <returns>
+        /// If a supported StartupType is provided, funciton returns true, otherwise false.
+        /// </returns>
+        internal static bool GetNativeStartupType(ServiceStartMode StartupType, out DWORD dwStartType)
```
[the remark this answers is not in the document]
SteveL-MSFT: Will change

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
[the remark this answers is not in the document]
iSazonov: Thanks for clafiry. I think we should leave as is.
Closed.

iSazonov (approved):

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -1684,22 +1682,14 @@ protected override void ProcessRecord()
                             || (ServiceStartMode)(-1) != StartupType)
                         {
                             DWORD dwStartType = NativeMethods.SERVICE_NO_CHANGE;
-                            switch (StartupType)
+                            if ((ServiceStartMode)(-1) != StartupType &&
```
daxian-dbw: It seems `(ServiceStartMode)(-1) != StartupType &&` should be removed.

SteveL-MSFT: Will change, but will need to change some other code to make this work.  Seems like `-1` shouldn't be needed at all.

SteveL-MSFT: Actually, still need `-1` which will be treated as equivalent to `SERVICE_NO_CHANGE`

daxian-dbw: By looking at the original code, for `New-Service`, `SERVICE_AUTO_START` would be used in case `StartupType` is -1. So it seems we cannot just use `dwStartType = NativeMethods.SERVICE_NO_CHANGE` for `(ServiceStartMode)(-1)`

SteveL-MSFT: New-Service [defaults](https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs#L1933) to Automatic whereas Set-Service [defaults](https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs#L1511) to -1 to indicate no change.

SteveL-MSFT:
@daxian-dbw feedback addressed

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
daxian-dbw: Please correct me if I'm wrong. The following snippet is the original code in `New-Service`. It indicates `dwStartType` should be `SERVICE_AUTO_START` when `StartupType == ((ServiceStartMode)(-1))`
```
DWORD dwStartType = NativeMethods.SERVICE_AUTO_START;
switch (StartupType)
{
    case ServiceStartMode.Automatic:
        dwStartType = NativeMethods.SERVICE_AUTO_START;
        break;
    case ServiceStartMode.Manual:
        dwStartType = NativeMethods.SERVICE_DEMAND_START;
        break;
    case ServiceStartMode.Disabled:
        dwStartType = NativeMethods.SERVICE_DISABLED;
        break;
    default:
        Diagnostics.Assert(
            ((ServiceStartMode)(-1)) == StartupType,
            "bad StartupType");
        break;
}
```
It's replaced with `if (!NativeMethods.TryGetNativeStartupType(StartupType, out DWORD dwStartType)) { // write error }` after the change, and `TryGetNativeStartupType` will set `dwStartType` to `SERVICE_NO_CHANGE` when `StartupType == ((ServiceStartMode)(-1))`.

So it looks to me changes the value of `dwStartType` in case of `StartupType == ((ServiceStartMode)(-1))`.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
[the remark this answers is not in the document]
daxian-dbw: In `New-Service`, dwStartType is also used in line 2045 when calling `NativeMethods.CreateServiceW`.
It turns out the following code works:
```
static void Blah ()
{
    if (!int.TryParse("1", out int result))
    {
        Console.WriteLine("Failed");
    }

    Console.WriteLine(result);
}
```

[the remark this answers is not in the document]
daxian-dbw: In `New-Service`, dwStartType is also used in line 2045 when calling `NativeMethods.CreateServiceW`.
It turns out the following code works:
```
static void Blah ()
{
    if (!int.TryParse("1", out int result))
    {
        Console.WriteLine("Failed");
    }

    Console.WriteLine(result);
}
```
It would be the best if we can do it in a consistent way in those 2 places.

daxian-dbw (approved):
LGTM except for a non-blocking comment.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
SteveL-MSFT: I'll fix it so if it's (-1), it should default to Auto for New-Service

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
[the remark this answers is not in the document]
SteveL-MSFT: Addressed

SteveL-MSFT:
@daxian-dbw can you review latest change?

daxian-dbw:
@SteveL-MSFT we cannot pass the value (-1) to `-StartupType` parameter. The parameter binding will fail because powershell cannot convert -1 to `[System.ServiceProcess.ServiceStartMode]`. That means `StartupType` in `New-Service` can never be `-1`, so we should be good without the last commit.

SteveL-MSFT:
@daxian-dbw yeah, I went through the code and for `New-Service`, StartupType defaults to Auto and can't ever be set to (-1), so I'll revert the last commit.

SteveL-MSFT:
Last commit was reverted.  We should be good to go.

iSazonov:
@SteveL-MSFT @daxian-dbw  Is the PR ready to merge?

daxian-dbw:
@iSazonov Yes, it's good to go.

PR #4805, opened 2017-09-11
Correct %c, %l, %k, %s, %j format in Get-Date -UFormat

### LINKED ISSUE #2195: "UNIX time stamp" different between windows and linux
## Steps to reproduce

[root@cent72 ~]# date
2016年 09月 07日 星期三 14:10:24 CST

PS /root> get-date -UFormat %s ; date +%s  
1473257427.28919
1473228627

PS C:\Windows\system32\WindowsPowerShell\v1.0>get-date
2016年9月7日 14:10:15
## Expected behavior

Today I encountered a script problem: Unix time stamp, I would like to use PowerShell to deal with, but：

get-date -UFormat %s,Seconds elapsed since January 1, 1970 00:00:00 ,
that is 【int format】! i think,but return decimal. and time zones problem.

Now powershell run in linux,return "UNIX time stamp" should be 【CONSISTENT】！
## Actual behavior

return:(i am chinese user,utc +8)
1473257427.28919
1473228627

8 \* 3600 secode different
## Environment data

win:

---

PSVersion                      5.1.14393.103
BuildVersion                   10.0.14393.103

timezone：utc+8

centos 7.2:

---

CLRVersion  

LANG=zh_CN.UTF-8

timezone：utc+8

iSazonov:
Related #4750.
Fix #2195.

In the PR we correct only obviuos formatting as set in ISO 8601 and [RFC 3339](https://tools.ietf.org/html/rfc3339#page-12)

1. Correct %c format in Get-Date -UFormat

Add leading zero: now day of the month is from 01 through 31.

Before fix:
```powershell
>Get-Date -UFormat %c
Thu Sep 7 17:49:48 2017
```
After the fix:
```powershell
>Get-Date -UFormat %c
Thu Sep 07 2017 17:49:48
```

2. Correct %l format in Get-Date -UFormat

Single digits are preceded by a blank.

Before fix:
```powershell
> get-date 2016-01-1 -UFormat %l
00
```
After the fix:
```powershell
> get-date 2016-01-1 -UFormat %l
 0
```

3. Correct %k format in Get-Date -UFormat

Single digits are preceded by a blank.

Before fix:
```powershell
> get-date 2016-01-1 -UFormat %k
01
```
After the fix:
```powershell
> get-date 2016-01-1 -UFormat %k
1
```

4. Correct %s format in Get-Date -UFormat

Removed fraction part of seconds.

Before fix:
```powershell
> get-date 2016-01-1 -UFormat %s -Millisecond 1
1451606400.001
```
After the fix:
```powershell
> get-date 2016-01-1 -UFormat %s -Millisecond 1
1451606400
```

5. Correct %j format in Get-Date -UFormat

Add leading zeros.

Before fix:
```powershell
>Get-date -Date 1/1/0030 -uformat %j
1
```
After the fix:
```powershell
>Get-date -Date 1/1/0030 -uformat %j
001
```


chuanjiao10:
#2195

iSazonov:
@chuanjiao10 Thanks for #2195 - the PR fix it.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
```diff
@@ -369,9 +370,7 @@ private string UFormatDateString(DateTime dateTime)
                             break;
 
                         case 'c':
-                            sb.Append("{0:ddd} {0:MMM} ");
-                            sb.Append(StringUtil.Format("{0,2} ", dateTime.Day));
-                            sb.Append("{0:HH}:{0:mm}:{0:ss} {0:yyyy}");
+                            sb.Append("{0:ddd} {0:MMM} {0:dd} {0:HH}:{0:mm}:{0:ss} {0:yyyy}");
```
SteveL-MSFT: This is different from Linux using `datetime -d "09/09/2017 08:30" +%c` which shows:

```none
Sat 09 Sept 2017 08:30:00 AM PDT
```

Not sure what the canonical representation should be

iSazonov: I'll move a day number to second position.

iSazonov: Fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
```diff
@@ -369,9 +370,7 @@ private string UFormatDateString(DateTime dateTime)
                             break;
 
                         case 'c':
-                            sb.Append("{0:ddd} {0:MMM} ");
-                            sb.Append(StringUtil.Format("{0,2} ", dateTime.Day));
-                            sb.Append("{0:HH}:{0:mm}:{0:ss} {0:yyyy}");
+                            sb.Append("{0:ddd} {0:dd} {0:MMM} {0:HH}:{0:mm}:{0:ss} {0:yyyy}");
```
SteveL-MSFT: Shouldn't this be

```c#
"{0:ddd} {0:dd} {0:MMM} {0:yyyy} {0:HH}:{0:mm}:{0:ss} {0:tt}"
```

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
iSazonov: If we add {0:tt} should we use {0:hh}?

JamesWTruher: I thought that the positioning of the month and day `({0:dd} and {0:MMM})` were locale dependent. It is for POSIX anyway (http://pubs.opengroup.org/onlinepubs/009604599/utilities/date.html). This routine might need to be a bit more clever

JamesWTruher (commented):
I'm concerned about the format of `%c`. It's supposed to be locale dependent

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
iSazonov: The problem is that we need to emulate UNIX behavior based on Unix locales to get the same that datetime returns but we can use only CoreFX locales.
We can get standard format strings:
```powershell
[CultureInfo]::new("en-US").DateTimeFormat
```
There I don't see that we can use.

I test on WSL
```
LC_TIME="en_US.UTF-8" date +%c
LC_TIME="ru_RU.UTF-8" date +%c
```
 and get the same format for all locales. Could you test on other Unix?
So I suggest use ""{0:ddd} {0:dd} {0:MMM} {0:yyyy} {0:HH}:{0:mm}:{0:ss}" as a compromise.

iSazonov:
@SteveL-MSFT Could we have a conclusion from PowerShell committee? The Unix like formatting very problematic. It was acceptable in Windows PowerShell but in PowerShell Core it is confusing because users can get (out of PowerShell Core) native Unix formatting and one is different from PowerShell Core formatting.
So we could:
1. Keep this as is for backward compatibility with Windows PowerShell.
2. Make this formatting closer as possible to Unix original that PR brings. It may look still bad.
3. Think about complex portable solution. I have not considered this particularly deeply, but so far I believe that this is not possible within existing Windows and Unix implementations. In particular, different Unix/Linux clones have different implementations.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and is ok with the improvements in the PR and defer the locale concern til later as a separate PR

TravisEz13:
@SteveL-MSFT  Can you update your review?  
@iSazonov merge conflicts need to be resolved?  If you don't want this reviewed, mark it as a `wip`

iSazonov:
Merge conflicts was resolved.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
SteveL-MSFT: @iSazonov looks like you didn't make this change (the year is still at the end)?

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
iSazonov: Done.

SteveL-MSFT (approved):

daxian-dbw:
@iSazonov Please review the PR summary again and see if the summary is still up-to-date. If things were changed in the final changes that were merged, then please update the PR summary accordingly. Thanks!

iSazonov:
@daxian-dbw I updated the PR description.

PR #4811, opened 2017-09-11
Validate product resource strings against resx files

JamesWTruher:
These tests validate that the resource strings represent the contents of the RESX files. This essentially ensures that we've run `-ResGen` when building. Additionally, these tests will ensure that we don't have ResourceIds which have a space in them (which we did). 

When there's a space in the ResourceId, the tooling converts the space to an "_" which is fine, but the compiled resource doesn't have that change, so when message retrieval is attempted, an empty string is returned.

Normally we generate an error and validate the ErrorId, but this test is different. We want to be sure that the message that we're giving the user is correct and the build included the `-ResGen` parameter.

### DIFF
--- a/test/powershell/engine/ResourceValidation/ConsoleHostResources.Tests.ps1
+++ b/test/powershell/engine/ResourceValidation/ConsoleHostResources.Tests.ps1
@@ -1,50 +1,9 @@
+. "$psscriptroot/TestRunner.ps1"
+
 $AssemblyName = "Microsoft.PowerShell.ConsoleHost"
-# this list is taken from ${AssemblyName}.csproj
+
 # excluded resources
 $excludeList = @("HostMshSnapinResources.resx")
 
-#
-# Boiler Plate
-#
-$repoBase = (Resolve-Path (Join-Path $psScriptRoot ../../../..)).Path
-$asmBase = Join-Path $repoBase "src/$AssemblyName"
-$resourceDir = Join-Path $asmBase resources
-$resourceFiles = Get-ChildItem $resourceDir -Filter *.resx | Where-Object {
-    $excludeList -notcontains $_.Name
-    }
-
-$bindingFlags = [reflection.bindingflags]"NonPublic,Static"
-
-# the resource generation is based on the file name of the .resx file
-# and is not in a namespace. We can find all of these classes in the
-# ${AssemblyName} assembly
-$ASSEMBLY = [appdomain]::CurrentDomain.GetAssemblies()|
-    Where-Object { $_.FullName -match "$AssemblyName" }
-
-# Validate that the binary you're running has the proper message
-# based on the resource files. It is possible that -ResGen could be
-# skipped and the messages might become out of sync with what is in 
-# source
-# you could argue that we should be generating conditions where the message is delivered
-# and then test for that, but that is a check against an ErrorId while this is ensuring
-# that the contents of the built binary matches the resx content
-Describe "Resources strings in $AssemblyName (was -ResGen used with Start-PSBuild)" -tag Feature {
-    foreach ( $resourceFile in $resourceFiles )
-    {
-        # in the event that the id has a space in it, it is replaced with a '_'
-        $classname = $resourcefile.Name -replace ".resx"
-        It "'$classname' should be an available type and the strings should be correct" {
-            # get the type from ASSEMBLY
-            $resourceType = $ASSEMBLY.GetType($classname, $false, $true)
-            # the properties themselves are static internals, so we need
-            # to using the appropriate bindingflags
-            $resourceType | Should Not BeNullOrEmpty
-
-            # check all the resource strings
-            $xmlData = [xml](Get-Content $resourceFile.Fullname)
-            foreach ( $inResource in $xmlData.root.data ) {
-                $resourceType.GetProperty($inResource.name,$bindingFlags).GetValue(0) | should be $inresource.value
-            }
-        }
-    }
-}
+# run the tests
+Test-ResourceStrings -AssemblyName $AssemblyName -ExcludeList $excludeList
--- a/test/powershell/engine/ResourceValidation/ConsoleHostResources.Tests.ps1
+++ b/test/powershell/engine/ResourceValidation/ConsoleHostResources.Tests.ps1
@@ -1,0 +1,50 @@
+$AssemblyName = "Microsoft.PowerShell.ConsoleHost"
+# this list is taken from ${AssemblyName}.csproj
+# excluded resources
+$excludeList = @("HostMshSnapinResources.resx")
+
+#
+# Boiler Plate
+#
+$repoBase = (Resolve-Path (Join-Path $psScriptRoot ../../../..)).Path
+$asmBase = Join-Path $repoBase "src/$AssemblyName"
+$resourceDir = Join-Path $asmBase resources
+$resourceFiles = Get-ChildItem $resourceDir -Filter *.resx | Where-Object {
+    $excludeList -notcontains $_.Name
+    }
+
+$bindingFlags = [reflection.bindingflags]"NonPublic,Static"
+
+# the resource generation is based on the file name of the .resx file
+# and is not in a namespace. We can find all of these classes in the
+# ${AssemblyName} assembly
+$ASSEMBLY = [appdomain]::CurrentDomain.GetAssemblies()|
+    Where-Object { $_.FullName -match "$AssemblyName" }
+
+# Validate that the binary you're running has the proper message
+# based on the resource files. It is possible that -ResGen could be
+# skipped and the messages might become out of sync with what is in 
+# source
+# you could argue that we should be generating conditions where the message is delivered
+# and then test for that, but that is a check against an ErrorId while this is ensuring
+# that the contents of the built binary matches the resx content
+Describe "Resources strings in $AssemblyName (was -ResGen used with Start-PSBuild)" -tag Feature {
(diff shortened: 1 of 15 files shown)


iSazonov:
Does it make sense to check for unused ResourceId in resx files?

JamesWTruher:
@iSazonov re: check for unused ResourceId in resx files
it would be good to check for these. I thought about how to do that, but couldn't come up something simple, other than brute force checking all the source files.

test/powershell/engine/ResourceValidation/TestRunner.ps1
```diff
@@ -0,0 +1,57 @@
+function Test-ResourceStrings
```
SteveL-MSFT: Would it be better to put this into an existing module or a new test module?

test/powershell/engine/ResourceValidation/ConsoleHostResources.Tests.ps1
```diff
@@ -0,0 +1,9 @@
+. "$psscriptroot/TestRunner.ps1"
+
+$AssemblyName = "Microsoft.PowerShell.ConsoleHost"
+
+# excluded resources
+$excludeList = @("HostMshSnapinResources.resx")
```
SteveL-MSFT: Should add comment why it's being excluded

test/powershell/engine/ResourceValidation/MPCUResource.Tests.ps1
```diff
@@ -0,0 +1,17 @@
+. "$psscriptroot/TestRunner.ps1"
+$AssemblyName = "Microsoft.PowerShell.Commands.Utility"
+# this list is taken from ${AssemblyName}.csproj
+# excluded resources
+$excludeList = "CoreMshSnapinResources.resx",
```
SteveL-MSFT: Add comment why these are excluded

test/powershell/engine/ResourceValidation/ManagementCommandsResources.Tests.ps1
```diff
@@ -0,0 +1,16 @@
+. "$psscriptroot/TestRunner.ps1"
+$AssemblyName = "Microsoft.PowerShell.Commands.Management"
+# this list is taken from ${AssemblyName}.csproj
+# excluded resources
+$excludeList = "EventlogResources.resx",
```
SteveL-MSFT: Added comment why excluded

test/powershell/engine/ResourceValidation/SMAResource.Tests.ps1
```diff
@@ -0,0 +1,8 @@
+. "$psscriptroot/TestRunner.ps1"
+$AssemblyName = "System.Management.Automation"
+# this list is taken from ${AssemblyName}.csproj
+# excluded resources
+$excludeList = "CoreMshSnapinResources.resx",
```
SteveL-MSFT: Comment why excluded

test/powershell/engine/ResourceValidation/SecurityResource.Tests.ps1
```diff
@@ -0,0 +1,11 @@
+. "$psscriptroot/TestRunner.ps1"
+
+$assemblyName = "Microsoft.PowerShell.Security"
+# this list is taken from ${AssemblyName}.csproj
+# excluded resources
+$excludeList = @("SecurityMshSnapinResources.resx")
```
SteveL-MSFT: Comment why excluded

test/powershell/engine/ResourceValidation/TestRunner.ps1
```diff
@@ -0,0 +1,57 @@
+function Test-ResourceStrings
+{
+    param ( $AssemblyName, $ExcludeList )
+
+    # determine the needed resource directory. If these tests are moved
+    # this logic will need to change
+    $repoBase = (Resolve-Path (Join-Path $psScriptRoot ../../../..)).Path
```
SteveL-MSFT: Is it better to use: git rev-parse --show-toplevel?

test/powershell/engine/ResourceValidation/TestRunner.ps1
```diff
@@ -0,0 +1,57 @@
+function Test-ResourceStrings
+{
+    param ( $AssemblyName, $ExcludeList )
+
+    # determine the needed resource directory. If these tests are moved
+    # this logic will need to change
+    $repoBase = (Resolve-Path (Join-Path $psScriptRoot ../../../..)).Path
+    $asmBase = Join-Path $repoBase "src/$AssemblyName"
+    $resourceDir = Join-Path $asmBase resources
+    $resourceFiles = Get-ChildItem $resourceDir -Filter *.resx -ea stop | Where-Object {
```
SteveL-MSFT: Why not use `-Exclude` parameter for `Get-ChildItem`?

SteveL-MSFT (changes requested):

test/powershell/engine/ResourceValidation/DotNetEventing.Tests.ps1
```diff
@@ -0,0 +1,10 @@
+. "$psscriptroot/TestRunner.ps1"
```
daxian-dbw: To be consistent with other files, maybe change the file name from `DotNetEventing.Tests.ps1` to `***Resource.Tests.ps1`

test/powershell/engine/ResourceValidation/MPCUResource.Tests.ps1
```diff
@@ -0,0 +1,17 @@
+. "$psscriptroot/TestRunner.ps1"
```
daxian-dbw: Since the file for management dll is `ManagementCommandsResources.Tests.ps1`, Maybe change `MPCUResource.Tests.ps1` to `Utility.Resource.Tests.ps1`?

daxian-dbw (approved):
Leave 2 comments. They don't block this PR.

test/powershell/engine/ResourceValidation/TestRunner.ps1
JamesWTruher: i wanted to do it this way because I do *not* want to make this a general function - we shouldn't be testing internal data structures, but there's no way to do this test without it.

test/powershell/engine/ResourceValidation/TestRunner.ps1
JamesWTruher: sure - I'll make that change

test/powershell/engine/ResourceValidation/TestRunner.ps1
JamesWTruher: i can do that

test/powershell/engine/ResourceValidation/DotNetEventing.Tests.ps1
JamesWTruher: will change

test/powershell/engine/ResourceValidation/MPCUResource.Tests.ps1
JamesWTruher: will change

test/powershell/engine/ResourceValidation/TestRunner.ps1
adityapatwardhan: Cannot assume that we have git tools. For example, the Code Coverage machine does not have git tools.

adityapatwardhan (approved):
LGTM

test/powershell/engine/ResourceValidation/TestRunner.ps1
SteveL-MSFT: ok

test/powershell/engine/ResourceValidation/TestRunner.ps1
SteveL-MSFT: in that case, this is fine

test/powershell/engine/ResourceValidation/TestRunner.ps1
JamesWTruher: huh - it turns out that we might have a bug - `get-childitem -filter *.resx -exclude <thing>` returns no results. I'll have to leave it as is.

JamesWTruher:
@SteveL-MSFT updated and pushed - see my comment about using `-filter` and `-exclude` together - I double checked with PowerShell 5 and the behavior is the same. if you use `-filter` and `-exclude` no results are returned, so at least it's not a regression, although i agree the behavior is a little surprising

test/powershell/engine/ResourceValidation/TestRunner.ps1
SteveL-MSFT: Now that you mention it, there's a number of issues with `-exclude`.  Ok with leaving as-is

SteveL-MSFT (approved):

iSazonov:
@JamesWTruher re: check for unused ResourceId in resx files

It seems @lzybkr once mentioned some tool.
As for me I think better way is to get native warning from the compiler that does still not implemented.

lzybkr:
@iSazonov - I use ReSharper to find unused references. There were some accuracy issues when the csproj file wasn't the same as how we built, but it worked fairly well anyway, so I'm guessing it's pretty accurate now.

iSazonov:
@lzybkr Thanks! 
I see there is free [ReSharper command tools](https://www.jetbrains.com/resharper/download/index.html#section=resharper-clt)

JamesWTruher:
@iSazonov @lzybkr  I created https://github.com/PowerShell/PowerShell/issues/4860 to track

Issue #4778, opened 2017-09-07
Windows system lockdown APIs are currently stubbed out


PaulHigin:
This means that PowerShell Core running on Windows will not detect system lockdown such as when DeviceGuard or AppLocker application allow policies are applied, and so won't automatically transition into constrained language mode.

This is a work item to re-implement these APIs for Windows platform and to finish porting the corresponding tests.

joeyaiello:
Not enough coverage on or usage of 6.0.0 yet to justify this for the 6.0.0 timeframe. Definitely necessary in the long-run though.

PaulHigin:
Yes, this issue is now closed.  Thanks.

PR #4816, opened 2017-09-12
Port tests for constrained language mode

PaulHigin:
These tests have been ported from Windows code base and converted to Pester.
Note that some tests are currently pending because PowerShell Core does not yet support Windows platform application allow APIs.  This is being tracked by issue #4778.


test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,890 @@
+﻿##
+## Tests for PowerShell system lock down and constrained language mode restrictions
+## These are Windows platform only tests
+##
+
+function Get-RandomFileName
```
JamesWTruher: this seems like a mighty useful function, perhaps it should be in the test/tools/Modules/HelpersCommon modules

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,890 @@
...
+    }
+'@
+
+    if (-not (Get-Command Invoke-LanguageModeTestingSupportCmdlet -ea Ignore))
+    {
+        $moduleName = Get-RandomFileName
+        $moduleDirectory = join-path $TestDrive\Modules $moduleName
+        if (-not (Test-Path $moduleDirectory))
+        {
+            mkdir $moduleDirectory -Force | Out-Null
```
JamesWTruher: this should be `$null = New-Item -ItemType Directory $moduleDirectory -Force`

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,890 @@
...
+        }
+
+        try 
+        { 
+            Add-Type -TypeDefinition $code -OutputAssembly $moduleDirectory\TestCmdletForConstrainedLanguage.dll -ErrorAction Ignore 
+        } catch {}
```
JamesWTruher: if this actually fails, you'll not be able to run any tests, correct? Should the catch really be silent?

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,890 @@
...
+                    Start-Job { [object]::Equals("A", "B") }
+                    throw "No Exception!"
+                }
+                catch 
+                {
+                    $exception = $_
+                }
+
+                Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
+
```
JamesWTruher: the above block should probably be:
```
try
{
    Start-Job { [object]::Equals("A","B") }
    throw "No Exception!"
}
catch
{
    $exception = $_
}
finally
{
    Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
}
```
and the same below

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,890 @@
...
+                $result.Direction | Should Be 1
+            }
+        }
+
+        AfterEach {
+            Remove-SystemLockDown
+        }
+    }
+
+    Describe "Mkdir command on function drive in constrained language" -Tags 'Feature' {
```
JamesWTruher: is the mkdir function the point of this, or should it be using `new-item -itemtype directory`?

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,890 @@
...
+
+            function MyDebuggerFunction3 {}
+            $null = Set-PSBreakpoint -Command MyDebuggerFunction3 -Action { $Global:dbgResult = [object]::Equals("A", "B") }
+
+            MyDebuggerFunction3
+            $expectedError = $Error[0]
+
+            Remove-SystemLockDown
+
+            $expectedError.FullyQualifiedErrorId | Should Match "MethodInvocationNotSupportedInConstrainedLanguage"
```
JamesWTruher: i would expect this could be done as `try { } catch { } finally { }` if `-ErrorAction Stop` was used

JamesWTruher (dismissed):

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Will do.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Will change.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: No, the failure will be at import-module

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Agreed.  I was trying to use Pester AfterAll, AfterEach.  But I prefer try/finally.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Mkdir.  This is a function test.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Ok.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Nope, you're right it should be new-item

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+                Invoke-LanguageModeTestingSupportCmdlet -EnableConstrainedLanguageMode
+
+                $result = (& $scriptblock)
+            }
+            finally
+            {
+                Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
+            }
+
+            ($result.Count -gt 0) | Should Be $true
```
SteveL-MSFT: Use: Should BeGreaterThan

Gives better test failure message

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+                catch 
+                {
+                    $exception = $_
+                }
+                finally
+                {
+                    Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
+                }
+
+                $exception.FullyQualifiedErrorId | Should Match "CannotStartJobInconsistentLanguageMode"
```
SteveL-MSFT: Since Start-Job throws, you should use ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            catch
+            {
+                $exception = $_
+            }
+            finally
+            {
+                Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
+            }
+
+            $exception.FullyQualifiedErrorId | Should Match "CannotDefineNewType"
```
SteveL-MSFT: Use ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+                catch
+                {
+                    $exception = $_
+                }
+                finally
+                {
+                    Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
+                }
+
+                $exception.FullyQualifiedErrorId | Should Match "CannotCreateTypeConstrainedLanguage"
```
SteveL-MSFT: Use ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+                catch
+                {
+                    $exception = $_
+                }
+                finally
+                {
+                    Invoke-LanguageModeTestingSupportCmdlet -RevertLockdownMode -EnableFullLanguageMode
+                }
+
+                $exception.FullyQualifiedErrorId | Should Match "CannotCreateComTypeConstrainedLanguage"
```
SteveL-MSFT: Use ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            catch
+            {
+                $exception = $_
+            }
+            finally
+            {
+                Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
+            }
+
+            $exception.FullyQualifiedErrorId | Should Match "NotSupported"
```
SteveL-MSFT: Use ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            catch
+            {
+                $exception = $_
+            }
+            finally
+            {
+                Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
+            }
+
+            $exception.FullyQualifiedErrorId | Should Match "CannotSetBreakpointInconsistentLanguageMode"
```
SteveL-MSFT: Use ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+                $null = Set-PSBreakpoint -Command MyDebuggerFunction3 -Action { $Global:dbgResult = [object]::Equals("A", "B") }
+
+                $restoreEAPreference = $ErrorActionPreference
+                $ErrorActionPreference = "Stop"
+                MyDebuggerFunction3
+            }
+            catch
+            {
+                $exception = $_
+                throw "No Exception!"
```
SteveL-MSFT: out of place throw

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            }
+            finally
+            {
+                if ($restoreEAPreference -ne $null) { $ErrorActionPreference = $restoreEAPreference }
+                Invoke-LanguageModeTestingSupportCmdlet -RevertLockdownMode -EnableFullLanguageMode
+            }
+
+            $exception.FullyQualifiedErrorId | Should Match "MethodInvocationNotSupportedInConstrainedLanguage"
```
SteveL-MSFT: Use ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+                $null = New-Event LockdownEvent
+                Wait-Job $job
+                Unregister-Event LockdownEvent
+            }
+            finally
+            {
+                Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
+            }
+
+            $job.Error.FullyQualifiedErrorId | Should Match "MethodInvocationNotSupportedInConstrainedLanguage"
```
SteveL-MSFT: Use ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            catch
+            {
+                $exception = $_
+            }
+            finally
+            {
+                Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
+            }
+
+            $exception.FullyQualifiedErrorId | Should Match "MethodInvocationNotSupportedInConstrainedLanguage"
```
SteveL-MSFT: ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            catch
+            {
+                $exception = $_
+            }
+            finally
+            {
+                $rs.Dispose()
+            }
+
+            $exception.FullyQualifiedErrorId | Should Match "DriveNotFoundException"
```
SteveL-MSFT: ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+
+    Describe "Get content syntax in constrained language mode" -Tags 'Feature','RequireAdminOnWindows' {
+
+        It "Verifies that the get content syntax returns null value in constrained language without provider" {
+
+            $iss = [initialsessionstate]::Create()
+            $iss.LanguageMode = "ConstrainedLanguage"
+            $rs = [runspacefactory]::CreateRunspace($iss)
+            $rs.Open()
+            $pl = $rs.CreatePipeline('${c:\temp\foo.txt}')
```
SteveL-MSFT: Use testdrive

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+    Describe "Stream redirection in constrained language mode" -Tags 'Feature','RequireAdminOnWindows' {
+
+        It "Verifies that stream redirection doesn't work in constrained language mode without provider" {
+
+            $iss = [initialsessionstate]::CreateDefault2()
+            $iss.Providers.Clear()
+            $iss.LanguageMode = "ConstrainedLanguage"
+            $rs = [runspacefactory]::CreateRunspace($iss)
+            $rs.Open()
+            $pl = $rs.CreatePipeline('"Hello" > c:\temp\foo.txt')
```
SteveL-MSFT: Use testdrive

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            catch
+            {
+                $exception = $_
+            }
+            finally
+            {
+                $rs.Dispose()
+            }
+
+            $exception.FullyQualifiedErrorId | Should Match "CmdletInvocationException"
```
SteveL-MSFT: ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            catch
+            {
+                $exception = $_
+            }
+            finally
+            {
+                Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
+            }
+
+            $exception.FullyQualifiedErrorId | Should Match "MethodInvocationNotSupportedInConstrainedLanguage"
```
SteveL-MSFT: ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
```
SteveL-MSFT: shouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
```
SteveL-MSFT: ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+                & $scriptblock
+                throw 'No Exception!'
+            }
+            catch
+            {
+                $exception = $_
+            }
+            finally
+            {
+                Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
```
SteveL-MSFT: You repeat this enough I think it would be better to have it in an AfterEach

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            }
+
+            $exception.FullyQualifiedErrorId | Should Match "NotSupported"
```
SteveL-MSFT: ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
```
SteveL-MSFT: ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+                $exception1 = InvokeDataSectionConstrained
+                # Repeat to make sure the first time properly restored the language mode to constrained.
+                $exception2 = InvokeDataSectionConstrained
+            }
+            finally
+            {
+                Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
+            }
+
+            $exception1.FullyQualifiedErrorId | Should Match "DataSectionAllowedCommandDisallowed"
```
SteveL-MSFT: ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            $exception2.FullyQualifiedErrorId | Should Match "DataSectionAllowedCommandDisallowed"
```
SteveL-MSFT: ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            catch
+            {
+                $exception = $_
+            }
+            finally
+            {
+                Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
+            }
+
+            $exception.FullyQualifiedErrorId | Should Match "DataSectionAllowedCommandDisallowed"
```
SteveL-MSFT: ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            catch
+            {
+                $exception = $_
+            }
+            finally
+            {
+                Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
+            }
+
+            $exception.FullyQualifiedErrorId | Should Match "CannotDefineSupportedCommand"
```
SteveL-MSFT: ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            catch
+            {
+                $exception = $_
+            }
+            finally
+            {
+                Invoke-LanguageModeTestingSupportCmdlet -EnableFullLanguageMode
+            }
+
+            $exception.FullyQualifiedErrorId | Should Match "MethodInvocationNotSupportedInConstrainedLanguage"
```
SteveL-MSFT: ShouldBeErrorId

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            param (
+                [string] $script
+            )
+
+            try
+            {
+                Invoke-LanguageModeTestingSupportCmdlet -EnableConstrainedLanguageMode
+
+                $sb = [scriptblock]::Create($script)
+                & sb
```
SteveL-MSFT: Seems better to make $script above as scriptblocks instead of strings, then just: & $script

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+    # End Describe blocks
+}
+finally
+{
+    if ($IsWindows)
+    {
+        Invoke-LanguageModeTestingSupportCmdlet -RevertLockdownMode -EnableFullLanguageMode
+    }
+
+    if ($defaultParamValues -ne $null)
```
SteveL-MSFT: I don't think this check is needed

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Changed.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: No, direct match to ErrorId is fragile because the provided Id can be augmented over time.  I prefer a match test.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Fixed

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: It is not needed here as the test is expected to fail with specific error id.  Just a path string is required.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: It is not needed here as the test is expected to fail with specific error id.  Just a path string is required.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.  BTW it is not necessary to add the same comment in each place :)

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: See Test Note comment at top of file.  Pester AfterEach and AfterAll is not reliable when session is constrained language or locked down.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: No.  Script block must be created within constrained language mode or test is not valid.  I'll add a comment.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: I agree that it is not needed now but I prefer to keep it in case something is added to the test in the future where an error can be thrown before $defaultParamValues is assigned.  It is essential that the global variable is restored or following tests will not run correctly.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Same reply.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
SteveL-MSFT: Ok, can you add a comment then since you're using a different pattern than what we recommend?

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
SteveL-MSFT: I realized it after the Nth time, but early on it makes it easy to check if changes were applied in all cases :)

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
SteveL-MSFT: We want to support ConstrainedLanguage on Linux/Mac as well, so having a Windows path is a bit misleading.  If you just need a path, I think $testdrive\foo.txt is better.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
SteveL-MSFT: same as above

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
SteveL-MSFT: Got it.  Thanks.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+                    $job = Start-Job -ScriptBlock { [object]::Equals("A", "B") } | Wait-Job
+                    $expectedError = $job.ChildJobs[0].Error
+                    $job | Remove-Job
+                }
+                finally
+                {
+                    Invoke-LanguageModeTestingSupportCmdlet -RevertLockdownMode
+                }
+
+                $expectedError | Should Not Be $null
```
adityapatwardhan: This check is not needed. The check below is sufficient,

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
```diff
@@ -0,0 +1,917 @@
...
+            $iss = [initialsessionstate]::Create()
+            $iss.LanguageMode = "ConstrainedLanguage"
+            $rs = [runspacefactory]::CreateRunspace($iss)
+            $rs.Open()
+            $pl = $rs.CreatePipeline('${c:\temp\foo.txt}')
+
+            $result = $pl.Invoke()
+            $rs.Dispose()
+
+            $result[0] | Should Be $null
```
adityapatwardhan: Please use ```Should BeNullOrEmpty```

adityapatwardhan (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
TravisEz13: maybe add a comment why it's ok to not do anything with the exception.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Removed.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Done.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: I don't think we should recommend this pattern unless we require FullyQualifiedErrorId to always be unchanging.  But this wasn't true in the past and as far as I know the only requirement is that it be unique and readable.  I have investigated numerous test failures where the ErrorId was augmented (sometimes by rolling up nested errors via method calls or the remoting system) and so the direct comparison failed even though the test succeeded and the error Id contained the correct error.  It might be better change ShouldBeErrorId to do a contains rather than an exact match, or create a new "ShouldContainErrorId" keyword.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Done.

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
PaulHigin: Done

test/powershell/Modules/Microsoft.PowerShell.Security/ConstrainedLanguageRestriction.Tests.ps1
SteveL-MSFT: According to https://msdn.microsoft.com/en-us/library/ms714465(v=vs.85).aspx the FQErrorId should be reproducible and same across systems, so not sure what you mean by augmented.

PaulHigin: What I mean by augmented is that the FQErrorId string is added to under some circumstances as errors are rolled up.  For example you may throw an error with error string "CannotDefineSupportedCommand", but the actual error Id you get has more information, something like:
```powershell
"CannotDefineSupportedCommand,Microsoft.PowerShell.Commands.InvokeExpressionCommand"
or
"CannotDefineSupportedCommand,PSRemotingTransportException"
```
depending on how the error came about.

The problem is that in the past the reported FQErrorId string changed due to code refactoring or new features and our tests broke because they required exact matches on the entire FQErrorId string, even when the actual error component was still in the string:
```powershell
"CannotDefineSupportedCommand,MethodCallFailed,PSRemotingTransportException"
FQErrorID -eq "CannotDefineSupportedCommand,PSRemotingTransportException" -> Test Failure
```

That is why I advocate for tests to match the expected *component* in a FQErrorId string:
```powershell
FQErrorID -match "CannotDefineSupportedCommand"  -> Test Succeeds
```

SteveL-MSFT: This seems to contradict the MSDN link above.  cc @BrucePay

PaulHigin: I don't think so.  PowerShell processing doesn't change or remove FQErrorIds for a specific code path, it just adds more information under certain circumstances.  You can still match to the specific error Id you care about.

Anyway I am not defending how it works, just relating my experience as to why I don't trust an exact match for tests.

SteveL-MSFT (approved):

adityapatwardhan (approved):
LGTM

adityapatwardhan:
@PaulHigin Please execute Feature tests as this PR include Feature tests. 
https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr

PaulHigin:
@JamesWTruher 
You have not approved this PR yet.  Thanks.

PaulHigin:
@adityapatwardhan 
All of the new constrained language tests passed.  A bunch of unrelated tests failed.

adityapatwardhan:
@PaulHigin A rebase is need to resolve those failures. Can you please rebase and push. Thanks!

adityapatwardhan:
@PaulHigin you may also want to squash some commits while rebasing.

PaulHigin:
Ok. I have performed a rebase.  Hopefully the tests will pass.

adityapatwardhan:
I pushed an empty commit with [Feature] so that the Feature tests are executed.

adityapatwardhan:
@PaulHigin Thanks for adding these tests. 👍

PaulHigin:
@adityapatwardhan 
Thanks!

daxian-dbw:
@PaulHigin Some tests look to me should be able to run on Unix platforms, such as the following:
```
Describe "Switch -file in constrained language mode"
Describe "Get content syntax in constrained language mode"
Describe "Stream redirection in constrained language mode"
Describe "Invoke-Expression in constrained language mode"
...
```
But currently, all those tests are made windows only. Is there something prevents them from running on Unix now?

(I didn't go through all conversations in this PR, so sorry that if you already answered a similar question)

PR #4822, opened 2017-09-12
Skip Certificate Authentication tests on CentOS and Mac

JamesWTruher:
These two tests have been failing on CentOS for some time. It is because the available libcurl library does not support client authentication certificates. Marking them as Pending for both CentOS and MacOS


dantraMSFT (approved):

SteveL-MSFT (approved):

adityapatwardhan (approved):
LGTM

PR #4824, opened 2017-09-12
Improve the guideline for perf consideration

daxian-dbw:


docs/dev-process/coding-guidelines.md
```diff
@@ -123,6 +123,15 @@ Some general guidelines:
...
   In this way you can avoid converting to the type twice.
 
+* Use generic collections instead of `ArrayList` to avoid type casting and unnecessary boxing whenever possible.
+
+* Use collection constructor overloads that take an initial capacity for collection types that have them.
+  Internally, `List<T>`, `Dictionary<TKey, TValue>`,
+  and the other generic collections use one or more arrays to hold valid data.
+  Whenever resizing is needed,
+  one or more new arrays double the size of existing arrays are created and items from the existing arrays are copied.
+  Setting an approximate initial capacity will reduce the number of resizing operations.
```
iSazonov: Interesting, I saw in MSDN some best practice coding guidance - we could collect it in the file. (Not in the PR)

docs/dev-process/coding-guidelines.md
```diff
@@ -123,6 +123,15 @@ Some general guidelines:
   Instead, use `var example = obj as Example` or the C# 7 syntax `if (obj is Example example) {...}` as appropriate.
   In this way you can avoid converting to the type twice.
 
+* Use generic collections instead of `ArrayList` to avoid type casting and unnecessary boxing whenever possible.
```
iSazonov: Maybe we should recommend use generic collections over classicc collections (not only `ArrayList`)?

iSazonov (approved):
LGTM with minor comment.

docs/dev-process/coding-guidelines.md
daxian-dbw: I mainly see `ArrayList` in past experiences. What other collection types do you refer to?

docs/dev-process/coding-guidelines.md
daxian-dbw: Yes, we should improve the guidelines gradually. I think the entries put here should be things general enough and commonly used in PS coding work.

docs/dev-process/coding-guidelines.md
iSazonov: I meant something like "Use generic collections _in particular_ instead of `ArrayList` to avoid type casting... "
It seems we use `Dictionary<TKey, TValue>`.

docs/dev-process/coding-guidelines.md
iSazonov: I meant "collent links to the MSDN guidances"

docs/dev-process/coding-guidelines.md
daxian-dbw: I think I get your point. I will mention `Stack`, `Queue` and `Hashtable` along with `ArrayList`.

docs/dev-process/coding-guidelines.md
daxian-dbw: That sounds good 👍

iSazonov:
LGTM.

PR #4829, opened 2017-09-13
Exclude directories from -Path in Select-String

### LINKED ISSUE #4820: Select-String's wildcard matching for -Path should exclude directories
`Select-String` only supports _file_ paths passed to `-LiteralPath` and (in resolved form) `-Path`.

As an aside: Even though the error message when you pass a _directory_ path is technically correct, it could be more helpful:

> Select-String: The file /path/to/dir cannot be read: Access to the path '/path/to/dir' is denied.

It would be convenient to automatically exclude _directory_ matches from what a wildcard expression passed to `-Path` resolves to, given that they won't be processed anyway.

This would allow searching all files in a directory, whatever extension they have or whether they have one at all, with `Select-String -Path * findme`

On a side note: `-Path`'s wildcard matching currently _invariably_ excludes _hidden_ files, so the only way to include hidden files in the search is to pipe from `Get-Item` / `Get-ChildItem` with `-Force`.

Current behavior
------------------

Setup: Create a directory that contains a subdirectory and a text file:

```powershell
Push-Location (New-Item -Type Directory -Force (Join-Path ([io.path]::GetTempPath()) $PID))
$null = New-Item -Type Directory subd
'hi' > t.txt
```

Using wildcard expression `*` reports an error, because `*` also matches the _directory_, `subd`:

```powershell
> Select-String -Path * hi
Select-String : The file .../subd cannot be read: Access to the path '.../subd' is denied.
...

t.txt:1:hi

```

Desired behavior
-----------------

The directory wildcard match should be ignored.

```powershell
> Select-String -Path * hi

t.txt:1:hi

```

Environment data
----------------

```powershell
PowerShell Core v6.0.0-beta.6 on macOS 10.12.6
PowerShell Core v6.0.0-beta.6 on Ubuntu 16.04.3 LTS
PowerShell Core v6.0.0-beta.6 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
Windows PowerShell v5.1.15063.502 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
```

iSazonov:
Fix #4820.

Select-String can search in files only so we should skip directory
objects.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
@@ -1347,6 +1347,11 @@ namespace Microsoft.PowerShell.Commands
             {
                 foreach (string filename in expandedPaths)
                 {
+                    if (System.IO.Directory.Exists(filename))
+                    {
+                        continue;
+                    }
+
                     bool foundMatch = ProcessFile(filename);
                     if (_quiet && foundMatch)
                         return;
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
@@ -1347,7 +1347,7 @@ namespace Microsoft.PowerShell.Commands
             {
                 foreach (string filename in expandedPaths)
                 {
-                    if (System.IO.Directory.Exists(filename))
+                    if (Directory.Exists(filename))
                     {
                         continue;
                     }
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
@@ -10,7 +10,7 @@
             $fileNameWithDots = $fileName.FullName.Replace("\", "\.\")
 
             $subDirName = Join-Path $TestDrive 'selectSubDir'
-            New-Item -Path $subDirName -ItemType Directory > $null
+            New-Item -Path $subDirName -ItemType Directory -Force > $null
             $pathWithoutWildcard = $TestDrive
             $pathWithWildcard = Join-Path $TestDrive '*'
 
@@ -30,11 +30,11 @@
         }
 
         It "Select-String does not throw on subdirectory (path without wildcard)" {
-            { select-string -Path  $pathWithoutWildcard "noExists" } | Should Not Throw
+            { select-string -Path  $pathWithoutWildcard "noExists" -ErrorAction Stop } | Should Not Throw
         }
 
         It "Select-String does not throw on subdirectory (path with wildcard)" {
-            { select-string -Path  $pathWithWildcard "noExists" } | Should Not Throw
+            { select-string -Path  $pathWithWildcard "noExists" -ErrorAction Stop } | Should Not Throw
         }
 
         It "LiteralPath with relative path" {
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
@@ -9,6 +9,11 @@
 
             $fileNameWithDots = $fileName.FullName.Replace("\", "\.\")
 
+            $subDirName = Join-Path $TestDrive 'selectSubDir'
+            New-Item -Path $subDirName -ItemType Directory > $null
+            $pathWithoutWildcard = $TestDrive
+            $pathWithWildcard = Join-Path $TestDrive '*'
+
             $tempFile = New-TemporaryFile
             "abc" | Out-File -LiteralPath $tempFile.fullname
 	        "bcd" | Out-File -LiteralPath $tempFile.fullname -Append
@@ -22,6 +27,14 @@
         AfterAll {
             Remove-Item $tempFile -Force -ErrorAction SilentlyContinue
             Pop-Location
+        }
+
+        It "Select-String does not throw on subdirectory (path without wildcard)" {
+            { select-string -Path  $pathWithoutWildcard "noExists" } | Should Not Throw
+        }
+
+        It "Select-String does not throw on subdirectory (path with wildcard)" {
+            { select-string -Path  $pathWithWildcard "noExists" } | Should Not Throw
         }
 
         It "LiteralPath with relative path" {


src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
```diff
@@ -1347,6 +1347,11 @@ protected override void ProcessRecord()
             {
                 foreach (string filename in expandedPaths)
                 {
+                    if (System.IO.Directory.Exists(filename))
```
daxian-dbw: There is a using directive for `System.IO.Directory`, so any reason to use the full type name here?

test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
```diff
@@ -24,6 +29,14 @@
             Pop-Location
         }
 
+        It "Select-String does not throw on subdirectory (path without wildcard)" {
+            { select-string -Path  $pathWithoutWildcard "noExists" } | Should Not Throw
+        }
+
+        It "Select-String does not throw on subdirectory (path with wildcard)" {
+            { select-string -Path  $pathWithWildcard "noExists" } | Should Not Throw
```
daxian-dbw: The current behavior is to throw a non-terminating error, so for this test, it will pass even if it actually failed. see `[+] ee 297ms` below
```
PS:122> describe "aa" { It "ee" { { Select-String -Path * hi } | should not throw } }
Describing aa
Select-String : The file F:\tmp\test\pp cannot be read: Access to the path 'F:\tmp\test\pp' is denied.
At line:1 char:29
+ describe "aa" { It "ee" { { Select-String -Path * hi } | should not t ...
+                             ~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Select-String], ArgumentException
    + FullyQualifiedErrorId : ProcessingFile,Microsoft.PowerShell.Commands.SelectStringCommand

 [+] ee 297ms
```

test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
```diff
@@ -24,6 +29,14 @@
             Pop-Location
         }
 
+        It "Select-String does not throw on subdirectory (path without wildcard)" {
+            { select-string -Path  $pathWithoutWildcard "noExists" } | Should Not Throw
```
daxian-dbw: Should not throw might not be a good test here because it doesn't throw with the current behavior -- a non-terminating error is written out.

daxian-dbw (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
dantraMSFT: I think this test is fine but you should also verify the reported error (ProcessingFile,Microsoft.PowerShell.Commands.SelectStringCommand) either in the test or in another test.

test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
dantraMSFT: Like the above should not throw test, you should verify the expected error (ProcessingFile,Microsoft.PowerShell.Commands.SelectStringCommand)

test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
```diff
@@ -9,6 +9,11 @@
 
             $fileNameWithDots = $fileName.FullName.Replace("\", "\.\")
 
+            $subDirName = Join-Path $TestDrive 'selectSubDir'
+            New-Item -Path $subDirName -ItemType Directory > $null
```
dantraMSFT: Suggest using -ErrorAction Stop.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
iSazonov: Fixed.

test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
iSazonov: I skipped '-ErrorAction Stop' - Fixed.

test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
iSazonov: I see we usualy use `-Force` - I added.
Please clarify about `-ErrorAction Stop` - I don't found such samples in our tests.

daxian-dbw (approved):

test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
dantraMSFT: For me, its all about failure diagnosability.  The test is dependent on the directory and if new-item fails, there's no point in the test continuing.  Using -EA Stop ensures you get the error at the point of the failure instead of later on when it may be muddled by test or product code.
Its just a suggestion.

dantraMSFT (approved):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
```diff
@@ -1347,6 +1347,11 @@ protected override void ProcessRecord()
             {
                 foreach (string filename in expandedPaths)
                 {
+                    if (Directory.Exists(filename))
```
mklement0: Hmm... I actually think we should only ignore directories if they're among the paths that result from _wildcard_ resolution.

If you specify an effectively literal directory path - whether via `-LiteralPath` or via `-Path` and a path that happens not to contain (unescaped) wildcard characters - you should continue to get an error (though arguably a better one than currently).

This will let users know that specifying directories is not supported. With this fix as-is, If you do something like `Select-String foo .`, you'll get no output and no error, which could be misinterpreted as the string not having been found in any files in the directory.

I think the behavior should be analogous to the following `Get-ChildItem` behavior:

```powershell
Get-ChildItem -Directory -LiteralPath /NoSuchDir  # error
Get-ChildItem -Directory -Path /NoSuchDir  # error too, because the path contains no (unescaped) wildcard chars.
Get-ChildItem -Directory -Path /NoSuchDir*  # NO error, because the wildcard expression matched nothing.
```

daxian-dbw: Good point. We can reactivate #4820 and get the behavior fixed. @iSazonov what do you think?

iSazonov: Now `Select-String` behavior is the same as `Unblock-File`. There may be other examples (?)
So I believe it is _another issue_.
I agree that we could fix this with a "breaking change" symptom. If @mklement0 would done the analysis (what cmdlets should we fix?), I'd have solved the Issue.

PR #4795, opened 2017-09-10
validate the expected output package exists when using start-pspackage

SteveL-MSFT:
Fix https://github.com/PowerShell/PowerShell/issues/4760
also fixed small typo

Issue #4760, opened 2017-09-06
Build is green if an MSI could not be produced due to e.g. a WiX compilation error


bergmeister:
[Here](https://ci.appveyor.com/project/PowerShell/powershell/build/6.0.0-beta.6-5027/artifacts) is an example of a build that is green but should have failed because it could not produce an MSI due to a WiX compilation error.
The WiX compilation error can also not be found in the log.

iSazonov:
@bergmeister Thanks for open the Issue. 
We have the same problem with creating all packages. So we can make the Issue more common.

PR #4831, opened 2017-09-13
Make the build output the WiX compilation log if it failed.

bergmeister:
Although PR #4795 fixed the main issue of #4760 (to make the build red if there is a `WiX` compilation error), it did not fix the problem that currently the `WiX` log is not outputted at all and therefore one cannot see the compilation error in the log. This PR fixes this and outputs the `WiX` log only if no `MSI` could be produced because the `WiX` log is quite verbose (around 250 lines) and would be unnecessary noise in a green build.

The reason why the log is currently not displayed in the log is because the resulting object is piped directly to `Write-Verbose` but is of type array and therefore needs to be converted to a string using `Out-String` beforehand.

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -1831,9 +1831,9 @@ function New-MSIPackage
         Write-Error -Message "Package already exists, use -Force to overwrite, path:  $msiLocationPath" -ErrorAction Stop
     }
 
-    & $wixHeatExePath dir  $ProductSourcePath -dr  $productVersionWithName -cg $productVersionWithName -gg -sfrag -srd -scom -sreg -out $wixFragmentPath -var env.ProductSourcePath -v | Write-Verbose
-    & $wixCandleExePath  "$ProductWxsPath"  "$wixFragmentPath" -out (Join-Path "$env:Temp" "\\") -ext WixUIExtension -ext WixUtilExtension -arch x64 -v | Write-Verbose
-    & $wixLightExePath -out $msiLocationPath $wixObjProductPath $wixObjFragmentPath -ext WixUIExtension -ext WixUtilExtension -dWixUILicenseRtf="$LicenseFilePath" -v | Write-Verbose
+    $WiXHeatLog = & $wixHeatExePath dir  $ProductSourcePath -dr  $productVersionWithName -cg $productVersionWithName -gg -sfrag -srd -scom -sreg -out $wixFragmentPath -var env.ProductSourcePath -v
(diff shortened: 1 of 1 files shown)


build.psm1
```diff
@@ -1844,6 +1844,9 @@ function New-MSIPackage
     }
     else
     {
+        $WiXHeatLog   | Out-String | Write-Verbose
+        $WiXCandleLog | Out-String | Write-Verbose
+        $WiXLightLog  | Out-String | Write-Verbose
```
daxian-dbw: I think you probably need `Write-Verbose -Verbose`, otherwise, the verbose message won't be written out unless `New-MSIPackage` is called with `-Verbose`.

daxian-dbw (approved):

build.psm1
bergmeister: Ok. I fixed it. But why is the build script no being called using the `-Verbose` option from the top level? As long as advanced functions are used, the `$VerbosePreference would be propagated through. If you always force verbose output, what is then the difference to `Write-Host`? I am thinking about [this](http://www.jsnover.com/blog/2013/12/07/write-host-considered-harmful/) blog post from Jeffrey.

daxian-dbw: `-Verbose` is not specified when calling [`Invoke-AppveyorFinish`](https://github.com/PowerShell/PowerShell/blob/master/appveyor.yml#L28) and also not specified when calling [`Start-PSPackage`](https://github.com/PowerShell/PowerShell/blob/master/tools/appveyor.psm1#L434), as well as when calling [`New-MSIPackage`](https://github.com/PowerShell/PowerShell/blob/master/tools/packaging/packaging.psm1#L167). Maybe it should be specified at the very top call.

But for this change, we'd better add `-Verbose` anyway because even if `-Verbose` is not specified when calling `New-MSIPackage`, it's still desired to write out the logs when no MSI was produced.

adityapatwardhan (approved):
LGTM

TravisEz13:
Our logs are already very long.  If we are on AppVeyor and not local, it may be better to add these logs as artifacts and reference them, If local we can just reference the logs.
Here are the docs on how to push an artifact
https://www.appveyor.com/docs/build-worker-api/#push-artifact
You can test if we are in AppVeyor using an environment variable:
https://www.appveyor.com/docs/environment-variables/

TravisEz13 (approved):
It's an improvement.  But these should be made into artifacts.

bergmeister:
@TravisEz13 : That's why the WiX log (around 250 lines) gets only output in the case of a failure to not pollute normal green builds.

TravisEz13:
unless the logs are very short, 1-5 line, an artifact would be better.  The logs are very long and become difficult to find details in.

PR #4835, opened 2017-09-14
Enable tab complete for '-Format' parameter of Get-Date

iSazonov:
Add ArgumentCompletions attribute to '-Format' parameter of Get-Date to enable
IntelliSense (tab complete).

## Fix
1. Added ArgumentCompletions attribute to enable IntelliSense but allow not only named format values ("FileDate", "FileDateUniversal", "FileDateTime", "FileDateTimeUniversal") but also custom format strings ("YYYY-MM-dd").
2. Added a commit to replace tabs in test file.
3. Added new ArgumentCompletionsAttribute type which allow tab completions (IntelliSense) for parameters without throw on "invalid" values in contrast to ValidateSet.
4. Added tests for `-Format` parameter with named format values.
5. Added tests for ArgumentCompletionsAttribute

## Additional considerations
No tab completion test added for Get-Date because it seems we have no requirements to check tab completion for every parameter in all cmdlets.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
@@ -206,7 +206,7 @@ namespace Microsoft.PowerShell.Commands
         /// Unix format string
         /// </summary>
         [Parameter(ParameterSetName = "net")]
-        [ArgumentCompleter("FileDate", "FileDateUniversal", "FileDateTime", "FileDateTimeUniversal")]
+        [ArgumentCompletions("FileDate", "FileDateUniversal", "FileDateTime", "FileDateTimeUniversal")]
         public string Format { get; set; }
 
         #endregion
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
@@ -206,7 +206,7 @@ namespace Microsoft.PowerShell.Commands
         /// Unix format string
         /// </summary>
         [Parameter(ParameterSetName = "net")]
-        [ValidateSetAttribute("FileDate", "FileDateUniversal", "FileDateTime", "FileDateTimeUniversal", AllowAll=true)]
+        [ArgumentCompleter("FileDate", "FileDateUniversal", "FileDateTime", "FileDateTimeUniversal")]
         public string Format { get; set; }
 
         #endregion
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
@@ -206,6 +206,7 @@ namespace Microsoft.PowerShell.Commands
         /// Unix format string
         /// </summary>
         [Parameter(ParameterSetName = "net")]
+        [ValidateSetAttribute("FileDate", "FileDateUniversal", "FileDateTime", "FileDateTimeUniversal", AllowAll=true)]
         public string Format { get; set; }
 
         #endregion
--- a/src/System.Management.Automation/engine/Attributes.cs
+++ b/src/System.Management.Automation/engine/Attributes.cs
@@ -1564,6 +1564,14 @@ namespace System.Management.Automation
         private static ConcurrentDictionary<Type, IValidateSetValuesGenerator> s_ValidValuesGeneratorCache = new ConcurrentDictionary<Type, IValidateSetValuesGenerator>();
 
         /// <summary>
+        /// If AllowAll==true 'ValidateElement' always pass all elements.
+        /// Useful if we need enable IntelliSense (TabComplete) but effectively disable validation
+        /// when a parameter can accept not only the listed values but also arbitrary.
+        /// Ex.: Get-Date -Format
+        /// </summary>
+        public bool AllowAll { get; set; } = false;
+
+        /// <summary>
         /// Gets or sets the custom error message that is displayed to the user
         ///
         /// The item being validated and a text representation of the validation set
@@ -1623,6 +1631,11 @@ namespace System.Management.Automation
                             "ArgumentIsEmpty",
                             null,
                             Metadata.ValidateNotNullFailure);
+            }
+
+            if (AllowAll == true)
+            {
+                return;
             }
 
             string objString = element.ToString();
--- a/src/System.Management.Automation/engine/Attributes.cs
+++ b/src/System.Management.Automation/engine/Attributes.cs
@@ -1564,14 +1564,6 @@ namespace System.Management.Automation
         private static ConcurrentDictionary<Type, IValidateSetValuesGenerator> s_ValidValuesGeneratorCache = new ConcurrentDictionary<Type, IValidateSetValuesGenerator>();
 
         /// <summary>
-        /// If AllowAll==true 'ValidateElement' always pass all elements.
-        /// Useful if we need enable IntelliSense (TabComplete) but effectively disable validation
-        /// when a parameter can accept not only the listed values but also arbitrary.
-        /// Ex.: Get-Date -Format
-        /// </summary>
-        public bool AllowAll { get; set; } = false;
-
-        /// <summary>
         /// Gets or sets the custom error message that is displayed to the user
         ///
         /// The item being validated and a text representation of the validation set
@@ -1631,11 +1623,6 @@ namespace System.Management.Automation
                             "ArgumentIsEmpty",
                             null,
                             Metadata.ValidateNotNullFailure);
-            }
-
-            if (AllowAll == true)
-            {
-                return;
             }
 
             string objString = element.ToString();
--- a/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
+++ b/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
@@ -2006,19 +2006,13 @@ namespace System.Management.Automation
             var argumentCompletionsAttribute = parameter.CompiledAttributes.OfType<ArgumentCompletionsAttribute>().FirstOrDefault();
             if (argumentCompletionsAttribute != null)
             {
-                try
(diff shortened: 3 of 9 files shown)


vors:
Should we use another attribute name just for the IntelliSense scenario?
`[ValidateSet(AllowAll)]` looks cumbersome. We can inherit from `ValidateSetAttribute` and use some name that doesn't mention Validation. It's also not very clean, but at least for the end user it will be not as confusing imo.

iSazonov:
ValidateSet is a sealed.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1564,6 +1564,14 @@ public sealed class ValidateSetAttribute : ValidateEnumeratedArgumentsAttribute
         private static ConcurrentDictionary<Type, IValidateSetValuesGenerator> s_ValidValuesGeneratorCache = new ConcurrentDictionary<Type, IValidateSetValuesGenerator>();
 
         /// <summary>
+        /// If AllowAll==true 'ValidateElement' always pass all elements.
+        /// Useful if we need enable IntelliSense (TabComplete) but effectively disable validation
+        /// when a parameter can accept not only the listed values but also arbitrary.
+        /// Ex.: Get-Date -Format
+        /// </summary>
+        public bool AllowAll { get; set; } = false;
```
lzybkr: This doesn't feel right to me - you shouldn't use a `Validation` attribute if you aren't validating.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
```diff
@@ -206,6 +206,7 @@ public int Millisecond
         /// Unix format string
         /// </summary>
         [Parameter(ParameterSetName = "net")]
+        [ValidateSetAttribute("FileDate", "FileDateUniversal", "FileDateTime", "FileDateTimeUniversal", AllowAll=true)]
```
lzybkr: I would suggest using an `ArgumentCompleter` attribute instead - it's a bit more cumbersome, but if this scenario makes sense (and I think it probably does), we can find some way to make it simpler.

lzybkr (changes requested):
Thanks for thinking about the usability here - I agree we should do something here, just not like this.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
iSazonov: Do you suggest enhance `ArgumentCompleter` to support `[ArgumentCompleter("FileDate", "FileDateUniversal", "FileDateTime", "FileDateTimeUniversal")] `?

lzybkr: Just brainstorming here - but maybe a new attribute `ArgumentCompletions` that also implements `IArgumentCompleter` - this probably requires some additional code elsewhere, probably similar to the `ValidateSet` code for completion.

iSazonov: A new attribute requires a lot of code. It seems easiest to add a new constructor `public ArgumentCompleterAttribute(params string[] completeStrings)`

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
```diff
@@ -206,7 +206,7 @@ public int Millisecond
         /// Unix format string
         /// </summary>
         [Parameter(ParameterSetName = "net")]
-        [ValidateSetAttribute("FileDate", "FileDateUniversal", "FileDateTime", "FileDateTimeUniversal", AllowAll=true)]
+        [ArgumentCompleter("FileDate", "FileDateUniversal", "FileDateTime", "FileDateTimeUniversal")]
```
lzybkr: This is much better, but I think still a little weird, I still think it's worth considering adding a new attribute `ArgumentCompletions`. It seems like you could take your class `StringArrayArgumentCompleter`, rename it, make it an attribute, and either duplicate a little bit of logic around `ValidateSet` or `ArgumentCompleter` in the tab completion code.

iSazonov: Is your only thought that ArgumentCompleter is a strange name and you want to rename it?

lzybkr: I wouldn't rename `ArgumentCompleter` even if it is a strange name (which I suppose it is.)

It's a really minor thing, but to me, `ArgumentCompleter` implies some dynamic code to compute the completions, whereas `ArgumentCompletions` is static.

powercode: I have missed that for a long time - the ability to easily complete a set without limiting the parameter to the completions. 

Is it a reasonable compromise that you have to go IArgumentCompleter if you want to provide rich `CompletionResult`s, or should be enable this with `ArgumentCompletions` too?

iSazonov: Done next iteration.

src/System.Management.Automation/engine/CommandCompletion/ExtensibleCompletion.cs
```diff
@@ -158,4 +158,56 @@ protected override void EndProcessing()
...
+        /// </summary>
+        public IEnumerable<CompletionResult> CompleteArgument(string commandName, string parameterName, string wordToComplete, CommandAst commandAst, IDictionary fakeBoundParameters)
+        {
+            var wordToCompletePattern = WildcardPattern.Get(string.IsNullOrWhiteSpace(wordToComplete) ? "*" : wordToComplete + "*", WildcardOptions.IgnoreCase);
+
+            foreach (var str in _completions)
+            {
+                if (wordToCompletePattern.IsMatch(str))
+                {
+                    yield return new CompletionResult(str, str, CompletionResultType.Text, str);
```
lzybkr: Should this be `CompletionResultType.ParameterValue`? I think that's the only place the attribute will have any effect.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -2002,6 +2002,26 @@ internal static string ConcatenateStringPathArguments(CommandElementAst stringAs
                 {
                 }
             }
+
+            var argumentCompletionsAttribute = parameter.CompiledAttributes.OfType<ArgumentCompletionsAttribute>().FirstOrDefault();
+            if (argumentCompletionsAttribute != null)
+            {
+                try
```
lzybkr: You are calling internal code which doesn't throw, so the try/catch is unnecessary.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1625,6 +1633,11 @@ protected override void ValidateElement(object element)
                             Metadata.ValidateNotNullFailure);
             }
 
+            if (AllowAll == true)
```
lzybkr: This should be reverted.

src/System.Management.Automation/engine/Attributes.cs
lzybkr: This should be reverted.

src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -1336,6 +1336,10 @@ private static Attribute NewValidateSetAttribute(AttributeAst ast)
                 {
                     result.IgnoreCase = s_attrArgToBoolConverter.Target(s_attrArgToBoolConverter, argValue);
                 }
+                else if (argumentName.Equals("AllowAll", StringComparison.OrdinalIgnoreCase))
```
lzybkr: This should be reverted.

lzybkr (changes requested):
Just a little cleanup and I think it looks good.

src/System.Management.Automation/engine/CommandCompletion/ExtensibleCompletion.cs
iSazonov: I agree. 
Fixed.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/Attributes.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/parser/Compiler.cs
iSazonov: Fixed.

lzybkr (approved):
Code changes are approved - but we should get a couple of tests for the new `ArgumentCompletions` attribute.

daxian-dbw:
@iSazonov Can you please update the PR description? Looks like the design has changed.

iSazonov:
Could you please help:  C# test failed on "ArgumentCompleter" abbreviation - full name ArgumentCompleterAttribute works. I cannot find how make "ArgumentCompleter" alias.

test/powershell/Language/Parser/ExtensibleCompletion.Tests.ps1
```diff
@@ -355,3 +355,97 @@ Describe "Additional type name completion tests" -Tags "CI" {
...
+        )
+
+        $testCasesCSharp = @(
+            @{ attributeName = "ArgumentCompletions"         ; cmdletName = "Get-ArgumentCompletions"  },
+            @{ attributeName = "ArgumentCompletionsAttribute"; cmdletName = "Get-ArgumentCompletions1" }
+        )
+    }
+
+    AfterAll {
+        #Remove-Module -ModuleInfo $testModule
```
daxian-dbw: Did you forget to uncomment this?

iSazonov: Yes 😄 Today test is passed. I wonder - I don't change a code.

daxian-dbw (approved):

iSazonov:
@lzybkr Thanks for great idea!

Issue #3961, opened 2017-06-07
Build Generic linux-x64 rpm and deb packages


TravisEz13:
Background: See https://github.com/PowerShell/PowerShell/issues/3721#issuecomment-306266883
tl;dr: This should build the assemblies in a way that they are portable to [all the supported Linux x64 platforms for .NET](https://github.com/dotnet/core/blob/master/roadmap.md#net-core-20---supported-os-versions), but requires that we package native binaries which `dotnet publish` will publish.
- [x] Update build so it can build and publish for the runtime `linux-x64`
- [x] Update packaging script to package native binaries published and remove unneeded dependencies
- [x] Work with Microsoft Repo owners to develop a package publishing story.  @raghushantha  and @travisez13 can get you started here.
- [x] Update documentation 
- [x] Create `.tar.gz` package for Linux
- [x] Add documentation about Linux package dependencies
- [x] Update docker files
- [x] Update from `2.0.0-preview1-002106-00` to `2.0.0-preview2-25303-04` that includes https://github.com/dotnet/corefx/pull/19070 to support Ubuntu 17 and Debian 9 (_update: powershell has moved to official .NET Core 2.0 release_)

janvorli:
@TravisEz13 I've located the very first version that supports Ubuntu 17 and Debian 9. It is `2.0.0-preview2-25303-04`.

borgdylan:
Has .NET Core 2.0 fixed ```SystemNative_SysConf```?

TravisEz13:
@borgdylan I see you just filed this issue, https://github.com/dotnet/corefx/issues/21386, about the same time you made this comment.  It would be more appropriate to ask on that issue if it has been fixed.  If you find something out from the .Net Core team that changes our decision to use portable Linux build,  that would be a good time to cross-post back to PowerShell, with a link to the issue.  Thanks

borgdylan:
Please keep tracking https://github.com/dotnet/corefx/issues/21386 . Using 2.0 will keep powershell broken for now until corefx gets fixed. A workaround would be to not use Environment.ProcessorCount et al in the Linux build until this gets fixed.

borgdylan:
The issues arose from a stale copy of System.Native installed system wide. After removing it, powershell starts normally. (tested against beta3)

iSazonov:
@borgdylan Can you repo this? If so please open new Issue for discussion and maybe fix if system-wide assemblies overlap local ones..

borgdylan:
[...]

PR #4841, opened 2017-09-14
Build powershell core using the generic RID 'linux-x64'

daxian-dbw:
Initial PR for #3961, to get the ball rolling.
- [x] Update build so it can build and publish for the runtime linux-x64

Now that openSUSE.42.1 is decommissioned, comparing to updating our build to support SUSE 42.2, it's better to spend our effort on moving toward the generic Linux package.

This PR focuses on enabling build/publish with 'linux-x64'. I also removed `win10-x64` and `win81-x64` RID's given that we produce a single windows package now.
I tried `Start-PSPackage` on Ubuntu after building with the changes of this PR, and the generated package works fine.

The next step is to figure out how to generate single DEB/RPM packages that work for all supported Linux distros.

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -638,17 +638,19 @@ function New-PSOptions {
         if ($Environment.IsLinux) {
             $Runtime = "linux-x64"
         } else {
-            $Runtime = dotnet --info | ForEach-Object {
+            $RID = dotnet --info | ForEach-Object {
                 if ($_ -match "RID") {
                     $_ -split "\s+" | Select-Object -Last 1
                 }
             }
 
-            # We plan to release packages targetting win7-x64 and win7-x86 RIDs,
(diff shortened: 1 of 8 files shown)


TravisEz13:
So, in this PR we build generically but still package for specific platforms?

daxian-dbw:
Yes, it targets the first task only. Need some pointers on how to proceed with the second task.

TravisEz13:
The second task I think will be a good bit of trial and error. General steps:

1. Remove dependencies we now include during packaging ( according to DotNet Core many native libs will be included when you publish)
2. Test the package on applicable platforms
3. Adjust the package based on the result and repeat 2

daxian-dbw:
@TravisEz13 I will take a stab at the second task.

Issue #4823, opened 2017-09-12
New-ModuleManifest should set ModuleVersion to 0.1 by default


joeyaiello:
Given our push for semantic versioning, I think it'd be helpful for the `New-ModuleManifest` defaults to set `ModuleVersion` to 0.1 

I don't *think* this is a breaking change because `-ModuleVersion` provides deterministic behavior. 

Steps to reproduce
------------------

```powershell
New-ModuleManifest -Path ./foo.psd1
```

Expected behavior
-----------------

```none
...
ModuleVersion = '0.1'
...
```

Actual behavior
---------------

```none
...
ModuleVersion = '1.0'
...
```

Environment data
----------------

```powershell
> $PSVersionTable

```

Jaykul:
## Nope, that's totally wrong!

See the spec:

2. A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and MUST NOT contain leading zeroes... [spec](http://semver.org/#spec-item-2)

SteveL-MSFT:
Leading zero means it can't be 001.002.003, but the major version can be 0.  0.1.0 is probably more correct.

```none
4. Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.
```

mklement0:
Aside from fixing the syntax issue (`0.1` -> `0.1.0` - though casting to `[semver]` curiously works with `0.1` too, and even with just `0`), can I suggest even more humble beginnings, namely `0.0.1`?

At least in the world of Node.js / npm, the `0.0.x` range is nowadays used for early development - see [here](http://markup.im/#q4_cRZ1Q) for background info.

iSazonov:
`0.1.0` looks good to me. We actually use it now, so we just need to add zero.

SteveL-MSFT:
SemVer recommends `0.1.0` although it doesn't say `0.0.1` isn't allowed.  Personally, I prefer `0.1.0` as the default and someone could always explicitly use `0.0.1` if they choose.

mklement0:
@SteveL-MSFT:

`0.1.0` is definitely an improvement over `1.0` and, as you state, is what SemVer recommends, so the following is just food for thought:

[SemVer](http://semver.org/) also imposes virtually _no_ rules on any version < `1.0.0`:

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

[...]

PR #4842, opened 2017-09-14
Updated default ModuleVersion in ModuleManifest

LDSpits:
This pull request updates the New-ModuleManifest command to generate a ModuleVersion manifest property with 0.0.1 instead of 1.0. specified by the discussion in #4823.

### DIFF
--- a/src/System.Management.Automation/engine/Modules/NewModuleManifestCommand.cs
+++ b/src/System.Management.Automation/engine/Modules/NewModuleManifestCommand.cs
@@ -123,7 +123,7 @@ namespace Microsoft.PowerShell.Commands
             get { return _moduleVersion; }
             set { _moduleVersion = value; }
         }
-        private Version _moduleVersion = new Version(1, 0);
+        private Version _moduleVersion = new Version(0, 0, 1);
 
         /// <summary>
         /// Set the module description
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/engine/Modules/NewModuleManifestCommand.cs
```diff
@@ -940,12 +940,12 @@ protected override void EndProcessing()
...
+                    filePath,
 #if UNIX
-                    new UTF8Encoding(false), // UTF-8, no BOM
-#else
-                    EncodingConversion.Unicode, // UTF-16 with BOM
-#endif
+                    new UTF8Encoding(false), // UTF-8, no BOM
+#else
+                    EncodingConversion.Unicode, // UTF-16 with BOM
+#endif
```
daxian-dbw: Is this part of change caused by some character encoding? If no intentional change was made, you can get rid of it by using `git reset HEAD^` and then `git add -p` (ignore this change section when adding).

daxian-dbw (changes requested):

src/System.Management.Automation/engine/Modules/NewModuleManifestCommand.cs
SteveL-MSFT: As part of https://github.com/PowerShell/PowerShell/pull/4761#issuecomment-327852126, this file has mixed EOL characters.  We can accept it to make the EOLs the same for this file or ignore it as I'm fixing this across the entire codebase in a separate PR.

daxian-dbw: OK, then we can accept this part of the changes.

daxian-dbw (approved):

adityapatwardhan (approved):
LGTM

Issue #4843, opened 2017-09-14
Credentials for Set-Service command


joandrsn:
Currently when creating a service using the `New-Service` from the module `Microsoft.PowerShell.Management` I can specify the credentials that my new service will be executed with.

`Set-Service` currently does not support setting the new logon account for the service.
The current workarounds that I have found:
* Use sc.exe to change credentials
* Use `Get-WmiObject -Class Win32_Service` and from there use the `Change` method
* Delete the service and create it again.

But I believe it would be nice to have `Set-Service` support credentials.
Is it possible to add functionality for setting the credentials using the `Set-Service` command?

SteveL-MSFT:
Shouldn't be too hard as Set-Service uses [ChangeServiceConfig](https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs#L1704) win32 api which accepts username/password.

joandrsn:
#4844 was merged

PR #4844, opened 2017-09-15
Added functionality to set credientials on Set-Service command

joandrsn:
Added functionality to set credientials on `Set-Service` command.
Using this you can now specifiy the `-Credential` parameter which makes the service you have specfied change the log-on account.

Issue #4843

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
@@ -1472,12 +1472,7 @@ namespace Microsoft.PowerShell.Commands
         /// <value></value>
         [Parameter]
         [Credential()]
-        public PSCredential Credential
-        {
-            get { return credential; }
-            set { credential = value; }
-        }
-        internal PSCredential credential = null;
+        public PSCredential Credential { get; set; }
 
 
 
@@ -1693,7 +1688,7 @@ namespace Microsoft.PowerShell.Commands
                             continue;
                         }
 
-                        // modify startup type or display name or credential
+                        // Modify startup type or display name or credential
                         if (!String.IsNullOrEmpty(DisplayName)
                             || (ServiceStartMode)(-1) != StartupType || null != Credential) 
                         {
@@ -1716,7 +1711,6 @@ namespace Microsoft.PowerShell.Commands
                                     break;
                             }
 
-                            // set up the Credential parameter
                             string username = null;
                             if (null != Credential)
                             {
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
@@ -1466,6 +1466,19 @@ namespace Microsoft.PowerShell.Commands
         }
         internal string displayName = null;
 
+        /// <summary>
+        /// Account under which the service should run
+        /// </summary>
+        /// <value></value>
+        [Parameter]
+        [Credential()]
+        public PSCredential Credential
+        {
+            get { return credential; }
+            set { credential = value; }
+        }
+        internal PSCredential credential = null;
+
 
 
         /// <summary>
@@ -1589,6 +1602,7 @@ namespace Microsoft.PowerShell.Commands
         {
             ServiceController service = null;
             string ServiceComputerName = null;
+            IntPtr password = IntPtr.Zero;
             foreach (string computer in ComputerName)
             {
                 bool objServiceShouldBeDisposed = false;
@@ -1679,9 +1693,9 @@ namespace Microsoft.PowerShell.Commands
                             continue;
                         }
 
-                        // modify startup type or display name
+                        // modify startup type or display name or credential
                         if (!String.IsNullOrEmpty(DisplayName)
-                            || (ServiceStartMode)(-1) != StartupType)
+                            || (ServiceStartMode)(-1) != StartupType || null != Credential) 
                         {
                             DWORD dwStartType = NativeMethods.SERVICE_NO_CHANGE;
                             switch (StartupType)
@@ -1701,6 +1715,14 @@ namespace Microsoft.PowerShell.Commands
                                         "bad StartupType");
                                     break;
                             }
+
+                            // set up the Credential parameter
+                            string username = null;
+                            if (null != Credential)
+                            {
+                                username = Credential.UserName;
+                                password = Marshal.SecureStringToCoTaskMemUnicode(Credential.Password);
+                            }
                             bool succeeded = NativeMethods.ChangeServiceConfigW(
                                 hService,
                                 NativeMethods.SERVICE_NO_CHANGE,
@@ -1710,8 +1732,8 @@ namespace Microsoft.PowerShell.Commands
                                 null,
                                 IntPtr.Zero,
                                 null,
-                                null,
-                                IntPtr.Zero,
+                                username,
+                                password,
                                 DisplayName
                                 );
                             if (!succeeded)
@@ -1847,6 +1869,10 @@ namespace Microsoft.PowerShell.Commands
                 } //End try
                 finally
                 {
+                    if (IntPtr.Zero != password)
+                    {
+                        Marshal.ZeroFreeCoTaskMemUnicode(password);
+                    }
                     if (objServiceShouldBeDisposed)
                     {
                         service.Dispose();
--- a/test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
@@ -110,35 +110,37 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
         $newServiceCommand.$parameter | Should Be $value
     }
 
-    It "Set-Service can change credentials of a service" -TestCases @(
-        @{name = "testsetcredential"; startUsername = "user1"; endUsername = "user2"}
-    ) {
-        param($name, $startUsername, $endUsername)
+    It "Set-Service can change credentials of a service" {
+        param()
         try {
+            $startUsername = "user1"
+            $endUsername = "user2"
             $testPass = "Secret123!"
-            $null = net user $startUsername $testPass /add
-            $null = net user $endUsername $testPass /add
+            $servicename = "testsetcredential"
+            net user $startUsername $testPass /add > $null
+            net user $endUsername $testPass /add > $null
             $password = ConvertTo-SecureString $testPass -AsPlainText -Force
             $creds = [pscredential]::new(".\$startUsername", $password)
             $parameters = @{
-                Name           = $name;
+                Name           = $servicename;
                 BinaryPathName = "$PSHOME\powershell.exe";
                 StartupType    = "Manual";
                 Credential     = $creds
             }
             $service = New-Service @parameters
             $service | Should Not BeNullOrEmpty
-            $service = Get-CimInstance Win32_Service -Filter "name='$name'"
+            $service = Get-CimInstance Win32_Service -Filter "name='$servicename'"
             $service.StartName | Should BeExactly $creds.UserName
+
             $creds = [pscredential]::new(".\$endUsername", $password)
-            Set-Service -Name $name -Credential $creds
-            $service = Get-CimInstance Win32_Service -Filter "name='$name'"
+            Set-Service -Name $servicename -Credential $creds
+            $service = Get-CimInstance Win32_Service -Filter "name='$servicename'"
             $service.StartName | Should BeExactly $creds.UserName
         }
         finally {
-            Get-CimInstance Win32_Service -Filter "name='$name'" | Remove-CimInstance -ErrorAction SilentlyContinue
-            $null = net user $startUsername /delete
-            $null = net user $endUsername /delete
+            Get-CimInstance Win32_Service -Filter "name='$servicename'" | Remove-CimInstance -ErrorAction SilentlyContinue
+            net user $startUsername /delete > $null
+            net user $endUsername /delete > $null
         }
     }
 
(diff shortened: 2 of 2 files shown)


src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -1466,6 +1466,19 @@ public new string DisplayName
...
+        /// </summary>
+        /// <value></value>
+        [Parameter]
+        [Credential()]
+        public PSCredential Credential
+        {
+            get { return credential; }
+            set { credential = value; }
+        }
+        internal PSCredential credential = null;
```
iSazonov: For the new code, it is preferable to use auto property. Especially because `credential` is not used.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -1679,9 +1693,9 @@ protected override void ProcessRecord()
                             continue;
                         }
 
-                        // modify startup type or display name
+                        // modify startup type or display name or credential
```
iSazonov: Please start comments with capitalized char. We shouldn't add bad patterns.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -1701,6 +1715,14 @@ protected override void ProcessRecord()
                                         "bad StartupType");
                                     break;
                             }
+
+                            // set up the Credential parameter
```
iSazonov: Please remove obvious comment.

iSazonov (commented):
@joandrsn Many thanks for your contribution!

Please add a description in the PR.
Also please add tests (see Set-Service.Tests.ps1).

joandrsn:
@iSazonov thank you for the feedback. I have updated my pull request. But I still need to create tests for the code I have written.
Unfortunatly I don't know the credentials of the computer that the tests will be running on, so I'm not sure how to write the testcase.
Do you have any documentation on what I can and can't write in the tests?

iSazonov:
@joandrsn Please see [tests for New-Service](https://github.com/PowerShell/PowerShell/blob/master/test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1#L104)
I think you should (1) create new test service (WMI?) (2) check that you can change its properties. (3) remove the test service.

Also you should add "[Feature]" in header first position of last commit that CI start the _feature_ tests.

joandrsn:
@iSazonov I have now added the test, but when executing the test with what it is written within the test, you will receive an error `The account name is invalid or does not exist, or the password is invalid for the account name specified` since the users `username` and `username2` does not exist.

I have tested with my own 2 users and it seems to work, but is this okay?

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -110,6 +110,39 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
         $newServiceCommand.$parameter | Should Be $value
     }
 
+    It "Set-Service can create a new service called '<name>' and change the credentials" -TestCases @(
```
iSazonov: "Set-Service can create a new service" - it is incorrect.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -110,6 +110,39 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
...
+            $service.StartName | Should Be $startCredential.UserName
+            Set-Service -Name $name -Credential $endCredential
+            $service = Get-CimInstance Win32_Service -Filter "name='$name'"
+            $service | Should Not BeNullOrEmpty
+            $service.StartName | Should Be $endCredential.UserName
+        }
+        finally {
+            $service = Get-CimInstance Win32_Service -Filter "name='$name'"
+            if ($service -ne $null) {
+                $service | Remove-CimInstance
```
iSazonov: Please add "-Force -ErrorAction SilentlyContinue"

iSazonov:
@joandrsn I think we should try to create a real user account for the test with "Log on as a service" privilege.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
joandrsn: Neither `Get-CimInstance` nor `Remote-CimInstance` has a force parameter. But I can add `-ErrorAction SilentlyContinue`

joandrsn:
@iSazonov 
I figured out that I could use some built in users from Windows.
Currently I'm using `.\Guest` when creating the service and `.\Administrator` when changing the credentials. Would you rather have it create a user and delete the user again?

iSazonov:
Guest and Administrator accounts haven't "Log on as a service" privilege and we don't know password. So we should create test users.

joandrsn:
They don't have the access, but the test passes. I guess that might not be a good way to test (using invalid credentials to validate a test).

Do you have suggestion on how you would add the user? `Add-LocalUser` has been disabled since #4272. I can't find any other test modules in the repository which creates and deletes a user.

SteveL-MSFT:
@joandrsn thanks for the contribution!  Look at what I've done [here](https://github.com/PowerShell/PowerShell/pull/4756/files#diff-c2f140f155813763bf7a3492a6a03196) to create my own user and removing at end of test

iSazonov:
For PowerShell team: We haven't tests to really create new service and change its properties. So we should refactor all tests after the PR will be merged. Maybe create tracking Issue?

SteveL-MSFT:
@iSazonov it seems the test case added by @joandrsn does a `new` and then `set`, were you thinking of something different?

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -110,6 +110,43 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
...
+            $parameters = @{
+                Name           = $name;
+                BinaryPathName = "$PSHOME\powershell.exe";
+                StartupType    = "Manual";
+                Credential     = $creds
+            }
+            $service = New-Service @parameters
+            $service | Should Not BeNullOrEmpty
+            $service = Get-CimInstance Win32_Service -Filter "name='$name'"
+            $service | Should Not BeNullOrEmpty
```
SteveL-MSFT: this check isn't necessary as the next line will fail

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -110,6 +110,43 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
...
+            $service.StartName | Should Be $creds.UserName
```
SteveL-MSFT: BeExactly?

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -110,6 +110,43 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
...
+            $creds = [pscredential]::new(".\$endUsername", $password)
+            Set-Service -Name $name -Credential $creds
+            $service = Get-CimInstance Win32_Service -Filter "name='$name'"
+            $service | Should Not BeNullOrEmpty
```
SteveL-MSFT: This check isn't necessary

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -111,35 +111,39 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
...
-            endCredential = [System.Management.Automation.PSCredential]::new(".\Administrator", 
-                (ConvertTo-SecureString "PlainTextPassword" -AsPlainText -Force))}
+        @{name = "testsetcredential"; startUsername = "user1"; endUsername = "user2"}
     ) {
-        param($name, $startCredential, $endCredential)
+        param($name, $startUsername, $endUsername)
         try {
+            $testPass = "Secret123!"
+            $null = net user $startUsername $testPass /add
+            $null = net user $endUsername $testPass /add
```
iSazonov: Minor comment - we usually use the pattern  `net user $startUsername $testPass /add > $null`

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -110,6 +110,43 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
...
+            $service | Should Not BeNullOrEmpty
+            $service.StartName | Should Be $creds.UserName
+            $creds = [pscredential]::new(".\$endUsername", $password)
+            Set-Service -Name $name -Credential $creds
+            $service = Get-CimInstance Win32_Service -Filter "name='$name'"
+            $service | Should Not BeNullOrEmpty
+            $service.StartName | Should Be $creds.UserName
+        }
+        finally {
+            $service = Get-CimInstance Win32_Service -Filter "name='$name'" -ErrorAction SilentlyContinue
```
SteveL-MSFT: You can simply this to just:

```powershell
Get-CimInstance ... | Remove-CimInstance -ErrorAction SilentlyContinue
```

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -111,35 +111,39 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
     }
 
     It "Set-Service can change credentials of a service" -TestCases @(
-        @{name = "testsetcredential"; 
-            startCredential = [System.Management.Automation.PSCredential]::new(".\Guest", 
-                (ConvertTo-SecureString "PlainTextPassword" -AsPlainText -Force)); 
-            endCredential = [System.Management.Automation.PSCredential]::new(".\Administrator", 
-                (ConvertTo-SecureString "PlainTextPassword" -AsPlainText -Force))}
+        @{name = "testsetcredential"; startUsername = "user1"; endUsername = "user2"}
```
iSazonov: It is single test and we should remove TestCases.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -111,35 +111,39 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
...
+                Credential     = $creds
             }
             $service = New-Service @parameters
             $service | Should Not BeNullOrEmpty
             $service = Get-CimInstance Win32_Service -Filter "name='$name'"
             $service | Should Not BeNullOrEmpty
-            $service.StartName | Should Be $startCredential.UserName
-            Set-Service -Name $name -Credential $endCredential
+            $service.StartName | Should Be $creds.UserName
+            $creds = [pscredential]::new(".\$endUsername", $password)
```
iSazonov: Please add new line to split into logical blocks.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -111,35 +111,39 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
...
-            $service.StartName | Should Be $endCredential.UserName
+            $service.StartName | Should Be $creds.UserName
         }
         finally {
             $service = Get-CimInstance Win32_Service -Filter "name='$name'" -ErrorAction SilentlyContinue
             if ($service -ne $null) {
                 $service | Remove-CimInstance -ErrorAction SilentlyContinue
             }
+            $null = net user $startUsername /delete
+            $null = net user $endUsername /delete
```
iSazonov: The same about $null

SteveL-MSFT (approved):

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -110,6 +110,40 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
         $newServiceCommand.$parameter | Should Be $value
     }
 
+    It "Set-Service can change credentials of a service" {
+        param()
```
iSazonov: Please remove.

iSazonov (approved):
@joandrsn Many thanks!

I'll merge on next week.

iSazonov:
>it seems the test case added by @joandrsn does a  new  and then  set , were you thinking of something different?
@SteveL-MSFT I've analyzed these tests more closely.
I don't understand why we need test "SetServiceCommand can be used as API for '<parameter>' with '<value>'". It seems we should do the checks on real service by invoke.
It seems we don't check that New- and Set-Service ignore System and Boot in StartupType.

SteveL-MSFT:
@iSazonov The "use as API" tests are specifically for covering the getter/setters for the parameters which is only used when called as an API and not when used as cmdlet parameter.  If we're missing some variations, we should add those, but perhaps as a different PR.

iSazonov:
@joandrsn Many thanks for your contribution!

PR #4847, opened 2017-09-15
Call 'DoComplete' when handling FileRedirection for CommandExpression

### LINKED ISSUE #4812: Direct redirection of a custom object ([pscustomobject] instance) to a file with ">" produces no output.
Even though `>` is not an _actual_ alias of `Out-File`, it usually is an _effective_ alias; the two generally behave the same.

However, when outputting a `[pscustomobject]` instance, `>` apparently produces an empty file (the file contains only the UTF16-LE BOM and a newline):

Steps to reproduce
------------------

```powershell
# Use Out-File explicitly.
[pscustomobject] @{ one = 1 } | Out-File tmp.txt; Get-Content tmp.txt

'---'

# Use > instead of Out-File, which should produce the same result.
[pscustomobject] @{ one = 1 } > tmp.txt; Get-Content tmp.txt
```

Expected behavior
-----------------

```none

one
---
  1

---

one
---
  1
```

Actual behavior
---------------

```none

one
---
  1

---

```

I.e., using `>`  instead of `Out-File` resulted in an empty file (just the UTF16-LE BOM and a newline).

Environment data
----------------

```powershell
PowerShell Core v6.0.0-beta.6 on macOS 10.12.6
PowerShell Core v6.0.0-beta.6 on Ubuntu 16.04.3 LTS
PowerShell Core v6.0.0-beta.6 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
Windows PowerShell v5.1.15063.502 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
```

daxian-dbw:
Fix #4812 

Summary
-----------

When handling file redirection for CommandExpression, we don't call 'DoComplete' on the underlying `PipelineProcessor` of the `FileRedirection` object, and thus the `EndProcessing` method is not called on `Out-File`, which causes different behaviors between `<expr> > out.txt` and `<expr> | Out-File out.txt`.

The fix is to make sure 'DoComplete' is called after the stream output has been written to the redirection pipe.

Also fix another issue
------------------------
This PR also fixes an issue that could mess up restoring the original pipes. Here is the repro:
```
PS> 1 *> b.txt > a.txt; 123
Cannot perform operation because object "PipelineProcessor" has already been disposed
```
The root cause is that we don't always restore pipes in the correct order. Please see the code changes in `Compiler.cs` for more details.

/cc @mklement0

### DIFF
--- a/src/System.Management.Automation/engine/parser/Compiler.cs
+++ b/src/System.Management.Automation/engine/parser/Compiler.cs
@@ -3219,7 +3219,7 @@ namespace System.Management.Automation.Language
                 // Expressions for this purpose are kept in 'extraFileRedirectExprs' and will be used later.
                 extraFileRedirectExprs.Add(Expression.Call(redirectionExpr, CachedReflectionInfo.FileRedirection_CallDoCompleteForExpression));
 
-                // The 'UnBind' and 'Dispose' operations on 'FileRedirection' objects must be done in the reversed order of 'Bind' operations.
+                // The 'UnBind' and 'Dispose' operations on 'FileRedirection' objects must be done in the reverse order of 'Bind' operations.
                 // Namely, it should be done is this order:
                 //   try {
                 //       // The order is A, B
--- a/src/System.Management.Automation/engine/parser/Compiler.cs
+++ b/src/System.Management.Automation/engine/parser/Compiler.cs
@@ -181,6 +181,8 @@ namespace System.Management.Automation.Language
 
         internal static readonly MethodInfo FileRedirection_BindForExpression =
             typeof(FileRedirection).GetMethod(nameof(FileRedirection.BindForExpression), instanceFlags);
+        internal static readonly MethodInfo FileRedirection_CallDoCompleteForExpression =
+            typeof(FileRedirection).GetMethod(nameof(FileRedirection.CallDoCompleteForExpression), instanceFlags);
         internal static readonly ConstructorInfo FileRedirection_ctor =
             typeof(FileRedirection).GetConstructor(instanceFlags, null, CallingConventions.Standard,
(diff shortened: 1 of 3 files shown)


daxian-dbw:
@lzybkr Can you please take a look when you have time? Thanks!

src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -3194,35 +3197,55 @@ private Expression GetRedirectedExpression(CommandExpressionAst commandExpr, boo
...
-                                                    CachedReflectionInfo.CommandRedirection_UnbindForExpression,
-                                                    _functionContext,
-                                                    savedPipes));
+
+                // We need to call 'DoComplete' on the file redirection pipeline processor after writing the stream output to redirection pipe,
+                // so that the 'EndProcessing' method of 'Out-File' would be called as expected.
+                // Expressions for this purpose are kept in 'extraFileRedirectExprs' and will be used later.
+                extraFileRedirectExprs.Add(Expression.Call(redirectionExpr, CachedReflectionInfo.FileRedirection_CallDoCompleteForExpression));
+
+                // The 'UnBind' and 'Dispose' operations on 'FileRedirection' objects must be done in the reversed order of 'Bind' operations.
```
lzybkr: done in the *reverse* order

lzybkr (approved):
This code has always bothered me - it's more complicated than a redirected command.

I wonder how possible it would be to move most of this logic into a helper method and avoid generating so much complicated code.

src/System.Management.Automation/engine/parser/Compiler.cs
daxian-dbw: Fixed. Thanks!

daxian-dbw:
I opened the issue #4879 to tracking the effort to revisit this implementation.

PR #4850, opened 2017-09-16
Void codemethod call should not crash

### LINKED ISSUE #4826: CodeMethod invoke fails with dot syntax but works with Foreach-Object -MemberName
Steps to reproduce
------------------

```powershell
using namespace System.Management.Automation
class Y{
    [int] $Id
}

class X{    
	static [void] OpenUri([psobject] $obj) {
		$id = $obj.Id
		start-process "http://baseuri/$id"
	}

    static [void] UpdateTypeData(){
        $td = [Runspaces.Typedata]::new([Y])		
		$td.Members.Add("OpenUri",
			[Runspaces.CodeMethodData]::new("OpenUri", [X]::GetMethodInfo([X], "OpenUri")));
		Update-TypeData -TypeData $td -Force
    }
    
    hidden static [Reflection.MethodInfo] GetMethodInfo([Type] $type, [string] $method) {
		return $type.GetMethod($method, [Reflection.BindingFlags]::Static -bor [Reflection.BindingFlags]::Public -bor [Reflection.BindingFlags]::IgnoreCase);
	}
}
[X]::UpdateTypeData()

$y = [Y]::new()

#works
$y | % OpenUri
#fails with No coercion operator is defined between types 'System.Void' and 'System.Object'.
$y.OpenUri()

```

Expected behavior
-----------------
Same behavior invoking method with dot syntax as when invoked via `Foreach-Object -MemberName`

Actual behavior
---------------
fails with
```
No coercion operator is defined between types 'System.Void' and 'System.Object'.
At line:1 char:1
+ $y.OpenUri()
```
```
PS C:\Program Files\PowerShell\6.0.0-beta.6> $y | gm

   TypeName: Y

Name        MemberType Definition
----        ---------- ----------
OpenUri     CodeMethod static void OpenUri(psobject obj)
```

Environment data
----------------
All versions (beta 6 and win 5.1)

powercode:
Fixes #4826 

Adding a check if the method to be called is void, and in that case wrap it in a block that returns null.

Looking at some other code, there seems to be a `CSharpBinderFlags.ResultDiscarded` when they to this in C#. Can reviewers please check if this or something similar needs to be used here?

### DIFF
--- a/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
+++ b/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
@@ -6571,7 +6571,7 @@ namespace System.Management.Automation.Language
                 Expression expr = InvokeMethod(codeMethod.CodeReference, null, args.Prepend(target).ToArray(), false, MethodInvocationType.Ordinary);
                 if (codeMethod.CodeReference.ReturnType == typeof(void))
                 {
-                    expr = Expression.Block(expr, Expression.Default(typeof(object)));
+                    expr = Expression.Block(expr, ExpressionCache.AutomationNullConstant);
                 }
                 else
                 {
--- a/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
+++ b/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
@@ -6568,9 +6568,16 @@ namespace System.Management.Automation.Language
             var codeMethod = methodInfo as PSCodeMethod;
             if (codeMethod != null)
             {
-                return new DynamicMetaObject(
-                    InvokeMethod(codeMethod.CodeReference, null, args.Prepend(target).ToArray(), false, MethodInvocationType.Ordinary)
-                    .Cast(typeof(object)), restrictions).WriteToDebugLog(this);
+                Expression expr = InvokeMethod(codeMethod.CodeReference, null, args.Prepend(target).ToArray(), false, MethodInvocationType.Ordinary);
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/engine/runtime/Binding/Binders.cs
```diff
@@ -6568,9 +6568,16 @@ public override DynamicMetaObject FallbackInvokeMember(DynamicMetaObject target,
             var codeMethod = methodInfo as PSCodeMethod;
             if (codeMethod != null)
             {
-                return new DynamicMetaObject(
-                    InvokeMethod(codeMethod.CodeReference, null, args.Prepend(target).ToArray(), false, MethodInvocationType.Ordinary)
-                    .Cast(typeof(object)), restrictions).WriteToDebugLog(this);
+                Expression expr = InvokeMethod(codeMethod.CodeReference, null, args.Prepend(target).ToArray(), false, MethodInvocationType.Ordinary);
+                if (codeMethod.CodeReference.ReturnType == typeof(void))
+                {
+                    expr = Expression.Block(expr, Expression.Default(typeof(object)));
```
lzybkr: This should be:
```C#
                    expr = Expression.Block(expr, ExpressionCache.AutomationNullConstant);
```
Otherwise it might be possible to use the return value.

lzybkr (changes requested):
Just one small change and I think it's good.

lzybkr (approved):

Issue #4845, opened 2017-09-15
Allow Capture of Response Headers with Invoke-RestMethod


markekraus:
# Problem

`Invoke-RestMethod` is extremely useful for working with APIs that return parsable objects. My single gripe has been that it is not possible to capture the response headers. Many API's provide (sometimes critical) details via the response headers, especially rate-limiting quotas and session timeouts. In situations where information in the the response headers are required, the user must currently fall back to `Invoke-WebRequest` and then perform their own Content-Type detection and response-to-object conversion.

# Proposal

Add a `-ResponseHeadersVariable` parameter to the Web Cmdlets that accepts string. this string will be used as the name of a variable in the calling scope. This variable will be set with the same contents of `BasicHtmlWebResponseObject.Headers`. The user can then access the variable to retrieve the headers from the previews call to `Invoke-RestMethod`

Example:

```powershell
$GetData = Invoke-RestMethod -Uri 'http://localhost:8083/Get/' -ResponseHeadersVariable 'ResonseHeaders'
$ResponseDate = [DateTime]$ResonseHeaders.Date[0]
```

This example converts the `Date` response header to a `DateTime`.

The functionality would be similar to `-SessionVariable`. This would be available for both Web Cmdlets, though it would be redundant on `Invoke-WebRequest` as it would be a mirror of the returned `Headers` property.

SteveL-MSFT:
This is a great idea!

PR #4853, opened 2017-09-16
Initialize Headers Dictionary Only Once

### LINKED ISSUE #3842: WebCmdlets: WebResponseObject.CoreClr.cs every get on Headers creates a new Dictionary
The Headers property creates a new Dictionary which then gets populated.  Users typically do something like

```powershell
if (obj.Headers.ContainsKey("Something"))
{
    obj.Headers["Something"] ...
}
```

which now instantiates a new Dictionary twice, we should optimize this code.  consider using a `ReadOnlyDictionary`

markekraus:
Switch the `WebResponseObject.Headers` Dictionary to initialize once instead of creating a new dictionary on every get. Moved logic to `WebResponseHelper.GetHeadersDictionary()` as this code will be reused outside of `WebResponseObject`.

Closes #3842 

This changes internal functionality only. Existing tests for the `Headers` property are sufficient.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebResponseHelper.CoreClr.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebResponseHelper.CoreClr.cs
@@ -4,6 +4,8 @@
 Copyright (c) Microsoft Corporation.  All rights reserved.
 --********************************************************************/
 
+using System;
+using System.Collections.Generic;
 using System.Net.Http;
 using System.Globalization;
 
@@ -15,6 +17,24 @@ namespace Microsoft.PowerShell.Commands
         {
             string characterSet = response.Content.Headers.ContentType.CharSet;
             return characterSet;
+        }
+        
+        internal static Dictionary<string, IEnumerable<string>> GetHeadersDictionary(HttpResponseMessage response)
+        {
+            var headers = new Dictionary<string, IEnumerable<string>>(StringComparer.OrdinalIgnoreCase);
+            foreach (var entry in response.Headers)
+            {
+                headers[entry.Key] = entry.Value;
+            }
+            if (response.Content != null)
+            {
+                foreach (var entry in response.Content.Headers)
+                {
+                    headers[entry.Key] = entry.Value;
+                }
+            }
+
+            return headers;
         }
 
         internal static string GetProtocol(HttpResponseMessage response)
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebResponseHelper.CoreClr.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebResponseHelper.CoreClr.cs
@@ -26,6 +26,10 @@ namespace Microsoft.PowerShell.Commands
             {
                 headers[entry.Key] = entry.Value;
             }
+            // In CoreFX, HttpResponseMessage separates content related headers, such as Content-Type to 
+            // HttpResponseMessage.Content.Headers. The remaining headers are in HttpResponseMessage.Headers.
+            // The keys in both should be unique with no duplicates between them.
+            // Added for backwards compatibility with PowerShell 5.1 and earlier.
             if (response.Content != null)
             {
                 foreach (var entry in response.Content.Headers)
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebResponseObject.CoreClr.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebResponseObject.CoreClr.cs
@@ -33,18 +33,7 @@ namespace Microsoft.PowerShell.Commands
             {
                 if(_headers == null)
                 {
-                    _headers = new Dictionary<string, IEnumerable<string>>(StringComparer.OrdinalIgnoreCase);
-                    foreach (var entry in BaseResponse.Headers)
-                    {
-                        _headers[entry.Key] = entry.Value;
-                    }
-                    if (BaseResponse.Content != null)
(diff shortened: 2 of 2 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebResponseObject.CoreClr.cs
```diff
@@ -31,23 +31,28 @@ public partial class WebResponseObject
...
+                            headers[entry.Key] = entry.Value;
+                        }
+                    }
                 }
 
                 return headers;
             }
         }
 
+        private Dictionary<string, IEnumerable<string>> headers = null;
```
iSazonov: Should it be `_headers`? [naming-conventions](https://github.com/PowerShell/PowerShell/blob/ede580e7d19c1aecdb708ba9a09a7beaee5a0a14/docs/dev-process/coding-guidelines.md#naming-conventions)

markekraus: Fixed, thanks.

markekraus:
I realized in planning for #4845 the headers dictionary logic would need to be available outside of `WebResponseObject` so I have moved it to `WebResponseHelper`.

iSazonov:
@markekraus Have we already tests? If so please add the comment in PR description. And about WebResponseHelper too.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebResponseHelper.CoreClr.cs
```diff
@@ -16,6 +18,24 @@ internal static string GetCharacterSet(HttpResponseMessage response)
...
+            var headers = new Dictionary<string, IEnumerable<string>>(StringComparer.OrdinalIgnoreCase);
+            foreach (var entry in response.Headers)
+            {
+                headers[entry.Key] = entry.Value;
+            }
+            if (response.Content != null)
+            {
+                foreach (var entry in response.Content.Headers)
+                {
+                    headers[entry.Key] = entry.Value;
```
iSazonov: Interesting, can there be duplicate headers in `response.Headers` and `response.Content.Headers` ? Have second ones a high ptiority?

markekraus: No. The headers in `Response.Content.Headers` will never exist in `response.Headers`. #4494 added this to address the lack of `Content-Type` in the `WebResponseObject.Headers`

markekraus:
@iSazonov I have updated the PR description RE: Testing and move to `WebResponseHelper`.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebResponseHelper.CoreClr.cs
iSazonov: Please clarify - my understanding from #4494 is that `headers` already contains `response.Content.Headers`, correct?

markekraus: That was add in #4494. Previous to that, `WebResponseObject.Headers` did not contain any of the headers in `HttpResponseMessage.Content.Headers`. CoreFX has split the response headers so that those related to content are always in `HttpResponseMessage.Content.Headers` and the rest of the response headers are in `HttpResponseMessage.Headers`. 

To clarify, I'm not adding or changing any of the logic in this PR for the creation of the Dictionary. Just moving it so that a new dictionary is not created on every Get to `Headers` and to make the logic available outside of `WebResponseObject`. The existing logic is there on purpose and we have existing tests to ensure it.

iSazonov: Thanks for clarify.
So `foreach (var entry in response.Content.Headers) ` is duplicate code and could we remove it?
Sorry for late question - if CoreFX split the headers why we join them again? Only for backward compatibility?

markekraus: We are joining them for backwards compat. In 5.1 and earlier all headers are in `WebResponseObject.Headers`. In 6.0 the content headers are buried in `WebResponseObject.BaseResponse.Content.Headers` unless we promote them to the `WebResponseObject.Headers` dictionary. One would reasonably expect the `Headers` property to contain all headers.

daxian-dbw: @markekraus can you please add a comment to the code to clarify that there won't be duplicate headers in `response.Headers` and `response.Content.Headers`? It would be very helpful to other people who look at the code later.

markekraus: @daxian-dbw should I also include a note about the content headers being added to the headers dictionary for backwards compatibility? or is the note about the distinct headers sufficient?

daxian-dbw: I think it would also be helpful to mention the backward compatibility issue. Thanks @markekraus!

markekraus: Comments added.

daxian-dbw (approved):
LGTM. Thanks!

iSazonov:
@markekraus Thanks for the fix!

PR #4855, opened 2017-09-17
reduce output of pester for CI

SteveL-MSFT:
added `-terse` switch to Start-PSPester to reduce output and suppress progress
update CI scripts to use new switch

example output:
```none
Describing Adapter Tests
+++++++++
Describing Adapter XML Tests
   Context Can set XML node property to non-string object
++++++++++
Describing DataRow and DataRowView Adapter tests
   Context DataRow Adapter tests
++
   Context DataRowView Adapter tests
++
Describing CIM Objects are adapted properly
+++++++++++++++
Describing Online help tests for PowerShell Core Cmdlets
!+++!+!+!+++++!++++++++++++!+++++!++++++++!++++++++++++++!+++++!++++++!++++++++++++++!+++!+++++!+++++++++++++!++!!++++++!+!!!!!!++++++++++!++!++!++!++++++++!!+!+++++++!+!+++!+++!+++++++++++++++++++++!++!++++++!++++++++!+++++!+++!+!++++++!+++++++!!+++!!+!!!!!+!!!+!++!!+!!!!+!++!!!++!+!!!!!!!+++!++!!++++++++!!+++!!!+!!!!!!!!!!!!!!!++++++
```

Runtime (on my laptop) went from 201.22s to 144.89s

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -842,6 +842,7 @@ function Start-PSPester {
         [Parameter(ParameterSetName='Unelevate',Mandatory=$true)]
         [switch]$Unelevate,
         [switch]$Quiet,
+        [switch]$Terse,
         [Parameter(ParameterSetName='PassThru',Mandatory=$true)]
         [switch]$PassThru,
         [switch]$IncludeFailingTest
@@ -883,6 +884,10 @@ function Start-PSPester {
 
     # All concatenated commands/arguments are suffixed with the delimiter (space)
     $Command = ""
+    if ($Terse)
+    {
+        $Command += "`$ProgressPreference = 'silentlyContinue'; "
+    }
 
     # Autoload (in subprocess) temporary modules used in our tests
     $Command += '$env:PSModulePath = '+"'$TestModulePath$TestModulePathSeparator'" + '+$($env:PSModulePath);'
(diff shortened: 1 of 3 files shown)


SteveL-MSFT:
@TravisEz13 it seems like Travis-CI is using a cached version of build.psm1 so it's not picking up my changes?

markekraus:
@SteveL-MSFT Travis-CI is using the `PassThru` switch which calls the `Start-NativeExecution` on line 1021 of your build.psm1. You don't have any logic there to run `Write-Terse` if `-Terse` is present.  Travis CI has to do `PassThru` in order to evaluate for test fails. 

At least that's my theory on why it failed to show up on Travis CI as opposed to a cached `build.psm1`.

SteveL-MSFT:
@markekraus thanks for the tip!

TravisEz13:
Perhaps we should tee the full output to a file and upload it as an artifact (this only really has a meaning on AppVeyor)

SteveL-MSFT:
@TravisEz13 you would still get all the test cases on AppVeyor

SteveL-MSFT:
@TravisEz13 I would suggest having it tee to a log be a separate PR if you really think it's needed

daxian-dbw:
I wonder how to check failed tests with the terse output? Sometimes it's useful to know the tests running before and after the failed tests.

TravisEz13:
@daxian-dbw The describes and contexts are always printed but this is something having the full output somewhere would be useful to resolve, but I agree with Steve that we can make adjustments after we merge this change.

TravisEz13 (approved):

markekraus:
@daxian-dbw @TravisEz13 Aren't the test fails printed in full in `Write-Terse`? It looks like it prints the full test name and the fail output in red and it's only "tersing" pass, skip, and pending tests.

daxian-dbw:
@markekraus Thanks for pointing it out. I didn't look into the changes when asking the question.
Having a full log would be useful and can be done in a separate PR.

Issue #4717, opened 2017-08-31
REQ: Remove-Service cmdlet


chucklu:
Hi,
    Since we have get-service,start-service,and some other functions about windows service.
    Is it possible to add intall-service and uninstall-service?
   
 I am using installutil.exe tool to do the installation and uninstallation.
    sometimes, the uninstallation did not work.
```
installutil /u 
```
   

   then I need to switch to cmd prompt,and use 
```
 sc delete
```

   If you can support install and uninstall service,it will be great.

Environment data
----------------

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.6
OS                             Microsoft Windows 10.0.10586
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

chucklu:
By the way, when the installutil.exe /u did not work. It is something like [this](https://stackoverflow.com/questions/13608157/installutil-wont-uninstall-the-specified-service-does-not-exist-as-an-install).

chuanjiao10:
Install-Service named New-Service:
name    New-Service
Creates a new Windows service.

but ps team not give us remove-service.

chucklu:
@chuanjiao10  thanks,new-service works.

SteveL-MSFT:
There isn't a `Remove-Service`.  If needed, it would be an enhancement.

chucklu:
@SteveL-MSFT  Would you add remove-service to an enhancement list?

SteveL-MSFT:
@chucklu I've updated the title of this to reflect the ask, currently no plans to implement this

chucklu:
@SteveL-MSFT  Thanks,I wish you guys can implement it  in the future.

SteveL-MSFT:
@chucklu workaround for now:

```powershell
get-ciminstance win32_service -filter "name='yourservicename'" | remove-ciminstance
```

chucklu:
@SteveL-MSFT  thanks,I also find the command in powershell [doc](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/new-service?view=powershell-6)
```powershell
PS C:\> sc.exe delete TestService
- or -
[...]

PR #4858, opened 2017-09-18
[Feature]Added Remove-service to Management module

joandrsn:
I have added the function `Remove-Service` to the Management module.
The functions `New-Service` and `Set-Service` were already present, I thought it would be a good idea to have a `Remove-Service`

Issue #4717


src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2167,6 +2167,212 @@ protected override void BeginProcessing()
...
+                        continue;
+                    }
+
+                    NakedWin32Handle hScManager = IntPtr.Zero;
+                    NakedWin32Handle hService = IntPtr.Zero;
+                    try
+                    {
+                        hScManager = NativeMethods.OpenSCManagerW(
+                            ServiceComputerName,
+                            null,
```
adityapatwardhan: Please use named parameters.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -1,4 +1,5 @@
-Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows" {
+Import-Module ".\Microsoft.PowerShell.Commands.Management.dll"
```
adityapatwardhan: This should not be required. Can you remove this?

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -195,6 +196,24 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
...
+            $service = New-Service @parameters
+            $service | Should Not BeNullOrEmpty
+            Remove-Service -Name $servicename
+            $service = Get-Service -Name $servicename -ErrorAction SilentlyContinue
+            $service | Should BeNullOrEmpty
+        }
+        finally {
+            Get-CimInstance Win32_Service -Filter "name='$servicename'" | Remove-CimInstance -ErrorAction SilentlyContinue
+        }
+    }
```
adityapatwardhan: Can you also add tests which validates the error cases?

adityapatwardhan (changes requested):

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2167,6 +2167,212 @@ protected override void BeginProcessing()
     } // class NewServiceCommand
     #endregion NewServiceCommand
 
+    #region RemoveServiceCommand
+    /// <summary>
+    /// This class implements the set-service command
+    /// </summary>
+    [Cmdlet(VerbsCommon.Remove, "Service", SupportsShouldProcess = true, DefaultParameterSetName = "Name")]
+    [OutputType(typeof(ServiceController))]
```
adityapatwardhan: Typically Remove* cmdlets do not return anything.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
adityapatwardhan: Also add tests for pipeline input.

adityapatwardhan (changes requested):

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2167,6 +2167,212 @@ protected override void BeginProcessing()
...
+    public class RemoveServiceCommand : ServiceBaseCommand
+    {
+        #region Parameters
+
+        /// <summary>
+        /// Name of the service to create
```
markekraus: Should be `Name of the service to remove`?

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2167,6 +2167,212 @@ protected override void BeginProcessing()
...
+        /// </summary>
+        /// <value></value>
+        [Parameter(Position = 0, Mandatory = true, ValueFromPipeline = true, ValueFromPipelineByPropertyName = true, ParameterSetName = "Name")]
+        [Alias("ServiceName", "SN")]
+        public string Name { get; set; }
+
+        /// <summary>
+        /// The following is the definition of the input parameter "InputObject".
+        /// Specifies ServiceController object representing the services to be stopped.
```
markekraus: Should be `the services to be removed.`?

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2167,6 +2167,212 @@ protected override void BeginProcessing()
...
+                            WriteNonTerminatingError(
+                                service,
+                                exception,
+                                "CouldNotRemoveService",
+                                ServiceResources.CouldNotSetService,
+                                ErrorCategory.PermissionDenied);
+                            continue;
+                        }
+
+
```
markekraus: Remove extra line.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
joandrsn: I don't quite follow you. Did you want to use the variable `ComputerName` ? Since the method supports both a `Name` on the local computer and a servicecontroller on a remote machine, I believe I have to use the `ServiceComputerName` variable.

markekraus (commented):
Minor formatting and comment/summary rewording suggestions.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -1,4 +1,5 @@
-Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows" {
+Import-Module "C:\Users\jsa\code\PowerShell\src\powershell-win-core\bin\Debug\netcoreapp2.0\win7-x64\Microsoft.PowerShell.Commands.Management.dll"
```
markekraus: Please Remove

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -195,6 +196,47 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
...
+        finally {
+            Get-CimInstance Win32_Service -Filter "name='$servicename'" | Remove-CimInstance -ErrorAction SilentlyContinue
+        }
+    }
+
+    It "Remove-Service cannot accept a service that does not exist" {
+        { Remove-Service -Name "testremoveservice" -ErrorAction 'Stop' } | Should Throw
+    }
+
+
```
markekraus: Remove extra line

markekraus (changes requested):

markekraus:
@joandrsn  it looks like the cmdlet doesn't export. I can repro on win10 x64. No clue why, though.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2167,6 +2167,210 @@ protected override void BeginProcessing()
...
+    /// </summary>
+    [Cmdlet(VerbsCommon.Remove, "Service", SupportsShouldProcess = true, DefaultParameterSetName = "Name")]
+    public class RemoveServiceCommand : ServiceBaseCommand
+    {
+        #region Parameters
+
+        /// <summary>
+        /// Name of the service to remove
+        /// </summary>
+        /// <value></value>
```
SteveL-MSFT: should probably remove this

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2167,6 +2167,210 @@ protected override void BeginProcessing()
...
+
+        #region Overrides
+        /// <summary>
+        /// Remove the service
+        /// </summary>
+        [ArchitectureSensitive]
+        protected override void ProcessRecord()
+        {
+            ServiceController service = null;
+            string ServiceComputerName = null;
```
SteveL-MSFT: for consistency, should use camelCasing

joandrsn:
@markekraus If I import the cmdlet with my import statement that I previously committed (and removed again), I can get it to work. Somehow the travis build gets the module imported (it doesn't fail).
I'm not sure why that is. 
If I try a `Get-Command Get-Service` I get a version 3.1.0.0 and after importing the .dll i get 6.0.0.0

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
SteveL-MSFT: see https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments

It makes the code more readable, particularly where null is passed.

adityapatwardhan:
@joandrsn I guess you need to add them to the exports here:

https://github.com/PowerShell/PowerShell/blob/master/src/Modules/Windows-Core/Microsoft.PowerShell.Management/Microsoft.PowerShell.Management.psd1

https://github.com/PowerShell/PowerShell/blob/master/src/Modules/Unix/Microsoft.PowerShell.Management/Microsoft.PowerShell.Management.psd1

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2167,6 +2167,210 @@ protected override void BeginProcessing()
...
+                        {
+                            bool succeeded = NativeMethods.CloseServiceHandle(hService);
+                            if (!succeeded)
+                            {
+                                int lastError = Marshal.GetLastWin32Error();
+                                Win32Exception exception = new Win32Exception(lastError);
+                                WriteNonTerminatingError(
+                                    service,
+                                    exception,
+                                    "CouldNotRemoveService",
```
SteveL-MSFT: This doesn't seem to the right error.  According to the docs, `CloseServiceHandle` only fails if it's an invalid handle which is probably never going to the be case here.  Anyways, even if this fails, it would seem that the service was still deleted.  Maybe this should just be a diagnostic assert?

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2167,6 +2167,210 @@ protected override void BeginProcessing()
...
+                                    ErrorCategory.PermissionDenied);
+                            }
+                        }
+
+                        if (IntPtr.Zero != hScManager)
+                        {
+                            bool succeeded = NativeMethods.CloseServiceHandle(hScManager);
+                            if (!succeeded)
+                            {
+                                int lastError = Marshal.GetLastWin32Error();
```
SteveL-MSFT: same as above, perhaps just an assert?

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -195,6 +195,46 @@ Describe "Set/New-Service cmdlet tests" -Tags "Feature", "RequireAdminOnWindows"
...
+    It "Remove-Service can accept pipeline input of a ServiceController" {
+        try {
+            $servicename = "testremoveservice"
+            $parameters = @{
+                Name           = $servicename;
+                BinaryPathName = "$PSHOME\powershell.exe"
+            }
+            $service = New-Service @parameters
+            $service | Should Not BeNullOrEmpty
+            Get-Service -Name $servicename | Remove-Service -ErrorAction SilentlyContinue
```
SteveL-MSFT: I don't think you want `-ErrorAction SilentlyContinue` there in case something fails

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2167,6 +2167,197 @@ protected override void BeginProcessing()
...
+                    else
+                    {
+                        serviceComputerName = computer;
+                        service = new ServiceController(Name, serviceComputerName);
+                        objServiceShouldBeDisposed = true;
+                    }
+                    Diagnostics.Assert(!String.IsNullOrEmpty(Name), "null ServiceName");
+                    // "new ServiceController" will succeed even if
+                    // there is no such service.  This checks whether
+                    // the service actually exists.
```
iSazonov: Should we put the comment before Diagnostics.Assert?
And please reformat the comment  - we can use wider lines.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2167,6 +2167,197 @@ protected override void BeginProcessing()
     } // class NewServiceCommand
     #endregion NewServiceCommand
 
+    #region RemoveServiceCommand
+    /// <summary>
+    /// This class implements the remove-service command
```
iSazonov: Case - Remove-Service.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2167,6 +2167,197 @@ protected override void BeginProcessing()
...
+        protected override void ProcessRecord()
+        {
+            ServiceController service = null;
+            string serviceComputerName = null;
+            foreach (string computer in ComputerName)
+            {
+                bool objServiceShouldBeDisposed = false;
+                try
+                {
+                    if (_ParameterSetName.Equals("InputObject", StringComparison.OrdinalIgnoreCase) && InputObject != null)
```
iSazonov: Why are we using `_ParameterSetName` in the file (twice)? In all other files we use `ParameterSetName`.
I believe we should fix this.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2167,6 +2167,197 @@ protected override void BeginProcessing()
...
+                    Diagnostics.Assert(!String.IsNullOrEmpty(Name), "null ServiceName");
+                    // "new ServiceController" will succeed even if
+                    // there is no such service.  This checks whether
+                    // the service actually exists.
+
+                    string unusedByDesign = service.DisplayName;
+                }
+                catch (ArgumentException ex)
+                {
+                    //Cannot use WriteNonterminatingError as service is null
```
iSazonov: Please add space `// Cannot`.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2167,6 +2167,197 @@ protected override void BeginProcessing()
...
+                    ErrorRecord er = new ErrorRecord(ex, "ArgumentException", ErrorCategory.ObjectNotFound, computer);
+                    WriteError(er);
+                    continue;
+                }
+                catch (InvalidOperationException ex)
+                {
+                    //Cannot use WriteNonterminatingError as service is null
```
iSazonov: Please add space `// Cannot`

iSazonov:
@joandrsn Please add `Remove-Service` in DefaultCommands.Tests.ps1 to pass CI.

test/powershell/engine/Basic/DefaultCommands.Tests.ps1
```diff
@@ -407,6 +407,7 @@ Describe "Verify approved aliases list" -Tags "CI" {
 "Cmdlet", "Set-PSDebug",                                        ,                     $($FullCLR -or $CoreWindows -or $CoreUnix)
 "Cmdlet", "Set-PSSessionConfiguration",                         ,                     $($FullCLR -or $CoreWindows -or $CoreUnix)
 "Cmdlet", "Set-Service",                                        ,                     $($FullCLR -or $CoreWindows              )
+"Cmdlet", "Remove-Service",                                     ,                     $($FullCLR -or $CoreWindows              )
```
markekraus: Can you move this up in the list so it is in alphabetical order? Should go between `Remove-PSSnapin` and `Remove-TypeData`

markekraus (changes requested):

markekraus (approved):

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
SteveL-MSFT: Looks like line 1604 used this and @joandrsn was just following this pattern.  Agree that we should change it to just `ParameterSetName` (including line 1604) for consistency.

joandrsn:
I think I figured it out.
I changed the inheritance from `Cmdlet` to `PSCmdlet` to use `ParameterSetName` instead of `_ParameterSetName`.
This meant I had to rewrite a test which used the class directly. Unfortunatly, the test was too complicated for me to rewrite in the same function as before, so I split the test into 4 separate tests.
Also I'm not sure if the last test actually does anything. On my computer, the service was already running. So maybe I should create a service and then change that service? Looking for feedback.

markekraus:
I think this probably needs to be reverted. I believe `ServiceBaseCommand` intentionally derives from `Cmdlet` so that the service cmdlets can be directly invoked. This would likely be a Bucket 1 or 2 Breaking Change.

joandrsn:
Do you have a suggestion on how I would be able to use `ParameterSetName` instead? I can't currently see the solution if changing the inheiritance to `PSCmdlet` would be a breaking change.

markekraus:
I think it was using `_ParameterSetName` "by design" as `ParameterSetName` is not available in commands that derive from `Cmdlet` instead of `PSCmdlet`. `RemoveServiceCommand` could derive from `PSCmdlet` but that would break the derived methods. `ServiceBaseCommand` could be switched to derive from `PSCmdlet`, but that would probably take a review committee approval.

Looking for guidance from other cmdlets, many of them are using 

```csharp
if (InputObject != null)
```
and
```csharp
if (Name != null)
```

Example [`RenameLocalGroupCommand`](https://github.com/PowerShell/PowerShell/blob/fb7c7b038b720fc78d92662ac3405029b8db625c/src/Microsoft.PowerShell.LocalAccounts/LocalAccounts/Commands/RenameLocalGroupCommand.cs#L188)

I lean towards either switching to the null checks or just leaving it as `_ParameterSetName`. I can see benefits to the Service Cmdlets being directly invoked fro host apps, so Moving to `PSCmdlet` seems like a bad option.

@SteveL-MSFT @iSazonov thoughts?

SteveL-MSFT:
Seems like the best solution here (preferring simplicity) is to revert back to deriving from `Cmdlet` as it originally was which perhaps was intentional.  Would like to see the tests back to being `-TestCases` as it seems there a bunch of similar code having split up due to the change in inheritance.

iSazonov:
Oops, I again got an artifact 😄 
I unlike using _ParameterSetName because it is marked as "internal" and available through a hack with  `InternalsVisibleTo`.
So I agree that we should revert last changes and remove `_ParameterSetName`.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
iSazonov: It seems we could remove `_ParameterSetName.Equals("InputObject", StringComparison.OrdinalIgnoreCase)` at all.

SteveL-MSFT: I agree.  InputObject check should be sufficient

SteveL-MSFT (approved):
Thanks for the contribution!

adityapatwardhan (approved):
LGTM

markekraus (approved):
LGTM

adityapatwardhan:
@joandrsn Thanks for your contribution!

PingPongSet:
What vesion of PowerShell is Remove-service available for? What about version 4?

PingPongSet:
@iSazonov What vesion of PowerShell is Remove-service available for?

iSazonov:
@PingPongSet I don' remember the version but you could look release notes for PowerShell Core 6.0. Also make only sense download latest PowerShell Core 6.0 version.

PR #4861, opened 2017-09-18
Only run resource validation tests on Windows, otherwise skip them.

JamesWTruher:
This is a global change rather than in the single WSMan tests. Since we don't have different resources based on platform, we can test on Windows only. If we have resources on non-Windows only then this will need to be a bit more complicated


adityapatwardhan (approved):
LGTM

SteveL-MSFT (approved):

PR #4863, opened 2017-09-19
Get '$PSVersionTable.PSVersion' from ProductVersion attribute

### LINKED ISSUE #3676: If powershell.version is present but empty, $PSVersionTable.GitCommitId reports "Index was outside the bounds of the array"

### LINKED ISSUE #3739: PSVersion in `$PSVersionTable` is hard-coded in PSVersionInfo.cs

iSazonov:
Resolve #3739.
Resolve #3676.

## Fix
Get a formatted product version from ProductVersion attribute.
Assign PSVersionInfo.PSVersion from a formatted product version as '6.0.0-Beta.7'
Assign PSVersionInfo.GitCommitId from a formatted product version as 'v6.0.0-beta.7-37-g0afa5250a...'
Remove dependence on 'powershell.version'.
Improve tests.

### DIFF
--- a/PowerShell.Common.props
+++ b/PowerShell.Common.props
@@ -31,6 +31,9 @@
       <PSCoreAdditionalCommits>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[2].Value)</PSCoreAdditionalCommits>
       <PSCoreCommitSHA>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[3].Value)</PSCoreCommitSHA>
 
+      <!--
+            Caution! 'PSVersion' and 'GitCommitId' from 'PSVersionInfo.cs' depend on the format of this version string.
+      -->
       <PSCoreFormattedVersion Condition = "'$(ReleaseTag)' != '' or '$(PSCoreAdditionalCommits)' == '0'">$(PSCoreBuildVersion) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
       <PSCoreFormattedVersion Condition = "'$(PSCoreFormattedVersion)' == ''">$(PSCoreBuildVersion) Commits: $(PSCoreAdditionalCommits) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
 
--- a/PowerShell.Common.props
+++ b/PowerShell.Common.props
@@ -31,9 +31,6 @@
       <PSCoreAdditionalCommits>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[2].Value)</PSCoreAdditionalCommits>
       <PSCoreCommitSHA>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[3].Value)</PSCoreCommitSHA>
 
-      <!--
-            Caution! PSVersionInfo.cs is very sensetive to the format of this string.
-      -->
       <PSCoreFormattedVersion Condition = "'$(ReleaseTag)' != '' or '$(PSCoreAdditionalCommits)' == '0'">$(PSCoreBuildVersion) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
       <PSCoreFormattedVersion Condition = "'$(PSCoreFormattedVersion)' == ''">$(PSCoreBuildVersion) Commits: $(PSCoreAdditionalCommits) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
 
--- a/PowerShell.Common.props
+++ b/PowerShell.Common.props
@@ -31,6 +31,9 @@
       <PSCoreAdditionalCommits>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[2].Value)</PSCoreAdditionalCommits>
       <PSCoreCommitSHA>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[3].Value)</PSCoreCommitSHA>
 
+      <!--
+            Caution! PSVersionInfo.cs is very sensetive to the format of this string.
+      -->
       <PSCoreFormattedVersion Condition = "'$(ReleaseTag)' != '' or '$(PSCoreAdditionalCommits)' == '0'">$(PSCoreBuildVersion) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
       <PSCoreFormattedVersion Condition = "'$(PSCoreFormattedVersion)' == ''">$(PSCoreBuildVersion) Commits: $(PSCoreAdditionalCommits) SHA: $(PSCoreCommitSHA)</PSCoreFormattedVersion>
 
--- a/build.psm1
+++ b/build.psm1
@@ -393,14 +393,6 @@ function Start-PSBuild {
         }
     }
 
-    # save git commit id to file for PowerShell to include in PSVersionTable
-    $gitCommitId = $ReleaseTag
-    if (-not $gitCommitId) {
-        # if ReleaseTag is not specified, use 'git describe' to get the commit id
-        $gitCommitId = Get-PSCommitId -WarningAction SilentlyContinue
-    }
-    $gitCommitId > "$psscriptroot/powershell.version"
-
     # create the telemetry flag file
     $null = new-item -force -type file "$psscriptroot/DELETE_ME_TO_DISABLE_CONSOLEHOST_TELEMETRY"
 
--- a/src/System.Management.Automation/engine/PSVersionInfo.cs
+++ b/src/System.Management.Automation/engine/PSVersionInfo.cs
@@ -60,12 +60,31 @@ namespace System.Management.Automation
             string assemblyPath = typeof(PSVersionInfo).Assembly.Location;
             string productVersion = FileVersionInfo.GetVersionInfo(assemblyPath).ProductVersion;
 
-            // Get 'ProductVersion' of the assembly. The product version string can be one of the following format examples:
-            //    when powershell is built from a commit      -- '6.0.0-beta.7 Commits: 29 SHA: 52c6b...'
-            //    when powershell is built from a release tag -- '6.0.0-beta.7 SHA: f1ec9...'
-            string rawGitCommitId = "v" + productVersion.Replace(" Commits: ", "-").Replace(" SHA: ", "-g");
+            // Get 'GitCommitId' and 'PSVersion' from the 'productVersion' assembly attribute.
+            //
+            // The strings can be one of the following format examples:
+            //    when powershell is built from a commit:
+            //      productVersion = 6.0.0-beta.7 Commits: 29 SHA: 52c6b...' convert to GitCommitId = '6.0.0-beta.7-29-g52c6b...'
+            //                                                                          PSVersion   = '6.0.0-beta.7'
+            //    when powershell is built from a release tag:
+            //      productVersion = '6.0.0-beta.7 SHA: f1ec9...'            convert to GitCommitId = '6.0.0-beta.7'
+            //                                                                          PSVersion   = '6.0.0-beta.7'
+            //    when powershell is built from a release tag for RTM:
+            //      productVersion = '6.0.0 SHA: f1ec9...'                   convert to GitCommitId = '6.0.0'
+            //                                                                          PSVersion   = '6.0.0'
+            string rawGitCommitId;
+            string mainVersion = productVersion.Substring(0, productVersion.IndexOf(' '));
 
-            s_psV6Version = new SemanticVersion(productVersion.Substring(0, productVersion.IndexOf(' ')));
+            if (productVersion.Contains(" Commits: "))
+            {
+                rawGitCommitId = "v" + productVersion.Replace(" Commits: ", "-").Replace(" SHA: ", "-g");
+            }
+            else
+            {
+                rawGitCommitId = "v" + mainVersion;
+            }
+
+            s_psV6Version = new SemanticVersion(mainVersion);
 
             s_psVersionTable[PSVersionInfo.PSVersionName] = s_psV6Version;
             s_psVersionTable[PSVersionInfo.PSEditionName] = PSEditionValue;
--- a/src/System.Management.Automation/engine/PSVersionInfo.cs
+++ b/src/System.Management.Automation/engine/PSVersionInfo.cs
@@ -45,7 +45,7 @@ namespace System.Management.Automation
         private static Version s_psV4Version = new Version(4, 0);
         private static Version s_psV5Version = new Version(5, 0);
         private static Version s_psV51Version = new Version(5, 1, NTVerpVars.PRODUCTBUILD, NTVerpVars.PRODUCTBUILD_QFE);
-        private static SemanticVersion s_psV6Version = new SemanticVersion(6, 0, 0, "beta");
+        private static SemanticVersion s_psV6Version;
 
         /// <summary>
         /// A constant to track current PowerShell Edition
@@ -56,6 +56,11 @@ namespace System.Management.Automation
         static PSVersionInfo()
         {
             s_psVersionTable = new PSVersionHashTable(StringComparer.OrdinalIgnoreCase);
+
+            string assemblyPath = typeof(PSVersionInfo).GetTypeInfo().Assembly.Location;
+            FileVersionInfo fileVersionInfo = FileVersionInfo.GetVersionInfo(assemblyPath);
+
+            s_psV6Version = new SemanticVersion(fileVersionInfo.ProductVersion);
 
             s_psVersionTable[PSVersionInfo.PSVersionName] = s_psV6Version;
             s_psVersionTable[PSVersionInfo.PSEditionName] = PSEditionValue;
--- a/src/System.Management.Automation/engine/PSVersionInfo.cs
+++ b/src/System.Management.Automation/engine/PSVersionInfo.cs
@@ -20,20 +20,13 @@ namespace System.Management.Automation
         internal const string PSRemotingProtocolVersionName = "PSRemotingProtocolVersion";
         internal const string PSVersionName = "PSVersion";
         internal const string PSEditionName = "PSEdition";
-        internal const string PSFormattedGitCommitIdName = "GitCommitId";
+        internal const string PSGitCommitIdName = "GitCommitId";
         internal const string PSCompatibleVersionsName = "PSCompatibleVersions";
         internal const string PSPlatformName = "Platform";
         internal const string PSOSName = "OS";
         internal const string SerializationVersionName = "SerializationVersion";
         internal const string WSManStackVersionName = "WSManStackVersion";
         private static PSVersionHashTable s_psVersionTable = null;
-
-        /// <summary>
-        /// The constant contains a raw git commit version.
-        /// 'git describe --abbrev=60 --long'
-        /// Ex.: '6.0.0-beta.7-29-52c6bfe1eae24dbaa1a162bffb3754ba3fdc1f4c'
-        /// </summary>
-        private static string s_rawGitCommitId;
 
         /// <summary>
         /// A constant to track current PowerShell Version.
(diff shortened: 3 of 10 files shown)


daxian-dbw:
Thanks @iSazonov! Can we close #3690 now?

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -57,6 +57,11 @@ static PSVersionInfo()
         {
             s_psVersionTable = new PSVersionHashTable(StringComparer.OrdinalIgnoreCase);
 
+            string assemblyPath = typeof(PSVersionInfo).GetTypeInfo().Assembly.Location;
```
daxian-dbw: I think `GetTypeInfo()` is not necessary now.
Another way to get the information version string is:
```CSharp
var infoVersion = typeof(PSVersionInfo).Assembly.GetCustomAttributes(typeof(AssemblyInformationalVersionAttribute), inherit: false)
infoVersion.InformationalVersion
```

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -57,6 +57,11 @@ static PSVersionInfo()
...
+            FileVersionInfo fileVersionInfo = FileVersionInfo.GetVersionInfo(assemblyPath);
+
+            s_psV6Version = new SemanticVersion(fileVersionInfo.ProductVersion);
```
daxian-dbw: The resulting `s_psV6Version` will look like this: `6.0.0-beta.7 SHA: d2fe36fb4137c2a7d6dbc0df22245d14256f3a74`.
I don't think the SHA or Commit info should be put in `$PSVersionTable.PSVersion`. I'm fine we change it from `6.0.0-beta` to `6.0.0-beta.7` though. How about split the information version string to get the `6.0.0-beta.7` part and create the SemanticVersion object with that?

daxian-dbw (changes requested):

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: I think we could try to put the information in another attribute (not ProductVersion) to exclude parsing.

daxian-dbw: Yes, that could be an option if we can find a reasonable attribute to store this info.
Splitting the information version string is fine by me too. And plus, we need to split that string anyway to reconstruct the value for `GitCommitId`.

iSazonov: I don't  found a suitable attribute - we are limited by MSBuild's AL task.

iSazonov:
I rebuilt GitCommitId from formatted one. It looks not very good solution 😕 
Currently $PSVersionTable show formatted GitCommitId and internally we use a rebuilt GitCommitId (raw). 
We can use raw GitCommitId in $PSVersionTable but it seems we want user friendly formatted GitCommitId  there. 
Or maybe refactor and add explicitly FormattedGitCommitId and RawGitCommitId (or GitCommitIdFormatted and GitCommitIdRaw) in $PSVersionTable?

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -57,9 +64,21 @@ static PSVersionInfo()
...
 
+            // Get '6.0.0-beta.7 Commits: 29 SHA: 52c6bfe1eae24dbaa1a162bffb3754ba3fdc1f4c'.
+            // The formatted string is displayed to users in '$PSVersionTable' and PowerShell banner.
+            string assemblyPath = typeof(PSVersionInfo).Assembly.Location;
+            string formattedGitCommitId = FileVersionInfo.GetVersionInfo(assemblyPath).ProductVersion;
+            s_psVersionTable[PSFormattedGitCommitIdName] = formattedGitCommitId;
+
+            // Convert to '6.0.0-beta.7-29-52c6bfe1eae24dbaa1a162bffb3754ba3fdc1f4c'.
+            // The raw git commit version string is widely used internally.
+            s_rawGitCommitId = "v" + formattedGitCommitId.Replace(" Commits: ", "-").Replace(" SHA: ", "-");
```
daxian-dbw: There is a `g` before the SHA value when got from `git describe`:
`v6.0.0-beta.7-30-gf1ec922fd93cf833f32d98fcf337a51f2919b6cf`

And also, when `Commit:` is not present, it indicates the build was from a release tag, so SHA shouldn't be put in the GitCommitId in that case, for example, it would be `v6.0.0-beta.7`.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -57,9 +64,21 @@ static PSVersionInfo()
         {
             s_psVersionTable = new PSVersionHashTable(StringComparer.OrdinalIgnoreCase);
 
+            // Get '6.0.0-beta.7 Commits: 29 SHA: 52c6bfe1eae24dbaa1a162bffb3754ba3fdc1f4c'.
+            // The formatted string is displayed to users in '$PSVersionTable' and PowerShell banner.
```
daxian-dbw: We shouldn't use the formatted string from the attribute directly. CommitId should be in the exact same format as before.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -57,9 +64,21 @@ static PSVersionInfo()
         {
             s_psVersionTable = new PSVersionHashTable(StringComparer.OrdinalIgnoreCase);
 
+            // Get '6.0.0-beta.7 Commits: 29 SHA: 52c6bfe1eae24dbaa1a162bffb3754ba3fdc1f4c'.
```
daxian-dbw: The format could also be `6.0.0-beta.7 SHA: 52c6bfe1eae24dbaa1a162bffb3754ba3fdc1f4c` without the Commit count. It's better to mention that too.
How about:
```
// Get ProductVersion of the assembly. The product version string can be one of the following format examples:
//    when powershell is built from a commit -- '6.0.0-beta.7 Commits: 29 SHA: 52c6b...'
//    when powershell is built from a release tag -- '6.0.0-beta.7 SHA: f1ec9...'
```

PowerShell.Common.props
```diff
@@ -31,6 +31,9 @@
       <PSCoreAdditionalCommits>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[2].Value)</PSCoreAdditionalCommits>
       <PSCoreCommitSHA>$([System.Text.RegularExpressions.Regex]::Match($(PowerShellVersion), $(RegexGitVersion)).Groups[3].Value)</PSCoreCommitSHA>
 
+      <!--
+            Caution! PSVersionInfo.cs is very sensetive to the format of this string.
```
daxian-dbw: It's better to be more specific about what depends on this formatted version string here. How about
```
Caution! PSVersion and GitCommitId from PSVersionInfo.cs depend on the format of this version string.
```

build.psm1
```diff
@@ -325,14 +325,6 @@ function Start-PSBuild {
...
     }
 
-    # save git commit id to file for PowerShell to include in PSVersionTable
-    $gitCommitId = $ReleaseTag
-    if (-not $gitCommitId) {
-        # if ReleaseTag is not specified, use 'git describe' to get the commit id
-        $gitCommitId = Get-PSCommitId -WarningAction SilentlyContinue
-    }
-    $gitCommitId > "$psscriptroot/powershell.version"
-
```
daxian-dbw: Need to remove `powershell.version` from `.gitignore` as well.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Done.

PowerShell.Common.props
iSazonov: Done.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Reverted.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -57,9 +64,21 @@ static PSVersionInfo()
...
+            // The formatted string is displayed to users in '$PSVersionTable' and PowerShell banner.
+            string assemblyPath = typeof(PSVersionInfo).Assembly.Location;
+            string formattedGitCommitId = FileVersionInfo.GetVersionInfo(assemblyPath).ProductVersion;
+            s_psVersionTable[PSFormattedGitCommitIdName] = formattedGitCommitId;
```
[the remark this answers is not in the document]
iSazonov: Reverted.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Fixed.

build.psm1
iSazonov: Fixed.

src/System.Management.Automation/engine/PSVersionInfo.cs
daxian-dbw: `GetTypeInfo()` can be removed.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -57,9 +57,18 @@ static PSVersionInfo()
         {
             s_psVersionTable = new PSVersionHashTable(StringComparer.OrdinalIgnoreCase);
 
+            string assemblyPath = typeof(PSVersionInfo).GetTypeInfo().Assembly.Location;
+            FileVersionInfo fileVersionInfo = FileVersionInfo.GetVersionInfo(assemblyPath);
+
+            // Convert to '6.0.0-beta.7-29-g52c6bfe1eae24dbaa1a162bffb3754ba3fdc1f4c'.
+            // The raw git commit version string is widely used internally.
```
daxian-dbw: Please add comments about the two formats of the product version strings. For example:
```
// Get ProductVersion of the assembly. The product version string can be one of the following format examples:
//    when powershell is built from a commit -- '6.0.0-beta.7 Commits: 29 SHA: 52c6b...'
//    when powershell is built from a release tag -- '6.0.0-beta.7 SHA: f1ec9...'
```
> // Convert to '6.0.0-beta.7-29-g52c6bfe1eae24dbaa1a162bffb3754ba3fdc1f4c'.

This is not accurate. When building from a release tag, it should be converted to `v6.0.0-beta.7`.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -57,9 +57,18 @@ static PSVersionInfo()
...
+            string rawGitCommitId = "v" + fileVersionInfo.ProductVersion.Replace(" Commits: ", "-").Replace(" SHA: ", "-g");
```
daxian-dbw: This won't work when it's built from a release tag. When `Commits:` is not present, SHA shouldn't be included in the reconstructed `GitCommitId`.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -57,9 +57,18 @@ static PSVersionInfo()
...
+
+            s_psV6Version = new SemanticVersion(fileVersionInfo.ProductVersion.Substring(0, fileVersionInfo.ProductVersion.IndexOf(' ')));
```
daxian-dbw: `fileVersionInfo.ProductVersion` is used 3 times here, so I suggest to hold the value in a local variable instead.

src/System.Management.Automation/engine/remoting/commands/CustomShellCommands.cs
```diff
@@ -1564,7 +1564,7 @@ internal static string GetWinrmPluginShellName()
             Hashtable versionTable = PSVersionInfo.GetPSVersionTable();
```
daxian-dbw: It seems `versionTable` is not used anymore. This line can be removed.

src/System.Management.Automation/engine/remoting/commands/CustomShellCommands.cs
```diff
@@ -1577,7 +1577,7 @@ internal static string GetWinrmPluginDllPath()
...
```
daxian-dbw: Ditto.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Sorry - I lost the commit.

Fixed.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Please clarify. It seems works as expected:
```
> "6.0.0-beta.7".Replace(" Commits: ", "-").Replace(" SHA: ", "-g")
6.0.0-beta.7
```

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/remoting/commands/CustomShellCommands.cs
iSazonov: Removed.

I see TODO comment - currently we can detect that we use 'Release Tag' and automate the name selection. Make sense?

src/System.Management.Automation/engine/remoting/commands/CustomShellCommands.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/PSVersionInfo.cs
daxian-dbw: `fileVersionInfo.ProductVersion` will never just be `6.0.0-beta.7`. In case of building with a release tag, it will be like this (for beta.7): 
```
PS> $s.VersionInfo.ProductVersion
6.0.0-beta.7 SHA: d2fe36fb4137c2a7d6dbc0df22245d14256f3a74
```

test/powershell/Host/PSVersionTable.Tests.ps1
```diff
@@ -15,20 +25,29 @@ Describe "PSVersionTable" -Tags "CI" {
...
+
+    It "PSVersion property" {
+       $PSVersionTable.PSVersion | Should BeOfType "System.Management.Automation.SemanticVersion"
+       $PSVersionTable.PSVersion | Should BeExactly $expectedVersion
+       $PSVersionTable.PSVersion.Major | Should Be 6
+    }
+
+    It "GitCommitId property" {
+       $PSVersionTable.GitCommitId | Should BeOfType "System.String"
+       [regex]::IsMatch($PSVersionTable.GitCommitId, "^v(\d+\.\d+\.\d+)-(.+)-(\d+)-g(.+)") | Should Be $true
```
daxian-dbw: Two things:
1. When it's building from a release tag, `$PSVersionTable.GitCommitId` will be like `v6.0.0-beta.7` and this match operation will fail.
2. It's better to use `... | Should Match ...`

daxian-dbw (changes requested):

src/System.Management.Automation/engine/remoting/commands/CustomShellCommands.cs
daxian-dbw: I don't really understand the intention of this comment. Maybe we should open an issue to track it (fix or remove).

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Please clarify. 
1. I see `v6.0.0-beta.7` in Beta.7 distributive  powershell.version file. In Build.psm1 I see the pattern for ReleaseTag "^v\d+\.\d+\.\d+(-\w+\.\d+)?$". If I run PowerShell I see:
```
PowerShell v6.0.0-beta.7
Copyright (C) Microsoft Corporation. All rights reserved.

PS C:\Program Files\PowerShell\6.0.0-beta.7>
```
-- no SHA have we.

2. "6.0.0-beta.7 SHA: d2fe36fb4137c2a7d6dbc0df22245d14256f3a74".Replace(" Commits: ", "-").Replace(" SHA: ", "-g")
Return: v6.0.0-beta.7-gd2fe36fb4137c2a7d6dbc0df22245d14256f3a74
-- the code works as expected.

daxian-dbw: Here is the `$PSVersionTable` for beta.7. Note that `GitCommitId` is ` v6.0.0-beta.7`.
```
PS:1> $PSVersionTable

```
The `ProductVersion` of `System.Management.Automation.dll` is
```
PS:6> $s = Get-Item C:\Arena\pscore\System.Management.Automation.dll
PS:7> $s.VersionInfo.ProductVersion
6.0.0-beta.7 SHA: d2fe36fb4137c2a7d6dbc0df22245d14256f3a74
```
And thus based on your code, `GitCommitId` will become:
```
PS> "v" + $s.VersionInfo.ProductVersion.Replace(" Commits: ", "-").Replace(" SHA: ", "-g")
v6.0.0-beta.7-gd2fe36fb4137c2a7d6dbc0df22245d14256f3a74
``` 
and it's obviously not `v6.0.0-beta.7`.

When building from a release tag, the `GitCommitId` should be exactly the same as of today when using `powershell.version` file.

iSazonov: 😄 I understood that I don't understand.
In official build Beta.7 we have $PSVertionTable.PSVersion= "6.0.0-beta".
In the PR  biuild - "6.0.0-beta.7" - it is hard coded. Should we fix this and remove ".7"?

daxian-dbw: Hope you understand my concern about the `GitCommitId` :)

As for `PSVersion`, it's safer to keep the same as it is, so just change your current code to use `productVersion.LastIndexOf('.')`
```
s_psV6Version = new SemanticVersion(productVersion.Substring(0, productVersion.LastIndexOf('.')));
```
For example, `6.0.0-beta.7 SHA: d2fe36fb4137c2a7d6dbc0df22245d14256f3a74` will get us `6.0.0-beta`

iSazonov: Yes,  GitCommitId and  PSVersion mess with my head.

About "point". What will RTM release tag? "v6.0.0-RTM"? If so I'd prefer to split by space and set PSVersion= "6.0.0-beta.7" and PSVersion= "6.0.0-RTM". If we is sensetive to ".7" or "Beta.7" we catch problems with "RTM".

daxian-dbw: I think by RTM, it will be just `v6.0.0`. I'm fine with having `PSVersion= "6.0.0-beta.7"`.
@SteveL-MSFT @JamesWTruher Do any of you have concerns?

SteveL-MSFT: Yes, the final version should just be `v6.0.0`.  `PSVersion="6.0.0-beta.N"` is correct.  Soon we should have `6.0.0-ReleaseCandidate` and if needed, `6.0.0-ReleaseCandidate.2`

JamesWTruher: I'm not a fan of completely removing the GitCommitId field. I don't mind if it has another name, but it's very useful for our telemetry and understanding how the product is being used and in what way. 

I see the GitCommitId and PSVersion being somewhat related, where GitCommitId is the most specific way to refer to the version of powershell being invoked

JamesWTruher: @daxian-dbw
Changing PSVersion to be a proper representation of any possible semantic version shouldn't break our telemetry. IIRC our telemetry is based on the GitCommitID (@SteveL-MSFT will know for sure). In any event, If it (our telemetry) doesn't support semantic versions, I would argue that we address that, as we _should_ handle a proper semantic version anyway.

daxian-dbw: Thanks @SteveL-MSFT and @JamesWTruher, appreciate your inputs.
@iSazonov Then let's go with `PSVersion= "6.0.0-beta.n`

SteveL-MSFT: Yes, I believe our telemetry is using GitCommitID and not PSVersion

test/powershell/Host/PSVersionTable.Tests.ps1
```diff
@@ -1,4 +1,29 @@
 Describe "PSVersionTable" -Tags "CI" {
+
+    BeforeAll {
+        $powershellProcess=Get-Process -id $pid
+        $rootPath = Split-Path -Path $powershellProcess.path -Parent
+        $sma = Get-Item (Join-Path $rootPath "System.Management.Automation.dll")
```
daxian-dbw: Why not just use `Join-Path $PSHome System.Management.Automation.dll`?

test/powershell/Host/PSVersionTable.Tests.ps1
```diff
@@ -15,20 +40,31 @@ Describe "PSVersionTable" -Tags "CI" {
...
+       $PSVersionTable.PSVersion | Should BeOfType "System.Management.Automation.SemanticVersion"
+       $PSVersionTable.PSVersion | Should BeExactly $expectedPSVersion
+       $PSVersionTable.PSVersion | Should Match $expectedVersionPattern
+       $PSVersionTable.PSVersion.Major | Should Be 6
+    }
+
+    It "GitCommitId property" {
+       $PSVersionTable.GitCommitId | Should BeOfType "System.String"
+       $PSVersionTable.GitCommitId | Should Match $expectedGitCommitIdPattern
+       $PSVersionTable.GitCommitId | Should Not Match $unexpectectGitCommitIdPattern
```
daxian-dbw: This `not match` test case is not necessary IMO since you are doing a `BeExactly` comparison below.

daxian-dbw (changes requested):

daxian-dbw:
@iSazonov Thanks for the update. It's almost there :)

test/powershell/Host/PSVersionTable.Tests.ps1
iSazonov: Join-Path returns string/ We need `FileInfo` to get `VersionInfo`.

daxian-dbw: I know, I meant using that to replace the existing `Join-Path ...` so you don't need the `get-process` and `split-path`.

test/powershell/Host/PSVersionTable.Tests.ps1
iSazonov: We do two checks - 1. for string values (line below). 2. for regex patterns (this line).
I split regex patterns to check GitCommitId and ReleaseTag cases and keep the patterns simple. The double "Should" is needed because `ReleaseTag` pattern can pass `GitCommitId`.

iSazonov:
@daxian-dbw The history is 4 months long - sorry that sometimes I lost the essence of the changes.

test/powershell/Host/PSVersionTable.Tests.ps1
iSazonov: 😄 Clear.

test/powershell/Host/PSVersionTable.Tests.ps1
daxian-dbw: I see the point. It's to guarantee that the reconstructed `rawGitCommitId` to be in the expected regx pattern. #Close

daxian-dbw (approved):

daxian-dbw:
@iSazonov finally we get it ready 😄 Thanks!

iSazonov:
@daxian-dbw Thank you for review and your patience!

Issue #4704, opened 2017-08-30
Wildcard character in registry paths results in "Illegal characters in path."


jared-is-ray:
In Win10 (10.0.16275.1000) and beta6, I am unable to use powershell to modify registry keys that are named "*". 

Steps to reproduce
------------------

```powershell
Remove-Item "HKCU:\Software\Classes\*\shell\"
# or try with escape
Remove-Item "HKCU:\Software\Classes\``*\shell\"
```

Expected behavior
-----------------

```none
Registry key is removed.
```

Actual behavior
---------------

```none
Remove-Item : Illegal characters in path.
Parameter name: path
At line:1 char:1
+ Remove-Item "HKCU:\Software\Classes\``*\shell\"
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Remove-Item], ArgumentException
    + FullyQualifiedErrorId : System.ArgumentException,Microsoft.PowerShell.Commands.RemoveItemCommand
```

Environment data
----------------

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.6
OS                             Microsoft Windows 10.0.16275
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

mklement0:
You did hit a bug, but there are two important asides:  

You should always use `-LiteralPath` to prevent (potentially unwanted) wildcard matching:

```powershell
Remove-Item -LiteralPath "HKCU:\Software\Classes\*\shell\"
```

Positionally, your argument binds to `-Path`, which performs wildcard matching, which would delete the `shell` subkeys of _all_ subkeys of `HKCU:\Software\Classes`.

Also, there is no escaping of wildcard expressions; `"HKCU:\Software\Classes\``*\shell\"` tries to match `` HKCU:\Software\Classes\`*\shell\ `` as a pattern, which is not the intent.

The only way to suppress wildcard matching is to use `-LiteralPath`.

mklement0:
A test case to demonstrate the problem succinctly:

```powershell
# Note that paths passed to New-Item are *always* interpreted as literals paths.
[...]

Issue #4330, opened 2017-07-24
Remove-Item: potential v6 regression on path like Function:\Global:*


nightroman:
Steps to reproduce
------------------

```powershell
    # works
    $null = New-Item Function:\Global:Test-Me -Value '"Test-Me1"'
    Test-Me

    # works
    Set-Item Function:\Global:Test-Me -Value '"Test-Me2"'
    Test-Me

    # works in v2-v5, fails in v6
    Remove-Item Function:\Global:Test-Me
```

Expected behavior
-----------------

`Remove-Item Function:\Global:Test-Me` removes the specified global function in v6,
like it does in v2-v5.

Actual behavior
---------------

`Remove-Item Function:\Global:Test-Me` fails in v6:

```none
    Remove-Item : The given path's format is not supported.
```

Environment data
----------------

```
    Name                           Value
    ----                           -----
    PSVersion                      6.0.0-beta
    PSEdition                      Core
    GitCommitId                    v6.0.0-beta.4
    OS                             Microsoft Windows 10.0.10586
    Platform                       Win32NT
    PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
    PSRemotingProtocolVersion      2.3
    SerializationVersion           1.1.0.1
    WSManStackVersion              3.0
```

joeyaiello:
This is a real bug, but removing the `Global:` works in some cases, and it seems like a bit of an edge case.

@nightroman do you have a scenario that this is blocking? Seems a bit academic.

nightroman:
My mocking tool creates temporary global functions.

SteveL-MSFT:
My PR for https://github.com/PowerShell/PowerShell/pull/4866 will also address this.  However, I noticed on both Windows PowerShell 5.1 and with my private build, you would need to do:

```powershell
remove-item function:\test-me
```

To actually remove the function

PR #4866, opened 2017-09-19
allow * to be used in registry path for remove-item

SteveL-MSFT:
Code is creating a DirectoryInfo instance to determine if the path is a directory or file, but it fails because `*` is not allowed, however, `*` is a valid character in a registry path.  Fix is to catch this exception and ignore it.  This means that using `-literalpath` with an asterisk and the filesystem provider will now return an error rather than returning quietly.

Fix https://github.com/PowerShell/PowerShell/issues/4704
Fix https://github.com/PowerShell/PowerShell/issues/4330

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
@@ -3266,6 +3266,10 @@ namespace Microsoft.PowerShell.Commands
                     {
                         // not a directory
                     }
+                    catch (System.ArgumentException)
+                    {
+                        // invalid characters (like asterisk) for a file path, but valid for other paths (like registry)
+                    }
 
                     if (!treatAsFile && !Recurse && hasChildren)
                     {
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
@@ -3107,6 +3107,18 @@ namespace Microsoft.PowerShell.Commands
                     try
                     {
                         resolvedPSPaths = SessionState.Path.GetResolvedPSPathFromPSPath(path, currentContext);
+                        if (null != LiteralPath && 0 == resolvedPSPaths.Count)
+                        {
+                            ItemNotFoundException pathNotFound =
+                                new ItemNotFoundException(
+                                    path,
+                                    "PathNotFound",
+                                    SessionStateStrings.PathNotFound);
+                            WriteError(new ErrorRecord(
+                                pathNotFound.ErrorRecord,
+                                pathNotFound));
+                            continue;
+                        }
                     }
                     finally
                     {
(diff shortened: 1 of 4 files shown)


src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
```diff
@@ -3266,6 +3266,10 @@ protected override void ProcessRecord()
                     {
                         // not a directory
                     }
+                    catch (System.ArgumentException)
```
anmenaga: Maybe it is better to skip that <DirectoryInfo code section> altogether if provider is not FileSystem provider?
e.g.: resolvedPath.Provider.Name.Equals(FileSystemProvider.ProviderName, StringComparison.OrdinalIgnoreCase)

anmenaga: ... that <DirectoryInfo code section> section altogether ...

anmenaga: ... that "DirectoryInfo code section" section altogether ...

src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
```diff
@@ -3266,6 +3266,10 @@ protected override void ProcessRecord()
...
+                    {
```
anmenaga: Need to make sure that FileSystem paths continue to work as they were before this new catch handler; 2 tests:
1) remove-item with -Path and filesystem path that contains *
2) remove-item with -LiteralPath and filesystem path that contains *

anmenaga (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Management/Registry.Tests.ps1
```diff
@@ -469,6 +469,22 @@ Describe "Extended Registry Provider Tests" -Tags @("Feature", "RequireAdminOnWi
...
+            try {
+                $tempPath = "HKCU:\_tmp"
+                $testPath = "$tempPath\*\sub"
+                New-Item -Force $testPath
+                $testPath | Should Exist
+                Remove-Item -LiteralPath $testPath
+                $testPath | Should Not Exist
+            }
+            finally {
+                Remove-Item -Recurse $tempPath
```
adityapatwardhan: Please add `-ErrorAction SilentlyContinue -Force`

test/powershell/Modules/Microsoft.PowerShell.Management/Registry.Tests.ps1
```diff
@@ -469,6 +469,22 @@ Describe "Extended Registry Provider Tests" -Tags @("Feature", "RequireAdminOnWi
             $result."$testPropertyName" | Should Be $testPropertyValue
         }
     }
+
+    Context "Validate -LiteralPath" {
+        It "Verify New-Item and Remove-Item work with asterisk" {
+            try {
+                $tempPath = "HKCU:\_tmp"
+                $testPath = "$tempPath\*\sub"
+                New-Item -Force $testPath
```
adityapatwardhan: Please assign to null.
```Powershell
$null = New-Item -Force $testPath
```

src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
adityapatwardhan: I agree checking for the provider seems to be a better option.

adityapatwardhan (changes requested):

src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
SteveL-MSFT: will make the change

src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
SteveL-MSFT: Looks like there's tests for -Path, I'll add a test for -LiteralPath with *

test/powershell/Modules/Microsoft.PowerShell.Management/Registry.Tests.ps1
SteveL-MSFT: will fix

test/powershell/Modules/Microsoft.PowerShell.Management/Registry.Tests.ps1
SteveL-MSFT: will add

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and is ok with the minor `breaking change`

adityapatwardhan (approved):
LGTM

adityapatwardhan:
@anmenaga Can you have a look again?

anmenaga (approved):

adityapatwardhan:
@SteveL-MSFT Can you update the PR description.

PR #4869, opened 2017-09-19
Replace httpbin.org/encoding Tests with WebListener

markekraus:
* Add Encoding/Utf8 test to WebListener
* Replace tests using httpbin.org/encoding/utf8 with Weblistener URL
* removed superfluous `Invoke-RestMethod` test. I'm not sure what the test was originally meant to accomplish but there is not an equivalent `Invoke-WebRequest` test and there are already other tests for `Invoke-RestMethod` which test `text/html` responses.  httpbin.org/encoding/utf8 did not return `text/plain` anyway. *shrugs

reference #2504


test/tools/WebListener/Controllers/EncodingController.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            string url = "/Encoding/Utf8";
+            ViewData["Url"] = url;
+            Response.Redirect(url, false);
+            return View("~/Views/Redirect/Index.cshtml");
+        }
+
+        public ActionResult Utf8()
+        {
+            MediaTypeHeaderValue mediaType = new MediaTypeHeaderValue("tex/html");
```
adityapatwardhan: Should it be "tex**t**/html"?

test/tools/WebListener/Views/Encoding/Utf8.cshtml
```diff
@@ -0,0 +1,220 @@
...
+  ╔══╦══╗  ┌──┬──┐  ╭──┬──╮  ╭──┬──╮  ┏━━┳━━┓  ┎┒┏┑   ╷  ╻ ┏┯┓ ┌┰┐    ▊ ╱╲╱╲╳╳╳
+  ║┌─╨─┐║  │╔═╧═╗│  │╒═╪═╕│  │╓─╁─╖│  ┃┌─╂─┐┃  ┗╃╄┙  ╶┼╴╺╋╸┠┼┨ ┝╋┥    ▋ ╲╱╲╱╳╳╳
+  ║│╲ ╱│║  │║   ║│  ││ │ ││  │║ ┃ ║│  ┃│ ╿ │┃  ┍╅╆┓   ╵  ╹ ┗┷┛ └┸┘    ▌ ╱╲╱╲╳╳╳
+  ╠╡ ╳ ╞╣  ├╢   ╟┤  ├┼─┼─┼┤  ├╫─╂─╫┤  ┣┿╾┼╼┿┫  ┕┛┖┚     ┌┄┄┐ ╎ ┏┅┅┓ ┋ ▍ ╲╱╲╱╳╳╳
+  ║│╱ ╲│║  │║   ║│  ││ │ ││  │║ ┃ ║│  ┃│ ╽ │┃  ░░▒▒▓▓██ ┊  ┆ ╎ ╏  ┇ ┋ ▎
+  ║└─╥─┘║  │╚═╤═╝│  │╘═╪═╛│  │╙─╀─╜│  ┃└─╂─┘┃  ░░▒▒▓▓██ ┊  ┆ ╎ ╏  ┇ ┋ ▏
+  ╚══╩══╝  └──┴──┘  ╰──┴──╯  ╰──┴──╯  ┗━━┻━━┛  ▗▄▖▛▀▜   └╌╌┘ ╎ ┗╍╍┛ ┋  ▁▂▃▄▅▆▇█
+                                               ▝▀▘▙▄▟
+
+</pre>
```
adityapatwardhan: Please add newline at the end.

adityapatwardhan (changes requested):

test/tools/WebListener/Controllers/EncodingController.cs
markekraus: Doh... fixed. thanks

test/tools/WebListener/Views/Encoding/Utf8.cshtml
markekraus: Fixed.

TravisEz13:
Tests failed due to a "spelling" error:  https://travis-ci.org/PowerShell/PowerShell/jobs/277524298#L4389

test/tools/WebListener/Controllers/EncodingController.cs
```diff
@@ -0,0 +1,36 @@
...
+            return View("~/Views/Redirect/Index.cshtml");
+        }
+
+        public ActionResult Utf8()
+        {
+            MediaTypeHeaderValue mediaType = new MediaTypeHeaderValue("text/html");
+            mediaType.Encoding = Encoding.UTF8;
+            return View();
+        }
+        public IActionResult Error()
```
iSazonov: Please add new line.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1376,19 +1372,14 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
...
-        $result = ExecuteWebCommand -command $command
+        $uri = Get-WebListenerUrl -Test 'Encoding/Utf8'
+        $command = "Invoke-RestMethod -Uri '$uri'"
 
-        # Validate response content
-        # TODO: There is a bug on ConvertFrom-Json that fails for utf8.
-        $result.headers.Host | Should Match "httpbin.org"
-        $result.headers.'User-Agent' | Should Match "WindowsPowerShell"
+        $result = ExecuteWebCommand -command $command        
+        $Result.Output | Should Match '⡌⠁⠧⠑ ⠼⠁⠒  ⡍⠜⠇⠑⠹⠰⠎ ⡣⠕⠌'
```
iSazonov: Is this test sufficient? Do we really check Utf8 not Utf16/32 here? 
Can we also protect from changing the encoding of this file? Does it make sense use `u{...} ?
The source file contains text in many languages. Does it make sense to ckeck multiple languages.

markekraus: I wasn't clear on what the original tests were trying to accomplish and I'm not proficient with encoding to know what is a good test or not. It was clear to me that they were definitely not sufficient for UTF8, so I tried to add something that would at least eliminate ASCII. After the response headers are available for `Invoke-RestMethod`, it would be possible to test the response encoding similar to the `Invoke-WebRequest`. But I'm definitely open to suggestion on how to improve these tests.

markekraus: On the source file. If we can narrow down a specific UTF8 character or string, We can switch to generating the string from the controller and remove the view. That way we do not need to worry about the encoding change. I'd need guidance on what to put there, though.

test/tools/WebListener/Controllers/EncodingController.cs
markekraus: fixed.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
TravisEz13: I think this is an improvement over what we had before.  If you think the tests can be further improved, perhaps an issue would be a better way to address the feedback.

TravisEz13:
@adityapatwardhan Can you update your review?

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: I think this is good. If my research on this is correct, It's no easy task to test a string for its encoding This at least include some UTf8 chars.

Opened #4890 for further improvements

adityapatwardhan (approved):
LGTM

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: Thanks for open tracking Issue!

Closed.

iSazonov (approved):
@markekraus Thanks for your contribution!

Issue #1920, opened 2016-08-19
Unix system executables terminate with error if transcription enabled


latkin:
## Steps to reproduce

Enable transcription (`Start-Transcript`)
Run any executable system command.
## Expected behavior

Command executes and terminates cleanly.
## Actual behavior

Command executes fine, but upon termination an error is always printed: `Program 'foo' failed to run: The method or operation is not implemented.At line:1 char:1`

Examples:

```
PS /> uname                                                                                                                                                                                                                                                              
Darwin
Program 'uname' failed to run: The method or operation is not implemented.At line:1 char:1
+ uname
+ ~~~~~.
At line:1 char:1
+ uname
+ ~~~~~
    + CategoryInfo          : ResourceUnavailable: (:) [], ApplicationFailedException
    + FullyQualifiedErrorId : NativeCommandFailed

PS /> curl google.com                                                                                                                                                                                                                                                    
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>301 Moved</TITLE></HEAD><BODY>
<H1>301 Moved</H1>
The document has moved
<A HREF="http://www.google.com/">here</A>.
</BODY></HTML>
Program 'curl' failed to run: The method or operation is not implemented.At line:1 char:1
+ curl google.com
+ ~~~~~~~~~~~~~~~.
At line:1 char:1
+ curl google.com
+ ~~~~~~~~~~~~~~~
    + CategoryInfo          : ResourceUnavailable: (:) [], ApplicationFailedException
    + FullyQualifiedErrorId : NativeCommandFailed
```

Does not happen if I capture the program's output into a variable:

```
PS /> $x = uname                                                                                                                                                                                                                                                         
PS /> $x                                                                                                                                                                                                                                                                 
Darwin
```

Stack:

```
PS /> $error[0].Exception.ToString()                                                                                                                                                                                                                                     
System.Management.Automation.ApplicationFailedException: Program 'curl' failed to run: The method or operation is not implemented.At line:1 char:1
+ curl google.com
+ ~~~~~~~~~~~~~~~. ---> System.NotImplementedException: The method or operation is not implemented.
   at Microsoft.PowerShell.ConsoleHostRawUserInterface.GetBufferContents(Rectangle rectangle)
   at System.Management.Automation.Internal.Host.InternalHostRawUserInterface.GetBufferContents(Rectangle r)
   at System.Management.Automation.NativeCommandProcessor.Complete()
[...]

PR #4871, opened 2017-09-20
Enable transcription of native commands on non-Windows

SteveL-MSFT:
Transcription was relying on reading the screen buffer to record output from native commands.
This resulted in an unhandled exception calling an unimplemented api on non-Windows.
Fix is to default to using redirected output if reading the screen buffer is not supported.

Fix https://github.com/PowerShell/PowerShell/issues/1920

### DIFF
--- a/src/System.Management.Automation/engine/NativeCommandProcessor.cs
+++ b/src/System.Management.Automation/engine/NativeCommandProcessor.cs
@@ -185,6 +185,17 @@ namespace System.Management.Automation
 
             //Create input writer for providing input to the process.
             _inputWriter = new ProcessInputWriter(Command);
+
+            try
+            {
+                Host.BufferCell[,] bufferContents = this.Command.Context.EngineHostInterface.UI.RawUI.GetBufferContents(
+                    new Host.Rectangle(_startPosition, _startPosition));
+                _scrapeHostOutput = true;
+            }
+            catch (NotImplementedException)
+            {
+                _scrapeHostOutput = false;
+            }
         }
 
         /// <summary>
@@ -398,12 +409,16 @@ namespace System.Management.Automation
             // redirecting anything. This is a bit tricky as we always run redirected so
             // we have to see if the redirection is actually being done at the topmost level or not.
 
-            //Calculate if input and output are redirected.
-            bool redirectOutput;
-            bool redirectError;
-            bool redirectInput;
+            // Calculate if input and output are redirected.
+            bool redirectOutput = true;
+            bool redirectError = true;
+            bool redirectInput = true;
 
-            CalculateIORedirection(out redirectOutput, out redirectError, out redirectInput);
+            // If we are transcribing and can't scrape, then continue to redirect so we have the output
+            if (!this.Command.Context.EngineHostInterface.UI.IsTranscribing || _scrapeHostOutput)
+            {
+                CalculateIORedirection(out redirectOutput, out redirectError, out redirectInput);
+            }
 
             // Find out if it's the only command in the pipeline.
             bool soloCommand = this.Command.MyInvocation.PipelineLength == 1;
@@ -417,7 +432,6 @@ namespace System.Management.Automation
             }
 
             _startPosition = new Host.Coordinates();
-            _scrapeHostOutput = false;
 
             Exception exceptionToRethrow = null;
             try
@@ -432,19 +446,10 @@ namespace System.Management.Automation
 
                     // Also, store the Raw UI coordinates so that we can scrape the screen after
                     // if we are transcribing.
-                    try
+                    if (this.Command.Context.EngineHostInterface.UI.IsTranscribing && _scrapeHostOutput)
                     {
-                        if (this.Command.Context.EngineHostInterface.UI.IsTranscribing)
-                        {
-                            _scrapeHostOutput = true;
-                            _startPosition = this.Command.Context.EngineHostInterface.UI.RawUI.CursorPosition;
-                            _startPosition.X = 0;
-                        }
-                    }
-                    catch (Host.HostException)
-                    {
-                        // The host doesn't support scraping via its RawUI interface
-                        _scrapeHostOutput = false;
+                        _startPosition = this.Command.Context.EngineHostInterface.UI.RawUI.CursorPosition;
+                        _startPosition.X = 0;
                     }
                 }
 
--- a/src/System.Management.Automation/engine/NativeCommandProcessor.cs
+++ b/src/System.Management.Automation/engine/NativeCommandProcessor.cs
@@ -401,9 +401,9 @@ namespace System.Management.Automation
             // we have to see if the redirection is actually being done at the topmost level or not.
 
             // Calculate if input and output are redirected.
-            bool redirectOutput = true;
-            bool redirectError = true;
-            bool redirectInput = this.Command.MyInvocation.ExpectingInput;
+            bool redirectOutput;
+            bool redirectError;
+            bool redirectInput;
 
             _startPosition = new Host.Coordinates();
 
@@ -1290,7 +1290,10 @@ namespace System.Management.Automation
                     redirectError = true;
                 }
             }
-            else if (_isTranscribing && (false == s_supportScreenScrape))
+
+            // if screen scraping isn't supported, we enable redirection so that the output is still transcribed
+            // as redirected output is always transcribed
+            if (_isTranscribing && (false == s_supportScreenScrape))
             {
                 redirectOutput = true;
                 redirectError = true;
--- a/src/System.Management.Automation/engine/NativeCommandProcessor.cs
+++ b/src/System.Management.Automation/engine/NativeCommandProcessor.cs
@@ -405,6 +405,8 @@ namespace System.Management.Automation
             bool redirectError = true;
             bool redirectInput = this.Command.MyInvocation.ExpectingInput;
 
+            _startPosition = new Host.Coordinates();
+
             if (null == s_supportScreenScrape)
             {
                 try
@@ -419,11 +421,7 @@ namespace System.Management.Automation
                 }
             }
 
-            // If we are transcribing and can't scrape, then continue to redirect so we have the output
-            if (!_isTranscribing || (true == s_supportScreenScrape))
-            {
-                CalculateIORedirection(out redirectOutput, out redirectError, out redirectInput);
-            }
+            CalculateIORedirection(out redirectOutput, out redirectError, out redirectInput);
 
             // Find out if it's the only command in the pipeline.
             bool soloCommand = this.Command.MyInvocation.PipelineLength == 1;
@@ -435,8 +433,6 @@ namespace System.Management.Automation
             {
                 throw new PipelineStoppedException();
             }
-
-            _startPosition = new Host.Coordinates();
 
             Exception exceptionToRethrow = null;
             try
@@ -707,8 +703,8 @@ namespace System.Management.Automation
                     ConsumeAvailableNativeProcessOutput(blocking: true);
                     _nativeProcess.WaitForExit();
 
-                    // Capture screen output if we are transcribing
-                    if (_isTranscribing && (true == s_supportScreenScrape))
+                    // Capture screen output if we are transcribing and running stand alone
+                    if (_isTranscribing && (true == s_supportScreenScrape) && _runStandAlone)
                     {
                         Host.Coordinates endPosition = this.Command.Context.EngineHostInterface.UI.RawUI.CursorPosition;
                         endPosition.X = this.Command.Context.EngineHostInterface.UI.RawUI.BufferSize.Width - 1;
@@ -1293,6 +1289,11 @@ namespace System.Management.Automation
                     redirectOutput = true;
                     redirectError = true;
                 }
+            }
+            else if (_isTranscribing && (false == s_supportScreenScrape))
+            {
+                redirectOutput = true;
+                redirectError = true;
             }
 
             _runStandAlone = !redirectInput && !redirectOutput && !redirectError;
--- a/src/System.Management.Automation/engine/NativeCommandProcessor.cs
+++ b/src/System.Management.Automation/engine/NativeCommandProcessor.cs
@@ -186,16 +186,7 @@ namespace System.Management.Automation
             //Create input writer for providing input to the process.
             _inputWriter = new ProcessInputWriter(Command);
 
-            try
-            {
-                Host.BufferCell[,] bufferContents = this.Command.Context.EngineHostInterface.UI.RawUI.GetBufferContents(
-                    new Host.Rectangle(_startPosition, _startPosition));
-                _scrapeHostOutput = true;
-            }
-            catch (NotImplementedException)
-            {
-                _scrapeHostOutput = false;
-            }
+            _isTranscribing = this.Command.Context.EngineHostInterface.UI.IsTranscribing;
         }
 
         /// <summary>
@@ -384,8 +375,8 @@ namespace System.Management.Automation
         /// </summary>
         private BlockingCollection<ProcessOutputObject> _nativeProcessOutputQueue;
 
-        private bool _scrapeHostOutput;
-
+        private static bool? s_supportScreenScrape = null;
+        private bool _isTranscribing;
         private Host.Coordinates _startPosition;
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
```diff
@@ -136,4 +135,19 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
...
+            $ps = [powershell]::Create()
+            $ps.addscript("Start-Transcript -path $transcriptFilePath").Invoke()
+            $ps.addscript("hostname").Invoke()
+            $ps.addscript("Stop-Transcript").Invoke()
+        } finally {
+            if ($null -ne $ps) {
+                $ps.Dispose()
+            }
+        }
+        Test-Path $transcriptFilePath | Should be $true
```
anmenaga: Also need to check that PS error stream is empty.

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
```diff
@@ -136,4 +135,19 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
...
+        $machineName = [System.Environment]::MachineName
+        $transcriptFilePath | Should contain $machineName
+    }
 }
```
anmenaga: Need newline in the end of file.

anmenaga (changes requested):

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
SteveL-MSFT: will add

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
SteveL-MSFT: will add

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
SteveL-MSFT: Actually, there's no reason to follow that pattern, I'm rewriting it as a scriptblock and any errors should just show up in the test run.

src/System.Management.Automation/engine/NativeCommandProcessor.cs
```diff
@@ -185,6 +185,17 @@ internal NativeCommandProcessor(ApplicationInfo applicationInfo, ExecutionContex
 
             //Create input writer for providing input to the process.
             _inputWriter = new ProcessInputWriter(Command);
+
+            try
+            {
+                Host.BufferCell[,] bufferContents = this.Command.Context.EngineHostInterface.UI.RawUI.GetBufferContents(
+                    new Host.Rectangle(_startPosition, _startPosition));
+                _scrapeHostOutput = true;
```
daxian-dbw: This can be made a static check so that we do it only once. And `_scrapeHostOutput` can be renamed to `s_supportScreenScrape` (or something similar).

src/System.Management.Automation/engine/NativeCommandProcessor.cs
```diff
@@ -398,12 +409,16 @@ private void InitNativeProcess()
...
             // we have to see if the redirection is actually being done at the topmost level or not.
 
-            //Calculate if input and output are redirected.
-            bool redirectOutput;
-            bool redirectError;
-            bool redirectInput;
+            // Calculate if input and output are redirected.
+            bool redirectOutput = true;
+            bool redirectError = true;
+            bool redirectInput = true;
```
daxian-dbw: Why do we want `redirectInput` to be true by default? Even if we are transcribing and screen scrape is not supported, we shouldn't need to always redirect the input.

src/System.Management.Automation/engine/NativeCommandProcessor.cs
```diff
@@ -398,12 +409,16 @@ private void InitNativeProcess()
...
 
-            CalculateIORedirection(out redirectOutput, out redirectError, out redirectInput);
+            // If we are transcribing and can't scrape, then continue to redirect so we have the output
+            if (!this.Command.Context.EngineHostInterface.UI.IsTranscribing || _scrapeHostOutput)
```
daxian-dbw: Suggestion: `this.Command.Context.EngineHostInterface.UI.IsTranscribing` is used at 3 places in this class. Maybe we can have an instance field to cache the value? Like `_isTranscribing = this.Command.Context.EngineHostInterface.UI.IsTranscribing` in the constructor?

src/System.Management.Automation/engine/NativeCommandProcessor.cs
```diff
@@ -432,19 +446,10 @@ private void InitNativeProcess()
...
-                    try
-                    {
-                        if (this.Command.Context.EngineHostInterface.UI.IsTranscribing)
-                        {
-                            _scrapeHostOutput = true;
-                            _startPosition = this.Command.Context.EngineHostInterface.UI.RawUI.CursorPosition;
-                            _startPosition.X = 0;
-                        }
-                    }
-                    catch (Host.HostException)
```
daxian-dbw: Looks like an HostException also indicates `The host doesn't support scraping via its RawUI interface` ...

src/System.Management.Automation/engine/NativeCommandProcessor.cs
```diff
@@ -185,6 +185,17 @@ internal NativeCommandProcessor(ApplicationInfo applicationInfo, ExecutionContex
...
+            try
+            {
+                Host.BufferCell[,] bufferContents = this.Command.Context.EngineHostInterface.UI.RawUI.GetBufferContents(
+                    new Host.Rectangle(_startPosition, _startPosition));
+                _scrapeHostOutput = true;
+            }
+            catch (NotImplementedException)
+            {
+                _scrapeHostOutput = false;
+            }
```
daxian-dbw: from line 444 below, it looks like a `HostException` also indicates `The host doesn't support scraping via its RawUI interface` ...

src/System.Management.Automation/engine/NativeCommandProcessor.cs
daxian-dbw: And form line 442 below, it indicates that we transcribe for native command only if `_runStandAlone` is true. Now it looks like you are assuming we can do transcribe regardless of `_runStandAlone`. That seems not right.

src/System.Management.Automation/engine/NativeCommandProcessor.cs
```diff
@@ -185,6 +185,17 @@ internal NativeCommandProcessor(ApplicationInfo applicationInfo, ExecutionContex
...
             //Create input writer for providing input to the process.
             _inputWriter = new ProcessInputWriter(Command);
+
+            try
+            {
+                Host.BufferCell[,] bufferContents = this.Command.Context.EngineHostInterface.UI.RawUI.GetBufferContents(
+                    new Host.Rectangle(_startPosition, _startPosition));
+                _scrapeHostOutput = true;
+            }
+            catch (NotImplementedException)
```
daxian-dbw: By putting the call to `GetBufferContents` in the constructor, it's possible that an arbitrary exception will be thrown from `GetBufferContents` in the constructor. When that happens, it would be wrapped to an `ApplicationFailedException` properly prior to this change, but after this change, the `NativeCommandProcessor` would just fail to be constructed, and thus I guess the error would look more obscure. So, I'm not sure if it's appropriate to put this in the constructor ...

daxian-dbw (changes requested):

daxian-dbw:
> Fix is to default to using redirected output if reading the screen buffer is not supported.

@SteveL-MSFT Could you please explain a little why redirecting the output/error would support transcribing in case that screen scraping is not supported? (just curious how it's done in that case)

SteveL-MSFT:
@daxian-dbw from what I can tell in the code, anything that gets redirected gets transcribed anyways

src/System.Management.Automation/engine/NativeCommandProcessor.cs
SteveL-MSFT: will change

src/System.Management.Automation/engine/NativeCommandProcessor.cs
SteveL-MSFT: will fix

src/System.Management.Automation/engine/NativeCommandProcessor.cs
```diff
@@ -698,8 +708,7 @@ internal override void Complete()
                     _nativeProcess.WaitForExit();
 
                     // Capture screen output if we are transcribing
-                    if (this.Command.Context.EngineHostInterface.UI.IsTranscribing &&
-                        _scrapeHostOutput)
+                    if (_isTranscribing && (true == s_supportScreenScrape))
```
daxian-dbw: I think there is a change of behavior here. Assuming `_isTranscribing == $true` and screen scraping is supported,
- before the change, `_scrapeHostOutput` is true only if `_runStandAlone` is true, so this block will run only the native command is running standalone.
- after the change, even if it's not running standalone (i.e. standard input/error redirected), this block of code still will run when transcribing, and in such case, `_startPosition` is not even initialized.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/NativeCommandProcessor.cs
```diff
@@ -398,12 +400,30 @@ private void InitNativeProcess()
...
+                    s_supportScreenScrape = true;
+                }
+                catch (NotImplementedException)
+                {
+                    s_supportScreenScrape = false;
+                }
+            }
+
+            // If we are transcribing and can't scrape, then continue to redirect so we have the output
+            if (!_isTranscribing || (true == s_supportScreenScrape))
```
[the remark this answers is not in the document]
SteveL-MSFT: Makes sense, will fix.

src/System.Management.Automation/engine/NativeCommandProcessor.cs
SteveL-MSFT: Will fix

src/System.Management.Automation/engine/NativeCommandProcessor.cs
```diff
@@ -1285,6 +1290,11 @@ private void CalculateIORedirection(out bool redirectOutput, out bool redirectEr
                     redirectError = true;
                 }
             }
+            else if (_isTranscribing && (false == s_supportScreenScrape))
```
daxian-dbw: Maybe remove the `else`? The `else if (Platform.IsWindowsDesktop && IsConsoleApplication(this.Path))` block doesn't necessarily set `redirectOutput` and `redirectError` to true.

daxian-dbw: And please add a comment that explains why we need to set output/error redirection in such case.

src/System.Management.Automation/engine/NativeCommandProcessor.cs
```diff
@@ -398,10 +400,26 @@ private void InitNativeProcess()
...
             // we have to see if the redirection is actually being done at the topmost level or not.
 
-            //Calculate if input and output are redirected.
-            bool redirectOutput;
-            bool redirectError;
-            bool redirectInput;
+            // Calculate if input and output are redirected.
+            bool redirectOutput = true;
+            bool redirectError = true;
+            bool redirectInput = this.Command.MyInvocation.ExpectingInput;
```
daxian-dbw: These changes now can be reverted.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/NativeCommandProcessor.cs
SteveL-MSFT: Will revert

src/System.Management.Automation/engine/NativeCommandProcessor.cs
SteveL-MSFT: yup

daxian-dbw (approved):

daxian-dbw:
@anmenaga Can you please take a look again?

SteveL-MSFT:
@markekraus the test failures are indicating the WebListener didn't start within the timeout.  Is this something you're familiar with?

markekraus:
@SteveL-MSFT News to me. I will take a look.

anmenaga (approved):

markekraus:
@SteveL-MSFT This appears to be related to the change in this PR. It looks native execution inside a job is broken. WebListener runs `dotnet WebListener.dll` in a `Start-Job` block. From your PR I can manually run the WebListener outside of a job in PowerShell. But inside a job it fails. But it's not limited to WebListener:

```poweshell
$Job = Start-Job {cmd.exe /c echo "test" }
$Job | Wait-Job | fl * -force
"-{0}-" -f $Job.ChildJobs[0].Output
```

expected behavior:
```none
State         : Completed
HasMoreData   : True
StatusMessage :
Location      : localhost
Command       : cmd.exe /c echo "test"
JobStateInfo  : Completed
Finished      : System.Threading.ManualResetEvent
InstanceId    : 7d654dd9-ba42-48f0-be72-9e91863e7286
Id            : 1
Name          : Job1
ChildJobs     : {Job2}
PSBeginTime   : 9/26/2017 3:41:02 PM
PSEndTime     : 9/26/2017 3:41:02 PM
PSJobTypeName : BackgroundJob
Output        : {}
Error         : {}
Progress      : {}
Verbose       : {}
Debug         : {}
Warning       : {}
Information   : {}

-test-
```

Actual behavior:
```none
State         : Failed
HasMoreData   : False
StatusMessage :
Location      : localhost
Command       : cmd.exe /c echo "test"
JobStateInfo  : Failed
Finished      : System.Threading.ManualResetEvent
InstanceId    : 58c30632-c4a1-41e3-b276-5aa9b24db3b0
Id            : 3
Name          : Job3
ChildJobs     : {Job4}
PSBeginTime   : 9/26/2017 3:42:41 PM
PSEndTime     : 9/26/2017 3:42:42 PM
PSJobTypeName : BackgroundJob
Output        : {}
Error         : {}
Progress      : {}
Verbose       : {}
Debug         : {}
Warning       : {}
Information   : {}

--
```

SteveL-MSFT:
@markekraus thanks!  I'll take a look.

src/System.Management.Automation/engine/NativeCommandProcessor.cs
```diff
@@ -398,11 +400,25 @@ private void InitNativeProcess()
...
+            {
+                _startPosition = this.Command.Context.EngineHostInterface.UI.RawUI.CursorPosition;
+                Host.BufferCell[,] bufferContents = this.Command.Context.EngineHostInterface.UI.RawUI.GetBufferContents(
+                    new Host.Rectangle(_startPosition, _startPosition));
+                s_supportScreenScrape = true;
+            }
+            catch (Exception)
+            {
+                // screen scraping not supported on non-Windows or as job
+            }
```
daxian-dbw: I think this check can still be done only once -- we do this check in `CalculateIORedirection` and only if `_runStandAlone` is true. Here the the code I propose:
```CSharp
private void CalculateIORedirection(out bool redirectOutput, out bool redirectError, out bool redirectInput)
{
    ...

    if (NativeCommandProcessor.IsServerSide)
    {
        redirectInput = true;
        redirectOutput = true;
        redirectError = true;
    }
    else if (Platform.IsWindowsDesktop && IsConsoleApplication(this.Path))
    {
        // On Windows desktops, if the command to run is a console application,
        // then allocate a console if there isn't one attached already...
        ConsoleVisibility.AllocateHiddenConsole();

        if (ConsoleVisibility.AlwaysCaptureApplicationIO)
        {
            redirectOutput = true;
            redirectError = true;
        }
    }
    _runStandAlone = !redirectInput && !redirectOutput && !redirectError;

    // !!!!! NOTE !!!!! -- unchanged until here

    if (_runStandAlone)
    {
        if (s_supportScreenScrape == null)
        {
            // Call GetBufferContents() to check and set s_supportScreenScrape to true or false accordingly.
        }

        // if screen scraping isn't supported, we enable redirection so that the output is still transcribed
        // as redirected output is always transcribed
        if (_isTranscribing && (false == s_supportScreenScrape))
        {
            redirectOutput = true;
            redirectError = true;
            _runStandAlone = false;
        }
    }
    
```
So the check of screen scraping support is done only if we are running the native command as standalone application.

In case of running it in a Job, I think the standard input and error are already redirected.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/NativeCommandProcessor.cs
SteveL-MSFT: I see.  I'll try that.

daxian-dbw (approved):
LGTM. Thanks!

Issue #4172, opened 2017-07-01
Never truncate when writing output to file


pedrogarcia:
When writing console output to file with *>>, lines should not be truncated. Currently, output is truncated to 85 characters. This makes Powershell fundamentally unusable for me.

pedrogarcia:
Actually, stuff should never be truncated in general. Pretty much every command I run in powershell has things truncated with ...
THere is plenty of space in my screen and in the window, yet, after 15 characters powershell will truncated some stuff. 
Man, how I really hate microsoft tools.

iSazonov:
@pedrogarcia Thanks for your report! The Issue is duplicate #1908. 
Currently PowerShell Core format all output based on console width, we want enhance the behavior for pipes with native commands. 
>Man, how I really hate microsoft tools.

PowerShell Core is an open project - feel free to show how need to build systems correctly. Welcome to make the enhancement and fix the Issue!

lzybkr:
@iSazonov - I'm reopening because fixing this issue might require different work than #1908, and this issue can be fixed without fixing #1908.

daxian-dbw:
@pedrogarcia could you please add a repro?

pedrogarcia:
Sure. 
Run a long copying task for rclone, with verbose output flag, for example:
‘
rclone -vv copy “drive:/folder1” “drive:/folder2” *>> cmd_output_log.log
‘

SteveL-MSFT:
I see the code in `Out-File` which computes the console width, but I'm not able to repro this.  In fact, the `-width` parameter doesn't even get honored.

```powershell
$text = "0123456789"
while ($text.Length -lt [console]::WindowWidth) { $text += $text }                
$text | Out-File ./test.txt -Encoding ascii
```

Not even able to repro with Windows PS 5.1

iSazonov:
@SteveL-MSFT I believe Out-File don't send simple objects to formatting system before output. If you replace string variable in your test with a hash or array you'll get repo.

SteveL-MSFT:
@iSazonov but the original repro was a native command outputting to stdout

iSazonov:
I wonder. Maybe Out-File detect a file provider in input strings?

SteveL-MSFT:
I tried using the native `echo` on Linux as well as `testexe -echoargs` to get long output, but nothing gets truncated.

Objects being truncated during formatting is currently expected (although I'm open to consider changing it...) and `-width` does work in that case.

[...]

PR #4876, opened 2017-09-20
Sort .spelling

### LINKED ISSUE #4875: .spelling Should Be Sorted
.spelling should be sorted. It's kind of awkward searching existing terms and adding new terms. I think it should be sorted "alphabetically".

markekraus:
closes #4875 

Sorting, separating and labeling the file overrides and words, and fixing incorrect `CODE_OF_CONDUCT.md` override.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -3,48 +3,73 @@
 # global dictionary is at the start, file overrides afterwards
 # one word per line, to define a file override use ' - filename'
 # where filename is relative to this configuration file
+#region Global Dictionary
+2ae5d07
+acl
 ActiveDirectory
+aditya
+adityapatwardhan
 alpha.15
 alpha.7
 alpha.8
 alpha.9
+AlternateStream
+amd64
 analyzing
 AppImage
 AppVeyor
+argumentlist
+arm32
 artifact
 artifacts
 ASP.NET
 AssemblyLoadContext
+authenticodesignature
 behavior
 behaviors
+beta.7
 booleans
 catalog
 cataloged
 CentOS
+childitem
+chucklu
+cimsession
+cimsupport
 cmake
 cmd
 cmdlet
 cmdlets
+codecov.io
 color
 colors
+comobject
 config
 ConsoleHost
 ConvertFrom-Json
+convertto-xml
 CoreCLR
 CoreFX
+credssp
 crontab
 crossgen
+cs
+CsPhysicallyInstalledMemory
+debughandler
 DevOps
 DockerFile
 DockerFiles
+DottedScopes
 eBook
 enum
 env
+ergo3114
 exe
 favor
 favorite
 frontload
 FullCLR
+functionprovider
 Get-Acl
 Get-AuthenticodeSignature
 Get-ChildItem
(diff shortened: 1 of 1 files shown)


daxian-dbw:
@markekraus Thanks!
@TravisEz13 will be very pleased to see this PR 😄

.spelling
```diff
@@ -3,690 +3,387 @@
 # global dictionary is at the start, file overrides afterwards
 # one word per line, to define a file override use ' - filename'
 # where filename is relative to this configuration file
+# File Overrides:
```
TravisEz13: You sorted it too broadly.  There is a general section then file specific sections.  I suspect the PR will fail because now all the words say they are for `CODE_OF_CONDUCT.md`

TravisEz13 (changes requested):
need to sort per section.

.spelling
SteveL-MSFT: Ideally, words used in multiple files would be at the top in the `global` section and each file override has it's own specific section which is sorted.

SteveL-MSFT (changes requested):

.spelling
markekraus: ah.. I misunderstood what was going on there. I will fix it.

markekraus:
@TravisEz13  https://gist.github.com/markekraus/7c019b19377e61a1c40dee3ba1e90801

TravisEz13:
It might be good to add a test or a tool that will keep it sorted.

.spelling
```diff
@@ -3,48 +3,73 @@
 # global dictionary is at the start, file overrides afterwards
 # one word per line, to define a file override use ' - filename'
 # where filename is relative to this configuration file
+#region Global Dictionary
```
SteveL-MSFT: Does VSCode support collapsing these regions?  I think from a readability standpoint, using whitespace between sections may be easier to read rather than decorating them with `#region` unless there's an editor benefit.

markekraus: User defined region folding is in the [iteration plan](https://github.com/Microsoft/vscode/issues/34160) for the 29th, that's why I included it. 😀 I could do white space too, though.

SteveL-MSFT: My personal preference is to go with whitespace for now.  We can always add regions later.  This is not a blocking issue for me, however.

SteveL-MSFT (approved):

markekraus:
Hmm https://travis-ci.org/PowerShell/PowerShell/jobs/277989665
This output is all over the place. I can't tell if it is a temporary glitch or something is actually broken.

SteveL-MSFT:
@markekraus that's intentional via https://github.com/PowerShell/PowerShell/pull/4855.  Failed tests will still show up verbatim, but passed/skipped/pending tests are just a symbol.  Some output (longer than console width) still sneaks through due to https://github.com/PowerShell/PowerShell/issues/4172

markekraus:
@SteveL-MSFT [This particular one is odd](https://travis-ci.org/PowerShell/PowerShell/jobs/277989665). It looks like the test ran 3 times or were at least outputed 3 times. The macOS and AppVeyor build look normal (well, post-terse normal anyway) but the Trusty build is odd.

SteveL-MSFT:
@markekraus you're right, didn't notice that.  In that run, the output implies the tests ran twice.  Looked at another CI run to compare and it didn't have that problem.  Need to keep an eye out to see if that was a one-off or something else.

TravisEz13 (approved):

PR #4888, opened 2017-09-21
Add ResponseHeadersVariable Parameter to Invoke-RestMethod

### LINKED ISSUE #4845: Allow Capture of Response Headers with Invoke-RestMethod

markekraus:
closes #4845

* Adds `-ResponseHeadersVariable` parameter to Invoke-RestMethod
* Adds `-RHV` alias for the parameter
* A variable with the provided name is created in the calling scope which contains the HTTP Response Headers as a dictionary constructed in the same manner `WebResponseObject.Headers` is constructed for `Invoke-WebRequest`
* Add tests for the parameter

# Documentation Needed
The new parameter will need to be documented.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
@@ -60,6 +60,13 @@ namespace Microsoft.PowerShell.Commands
             set { base._maximumFollowRelLink = value; }
         }
 
+        /// <summary>
+        /// Gets or sets the ResponseHeadersVariable property.
+        /// </summary>
+        [Parameter]
+        [Alias("HV")]
+        public string ResponseHeadersVariable { get; set; }
+
         #endregion Parameters
 
         #region Helper Methods
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
@@ -64,7 +64,7 @@ namespace Microsoft.PowerShell.Commands
         /// Gets or sets the ResponseHeadersVariable property.
         /// </summary>
         [Parameter]
-        [Alias("HV")]
+        [Alias("RHV")]
         public string ResponseHeadersVariable { get; set; }
 
         #endregion Parameters
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/InvokeRestMethodCommand.CoreClr.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/InvokeRestMethodCommand.CoreClr.cs
@@ -95,6 +95,12 @@ namespace Microsoft.PowerShell.Commands
                 {
                     StreamHelper.SaveStreamToFile(responseStream, QualifiedOutFile, this);
                 }
+
+                if (!String.IsNullOrEmpty(ResponseHeadersVariable))
+                {
+                    PSVariableIntrinsics vi = SessionState.PSVariable;
+                    vi.Set(ResponseHeadersVariable, WebResponseHelper.GetHeadersDictionary(response));
+                }
             }
         }
 
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -2080,14 +2080,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
             $headers.'Content-Type' | Should Be 'text/html; charset=utf-8'
             $headers.Server | Should Be 'Kestrel'
         }
-        It "Verifies Invoke-RestMethod supports -HV alias" {
-            $uri = Get-WebListenerUrl -Test '/'
-            $response = Invoke-RestMethod -Uri $uri -HV 'headers'
-
-            $headers | Should Not BeNullOrEmpty
-            $headers.'Content-Type' | Should Be 'text/html; charset=utf-8'
-            $headers.Server | Should Be 'Kestrel'
-        }
+        
         It "Verifies Invoke-RestMethod supports -ResponseHeadersVariable overwriting existing variable" {
             $uri = Get-WebListenerUrl -Test '/'
             $headers = 'prexisting'
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -2076,7 +2076,6 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
             $uri = Get-WebListenerUrl -Test '/'
             $response = Invoke-RestMethod -Uri $uri -ResponseHeadersVariable 'headers'
 
-            $headers | Should Not BeNullOrEmpty
             $headers.'Content-Type' | Should Be 'text/html; charset=utf-8'
             $headers.Server | Should Be 'Kestrel'
         }
@@ -2087,7 +2086,6 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
             $response = Invoke-RestMethod -Uri $uri -ResponseHeadersVariable 'headers'
 
             $headers | Should Not Be 'prexisting'
-            $headers | Should Not BeNullOrEmpty
             $headers.'Content-Type' | Should Be 'text/html; charset=utf-8'
             $headers.Server | Should Be 'Kestrel'
         }
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -2071,6 +2071,35 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
 
     #endregion charset encoding tests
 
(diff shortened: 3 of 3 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
```diff
@@ -60,6 +60,13 @@ public int MaximumFollowRelLink
             set { base._maximumFollowRelLink = value; }
         }
 
+        /// <summary>
+        /// Gets or sets the ResponseHeadersVariable property.
+        /// </summary>
+        [Parameter]
+        [Alias("HV")]
```
iSazonov: Maybe 'RHV'?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
```diff
@@ -60,6 +60,13 @@ public int MaximumFollowRelLink
...
+        public string ResponseHeadersVariable { get; set; }
```
iSazonov: Should we add any validation attributes?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/InvokeRestMethodCommand.CoreClr.cs
```diff
@@ -95,6 +95,12 @@ internal override void ProcessResponse(HttpResponseMessage response)
                 {
                     StreamHelper.SaveStreamToFile(responseStream, QualifiedOutFile, this);
                 }
+
+                if (!String.IsNullOrEmpty(ResponseHeadersVariable))
+                {
+                    PSVariableIntrinsics vi = SessionState.PSVariable;
+                    vi.Set(ResponseHeadersVariable, WebResponseHelper.GetHeadersDictionary(response));
```
iSazonov: Should we return ReadOnlyDictionary?
Does it make sense to return an _entire_ response?

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -2071,6 +2071,35 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
...
+    Context 'Invoke-RestMethod ResponseHeadersVariable Tests' {
+        It "Verifies Invoke-RestMethod supports -ResponseHeadersVariable" {
+            $uri = Get-WebListenerUrl -Test '/'
+            $response = Invoke-RestMethod -Uri $uri -ResponseHeadersVariable 'headers'
+
+            $headers | Should Not BeNullOrEmpty
+            $headers.'Content-Type' | Should Be 'text/html; charset=utf-8'
+            $headers.Server | Should Be 'Kestrel'
+        }
+        It "Verifies Invoke-RestMethod supports -HV alias" {
```
iSazonov: We never test parameter aliases. If we want test parameter aliases we must create one complex test. I think we should remove the test.

iSazonov (commented):
Leave a comment

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
markekraus: `SessionVariable` does something similar and it doesn't have any validation. I'm not sure what value it would add. Aside from being null or empty (which is tested to set it on line 99) a variable name can be any valid string.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/InvokeRestMethodCommand.CoreClr.cs
markekraus: I have seen cases where users are taking the returned dictionary, updating it, and then supplying back to further calls. I don't see much benefit. Also, `System.Net.Http.Headers.HttpResponseHeaders` is not read-only.

```powershell
$Res = invoke-webrequest https://google.com
$Res.BaseResponse.Headers.TryAddWithoutValidation('testy','lala')
$Res.BaseResponse.Headers.GetValues('testy')
```
results with `lala`

As for the returning entire response, that would be somewhat pointless with `Invoke-RestMethod`, IMO. If the user needs the full response object, they can use `Invoke-WebRequest`. This is just adding the ability to get the response headers which many API's use to provide information along with the JSON/XML response. But, a separate PR could be made to return the response object if that is something people really want.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
markekraus: I'm up for suggestions on this one. I was following suit with `SessionVariable`'s `SV`. I would have used `-HeadersVariable` for the name if that wouldn't have been mistaken for `-Headers`

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: Fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
adityapatwardhan: I prefer RHV too.

adityapatwardhan (approved):
LGTM

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
markekraus: updated

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -2071,6 +2071,28 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
...
+            $headers.Server | Should Be 'Kestrel'
+        }
+        
+        It "Verifies Invoke-RestMethod supports -ResponseHeadersVariable overwriting existing variable" {
+            $uri = Get-WebListenerUrl -Test '/'
+            $headers = 'prexisting'
+            $response = Invoke-RestMethod -Uri $uri -ResponseHeadersVariable 'headers'
+
+            $headers | Should Not Be 'prexisting'
+            $headers | Should Not BeNullOrEmpty
```
SteveL-MSFT: Seems like this test isn't needed as the next line would fail

SteveL-MSFT (commented):
Thanks for the contribution!  My comments are not blocking.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
```diff
@@ -60,6 +60,13 @@ public int MaximumFollowRelLink
             set { base._maximumFollowRelLink = value; }
         }
 
+        /// <summary>
+        /// Gets or sets the ResponseHeadersVariable property.
+        /// </summary>
+        [Parameter]
+        [Alias("RHV")]
+        public string ResponseHeadersVariable { get; set; }
```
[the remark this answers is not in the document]
markekraus: I chose `Headers` because we are using the `-Headers` parameter as well as `WebResponseObject.Headers` (which this provides parity for). I can change it, but it would be the odd-man-out.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: Fixed. and removed from the other test as well.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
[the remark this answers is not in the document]
SteveL-MSFT: I'm fine with it

SteveL-MSFT (approved):

iSazonov (approved):

adityapatwardhan:
@markekraus Thanks for your contribution.

pr3sidentspence:
@markekraus Respo**sn**e typo still exists in the body of OP. 

> Adds `-ResposneHeadersVariable` parameter ...

PR #4892, opened 2017-09-21
fix tests not setting back $psdefaultparametervalues back correctly

SteveL-MSFT:
Due to how Pester works, when $psdefaultparametervalues is first set, it is set at global scope so needs to be reset back in global scope


markekraus:
`test\powershell\Language\Parser\RedirectionOperator.Tests.ps1` is also not cloning:

```powershell
        $oldDefaultParameterValues = $psDefaultParameterValues
        $psDefaultParameterValues = @{}
```
Not sure if you want to fix this in this PR or address it in another.

SteveL-MSFT:
@markekraus I think it makes sense to fix it with this PR, I'll take care of it.  Thanks!

adityapatwardhan (approved):
LGTM

TravisEz13 (approved):

PR #4896, opened 2017-09-22
[ Don't merge!!! ] Add Test-Json cmdlet

### LINKED ISSUE #4220: Test-Json cmdlet (JSON lint)
Placeholder for now as I follow up on some customer scenarios: it would be really nice to have a cmdlet that could validate JSON before it's used as input. I'd expect for something like this to already exist in Newtonsoft.Json. 

IMHO, it's a nice-to-have for 6.0.0 depending on how we're looking with `HighPriority` over the next couple months. If not, we could move it out to 6.1.0 without a real problem.

iSazonov:
Resolve #4220.

The cmdlet is based on [Newtonsoft.Json.Schema](https://github.com/JamesNK/Newtonsoft.Json.Schema).
It allow to check:
- Json by only parsing
- Json against Schema
- implicitly check Schema by parsing (based on previous line check)

Please see [license](https://github.com/JamesNK/Newtonsoft.Json.Schema/blob/master/LICENSE.md) of the framework (GNU Affero General Public License). /cc @joeyaiello @SteveL-MSFT

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
@@ -1,0 +1,94 @@
+using System;
+using System.Collections.Generic;
+using System.Management.Automation;
+using System.Management.Automation.Internal;
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Schema;
+
+namespace Microsoft.PowerShell.Commands
+{
+    /// <summary>
+    /// A cmdlet to validate an Json.
+    /// </summary>
+    [Cmdlet(VerbsDiagnostic.Test, "Json", HelpUri = "")]
+    public class TestJsonCommand : PSCmdlet
+    {
+        /// <summary>
+        /// An Json to be validated.
+        /// </summary>
+        [Parameter(Position = 0, Mandatory = true, ValueFromPipeline = true)]
+        public String Json { get; set; }
+
+        /// <summary>
+        /// A schema to validate the Json against.
+        /// </summary>
+        [Parameter(Position = 1)]
+        [ValidateNotNullOrEmpty()]
+        public String Schema { get; set; }
+
+        private JSchema _jschema = null;
+
+        /// <summary>
+        /// Prepare an Json schema.
+        /// </summary>
+        protected override void BeginProcessing()
+        {
+            if (Schema != null)
+            {
+                try
+                {
+                    _jschema = JSchema.Parse(Schema);
+                }
+                catch (Exception exc)
+                {
+                    Exception exception = new Exception(TestJsonCmdletStrings.InvalidJsonSchema, exc);
+                    ThrowTerminatingError(new ErrorRecord(exception, "InvalidJsonSchema", ErrorCategory.InvalidData, null));
+                }
+            }
+        }
+
+        /// <summary>
+        /// Validate an Json.
+        /// </summary>
+        protected override void ProcessRecord()
+        {
+            JObject parsedJson = null;
+            bool result = true;
+
+            try
+            {
+                parsedJson = JObject.Parse(Json);
+
+                if (_jschema != null)
+                {
+                    if (!parsedJson.IsValid(_jschema, out IList<string> errorMessages))
+                    {
+                        result = false;
+
+                        Exception exception = new Exception(TestJsonCmdletStrings.InvalidJsonAgainistSchema);
+                        if (errorMessages != null)
+                        {
+                            foreach (var message in errorMessages)
+                            {
+                                ErrorRecord errorRecord = new ErrorRecord(exception, "InvalidJsonAgainistSchema", ErrorCategory.InvalidData, null);
+                                errorRecord.ErrorDetails = new ErrorDetails(message);
+                                WriteError(errorRecord);
+                            }
+                        }
+                    }
+                }
+            }
+            catch (Exception exc)
+            {
+                result = false;
+
+                Exception exception = new Exception(TestJsonCmdletStrings.InvalidJson, exc);
+                WriteError(new ErrorRecord(exception, "InvalidJson", ErrorCategory.InvalidData, Json));
+            }
+
+            WriteObject(result);
+        }
+    }
+}
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
@@ -33,7 +33,7 @@ namespace Microsoft.PowerShell.Commands
         [ValidateNotNullOrEmpty()]
         public String Schema { get; set; }
 
-        private JSchema _jschema = null;
+        private JSchema _jschema;
 
         /// <summary>
         /// Prepare an Json schema.
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
@@ -23,6 +23,11 @@ namespace Microsoft.PowerShell.Commands
 
         /// <summary>
         /// A schema to validate the Json against.
+        /// It is optional parameter.
+        /// If the parameter is absent the cmdlet only try to parse the Json.
+        /// If the parameter present the cmdlet try to parse the Json and
+        /// then check the Json against the schema. Before the check
+        /// the cmdlet parse the schema doing implicitly check the schema too.
         /// </summary>
         [Parameter(Position = 1)]
         [ValidateNotNullOrEmpty()]
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
@@ -10,7 +10,7 @@ using Newtonsoft.Json.Schema;
 namespace Microsoft.PowerShell.Commands
 {
     /// <summary>
-    /// A cmdlet to validate an Json.
+    /// This class implements Test-Json command.
     /// </summary>
     [Cmdlet(VerbsDiagnostic.Test, "Json", HelpUri = "")]
     public class TestJsonCommand : PSCmdlet
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/Test-Json.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/Test-Json.Tests.ps1
@@ -1,0 +1,104 @@
+Describe "Test-Json" -Tags "CI" {
+    BeforeAll {
+        $validSchemaJson = @"
+            {
+            'description': 'A person',
+            'type': 'object',
+            'properties': {
+                'name': {'type': 'string'},
+                'hobbies': {
+                'type': 'array',
+                'items': {'type': 'string'}
+                }
+                }
+            }
+"@
+
+        $invalidSchemaJson = @"
+            {
+            'description',
+            'type': 'object',
+            'properties': {
+                'name': {'type': 'string'},
+                'hobbies': {
+                'type': 'array',
+                'items': {'type': 'string'}
+                }
+                }
+            }
+"@
+
+        $validJson = @"
+            {
+                'name': 'James',
+                'hobbies': ['.NET', 'Blogging', 'Reading', 'Xbox', 'LOLCATS']
+            }
+"@
+
+        $invalidTypeInJson = @"
+            {
+                'name': 123,
+                'hobbies': ['.NET', 'Blogging', 'Reading', 'Xbox', 'LOLCATS']
+            }
+"@
+
+        $invalidTypeInJson2 = @"
+            {
+                'name': 123,
+                'hobbies': [456, 'Blogging', 'Reading', 'Xbox', 'LOLCATS']
+            }
+"@
+
+        $invalidNodeInJson = @"
+            {
+                'name': 'James',
+                'hobbies': ['.NET', 'Blogging', 'Reading', 'Xbox', 'LOLCATS']
+                errorNode
+            }
+"@
+}
+
+    It "Json is valid" {
+        Test-Json -Json $validJson | Should Be $true
+    }
+
+    It "Json is valid aganist a valid schema" {
+        Test-Json -Json $validJson -Schema $validSchemaJson | Should Be $true
+    }
+
+    It "Json is invalid" {
+        Test-Json -Json $invalidNodeInJson -ErrorAction SilentlyContinue | Should Be $false
+    }
+
+    It "Json is invalid aganist a valid schema" {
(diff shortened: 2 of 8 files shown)


SteveL-MSFT:
I’ll check if we can accept this license.  For now, you’ll need to add it to https://github.com/PowerShell/PowerShell/blob/master/ThirdPartyNotices.txt as part of this PR

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Management.Automation.Internal;
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Schema;
+
+namespace Microsoft.PowerShell.Commands
+{
+    /// <summary>
+    /// A cmdlet to validate an Json.
```
markekraus: Can this include a note that it implements the `Test-Json` command?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+                result = false;
+
+                Exception exception = new Exception(TestJsonCmdletStrings.InvalidJson, exc);
+                WriteError(new ErrorRecord(exception, "InvalidJson", ErrorCategory.InvalidData, Json));
+            }
+
+            WriteObject(result);
+        }
+    }
+}
```
markekraus: Add newline

iSazonov:
@SteveL-MSFT I can not pull changes in ThirdPartyNotices.txt - it is excluded file.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
iSazonov: Done.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
iSazonov: Fixed.

SteveL-MSFT:
@iSazonov ok, I'll take care of that file then once I get approval.  We should not merge this PR until the license is approved for this project.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Schema;
+
+namespace Microsoft.PowerShell.Commands
+{
+    /// <summary>
+    /// This class implements Test-Json command.
+    /// </summary>
+    [Cmdlet(VerbsDiagnostic.Test, "Json", HelpUri = "")]
```
PaulHigin: I don't know what the process is for creating Cmdlet help/helpuri.  Should we have an issue tracking this so that the work item is not forgotten?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+    public class TestJsonCommand : PSCmdlet
+    {
+        /// <summary>
+        /// An Json to be validated.
+        /// </summary>
+        [Parameter(Position = 0, Mandatory = true, ValueFromPipeline = true)]
```
PaulHigin: I think [ValidateNotNullOrEmpty()] is needed here too.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+        public String Schema { get; set; }
+
+        private JSchema _jschema = null;
+
+        /// <summary>
+        /// Prepare an Json schema.
+        /// </summary>
+        protected override void BeginProcessing()
+        {
+            if (Schema != null)
```
PaulHigin: I don't think this check is necessary since Schema property is validated as not null.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+        public String Json { get; set; }
+
+        /// <summary>
+        /// A schema to validate the Json against.
+        /// </summary>
+        [Parameter(Position = 1)]
+        [ValidateNotNullOrEmpty()]
+        public String Schema { get; set; }
+
+        private JSchema _jschema = null;
```
PaulHigin: [Nitpick] We normally don't explicitly initialize fields to null since C# guarantees fields are initialized to zero.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+        protected override void ProcessRecord()
+        {
+            JObject parsedJson = null;
+            bool result = true;
+
+            try
+            {
+                parsedJson = JObject.Parse(Json);
+
+                if (_jschema != null)
```
PaulHigin: I don't think this null check is necessary since we throw terminating error for invalid schema.

PaulHigin (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
iSazonov: @Francisco-Gamino @daxian-dbw Could you help with new HelpUri?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
iSazonov: @daxian-dbw It seems we should add this in our code guidance.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
iSazonov: My thoughts is:
 - Here we can get values from pipeline
 - The cmdlet is specific - it _tests_ input objects 
-- so for consistency it is better to be based on Json library check logic.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
iSazonov: It is a check that the property present.

test/powershell/Modules/Microsoft.PowerShell.Utility/Test-Json.Tests.ps1
```diff
@@ -0,0 +1,104 @@
...
+        $invalidNodeInJson = @"
+            {
+                'name': 'James',
+                'hobbies': ['.NET', 'Blogging', 'Reading', 'Xbox', 'LOLCATS']
+                errorNode
+            }
+"@
+}
+
+    It "Json is valid" {
```
adityapatwardhan: First 4 It can be converted to one with testcases and splatting.

test/powershell/Modules/Microsoft.PowerShell.Utility/Test-Json.Tests.ps1
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    It "Test-Json write an error on invalid (type) Json aganist a valid schema" {
+        $errorVar = $null
+        Test-Json -Json $invalidTypeInJson -Schema $validSchemaJson -ErrorVariable errorVar -ErrorAction SilentlyContinue
+
+        $errorVar.FullyQualifiedErrorId | Should BeExactly "InvalidJsonAgainistSchema,Microsoft.PowerShell.Commands.TestJsonCommand"
+    }
+
+    It "Test-Json write an error on invalid (node) Json aganist a valid schema" {
```
adityapatwardhan: This one and the one above can be combined into single It with testcases.

adityapatwardhan (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
iSazonov: Above we check whether Schema present - it is optional parameter. If Schema is absent here _jschema is null and we skip the test aganist a json schema.

test/powershell/Modules/Microsoft.PowerShell.Utility/Test-Json.Tests.ps1
iSazonov: Done.

test/powershell/Modules/Microsoft.PowerShell.Utility/Test-Json.Tests.ps1
iSazonov: I tried to do it, but there are too many parameters, and these tests are losing visibility. Maybe leave as is?

SteveL-MSFT:
To set expectation, since this depends on a component that has a non-permissive license, it may take some time to get this approved.  Also, since the [Ignite](https://www.microsoft.com/en-us/ignite/default.aspx) conference is this week, many of the approvers are not here.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
PaulHigin: Ok.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
PaulHigin: This is a required property.  Json check cannot be run without a valid schema property.  Make this property "mandatory" and let PowerShell do the validation rather than have an explicit check.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
PaulHigin: Schema should be a required "mandatory" property since the Json check cannot be run without it.  Make Schema mandatory and then the two explicit null checks are not necessary.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
iSazonov: What if an user want only check json parsing? Why we should block the scenario?

PaulHigin: Ok, I understand now.  Can you add a comment making it clear that Schema is optional and without it Json parsing is still performed and of value?  I assume this validates that the json is well formed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// Validate an Json.
+        /// </summary>
+        protected override void ProcessRecord()
+        {
+            JObject parsedJson = null;
+            bool result = true;
+
+            try
+            {
+                parsedJson = JObject.Parse(Json);
```
PaulHigin: I am not familiar with this framework.  Do we need to be concerned about code injection here when passing the input string directly to the Parse() API?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
iSazonov: I added the comment in Schema parameter description and in the PR description.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
iSazonov: The initializer was removed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
iSazonov: Good question.
@JamesNK Could you please comment? Should we be concerned about code injection in `JObject.Parse(Json)` ?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
PaulHigin: Please disregard.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/TestJsonCommand.cs
iSazonov: We can look at the [Parse source code](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Linq/JObject.cs#L423). It uses `StringReader` at low level and then tokenize the string stream. 
I don't see threats but I haven't an experience in the security area.

PaulHigin: I agree that it is an unlikely threat with this Json reader/parser, but I wanted to raise awareness for possible code injection vulnerabilities when we pass external input to a 3rd party API.

TravisEz13:
FYI, there are merge conflicts.

iSazonov:
Rebased to resolve merge conflicts.

iSazonov:
@SteveL-MSFT Could you ping license approvers? Also can we ask the author of the package to change/add MIT license?

SteveL-MSFT:
@iSazonov since it appears that [Json.Net Schema](https://www.newtonsoft.com/jsonschema) is how NewtonSoft makes money, I doubt they will change the license.  Still waiting on the approval, but to be honest with a non-permissive license, this will likely not be approved as I don't think it's compatible with the MIT license.  Your best option might be to publish this as a separate module on the PowerShellGallery as long as it adheres to the AGPL.

iSazonov:
@SteveL-MSFT Thanks. I hope that if they offer AGPL for open source projects, they don't lose anything with adding MIT.

SteveL-MSFT:
@iSazonov so after talking to legal, the end result is that if we accept this PR which will pull in the AGPL assembly and we distribute that package, then we would have to re-license PSCore6 under the AGPL to comply with the AGPL.  I think the alternate path is for you to publish this code under AGPL in your own repo and publish the resulting module on PowerShellGallery and then there is no issue with someone using a AGPL module with our MIT PSCore6.

iSazonov:
@SteveL-MSFT I've made a request to obtain a suitable license.
If the answer will be no, I guess we have to find another way, because Json is very desirable to have inside.

iSazonov:
Because of the licensing problem, I closed the PR and I'll open a new PR with another library.

PR #4900, opened 2017-09-22
Fix importing module remote using version filters and added tests

### LINKED ISSUE #4161: Add tests for Import-Module / Get-Module over remoting
Add tests for Import-Module / Get-Module over PSRP and CIMSession
Related code coverage report:

https://codecov.io/gh/PowerShell/PowerShell/tree/9e41c647a9407e05d86f4971a6bd25719a7130d0/src/System.Management.Automation/engine/Modules

Environment data
----------------

```powershell
> $PSVersionTable
OS                             Microsoft Windows 10.0.16204

```

SteveL-MSFT:
added tests for remote import-module
fixed issue with remotely importing module where it was checking the version filter incorrectly against the proxy module
  - the filters are applied when importing the module remotely
  - after proxy module is generated it always has a module version of 1.0, so the filters will always fail when importing the proxy locally

Fix https://github.com/PowerShell/PowerShell/issues/4161

### DIFF
--- a/src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs
+++ b/src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs
@@ -989,11 +989,27 @@ namespace Microsoft.PowerShell.Commands
                 try
                 {
                     this.ArgumentList = new object[] { psSession };
-                    // since we already applied these filters remotely, we don't want to apply it locally where the generated module is always 1.0
+
+                    // The correct module version has already been imported from the remote session and created locally. 
+                    // The locally created module always has a version of 1.0 regardless of the actual module version
+                    // imported from the remote session, and version checking is no longer needed and will not work while 
+                    // importing this created local module.
+                    Version originalBaseMinimumVersion = BaseMinimumVersion;
+                    Version originalBaseMaximumVersion = BaseMaximumVersion;
+                    Version originalBaseRequiredVersion = BaseRequiredVersion;
                     BaseMinimumVersion = null;
                     BaseMaximumVersion = null;
                     BaseRequiredVersion = null;
-                    ImportModule_LocallyViaName(importModuleOptions, wildcardEscapedPsd1Path);
+                    try
+                    {
+                        ImportModule_LocallyViaName(importModuleOptions, wildcardEscapedPsd1Path);
+                    }
+                    finally
+                    {
+                        BaseMinimumVersion = originalBaseMinimumVersion;
+                        BaseMaximumVersion = originalBaseMaximumVersion;
+                        BaseRequiredVersion = originalBaseRequiredVersion;
+                    }
                 }
                 finally
                 {
--- a/src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs
+++ b/src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs
@@ -986,6 +986,9 @@ namespace Microsoft.PowerShell.Commands
                 // import the proxy module just as any other local module
                 //
                 object[] oldArgumentList = this.ArgumentList;
+                Version originalBaseMinimumVersion = BaseMinimumVersion;
+                Version originalBaseMaximumVersion = BaseMaximumVersion;
+                Version originalBaseRequiredVersion = BaseRequiredVersion;
                 try
                 {
                     this.ArgumentList = new object[] { psSession };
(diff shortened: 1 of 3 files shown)


src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs
```diff
@@ -989,6 +989,10 @@ private PSModuleInfo ImportModule_LocallyViaName(ImportModuleOptions importModul
                 try
                 {
                     this.ArgumentList = new object[] { psSession };
+                    // since we already applied these filters remotely, we don't want to apply it locally where the generated module is always 1.0
```
PaulHigin: I am a little worried about changing these object properties here permanently.  It looks like this method is called within a loop of many kinds of different modules, including CIM and remote modules.  I think we should change and restore in the try/finally to be safe.

This is pretty confusing and I wonder if we can add a clearer comment?  Something like: "The correct module version has been imported from the remote session and created locally.  The locally created module always has a version of 1.0 regardless of the actual module version imported from the remote session, and version checking is no longer needed and will not work while importing this created local module".

I suppose the right thing to do is to ensure the generated module has the same version info as the module imported from the remote session.  But that is a much bigger change and can be handled in a separate issue.

src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs
```diff
@@ -989,6 +989,10 @@ private PSModuleInfo ImportModule_LocallyViaName(ImportModuleOptions importModul
                 try
                 {
                     this.ArgumentList = new object[] { psSession };
```
PaulHigin: I don't understand why the PSSession is added/reverted from the ArgumentList here.  The remote session has already been imported (via the Export-PSSession Cmdlet) and is no longer needed.  It looks like this can be removed.

test/powershell/Modules/Microsoft.PowerShell.Core/RemoteImportModule.Tests.ps1
```diff
@@ -0,0 +1,111 @@
...
+        Remove-Module TestImport -Force -ErrorAction SilentlyContinue
+    }
+
+    It "Import-Module can be called as an API with '<parameter>' = '<value>'" -TestCases @(
+        @{parameter = "Global"             ; value = $true},
+        @{parameter = "Global"             ; value = $false},
+        @{parameter = "Prefix"             ; value = "Hello"},
+        @{parameter = "Name"               ; value = "foo","bar"},
+        @{parameter = "FullyQualifiedName" ; value = @{ModuleName='foo';RequiredVersion='0.0'},@{ModuleName='bar';RequiredVersion='1.1'}},
+        @{parameter = "Assembly"           ; script = { [System.AppDomain]::CurrentDomain.GetAssemblies() | Select-Object -First 2 }}
```
PaulHigin: Can't the Assembly array be pre-computed and passed in as a "value" parameter?  Then you wouldn't need the special case for "script" parameter.

PaulHigin (changes requested):

src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs
PaulHigin: Actually this does need to be here.  It is used to fill the PSSession property on the module info.

src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs
SteveL-MSFT: I'll update the comment with your suggestion.  

I initially looked to see about updating the proxy module version, but there were a couple of problems.  Not only was it not trivial to update the module version, based on my reading of the code, because you can export-pssession multiple modules, you actually don't have a single module version to use so from that perspective 1.0 makes sense.  Of course, we could special case single modules to have the same version number, but a bunch of plumbing code would be needed.

test/powershell/Modules/Microsoft.PowerShell.Core/RemoteImportModule.Tests.ps1
SteveL-MSFT: My thinking is that if we add new test cases here, we can reuse `script` even though it's only used once here.  Also, I don't like having variables defined at the top of the script in a `BeforeAll` statement if it's only used once for a specific `It`, prefer to have it closer to the code where it's being used.

src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs
SteveL-MSFT: As far as I could tell, it seemed safe to change it, but I can store it and change it back.

SteveL-MSFT:
@PaulHigin feedback addressed

src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs
```diff
@@ -989,7 +989,27 @@ private PSModuleInfo ImportModule_LocallyViaName(ImportModuleOptions importModul
...
+                    // The locally created module always has a version of 1.0 regardless of the actual module version
+                    // imported from the remote session, and version checking is no longer needed and will not work while 
+                    // importing this created local module.
+                    Version originalBaseMinimumVersion = BaseMinimumVersion;
+                    Version originalBaseMaximumVersion = BaseMaximumVersion;
+                    Version originalBaseRequiredVersion = BaseRequiredVersion;
+                    BaseMinimumVersion = null;
+                    BaseMaximumVersion = null;
+                    BaseRequiredVersion = null;
+                    try
```
PaulHigin: You can just use the existing try/catch (for this.ArgumentList).  I don't think a new one is needed.

SteveL-MSFT: Yes, didn't notice that.  Will fix.

PaulHigin (approved):
LGTM

daxian-dbw:
@SteveL-MSFT Can you run `git submodule update`? The PR has changs to Pester files.

SteveL-MSFT:
@daxian-dbw Fixed.

daxian-dbw:
@SteveL-MSFT there are 2 test failures in AppVeyor run. Maybe they are not related to your changes but can you please take a look?

daxian-dbw:
Oh, and it seems the line 3200 and 3201 are rendered as red incorrectly :)
```
Describing Validate Invoke-WebRequest and Invoke-RestMethod -InFile
   Context InFile parameter negative tests
```

SteveL-MSFT:
@daxian-dbw webcmdlets are failing on two tests and it's affecting two of my PRs, trying a rebase to see if changes in master fixed something

daxian-dbw:
@SteveL-MSFT I don't think those 2 failing tests are related to changes in this PR. I will merge it.

PR #4902, opened 2017-09-22
Create generic Linux-x64 packages that are portable to all supported Linux distros

daxian-dbw:
Update build and package scripts to create portable Linux packages.
Fix the second task of #3961
- [x] Update packaging script to package native binaries published and remove unneeded dependencies

Help Needed
--------------
**I need help from the team to verify the single DEB/RPM package on different distros.**

Background Information
--------------------
- "RID `Linux-x64` isn't any different from `Ubuntu.16.04-x64`. The change we've made to enable the portable Linux-x64 rid was that we dynamically load OpenSSL and ICU libraries at runtime and we can also handle few missing APIs in various versions of CURL, so we can handle subtle differences between the distros. We don't build any distro specific binaries, so the `dotnet publish` publishes the same files when using those 2 RIDs respectively." See https://github.com/PowerShell/PowerShell/issues/3961#issuecomment-330796735

- "if you build the libpsl-native.so on CentOS 7, then it should be as portable as the dotnet stuff (we build it on CentOS 7 to ensure that it doesn't depend on newer glibc)." See https://github.com/PowerShell/PowerShell/issues/3961#issuecomment-331009082

Therefore, we need to use the `Linux-x64` `libpsl-native.so` built on CentOS 7 for all Linux distros. The way to do that is to create nuget package for `libpsl` and consume it from our build.

Native Dependencies of `dotnet-runtime-2.0.0` packages
--------------------------------------------------------------
The dependencies below were got by inspecting the `dotnet-runtime-2.0.0` packages (You can get those packages from https://www.microsoft.com/net/download/linux).
### DEB
- Ubuntu 16.04 depends
   - libc6, libcurl3, libgcc1, libgssapi-krb5-2, liblttng-ust0, libstdc++6, libunwind8, libuuid1, zlib1g, libssl1.0.0, libicu55
- Ubuntu 14.04 depends
   - libc6, libcurl3, libgcc1, libgssapi-krb5-2, liblttng-ust0, libstdc++6, libunwind8, libuuid1, zlib1g, libssl1.0.0, libicu52
- Ubuntu 17.04 depends
   - libc6, libcurl3, libgcc1, libgssapi-krb5-2, liblttng-ust0, libstdc++6, libunwind8, libuuid1, zlib1g, libssl1.0.0, libicu57
- Debian-stretch (9) depends
   - libc6, libcurl3, libgcc1, libgssapi-krb5-2, liblttng-ust0, libstdc++6, libunwind8, libuuid1, zlib1g, libssl1.0.2, libicu57
- Debian-jessie (8) depends
   - libc6, libcurl3, libgcc1, libgssapi-krb5-2, liblttng-ust0, libstdc++6, libunwind8, libuuid1, zlib1g, libssl1.0.0, libicu52

### RPM
- CentOS 7/Oracle Linux 7/openSUSE depend
   - openssl-libs, libcurl
   - However, when installing `dotnet-sdk-2.0.0`, two more packages are required to be manually installed: libunwind, libicu.
- Fedora 26
   - openssl-libs, libcurl
   - However, when installing `dotnet-sdk-2.0.0`, three more packages are required to be manually installed: libunwind, libicu, compat-openssl10

As shown above:
The only difference in DEB world is libicu (libicu52, libicu55, libicu57).
The only difference in RPM world is that Fedora needs `compat-openssl10` (see https://www.microsoft.com/net/core#linuxfedora). **`compat-openssl10` might be required only for dotnet-sdk toolings, but I'm not sure.**

Major Changes
-----------------
- Move the Unix native component build out of `Start-PSBuild` to `Start-BuildNativeUnixBinaries`
- Pack `libps1-native` into a nuget package `libpsl`. It includes binaries for `linux-x64`, `linux-arm` and `osx`. **NOTE:** the `linux-x64` version `libps1-native.so` is built on CentOS 7.
- Update the package script to use the correct dependencies and also use names for .deb and .rpm names.
   - **NOTE1:** for DEB, since `libicu` is different on different distros, I decided to use `libicu-dev` which will pull in the correct packages for `libicu` for different distros.
   - **NOTE2:** for RPM, since `compat-openssl10` is only available to Fedora, I don't include it in the dependencies. We can do the same as .NET Core (https://www.microsoft.com/net/core#linuxfedora), asking Fedora users to install `compat-openssl10` manually.
   - **NOTE3:** the updated package names are: `powershell_6.0.0-beta.20_amd64.deb` and `powershell-6.0.0_beta.22-1.rhel.7.x86_64.rpm`. I use `rhel.7` because dotnet-runtime-2.0.0 use it in the name `dotnet-runtime-2.0.0-rhel.7-x64.rpm`.

Distros that I have verified
-----------------------------
1. DEB package built from Ubuntu.16.04
    - Works on Ubuntu.16.04 and Ubuntu.14.04
    - Haven't tried on Ubuntu.17.04, Debian 8 and Debian 9
2. RPM package built from CentOS 7
    - Works on CentOS7, openSUSE.42.2
       - **Attention:** installing RPM on openSUSE has the same issue as `dotnet-runtime-2.0.0` package. It's quoted below from https://www.microsoft.com/net/core#linuxopensuse.
       - Haven't tried on other RedHat distros
> "Note: When installing the SDK, SUSE and OpenSUSE may report that nothing provides libcurl. libcurl should already be installed on supported versions of both distros. Run zypper search libcurl to confirm. The error will present 2 'solutions'. Choose 'Solution 2' to continue installing .NET Core."

### DIFF
--- a/tools/packaging/packaging.psm1
+++ b/tools/packaging/packaging.psm1
@@ -246,10 +246,6 @@ function New-UnixPackage {
         [Parameter(Mandatory)]
         [string]$Version,
 
-        # Package iteration version (rarely changed)
-        # This is a string because strings are appended to it
-        [string]$Iteration = "1",
-
         [Switch]
         $Force
     )
--- a/tools/packaging/packaging.psm1
+++ b/tools/packaging/packaging.psm1
@@ -30,7 +30,7 @@ function Start-PSPackage {
         [Switch] $Force,
 
         [Switch] $IncludeSymbols,
-        
+
         [Switch] $SkipReleaseChecks
     )
 
@@ -43,11 +43,9 @@ function Start-PSPackage {
         New-PSOptions -Configuration "Release" -WarningAction SilentlyContinue | ForEach-Object { $_.Runtime, $_.Configuration }
     }
 
-    # We convert the runtime to win7-x64 or win7-x86 to build the universal windows package.
     if($Environment.IsWindows) {
-        $Runtime = $Runtime -replace "win\d+", "win7"
-
-        # Build the name suffix for win-plat packages
+        # Runtime will always be win7-x64 or win7-x86 on Windows.
+        # Build the name suffix for universal win-plat packages.
         $NameSuffix = $Runtime -replace 'win\d+', 'win'
     }
 
@@ -134,7 +132,7 @@ function Start-PSPackage {
     }
     log "Packaging Type: $Type"
 
-    # Add the symbols to the suffix 
+    # Add the symbols to the suffix
     # if symbols are specified to be included
     if($IncludeSymbols.IsPresent -and $NameSuffix) {
         $NameSuffix = "symbols-$NameSuffix"
@@ -186,7 +184,7 @@ function Start-PSPackage {
             {
                 throw "AppImage does not support packaging '-IncludeSymbols'"
             }
-            
+
             if ($Environment.IsUbuntu14) {
                 $null = Start-NativeExecution { bash -iex "$PSScriptRoot/../appimage.sh" }
                 $appImage = Get-Item PowerShell-*.AppImage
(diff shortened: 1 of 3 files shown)


tools/packaging/packaging.psm1
```diff
@@ -408,73 +402,28 @@ function New-UnixPackage {
             "libc6",
             "libcurl3",
             "libgcc1",
-            "libssl1.0.0",
+            "libgssapi-krb5-2",
```
TravisEz13: The package changes have been tested with currently supported platforms at least?

tools/packaging/packaging.psm1
```diff
@@ -408,73 +402,28 @@ function New-UnixPackage {
...
+            "liblttng-ust0",
             "libstdc++6",
-            "libtinfo5",
             "libunwind8",
             "libuuid1",
-            "zlib1g"
+            "zlib1g",
+            "libssl1.0.0",
+            "libicu-dev"
```
TravisEz13: We should probably open an issue to see if we can get to use the runtime package instead of the dev package.

TravisEz13 (approved):

TravisEz13:
@adityapatwardhan  Can you review?

adityapatwardhan (approved):
LGTM

tools/packaging/packaging.psm1
daxian-dbw: I have tested the DEB package on Ubuntu.16.04 and Ubuntu.14.04, but not on Debian 8 and Debian 9 yet.

tools/packaging/packaging.psm1
daxian-dbw: Please add one more task to the existing #3961 to track it.

Issue #3957, opened 2017-06-06
Enable console GUIs with PowerShell


SteveL-MSFT:
As people are going back to the console for management (instead of primarily GUI), we should support curses to enable console based GUIs.

- [ ] out-gridview implementation
- [ ] get-member interactive navigation
- [ ] get-help -showwindow (for this, my suggestion is to not literally implement `showwindow`, but to improve console rendering of help with interactive navigation)
- [ ] show-command

iSazonov:
Show-Command?

Bartolomeus-649:
People are not "going back" to consoles for management!  They are forced to use it as there is no real alternatives!

MaximoTrinidad:
In one of the posted issues: https://github.com/PowerShell/PowerShell/issues/1808 

I mention about using Python Tk with PowerShell.  I think Python would be a BIG candidate to help overcoming the missing GUI component in PowerShell.

![myazuredemo_02_2017-06-11_12-03-51](https://user-images.githubusercontent.com/19892435/27436077-19fc682e-572c-11e7-9a8c-e616f5884e6c.jpg)

:)

iSazonov:
I guess Microsoft love Xamarin 😄

SQLDBAWithABeard:
Yes to enabling a seperate searchable help window a la -showwindow. That's a useful capability

megamorf:
Even though I already 👍 this months ago, I wanted to let you guys know that Out-Gridview is a command that lots of people use to quickly preview and filter datasets. It's become a great asset in the typical admin toolbelt :-)

swngdnz:
I view curses as only a stop-gap measure. OGV based on curses is going to be quite weak in comparison to OGV on FullCLR (and yes, I'm another person that uses OGV extensively when developing solutions to examine result data-sets, even though OGV may not end up in my published solution). Whether it's tk or something else, I believe you will need to pick a true xplat gui solution.

JanMulkens:
Haven't used PS6 yet, but this is going to be missed when i do migrate

dkateiva:
I hope ogv can be implemented in some way. It is really convenient sometimes

SteveL-MSFT:
@swngdnz for "true xplat gui", there's really only two viable solutions: vt100 and html.  With vt100, it'll work across SSH, locally, as well as in CloudShell.  Html is more limited as it'll only work locally, but it also simpler to implement.  I encourage someone from the community to build such a module.

[...]

PR #4903, opened 2017-09-23
remove unsupported -showwindow switch from get-help

SteveL-MSFT:
`-showwindow` relies on WPF which is not supported on CoreCLR
also removed any `#if` for CoreCLR

doc update https://github.com/PowerShell/PowerShell-Docs/pull/1685


iSazonov:
Should we track this in #3957 ?

SteveL-MSFT:
@iSazonov I'll add it to #3957

adityapatwardhan (approved):
LGTM

WilliamXieMSFT:
With recent announcements for wpf support on dotnetcore 3, would readding this feature be considered in the future?

PR #4905, opened 2017-09-23
Replace httpbin.org/delay Tests with WebListener

markekraus:
Unfortunately, there were some odd things coupled with this change that were all somewhat interdependent.

* Adds `Delay` test to WebListener 
* Adds `-TestValue` parameter to `Get-WebListenerUrl` to support tests which take a value such as encoding, redirect, and delay
* Adjusts existing tests for encoding and redirect to use `-TestValue`
* Removes a repeated timeout test
* Replaces httpbin.org/delay tests with Weblistener URL
* Fixes the tests

I discovered in switching this that the URL's that were being used (e.g. `http://httpbin.org/delay/:5`) were actually pointing to a page that immediately returned a `505` error. The pages were not actually delayed and the web cmdlets were not actually timing out. The tests were passing, but they were checking for the wrong Error ID. I had to correct the error ID in all the tests.

~~Ironically, httpbin.org is having a ton of issues right now. I will push a feature tag later after they have stabilized.~~ Feature tag pushed.

reference #2504

### DIFF
--- a/test/tools/WebListener/README.md
+++ b/test/tools/WebListener/README.md
@@ -59,6 +59,30 @@ Response when certificate is not provided in request:
 }
 ```
 
+## /Delay/
+
+Returns the same results as the Get test. If a number is supplied, the server will wait that many seconds before returning a response. This can be used to test timeouts.
+
+```powershell
+Invoke-WebRequest -Uri 'http://localhost:8083/Delay/5'
+```
+
+After 5 Seconds:
+
+```json
+{
+  "args": {
+    
+  },
+  "origin": "127.0.0.1",
+  "headers": {
+    "User-Agent": "Mozilla/5.0 (Windows NT; Windows NT 10.0; en-US) WindowsPowerShell/5.1.15063.608",
+    "Host": "localhost:8083"
+  },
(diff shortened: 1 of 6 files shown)


TravisEz13 (approved):

adityapatwardhan (approved):
LGTM

test/tools/WebListener/README.md
```diff
@@ -59,6 +59,30 @@ Response when certificate is not provided in request:
 }
 ```
 
+## /Delay/
+
+Returns the same results as the Get test. If a number is supplied, the server will wait that many seconds before returning a response. This can be used to test timeouts.
+
+```powershell
+Invoke-WebRequest -Uri 'http://localhost:8083/Delay/5'
```
dantraMSFT: Most of the tests are using the Get-WebListenerUrl function. Suggest adding an example for that as well.

$uri = Get-WebListenerUrl -Test 'Delay' -TestValue '5'

markekraus: I will add those on the next iteration. Created #4918 to rack.

Issue #2579, opened 2016-11-01
Get-Service should return the user account the services are running under


joeyaiello:
Currently the #1 item on the PowerShell UserVoice: https://windowsserver.uservoice.com/forums/301869-powershell/suggestions/12336204-get-service-should-return-the-user-account-the-ser

Arnaud Petitjean:
> The Get-Service cmdlet should return the service account associated to each service.

> We can get this information easily from the Win32_Service WMI class, but to me, it should be returned by Get-Service.

Expected behavior
-----------------
```powershell
Get-Service Spooler | Format-List Name,LogOnAs

Name : Spooler
LogOnAs : Local System
```

Actual behavior
---------------
```powershell
Get-Service Spooler | Format-List Name,LogOnAs

Name : Spooler
```

Environment data
----------------
All versions of PowerShell

copdips:
please also return StartupType and ServicePath

joeyaiello:
@copdips StartType as been included in `Get-Service` output since 5.0:  

```powershell
C:\Users\jaiello> get-service sshd | fl *


Name                : sshd
RequiredServices    : {ssh-agent}
CanPauseAndContinue : False
CanShutdown         : False
CanStop             : False
DisplayName         : sshd
DependentServices   : {}
MachineName         : .
ServiceName         : sshd
ServicesDependedOn  : {ssh-agent}
ServiceHandle       :
Status              : Stopped
ServiceType         : Win32OwnProcess
StartType           : Manual
Site                :
Container           :
```

joeyaiello:
But yeah, we should probably have `ServicePath` too

charub:
@SteveL-MSFT , @lzybkr  @joeyaiello  : As mentioned, Service Account info as well service path is available from Win32_Service WMI class. Should I get started on this?

![image](https://cloud.githubusercontent.com/assets/18017330/20025565/c509e960-a2ac-11e6-9281-6d5a87da6b38.png)

SteveL-MSFT:
@charub you can always take up issues marked as Backlog, although it's preferred if you can prioritize issues you can fix marked for 6.0.0 milestone

KirkMunro:
Some food for thought: Don't forget the Description property, and consider device drivers as well.

Years ago when I used to work on PowerGUI (PowerShell v1 days, before StartType was added), we solved this problem like this:

```PowerShell
$wmiServices = @{}
[...]

PR #2850, opened 2016-12-07
Adding Parent extended property to System.Diagnostics.Process

powercode:
Fixes #2763 
System.Management.Automation:
Adding ProcessCodeMethods.cs to do the pinvokes and the mandatory timestamp check
for parent processes.
Adding 'Parent' CodeProperty to Process in Types_Ps1Xml.generates.cs 
For unix, ```/proc/<pid>/stat``` is parsed.
Moving addition of process PSProperties to SMA ClrFacade.

PR #4907, opened 2017-09-23
Added BinPath, Description, UserName and Delayed Auto Start to Get-Service

joandrsn:
I had the time to look into this issue #2579

I have added the following to the `Get-Service` command:
* BinaryPathName (Was suggested as `ServicePath`)
* Description
* UserName (Was suggested as `LogOnAs`)
* DelayedAutoStart
* ServiceStartupType

The ServiceStartupType was also added, providing the user access to use services like in `services.msc`, having `Automatic`, `Automatic (Delayed Start)`, `Manual` and `Disabled` as options (Also includes `InvalidValue` as an initial value for `ServiceStartupType`)

I'm not completely sure if my naming of the variables are correct and would like some feedback.
I have tried to get a complete overview of what the different properties are called in the different ways to interact with services (With my additions marked like ***this***).

| New-Service | Set-Service  | Get-Service            | Get-CimInstance     |
| ----------- | ------------ | ---------------------- | ------------------- |
| N/A         | Computername | N/A                    | SystemName?         |
| Name        | Name         | Name                   | Name                |
| DisplayName | DisplayName  | DisplayName            | DisplayName/Caption |
| Description | Description  | ***Description***      | Description         |
| StartupType | StartupType  | ***StartupType***      | Startmode (string)  |
| N/A         | Status       | Status                 | State               |
| Credentials | Credentials  | ***UserName****        | StartName*          |
| BinaryPathName     | N/A          | ***BinaryPathName***          | PathName            |
| DependsOn   | N/A          | RequiredServices       | N/A                 |
| N/A         | N/A          | N/A                    | DelayedAutoStart    |
| N/A         | N/A          | N/A                    | ProcessId           |

*Only username of the account running the service.

When adding the properties, I used NotePropertie and I don't know if it's possible to make autocomplete for this.
I would love to have the ability to type:
`$service = Get-Service -Name 'spooler'`
`$service.Use<TAB>` => `$service.UserName`

Another thing: I'm not sure if I should/can split my function for adding properties up into smaller pieces since it's about 100 lines currently.

I might remove DelayedAutoStart from the pull request since people might rather want it as a StartupType instead. I currently have it in since people might find the information useful on its own.


src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -689,6 +689,207 @@ protected override void ProcessRecord()
...
+            PSObject serviceAsPSObj = PSObject.AsPSObject(service);
+            try
+            {
+                hScManager = NativeMethods.OpenSCManagerW(
+                    lpMachineName: service.MachineName,
+                    lpDatabaseName: null,
+                    dwDesiredAccess: NativeMethods.SC_MANAGER_CONNECT
+                );
+                if (IntPtr.Zero == hScManager) {
+                    lastError = Marshal.GetLastWin32Error();
```
SteveL-MSFT: Seems like by the time this is called, we know `service` is valid so I suppose you can reasonably assume any failures are due to Access Denied (however, I didn't look at all the possible return values from the win32 apis).  However, I think you should at least add a comment about why you're not checking the `lastError` code.  Applies to the other ones below as well.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -689,6 +689,207 @@ protected override void ProcessRecord()
...
+                    Win32Exception exception = new Win32Exception(lastError);
+                    WriteNonTerminatingError(
+                        service,
+                        exception,
+                        "CouldNotGetServiceInfo",
+                        ServiceResources.CouldNotGetServiceInfo,
+                        ErrorCategory.PermissionDenied);
+                }
+                NativeMethods.SERVICE_DELAYED_AUTO_START_INFO autostartInfo = (NativeMethods.SERVICE_DELAYED_AUTO_START_INFO)Marshal.PtrToStructure(lpBuffer, typeof(NativeMethods.SERVICE_DELAYED_AUTO_START_INFO));
```
SteveL-MSFT: I don't have a strong opinion on this one, but my preference is to have it as part of StartupType similar to what you see in services.msc

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -689,6 +689,207 @@ protected override void ProcessRecord()
...
+                        exception,
+                        "CouldNotGetServiceInfo",
+                        ServiceResources.CouldNotGetServiceInfo,
+                        ErrorCategory.PermissionDenied);
+                }
+
+                // Description information
+                IntPtr lpBuffer = IntPtr.Zero;
+                DWORD bufferSizeNeeded = 0;
+                bool status = NativeMethods.QueryServiceConfig2W(
```
SteveL-MSFT: If seems like the `QueryServiceConfig2W` api along with the error handling could be an internal helper api since it's repeated pretty often.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -689,6 +689,207 @@ protected override void ProcessRecord()
...
+                        "CouldNotGetServiceInfo",
+                        ServiceResources.CouldNotGetServiceInfo,
+                        ErrorCategory.PermissionDenied);
+                }
+                NativeMethods.SERVICE_DELAYED_AUTO_START_INFO autostartInfo = (NativeMethods.SERVICE_DELAYED_AUTO_START_INFO)Marshal.PtrToStructure(lpBuffer, typeof(NativeMethods.SERVICE_DELAYED_AUTO_START_INFO));
+
+                // General information
+                lpBuffer = IntPtr.Zero;
+                bufferSizeNeeded = 0;
+                status = NativeMethods.QueryServiceConfigW(
```
SteveL-MSFT: Same with `QueryServiceConfigW` as helper api

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -689,6 +689,207 @@ protected override void ProcessRecord()
...
+                }
+                hService = NativeMethods.OpenServiceW(
+                    hScManager,
+                    service.ServiceName,
+                    NativeMethods.SERVICE_QUERY_CONFIG
+                );
+                if (IntPtr.Zero == hService) {
+                    lastError = Marshal.GetLastWin32Error();
+                    Win32Exception exception = new Win32Exception(lastError);
+                    WriteNonTerminatingError(
```
SteveL-MSFT: I think rather than writing a non-terminating error for each property that fails to be retrieved, it might be better to just write one non-terminating error if any fail.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -235,6 +235,85 @@ Describe "Set/New/Remove-Service cmdlet tests" -Tags "Feature", "RequireAdminOnW
         { Remove-Service -Name "testremoveservice" -ErrorAction 'Stop' } | ShouldBeErrorId "InvalidOperationException,Microsoft.PowerShell.Commands.RemoveServiceCommand"
     }
 
+    It "Get-Service can get the username of a service" {
+        try {
+            $userName = "user1"
```
SteveL-MSFT: Because there's very similar code for these test cases, you could turn them into `-TestCases` like:

```powershell

# move user creation to BeforeAll{} and cleanup in AfterAll{}

It "Get-Service can get the <property> of a service" -TestCases @(
  @{property="username"; value="user1"; parameters = @{ Credential = $creds },
  @{property="binpath";value="$PSHOME\powershell.exe"},
 ...
){
 param ($property, $value, $parameters)

  if ($parameters -eq $null) {
    $parameters = @{}
  }
  $parameters += @{Name=$servicename;binarypathname="...}
  $service = New-Service @parameters
  ...
  $service.$property | Should BeExactly $value
}
```

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
joandrsn: Can we extend Enums using PowerShell somehow?
C# does not seem to be able to do it using just inheritance.
Currently the [ServiceStartMode][0] class does not support `Automatic (Delayed start)`. 
I could also make my own class, but that would be a breaking change, correct?

[0]: https://msdn.microsoft.com/en-us/library/system.serviceprocess.servicestartmode(v=vs.110).aspx

SteveL-MSFT: Seems like having it separate would be simplest.

joandrsn:
I have converted the `StartupType` to a custom enum. It's now possible to use to specify AutomaticDelayedStart as a StartupType.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -1506,17 +1609,7 @@ public string Description
...
-            get { return startupType; }
-            set
-            {
-                startupType = value;
-            }
-        }
-        // We set the initial value to an invalid value so that we can
-        // distinguish when this is and is not set.
-        internal ServiceStartMode startupType = (ServiceStartMode)(-1);
+        public ServiceStartupType StartupType { get; set; } = (ServiceStartupType)(-1);
```
iSazonov: It seems the change is not related to the PR purpose. I'd prefer to move it in separate PR.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -1756,7 +1859,31 @@ protected override void ProcessRecord()
...
+                        if (!status)
+                        {
+                            int lastError = Marshal.GetLastWin32Error();
+                            Win32Exception exception = new Win32Exception(lastError);
+                            WriteNonTerminatingError(
+                                Name,
+                                DisplayName,
+                                Name,
+                                exception,
+                                "CouldNotNewServiceDescription",
```
SteveL-MSFT: Shouldn't this be related to DelayedAutoStart?

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2112,6 +2239,34 @@ protected override void BeginProcessing()
...
+                    if (!succeeded)
+                    {
+                        int lastError = Marshal.GetLastWin32Error();
+                        Win32Exception exception = new Win32Exception(lastError);
+                        WriteNonTerminatingError(
+                            Name,
+                            DisplayName,
+                            Name,
+                            exception,
+                            "CouldNotNewServiceDescription",
```
SteveL-MSFT: Shouldn't this be related to DelayedAutoStart?

SteveL-MSFT (changes requested):

iSazonov:
Does it make sense to use ETS here?

/cc @lzybkr

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
joandrsn: I'm not checking `lastError` because the documentation says `If the function fails, the return value is NULL.` [Source][0] and also because the other service classes called the function the same way.

I thought it would be better to have the potential exception speak for itself rather than assuming that the error is a Access Denied.

But I might be wrong. If I am, please tell me.

[0]: https://msdn.microsoft.com/en-us/library/windows/desktop/ms684323(v=vs.85).aspx

SteveL-MSFT: Thanks for the explanation, I'm ok with this.

joandrsn:
It has been 3 days since the latest update on this issue, so I wanted to bring the issue up again.
Are we waiting for @lzybkr to review this?

@iSazonov, you mention ETS. 
I initially tried to use ETS but it seemed at bit hard. Then I saw what pull request #2850 did it and used the `Properties`-method.
Are there any benifits to using ETS over using `Properties` directly?
Or does it generate more readable code?
Or does it maybe help PowerShell recognize properties and then enable tab-completion of the properties added?
I am curious :)

SteveL-MSFT:
I think we can defer the ETS discussion and redo it if necessary.

SteveL-MSFT (approved):
LGTM Thanks!

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -1756,7 +1859,31 @@ protected override void ProcessRecord()
                                 ErrorCategory.PermissionDenied);
                         }
 
+                        // Set the delayed auto start
+                        NativeMethods.SERVICE_DELAYED_AUTO_START_INFO ds = new NativeMethods.SERVICE_DELAYED_AUTO_START_INFO();
+                        ds.fDelayedAutostart = StartupType == ServiceStartupType.AutomaticDelayedStart;
+                        size = Marshal.SizeOf(ds);
+                        buffer = Marshal.AllocCoTaskMem(size);
+                        Marshal.StructureToPtr(ds, buffer, false);
```
iSazonov: I wonder where we destroy and free?

src/System.Management.Automation/utils/PInvokeDllNames.cs
```diff
@@ -136,5 +136,7 @@ internal static class PinvokeDllNames
         internal const string Process32NextDllName = "api-ms-win-core-toolhelp-l1-1-0";                      /*122*/
         internal const string GetACPDllName = "api-ms-win-core-localization-l1-2-0.dll";                     /*123*/
         internal const string DeleteServiceDllName = "api-ms-win-service-management-l1-1-0.dll";             /*124*/
+        internal const string QueryServiceConfigDllName = "api-ms-win-service-management-l2-1-0.dll";        /*125*/
+        internal const string QueryServiceConfig2DllName = "api-ms-win-service-management-l2-1-0.dll";       /*126*/
```
iSazonov: Have we the ApiSet on Windows 7?
/cc @mirichmo

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
iSazonov: @daxian-dbw Could you please comment?

daxian-dbw: Good catch! I don't find the allocated memory is freed anywhere. `FreeCoTaskMem` should be called once done using the memory. The same issue also happens in some other places in this file.

joandrsn:
This is my first time freeing memory.

I think I still have some memory which does not get dereferenced, but I just can't find it.
If I run this command: 
`1..10000 | % { $null = Get-Service -Name 'spooler' }`
I am not getting any less memory allocated, but only more (Very minor, but still relevant).

I'm not sure if I should use [AllocCoTaskMem][0] or [AllocHGlobal][1] and I have searched a lot to figure out which one I should use.
Currently, I'm using both, but I should probably use only one of the methods since it gives better readability.
I am however using the correct method for freeing memory, as [I should not mix][2].

I would also free the memory when the other functions are called, but that is a different issue and I won't include it in this pull request.

[0]: https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.marshal.alloccotaskmem(v=vs.110).aspx
[1]: https://msdn.microsoft.com/en-us/library/5cyb68cy(v=vs.110).aspx
[2]: https://msdn.microsoft.com/en-us/library/windows/desktop/aa366533(v=vs.85).aspx

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
iSazonov: I found `AllocCoTaskMem` in 4 files and ` FreeCoTaskMem` in 17 files - It seems we should widely review. Maybe don't block the PR and open a tracking Issue for new PR?

iSazonov:
@joandrsn AllocHGlobal is from 16-bit Windows. See https://msdn.microsoft.com/en-us/library/aa366596.aspx
AllocCoTaskMem is recommended.

I agree that we should open a tracking Issue to review our code for the problem with memory free.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2642,8 +2894,42 @@ internal static bool TryGetNativeStartupType(ServiceStartMode StartupType, out D
...
+                    break;
+            }
+            return result;
+        }
     }
     #endregion NativeMethods
+
+    #region ServiceStartupType
+    ///<summary>
+    ///Enum for usage with StartupType. Automatic, Manual & Disabled index matched from System.ServiceProcess.ServiceStartMode
```
iSazonov: Why is GitHub paints '&' in red?

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -4,9 +4,19 @@ Describe "Set/New/Remove-Service cmdlet tests" -Tags "Feature", "RequireAdminOnW
         if ( -not $IsWindows ) {
             $PSDefaultParameterValues["it:skip"] = $true
         }
+        if($IsWindows) {
```
iSazonov: Please add spaces - use common formatting convention. Below too.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -235,6 +241,33 @@ Describe "Set/New/Remove-Service cmdlet tests" -Tags "Feature", "RequireAdminOnW
         { Remove-Service -Name "testremoveservice" -ErrorAction 'Stop' } | ShouldBeErrorId "InvalidOperationException,Microsoft.PowerShell.Commands.RemoveServiceCommand"
     }
 
+    It "Get-Service can get the '<property>' of a service"  -TestCases @(
```
iSazonov: Extra space before "-TestCases".

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -4,9 +4,19 @@ Describe "Set/New/Remove-Service cmdlet tests" -Tags "Feature", "RequireAdminOnW
         if ( -not $IsWindows ) {
             $PSDefaultParameterValues["it:skip"] = $true
         }
+        if($IsWindows) {
+          $userName = "testuserservices"
```
iSazonov: Please use indentation with 4 spaces. Below too.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -235,6 +241,33 @@ Describe "Set/New/Remove-Service cmdlet tests" -Tags "Feature", "RequireAdminOnW
         { Remove-Service -Name "testremoveservice" -ErrorAction 'Stop' } | ShouldBeErrorId "InvalidOperationException,Microsoft.PowerShell.Commands.RemoveServiceCommand"
     }
 
+    It "Get-Service can get the '<property>' of a service" -TestCases @(
```
iSazonov: Please correct indentation in the test.

iSazonov:
@SteveL-MSFT We have new commits - Could you please take another look?

@daxian-dbw Could you please review and approve as code owner?

@mirichmo Could you please comment [about ApiSets](https://github.com/PowerShell/PowerShell/pull/4907#discussion_r141940852)?

iSazonov:
@joandrsn Thanks for great work! Please wait improvements from MSFT experts.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -689,6 +689,114 @@ protected override void ProcessRecord()
...
+                        exception,
+                        "CouldNotGetServiceInfo",
+                        ServiceResources.CouldNotGetServiceInfo,
+                        ErrorCategory.PermissionDenied);
+                }
+
+                bool queryStatus = NativeMethods.QueryServiceConfig2(hService, NativeMethods.SERVICE_CONFIG_DESCRIPTION, out descriptionStructPtr);
+                NativeMethods.SERVICE_DESCRIPTIONW description = (NativeMethods.SERVICE_DESCRIPTIONW)Marshal.PtrToStructure(descriptionStructPtr, typeof(NativeMethods.SERVICE_DESCRIPTIONW));
+
+                queryStatus = queryStatus & NativeMethods.QueryServiceConfig2(hService, NativeMethods.SERVICE_CONFIG_DELAYED_AUTO_START_INFO, out delayedAutoStartStuctPtr);
```
SteveL-MSFT: Shouldn't this be a `|`?

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -689,6 +689,114 @@ protected override void ProcessRecord()
...
+                NativeMethods.SERVICE_DELAYED_AUTO_START_INFO autostartInfo = (NativeMethods.SERVICE_DELAYED_AUTO_START_INFO)Marshal.PtrToStructure(delayedAutoStartStuctPtr, typeof(NativeMethods.SERVICE_DELAYED_AUTO_START_INFO));
+
+                queryStatus = queryStatus & NativeMethods.QueryServiceConfig(hService, out serviceConfigStructPtr);
```
SteveL-MSFT: Shouldn't this be a `|`?

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -689,6 +689,114 @@ protected override void ProcessRecord()
...
+
+                noteProperty = new PSProperty("Description", description.lpDescription);
+                serviceAsPSObj.Properties.Add(noteProperty, true);
+                serviceAsPSObj.TypeNames.Insert(0, "System.Service.ServiceController#Description");
+
+                noteProperty = new PSProperty("DelayedAutoStart", autostartInfo.fDelayedAutostart);
+                serviceAsPSObj.Properties.Add(noteProperty, true);
+                serviceAsPSObj.TypeNames.Insert(0, "System.Service.ServiceController#DelayedAutoStart");
+
+                noteProperty = new PSProperty("BinPath", serviceInfo.lpBinaryPathName);
```
SteveL-MSFT: Should this be spelled out as `BinaryPath`?

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2606,6 +2811,52 @@ internal static extern
...
 
+        internal static bool QueryServiceConfig(NakedWin32Handle hService, out IntPtr structurePointer)
+        {
+            IntPtr lpBuffer = IntPtr.Zero;
+            DWORD bufferSize, bufferSizeNeeded = 0;
+            bool status = NativeMethods.QueryServiceConfigW(
+                hSCManager: hService,
+                lpServiceConfig: lpBuffer,
+                cbBufSize: 0,
+                pcbBytesNeeded: out bufferSizeNeeded);
```
SteveL-MSFT: we should return immediately if status != true && GetLastError() != ERROR_INSUFFICIENT_BUFFER

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2606,6 +2811,52 @@ internal static extern
...
+        internal static bool QueryServiceConfig2(NakedWin32Handle hService, DWORD infolevel, out IntPtr structurePointer)
+        {
+            IntPtr lpBuffer = IntPtr.Zero;
+            DWORD bufferSize, bufferSizeNeeded = 0;
+            bool status = NativeMethods.QueryServiceConfig2W(
+                hService: hService,
+                dwInfoLevel: infolevel,
+                lpBuffer: lpBuffer,
+                cbBufSize: 0,
+                pcbBytesNeeded: out bufferSizeNeeded);
```
SteveL-MSFT: we should return immediately if status != true && GetLastError() != ERROR_INSUFFICIENT_BUFFER

src/Microsoft.PowerShell.Commands.Management/resources/ServiceResources.resx
```diff
@@ -162,15 +162,24 @@
...
+    <value>Service '{1} ({0})' automatic (delayed start) cannot be configured due to the following error: {2}</value>
+  </data>
   <data name="CouldNotNewService" xml:space="preserve">
     <value>Service '{1} ({0})' cannot be created due to the following error: {2}</value>
   </data>
   <data name="CouldNotNewServiceDescription" xml:space="preserve">
     <value>Service '{1} ({0})' was created, but its description cannot be configured due to the following error: {2}</value>
   </data>
+  <data name="CouldNotNewServiceDelayedAutoStart" xml:space="preserve">
+    <value>Service '{1} ({0})' was created, but its startuptype, automatic (delayed start), cannot be configured due to the following error: {2}</value>
```
SteveL-MSFT: I would suggesting changing this to:

```none
... its StartupType 'Automatic (Delayed Start)' could not be...
```

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
joandrsn: I renamed the variable to `querySuccessful` to help with the readability. I'm quite sure it's `&`. I also changed it to `&&` to make a short-circuit

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2112,6 +2245,35 @@ protected override void BeginProcessing()
...
+                        WriteNonTerminatingError(
+                            Name,
+                            DisplayName,
+                            Name,
+                            exception,
+                            "CouldNotNewServiceDelayedAutoStart",
+                            ServiceResources.CouldNotNewServiceDelayedAutoStart,
+                            ErrorCategory.PermissionDenied);
+                    }
+                    Marshal.FreeCoTaskMem(buffer);
```
mirichmo: Finally here as well

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2482,6 +2648,25 @@ internal static extern
             DWORD dwDesiredAccess
             );
 
+        [DllImport(PinvokeDllNames.QueryServiceConfigDllName, CharSet = CharSet.Unicode, SetLastError = true)]
```
mirichmo: @daxian-dbw Are we still following the PInvokeDLLNames convention? It no longer makes sense since we stripped out the full CLR code

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2606,6 +2811,52 @@ internal static extern
...
+            IntPtr lpBuffer = IntPtr.Zero;
+            DWORD bufferSize, bufferSizeNeeded = 0;
+            bool status = NativeMethods.QueryServiceConfig2W(
+                hService: hService,
+                dwInfoLevel: infolevel,
+                lpBuffer: lpBuffer,
+                cbBufSize: 0,
+                pcbBytesNeeded: out bufferSizeNeeded);
+
+            lpBuffer = Marshal.AllocCoTaskMem((int)bufferSizeNeeded);
```
mirichmo: A check should be made to confirm that status == ERROR_INSUFFICIENT_BUFFER before proceeding.

https://msdn.microsoft.com/en-us/library/windows/desktop/ms684935.aspx

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -2606,6 +2811,52 @@ internal static extern
...
+        {
+            IntPtr lpBuffer = IntPtr.Zero;
+            DWORD bufferSize, bufferSizeNeeded = 0;
+            bool status = NativeMethods.QueryServiceConfigW(
+                hSCManager: hService,
+                lpServiceConfig: lpBuffer,
+                cbBufSize: 0,
+                pcbBytesNeeded: out bufferSizeNeeded);
+
+            lpBuffer = Marshal.AllocCoTaskMem((int)bufferSizeNeeded);
```
mirichmo: Same issue here. It should not proceed unless it gets the correct status value.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
SteveL-MSFT: resolved

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
SteveL-MSFT: resolved

SteveL-MSFT (approved):
Great!  LGTM

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -689,6 +689,114 @@ protected override void ProcessRecord()
...
+                    WriteNonTerminatingError(
+                        service,
+                        exception,
+                        "CouldNotGetServiceInfo",
+                        ServiceResources.CouldNotGetServiceInfo,
+                        ErrorCategory.PermissionDenied);
+                }
+
+                bool queryStatus = NativeMethods.QueryServiceConfig2(hService, NativeMethods.SERVICE_CONFIG_DESCRIPTION, out descriptionStructPtr);
+                NativeMethods.SERVICE_DESCRIPTIONW description = (NativeMethods.SERVICE_DESCRIPTIONW)Marshal.PtrToStructure(descriptionStructPtr, typeof(NativeMethods.SERVICE_DESCRIPTIONW));
```
mirichmo: I'd like to see these method calls organized differently to protect against potential memory leaks. The pattern is repeated in NativeMethods.QueryServiceConfig2 and I'd like all instances of it fixed. Since NativeMethods.QueryServiceConfig() is a helper function, its output parameter, instead of being an IntPtr structurePointer, should be the managed object NativeMethods.SERVICE_DESCRIPTIONW. That way, the unmanaged memory that is allocated within that function is also cleaned up within that function. The managed object carries the values out of the function and eventually gets garbage collected. To fix this instance, line 741 should be moved within the helper function. You can probably make Nativemethods.QueryServiceConfig2() a generic method as well to avoid having to use a switch for infoLevel.

mirichmo (changes requested):

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
iSazonov: Interesting, what is best practice? Could we exclude second call by pre-allocate memory buffer?

mirichmo: This technique is a common pattern in Windows APIs. It is designed to be called twice with the first call returning the buffer size needed so the correct buffer size can be allocated and supplied on the second call. You can do it all in one call, but you have to either guess how big the buffer should be or allocate a max size buffer. If you guess wrong, you'll need to resize it and call the method again anyway. A max size buffer will work, but will incur unnecessary memory allocation. With that in mind, it is optimal to call it twice and follow the expected pattern.

joandrsn: I can't find this. I think your comment might be on an outdated version of the code. I made the check [here][0] and [here][1]

[0]: https://github.com/PowerShell/PowerShell/pull/4907/files#diff-5ee347b353d729b7e743da3c535d1a21R2835
[1]: https://github.com/PowerShell/PowerShell/pull/4907/files#diff-5ee347b353d729b7e743da3c535d1a21R2863

mirichmo: Agreed. Steve and I made the same review comment around the same time. This has been addressed.

src/System.Management.Automation/utils/PInvokeDllNames.cs
iSazonov: Is there MSDN public docs? I only see for Windows 10 and Windows 8.1. 😕

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
iSazonov: Closed.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -689,6 +689,114 @@ protected override void ProcessRecord()
...
+                    WriteNonTerminatingError(
+                        service,
+                        exception,
+                        "CouldNotGetServiceInfo",
+                        ServiceResources.CouldNotGetServiceInfo,
+                        ErrorCategory.PermissionDenied);
+                }
+
+                bool querySuccessful = NativeMethods.QueryServiceConfig2(hService, NativeMethods.SERVICE_CONFIG_DESCRIPTION, out descriptionStructPtr);
+                NativeMethods.SERVICE_DESCRIPTIONW description = (NativeMethods.SERVICE_DESCRIPTIONW)Marshal.PtrToStructure(descriptionStructPtr, typeof(NativeMethods.SERVICE_DESCRIPTIONW));
```
iSazonov: @joandrsn Could you please address @mirichmo previous request:

I'd like to see these method calls organized differently to protect against potential memory leaks. The pattern is repeated in NativeMethods.QueryServiceConfig2 and I'd like all instances of it fixed. Since NativeMethods.QueryServiceConfig() is a helper function, its output parameter, instead of being an IntPtr structurePointer, should be the managed object NativeMethods.SERVICE_DESCRIPTIONW. That way, the unmanaged memory that is allocated within that function is also cleaned up within that function. The managed object carries the values out of the function and eventually gets garbage collected. To fix this instance, line 741 should be moved within the helper function. You can probably make Nativemethods.QueryServiceConfig2() a generic method as well to avoid having to use a switch for infoLevel.

iSazonov (commented):
Leave a comment

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
joandrsn: This sounds like a great idea and I'm all for it.
Could you describe a bit more what you would change? 
Like somewhere with an example.
The problem I'm having now is if I return the `struct` as an `out`, I have to set the variable before returning. I could set default values, but that means I have to check the values when adding the properties to my `Service`-object.

joandrsn: @mirichmo, could you comment on my last comment?

iSazonov: The idea is to put (NativeMethods.QUERY_SERVICE_CONFIG)Marshal.PtrToStructure(serviceConfigStructPtr, typeof(NativeMethods.QUERY_SERVICE_CONFIG)); and Marshal.FreeCoTaskMem in QueryServiceConfig2 and return the managed object NativeMethods.SERVICE_DESCRIPTIONW

mirichmo: I had something like this in mind. The code is a bit messy and should be cleaned up, but illustrates the idea. The idea is that the helper function is a generic method that handles all the types in a similar manner. It also keeps allocation and freeing within the same method.

### Method

```C#
internal static bool QueryServiceConfig2<T>(NakedWin32Handle hService, DWORD infolevel, out T configStructure)
{
            IntPtr lpBuffer = IntPtr.Zero;
            structurePointer = IntPtr.Zero;
            DWORD bufferSize, bufferSizeNeeded = 0;

            bool status = NativeMethods.QueryServiceConfig2W(
                hService: hService,
                dwInfoLevel: infolevel,
                lpBuffer: lpBuffer,
                cbBufSize: 0,
                pcbBytesNeeded: out bufferSizeNeeded);

            if (status != true && Marshal.GetLastWin32Error() != ERROR_INSUFFICIENT_BUFFER) {
                return status;
            }

    try {
            lpBuffer = Marshal.AllocCoTaskMem((int)bufferSizeNeeded);
            bufferSize = bufferSizeNeeded;
           status = NativeMethods.QueryServiceConfig2W(
               hService,
               infolevel,
                lpBuffer,
                bufferSize,
                out bufferSizeNeeded);

        configStructure = (T)Marshal.PtrToStructure(lpBuffer, typeof(T));
    }
    finally
    {
        Marshal.FreeCoTaskMem(lpBuffer);
    }
            return status;
}
```

### Caller

```C#
NativeMethods.SERVICE_DESCRIPTIONW descriptionStruct = new NativeMethods.SERVICE_DESCRIPTIONW();

bool querySuccessful = NativeMethods.QueryServiceConfig2<NativeMethods.SERVICE_DESCRIPTIONW>(hService, NativeMethods.SERVICE_CONFIG_DESCRIPTION, out descriptionStruct);

// No free is needed here
```

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
iSazonov: @daxian-dbw Could you please comment?

daxian-dbw: I think there is still one benefit -- you can easily know what native dependencies powershell has.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
joandrsn: Thank you! 👍  I made the updates. Could you please review @mirichmo

iSazonov:
@mirichmo Could you please take another look?

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -1413,7 +1515,7 @@ public ServiceStartMode StartupType
         }
         // We set the initial value to an invalid value so that we can
         // distinguish when this is and is not set.
-        internal ServiceStartMode startupType = (ServiceStartMode)(-1);
+        internal ServiceStartupType startupType = (ServiceStartupType)(-1);
```
mirichmo: Sorry, I just noticed this. Your use of a wrapper enum + conversion functions makes it possible to define -1 as an ServiceStartupType value so it does not need to be casted throughout. Something like ServiceStartupType.InvalidValue or ServiceStartupType.Undefined would work.

mirichmo (changes requested):
Just one small comment. Other than that this looks good and it almost ready for merge.

joandrsn:
Thanks @mirichmo! I believe I fixed it.

iSazonov:
@joandrsn Please add `ServiceStartupType.InvalidValue` in the PR description.

joandrsn:
@iSazonov did I understand you correctly? I just updated my initial PR comment to reflect how `ServiceStartupType` can be used.

iSazonov:
LGTM.

iSazonov:
@SteveL-MSFT We've got new commits - could you please take another look?

mirichmo (approved):

iSazonov:
@joandrsn Many thanks for your contribution! Welcome back with new PRs!

jcoffi:
Whatever happened to this? It looks like it never made it into a final build as I'm not seeing it in 7.0.3.

iSazonov:
@jcoffi You can see all properties with:
```powershell
Get-Service wuauserv | fl * -Force
```

Issue #4910, opened 2017-09-24
Convert all files to ASCII encoding


SteveL-MSFT:
Based on discussion from https://github.com/PowerShell/PowerShell/pull/4761 we should ensure consistent line endings (LF on the repository side) and encodings (ASCII).  Since this touches many files, we'll do this after 6.0.0 is final and before work on 6.1.0 starts.

SteveL-MSFT:
Due to the change in .gitattributes to force auto eol correction, it was showing incorrect files as modified when doing a fetch/pull/clone.  So to avoid the confusion, I force changed all EOL to LF as part of https://github.com/PowerShell/PowerShell/pull/4943 and https://github.com/PowerShell/PowerShell/pull/4956

SteveL-MSFT:
I believe this is already done

PR #4912, opened 2017-09-24
Enable auto EOL on git repo side, fix some character encoding issues

SteveL-MSFT:
Some of the cleanup work from https://github.com/PowerShell/PowerShell/pull/4761

-  add ending newline to files missing it
- update .gitattributes to enforce autocrlf on all text files (this means some of the files show broader changes because it had CRLF on the repo side, but should just be LF)
- added missing <copyright> opening element
- fix mis-encoded character to apostrophe
- fix copyright symbol where it didn't show up correctly in vscode (although it shows up fine here, it's using the unicode symbol instead of the extended ascii one)

cc @mklement0 @markekraus


markekraus:
Just to state the obvious, but, the hashes in the tests will need to be updated for the files which were modified.

SteveL-MSFT:
Not sure why the FileCatalog test is failing on AppVeyor.  Passing on my WS2016 and WS2012R2 boxes.

test/powershell/engine/Cdxml/assets/CimTest/CdxmlTest.psd1
```diff
@@ -1,14 +1,14 @@
...
-    NestedModules = @( 'CimTest.cdxml')
-    AliasesToExport = @()
-    CmdletsToExport = @()
-    FunctionsToExport = @( 'Get-CimTest', 'Remove-CimTest', 'New-CimTest', 'Set-CimTest' )
-}
+@{
+    GUID = '41486F7D-842F-40F1-ACE4-8405F9C2ED9B'
+    Author="Microsoft Corporation"
+    CompanyName="Microsoft Corporation"
+    Copyright="(c) Microsoft Corporation. All rights reserved."
```
iSazonov: Why we use defferent "c"? I see:
-  Copyright (c) 
-  Copyright (C) 
-  Copyright (©)

markekraus: I vaguely recall `(c)` is completely wrong, `(C)` has no legal weight, and that `©` should be used unless the intended process or system cannot parse it (in which case fall back to `(C)`. But I can't find a source.

SteveL-MSFT: IANAL, but doing a search tells me the Berne Convention says none of those symbols are actually needed.  Let me follow up with our legal team to see what they recommend so we have consistency.  Based on prior experience, I don't expect them to get back to me very quickly, so I suggest we have this as a separate issue https://github.com/PowerShell/PowerShell/issues/4917

iSazonov: We can look CoreFX experience.

.gitattributes
```diff
@@ -1 +1,3 @@
 CHANGELOG.md merge=union
+* text=auto
```
TravisEz13: Based on my reading of the docs, this will only work if the files are not committed  CRLF.  

>When text is set to "auto", the path is marked for automatic end-of-line conversion. If Git decides that the content is text, its line endings are converted to LF on checkin. When the file has been committed with CRLF, **no conversion is done**.
```

Comment (SteveL-MSFT) [reply .gitattributes]: From https://www.git-scm.com/docs/gitattributes:

```none
If you want to ensure that text files that any contributor introduces to the repository have their line endings normalized, you can set the text attribute to "auto" for all files.

*	text=auto
```

The actual conversion of files that have CRLF (or worse mixed EOL) will be part of https://github.com/PowerShell/PowerShell/issues/4910 which we'll do after 6.0.0 final since it touches so many files.

TravisEz13 (approved):

TravisEz13:
@SteveL-MSFT Do you want to wait for the copyright issue to be resolved before merging?

SteveL-MSFT:
@TravisEz13 no, I created a separate issue for it https://github.com/PowerShell/PowerShell/issues/4917

daxian-dbw:
@SteveL-MSFT quick question, I noticed that some `©` has been changed to `(c)` and some are not (e.g. https://github.com/PowerShell/PowerShell/pull/4912/files?w=1#diff-fd6789ade5008bce347f09499cd96bf5R3). Is that intentional?

SteveL-MSFT:
@daxian-dbw I only changed the ones that were using the Unicode version that doesn't show up in VSCode (easy to search for).  I left the ones that use extended ascii.  So intentional.

PR #4914, opened 2017-09-25
Change UserAgent App WindowsPowerShell -> PowerShell

### LINKED ISSUE #4911: Default User Agent Reports WindowsPowerShell
https://github.com/PowerShell/PowerShell/blob/1c921cdf85e83c431a4888813302b0c241e384ce/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs#L116

This should be  just `PowerShell` and not `WindowsPowerShell`.

Steps to reproduce
------------------

```powershell
$bingdings = [System.Reflection.BindingFlags]::NonPublic -bxor [System.Reflection.BindingFlags]::Static
[Microsoft.PowerShell.Commands.PSUserAgent].GetProperty('App',$bingdings).GetValue($null,$null)
```

Expected behavior
-----------------

```none
PowerShell/6.0.0
```

Actual behavior
---------------

```none
WindowsPowerShell/6.0.0
```

Environment data
----------------

```powershell
> $PSVersionTable
```
```none
```
and
```none
OS                             Linux 4.4.0-96-generic #119-Ubuntu SMP Tue Se...

```

## Note on changing this
There are several tests, like the one below, which will also need to be updated.

https://github.com/PowerShell/PowerShell/blob/1c921cdf85e83c431a4888813302b0c241e384ce/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1#L466

markekraus:
fixes #4911

* renames the App portion of the user agent to `PowerShell`
* Adjusts tests.

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -465,7 +465,7 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
         # Validate response content
         $jsonContent = $result.Output.Content | ConvertFrom-Json
-        $jsonContent.headers.'User-Agent' | Should Be $DefaultUserAgent
+        $jsonContent.headers.'User-Agent' | Should Match '(?<!Windows)PowerShell\/\d+\.\d+\.\d+.*'
     }
 
     It "Invoke-WebRequest returns headers dictionary" {
@@ -1326,7 +1326,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
         $result = ExecuteWebCommand -command $command
 
         # Validate response
-        $result.Output.headers.'User-Agent' | Should Be $DefaultUserAgent
+        $result.Output.headers.'User-Agent' | Should Be '(?<!Windows)PowerShell\/\d+\.\d+\.\d+.*'
     }
 
     It "Invoke-RestMethod returns headers dictionary" {
@@ -2187,12 +2187,5 @@ Describe "Web cmdlets tests using the cmdlet's aliases" -Tags "CI" {
     It "Execute Invoke-RestMethod" {
         $result = irm "http://localhost:8082/PowerShell?test=response&output={%22hello%22:%22world%22}&contenttype=application/json" -TimeoutSec 5
         $result.Hello | Should Be "world"
-    }
-}
-
-Describe "PSUserAgent Tests" -Tags "CI" {
-    It "App Should Match ^PowerShell/\d+\.\d+\.\d+.*" {
-        $app = [Microsoft.PowerShell.Commands.PSUserAgent].GetProperty('App',@('Static','NonPublic')).GetValue($null,$null) 
-        $app | Should Match '^PowerShell/\d+\.\d+\.\d+.*'
     }
 }
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -463,7 +463,7 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
         # Validate response content
         $jsonContent = $result.Output.Content | ConvertFrom-Json
-        $jsonContent.headers.'User-Agent' | Should Match '(?<!Windows)PowerShell\/\d+\.\d+\.\d+.*'
+        $jsonContent.headers.'User-Agent' | Should MatchExactly '(?<!Windows)PowerShell\/\d+\.\d+\.\d+.*'
     }
 
     It "Invoke-WebRequest returns headers dictionary" {
@@ -1317,7 +1317,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
         $result = ExecuteWebCommand -command $command
 
         # Validate response
-        $result.Output.headers.'User-Agent' | Should Match '(?<!Windows)PowerShell\/\d+\.\d+\.\d+.*'
+        $result.Output.headers.'User-Agent' | Should MatchExactly '(?<!Windows)PowerShell\/\d+\.\d+\.\d+.*'
     }
 
     It "Invoke-RestMethod returns headers dictionary" {
@@ -1427,7 +1427,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
         $result = ExecuteWebCommand -command $command
 
         # Validate response
-        $result.Output.headers.'User-Agent' | Should Match '(?<!Windows)PowerShell\/\d+\.\d+\.\d+.*'
+        $result.Output.headers.'User-Agent' | Should MatchExactly '(?<!Windows)PowerShell\/\d+\.\d+\.\d+.*'
     }
 
     # Perform the following operation for Invoke-RestMethod
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -421,14 +421,13 @@ $PendingCertificateTest = $false
 if ( $IsMacOS ) { $PendingCertificateTest = $true }
 if ( test-path /etc/centos-release ) { $PendingCertificateTest = $true }
 
-# Get the default UserAgent through reflection
-$DefaultUserAgent = [Microsoft.PowerShell.Commands.PSUserAgent].GetProperty('UserAgent',@('Static','NonPublic')).GetValue($null,$null)
-
 Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
     BeforeAll {
         $response = Start-HttpListener -Port 8080
         $WebListener = Start-WebListener
+        # Get the default UserAgent through reflection
+        $DefaultUserAgent = [Microsoft.PowerShell.Commands.PSUserAgent].GetProperty('UserAgent',@('Static','NonPublic')).GetValue($null,$null)
     }
 
     AfterAll {
@@ -1310,6 +1309,8 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
     BeforeAll {
         $response = Start-HttpListener -Port 8081
         $WebListener = Start-WebListener
+        # Get the default UserAgent through reflection
+        $DefaultUserAgent = [Microsoft.PowerShell.Commands.PSUserAgent].GetProperty('UserAgent',@('Static','NonPublic')).GetValue($null,$null)
     }
 
     AfterAll {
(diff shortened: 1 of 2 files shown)


SteveL-MSFT (approved):
Thanks!

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -720,7 +720,7 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         $result = ExecuteRequestWithOutFile -cmdletName "Invoke-WebRequest" -uri $uri
         $jsonContent = $result.Output | ConvertFrom-Json
         $jsonContent.headers.Host | Should Be $uri.Authority
-        $jsonContent.headers.'User-Agent' | Should Match "WindowsPowerShell"
+        $jsonContent.headers.'User-Agent' | Should Match "PowerShell"
```
adityapatwardhan: This match would work with `WindowsPowerShell` as well. This should be `^PowerShell`. 

In addition to that, it should be `MatchExactly`

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1322,7 +1322,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
         $result = ExecuteWebCommand -command $command
 
         # Validate response
-        $result.Output.headers.'User-Agent' | Should Match "WindowsPowerShell"
+        $result.Output.headers.'User-Agent' | Should Match "PowerShell"
```
adityapatwardhan: Same as above.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1334,7 +1334,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
 
         # Validate response
         $result.Output.headers.Host | Should Be $Uri.Authority
-        $result.Output.headers.'User-Agent' | Should Match "WindowsPowerShell"
+        $result.Output.headers.'User-Agent' | Should Match "PowerShell"
```
adityapatwardhan: Same as above

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1347,7 +1347,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
 
         # Validate response
         $result.Output.headers.Host | Should Be $uri.Authority
-        $result.Output.headers.'User-Agent' | Should Match "WindowsPowerShell"
+        $result.Output.headers.'User-Agent' | Should Match "PowerShell"
```
adityapatwardhan: Same as above

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1360,7 +1360,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
 
         # Validate response
         $result.Output.headers.Host | Should Match $uri.Authority
-        $result.Output.headers.'User-Agent' | Should Match "WindowsPowerShell"
+        $result.Output.headers.'User-Agent' | Should Match "PowerShell"
```
adityapatwardhan: Same as above

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1435,7 +1435,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
         $result = ExecuteWebCommand -command $command
 
         # Validate response
-        $result.Output.headers.'User-Agent' | Should Match "WindowsPowerShell"
+        $result.Output.headers.'User-Agent' | Should Match "PowerShell"
```
adityapatwardhan: Same as above

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1497,7 +1497,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
 
                 # Validate response
                 $result.Output.url | Should Match $uri
-                $result.Output.headers.'User-Agent' | Should Match "WindowsPowerShell"
+                $result.Output.headers.'User-Agent' | Should Match "PowerShell"
```
adityapatwardhan: Same as above

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1525,7 +1525,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
 
         # Validate response
         $result.Output.url | Should Match $uri
-        $result.Output.headers.'User-Agent' | Should Match "WindowsPowerShell"
+        $result.Output.headers.'User-Agent' | Should Match "PowerShell"
```
adityapatwardhan: Same as above

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1560,7 +1560,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
         $result = ExecuteRequestWithOutFile -cmdletName "Invoke-RestMethod" -uri $uri
         $jsonContent = $result.Output | ConvertFrom-Json
         $jsonContent.headers.Host | Should Be $uri.Authority
-        $jsonContent.headers.'User-Agent' | Should Match "WindowsPowerShell"
+        $jsonContent.headers.'User-Agent' | Should Match "PowerShell"
```
adityapatwardhan: Same as above

adityapatwardhan (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: User-Agent can be `PowerShell/6.0.0-Beta.8` or `PowerShell/6.0.0`- so if we use  MatchExactly the pattern should be "^PowerShell/\d+\.\d+\.\d+.*"

adityapatwardhan: Makes sense.

markekraus: It's actually more like `Mozilla/5.0 (Windows NT; Microsoft Windows 10.0.15063 ; en-US) WindowsPowerShell/6.0.0` and it will be more dynamic "soon"

we could pull the exact one through reflection in `BeforeAll`. Honestly, I'm not even sure this is a good test. `Should Be $uri.Authority` should be sufficient, IMO.

markekraus:
Test fails due to #4919

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -421,6 +421,9 @@ $PendingCertificateTest = $false
 if ( $IsMacOS ) { $PendingCertificateTest = $true }
 if ( test-path /etc/centos-release ) { $PendingCertificateTest = $true }
 
+# Get the default UserAgent through reflection
+$DefaultUserAgent = [Microsoft.PowerShell.Commands.PSUserAgent].GetProperty('UserAgent',@('Static','NonPublic')).GetValue($null,$null)
```
iSazonov: It is dummy test: if C# code will be changed the test will be automatically passed.  I believe we should use explicit desired pattern.
Also please move to `BeforeAll`.

iSazonov (commented):
Leave a comment

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: That's not the purpose of the tests. They are just testing the response from the server to ensure that expected data sent is also returned. if we want to test the patterns in the UserAgent, that would be a separate set of tests for the `PSUserAgent` class.

I will add a separate describe block for `PSUserAgent` tests and a test for the `App` member (there will be other tests possibly for #4193). I will create a separate issue to add and track adding other tests for `PSUserAgent`.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -2188,3 +2189,10 @@ Describe "Web cmdlets tests using the cmdlet's aliases" -Tags "CI" {
         $result.Hello | Should Be "world"
     }
 }
+
+Describe "PSUserAgent Tests" {
+    It "App Should Match ^PowerShell/\d+\.\d+\.\d+.*" {
+        $app = [Microsoft.PowerShell.Commands.PSUserAgent].GetProperty('App',@('Static','NonPublic')).GetValue($null,$null) 
+        $app | Should Match '^PowerShell/\d+\.\d+\.\d+.*'
```
iSazonov: Here I don't see the point of checking internal property - we are free to change (add/remove) internals in any time. We should test public API. Here "public API" is User-Agent which we send to a server and get it back. We should test the User-Agent format if we want to control the User-Agent. So it is good to use regex pattern. I think that check User-Agent 14 times it's redundant - we need to have one check for each web cmdlet.

markekraus: Again, the point of most of those tests is not to test the User-Agent that is sent. Most of those assertions are being used to ensure meaningful data is returned in a response after a set of parameters has been used to make web requests. The only reason I updated them at all was because the change from `WindowsPowerShell` to `PowerShell` would have caused all these tests to false fail.

I will change the two tests that are explicitly about the User-Agent.

iSazonov: As you mentioned `$uri.Authority` is sufficient. So I'd remove `$DefaultUserAgent` at all. I unlike it as you see 😄

markekraus: I'm ok with that. This entire file needs to be more deeply analyzed and cleaned up. Maybe after I'm done making radical changes to it every day 😆.

iSazonov (approved):

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -463,7 +463,7 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
         # Validate response content
         $jsonContent = $result.Output.Content | ConvertFrom-Json
-        $jsonContent.headers.'User-Agent' | Should Match "WindowsPowerShell"
+        $jsonContent.headers.'User-Agent' | Should Match '(?<!Windows)PowerShell\/\d+\.\d+\.\d+.*'
```
JamesWTruher: I think you might use `MatchExactly` which does a case sensitive match (`-cmatch`). Especially since you're explicitly not matching WindowsPowerShell

JamesWTruher (approved):

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1

adityapatwardhan (approved):
LGTM

PR #4921, opened 2017-09-26
Enable network Select-String test

### LINKED ISSUE #4827: Enable Select-String test with network path
We have the skipped [Select-String test](https://github.com/PowerShell/PowerShell/blob/02b5f357a20e6dee9f8e60e3adb9025be3c94490/test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1#L39)
```powershell
        It "Network path" -skip:($IsCoreCLR) {
            (select-string -LiteralPath $fileNameAsNetworkPath "b").count | Should Be 2
        }
```

Should we enable the test or remove?

If enable we should:
1. ~Run it with admin privileges~
2. Fix #4828  `New-TemporaryFile` - it don't return extended provider properties like PSDrive, PSPath...

iSazonov:
Fix #4827 

Enable pending network Select-String test.

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
@@ -14,7 +14,7 @@
             $pathWithoutWildcard = $TestDrive
             $pathWithWildcard = Join-Path $TestDrive '*'
 
-            # Here Get-ChildItem add 'PSDrive' property
+            # Here Get-ChildItem adds 'PSDrive' property
             $tempFile = New-TemporaryFile | Get-ChildItem
             "abc" | Out-File -LiteralPath $tempFile.fullname
 	        "bcd" | Out-File -LiteralPath $tempFile.fullname -Append
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
@@ -14,7 +14,8 @@
             $pathWithoutWildcard = $TestDrive
             $pathWithWildcard = Join-Path $TestDrive '*'
 
-            $tempFile = New-TemporaryFile
+            # Here Get-ChildItem add 'PSDrive' property
+            $tempFile = New-TemporaryFile | Get-ChildItem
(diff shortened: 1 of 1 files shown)


test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
```diff
@@ -14,7 +14,8 @@
             $pathWithoutWildcard = $TestDrive
             $pathWithWildcard = Join-Path $TestDrive '*'
 
-            $tempFile = New-TemporaryFile
+            # Here Get-ChildItem add 'PSDrive' property
```
markekraus: I had trouble understanding this comment. Can you correct `add` to `adds`?

markekraus (approved):
Minor comment suggestion, otherwise LGTM

test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
iSazonov: Fixed.

markekraus (approved):

test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
```diff
@@ -14,7 +14,8 @@
             $pathWithoutWildcard = $TestDrive
             $pathWithWildcard = Join-Path $TestDrive '*'
 
-            $tempFile = New-TemporaryFile
+            # Here Get-ChildItem adds 'PSDrive' property
+            $tempFile = New-TemporaryFile | Get-ChildItem
```
SteveL-MSFT: Although in this case, it'll have the same result, I think

```powershell
New-TemporaryFile | Get-Item
```

is more correct, particularly if other people use this pattern until we address `New-TemporaryFile`

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/string.tests.ps1
iSazonov: Fixed.

SteveL-MSFT (approved):

PR #4922, opened 2017-09-26
Exclude -Comobject parameter of New-Object cmdlet on unsupported platforms

### LINKED ISSUE #4897: -ComObject should not be exposed on systems where it is not supported
Hello everybody, i am encountering an issue.
I need to convert some old XML files (encoded with word 2006 .xml schema) as PDF.
I already used PowerShell and ComObject to do this on a Windows system :

```powershell
$word = New-Object -ComObject Word.Application
$word.Documents.Open($filename).SaveAs($filename.replace(".xml", ".pdf"), 17)
```

but now I have to automate this process on a Linux system (every XML files are generated onto).
PowerShell on Linux able to me to use these commands but not ComObject due to a missing of api-ms-win-core-com-l1-1-0.dll

Is there any way to load this dll into a Linux system ? If not, is there any other way to do the same ?

Command
-----------------

```powershell
$word = New-Object -ComObject Word.Application
```

Expected behavior
-----------------

```none

```

Actual behavior
-----------------

```powershell
New-Object : Unable to load DLL 'api-ms-win-core-com-l1-1-0.dll': The specified module or one of its dependencies could not be found.
 (Exception from HRESULT: 0x8007007E)
At line:1 char:9
+ $word = New-Object -ComObject Word.application
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [New-Object], DllNotFoundException
    + FullyQualifiedErrorId : System.DllNotFoundException,Microsoft.PowerShell.Commands.NewObjectCommand
```

Environment data
-----------------

```powershell
> $PSVersionTable
OS                             Linux 4.4.0-38-generic #57-Ubuntu SMP Tue Sep 6 15:42:33 UTC 2016
```

iSazonov:
Fix #4897.

New-Object -Comobject is not supported on
- Linux and MacOs - fix is remove the code.
- Windows Core and IoT - fix is throw `NotSupportedException` exception.

First commit reformat test file.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/new-object.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/new-object.cs
@@ -31,12 +31,12 @@ namespace Microsoft.PowerShell.Commands
         [Parameter(ParameterSetName = netSetName, Mandatory = true, Position = 0)]
         public string TypeName { get; set; } = null;
 
-
+#if !UNIX
         private Guid _comObjectClsId = Guid.Empty;
         /// <summary> the ProgID of the Com object</summary>
         [Parameter(ParameterSetName = "Com", Mandatory = true, Position = 0)]
         public string ComObject { get; set; } = null;
-
+#endif
 
         /// <summary>
         /// The parameters for the constructor
@@ -245,8 +245,15 @@ namespace Microsoft.PowerShell.Commands
                      "CannotFindAppropriateCtor",
                      ErrorCategory.ObjectNotFound, null));
             }
+#if !UNIX
             else // Parameterset -Com
             {
+                if (!Platform.IsWindowsDesktop)
+                {
+                    Exception exc = new NotSupportedException(NewObjectStrings.ComObjectPlatformIsNotSupported);
+                    ThrowTerminatingError(new ErrorRecord(exc, "ComObjectPlatformIsNotSupported", ErrorCategory.NotImplemented, null));
+                }
+
                 int result = NewObjectNativeMethods.CLSIDFromProgID(ComObject, out _comObjectClsId);
 
                 // If we're in ConstrainedLanguage, do additional restrictions
@@ -295,6 +302,7 @@ namespace Microsoft.PowerShell.Commands
                 }
                 WriteObject(comObject);
             }
+#endif
         }//protected override void BeginProcessing()
 
         #endregion Overrides
@@ -407,6 +415,7 @@ namespace Microsoft.PowerShell.Commands
         }
 #endif
 
+#if !UNIX
         private object CreateComObject()
         {
             Type type = null;
@@ -465,7 +474,7 @@ namespace Microsoft.PowerShell.Commands
                 return null;
             }
         }
-
+#endif
         #endregion Com
 
         // HResult code '-2147417850' - Cannot change thread mode after it is set.
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/new-object.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/new-object.cs
@@ -248,12 +248,6 @@ namespace Microsoft.PowerShell.Commands
 #if !UNIX
             else // Parameterset -Com
             {
-                if (!Platform.IsWindowsDesktop)
-                {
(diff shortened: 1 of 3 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/new-object.cs
```diff
@@ -245,8 +245,15 @@ protected override void BeginProcessing()
                      "CannotFindAppropriateCtor",
                      ErrorCategory.ObjectNotFound, null));
             }
+#if !UNIX
             else // Parameterset -Com
             {
+                if (!Platform.IsWindowsDesktop)
+                {
+                    Exception exc = new NotSupportedException(NewObjectStrings.ComObjectPlatformIsNotSupported);
```
daxian-dbw: COM works on NanoServer as well, as long as there it has a workable COM component
PS: the COM support was enabled when powershell core was still NanoServer powershell.

daxian-dbw (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/new-object.cs
```diff
@@ -465,7 +474,7 @@ private object CreateComObject()
                 return null;
             }
         }
-
+#endif
         #endregion Com
```
PaulHigin: I think the entire "com" region here should be under the #if !UNIX.

PaulHigin (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/new-object.cs
iSazonov: So should we exclude the parameter only on Unix?

daxian-dbw: The COM components we use in our tests are not available in NanoServer and IoT.
Yes, we should exclude the parameter only for Unix plats.

iSazonov: Does "Shell.Application" work on NanoServer and IoT? can I use this in a test?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/new-object.cs
iSazonov: Fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/new-object.cs
daxian-dbw: Shell itself is not available on NanoServer and IoT. You can use it as a test, but just follow the existing tests in `test\powershell\engine\COM`.

PaulHigin (approved):
LGTM

src/Microsoft.PowerShell.Commands.Utility/commands/utility/new-object.cs
iSazonov: Why shouldn't we targeting our test to NanoServer and IoT?

test/powershell/Modules/Microsoft.PowerShell.Utility/New-Object.Tests.ps1
```diff
@@ -1,203 +1,211 @@
 Describe "New-Object" -Tags "CI" {
+    It "Support 'ComObject' parameter on platforms" {
+        if ($IsLinux -or $IsMacOs ) {
+            { New-Object -ComObject "Shell.Application" } | ShouldBeErrorId "NamedParameterNotFound,Microsoft.PowerShell.Commands.NewObjectCommand"
+        } else {
+            { New-Object -ComObject "Shell.Application" } | Should Not Throw
```
daxian-dbw: This will fail when running tests on NanoServer ... Maybe change it to check if `New-Object` has the parameter `ComObject` instead of creating the `Shell.Application` object?
```
$s = gcm New-Object
$s.Parameters.ContainsKey("ComObject")
True
```

daxian-dbw (changes requested):

daxian-dbw (approved):

PR #4926, opened 2017-09-26
Add Tests for *-computer cmdlets

JamesWTruher:
This PR also removes code protected by #!CORE in Computer.cs which removed a very large amount of code.

In order to test without actually rebooting the system, I implemented a number of InternalTestHooks to avoid making the call to actually calling the CIM method to actually reboot/halt the system. While I was in the code I simplified a number of parameters, especially throttling, which was way more complicated than it needed to be.

the breaking change is specific to parameter values which were silently "corrected" for throttling (which FWIW isn't even implemented)

Fix https://github.com/PowerShell/PowerShell/issues/4146


src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs
```diff
@@ -198,18 +183,13 @@ public class TestConnectionCommand : PSCmdlet
         /// </summary>
         [Parameter(ParameterSetName = SourceParameterSet)]
         [Parameter(ParameterSetName = RegularParameterSet)]
-        [ValidateRange(int.MinValue, (int)1000)]
+        [ValidateRange((int)1, (int)1000)]
```
SteveL-MSFT: The current [documentation](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/restart-computer?view=powershell-5.1) says using 0 == 32.  I think you should support that to avoid a breaking change.  We can probably safely not support negative values.

src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs
```diff
@@ -244,13 +224,6 @@ public SwitchParameter Quiet
...
 #region "Overrides"
 
-#if !CORECLR
-        ///// <summary>
-        ///// To Store the output for each ping reply
-        ///// </summary>
-        private ManagementObjectSearcher searcher;
-
-#endif
         private TransportProtocol _transportProtocol = TransportProtocol.DCOM;
```
SteveL-MSFT: This should be TransportProtocol.WSMAN as DCOM isn't supported

dantraMSFT:
>                 throttlelimit = value;

Shouldn't this logic remain for cases where the cmdlet is called programmatically?

---
Refers to: src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs:207 in e283582. [](commit_id = e28358202134177049fe78beba58357ff878b888, deletion_comment = True)

dantraMSFT:
>         [ValidateSet(ComputerWMIHelper.DcomProtocol, ComputerWMIHelper.WsmanProtocol)]

The ValidateSet should be updated as well.

---
Refers to: src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs:102 in e283582. [](commit_id = e28358202134177049fe78beba58357ff878b888, deletion_comment = False)

src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs
```diff
@@ -672,767 +486,438 @@ private void ProcessWSManProtocolForTestConnection()
...
+        /// Packet = 4,
+        /// PacketIntegrity = 5,
+        /// PacketPrivacy = 6.
         /// </summary>
-        /// <param name="disposing"></param>
-        public void Dispose(bool disposing)
+        [Parameter]
+        [Alias("Authentication")]
+        [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly")]
+        public AuthenticationLevel DcomAuthentication
```
SteveL-MSFT: We should remove this since DcomAuthentication isn't supported.  Will need documentation update.

dantraMSFT:
>         [ValidateSet(ComputerWMIHelper.DcomProtocol, ComputerWMIHelper.WsmanProtocol)]

I also think that validation should occur here, since it is removed from BeginProcessing.

---
In reply to: [332362109](https://github.com/PowerShell/PowerShell/pull/4926#issuecomment-332362109) [](ancestors = 332362109)

---
Refers to: src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs:102 in e283582. [](commit_id = e28358202134177049fe78beba58357ff878b888, deletion_comment = False)

src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs
```diff
@@ -672,767 +486,438 @@ private void ProcessWSManProtocolForTestConnection()
...
-                }
+                _dcomAuthentication = value;
+                _isDcomAuthenticationSpecified = true;
             }
         }
-
-#endregion "IDisposable Members"
-
-#region Overrides
+        private AuthenticationLevel _dcomAuthentication = AuthenticationLevel.Packet;
```
SteveL-MSFT: remove, not supported

src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs
```diff
@@ -672,767 +486,438 @@ private void ProcessWSManProtocolForTestConnection()
...
+        private bool _isDcomAuthenticationSpecified = false;
```
SteveL-MSFT: remove, not supported

dantraMSFT:
>             switch (_transportProtocol)

Is the switch statement unnecessary? The transport can only be one value.

---
Refers to: src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs:283 in e283582. [](commit_id = e28358202134177049fe78beba58357ff878b888, deletion_comment = False)

src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs
```diff
@@ -672,767 +486,438 @@ private void ProcessWSManProtocolForTestConnection()
...
+        ///
+        /// Default = 0,
+        /// Anonymous = 1,
+        /// Identify = 2,
+        /// Impersonate = 3,
+        /// Delegate = 4.
         /// </summary>
-        protected override void BeginProcessing()
+        [Parameter]
+        public ImpersonationLevel Impersonation
```
SteveL-MSFT: remove, not supported (I believe this only applies to DCOM)

src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs
```diff
@@ -672,767 +486,438 @@ private void ProcessWSManProtocolForTestConnection()
...
-                            ErrorRecord errorRecord = new ErrorRecord(e, "COMException", ErrorCategory.InvalidOperation, null);
-                            WriteError(errorRecord);
-                        }
-                    }
-                }
+                _impersonation = value;
+                _isImpersonationSpecified = true;
             }
         }
+        private ImpersonationLevel _impersonation = ImpersonationLevel.Impersonate;
```
SteveL-MSFT: remove, not supported

src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs
```diff
@@ -672,767 +486,438 @@ private void ProcessWSManProtocolForTestConnection()
...
+        private bool _isImpersonationSpecified = false;
```
SteveL-MSFT: remove, not supported

src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs
```diff
@@ -672,767 +486,438 @@ private void ProcessWSManProtocolForTestConnection()
...
+            "Digest",
+            "Kerberos")] // can be used with and without credential (not sure about implications)
+        [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly")]
+        public string WsmanAuthentication { get; set; }
+
+        /// <summary>
+        /// Specify the protocol to use
+        /// </summary>
+        [Parameter(ParameterSetName = DefaultParameterSet)]
+        [ValidateSet(ComputerWMIHelper.DcomProtocol, ComputerWMIHelper.WsmanProtocol)]
```
SteveL-MSFT: DCOM not supported.  We should keep this if we decide to support SSH directly.

dantraMSFT:
>         [ValidateSet(ComputerWMIHelper.DcomProtocol, ComputerWMIHelper.WsmanProtocol)]

The same comments as above; the ValidateSet should be updated and validation inserted to ensure only WsmanProtocol

---
Refers to: src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs:713 in e283582. [](commit_id = e28358202134177049fe78beba58357ff878b888, deletion_comment = False)

src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs
```diff
@@ -1455,4822 +940,1464 @@ public void Dispose(bool disposing)
...
-                Protocol.Equals(ComputerWMIHelper.DcomProtocol, StringComparison.OrdinalIgnoreCase))
-            {
-                InvalidOperationException ex = new InvalidOperationException(ComputerResources.InvalidParameterDCOMNotSupported);
-                ThrowTerminatingError(new ErrorRecord(ex, "InvalidParameterDCOMNotSupported", ErrorCategory.InvalidOperation, null));
-            }
-
-            // TODO:CORECLR This should be re-visited if we decide to add double hop remoting to CoreCLR (outgoing connections)
-            if (ParameterSetName.Equals(AsJobParameterSet, StringComparison.OrdinalIgnoreCase))
-            {
-                InvalidOperationException ex = new InvalidOperationException(ComputerResources.InvalidParame
```
SteveL-MSFT: remove, not supported

SteveL-MSFT: remove, not supported

SteveL-MSFT: remove, not supported

SteveL-MSFT: shouldn't be needed anymore

SteveL-MSFT: no longer needed

SteveL-MSFT: no longer needed

SteveL-MSFT: no longer needed

SteveL-MSFT: WSMAN

SteveL-MSFT: remove, not supported

SteveL-MSFT: remove DCOM.  Perhaps we should just remove Protocol as a paramter until we support something other than WSMan (applies to other cmdlets as well)

SteveL-MSFT: remove, only applies to DCOM

SteveL-MSFT: remove, not applicable

SteveL-MSFT: remove, not applicable

SteveL-MSFT: remove, not applicable

SteveL-MSFT: remove, not applicable

SteveL-MSFT: remove, not applicable

SteveL-MSFT: remove, not applicable

SteveL-MSFT: remove, not applicable

SteveL-MSFT: WSMAN

SteveL-MSFT: should just be WSMan, but this is probably no longer needed

SteveL-MSFT: remove, not applicable

SteveL-MSFT: remove, not applicable

test/powershell/Modules/Microsoft.PowerShell.Management/Rename-Computer.Tests.ps1
```diff
@@ -0,0 +1,97 @@
...
+
+# protect the user from from stop their computer 
+function Test-TesthookIsSet() {
+    try {
+        return ${TesthooksType}.GetField($TesthookName, "NonPublic,Static").GetValue($null)
+    }
+    catch {
+        # fall through
+    }
+    return $false
```
SteveL-MSFT: why not have this in the catch?

test/powershell/Modules/Microsoft.PowerShell.Management/Rename-Computer.Tests.ps1
```diff
@@ -0,0 +1,97 @@
...
+}
+
+try 
+{
+    # set up for testing
+    $PSDefaultParameterValues["it:skip"] = ! $IsWindows
+    ${TesthooksType}::SetTestHook($TesthookName,$true)
```
SteveL-MSFT: inconsistent whitespace with line 28, probably easier to read to include it

test/powershell/Modules/Microsoft.PowerShell.Management/Rename-Computer.Tests.ps1
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        It "Should not produce a reboot warning when renaming a computer with the reboot flag" {
+            Set-HookResult -result $defaultResultValue
+            $newname = "mynewname"
+            $result = Rename-Computer -ea Stop -ComputerName . -NewName "$newname" -Pass -WarningAction SilentlyContinue -WarningVariable WarnVar -Restart
+            $result.HasSucceeded | should be $true
+            $result.NewComputerName | should be $newname
+            $WarnVar | should BeNullOrEmpty
+
```
SteveL-MSFT: remove whitespace

test/powershell/Modules/Microsoft.PowerShell.Management/Rename-Computer.Tests.ps1
```diff
@@ -0,0 +1,97 @@
...
+            Set-HookResult -result $defaultResultValue
+            $newname = "mynewname"
+            $result = Rename-Computer -ea Stop -ComputerName . -NewName "$newname" -Pass -WarningAction SilentlyContinue -WarningVariable WarnVar -Restart
+            $result.HasSucceeded | should be $true
+            $result.NewComputerName | should be $newname
+            $WarnVar | should BeNullOrEmpty
+            
+        }
+
+
```
SteveL-MSFT: extra whitespace

SteveL-MSFT (changes requested):
Don't bother responding to my comments unless you disagree

src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs
JamesWTruher: yah - specifying a throttle of 0 which means 32 doesn't make any sense to me at all. The default is 32, which is fine (FWIW, throttling doesn't seem to be actually implemented). We have the opportunity to fix this somewhat strange behavior, by specifying a range which is actually useful.

test/powershell/Modules/Microsoft.PowerShell.Management/Rename-Computer.Tests.ps1
JamesWTruher: cause I need to return false from here anyway - no reason to have
```
catch {
    return $false
}
return $false
```

JamesWTruher:
I hadn't intended a complete overhaul, but that's what it looks like is needed. I will update accordingly

src/Microsoft.PowerShell.Commands.Management/commands/management/Computer.cs
SteveL-MSFT: I'm ok with updating the documentation and making this breaking change

SteveL-MSFT:
Added issue this fixes

PR #4927, opened 2017-09-26
Fix credential scan issues

TravisEz13:
Add suppression for lines with secrets.
Add suppressions for files which are secrets.
This will be consumed in the VSTS daily build.

Renamed tests with `ConvertTo-SecureString` to avoid false positives

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-SecureString.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-SecureString.Tests.ps1
@@ -1,8 +1,9 @@
-Describe "ConvertTo-SecureString" -Tags "CI" {
+	Describe "ConvertTo--SecureString" -Tags "CI" {
 
-    Context "Checking return types of ConvertTo-SecureString" {
+    Context "Checking return types of ConvertTo--SecureString" {
 
 	It "Should return System.Security.SecureString after converting plaintext variable"{
+        #[SuppressMessage("Microsoft.Security", "CS002:SecretInNextLine", Justification="Demo/doc/test secret.")]
 	    $PesterTestConvert = (ConvertTo-SecureString "plaintextpester" -AsPlainText -force)
 	    $PesterTestConvert | Should BeOfType securestring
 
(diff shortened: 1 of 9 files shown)


SteveL-MSFT (approved):
LGTM

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-SecureString.Tests.ps1
```diff
@@ -1,8 +1,9 @@
-Describe "ConvertTo-SecureString" -Tags "CI" {
+	Describe "ConvertTo--SecureString" -Tags "CI" {
```
adityapatwardhan: Seems to have an extra - between ConvertTo and SecureString

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-SecureString.Tests.ps1
```diff
@@ -1,8 +1,9 @@
...
 
-    Context "Checking return types of ConvertTo-SecureString" {
+    Context "Checking return types of ConvertTo--SecureString" {
```
adityapatwardhan: Seems to have an extra - between ConvertTo and SecureString

adityapatwardhan (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-SecureString.Tests.ps1
TravisEz13: See the previous comment

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-SecureString.Tests.ps1
TravisEz13: I have to rename it to something other than convertto-securestring.

adityapatwardhan (approved):

megamorf:
Just out of curiosity, is the `CS002:SecretInNextLine` suppression VSTS specific? Haven't seen that before.

TravisEz13:
@megamorf It's specific to the tool we use.  Unfortunately, the tool itself is internal to Microsoft only.  That is why we run it in VSTS and not are normal CI systems.

megamorf:
@TravisEz13 Ah, that makes sense. Thanks for the answer :-)

Issue #4836, opened 2017-09-14
Enumerate and run installed PowerShell versions


iSazonov:
Currently we changed semantic of `-version` command line parameter from "run Windows PowerShell 2.0" to "show current PowerShell version".

Now we need a way to address a side-by-side scenario - to enumerate installed PowerShell versions and run needed PowerShell version.

### Suggesion
[ ] Enumerate installed versions:
```
>powershell -versions
```

[ ] Run needed version:
```
powershell -versions <ver spec>
```

iSazonov:
I am ready to implement enumeration of installed versions based on https://github.com/PowerShell/vscode-powershell/blob/c3e987c9e9add273e6699972bb017ccfbc47e873/src/platform.ts#L110

joeyaiello:
I totally understand the need for this, but I think that it probably shouldn't be implemented in the host binary. Ultimately, this is the job of the package manager, but I get why we might need something like [virtualenv](https://virtualenv.pypa.io/en/stable/) as well.

However, at this point in the release cycle, it's a pretty high-risk change to bring in (and potentially get wrong), especially when you consider that there won't be any production scenarios for side-by-side 6.x's for at least a little while. 

When we do want to bring it in, we should have an RFC for it (and I think we should have some alternate designs that include a PS module that enumerates versions as well as whether it should be a native tool to support bootstrap scenarios).

@iSazonov do you have a scenario today where this is critical? I think unpacking that is the first step to understanding the proper implementation here (and where having a `pwsh` alias for PS Core isn't enough).

iSazonov:
@joeyaiello I see that VS Code (see link above) discover installed PowerShell versions and allow users run desired version.
Also I know Microsoft SCOR 2012 product. It is based on PowerShell 2.0 (!!!). If users need new features they are forced to run needed version explicitly.
I believe it is better to have the discover feature natively (specially on Windows) - this will make other products more stable.
I'm not sure about launching, but it's also interesting, all the more so when we had `-Version 2`.

iSazonov:
Any news?

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and believes this is should be an RFC for a proposal that can be reviewed in its entirety with considerations for different potential design proposals

KirkMunro:
RFC: https://github.com/PowerShell/PowerShell-RFC/pull/202.

PR #4931, opened 2017-09-27
Write an error for powershell -version <any value>

### LINKED ISSUE #4834: powershell -version 2 should return error
As far as I know, PowerShell v2 was recently declared deprecated.
Thus, it normal to expect some kind of related errors in the latest v6.

The behaviour is strange, though. If I try to invoke something (anything
actually, including gibberish) with `-version 2` then PowerShell succeeds
(`$LASTEXITCODE` is 0) and prints the same result `powershell v6.0.0-beta.7`.

Steps to reproduce
------------------

````powershell
    powershell -version 2 gibberish
    $LASTEXITCODE
````

Expected behavior
-----------------

Non-zero `$LASTEXITCODE` and a clear error message.

Actual behavior
---------------

Zero `$LASTEXITCODE` and some version-like text, the same regardless of a command.

````
    powershell v6.0.0-beta.7
    0
````

Environment data
----------------

PowerShell-6.0.0-beta.7-win-x64.zip

````
OS                             Microsoft Windows 10.0.10586
````

### PR DESCRIPTION
**Pull request by iSazonov**
Fix #4834

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
@@ -552,7 +552,7 @@ namespace Microsoft.PowerShell
                     ++i;
                     if (i < args.Length)
                     {
-                        if (LanguagePrimitives.TryConvertTo<int>(args[i], out int verNumber) && (verNumber == 2 || verNumber == 3)
+                        if (LanguagePrimitives.TryConvertTo<int>(args[i], out int verNumber) && (verNumber == 2 || verNumber == 3))
                         {
                             WriteCommandLineError(CommandLineParameterParserStrings.DeprecatedVersionParameter);
                         }
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
@@ -552,10 +552,7 @@ namespace Microsoft.PowerShell
                     ++i;
                     if (i < args.Length)
                     {
-                        if (LanguagePrimitives.TryConvertTo<int>(args[i], out int verNumber) && (verNumber == 2 || verNumber == 3))
-                        {
-                            WriteCommandLineError(CommandLineParameterParserStrings.DeprecatedVersionParameter);
-                        }
+                        WriteCommandLineError(string.Format(CultureInfo.CurrentCulture, CommandLineParameterParserStrings.DeprecatedVersionParameter,args[i]));
                     }
                     break;
                 }
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
@@ -548,6 +548,15 @@ namespace Microsoft.PowerShell
                     _noInteractive = true;
                     _skipUserInit = true;
                     _noExit = false;
+
+                    ++i;
+                    if (i < args.Length)
+                    {
+                        if (LanguagePrimitives.TryConvertTo<int>(args[i], out int verNumber) && verNumber == 2)
+                        {
+                            WriteCommandLineError(CommandLineParameterParserStrings.DeprecatedVersionParameter);
+                        }
+                    }
                     break;
                 }
 
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
@@ -552,7 +552,7 @@ namespace Microsoft.PowerShell
                     ++i;
                     if (i < args.Length)
                     {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx [hunk 1]
```diff
@@ -272,4 +272,7 @@ EXAMPLES
   <data name="InvalidArgument" xml:space="preserve">
     <value>Invalid argument '{0}', did you mean:</value>
   </data>
+  <data name="DeprecatedVersionParameter" xml:space="preserve">
+    <value>Old functionality of '-version 2.0' is deprecated. Now '-version' parameter shows current PowerShell version. </value>
```
SteveL-MSFT: I think we should change this error message to:

```none
Previous usage of '-version {0}' is not supported.  '-version' currently only supports returning the current PowerShell version.
```

cc @joeyaiello

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -548,6 +548,15 @@ private void ParseHelper(string[] args)
                     _noInteractive = true;
                     _skipUserInit = true;
                     _noExit = false;
+
+                    ++i;
+                    if (i < args.Length)
+                    {
+                        if (LanguagePrimitives.TryConvertTo<int>(args[i], out int verNumber) && verNumber == 2)
```
SteveL-MSFT: I think we should currently error if any value is passed to `-v`

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
iSazonov: Now we write the error and then show current version.
Please clarify your request.

SteveL-MSFT: Aren't you only writing error if `-v 2`?  Should also error if `-v 3`

iSazonov: I agree. Will add the error for `-v 3` too.

SteveL-MSFT: We should error on any value passed to `-v`.  In some future time if we decide to support launching different installed versions of PSCore6, we can decide if we want to overload `-v` or even introduce a new parameter.

iSazonov: Currently we ignore all values after `-v`. The docs says the same: accept only 2.0 or 3.0. If anybody use the parameter the breaking change is related only to 2.0 or 3.0 values - we should write an error for the cases only.

src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
iSazonov: I changed the error message - please review.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
iSazonov: Could you comment #4836 - I am ready to implement enumeration of installed versioms.

anmenaga: If we currently ignore all values after -v, then it seems logical to write error on any value passed to -v saying that parameter doesn't have any effect.
Having a parameter that quietly does nothing doesn't look good.

anmenaga (changes requested):
Please add tests.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -548,6 +548,15 @@ private void ParseHelper(string[] args)
                     _noInteractive = true;
                     _skipUserInit = true;
                     _noExit = false;
+
+                    ++i;
+                    if (i < args.Length)
+                    {
+                        if (LanguagePrimitives.TryConvertTo<int>(args[i], out int verNumber) && (verNumber == 2 || verNumber == 3))
```
SteveL-MSFT: I think we should error on any value passed to `-v` as part of this PR.

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
iSazonov: It is an enhancement. It make sense. Thanks!

Done.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
iSazonov: Done.

SteveL-MSFT (approved):

anmenaga (approved):

PR #4933, opened 2017-09-27
Enable security flags in native compiler

TravisEz13:
Enable strong stack protection
Enable  address space layout randomization (ASLR) or PIE
https://en.wikipedia.org/wiki/Position-independent_code#Position-independent_executables


SteveL-MSFT (commented):
Do we need similar flags for arm?

TravisEz13:
Seems like a good idea.  I enabled it and verified the build still works.

daxian-dbw (approved):

PR #4937, opened 2017-09-28
User Agent now reports the OS platform

### LINKED ISSUE #4913: User Agent Statically Reports "Windows NT" as the platform.
https://github.com/PowerShell/PowerShell/blob/1c921cdf85e83c431a4888813302b0c241e384ce/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs#L125

Steps to reproduce
------------------

```powershell
$bingdings = [System.Reflection.BindingFlags]::NonPublic -bxor [System.Reflection.BindingFlags]::Static
[Microsoft.PowerShell.Commands.PSUserAgent].GetProperty('Platform',$bingdings).GetValue($null,$null)
```

Expected behavior
-----------------

This should change based on the platform 

Examples

| OS | Platform |
|--|--|
| Windows | Windows NT |
| macOS | Macintosh |
| Linux | Linux |

Actual behavior
---------------

```none
Windows NT
```

Environment data
----------------

```powershell
> $PSVersionTable
```

```none
OS                             Linux 4.4.0-96-generic #119-Ubuntu SMP Tue Sep 12 14:59:54 UTC 2017
```

LDSpits:
This pr fixes #4913, I have not written any tests, If these are required then please notify me of this and Ill add them

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
@@ -125,7 +125,10 @@ namespace Microsoft.PowerShell.Commands
             {
                 if (Platform.IsWindows)
                 {
-                    return "Windows NT";
+                    // find the version in the windows operating system description
+                    string versionText = PSUserAgent.OS.Substring(PSUserAgent.OS.TrimEnd().LastIndexOf(" ") +1);
+                    Version windowsPlatformversion = new Version(versionText);
+                    return $"Windows NT {windowsPlatformversion.Major}.{windowsPlatformversion.Minor}";
                 }
                 else if (Platform.IsMacOS)
                 {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
@@ -22,7 +22,7 @@ namespace Microsoft.PowerShell.Commands
                 // format the user-agent string from the various component parts
                 string userAgent = string.Format(CultureInfo.InvariantCulture,
                     "{0} ({1}; {2}; {3}) {4}",
-                    Compatibility, Platform, OS, Culture, App);
+                    Compatibility, PlatformName, OS, Culture, App);
                 return (userAgent);
             }
         }
@@ -37,7 +37,7 @@ namespace Microsoft.PowerShell.Commands
                 // format the user-agent string from the various component parts
                 string userAgent = string.Format(CultureInfo.InvariantCulture,
                     "{0} (compatible; MSIE 9.0; {1}; {2}; {3})",
-                    Compatibility, Platform, OS, Culture);
+                    Compatibility, PlatformName, OS, Culture);
                 return (userAgent);
             }
         }
@@ -52,7 +52,7 @@ namespace Microsoft.PowerShell.Commands
                 // format the user-agent string from the various component parts
                 string userAgent = string.Format(CultureInfo.InvariantCulture,
                     "{0} ({1}; {2}; {3}) Gecko/20100401 Firefox/4.0",
-                    Compatibility, Platform, OS, Culture);
+                    Compatibility, PlatformName, OS, Culture);
                 return (userAgent);
             }
         }
@@ -67,7 +67,7 @@ namespace Microsoft.PowerShell.Commands
                 // format the user-agent string from the various component parts
                 string userAgent = string.Format(CultureInfo.InvariantCulture,
                     "{0} ({1}; {2}; {3}) AppleWebKit/534.6 (KHTML, like Gecko) Chrome/7.0.500.0 Safari/534.6",
-                    Compatibility, Platform, OS, Culture);
+                    Compatibility, PlatformName, OS, Culture);
                 return (userAgent);
             }
         }
@@ -82,7 +82,7 @@ namespace Microsoft.PowerShell.Commands
                 // format the user-agent string from the various component parts
                 string userAgent = string.Format(CultureInfo.InvariantCulture,
                     "Opera/9.70 ({0}; {1}; {2}) Presto/2.2.1",
-                    Platform, OS, Culture);
+                    PlatformName, OS, Culture);
                 return (userAgent);
             }
         }
@@ -97,7 +97,7 @@ namespace Microsoft.PowerShell.Commands
                 // format the user-agent string from the various component parts
                 string userAgent = string.Format(CultureInfo.InvariantCulture,
                     "{0} ({1}; {2}; {3}) AppleWebKit/533.16 (KHTML, like Gecko) Version/5.0 Safari/533.16",
-                    Compatibility, Platform, OS, Culture);
+                    Compatibility, PlatformName, OS, Culture);
                 return (userAgent);
             }
         }
@@ -120,19 +120,19 @@ namespace Microsoft.PowerShell.Commands
             }
         }
 
-        internal static string Platform
+        internal static string PlatformName
         {
             get
             {
-                if(System.Management.Automation.Platform.IsWindows)
+                if (Platform.IsWindows)
                 {
                     return "Windows NT";
                 }
-                else if(System.Management.Automation.Platform.IsMacOS)
+                else if (Platform.IsMacOS)
                 {
                     return "Macintosh";
                 }
-                else if(System.Management.Automation.Platform.IsLinux)
+                else if (Platform.IsLinux)
                 {
                     return "Linux";
                 }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
@@ -4,6 +4,7 @@ Copyright (c) Microsoft Corporation.  All rights reserved.
 
 using System;
 using System.Management.Automation;
+using System.Runtime.InteropServices;
 using System.Globalization;
 
 namespace Microsoft.PowerShell.Commands
@@ -122,7 +123,16 @@ namespace Microsoft.PowerShell.Commands
         {
             get
             {
-                return ("Windows NT");
+                if(RuntimeInformation.IsOSPlatform(OSPlatform.Windows)){
+                    return "Windows NT";
+                }else if(RuntimeInformation.IsOSPlatform(OSPlatform.OSX)){
+                    return "Macintosh";
+                }else if(RuntimeInformation.IsOSPlatform(OSPlatform.Linux)){
+                    return "Linux";
+                }else{
+                    // unknown/unsupported platform
+                    throw new PlatformNotSupportedException("Platform is not supported");
+                }
             }
         }
 
@@ -130,7 +140,7 @@ namespace Microsoft.PowerShell.Commands
         {
             get
             {
-                return System.Runtime.InteropServices.RuntimeInformation.OSDescription;
+                return RuntimeInformation.OSDescription
             }
         }
 
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
@@ -6,7 +6,6 @@ using System;
 using System.Management.Automation;
 using System.Runtime.InteropServices;
 using System.Globalization;
-using System.Diagnostics;
 
 namespace Microsoft.PowerShell.Commands
 {
@@ -139,7 +138,7 @@ namespace Microsoft.PowerShell.Commands
                 else
                 {
                     // unknown/unsupported platform
-                    Debug.Assert(true, "Unable to determine Operating System Platform");
+                    Diagnostics.Assert(false, "Unable to determine Operating System Platform");
                     return String.Empty;
                 }
             }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
@@ -6,6 +6,7 @@ using System;
 using System.Management.Automation;
 using System.Runtime.InteropServices;
 using System.Globalization;
+using System.Diagnostics;
 
 namespace Microsoft.PowerShell.Commands
 {
@@ -123,15 +124,23 @@ namespace Microsoft.PowerShell.Commands
         {
             get
             {
-                if(RuntimeInformation.IsOSPlatform(OSPlatform.Windows)){
+                if(System.Management.Automation.Platform.IsWindows)
+                {
                     return "Windows NT";
-                }else if(RuntimeInformation.IsOSPlatform(OSPlatform.OSX)){
+                }
+                else if(System.Management.Automation.Platform.IsMacOS)
+                {
                     return "Macintosh";
-                }else if(RuntimeInformation.IsOSPlatform(OSPlatform.Linux)){
+                }
+                else if(System.Management.Automation.Platform.IsLinux)
+                {
                     return "Linux";
-                }else{
+                }
+                else
+                {
                     // unknown/unsupported platform
-                    throw new PlatformNotSupportedException("Platform is not supported");
+                    Debug.Assert(true, "Unable to determine Operating System Platform");
+                    return String.Empty;
                 }
             }
         }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
@@ -131,7 +131,7 @@ namespace Microsoft.PowerShell.Commands
                     // only generate the windows user agent once
                     if(s_windowsUserAgent == null){
                         // find the version in the windows operating system description
-                        string versionText = OS.TrimEnd().Substring(OS.LastIndexOf(" ") +1);
+                        string versionText = OS.Substring(OS.LastIndexOf(" ") +1);
                         Version windowsPlatformversion = new Version(versionText);
                         s_windowsUserAgent = $"Windows NT {windowsPlatformversion.Major}.{windowsPlatformversion.Minor}";
                     }
@@ -159,7 +159,7 @@ namespace Microsoft.PowerShell.Commands
         {
             get
             {
-                return RuntimeInformation.OSDescription;
+                return RuntimeInformation.OSDescription.Trim();
             }
         }
 
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
@@ -140,7 +140,7 @@ namespace Microsoft.PowerShell.Commands
         {
             get
             {
-                return RuntimeInformation.OSDescription
+                return RuntimeInformation.OSDescription;
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
```diff
@@ -122,15 +123,24 @@ internal static string Platform
         {
             get
             {
-                return ("Windows NT");
+                if(RuntimeInformation.IsOSPlatform(OSPlatform.Windows)){
```
iSazonov: I suggest use `public static class Platform` from CorePsPlatform.cs and rename `Platform` in PSUserAgent.cs to PSUserAgentPlatform.

Also please use common formatting pattern:
```c#
if ( ... )
{
    ...
}
else if ( ... )
{
    ...
}
```

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
```diff
@@ -122,15 +123,24 @@ internal static string Platform
...
+                    return "Windows NT";
+                }else if(RuntimeInformation.IsOSPlatform(OSPlatform.OSX)){
+                    return "Macintosh";
+                }else if(RuntimeInformation.IsOSPlatform(OSPlatform.Linux)){
+                    return "Linux";
+                }else{
+                    // unknown/unsupported platform
+                    throw new PlatformNotSupportedException("Platform is not supported");
```
iSazonov: Maybe better `Debug.Assert` and returm String.Empty?

iSazonov (commented):
@LDSpits Thanks for your contribution!

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
```diff
@@ -123,15 +124,23 @@ internal static string Platform
         {
             get
             {
-                if(RuntimeInformation.IsOSPlatform(OSPlatform.Windows)){
+                if(System.Management.Automation.Platform.IsWindows)
```
iSazonov: Please add a space "if". Below too.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
```diff
@@ -123,15 +124,23 @@ internal static string Platform
...
+                {
                     return "Windows NT";
-                }else if(RuntimeInformation.IsOSPlatform(OSPlatform.OSX)){
+                }
+                else if(System.Management.Automation.Platform.IsMacOS)
```
iSazonov: Above I suggest rename _internal_ Platform property to exclude using "System.Management.Automation".

LDSpits: any preference on the name? how about OSPlatform? @iSazonov

iSazonov: It is a string name - maybe PlatformName?

dantraMSFT:
Minor Nit: I suggest you change the title of the PR to something like 'User Agent now reports the OS platform'.  When I first read the title, I was expecting to find a behavioral change that is dependent on the OS platform.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
```diff
@@ -118,19 +120,36 @@ internal static string App
...
+                    return "Macintosh";
+                }
+                else if (Platform.IsLinux)
+                {
+                    return "Linux";
+                }
+                else
+                {
+                    // unknown/unsupported platform
+                    Debug.Assert(true, "Unable to determine Operating System Platform");
```
daxian-dbw: I think this should be `Debug.Assert(false, "...")`. And it's recommended to use `Diagnostics.Assert` which is from the `System.Management.Automation` namespace. See [the code here](https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.Commands.Utility/commands/utility/compare-object.cs#L154) as an example.

daxian-dbw (dismissed):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
LDSpits: @daxian-dbw I have adressed your feedback.

daxian-dbw:
@LDSpits Thanks for the update. Can you please add corresponding tests? You can use https://github.com/PowerShell/PowerShell/blob/master/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1#L456 as an example to get `user-agent`.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
```diff
@@ -118,19 +119,36 @@ internal static string App
...
 
-        internal static string Platform
+        internal static string PlatformName
         {
             get
             {
-                return ("Windows NT");
+                if (Platform.IsWindows)
+                {
+                    return "Windows NT";
```
markekraus: I think we should add tests for these. These could either be a separate test or the existing 'User-Agent` tests could be expanded. 

https://github.com/PowerShell/PowerShell/blob/b07f24e3c9fec8bdeee0f70cb31edbbff0e29b9b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1#L456

https://github.com/PowerShell/PowerShell/blob/b07f24e3c9fec8bdeee0f70cb31edbbff0e29b9b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1#L1308

markekraus (changes requested):
Please add tests.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
iSazonov: The test should be patform sensetive - I'd prefer new separate test.

markekraus: @iSazonov I agree. 
@LDSpits you can use these tests as a base. Please include tests for both `Invoke-WebRequest` and `Invoke-RestMethod`

LDSpits: alright, will start work on them later today.

markekraus: @LDSpits Also, it looks like the Windows platform part is supposed to include the `<major>.<minor>` version of windows (6.1, 10.0, etc.). I wouldn't call this a requirement, but if you could add that too it would be great!

iSazonov: I think it's better to put it off. If we want more accurate emulation of User-Agent we should discuss in new Issue.

markekraus: It depends on how you look at it. if you view the current issue as "the platform is broken and wrong" then adding the version to the Windows in this PR is correct. After 6.0.0 release I wouldn't recommend changing the platform format until 7.0.0 as changes to it will lead to very undesirable results for users. If we do track that as a separate issue, it should also be 6.0.0-HighPriority.

daxian-dbw: There are a few ways to get the `<major>.<minor>` version of windows:
- Call [`GetVersionEx`](https://msdn.microsoft.com/en-us/library/windows/desktop/ms724451(v=vs.85).aspx) via P/Invoke.
- Inspect the `VersionInfo` of a system library. However, the tricky part of this option is to pick the right system binary to inspect because the one you pick might not be available on NanoServer.
```
> $s = Get-Item C:\Windows\System32\kernel32.dll
> $s.VersionInfo.ProductMajorPart
10
> $s.VersionInfo.ProductMinorPart
0
```
- Parse `[RuntimeInformation]::OSDescription`. It should be simple enought too.
```
$osdesp = [System.Runtime.InteropServices.RuntimeInformation]::OSDescription.Trim()
$vString = $osdesp.Substring($osdesp.LastIndexOf(" ")+1)
$version = [version]::new($vString)
$version.Major
$version.Minor
```

daxian-dbw:
@LDSpits One more thing, when you push new commits, please make sure the message of the last commit contains `[Feature]` in it so that it will trigger a full test run. Web cmdlets tests by default are not included in CI runs, so we need a full test run to validate changes related to them. Thanks!

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
iSazonov: GetVersionEx looks good. And we could cache the result.

daxian-dbw: @iSazonov I actually personally like parsing `[RuntimeInformation]::OSDescription`. The parsing is a one-time effort and pretty easy. I know you might concern about the consistency of the string value returned from this property ... :)

iSazonov: Yes. Ex. - What about Unix? I wouldn't trust `OSDescription`. I'd prefer to discuss all our User-Agent string (properties) before enhance this one.

daxian-dbw: I don't think we need the version info for Linux. This is what I got when using chrome to send request to `http://httpbin.org/user-agent`:
Windows: (Windows NT 10.0; Win64; x64)
Ubuntu: (X11; Linux x86_64)

iSazonov: Please see [FireFox](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent/Firefox). Specially MacOS and Android.

markekraus: @daxian-dbw is correct. The OS version numbers in the platform are only sent for Windows platforms. Linux platforms send either `X11` or `Linux` (`Linux` is used more in CLI based HTTP agents). macOS sends `Machintosh`. None of those with versions.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
```diff
@@ -118,19 +119,39 @@ internal static string App
...
+        internal static string PlatformName
         {
             get
             {
-                return ("Windows NT");
+                if (Platform.IsWindows)
+                {
+                    // find the version in the windows operating system description
+                    string versionText = PSUserAgent.OS.Substring(PSUserAgent.OS.TrimEnd().LastIndexOf(" ") +1);
+                    Version windowsPlatformversion = new Version(versionText);
```
markekraus: Since this would be dynamically generated each time the Web Cmdlets are called, but is static beyond the initial call, perhaps adding a private field `_windowsPlatformversionString`  (or something) generated once on the first call (`if (_windowsPlatformversionString == null)`) and then reference it directly every time thereafter. `return  _windowsPlatformversionString`

markekraus (changes requested):

LDSpits:
@iSazonov @daxian-dbw @markekraus Is it really nessecary to move the tests to a new file? shouldn't we add the test to WebCmdlets.tests.ps1 since most of the infrastructure required for the tests is already inside of there?

markekraus:
@LDSpits My preference would be for the tests to remain in `WebCmdlets.tests.ps1` along with the other `User-Agent` test. They should be new `It` blocks in the same file preferably right after the current ones. Here is the one for `Invoke-WebRequest`:

https://github.com/PowerShell/PowerShell/blob/2e901e53c3c1bfc3eebd66c3ae3223cd3ca6a499/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1#L456-L467

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
iSazonov: It is good candidate for private static variable and private static initialization function.

LDSpits:
@iSazonov @daxian-dbw @markekraus can I ask a few questions about tests? I was in the process of adding tests for ``execute-webrequest``. I have copied over most of the code from the ``Invoke-WebRequest returns User-Agent`` test with only changing the regex to ``'.*\(Windows NT \d+\.\d*;.*\) PowerShell\/\d+\.\d+\.\d+.*'``. this test failed however. so I tested the regex using powershell (to ensure i was using the same regex parsing engine as pester). However this returned true. So now im puzzled as to why it failed. Could you maybe give some pointers?

powershell code used to test regex:
```powershell
PS C:\Users\spits\Desktop\Development\Powershell\PowerShell> $bingdings =[System.Reflection.BindingFlags]::NonPublic -b
xor [System.Reflection.BindingFlags]::Static
PS C:\Users\spits\Desktop\Development\Powershell\PowerShell> [Microsoft.PowerShell.Commands.PSUserAgent].GetProperty('Us
erAgent',$bingdings).GetValue($null,$null) -match '.*\(Windows NT \d+\.\d*;.*\) PowerShell\/\d+\.\d+\.\d+.*'
True
```

Current (windows) test:

```powershell
It "Invoke-WebRequest returns Correct User-Agent on Windows" {

        $uri = Get-WebListenerUrl -Test 'Get'
        $command = "Invoke-WebRequest -Uri '$uri' -TimeoutSec 5"

        ValidateResponse -response $result

        # Validate response content
        $jsonContent = $result.Output.Content | ConvertFrom-Json
        $jsonContent.headers.'User-Agent' | Should MatchExactly '.*\(Windows NT \d+\.\d*;.*\) PowerShell\/\d+\.\d+\.\d+.*'
    }
```

pester test message:
```powershell
Describing Invoke-WebRequest tests
 [-] Invoke-WebRequest returns Correct User-Agent on MacOSX 6.87s
   Expected: {Mozilla/5.0 (Windows NT; Microsoft Windows 10.0.16296 ; nl-NL) PowerShell/6.0.0} to exactly match the expression {.*\(Macintosh;.*\) PowerShell\/\d+\.\d+\.\d+.*}
   at line: 466 in C:\Users\spits\Desktop\Development\Powershell\PowerShell\test\powershell\Modules\Microsoft.PowerShell
.Utility\WebCmdlets.Tests.ps1
   466:         $jsonContent.headers.'User-Agent' | Should MatchExactly '.*\(Macintosh;.*\) PowerShell\/\d+\.\d+\.\d+.*'
 [-] Invoke-WebRequest returns Correct User-Agent on Linux 1.14s
   Expected: {Mozilla/5.0 (Windows NT; Microsoft Windows 10.0.16296 ; nl-NL) PowerShell/6.0.0} to exactly match the expr
ession {.*\(Linux;.*\) PowerShell\/\d+\.\d+\.\d+.*}
   at line: 479 in C:\Users\spits\Desktop\Development\Powershell\PowerShell\test\powershell\Modules\Microsoft.PowerShell
.Utility\WebCmdlets.Tests.ps1
   479:         $jsonContent.headers.'User-Agent' | Should MatchExactly '.*\(Linux;.*\) PowerShell\/\d+\.\d+\.\d+.*'
 [-] Invoke-WebRequest returns Correct User-Agent on Windows 1.08s
   Expected: {Mozilla/5.0 (Windows NT; Microsoft Windows 10.0.16296 ; nl-NL) PowerShell/6.0.0} to exactly match the expression {.*\(Windows NT \d+\.\d*;.*\) PowerShell\/\d+\.\d+\.\d+.*}
   at line: 492 in C:\Users\spits\Desktop\Development\Powershell\PowerShell\test\powershell\Modules\Microsoft.PowerShell
.Utility\WebCmdlets.Tests.ps1
   492:         $jsonContent.headers.'User-Agent' | Should MatchExactly '.*\(Windows NT \d+\.\d*;.*\) PowerShell\/\d+\.\
d+\.\d+.*'
```

markekraus:
@LDSpits It looks like you are not using a version of PowerShell Compiled with your changes. The pester output shows `Mozilla/5.0 (Windows NT; Microsoft Windows 10.0.16296 ; nl-NL) PowerShell/6.0.0` as the user-agent. not that the `<major>.<minor>` is missing after `Windows NT`

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
```diff
@@ -13,14 +14,17 @@ namespace Microsoft.PowerShell.Commands
     /// </summary>
     public static class PSUserAgent
     {
+
+        private static string _windowsUserAgent;
```
iSazonov: Please use 's_' prefix for static variables.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
```diff
@@ -118,19 +122,44 @@ internal static string App
...
-        internal static string Platform
+        internal static string PlatformName
         {
             get
             {
-                return ("Windows NT");
+                if (Platform.IsWindows)
+                {
+                    // only generate the windows user agent once
+                    if(PSUserAgent._windowsUserAgent == null){
```
iSazonov: Please remove 'PSUserAgent' prefix. Below too.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/PSUserAgent.cs
```diff
@@ -118,19 +122,44 @@ internal static string App
...
+                        // find the version in the windows operating system description
+                        string versionText = PSUserAgent.OS.Substring(PSUserAgent.OS.TrimEnd().LastIndexOf(" ") +1);
```
iSazonov: Do we really need `TrimEnd()`?
If so it is better mobe it in OS property.

Please remove 'PSUserAgent' prefix.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -453,8 +453,22 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         $response.Output.Content | Should Not Be $null
     }
 
-    It "Invoke-WebRequest returns User-Agent" {
+    #User-Agent changes on different platforms, so tests should only be run if on the correct platform
+    It "Invoke-WebRequest returns Correct User-Agent on MacOSX" -Pending:(!$IsMacOS) {
```
iSazonov: We use `Pending` to temporary turn out broken tests.
Please use `Skip`. Below too.

markekraus (approved):

LDSpits:
@iSazonov @daxian-dbw @markekraus the Travis build is erroring? had a problem with the previous build on the http proxy tests...

markekraus:
@LDSpits Its related to #5005. We need to fix the other delay tests. I'll create an issue for it.

daxian-dbw:
#5035 has been merged and I restarted the CI builds. Hopefully the CI will pass this time.

LDSpits:
Let's hope they'll pass this time 😄

markekraus (approved):
LGTM

daxian-dbw:
@LDSpits Thanks for your contribution!
@markekraus @iSazonov Thank you for the thorough review!

markekraus:
@LDSpits Thanks for doing this!

LDSpits:
@daxian-dbw @markekraus No problem, expect to see me around 😄

PR #4942, opened 2017-09-28
removed RunspaceConfiguration support

SteveL-MSFT:
For PSCore6, we are only supporting InitialSessionState.  The RunspaceConfiguration api was already made internal.  This PR removes all the code related to RunspaceConfiguration.  This also means that some public apis have changed.  Was deciding between leaving the RunspaceConfiguration parameters and throwing Unsupported, but thought it was better to have it a compile time error.  This should simplify the code base.


daxian-dbw:
@SteveL-MSFT `powershell.Invoke()` (maybe some other part of the powershell class) goes through `RunspaceConfiguration`, for example the following script:
```
$ps = [powershell]::Create()
$ps.AddScript("gcm")
$ps.Invoke()
```
Did you change it to not depend on `RunspaceConfiguration`?

SteveL-MSFT:
@daxian-dbw yes, I changed it to use InitialSessionState

daxian-dbw:
@SteveL-MSFT Awesome! Can you please point me to the part of changes that related to `powershell.invoke()`?

SteveL-MSFT:
@daxian-dbw after ripping out the RunspaceConfig code, to get it working (and those tests passing) I [create a ISS instance with defaults](https://github.com/PowerShell/PowerShell/pull/4942/files#diff-1eb100aa3dfb92932b58036835b262bfR46).

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
```diff
@@ -109,20 +67,6 @@ internal void AddSessionStateEntry(SessionStateProviderEntry providerEntry)
             );
         }
 
-        /// <summary>
-        /// Internal method used by RunspaceConfig for updating providers.
-        /// </summary>
-        /// <param name="providerConfig"></param>
-        private ProviderInfo AddProvider(ProviderConfigurationEntry providerConfig)
```
PaulHigin: It looks like RemoveProvider on line 1523 should be removed too, since it is only used for RunspaceConfiguration.

PaulHigin (changes requested):
Not quite half way through the code review, but I am publishing the comments I have so far.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Update-TypeData.cs
```diff
@@ -1055,11 +985,6 @@ protected override void ProcessRecord()
                     if (entries.Count > 0)
                     {
                         Context.FormatDBManager.UpdateDataBase(entries, this.Context.AuthorizationManager, this.Context.EngineHostInterface, false);
-                        FormatAndTypeDataHelper.ThrowExceptionOnError(
```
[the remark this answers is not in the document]
SteveL-MSFT: From the code, it appears that this helper is only used with RunspaceConfiguration processing Format and Type data.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -1743,46 +1709,10 @@ private void DoRunspaceInitialization(bool importSystemModules, bool skipProfile
...
-                }
-
                 // If -iss has been specified, then there won't be a runspace
                 // configuration to get the shell ID from, so we'll use the default...
                 string shellId = null;
-                if (_configuration != null)
-                    shellId = _configuration.ShellId;
-                else
-                    shellId = "Microsoft.PowerShell"; // TODO: what will happen for custom shells built using Make-Shell.exe
+                shellId = "Microsoft.PowerShell"; // TODO: what will happen for custom shells built using Make-Shell.exe
```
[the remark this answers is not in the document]
SteveL-MSFT: I think the commend can be removed

src/Microsoft.PowerShell.ConsoleHost/host/msh/ManagedEntrance.cs
```diff
@@ -55,54 +55,37 @@ public static int Start(string consoleFilePath, [MarshalAs(UnmanagedType.LPArray
...
-                        ConsoleHost.DefaultInitialSessionState = InitialSessionState.CreateDefault2();
-                        configuration = null;
-                    }
-                    else
-                    {
-                        ConsoleHost.DefaultInitialSessionState = InitialSessionState.CreateDefault2();
-                        configuration = null;
-                    }
-#else
+            if (args.Length > 0 && !String.IsNullOrEmpty(args[0]) && args[0].Equals("-iss", StringComparison.OrdinalIgnoreCase))
```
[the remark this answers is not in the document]
SteveL-MSFT: Yes, will move it out

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1114,31 +1098,21 @@ internal void RunspaceClosingNotification()
...
                 return _typeTable;
             }
-            // This needs to exist so that RunspaceConfiguration can
-            // push it's shared type table into ExecutionContext
             set
             {
-                if (this.RunspaceConfiguration != null)
-                    throw new NotImplementedException("set_TypeTable()");
                 _typeTable = value;
                 _typeTableWeakReference = value != null ? new WeakReference<TypeTable>(value) : null;
```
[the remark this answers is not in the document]
SteveL-MSFT: I can fix that

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1585,14 +1483,8 @@ private bool IsModuleCommandCurrentlyRunning(out Cmdlet command, out string erro
         /// <param name="hostInterface">
         /// Interface that should be used for interaction with host
         /// </param>
-        /// <param name="runspaceConfiguration">
-        /// RunspaceConfiguration information
-        /// </param>
-        internal ExecutionContext(AutomationEngine engine, PSHost hostInterface, RunspaceConfiguration runspaceConfiguration)
+        internal ExecutionContext(AutomationEngine engine, PSHost hostInterface)
```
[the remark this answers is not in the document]
SteveL-MSFT: I'll check; if this isn't used anymore I'll remove it

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -5070,9 +5069,6 @@ internal void SaveAsConsoleFile(string path)
             {
                 throw PSTraceSource.NewArgumentException("path", ConsoleInfoErrorStrings.BadConsoleExtension);
             }
-
```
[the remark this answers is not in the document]
SteveL-MSFT: PSConsoleFile is not supported as it was only used with RunspaceConfiguration

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
SteveL-MSFT: Will remove

PaulHigin (changes requested):

src/System.Management.Automation/engine/InitialSessionState.cs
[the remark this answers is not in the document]
SteveL-MSFT: Yes, SaveAsConsoleFile() should be removed

src/System.Management.Automation/engine/hostifaces/ConnectionBase.cs
```diff
@@ -37,22 +37,14 @@ internal abstract class RunspaceBase : Runspace
...
             if (host == null)
             {
                 throw PSTraceSource.NewArgumentNullException("host");
             }
-            if (runspaceConfiguration == null)
-            {
-                throw PSTraceSource.NewArgumentNullException("runspaceConfiguration");
-            }
-
+            InitialSessionState = InitialSessionState.CreateDefault2();
```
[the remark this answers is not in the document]
SteveL-MSFT: Yes, this is a breaking change.  Our test coverage may not be sufficient to cover this so I need someone with experience here to point out how this may have adverse effects.

src/System.Management.Automation/engine/hostifaces/RunspacePoolInternal.cs
```diff
@@ -76,21 +72,14 @@ internal class RunspacePoolInternal
...
-            {
-                throw PSTraceSource.NewArgumentNullException("runspaceConfiguration");
-            }
-
             if (host == null)
             {
                 throw PSTraceSource.NewArgumentNullException("host");
             }
 
-            rsConfig = runspaceConfiguration;
```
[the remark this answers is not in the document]
SteveL-MSFT: I'll create a default one

src/System.Management.Automation/minishell/api/RunspaceConfiguration.cs
```diff
@@ -758,34 +45,4 @@ internal enum RunspaceConfigurationCategory
         Formats,
```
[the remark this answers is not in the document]
SteveL-MSFT: I think I originally wanted to remove it, but it contains an enum used by FormatAndTypeDataHelper which I think is still needed for some PSSnapins we ship even if we don't support custom PSSnapins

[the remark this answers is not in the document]
lzybkr: You could rename - it's being used in `Update-FormatData` which is not snap-in related.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and agreed on removal on RunspaceConfiguration code

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
```diff
@@ -1749,7 +1468,7 @@ internal int ProviderCount
                 return count;
             }
         } // ProviderCount
-    } // SessionStateInternal class
+    } // SessionStateInternal class}
```
PaulHigin: This looks strange.  Can you remove the bracket from the comment?

src/System.Management.Automation/engine/hostifaces/ConnectionBase.cs
```diff
@@ -43,7 +43,7 @@ protected RunspaceBase(PSHost host)
             {
                 throw PSTraceSource.NewArgumentNullException("host");
             }
-            InitialSessionState = InitialSessionState.CreateDefault2();
+            InitialSessionState = InitialSessionState.CreateDefault();
```
PaulHigin: Are we sure that we want to use CreateDefault() and load all commands/functions?  Or CreateDefault2() where just Core commands are loaded (and is faster)?  Please add a comment to justify which default we use.

src/System.Management.Automation/engine/hostifaces/RunspacePoolInternal.cs
```diff
@@ -84,6 +84,7 @@ internal class RunspacePoolInternal
             pool = new Stack<Runspace>();
             runspaceRequestQueue = new Queue<GetRunspaceAsyncResult>();
             ultimateRequestQueue = new Queue<GetRunspaceAsyncResult>();
+            _initialSessionState = InitialSessionState.CreateDefault();
```
PaulHigin: Are we sure that we want to use CreateDefault() and load all commands/functions?  Or CreateDefault2() where just Core commands are loaded (and is faster)?  Please add a comment to justify which default we use.

PaulHigin (changes requested):

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
SteveL-MSFT: Yes, that was probably just a mistake

src/System.Management.Automation/engine/hostifaces/ConnectionBase.cs
SteveL-MSFT: This is where I need some expertise to mimic the previous RunspaceConfiguration behavior.  @lzybkr @daxian-dbw ?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Update-TypeData.cs
lzybkr: I believe we still need this code. A test might look like:

```powershell
"@
<Configuration>
  <ViewDefinitions>    
    <View>
      <Name>T2</Name>
    </View>   
  </ViewDefinitions>
</Configuration>
@" > test.format.ps1xml

{ Update-FormatData test.format.ps1xml } | Should Throw
```

---
In reply to: [142809956](https://github.com/PowerShell/PowerShell/pull/4942#discussion_r142809956) [](ancestors = 142809956)

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
```diff
@@ -653,17 +650,10 @@ private void DoOpenHelper()
...
-                    _engine = new AutomationEngine(Host, null, InitialSessionState);
-                }
-                else
-                {
-                    _engine = new AutomationEngine(Host, RunspaceConfiguration, null);
-                }
+                // All ISS-based configuration of the engine itself is done by AutomationEngine,
+                // which calls InitialSessionState.Bind(). Anything that doesn't
+                // require an active and open runspace should be done in ISS.Bind()
+                _engine = new AutomationEngine(Host, InitialSessionState);
```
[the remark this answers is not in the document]
SteveL-MSFT: yes, added

src/Microsoft.PowerShell.ConsoleHost/host/msh/ManagedEntrance.cs
```diff
@@ -55,54 +55,26 @@ public static int Start(string consoleFilePath, [MarshalAs(UnmanagedType.LPArray
...
-                    }
-#else
-                ConsoleHost.DefaultInitialSessionState = InitialSessionState.CreateDefault2();
-                configuration = null;
-#endif
-            }
-            else
+            // Special switch for debug mode
+            if (args.Length > 0 && !String.IsNullOrEmpty(args[0]) && args[0].Equals("-isswait", StringComparison.OrdinalIgnoreCase))
             {
```
lzybkr: I'd move this code earlier, like in main. I needed similar code in the past and never noticed it here.
My version was little nicer, waiting for the debugger to attach (in a loop, with time outs) and then hit a breakpoint.

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1090,13 +1079,6 @@ internal ConfirmImpact ConfirmPreferenceVariable
 
         internal void RunspaceClosingNotification()
         {
-            if (this.RunspaceConfiguration != null)
-            {
-                this.RunspaceConfiguration.Unbind(this);
-            }
-
-            EngineSessionState.RunspaceClosingNotification();
```
lzybkr: Why is this line deleted?

src/System.Management.Automation/engine/SessionState.cs
```diff
@@ -391,11 +391,6 @@ internal void SetConsoleVariable()
         {
```
lzybkr: Consider removing this whole function and the special variable now that console files went away.

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
```diff
@@ -1510,285 +1452,6 @@ private void NewProviderEntry(ProviderInfo provider)
...
-                        }
-                    }
-
-                    // Now tell the provider that they are going to be removed by
-                    // calling the Stop method
-
-                    try
-                    {
-                        providerBase.Stop(context);
-                    }
```
lzybkr: It's hard to be certain all of this code can be safely deleted, but this line is worrisome - it potentially calls third party code to clean up resources.

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
```diff
@@ -1510,285 +1452,6 @@ private void NewProviderEntry(ProviderInfo provider)
...
-                        providerName));
-            }
-            finally
-            {
-                if (force || !errors)
-                {
-                    // Log the provider stopped event
-
-                    MshLog.LogProviderLifecycleEvent(
-                        this.ExecutionContext,
```
lzybkr: It's possible this event is used by forensic tools - as much as I'd like to get rid of it.

src/System.Management.Automation/engine/SessionState.cs
```diff
@@ -487,34 +482,6 @@ private SessionStateEntryVisibility checkPathVisibility(List<string> list, strin
...
-        internal void RunspaceClosingNotification()
-        {
-            if (this != ExecutionContext.TopLevelSessionState && Providers.Count > 0)
-            {
-                // Remove all providers at the top level...
-
-                CmdletProviderContext context = new CmdletProviderContext(this.ExecutionContext);
-
-                Collection<string> keys = new Collection<string>();
-                foreach (string key in Providers.Keys)
```
lzybkr: I'm not sure we should remove this method, but this extra loop is pointless - just copying the keys to a temp collection.

src/System.Management.Automation/engine/TypeTable.cs
```diff
@@ -4218,9 +4218,6 @@ public void AddType(TypeData typeData)
 
             Update(errors, typeData, false);
             StandardMembersUpdated();
-
-            // Throw exception if there are any errors
-            FormatAndTypeDataHelper.ThrowExceptionOnError("ErrorsUpdatingTypes", errors, RunspaceConfigurationCategory.Types);
         }
```
lzybkr: I think this is still needed.

src/System.Management.Automation/engine/TypeTable.cs
```diff
@@ -4245,9 +4242,6 @@ public void RemoveType(string typeName)
 
             Update(errors, typeData, true);
             StandardMembersUpdated();
-
-            // Throw exception if there are any errors
-            FormatAndTypeDataHelper.ThrowExceptionOnError("ErrorsUpdatingTypes", errors, RunspaceConfigurationCategory.Types);
         }
```
lzybkr: I think this is still needed.

src/System.Management.Automation/engine/hostifaces/ConnectionBase.cs
lzybkr: I'm not sure this api makes sense anymore - it was accepting a runspace config, so the replacement would take an initial session state, but those apis already exist.

---
In reply to: [144347463](https://github.com/PowerShell/PowerShell/pull/4942#discussion_r144347463) [](ancestors = 144347463)

src/System.Management.Automation/engine/hostifaces/ConnectionFactory.cs
```diff
@@ -67,72 +65,12 @@ public static Runspace CreateRunspace(PSHost host)
                 throw PSTraceSource.NewArgumentNullException("host");
             }
 
-            return CreateRunspace(host, RunspaceConfiguration.Create());
+            return new LocalRunspace(host);
         }
```
lzybkr: We should construct the InitialSessionState here and call the appropriate constructor.

As for using CreateDefault vs. CreateDefault2, the argument that CreateDefault2 is faster is apparently very wrong, I measure it as 8X slower:

```
#88 PS> 'CreateDefault','CreateDefault2' | %{  measure-command {
>> $iss = [InitialSessionState]::$_();
>> 1..100 | % { $ps = [powershell]::Create($iss).AddCommand('echo').AddArgument(1); $ps.Invoke(); $ps.Dispose() } } }

1.003s
8.247s
```

If I change the example to not load any commands, then they are more similar:

#89 PS> 'CreateDefault','CreateDefault2' | %{  measure-command {
>> $iss = [InitialSessionState]::$_();
>> 1..100 | % { $ps = [powershell]::Create($iss).AddScript('1'); $ps.Invoke(); $ps.Dispose() } } }

1.209s
1.037s

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
```diff
@@ -38,12 +38,9 @@ internal sealed partial class LocalRunspace : RunspaceBase
...
         /// The explicit PSHost implementation
         /// </param>
-        /// <param name="runspaceConfig">
-        /// configuration information for this minshell.
-        /// </param>
         [SuppressMessage("Microsoft.Maintainability", "CA1505:AvoidUnmaintainableCode")]
-        internal LocalRunspace(PSHost host, RunspaceConfiguration runspaceConfig)
-            : base(host, runspaceConfig)
+        internal LocalRunspace(PSHost host)
+            : base(host)
```
lzybkr: Does this api make sense anymore?

lzybkr:
>                 {"Get-PSSnapin",                      new SessionStateCmdletEntry("Get-PSSnapin", typeof(GetPSSnapinCommand), helpFile) },

You can delete this line.

---
Refers to: src/System.Management.Automation/engine/InitialSessionState.cs:5572 in 6121d89. [](commit_id = 6121d891d98cf4f813ac013ef8faa1bc8e68296d, deletion_comment = False)

lzybkr:
>                 {"Get-PSSnapin",                      new SessionStateCmdletEntry("Get-PSSnapin", typeof(GetPSSnapinCommand), helpFile) },

And the *PSSnapin lines.

---
In reply to: [336302306](https://github.com/PowerShell/PowerShell/pull/4942#issuecomment-336302306) [](ancestors = 336302306)

---
Refers to: src/System.Management.Automation/engine/InitialSessionState.cs:5572 in 6121d89. [](commit_id = 6121d891d98cf4f813ac013ef8faa1bc8e68296d, deletion_comment = False)

lzybkr (dismissed):
:clock1:

src/Microsoft.PowerShell.ConsoleHost/host/msh/ManagedEntrance.cs
SteveL-MSFT: Is this code needed anymore?  In VSCode, it will break in main() anyways.  Plus this code looks to be an artifact of when ISS was new as it's blocking at the creation of ISS.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Update-TypeData.cs
SteveL-MSFT: I misunderstood the relationship between the formatdata code and the runspace config code.  Will put it back and add the test.

SteveL-MSFT: I tried this (fixing the @" "@ here string) and it doesn't throw on Beta.8.

src/System.Management.Automation/engine/ExecutionContext.cs
SteveL-MSFT: Mistake.

src/System.Management.Automation/engine/SessionState.cs
SteveL-MSFT: will remove

src/Microsoft.PowerShell.ConsoleHost/host/msh/ManagedEntrance.cs
lzybkr: Sometimes you need to launch from another process, so it can be useful. But you could just remove it and the next time it's needed, something like it will come back.

src/System.Management.Automation/engine/SessionState.cs
SteveL-MSFT: I wasn't sure about this method.  I set a breakpoint in it and couldn't get it to hit within the if statement using runspace creation.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Update-TypeData.cs
lzybkr: It does for me, Windows PowerShell and Beta 8 on Windows.

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
SteveL-MSFT: Best that I could tell, RemoveProvider() is only used for RunspaceConfiguration

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
SteveL-MSFT: If we know that RemoveProvider is used outside of RunspaceConfig, I can put this back, but currently it doesn't seem to be the case.

src/System.Management.Automation/engine/TypeTable.cs
SteveL-MSFT: Will put back

src/System.Management.Automation/engine/TypeTable.cs
SteveL-MSFT: Will put back

src/System.Management.Automation/engine/hostifaces/ConnectionFactory.cs
SteveL-MSFT: will update

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
SteveL-MSFT: will remove due to change above

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
lzybkr: I think there is a bug in `LocalRunspace.DoCloseHelper` then - there is an early return in there, and if it always returns, there are multiple things we're not doing at shutdown, e.g. we're not firing the exiting event.

Bottom line - not hitting the code doesn't mean much w.r.t. RunspaceConfiguration, at least based on my review of the code and quick debugging of process exit.

src/System.Management.Automation/engine/hostifaces/RunspacePoolInternal.cs
SteveL-MSFT: Based on Jason's test, CreateDefault() is the faster one while CreateDefault2() does more, I'll update the comments on those methods

lzybkr: Note that it feels like something is terribly wrong with `CreateDefault2`, so that's investigating. It should be faster like Paul suggested.

src/System.Management.Automation/engine/hostifaces/ConnectionFactory.cs
SteveL-MSFT: On my MacBookPro, CreateDefault2() is 3x slower and on Windows it's 2.5x slower, but consistently slower to use CreateDefault2()

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
SteveL-MSFT: I'll change the DoCloseHelper() method.  Based on the comments, it's preventing stopping transcription until all the runspaces are closed, so I'll just use a flag to catch if there's any open runspaces and will continue with the rest of DoCloseHelper() to shutdown.

src/System.Management.Automation/engine/hostifaces/RunspacePoolInternal.cs
SteveL-MSFT: I think that should be a separate issue.  https://github.com/PowerShell/PowerShell/issues/5104

src/Microsoft.PowerShell.ConsoleHost/host/msh/ManagedEntrance.cs
SteveL-MSFT: Will remove

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
SteveL-MSFT: based on our code coverage that code has never been hit

src/System.Management.Automation/minishell/api/RunspaceConfiguration.cs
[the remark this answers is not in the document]
SteveL-MSFT: Moving enum to FormatAndTypeDataHelper.cs and moving that file to utils.  Removing the last of the RunspaceConfiguration*.cs files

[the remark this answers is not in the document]
SteveL-MSFT: I take that back, RunspaceConfigurationEntry.cs is still needed.  I'll move the enum to this file put it under engine.  We can get rid of the minishell folder.

SteveL-MSFT:
I believe all feedback has been addressed

src/System.Management.Automation/engine/CommandSearcher.cs
```diff
@@ -133,18 +133,6 @@ public bool MoveNext()
                 }
 
                 // Advance the state
-                _currentState = SearchState.SearchingBuiltinScripts;
```
daxian-dbw: Is it safe to remove the enum number `SearchState.SearchingBuiltinScripts`? It doesn't seem to be needed anymore.

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -169,23 +169,17 @@ internal bool ShouldTraceStatement
...
+        /// True if the InitialSessionState is for a single shell or false otherwise.
         /// </summary>
         ///
         internal bool IsSingleShell
         {
             get
             {
-                RunspaceConfigForSingleShell runSpace = RunspaceConfiguration as RunspaceConfigForSingleShell;
-                return runSpace != null || InitialSessionState != null;
+                return InitialSessionState != null;
```
daxian-dbw: It looks to me it's always SingleShell now, as `InitialSessionState` is initialized in the constructor and cannot be null.
So, all `IsSingleShell` related the code can be cleaned as well.

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1194,12 +1159,8 @@ internal TypeInfoDataBaseManager FormatDBManager
                 return _formatDBManager;
             }
 
-            // This needs to exist so that RunspaceConfiguration can
-            // push it's shared format database table into ExecutionContext
             set
             {
-                if (this.RunspaceConfiguration != null)
-                    throw new NotImplementedException("set_FormatDBManager()");
                 _formatDBManager = value;
```
daxian-dbw: It looks to me we can remove the set method now, as it's only used for a RunspaceConfiguration scenario.

src/System.Management.Automation/engine/SessionState.cs
```diff
@@ -379,28 +379,6 @@ internal void InitializeFixedVariables()
...
-        internal void SetConsoleVariable()
-        {
-            // $Console - set the console file for this shell, if there is one, "" otherwise...
-            string consoleFileName = string.Empty;
-            RunspaceConfigForSingleShell rcss = ExecutionContext.RunspaceConfiguration as RunspaceConfigForSingleShell;
-            if (rcss != null && rcss.ConsoleInfo != null && !string.IsNullOrEmpty(rcss.ConsoleInfo.Filename))
-            {
-                consoleFileName = rcss.ConsoleInfo.Filename;
-            }
-            PSVariable v = new PSVariable(SpecialVariables.ConsoleFileName,
```
daxian-dbw: `SpecialVariables.ConsoleFileName` in SpecialVariables.cs can be removed now

src/System.Management.Automation/engine/SessionState.cs
daxian-dbw: The if statement looks to me is checking if the current `EngineSessionState` is a module session state and if the module defines providers. This seems will happen when a Runspace is closing while running cmdlets from a module.

`RemoveProvider(string providerName, bool force, CmdletProviderContext context)` doesn't look like solely related to `RunspaceConfigration`.

I think we should keep this method but remove the extra loop.

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
```diff
@@ -1510,285 +1452,6 @@ private void NewProviderEntry(ProviderInfo provider)
...
-        /// All drives associated with the provider must be removed before the provider
-        /// can be removed. Call SessionState.GetDrivesForProvider() to determine if there
-        /// are any drives associated with the provider. A SessionStateException
-        /// will be written to the context if any such drives do exist.
-        /// </remarks>
-        ///
-        internal void RemoveProvider(
-            string providerName,
-            bool force,
-            CmdletProviderContext context)
```
daxian-dbw: Same question here. This method doesn't look like solely related to RunspaceConfiguration.

src/System.Management.Automation/engine/TypeTable.cs
```diff
@@ -4221,7 +4221,7 @@ public void AddType(TypeData typeData)
 
             // Throw exception if there are any errors
             FormatAndTypeDataHelper.ThrowExceptionOnError("ErrorsUpdatingTypes", errors, RunspaceConfigurationCategory.Types);
-        }
+         }
```
daxian-dbw: This change seems unnecessary. The extra space should be removed.

PaulHigin (approved):
I was surprised that CreateDefault2() was so much slower creating commands than with CreateDefault() initialization.  I believe you said somewhere that you created an Issue to track this perf issue.

src/System.Management.Automation/engine/hostifaces/ConnectionBase.cs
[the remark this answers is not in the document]
daxian-dbw: If we want to mimic the existing behavior, then we should use `InitialSessionState.CreateDefault()` here. When using `RunspaceConfiguration`. all built-in powershell modules are loaded as snapins, so you will get nothing when running `Get-Module` even after executing `Get-Process`. With `CreateDefault()`, built-in powershell modules are loaded as snapins as well. But with `CreateDefault2()`, only the `Microsoft.PowerShell.Core` is loaded as snapin and the rest are loaded as modules, like a regular powershell console session does.

daxian-dbw (changes requested):
I'm now looking at `ConnectionFactory.cs`, but want to share my existing feedback early.

src/System.Management.Automation/engine/hostifaces/ConnectionFactory.cs
```diff
@@ -67,72 +65,12 @@ public static Runspace CreateRunspace(PSHost host)
                 throw PSTraceSource.NewArgumentNullException("host");
             }
 
-            return CreateRunspace(host, RunspaceConfiguration.Create());
+            return new LocalRunspace(host, InitialSessionState.CreateDefault2());
```
daxian-dbw: Same here, if we want to mimic the existing behavior, then we should use `CreateDefault`.

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
```diff
@@ -867,19 +838,21 @@ private void DoCloseHelper()
...
                         {
-                            return;
+                            stopTranscription = false;
+                            break;
                         }
                     }
 
                     PSHostUserInterface host = executionContext.EngineHostInterface.UI;
-                    if (host != null)
+                    if (host != null && stopTranscription)
```
daxian-dbw: Sorry if you already explained this change in some previous conversations, but I didn't find any.
The original code seems already makes sure that `StopAllTranscribing` will be called only if all runspaces are at closing state. As for `AmsiUtils.Uninitialize()`, it looks to me doesn't need to be called for every Runspace close operation, but stead should only be called once when the last Runspace is closed.

src/System.Management.Automation/engine/hostifaces/RunspacePoolInternal.cs
```diff
@@ -76,25 +72,19 @@ internal class RunspacePoolInternal
...
             {
                 throw PSTraceSource.NewArgumentNullException("host");
             }
 
-            rsConfig = runspaceConfiguration;
             this.host = host;
             pool = new Stack<Runspace>();
             runspaceRequestQueue = new Queue<GetRunspaceAsyncResult>();
             ultimateRequestQueue = new Queue<GetRunspaceAsyncResult>();
+            _initialSessionState = InitialSessionState.CreateDefault2();
```
daxian-dbw: Same here. If we want to mimic `RunspaceConfiguration`, then `CreateDefault()` should be used.

src/System.Management.Automation/help/HelpProvider.cs
```diff
@@ -246,9 +246,7 @@ internal string GetDefaultShellSearchPath()
         /// <returns>true if supported,false otherwise.</returns>
         internal bool AreSnapInsSupported()
```
daxian-dbw: I think this method should be removed, and update the two call sites in `HelpProviderWithFullCache.cs`:
`if (!this.CacheFullyLoaded || AreSnapInsSupported())` ==> `if (!this.CacheFullyLoaded)`

daxian-dbw (changes requested):
Done through all changes.

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
SteveL-MSFT: @lzybkr had a question about this section in that previously it would return out of this method early 
 if it's not the last runspace so that transcription would not be stopped.  Because it returned early, the rest of the closing actions would not be executed.

daxian-dbw: The rest of the closing method is `AmsiUtils.Uninitialize();` which I think should only execute once when the last Runspace is closing. It looks to me `AmsiUtils.Init()` should run only once and then can be shared across all Runspaces.

daxian-dbw: I remember `AMSI` is thread safe and @PaulHigin even fixed a bug previously to remove the locks when calling `AmsiUtils.ScanContent`. So I think it should be initialized once and used across all Runspaces.

src/System.Management.Automation/engine/CommandSearcher.cs
SteveL-MSFT: Yes, looks safe to remove as it's no longer being used

src/System.Management.Automation/engine/ExecutionContext.cs
SteveL-MSFT: yup, will clean up

src/System.Management.Automation/engine/ExecutionContext.cs
SteveL-MSFT: will remove

src/System.Management.Automation/engine/SessionState.cs
SteveL-MSFT: will remove

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
SteveL-MSFT: Putting back

src/System.Management.Automation/engine/TypeTable.cs
SteveL-MSFT: will fix

src/System.Management.Automation/engine/hostifaces/ConnectionFactory.cs
SteveL-MSFT: will change

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
SteveL-MSFT: I'll change it so if there's no open runspaces left, transcription is stopped and AmsiUtils.Uninitialize() is called.

src/System.Management.Automation/engine/hostifaces/RunspacePoolInternal.cs
SteveL-MSFT: will change

src/System.Management.Automation/help/HelpProvider.cs
SteveL-MSFT: will update and remove no longer relevant comment

src/System.Management.Automation/engine/ExecutionContext.cs
SteveL-MSFT: looks like InitialSessionState.cs:3582 uses it

src/Microsoft.PowerShell.ConsoleHost/host/msh/ManagedEntrance.cs
```diff
@@ -55,54 +55,15 @@ public static int Start(string consoleFilePath, [MarshalAs(UnmanagedType.LPArray
...
-            {
-#if DEBUG
-// Special switches for debug mode to allow self-hosting on InitialSessionState instead
-// of runspace configuration...
-                    if (args.Length > 0 && !String.IsNullOrEmpty(args[0]) && args[0].Equals("-iss", StringComparison.OrdinalIgnoreCase))
-                    {
-                        ConsoleHost.DefaultInitialSessionState = InitialSessionState.CreateDefault2();
-                        configuration = null;
-                    }
-                    else if (args.Length > 0 && !String.IsNullOrEmpty(args[0]) && args[0].Equals("-isswait", StringComparison.OrdinalIgnoreCase))
```
daxian-dbw: I just ran into a scenario where I need this flag to help debugging: I want to debug the module path resolution code in `ModuleIntrisics.cs` when starting powershell core in Windows PowerShell. I cannot just start/debug powershell core in VSCode, as I need to start it from Windows PowerShell, and `-isswait` is very helpful in this situation.

So I think we probably should keep this hidden flag for debugging purpose.

lzybkr: When I needed it, I needed it even earlier than this, so I never saw this code.

A slightly nicer version does something like:

```c#
while (!System.Diagnostics.Debugger.IsAttached) {
    Thread.Sleep(100);
}
System.Diagnostics.Debugger.Break();
```

This way, you don't need to press a key.

SteveL-MSFT: I'll put in the updated version Jason is suggesting

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Update-TypeData.cs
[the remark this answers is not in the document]
SteveL-MSFT: added back

SteveL-MSFT: repro'ing for me now, will add test

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
daxian-dbw: These following two methods seem OK to be removed:
```
private void RemoveProvider(ProviderConfigurationEntry entry)
private string GetProviderName(ProviderConfigurationEntry entry)
```

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
```diff
@@ -867,24 +838,25 @@ private void DoCloseHelper()
...
+                            break;
                         }
                     }
 
                     PSHostUserInterface host = executionContext.EngineHostInterface.UI;
-                    if (host != null)
+                    if (host != null && openRunspaces == false)
                     {
                         host.StopAllTranscribing();
+                        AmsiUtils.Uninitialize();
```
daxian-dbw: `AmsiUtils.Uninitialize()` should be called even if `host == null`.
It looks to me we don't need to change this method. Here is the logic of the existing method:
If there is still an open Runsapce, then this method will just return; if there is no open Runspace, then the method will continue: call `StopAllTranscribing` if `host != null`, and then call `AmsiUtils.Uninitialize()`.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ManagedEntrance.cs
```diff
@@ -55,54 +55,32 @@ public static int Start(string consoleFilePath, [MarshalAs(UnmanagedType.LPArray
...
             else
             {
-                //TODO : Deprecate RunspaceConfiguration and use InitialSessionState
-                configuration =
-                    RunspaceConfigForSingleShell.Create(consoleFilePath, out warning);
+                ConsoleHost.DefaultInitialSessionState = InitialSessionState.CreateDefault2();
             }
+#else
+            ConsoleHost.DefaultInitialSessionState = InitialSessionState.CreateDefault2();
+#endif
```
daxian-dbw: How about change it to the following to remove 2 duplicate lines of `InitialSessionState.CreateDefault2()`?
```
#if DEBUG
            if (args.Length > 0 && !String.IsNullOrEmpty(args[0]) && args[0].Equals("-isswait", StringComparison.OrdinalIgnoreCase))
            {
                Console.WriteLine("Attach the debugger to continue...");
                while (!System.Diagnostics.Debugger.IsAttached) {
                    Thread.Sleep(100);
                }
                System.Diagnostics.Debugger.Break();
            }
#endif
            ConsoleHost.DefaultInitialSessionState = InitialSessionState.CreateDefault2();
```

daxian-dbw (changes requested):

daxian-dbw:
There is also a comment at https://github.com/PowerShell/PowerShell/pull/4942#discussion_r145203074.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ManagedEntrance.cs
SteveL-MSFT: sounds good

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
SteveL-MSFT: will revert previous change

src/System.Management.Automation/engine/SessionStateProviderAPIs.cs
SteveL-MSFT: Will remove.

daxian-dbw (approved):
LGTM. Thanks @SteveL-MSFT for getting this done!

SteveL-MSFT:
New new Update-FormatData with invalid xml is causing a problem with PowerShellGet tests to fail.  I repro'd it with binaries built from master so it's not the result of these changes.  Created https://github.com/PowerShell/PowerShell/issues/5154 to track.

PR #4943, opened 2017-09-29
Autocorrected CRLF to LF

SteveL-MSFT:
Fixed Parser.Tests.ps1 to use regex:

```powershell
$result | should match "\r?\nfoo``u\{2195\}abc\r?\n"
```


daxian-dbw:
Add `?w=1` to the URL to see the diff with whitespace ignored --
 https://github.com/PowerShell/PowerShell/pull/4943/files?w=1. It looks good to me.

daxian-dbw (approved):

SteveL-MSFT:
@daxian-dbw that's an awesome tip!

test/powershell/Language/Parser/Parser.Tests.ps1
```diff
@@ -1,918 +1,918 @@
...
+
+        It 'Test that Unicode escape sequence in single quoted here string is not processed.' {
+            $result = ExecuteCommand @"
+@'
+
+foo``u{2195}abc
+
+'@
+"@
+            $result | should match "\r?\nfoo``u\{2195\}abc\r?\n"
```
daxian-dbw: Actually, one question about `\r?\n` here. Is it because we use CRLF on windows and LF on Unix? If so, can we do the following instead?
```
$newLine = [System.Environment]::NewLine
$result | should be "${newLine}foo``u\{2195\}abc${newLine}"
```

SteveL-MSFT: Yes, it is because git converts files to CRLF on Windows.  I like your suggestion and will make the change.

SteveL-MSFT:
@daxian-dbw I'm going to revert that last change, seems that on Windows you might have git set to not convert to CRLF which may be why it's failing on AppVeyor as it was just LF.

daxian-dbw (approved):

Issue #4953, opened 2017-09-29
Remove ApartmentState support


SteveL-MSFT:
ApartmentState STA/MTA is only used to interop with COM and not supported on non-Windows.  It also seems that dotnet can [marshal between STA and MTA](https://docs.microsoft.com/en-us/dotnet/framework/interop/interop-marshaling#marshaling_and_com_apartments) just fine so allowing the user to set it is just a perf optimization.

I propose we simplify our code by removing support.  The code base today already has ApartmentState commented out as we run MTA by default (as that is default for dotnet) so this is simply removing the `#if !CORECLR` conditional code.

SteveL-MSFT:
@PowerShell/powershell-committee discussed this and decided to leave the code as-is for 6.0.0 release

bergmeister:
@stevel-msft I want to add that removing support for appartmentstate would not have be good. I have seen cases where certain COM APIs can cause threading issues and lead to very bad performance, therefore the ability to optionally execute as STA is critical.
Therefore it is good that it was decided to not remove support for it. I just wanted to bring this up in case this issue gets re-considered.

PR #4958, opened 2017-09-30
fix `powershell -version` and help

SteveL-MSFT:
`PowerShell -v` behavior updated to align with other tools like git, curl, and bash where args after `-v` are silently ignored.

Built-in help updated to reflect changes we've made to the console host in PSCore6 removing unsupported parameters.

Doc update aligned with these changes (and general cleanup of parameters we don't support in PSCore6) https://github.com/PowerShell/PowerShell-Docs/pull/1707

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
+++ b/src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
@@ -273,6 +273,6 @@ EXAMPLES
     <value>Invalid argument '{0}', did you mean:</value>
   </data>
   <data name="DeprecatedVersionParameter" xml:space="preserve">
-    <value>Usage of '-Version {0}' is not supported. '-Version' currently only returns the current PowerShell version.</value>
+    <value>Unsupported value '{0}' passed to `-Version`. '-Version' only returns the current PowerShell version and must be used standalone.</value>
   </data>
 </root>
--- a/src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
+++ b/src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
@@ -129,89 +129,6 @@
   <data name="UnknownParameter" xml:space="preserve">
     <value>Unrecognized parameter: '{0}'.</value>
   </data>
-  <data name="DefaultHelp" xml:space="preserve">
-    <value>PowerShell[.exe] [-PSConsoleFile &lt;file&gt; | -Version &lt;version&gt;]
-    [-NoLogo] [-NoExit] [-NoProfile] [-NonInteractive] [-STA]
-    [-OutputFormat {Text | XML}] [-InputFormat {Text | XML}]
-    [-ConfigurationName &lt;string&gt;]
-    [-File fileName [arguments...]] [-ExecutionPolicy &lt;ExecutionPolicy&gt;]
-    [-Command { - | &lt;script-block&gt; [-args &lt;arg-array&gt;]
-                  | &lt;string&gt; [&lt;CommandParameters&gt;] } ]
-
-PowerShell[.exe] -Help | -? | /?
-
--PSConsoleFile
-    Loads the specified PowerShell console file. To create a console
-    file, use Export-Console in PowerShell.
-
--Version
-    Starts the specified version of PowerShell.
-
--NoLogo
-    Hides the copyright banner at startup.
-
--NoExit
-    Does not exit after running startup commands.
-
--NoProfile
-    Does not use the user profile.
-
--NonInteractive
-    Does not present an interactive prompt to the user.
-
--STA
-    Uses a single-threaded apartment for the execution thread.
-
--OutputFormat
-    Determines how output from PowerShell is formatted. Valid values
-    are "Text" (text strings) or "XML" (serialized CLIXML format).
-
--InputFormat
-    Describes the format of data sent to PowerShell. Valid values are
-    "Text" (text strings) or "XML" (serialized CLIXML format).
-
--ConfigurationName
-    Specifies a configuration endpoint in which PowerShell is run.
-    This can be any endpoint registered on the local machine including the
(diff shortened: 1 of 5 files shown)


daxian-dbw:
The same test didn't fail in AppVeyor. In AppVeyor, the failed test is `Version should write an error if a value is present`, see https://ci.appveyor.com/project/PowerShell/powershell/build/6.0.0-beta.7-5642#L384

On Travis, a different test failed: `Version should ignore other parameters`. See https://travis-ci.org/PowerShell/PowerShell/jobs/281514325#L1461

So probably we shouldn't just remove that test.

SteveL-MSFT:
We still need to remove this test as the change is to return an error if there's anything after `-version` whereas previously we silently ignored it.  I think the current behavior is correct.

SteveL-MSFT:
Need to fix the other test and also make a change to the `-v` handling.  I should have caught this in the code review.  Apologies.

test/powershell/Host/ConsoleHost.Tests.ps1
```diff
@@ -171,20 +171,11 @@ Describe "ConsoleHost unit tests" -tags "Feature" {
...
-        }
-
         It "-Version should write an error if a value is present" {
             $versionValue = "abrakadabra"
-            $tempFile = Join-Path $testdrive "expectedError.txt"
-            $observed = & $powershell -version $versionValue > $tempFile
-            $expectedError = (Get-Content $tempFile)[0]
-
-            $expectedError | Should Match $versionValue
+            $observed = & $powershell -version $versionValue 2>&1
```
iSazonov: 😕 I skipped [Feature] in previous PR. I saw "CI" on top the test file but we have "Feature" tests below. 
 _We need an idea how exclude this in future._
(I wonder that the test was passed locally.)

In CI log I see messages:
```
[00:08:17] Usage of '-Version -command' is not supported. '-Version' currently only returns the current PowerShell version.
[00:08:17] +
[00:08:17] Usage of '-Version abrakadabra' is not supported. '-Version' currently only returns the current PowerShell version.
```
The first message looks incorrect. Maybe: "Error in '-Version {0}'. Now '-Version' shows the current PowerShell version and don't accept an argument."

Also previous test helped us to catch the problem - so I prefer to don't remove it because it may test another code path in future.

iSazonov (commented):
Leave a comment

test/powershell/Host/ConsoleHost.Tests.ps1
SteveL-MSFT: The test I removed was when we supported:

```bash
poweshell -version -command 1+1
```

I think this syntax doesn't make sense as what it ended up doing was outputting the version string and ignoring everything after it.  It seems that this should have caught an unintentional user error which is why I removed that test since the new behavior prevents that old behavior.

I agree that the error message could be better particularly in the above scenario.  I'll update it.

SteveL-MSFT: I think I'll change it to

```none
Unsupported value '{0}' passed to `-Version`. '-Version' currently only returns the current PowerShell version and must be used standalone.
```

src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
```diff
@@ -273,6 +273,6 @@ EXAMPLES
     <value>Invalid argument '{0}', did you mean:</value>
   </data>
   <data name="DeprecatedVersionParameter" xml:space="preserve">
-    <value>Usage of '-Version {0}' is not supported. '-Version' currently only returns the current PowerShell version.</value>
+    <value>Unsupported value '{0}' passed to `-Version`. '-Version' only returns the current PowerShell version and must be used standalone.</value>
```
iSazonov: I still unlike this:
1. `-Version`. '-Version'  - looks bad.
2. Unsupported value - implies that there are supported values.

SteveL-MSFT: If we follow the example of some other tools like: git, curl, bash (although each of them use `--version`), they basically ignore anything after `--version` and simply output the version string.  Perhaps that's the right behavior.

SteveL-MSFT: For now, I think I'll follow the other tools and will update the documentation as well.

iSazonov: Initial idea for PowerShell was to inform users that we change -Version semantics. So I write information error and PowerShell version too.

SteveL-MSFT:
@iSazonov if you're fine with the changes, can you merge so we can unblock other PRs?  If you have specific concerns with the expected behavior, I think we can continue that discussion in a new issue.  Thanks!

iSazonov:
One concern is that we show full help in case of an error.
It's very annoying in interactive session if I typo - I'm forced to scroll back to see error message and lost my history before eyes. I'd prefer to see an error message and shot hint "Use powershell -? or -h to get more help" .

SteveL-MSFT:
@iSazonov you referring to existing behavior and not something I introduced, correct?

iSazonov:
Yes, I wanted to say that we need more review and cleanups in the file. I see:
1. #if !CORE
2. _ShowHelp=true
3. maybe formattings

SteveL-MSFT:
@iSazonov can submit a new PR?  the `#if CORECLR` I was going to take care of as part of https://github.com/PowerShell/PowerShell/issues/4953

iSazonov:
I'll try to review on next week.

iSazonov:
@SteveL-MSFT Please clarify. In #4834 we decided to write an error to inform users about changing -Version behavior. In the PR we remove the error. Should we write in #4834 that we reject the request now?

SteveL-MSFT:
@iSazonov added comment and closed https://github.com/PowerShell/PowerShell/issues/4834

1. we should be consistent and, in general, use existing norms like `bash`
2. we should not output help either
3. agree with the proposed changes
4. absolutely!

PR #4959, opened 2017-09-30
enable support of folders and files with colon in name on Unix

### LINKED ISSUE #4889: Linux: cd and find commands fail when a colon is in the directory or file name
The /sys and /proc directories have sub-directories with colons in the names.  Commands like cd and find fail when trying to access directories with ':'.

PS /> cd /sys/devices/pci0000:00/                                                                                                                                                            
cd : Cannot find drive. A drive with the name 'sys/devices/pci0000' does not exist.
At line:1 char:1
+ cd /sys/devices/pci0000:00/
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (sys/devices/pci0000:String) [Set-Location], DriveNotFoundException
    + FullyQualifiedErrorId : DriveNotFound,Microsoft.PowerShell.Commands.SetLocationCommand
 

PS /> find /sys -name '*:*'                                                                                                                                                                  
Cannot find drive. A drive with the name '*' does not exist.
At line:1 char:1
+ find /sys -name '*:*'
+ ~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (*:String) [], DriveNotFoundException
    + FullyQualifiedErrorId : DriveNotFound
 
The ls command does work.
PS /> ls /sys/devices/pci0000:00/                                                                                                                                                            
0000:00:00.0  0000:00:16.0  0000:00:19.0  0000:00:1b.0  0000:00:1c.1  0000:00:1c.3  0000:00:1d.0  0000:00:1f.2  firmware_node  power       uevent
0000:00:02.0  0000:00:16.3  0000:00:1a.0  0000:00:1c.0  0000:00:1c.2  0000:00:1c.5  0000:00:1f.0  0000:00:1f.3  pci_bus        SMO8800:00

------------------

```powershell

```

Expected behavior
-----------------
For files and directories with ':' to be processed correctly, and not as drive names.

Actual behavior
---------------

```none

```

Environment data
----------------

OS                             Linux 4.4.0-92-generic #115-Ubuntu SMP Thu Aug 10 09:04:33 UTC 2017                                                                                          

SteveL-MSFT:
allow colons if it shows up after a path separator

fix https://github.com/PowerShell/PowerShell/issues/4889

### DIFF
--- a/src/System.Management.Automation/namespaces/LocationGlobber.cs
+++ b/src/System.Management.Automation/namespaces/LocationGlobber.cs
@@ -1665,13 +1665,15 @@ namespace System.Management.Automation
 
                 if (index > 0)
                 {
-                    int separator = path.IndexOf(StringLiterals.DefaultPathSeparator, 0, index);
+                    // see if there are any path separators before the colon which would mean the
+                    // colon is part of a file or folder name and not a drive: ./foo:bar vs foo:bar
+                    int separator = path.IndexOf(StringLiterals.DefaultPathSeparator, 0, index-1);
                     if (separator == -1)
                     {
-                        separator = path.IndexOf(StringLiterals.AlternatePathSeparator, 0, index);
+                        separator = path.IndexOf(StringLiterals.AlternatePathSeparator, 0, index-1);
                     }
                     if (separator == -1 || index < separator)
-                    {                        
+                    {
                         // We must have a drive specified
                         result = true;
                     }
--- a/src/System.Management.Automation/namespaces/LocationGlobber.cs
+++ b/src/System.Management.Automation/namespaces/LocationGlobber.cs
@@ -1649,7 +1649,6 @@ namespace System.Management.Automation
                 }
 
                 int index = path.IndexOf(":", StringComparison.Ordinal);
-                int separator = path.IndexOf(StringLiterals.DefaultPathSeparatorString, StringComparison.Ordinal);
 
                 if (index == -1)
                 {
@@ -1664,11 +1663,18 @@ namespace System.Management.Automation
                 // must assume that it is part of the path, and not
                 // delimiting the drive name.
 
-                if (index > 0 && index < separator)
+                if (index > 0)
                 {
-                    // We must have a drive specified
-
-                    result = true;
+                    int separator = path.IndexOf(StringLiterals.DefaultPathSeparator, 0, index);
+                    if (separator == -1)
+                    {
+                        separator = path.IndexOf(StringLiterals.AlternatePathSeparator, 0, index);
+                    }
+                    if (separator == -1 || index < separator)
+                    {                        
+                        // We must have a drive specified
+                        result = true;
+                    }
                 }
             } while (false);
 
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/namespaces/LocationGlobber.cs
```diff
@@ -1649,6 +1649,7 @@ internal static bool IsAbsolutePath(string path)
                 }
 
                 int index = path.IndexOf(":", StringComparison.Ordinal);
+                int separator = path.IndexOf(StringLiterals.DefaultPathSeparatorString, StringComparison.Ordinal);
```
lzybkr: This is inefficient to do unconditionally for an uncommon case, but it also doesn't fully fix the problem. Is it not possible to check if it looks like a drive, and if not, then it moist be a file? There are downsides to that approach though, so if it was rejected, I think it's worth noting why in the code.

SteveL-MSFT: The problem with the current code is that it always checks for the colon you see above and if it's there, it assumes it's a drive.  What I'm doing here is what you are saying.  If the colon is after a path separator, it doesn't look like a drive.  If you mean to see if the left of the colon is a drive, wouldn't that be a bigger perf hit?  I could certainly only look for the separator only if there is a colon.

SteveL-MSFT: With this change this still won't work:

```powershell
1+1 > foo:bar.txt
```

instead you have to do:

```powershell
1+1 > ./foo:bar.txt
```

Which seems fine to me

SteveL-MSFT:
CI failures due to https://github.com/PowerShell/PowerShell/pull/4958

src/System.Management.Automation/namespaces/LocationGlobber.cs
lzybkr: At a minimum, yes, only look for the separator if there is a colon.

But the inconsistency of requiring a separator only if there is a colon - that seems silly and not easy to discover.

I guess I was suggesting to assume it's a drive (maybe not if there is a separator) because we're doing that already, and if it's not a drive, then don't error out, treat it as a filesystem path.

Of course this does introduce an ambiguity:

```powershell
Set-Content -Path env:abc -Value 123
```

This sets the environment variable, not the file content. In cases like this, you'd have to use a separator. This example might seem not interesting, but something like this is more problematic:

```powershell
Set-Content -Path "${prefix}:${suffix}" -Value 123
```

SteveL-MSFT: Personally, I like having the distinction between a drive and a file with the separator.

SteveL-MSFT:
@lzybkr are you ok with the current changes?  would like to get it in for beta.8 if possible

src/System.Management.Automation/namespaces/LocationGlobber.cs
```diff
@@ -1665,9 +1665,16 @@ internal static bool IsAbsolutePath(string path)
 
                 if (index > 0)
                 {
-                    // We must have a drive specified
-
-                    result = true;
+                    int separator = path.IndexOf(StringLiterals.DefaultPathSeparator, 0, index);
```
mirichmo: Based on the [documentation](https://msdn.microsoft.com/en-us/library/ms131434(v=vs.110).aspx) this does not look correct to me. It searches the string starting from the beginning though index count of characters for the matching separator. "C:\something.txt" Will always fail because the separator never matches the first character of the string ("C").

I think you might need:
```c#
separator = path.IndexOf(StringLiterals.AlternatePathSeparator, index, 1); 
```
This checks the character immediately after the index. But if you are going to do that, you might as well directly test index+1.

mirichmo (changes requested):

src/System.Management.Automation/namespaces/LocationGlobber.cs
SteveL-MSFT: `index` points to location of the colon.  So this test is to determine if there is a path separator between the beginning and the colon.  So it should match `./foo:bar` but not match `foo:bar` where the former is a file/directory and the latter is a drive path.  One optimization I can see is I only need to search up to `index-1` since I already know there is a colon at `index`.

SteveL-MSFT:
@mirichmo can you review again?

mirichmo (approved):

PR #4963, opened 2017-09-30
correct passing of parameter so args show up as args instead of input

### LINKED ISSUE #4923: CommandInvocationIntrinsics.InvokeScript bind arguments to $input instead of $args
[This lines](https://github.com/PowerShell/PowerShell/blob/master@%7B2017-09-26%7D/src/System.Management.Automation/engine/MshCmdlet.cs#L687-L690):
```csharp
public Collection<PSObject> InvokeScript(string script, params object[] args)
{
    return InvokeScript(script, true, PipelineResultTypes.None, args);
}
``` 
Pass `args` parameter into `input` parameter of called function, thus mixing them up.

Steps to reproduce
------------------

```powershell
$ExecutionContext.SessionState.InvokeCommand.InvokeScript('"`$args:($args); `$input:($input)"', 1, 2, 3)
```

Expected behavior
-----------------

```none
$args:(1 2 3); $input:()
```

Actual behavior
---------------

```none
$args:(); $input:(1 2 3)
```

Environment data
----------------

Tested in v2.0 and v5.1.15063.608

SteveL-MSFT:
incorrect position of a parameter which resulted in the args passed as input instead of as args

fix https://github.com/PowerShell/PowerShell/issues/4923

### DIFF
--- a/src/System.Management.Automation/engine/MshCmdlet.cs
+++ b/src/System.Management.Automation/engine/MshCmdlet.cs
@@ -686,7 +686,7 @@ namespace System.Management.Automation
         /// <exception cref="FlowControlException"></exception>
         public Collection<PSObject> InvokeScript(string script, params object[] args)
         {
-            return InvokeScript(script, true, PipelineResultTypes.None, args);
+            return InvokeScript(script, true, PipelineResultTypes.None, null, args);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


SteveL-MSFT:
CI failures waiting on https://github.com/PowerShell/PowerShell/pull/4958

daxian-dbw (approved):

Issue #4358, opened 2017-07-27
Command line parsing bug: spaces and trailing backslash


akervinen:
This has already been posted on UserVoice ([here](https://windowsserver.uservoice.com/forums/301869-powershell/suggestions/20102977-command-line-parsing-bug-spaces-and-trailing-slas)), but appears to still happen in 6.0.

Steps to reproduce
------------------

Using any .exe capable of outputting its command line arguments:
```powershell
PS X:\scratch> .\ps-args-test.exe .\test1\ '.\test 2\' .\test1\
```
Auto-completion uses this format with a trailing backslash, so it's expected to work correctly.

Expected behavior
-----------------

```none
.\test1\
.\test 2\
.\test1\
```

Actual behavior
---------------

```none
.\test1\
.\test 2" .\test1\
```

Environment data
----------------

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.4
OS                             Microsoft Windows 10.0.14393
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

mklement0:
@iSazonov: 

This is only indirectly related to IntelliSense (which _suggests_ passing the arguments like that).

The real issue is how the command line is rebuilt _behind the scenes_ by PowerShell:

Argument list `.\test1\ '.\test 2\' .\test1\` is translated to `.\test1\ ".\test 2\" .\test1\` behind the scenes, which any external utility that uses `\` as the escape character (most of them) parses as reflected in the initial post, causing the initial argument partitioning to be misinterpreted:

* The 2nd `"` in `".\test 2\"`, due to being preceded by `\` is interpreted as an _escaped_ `"`, causing the remainder of the string - despite a then-missing closing `"` to be interpreted as part of the same argument.

---

The - cumbersome - **workaround** on **Windows** is to add an extra `\`, though, given that PowerShell does re-quoting behind the scenes, the **fix** is to let _PowerShell_  do  that _automatically_:

    > .\ps-args-test.exe .\test1\ '.\test 2\\' .\test1\  # Note the '\\'

[...]

PR #4965, opened 2017-10-01
properly escape trailing backslash so that it doesn't end up escaping the quotes

SteveL-MSFT:
When we create the arg string, an arg of `.\test 1\` ends up being `".\test 1\"` which is stored in StartInfo.Arguments which then corefx [passes as-is](https://github.com/dotnet/corefx/blob/93c5f5d795a464c18d96da47bcb5b2f7f571d26a/src/System.Diagnostics.Process/src/System/Diagnostics/Process.Win32.cs#L48) to SHELLEXECUTEINFO.

The native command receives the arg as `.\test 1"` as the last `\"` is treated as escaping the quotes.  Fix is to add an extra backslash to escape the last enclosing quote.

Fix https://github.com/PowerShell/PowerShell/issues/4358

### DIFF
--- a/src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
+++ b/src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
@@ -197,25 +197,11 @@ namespace System.Management.Automation
                         {
                             _arguments.Append('"');
                             // need to escape all trailing backslashes so the native command receives it correctly
-                            if (arg.EndsWith('\\'))
+                            // according to http://www.daviddeley.com/autohotkey/parameters/parameters.htm#WINCRULESDOC
+                            _arguments.Append(arg);
+                            for (int i = arg.Length-1; i >= 0 && arg[i] == '\\'; i--)
                             {
-                                // find index of last character that isn't a backslash
-                                var chars = arg.ToCharArray();
-                                int index = 0;
-                                for (int i = chars.Length-1; i > 0; i--)
-                                {
-                                    if (chars[i] != '\\')
-                                    {
-                                        index = i;
-                                        break;
-                                    }
-                                }
-                                _arguments.Append(arg.Substring(0, index));
-                                _arguments.Append(arg.Substring(index).Replace("\\","\\\\"));
-                            }
-                            else
-                            {
-                                _arguments.Append(arg);
+                                _arguments.Append('\\');
                             }
                             _arguments.Append('"');
                         }
--- a/src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
+++ b/src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
@@ -196,16 +196,28 @@ namespace System.Management.Automation
                         if (NeedQuotes(arg))
                         {
                             _arguments.Append('"');
-                            _arguments.Append(arg);
-                            // if trailing backslash, we need to escape it so that it doesn't escape the quotes
+                            // need to escape all trailing backslashes so the native command receives it correctly
                             if (arg.EndsWith('\\'))
                             {
-                                _arguments.Append("\\\"");
+                                // find index of last character that isn't a backslash
+                                var chars = arg.ToCharArray();
+                                int index = 0;
+                                for (int i = chars.Length-1; i > 0; i--)
+                                {
+                                    if (chars[i] != '\\')
+                                    {
+                                        index = i;
+                                        break;
+                                    }
+                                }
+                                _arguments.Append(arg.Substring(0, index));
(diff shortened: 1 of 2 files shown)


PetSerAl:
Since not every application use `\` as escapement it will also affect behavior of that programs. For example:
```powershell
cmd /c 'echo \'
```
will print `\\` instead of `\`.

Also you should not only double one `\` at the end, but whole sequence of `\` in the end. For example:
```powershell
cmd /c 'echo \'
```
will print `\\\\`, while it should be `\\\\\\` to be properly recognized as three literal `\`.

test/powershell/Language/Scripting/NativeExecution/NativeCommandArguments.Tests.ps1
```diff
@@ -31,4 +31,11 @@ Describe "Native Command Arguments" -tags "CI" {
         $lines[0] | Should Be 'Arg 0 is <a"b c"d>'
         $lines[1] | Should Be 'Arg 1 is <a"b c"d>'
     }
+
+    It "Should correctly quote paths with spaces" {
+        $lines = testexe -echoargs '.\test 1\' '.\test 2\'
```
iSazonov: Should we check double quotes?

SteveL-MSFT: Will add test

SteveL-MSFT:
@PetSerAl for `cmd echo`, perhaps the only thing we can do is suggest using `cmd --% /c echo \`

PetSerAl:
How about multiple trailing `\`?:
```powershell
testexe -echoargs '.\test 1\\' ".\test 2\\"
```

SteveL-MSFT:
@PetSerAl I'll work on that and add a test

SteveL-MSFT:
Apparently these strange trailing backslash and double quotes rules are [documented](http://www.daviddeley.com/autohotkey/parameters/parameters.htm#WINCRULESDOC)

src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
```diff
@@ -196,16 +196,28 @@ private void appendOneNativeArgument(ExecutionContext context, object obj, char
                         if (NeedQuotes(arg))
                         {
```
PetSerAl: How about:
```csharp
_arguments.Append('"'); 
_arguments.Append(arg); 
for (int i = arg.Length-1; i >= 0 && arg[i] == '\\'; i--)
{
    _arguments.Append('\\');
}
_arguments.Append('"'); 
```
?

SteveL-MSFT: That's much better than my attempt.  Will change.

src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
```diff
@@ -197,25 +197,11 @@ private void appendOneNativeArgument(ExecutionContext context, object obj, char
...
-                                _arguments.Append(arg.Substring(0, index));
-                                _arguments.Append(arg.Substring(index).Replace("\\","\\\\"));
-                            }
-                            else
+                            // according to http://www.daviddeley.com/autohotkey/parameters/parameters.htm#WINCRULESDOC
+                            _arguments.Append(arg);
+                            for (int i = arg.Length-1; i >= 0 && arg[i] == '\\'; i--)
                             {
-                                _arguments.Append(arg);
+                                _arguments.Append('\\');
```
iSazonov: Shouldn't it be "\\\\"?

PetSerAl: No, unless you want to triplicate trailing `\`. `arg` already have one copy of `\`. This just adding second copy on top of it.

iSazonov: Thanks for clarify!

iSazonov (approved):

iSazonov:
@daxian-dbw @lzybkr Could you please review?

iSazonov:
@daxian-dbw @lzybkr Could you please review?

SteveL-MSFT:
@daxian-dbw is in training today, @anmenaga can you review?

daxian-dbw:
Sorry that I missed the the first "@mention" notification. I will take a look later today.
I wish @mklement0 can review this if possible, as he has a better understanding of the native argument parsing user experience in PowerShell.

daxian-dbw (approved):

anmenaga (approved):
LGTM

Issue #2358, opened 2016-09-26
Add Standard Deviation to Measure-Object output


powercode:
Standard Deviation is a measure that is used to quantify the amount of variation or dispersion of a set of data values. A low standard deviation indicates that the data points tend to be close to the mean (also called the expected value) of the set, while a high standard deviation indicates that the data points are spread out over a wider range of values.

This is a commonly used to get an insight about a set of values, and would be easy to implement as an addition to Measure-Object.

iSazonov:
This requires a double iterate: first, to calculate the mean value, the second, to calculate the variance.
Now Measure-Object don't collect values for second iterate. It is too complicate the cmdlet.
Also there's the challenge of the accuracy of the calculations for real data.
I believe this is a task for specialized statistical module.

powercode:
@iSazonov I'd turn the argument around: Give me something easy and useful as a build in tool, and write a specialized statistical module if you want higher accuracy or better performance.

vors:
Fixed in #6238

PR #4969, opened 2017-10-01
First pass at adding std deviation to measure-object

dfinke:
Fix https://github.com/PowerShell/PowerShell/issues/2358

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
@@ -35,7 +35,7 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         public GenericMeasureInfo()
         {
-            Average = Sum = Maximum = Minimum = null;
+            Average = Sum = Maximum = Minimum = StdDeviation = null;
         }
 
         /// <summary>
@@ -72,6 +72,13 @@ namespace Microsoft.PowerShell.Commands
         ///
         /// </summary>
         public double? Minimum { get; set; }
+
+        /// <summary>
+        ///
+        /// The StdDeviation of property values
+        ///
+        /// </summary>
+        public double? StdDeviation { get; set; }
     }
 
     /// <summary>
@@ -91,7 +98,7 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         public GenericObjectMeasureInfo()
         {
-            Average = Sum = null;
+            Average = Sum = StdDeviation = null;
             Maximum = Minimum = null;
         }
 
@@ -129,6 +136,13 @@ namespace Microsoft.PowerShell.Commands
         ///
         /// </summary>
         public object Minimum { get; set; }
+
+        /// <summary>
+        ///
+        /// The StdDeviation of property values
+        ///
+        /// </summary>
+        public double? StdDeviation { get; set; }
     }
 
 
@@ -227,6 +241,8 @@ namespace Microsoft.PowerShell.Commands
 
             // Generic/Numeric statistics
             internal double sum = 0.0;
+            internal double stdDeviation = 0.0;
+            internal List<double> stdDeviationNumbers = new List<double>();
             internal object max = null;
             internal object min = null;
 
@@ -264,6 +280,25 @@ namespace Microsoft.PowerShell.Commands
         public string[] Property { get; set; } = null;
 
         #endregion Common parameters in both sets
+
+        /// <summary>
+        /// Set to true is StdDeviation is to be returned
+        /// </summary>
+        /// <value></value>
+        [Parameter(ParameterSetName = GenericParameterSet)]
+        public SwitchParameter StdDeviation
+        {
+            get
+            {
+                return _measureStdDeviation;
+            }
+            set
+            {
+                _measureStdDeviation = value;
+            }
+        }
+
+        private bool _measureStdDeviation;
 
         /// <summary>
         /// Set to true is Sum is to be returned
@@ -713,6 +748,8 @@ namespace Microsoft.PowerShell.Commands
         {
             if (_measureSum || _measureAverage)
                 stat.sum += numValue;
+            if (_measureStdDeviation)
+                stat.stdDeviationNumbers.Add(numValue);
         }
 
         /// <summary>
@@ -793,6 +830,7 @@ namespace Microsoft.PowerShell.Commands
         {
             double? sum = null;
             double? average = null;
+            double? stdDeviation = null;
             object max = null;
             object min = null;
 
@@ -800,8 +838,21 @@ namespace Microsoft.PowerShell.Commands
             {
                 if (_measureSum)
                     sum = stat.sum;
+
                 if (_measureAverage && stat.count > 0)
                     average = stat.sum / stat.count;
+
+                if (_measureStdDeviation && _measureAverage && stat.count > 0)
+                {
+                    var popdev = 0.0;
+
+                    foreach (double n in stat.stdDeviationNumbers)
+                    {
+                        popdev += Math.Pow((n - (double)average), 2);
+                    }
+
+                    stdDeviation = Math.Round(Math.Sqrt(popdev / (stat.stdDeviationNumbers.Count - 1)), 4);
+                }
             }
 
             if (_measureMax)
@@ -838,6 +889,7 @@ namespace Microsoft.PowerShell.Commands
                 gmi.Count = stat.count;
                 gmi.Sum = sum;
                 gmi.Average = average;
+                gmi.StdDeviation = stdDeviation;
                 if (null != max)
                 {
                     gmi.Maximum = (double)max;
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
@@ -843,8 +843,14 @@ namespace Microsoft.PowerShell.Commands
                     average = stat.sum / stat.count;
 
                 if(_measureStdDeviation && !_measureAverage) {
-                    var message = "StdDeviation was requested and requires the average to be calculated, please add the -Average switch";
-                    throw new ArgumentException(message);
+                    ErrorRecord errorRecord = new ErrorRecord(
+                        PSTraceSource.NewArgumentException("Average"),
+                        "AverageSwitchNotSet",
+                        ErrorCategory.InvalidArgument,
+                        null);
+
+                    errorRecord.ErrorDetails = new ErrorDetails(this, "MeasureObjectStrings", "AverageSwitchNotSet", "Average");
+                    WriteError(errorRecord);
                 }
 
                 if (_measureStdDeviation && _measureAverage && stat.count > 0)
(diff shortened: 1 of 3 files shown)


SteveL-MSFT:
Added the issue

iSazonov:
@dfinke Thanks for your contribution!

The tests file contains multiple formatting issue. We should split code and formatting changes. So I open #4972 - please rebase after it will be merged.

We should think about optimizations. Currently we do multiple checks `_measureSum`, `_measureAverage`, `_measureStdDeviation` - can we refactor and optimize?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
```diff
@@ -72,6 +72,13 @@ public GenericMeasureInfo()
         ///
         /// </summary>
         public double? Minimum { get; set; }
+
+        /// <summary>
+        ///
+        /// The StdDeviation of property values
```
SteveL-MSFT: For the comments, we should spell out `Standard Deviation`

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
```diff
@@ -129,6 +136,13 @@ public GenericObjectMeasureInfo()
         ///
         /// </summary>
         public object Minimum { get; set; }
+
+        /// <summary>
+        ///
+        /// The StdDeviation of property values
```
SteveL-MSFT: For the comments, we should spell out `Standard Deviation`

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
```diff
@@ -266,6 +282,25 @@ public MeasureObjectCommand()
...
+        [Parameter(ParameterSetName = GenericParameterSet)]
+        public SwitchParameter StdDeviation
+        {
+            get
+            {
+                return _measureStdDeviation;
+            }
+            set
+            {
+                _measureStdDeviation = value;
```
SteveL-MSFT: stddeviation requires _measureAverage, so I think we should set _measureAverage = true.  If _measureAverage = false, perhaps we should error.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
```diff
@@ -793,15 +830,29 @@ private MeasureInfo CreateGenericMeasureInfo(Statistics stat, bool shouldUseGene
...
                 if (_measureAverage && stat.count > 0)
                     average = stat.sum / stat.count;
+
+                if (_measureStdDeviation && _measureAverage && stat.count > 0)
+                {
+                    var popdev = 0.0;
+
+                    foreach (double n in stat.stdDeviationNumbers)
+                    {
+                        popdev += Math.Pow((n - (double)average), 2);
```
SteveL-MSFT: My understanding is that A*A is faster than Math.Pow(A,2)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
```diff
@@ -793,15 +830,29 @@ private MeasureInfo CreateGenericMeasureInfo(Statistics stat, bool shouldUseGene
...
             {
                 if (_measureSum)
                     sum = stat.sum;
+
                 if (_measureAverage && stat.count > 0)
                     average = stat.sum / stat.count;
+
+                if (_measureStdDeviation && _measureAverage && stat.count > 0)
+                {
+                    var popdev = 0.0;
```
SteveL-MSFT: I think `sumOfDerivation` is more descriptive than `popdev`

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
dfinke: Makes sense. Where should it throw, in `CreateGenericMeasureInfo`?

dfinke:
I noticed that the property `StdDeviation` doesn't show up in `help measure`. Do the maml files need to get updated and rebuilt? I tweaked one of the maml files, but no luck.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
SteveL-MSFT: Yes, I think that's the right place.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
```diff
@@ -793,15 +830,35 @@ private MeasureInfo CreateGenericMeasureInfo(Statistics stat, bool shouldUseGene
...
             if (!_nonNumericError)
             {
                 if (_measureSum)
                     sum = stat.sum;
+
                 if (_measureAverage && stat.count > 0)
                     average = stat.sum / stat.count;
+
+                if(_measureStdDeviation && !_measureAverage) {
+                    var message = "StdDeviation was requested and requires the average to be calculated, please add the -Average switch";
```
SteveL-MSFT: messages should be in the corresponding .resx file.  See line 537 as an example.

SteveL-MSFT:
@dfinke the MAML files are generated.  You need to update the help found in https://github.com/powershell/powershell-docs

src/Microsoft.PowerShell.Commands.Utility/resources/MeasureObjectStrings.resx
```diff
@@ -126,4 +126,7 @@
   <data name="NonNumericInputObject" xml:space="preserve">
     <value>Input object "{0}" is not numeric.</value>
   </data>
+  <data name="AverageSwitchNotSet" xml:space="preserve">
+    <value>StdDeviation was requested and requires the average to be calculated, please add the -Average switch.</value>
```
SteveL-MSFT: I think we should default `-Average` to be set if `-StdDeviation` is set, I think the error only applies if someone explicitly used `-Average:$false`.  So I think the error message should be a bit more specific:

```none
StdDeviation was requested and requires the average to be calculated, however '-Average' was set to $false.
```

SteveL-MSFT (changes requested):

iSazonov:
@dfinke #4972 was merged.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
```diff
@@ -793,15 +830,41 @@ private MeasureInfo CreateGenericMeasureInfo(Statistics stat, bool shouldUseGene
...
+                    WriteError(errorRecord);
+                }
+
+                if (_measureStdDeviation && _measureAverage && stat.count > 0)
+                {
+                    var sumOfDerivation = 0.0;
+
+                    foreach (double n in stat.stdDeviationNumbers)
+                    {
+                        var m = n - (double)average;
```
iSazonov: Should we move the cast out of foreach ?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
```diff
@@ -793,15 +830,41 @@ private MeasureInfo CreateGenericMeasureInfo(Statistics stat, bool shouldUseGene
...
+                        sumOfDerivation += m * m;
+                    }
+
+                    stdDeviation = Math.Round(Math.Sqrt(sumOfDerivation / (stat.stdDeviationNumbers.Count - 1)), 4);
```
iSazonov: Please clarify why we round? I believe it must be a double.

lzybkr:
Maybe not for this PR - but the output of `Measure-Object` is not great - most of the properties are `$null`, and adding a new property will make it slightly worse.

I was going to propose something like what I use in my profile (see [here](https://gist.github.com/lzybkr/eddc79020dfddc821782cfce8b671f05)), but that is messy and not quite right (doesn't support strict mode properly), so I might actually want a new formatting feature to make it simpler to specify optional properties, something like:
```
<ListItem>
    <PropertyName WhenPresent="true">Average</PropertyName>
</ListItem>
```

This could test if the property is present and non-null before selecting the item.

iSazonov:
Can we create a custom PSObject with non-null properties in the cmdlet?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
SteveL-MSFT: Thinking about this, it seems that `-Average:$false -StdDeviation` should not be a valid combination and perhaps a better way to solve this is to make them different parameter sets.  The problem here is that depending on which one is set first, you may get unintended behavior `-StdDeviation -Average:$false` since the code here overwrites `_measureAverage`.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
```diff
@@ -793,15 +830,41 @@ private MeasureInfo CreateGenericMeasureInfo(Statistics stat, bool shouldUseGene
...
 
             if (!_nonNumericError)
             {
                 if (_measureSum)
                     sum = stat.sum;
+
                 if (_measureAverage && stat.count > 0)
                     average = stat.sum / stat.count;
+
+                if(_measureStdDeviation && !_measureAverage) {
```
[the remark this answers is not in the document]
SteveL-MSFT: With different parameter sets, this code is not needed.

SteveL-MSFT (changes requested):

dfinke:
Another option might be to calculate average when std deviation is specified. Or use the already calculated average if the `-Average` switch was passed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
iSazonov: We need new flag to process a sum. We can set it in BeginProcessing.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Measure-Object.cs
dfinke: I rounded for no particular reason. I can remove it.

dfinke:
I'm confused. I checked the type etc. In the `.cs` file I took out the `Math.Round`.

image

iSazonov:
@dfinke Please remove ` Math.Round ` because we should return honest double - the round is only formatting. We have separate formatting engine as @lzybkr mentioned above. If we want better formatting we should make it there. I believe we can skip this in the PR.

Please resolve merge conflict and address feedbacks.

iSazonov:
@dfinke Could you please continue?

dfinke:
@iSazonov sorry, my free time got pulled in a different direction. I also got blocked testing. Returning the double from the cmdlet and testing it in Pester. The number looked correct from the cmdlet and in Pester to several decimal places. 

I won't be able to get back to this in the near future.

iSazonov:
@dfinke  Very sorry. Welcome back with new ideas and contributions!

PR #4970, opened 2017-10-01
Add CertificateValidationScript Parameter to Web Cmdlets

### LINKED ISSUE #4899: Add User Supplied ServerCertificateCustomValidationCallback Support to Web Cmdlets
# Problem

Currently users only have 2 options for dealing with Server SSL/TLS Certificates with `Invoke-WebRequest` and `Invoke-RestMethod`: the default validation and to skip validation. Some scenarios warrant tighter security on web requests where a certificate is not fully trusted by the host environment but is known to be trusted by the user. This could include internal web APIs that use a self signed certificate with a specific thumbprint or from a known CA that is not trusted by the host. Or if a user wishes to ensure a certain CA/Thumbprint/Subject is blocked (a known bad actor).

Also `[System.Net.ServicePointManager]::ServerCertificateValidationCallback` has no effect in Core and `HttpClient` only uses the settings provided by `HttpClientHandler`.

# Proposal

Add a parameter of type `Func<HttpRequestMessage,X509Certificate2,X509Chain,SslPolicyErrors,Boolean>` that accepts a ScriptBlock to both Web Cmdlets. This is to be set on `HttpClientHandler.ServerCertificateCustomValidationCallback`. The `-SkipCertificateCheck` would have priority, meaning if both were supplied either a parameter exception is thrown or `-SkipCertificateCheck` would be applied and the callback ignored. 

The Parameter will be named `CertificateValidationScript`

# For consideration

* What to name this parameter?
* Throw or no throw on clash with `-SkipCertificateCheck`?

markekraus:
closes #4899

This adds a `-CertificateValidationScript` parameter to `Invoke-RestMethod` and `Invoke-WebRequest` which allows the user to supply a custom validation script as a PowerShell `ScriptBlock` to validate Server Certificates for HTTPS connections.

The ScriptBlock also has 4 automatic variables created which allows the user to use either `$args`, `Param()`, or the automatic variables to inspect the `HttpRequestMessage` being sent to the server, `X509Certificate2` certificate returned by the server, the `X509Chain` certificate chain of the server certificate, and the `SslPolicyErrors`.  The ScriptBlock is converted to a Closure and is wrapped in a `Func` delegate which is used as the `HttpClientHandler.ServerCertificateCustomValidationCallback`. The ScriptBlock has access to the calling scope and any ScriptBlock exceptions are treated as a certificate failure.

# Documentation Needed
This new parameter will need to be documented and an example added. The documentation will need to note that the `$using:` scope is not supported and will result in a certificate failure. it will also need to note that if the site redirects, the same validation script will apply for all hops. Finally, it will also need to state that `-SkipCertificateCheck` will override anything supplied in `-CertificateValidationScript`.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
@@ -165,7 +165,7 @@ namespace Microsoft.PowerShell.Commands
         public virtual SwitchParameter SkipCertificateCheck { get; set; }
 
         /// <summary>
-        /// gets or sets the CertificateValidationScript property
+        /// gets or sets the CertificateValidationScript property. This will be ignored if SkipCertificateCheck is set.
         /// </summary>
         [Parameter]
         public virtual ScriptBlock CertificateValidationScript { get; set; }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
@@ -165,6 +165,12 @@ namespace Microsoft.PowerShell.Commands
         public virtual SwitchParameter SkipCertificateCheck { get; set; }
 
         /// <summary>
+        /// gets or sets the CertificateValidationScript property
+        /// </summary>
+        [Parameter]
+        public virtual ScriptBlock CertificateValidationScript { get; set; }
+
+        /// <summary>
         /// Gets or sets the TLS/SSL protocol used by the Web Cmdlet
         /// </summary>
         [Parameter]
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
@@ -182,17 +182,28 @@ namespace Microsoft.PowerShell.Commands
                 Func<HttpRequestMessage,X509Certificate2,X509Chain,SslPolicyErrors,bool> validationCallBackWrapper = 
                     delegate(HttpRequestMessage httpRequestMessage, X509Certificate2 x509Certificate2, X509Chain x509Chain, SslPolicyErrors sslPolicyErrors)
                     {
+                        InitialSessionState iss = InitialSessionState.CreateDefault();
+                        
+                        // Add $_ variable with the Delegate parameters as members
+                        PSObject dollarUnderbar = new PSObject();
+                        dollarUnderbar.Members.Add(new PSNoteProperty("Request", httpRequestMessage));
+                        dollarUnderbar.Members.Add(new PSNoteProperty("Certificate", x509Certificate2));
+                        dollarUnderbar.Members.Add(new PSNoteProperty("CertificateChain", x509Chain));
+                        dollarUnderbar.Members.Add(new PSNoteProperty("SslErrors", sslPolicyErrors));
+                        iss.Variables.Add(new SessionStateVariableEntry(name: "_", value: dollarUnderbar, description: String.Empty));
+
                         Boolean result = false;
                         try
                         {
-                            result = LanguagePrimitives.IsTrue(
-                                 CertificateValidationScript.InvokeReturnAsIs(
-                                    httpRequestMessage, 
-                                    x509Certificate2, 
-                                    x509Chain, 
-                                    sslPolicyErrors
-                                )
-                            );
+                            using (Runspace rs = RunspaceFactory.CreateRunspace(iss))
+                            using (var ps = System.Management.Automation.PowerShell.Create())
+                            {
+                                ps.Runspace = rs;
+                                rs.Open();
+                                ps.AddScript(CertificateValidationScript.ToString());
+
+                                result = LanguagePrimitives.IsTrue(ps.Invoke().First());
+                            }
                         }
                         catch // Treat all exceptions as Certificate failures.
                         {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
@@ -4,15 +4,18 @@ Copyright (c) Microsoft Corporation. All rights reserved.
 
 using System;
 using System.Management.Automation;
+using System.Management.Automation.Runspaces;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Net.Security;
 using System.IO;
 using System.Text;
 using System.Collections;
 using System.Globalization;
 using System.Security.Authentication;
 using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Xml;
 using System.Collections.Generic;
@@ -172,6 +175,50 @@ namespace Microsoft.PowerShell.Commands
             {
                 handler.ServerCertificateCustomValidationCallback = HttpClientHandler.DangerousAcceptAnyServerCertificateValidator;
                 handler.ClientCertificateOptions = ClientCertificateOption.Manual;
+            }
+            else if (CertificateValidationScript != null)
+            {
+                // validationCallBackWrapper wraps the certificateValidationClosure ScriptBlock so the async callback can set the default
+                // PowerShell Runspace in the async thread to the Runspace of the current thread. This provides the ScriptBlock access to 
+                // the current scope.
+                Runspace currentRunspace = Runspace.DefaultRunspace;
+                ScriptBlock certificateValidationClosure = CertificateValidationScript.GetNewClosure();
+                Func<HttpRequestMessage,X509Certificate2,X509Chain,SslPolicyErrors,bool> validationCallBackWrapper = 
+                    delegate(HttpRequestMessage httpRequestMessage, X509Certificate2 x509Certificate2, X509Chain x509Chain, SslPolicyErrors sslPolicyErrors)
+                    {
+                        Runspace previousRunspace = Runspace.DefaultRunspace;
+                        Runspace.DefaultRunspace = currentRunspace;
+
+                        // Set Script scope automatic variables for the closure.
+                        PSVariableIntrinsics closureVi = certificateValidationClosure.Module.SessionState.PSVariable;
+                        closureVi.Set("HttpRequestMessage", httpRequestMessage);
+                        closureVi.Set("X509Certificate2", x509Certificate2);
+                        closureVi.Set("X509Chain", x509Chain);
+                        closureVi.Set("SslPolicyErrors", sslPolicyErrors);
+                        closureVi.Set("ErrorActionPreference", "Stop");
+
+                        Boolean result = false;
+                        try
+                        {
+                            result = LanguagePrimitives.IsTrue(
+                                certificateValidationClosure.InvokeReturnAsIs(
+                                    httpRequestMessage, 
+                                    x509Certificate2, 
+                                    x509Chain, 
+                                    sslPolicyErrors
+                                )
+                            );
+                        }
+                        catch // Treat all exceptions as Certificate failures.
+                        {
+                            result = false;
+                        }
+
+                        Runspace.DefaultRunspace = previousRunspace;
+                        return result;
+                    };
+
+                handler.ServerCertificateCustomValidationCallback = validationCallBackWrapper;
             }
 
             // This indicates GetResponse will handle redirects.
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
@@ -178,30 +178,15 @@ namespace Microsoft.PowerShell.Commands
             }
             else if (CertificateValidationScript != null)
             {
-                // validationCallBackWrapper wraps the certificateValidationClosure ScriptBlock so the async callback can set the default
-                // PowerShell Runspace in the async thread to the Runspace of the current thread. This provides the ScriptBlock access to 
-                // the current scope.
-                Runspace currentRunspace = Runspace.DefaultRunspace;
-                ScriptBlock certificateValidationClosure = CertificateValidationScript.GetNewClosure();
+                // validationCallBackWrapper wraps the CertificateValidationScript ScriptBlock so the async callback properly execute ScriptBlock
                 Func<HttpRequestMessage,X509Certificate2,X509Chain,SslPolicyErrors,bool> validationCallBackWrapper = 
                     delegate(HttpRequestMessage httpRequestMessage, X509Certificate2 x509Certificate2, X509Chain x509Chain, SslPolicyErrors sslPolicyErrors)
                     {
-                        Runspace previousRunspace = Runspace.DefaultRunspace;
-                        Runspace.DefaultRunspace = currentRunspace;
-
-                        // Set Script scope automatic variables for the closure.
-                        PSVariableIntrinsics closureVi = certificateValidationClosure.Module.SessionState.PSVariable;
-                        closureVi.Set("HttpRequestMessage", httpRequestMessage);
-                        closureVi.Set("X509Certificate2", x509Certificate2);
-                        closureVi.Set("X509Chain", x509Chain);
-                        closureVi.Set("SslPolicyErrors", sslPolicyErrors);
-                        closureVi.Set("ErrorActionPreference", "Stop");
-
                         Boolean result = false;
                         try
                         {
                             result = LanguagePrimitives.IsTrue(
-                                certificateValidationClosure.InvokeReturnAsIs(
+                                 CertificateValidationScript.InvokeReturnAsIs(
                                     httpRequestMessage, 
                                     x509Certificate2, 
                                     x509Chain, 
@@ -214,7 +199,6 @@ namespace Microsoft.PowerShell.Commands
                             result = false;
                         }
 
-                        Runspace.DefaultRunspace = previousRunspace;
                         return result;
                     };
 
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -1583,23 +1583,20 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
             $jsonResult.Headers.Host | Should BeExactly $params.Uri.Authority
         }
 
-        It "Verifies Invoke-WebRequest -CertificateValidationScript script has access to the calling scope" -Pending:$PendingCertificateTest {
-            # WebListener's Certificate Thumbprint
-            $thumbprint = 'C8747A1C4A46E52EEC688A6766967010F86C58E3'
-            $scriptHash = @{Subject = $null}
+        It 'Verifies Invoke-WebRequest -CertificateValidationScript script has a $_' -Pending:$PendingCertificateTest {
             $params = @{
                 Uri = Get-WebListenerUrl -Test 'Get' -Https
                 ErrorAction = 'Stop'
                 CertificateValidationScript = { 
-                    # $args[1] is the remote server's X509Certificate2 certificate
-                    $scriptHash['Subject'] = $args[1].Subject
-                    return ($args[1].Thumbprint -eq $thumbprint)
+                    $_.Certificate.Thumbprint -eq 'C8747A1C4A46E52EEC688A6766967010F86C58E3' -and
+                    $_.CertificateChain.ChainElements[0].Certificate.Thumbprint -eq 'C8747A1C4A46E52EEC688A6766967010F86C58E3' -and
+                    $_.SslErrors -eq 'RemoteCertificateChainErrors' -and
+                    $_.Request.Method.Method -eq 'GET'
                 }
             }
             $result = Invoke-WebRequest @Params
             $jsonResult = $result.Content | ConvertFrom-Json
             $jsonResult.Headers.Host | Should BeExactly $params.Uri.Authority
-            $scriptHash.Subject | Should BeExactly 'CN=localhost'
         }
 
         It "Verifies Invoke-WebRequest -CertificateValidationScript treats exceptions as Certificate failures" -Pending:$PendingCertificateTest {
@@ -2708,22 +2705,20 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
             $result.Headers.Host | Should BeExactly $params.Uri.Authority
         }
         
-        It "Verifies Invoke-RestMethod -CertificateValidationScript script has access to the calling scope" -Pending:$PendingCertificateTest {
-            # WebListener's Certificate Thumbprint
-            $thumbprint = 'C8747A1C4A46E52EEC688A6766967010F86C58E3'
-            $scriptHash = @{Subject = $null}
+        It 'Verifies Invoke-RestMethod -CertificateValidationScript script has a $_' -Pending:$PendingCertificateTest {
             $params = @{
                 Uri = Get-WebListenerUrl -Test 'Get' -Https
                 ErrorAction = 'Stop'
-                CertificateValidationScript = { 
-                    # $args[1] is the remote server's X509Certificate2 certificate
-                    $scriptHash['Subject'] = $args[1].Subject
-                    return ($args[1].Thumbprint -eq $thumbprint)
+                CertificateValidationScript = {
+                    $WebListenerThumbprint = 'C8747A1C4A46E52EEC688A6766967010F86C58E3'
+                    $_.Certificate.Thumbprint -eq $WebListenerThumbprint -and
+                    $_.CertificateChain.ChainElements[0].Certificate.Thumbprint -eq $WebListenerThumbprint -and
+                    $_.SslErrors -eq 'RemoteCertificateChainErrors' -and
+                    $_.Request.Method.Method -eq 'GET'
                 }
             }
-            $result = Invoke-RestMethod @Params
+            $result = Invoke-RestMethod @Params 
             $result.Headers.Host | Should BeExactly $params.Uri.Authority
-            $scriptHash.Subject | Should BeExactly 'CN=localhost'
         }
 
         It "Verifies Invoke-RestMethod -CertificateValidationScript treats exceptions as Certificate failures" -Pending:$PendingCertificateTest {
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -1557,6 +1557,78 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
     }
 
+    Context "Invoke-WebRequest CertificateValidationScript Tests" {
+        It "Verifies Invoke-WebRequest -CertificateValidationScript can accept all certificates" {
+            $params = @{
+                Uri = Get-WebListenerUrl -Test 'Get' -Https
+                ErrorAction = 'Stop'
+                CertificateValidationScript = { $true }
+            }
+            # WebListener uses a self-signed cert. Without -SkipCertificateCheck this would normally fail
+            $result = Invoke-WebRequest @Params
+            $jsonResult = $result.Content | ConvertFrom-Json
+            $jsonResult.Headers.Host | Should BeExactly $params.Uri.Authority
+        }
+
+        It "Verifies Invoke-WebRequest -CertificateValidationScript is ignored when -SkipCertificateCheck is present" {
+            $params = @{
+                Uri = Get-WebListenerUrl -Test 'Get' -Https
+                ErrorAction = 'Stop'
+                SkipCertificateCheck = $true
+                # This script would fail all certificates
+                CertificateValidationScript = { $false }                
+            }
+            $result = Invoke-WebRequest @Params
+            $jsonResult = $result.Content | ConvertFrom-Json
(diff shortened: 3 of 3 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -174,6 +177,50 @@ internal virtual HttpClient GetHttpClient(bool handleRedirect)
...
+                        Runspace previousRunspace = Runspace.DefaultRunspace;
+                        Runspace.DefaultRunspace = currentRunspace;
+
+                        // Set Script scope automatic variables for the closure.
+                        PSVariableIntrinsics closureVi = certificateValidationClosure.Module.SessionState.PSVariable;
+                        closureVi.Set("HttpRequestMessage", httpRequestMessage);
+                        closureVi.Set("X509Certificate2", x509Certificate2);
+                        closureVi.Set("X509Chain", x509Chain);
+                        closureVi.Set("SslPolicyErrors", sslPolicyErrors);
+                        closureVi.Set("ErrorActionPreference", "Stop");
```
iSazonov: Maybe put `x509Certificate2`on first position because the callback is for it?
Another thought is to pass a PSObject as $PSItem with the properties (x509Certificate2, httpRequestMessage, ...).

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -94,6 +94,12 @@ public abstract partial class WebRequestPSCmdlet : PSCmdlet
         [Parameter]
         public virtual SwitchParameter SkipCertificateCheck { get; set; }
 
+        /// <summary>
+        /// gets or sets the CertificateValidationScript property
+        /// </summary>
```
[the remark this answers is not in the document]
rkeithhill: In that case, perhaps the two parameters should be in mutually exclusive parameter sets?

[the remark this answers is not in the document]
markekraus: @rkeithhill  We don't want these to be full mutually exclusive via parameter set. we want to support thuis usage:

```powershell
$script = { <# some validation script #> }
$PSDefaultParameterValues['Invoke-RestMethod:CertificateValidationScript'] = $script

# All calls will use the validation script
Invoke-RestMethod https://unknonwnurl.com/

# But the user may want to override their normal validation for a certain Url
Invoke-RestMethod https://devsite.com/ -SkipCertificateCheck
```

Also, the Web Cmdlets have traditionally had a single parameter set and there are other parameters that are exclusive (i.e. `-Infile` and `-Body`). Introducing a parameter set here would be awkward.

@iSazonov I can update the summary, but other parameters with this kind of behavior to not call it out in their summaries. I believe these summaries are only visible when coding in C#, correct? Since these are `PSCmdlet` derived, you don't even see these summaries from with in the same assembly. In which case, this is visible only when working in classes that derive from `WebRequestPSCmdlet`.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
markekraus: All 4 are equally important for validating the certificate. 

```powershell
$Script = {
    Return (
        $HttpRequestMessage.RequestUri.AbsoluteUri -eq 'https://www.google.com/' -and
        $X509Certificate2.Subject -eq 'CN=www.google.com, O=Google Inc, L=Mountain View, S=California, C=US' -and
        $X509Chain.ChainElements[2].Certificate.Thumbprint -eq 'DE28F4A4FFE5B92FA3C503D1A349A7F9962A8212' -and
        $SslPolicyErrors -eq 'None'
    )
}
```

I wanted to mirror the ,NET callback as much as possible to prevent confusion.

To be clear, `$args[0]`, `$X509Certificate2`, and `Param( [Parameter(Position=0)]$certificate <#...#> )` are all possible to the user.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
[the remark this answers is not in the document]
iSazonov: If Web Cmdlets have traditionally had a single parameter set I agree to follow the pattern.

My thought was that we should add the comment somewhere. The place is best.  It's good for documenting code. Also we generate XML help files - I guess it can be used for creation of HTLM documentation.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
iSazonov: I believe if we say about a script block we shouldn't concerned about .Net templates. Script writes don't see underlying implementations and usualy expect implicit `$PSItem` or explicit `Param`.
Also the variables actually is automatic variables and we should add its in list of automatic variables. This seems unjustified.
I suggest we wait for @lzybkr  opinion.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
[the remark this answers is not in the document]
markekraus: If I'm not mistaken, the help files are generated from the [PowerShell/PowerShell-Docs](https://github.com/PowerShell/PowerShell-Docs) markdown? I can certainly say that I have never seen `gets or sets the Body property` (the summary for `-Body`) in anything. I certainly don't see any of the parameter summaries in any of the help xmls. The only place this shows up is the `Microsoft.PowerShell.Commands.Utility.xml` file and that has the same unhelpful summaries as the other parameters.

I will update it. I just think it wont be actually that helpful to anyone.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
markekraus: It's automatic to the ScriptBlock only. These do not show up in the calling scope. I don't see the problem? Please elaborate "we should add its in list of automatic variables". Do Script Block have their own list of automatic variables?

And, we *should* be concerned about the underlying .NET. It would be confusing to maintainers if the `Func` delegate has one argument order but the script block has another. The user gets the convenience of the automatic variables or `Param()` if they don't like that `$args[0]` is the `HttpMessageRequest` . As I said, all 4 arguments are equally important. Some users may have rules which never even test the certificate (URL based acceptance).

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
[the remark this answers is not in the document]
SteveL-MSFT: There's two types of help, there's the updatable help that exists in PowerShell-Docs and there's api reference docs that get generated from the assemblies.  Any public api should be documented in code as the documentation is generated from that xml.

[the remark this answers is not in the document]
markekraus: @SteveL-MSFT Is this the documentation you are referring to? https://msdn.microsoft.com/en-us/library/microsoft.powershell.commands.webrequestpscmdlet(v=vs.85).aspx If not, what exactly? I ask because the other summaries are very unhelpful and if these are being documented somewhere they could use some serious attention. I have never seen these summaries outside of the XML file and the source code.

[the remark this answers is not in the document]
iSazonov: We already add in our coding guigance that we welcome good comments everywhere. I hope in time we'll see a well-documented code 😄

markekraus:
@lzybkr @SteveL-MSFT Is there something more I need to do for this PR?

iSazonov:
LGTM but we need review from MSFT experts.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
lzybkr: Why set the variables **and** pass them as args?

Also, why set `$ErrorActionPreference`? That might not be desirable.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -174,6 +177,50 @@ internal virtual HttpClient GetHttpClient(bool handleRedirect)
...
+                                    sslPolicyErrors
+                                )
+                            );
+                        }
+                        catch // Treat all exceptions as Certificate failures.
+                        {
+                            result = false;
+                        }
+
+                        Runspace.DefaultRunspace = previousRunspace;
```
lzybkr: I'd sleep better if this was in a finally - then I don't need to worry about a change to let some exceptions through.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -174,6 +177,50 @@ internal virtual HttpClient GetHttpClient(bool handleRedirect)
...
+                // validationCallBackWrapper wraps the certificateValidationClosure ScriptBlock so the async callback can set the default
+                // PowerShell Runspace in the async thread to the Runspace of the current thread. This provides the ScriptBlock access to 
+                // the current scope.
+                Runspace currentRunspace = Runspace.DefaultRunspace;
+                ScriptBlock certificateValidationClosure = CertificateValidationScript.GetNewClosure();
+                Func<HttpRequestMessage,X509Certificate2,X509Chain,SslPolicyErrors,bool> validationCallBackWrapper = 
+                    delegate(HttpRequestMessage httpRequestMessage, X509Certificate2 x509Certificate2, X509Chain x509Chain, SslPolicyErrors sslPolicyErrors)
+                    {
+                        Runspace previousRunspace = Runspace.DefaultRunspace;
+                        Runspace.DefaultRunspace = currentRunspace;
```
lzybkr: This is worrisome. From reading just this code, I have no clue if `currentRunspace` will be blocked while the delegate is running.

For example, if we're streaming objects and in the middle of a call to `WriteObject` in the cmdlet's thread, that could yield to a random PowerShell event handler, which will be a race condition with whatever this delegate might do.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
markekraus: > Why set the variables and pass them as args?

I wanted to support parameter binding by type i.e. 

```powershell
param(
    [System.Security.Cryptography.X509Certificates.X509Certificate2]$cert,
# ...
````
I was trying to be flexible. 

> Also, why set `$ErrorActionPreference`? That might not be desirable.

The [planned] documented behavior is that that errors and exceptions result in certificate failures. I ran into some strange scenarios where if errors weren't terminating it would hang the thread. but, this was before I switched to a closure and before I was using `LanguagePrimitives.IsTrue()`. I'm not sure if it s still needed, I can check though.

File: src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs (hunk 4 above)
Comment (markekraus) [reply src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs]: I'm not sure how to make this work in a different runspace and supporting access to the current scope. To me, personally, that is not an issue as I can live with this not having scope access. But in the discussion the issue was brought up that users might find that confusing. or difficult to work with.

Can you mock up something that would cause this race condition so I can test it? I did some basic tests with this running in a pipeline and as part of a process block, but I was never able to cause a hang. or unexpected behavior. I thought maybe since we are forcing the call to run synchronously it might nor be an issue:

https://github.com/PowerShell/PowerShell/blob/aea561f08df3cc1c6ce92135db47604868fade2c/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs#L468

But, honestly, I'm not anywhere near an expert on this area. I just know that if I don't specify a run space here I get the exception about a missing runspace.

File: src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs (hunk 3 above)
Comment (markekraus) [reply src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs]: I can do that.

File: src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs (hunk 4 above)
Comment (lzybkr) [reply src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs]: Sure, before running your code, run something like:

```powershell
$t = [System.Timers.Timer]::new(10)
$t.AutoReset = $true
Register-ObjectEvent -InputObject $t -EventName Elapsed -Action { Write-Host 'event' }
$t.Enabled = $true
```

If this doesn't cause problems, try something less trivial in the action - call functions, maybe a recursive function calculating fibonacci or something.

In your certificate validation script, also do something similarly non-trivial, e.g. compute a different fibonacci number.

I'm thinking some state will be corrupted, e.g. the current scope.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
lzybkr: I'm definitely not in favor of both variables and parameters. Either way you need good documentation, but I prefer parameters.

And please do verify that setting `$ErrorActionPreference` is not needed.

markekraus: This is the draft documentation:
https://github.com/PowerShell/PowerShell-Docs/compare/staging...markekraus:CertificatevalidationScriptDraft 

I'm not sure I agree that parameters are preferable. Personally, I'm OK with either. But, I think most people would rather have auto-variables than have to declare  `param()` or try to remember which `$args` is which. in the end, this is a documentation heavy dependent feature no matter which way it is implemented.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
markekraus: Using that event and a simple block caused it to hang the thread. So definitely some corruption. ☹️ I'm looking at some other options.

lzybkr: 10ms might be too fast, maybe try a little slower, say 100ms. The goal was to hit a race as quickly as possible and get something other than a hang, though I suppose a hang is one possible outcome of a race.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
lzybkr: I'll let the @PowerShell/powershell-committee chime in on this one.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
markekraus: @lzybkr actually... I think that test may kill any callback, not just the one in this issue.

I reduced the code to this
```csharp
Func<HttpRequestMessage,X509Certificate2,X509Chain,SslPolicyErrors,bool> validationCallBackWrapper = 
    delegate(HttpRequestMessage httpRequestMessage, X509Certificate2 x509Certificate2, X509Chain x509Chain, SslPolicyErrors sslPolicyErrors)
    {
        Boolean result = false;
        try
        {
            result = LanguagePrimitives.IsTrue(
                CertificateValidationScript.InvokeReturnAsIs(
                    httpRequestMessage, 
                    x509Certificate2, 
                    x509Chain, 
                    sslPolicyErrors
                )
            );
        }
        catch // Treat all exceptions as Certificate failures.
        {
            result = false;
        }
        return result;
    };

handler.ServerCertificateCustomValidationCallback = validationCallBackWrapper;
```
No more closure, no more setting the default runspace, no more auto variables or erroraction. Just wrapping the `InvokeReturnAsIs()` in the delegate and this hangs:

```powershell
$t = [System.Timers.Timer]::new(10)
$t.AutoReset = $true
Register-ObjectEvent -InputObject $t -EventName Elapsed -Action { write-host =  'event' }
$t.Enabled = $true
$script = { return $true }
Invoke-RestMethod -uri https://httpbin.org/get -CertificateValidationScript $Script
```

if I reduce it back to 

```csharp
Func<HttpRequestMessage,X509Certificate2,X509Chain,SslPolicyErrors,bool> validationCallBackWrapper =
    LanguagePrimitives.ConvertTo<Func<HttpRequestMessage,X509Certificate2,X509Chain,SslPolicyErrors,bool>>(CertificateValidationScript);
handler.ServerCertificateCustomValidationCallback = validationCallBackWrapper;
```

I get this:

```none
Invoke-RestMethod : There is no Runspace available to run scripts in this thread. You can provide one in the
DefaultRunspace property of the System.Management.Automation.Runspaces.Runspace type. The script block you attempted
to invoke was:  return $true
```
```

Comment (markekraus) [reply src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs]: 100ms didn't make a difference. ☹️ Still hangs on all the scenarios.

**Comment by SteveL-MSFT** [comment]
@PowerShell/powershell-committee reviewed the topic of params or automatic variables passed to the script block delegate.  There was concerns about new automatic variables scoped only to this usage but would needed to be documented in the about_automaticvariables topic which may create some confusion.  In addition, params seems to be a general pattern already in use across PowerShell, so the request is to use params for the script block (and update the doc draft accordingly).

There were other concerns we didn't have time to discuss such as which runspace the delegate ran in.  For now, continue that discussion in this PR.

File: src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs (hunk 3 above)
Comment (markekraus) [reply src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs]: no longer applicable.

File: src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs (hunk 1 above)
Comment (markekraus) [reply src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs]: Automatic variables removed per @PowerShell/powershell-committee decision. `$ErrorActionPreference` also appears to be no longer needed.

File: src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs (hunk 4 above)
Comment (markekraus) [reply src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs]: This code has been changed and the setting/unsetting of the runspace is no longer a part of it. However, the callback still results in a thread hang when using the event trigger(s) provided by @lzybkr. I'm not sure where to go on that issue.

**Comment by markekraus** [comment]
Well that's baffling. It's not a surprise that the certificate validation doesn't work on macOS. I ran into similar issues enabling `-Certificate` because macOS (and some linux distors) use a different implementation than openssl but, how did these tests manage to pass on macOS in previous commits? 

* working: https://travis-ci.org/PowerShell/PowerShell/jobs/282050275
* not working: https://travis-ci.org/PowerShell/PowerShell/jobs/287024709

It should have failed before.

**Comment by alx9r** [comment]
> There was concerns about new automatic variables scoped only to this usage but would needed to be documented in the about_automaticvariables topic which may create some confusion. 

@SteveL-MSFT Was the prospect of providing the callback parameters to the scriptblock by way of properties on `$_` contemplated?  That feels like the most PowerShell-ey style to me.  There is precedent for non-pipeline use of `$_` instead of `param()` in [ValidateScript()](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_parameters?view=powershell-5.1#validatescript-validation-attribute).  

I've been favoring `$_` over `param()` for scriptblock predicates like `ServerCertificateCustomValidationCallback` mainly because the `param()` block ends up being repeated, often overshadows the predicate in character count, and usually doesn't provide much useful information to a reader.

Here's a mock-up of `param()` vs `$_` for illustration:

```PowerShell
{
    param($Message,$Cert,$Chain,$PolicyErrors)
    Assert-CertificateCompliance -Policy Strict -Message $Message -Cert $Cert -Chain $Chain -PolicyErrors $PolicyErrors
}

{$_ | Assert-CertificateCompliance -Policy Strict}
```

For me the point of this PR is to be able to impose per-URL certificate requirements.  So I'll end up with policy data structures looking something like this:

```PowerShell
@{
    Uri = 'https://github.com'
    CertificateValidator = {
        $_ | Assert-CertificateCompliance -Policy Strict
        $_ | Assert-Certificate Server -Thumbprint d79f076110b39293e349ac89845b0380c19e2f8b
    }
}

@{
    Uri = 'https://www.powershellgallery.com'
    CertificateValidator = {
        $_ | Assert-CertificateCompliance -Policy Strict
        $_ | Assert-Certificate Intermediate -Thumbprint ‎98af62baa639b0827f887c75a77f4afd4eff2685
    }
}
```

The validation scriptblock gets repeated for each uri so noise from `param()` adds up.

SteveL-MSFT:
@alx9r no, using members of `$_` wasn't something that was considered and is an interesting approach.  Would like some others to chime in on this proposal.  cc @lzybkr @JamesWTruher @HemantMahawar

markekraus:
@SteveL-MSFT I had an early draft that did it that way. I had dropped it in favor of the automatic variables before submitting the PR.

SteveL-MSFT:
@markekraus sorry, must have missed it.  Getting myself involved in too many discussions on here to keep track of them (something I'm learning not to do!)

markekraus:
@SteveL-MSFT I don't think the draft ever made it's way to my public repo so I don't think you missed anything, actually. My point was more that I'm on board with the idea of a `$_` with members if that lends any weight in the decision making.

markekraus:
After much trial and error I determined that the only safe way to do this where any kind of event triggers were in place was run the `ScriptBlock` in a completely new session state and runspace. Variables could easily be imported, but I did tests where the event triggers updated a variable at that same time as callback and that resulted in a hang. Importing functions and modules leaves them in a disconnected state from the user's context (module scoped variables are lost).

I have incorporated the `$_`. I'm open for suggestions on renaming the members. Right now they are `Request`, `Certificate`, `CertificateChain`, and `SslErrors`. This current implementation does not support `param()`. I'm assuming we want to go with one or the other and not both.

I have tested the current code with the event triggers @lzybkr provided and it works. no hangs.

The down side is that the `ScriptBlock` is now devoid of the calling scope. I was originally under the impression it would need to be done this way and I think that so long as it is documented it should be acceptable. Any of the other means of making the scope available results in either unpredictable behavior or hangs. I think no scope is better than some broken scope.

alx9r:
> ...I did tests where the event triggers updated a variable at that same time as callback and that resulted in a hang.

I'm curious what these tests looked like.  It doesn't seem like `Register-ObjectEvent` naturally accepts `PSVariable`s in its parameters.  It also doesn't seem like the `Register-ObjectEvent -Action` scriptblock has access the call site's variables.  In the tests that hung how did the event handler get a reference to the same variable as the callback?

> I have incorporated the $_. I'm open for suggestions on renaming the members. Right now they are Request, Certificate, CertificateChain, and SslErrors.

I see that [the current documentation for `HttpClientHandler.ServerCertificateCustomValidationCallback`](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.servercertificatecustomvalidationcallback?view=netcore-2.0#System_Net_Http_HttpClientHandler_ServerCertificateCustomValidationCallback) doesn't name its parameters.  For that matter, there is very little documentation about the semantics of that callback altogether.  dotnet/corefx#24774 might improve that documentation and that might lead to parameter names.  If `ServerCertificateCustomValidationCallback` does end up with documented parameter names, it seems like the properties on `$_` should have identical names.  The closest documented callback interface seems to be [full .net's `System.Net.Security.RemoteCertificateValidationCallback`](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.remotecertificatevalidationcallback?view=netframework-4.7.1).  I propose adopting the corresponding names from that callback interface in the meantime.

> The down side is that the ScriptBlock is now devoid of the calling scope. I was originally under the impression it would need to be done this way and I think that so long as it is documented it should be acceptable. Any of the other means of making the scope available results in either unpredictable behavior or hangs. I think no scope is better than some broken scope.

Is there a reason that the scope rules wouldn't be the same as for other scriptblocks using [remote variables](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_remote_variables?view=powershell-5.1)?  [`Invoke-Command`](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/invoke-command?view=powershell-5.1)` -Session -ScriptBlock`, for example, provides access to local variables in the remote session by way of the `$using:` expression.

FWIW, I think this would be a pretty awkward interface without having some way of making user objects available in the callback scriptblock.  A natural use would be as follows:

```PowerShell
@{
    u='https://github.com/some/important/file'
    thumbprint = 'd79f076110b39293e349ac89845b0380c19e2f8b'
},
@{
    u='https://powershellgallery.com/some/other/important/file'
    cv = '98af62baa639b0827f887c75a77f4afd4eff2685'
} |
    % { 
        # ... do some prep things
        Invoke-WebRequest $_.u -CertificateValidationScript {
            # ... a bunch of tests that are identical for every certificate
            $_.Certificate.Thumbprint -eq $using:_
        }
        # ... do some cleanup things
    }
```

Without `$using:`, arguments, or some other means of passing data, it will be tough to separate the concern of which thumbprints are acceptable for which url from the concern of checking for certificate revokation, for example.

markekraus:
> I'm curious what these tests looked like. 
I just used the same variable in @lzybkr's example and used it

```powershell
$t = [System.Timers.Timer]::new(10)
$t.AutoReset = $true
Register-ObjectEvent -InputObject $t -EventName Elapsed -Action { Write-Host 'event' }
$t.Enabled = $true
$script = {$t.Enabled = $false; $true}
irm -uri https://expired.badssl.com
#hangs
```

`$t` was being populated as a PSVarabile in the iss.

>  I propose adopting the corresponding names from that callback interface in the meantime.

I disagree. Unless it is documented on this API, we should use something generic that is easier for PowerShell users to use. `Sender` is a terrible choice when it's an `HttpRequestMessage`. If Sender were the documented name for this API, then I would concede that's probably better. 

> Is there a reason that the scope rules wouldn't be the same as for other scriptblocks using

`$using` in every instance I've seen is a unique implementation. If there were a generic API for adding this capability to a `ScriptBlock`, then `$using:` would be awesome to have here. but I see no point in adding yet-another-using-implimentation for validation callbacks on cmdlets that don't normally do any work with `ScriptBlock`s.

> Without $using:, arguments, or some other means of passing data,..

Erm, not really? I mean yea, not on the ease of use level that using variables from the calling scope or `$using:`, but, most validation logic is fairly static, in my experience. Create module, Import it in the ScriptBlock, magic. or use CSV, Clixml, JSON files for confgis if you don't wants it hard coded in the the PS Script.

I'm not saying this is ideal. What I am saying is that I can't see a safe way to get the calling scope available in the remote scope with out severely over-complicating the implementation.

Also, these validations scripts are "per call". For 6.1 I want to look at adding session persistent options. One of them will be add your own .NET delegates for cert validation, So if you want to have some variable magic then you can. The goal in this PR is to provide the basic implementation that will work for most people on a "per call" basis. I think most will just need to ensure a static thumbprint or 2 against URLs. Complex validation scenarios are not the goal of the implementation in this PR, but are on my radar for the futre.

alx9r:
> Complex validation scenarios are not the goal of the implementation in this PR, but are on my radar for the futre.

OK.  Got it.  Thanks @markekraus

markekraus:
rebased to fix merge conflict.

iSazonov:
@markekraus Athother merge conflict.

markekraus:
rebased.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and recommends using the param() format.  Specifically, $_ provides no context on what it represents.  Typically $_ means the current object and reusing it in the scriptblock changes that expectation.  Using param() makes the code more readable and therefore maintainable.

iSazonov:
@markekraus Please rebase with next new commit to get latest updates and pass CIs.

markekraus:
@iSazonov I still need to incorporate the changes per the committee's decision. I will hopefully have free time this week. I will rebase before I commit the changes.

markekraus:
FYI, I'm waiting on #5518 as this will need to rebased again and more drastically.

markekraus:
Due to the issues of rebasing this and that the code still needs to be updated to fit the review comitee's decision, I'm going to close this PR. I will attempt to address this feature before 6.1.0.

PR #4972, opened 2017-10-02
Reformat Measure-Object test

iSazonov:
The Measure-Object test file contains multiple tabs and formatting issues.
In #4969 we need add new tests in the file but the formatting issues block us.
Please fast review to unblock #4969.


SteveL-MSFT (approved):
LGTM

Issue #3726, opened 2017-05-07
Get-ChildItem ignores the -Depth parameter when -Include or -Exclude are specified


mklement0:
Steps to reproduce
------------------

```powershell
Set-Location ~
Get-ChildItem . -Depth 0 | Measure-Object
Get-ChildItem . -Depth 0 -Include * | Measure-Object
Get-ChildItem . -Depth 0 -Exclude NoSuchItemImSure | Measure-Object
```

Expected behavior
-----------------

All 3 commands should output the same number - the count of items in the current dir.

Actual behavior
---------------

* The 1st command reports the number of items in the current directory only.

* The 2nd and 3rd commands unexpectedly traverse the entire subtree, ignoring `-Depth 0`. 

Environment data
----------------

```powershell
PowerShell Core v6.0.0-alpha (v6.0.0-alpha.18) on macOS 10.12.4
PowerShell Core v6.0.0-alpha (v6.0.0-alpha.18) on Microsoft Windows 10 Pro (64-bit; v10.0.14393)
```

Windos:
Given that I spent longer than I care to admit being tripped up by this this morning, I figure it couldn't hurt to add a comment on this issue in support of sorting it.

To my mind I see two potential courses of action:

1. Make Include/Exclude and Depth mutually exclusive. Not intuitive, but at least the user will get an error rather than unexpected results.
2. Adjust the cmdlet so that using these combinations of parameters works as one would expect them to.

I'd say the second one is more desirable, but I'm approaching this from a high-level as a user so I don't fully appreciate the amount of work or if it would result in any breaking changes.

psmolkin:
The same behavior in 5.1.15063.608.
`Get-ChildItem  -Depth 0 -Filter 'mask'` works good, but
`Get-ChildItem  -Depth 0 -Include 'mask'` returns all 'mask' elemens of tree.

jszabo98:
This is still true to this day in powershell 5.1  (5.1.19041.4170).

PR #4985, opened 2017-10-03
Make Get-ChildItem honor Depth parameter with Include/Exclude

Windos:
Addresses issue #3726 by adding decrementing depth to ProcessPathItems.

Includes tests specifically for the examples listed in the issue and also testing that recursion still works when excluding or including a string.

### DIFF
--- a/src/System.Management.Automation/engine/SessionStateContainer.cs
+++ b/src/System.Management.Automation/engine/SessionStateContainer.cs
@@ -1593,7 +1593,7 @@ namespace System.Management.Automation
                             }
 
                             int unUsedChildrenNotMatchingFilterCriteria = 0;
-                            ProcessPathItems(providerInstance, providerPath, recurse, context, out unUsedChildrenNotMatchingFilterCriteria, ProcessMode.Enumerate);
+                            ProcessPathItems(providerInstance, providerPath, recurse, context, out unUsedChildrenNotMatchingFilterCriteria, ProcessMode.Enumerate, depth: depth);
                         }
                     }
                     else
@@ -1836,6 +1836,10 @@ namespace System.Management.Automation
         ///
         /// <param name="skipIsItemContainerCheck">a hint used to skip IsItemContainer checks</param>
         ///
+        /// <param name="depth">
+        /// Limits the depth of recursion; uint.MaxValue performs full recursion.
+        /// </param>
+        ///
         /// <exception cref="ProviderNotFoundException">
         /// If the <paramref name="path"/> refers to a provider that could not be found.
         /// </exception>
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/engine/SessionStateContainer.cs
```diff
@@ -1593,7 +1593,7 @@ internal Collection<PSObject> GetChildItems(string[] paths, bool recurse, uint d
                             }
 
                             int unUsedChildrenNotMatchingFilterCriteria = 0;
-                            ProcessPathItems(providerInstance, providerPath, recurse, context, out unUsedChildrenNotMatchingFilterCriteria, ProcessMode.Enumerate);
+                            ProcessPathItems(providerInstance, providerPath, recurse, context, out unUsedChildrenNotMatchingFilterCriteria, ProcessMode.Enumerate, depth: depth);
```
iSazonov: I'd prefer to exclude parameters with defaults and use overloads.

Windos: Just pushed new commit swapping optional parameter for overload. Is it along the right track?

iSazonov:
@daxian-dbw Can we ask @Windos to fix (replace with overloads) skipIsItemContainerCheck defaults in the PR?

daxian-dbw:
Well, the optional parameter was introduced to be an alternative of overloads, to reduce the number of overloads of a method.
If we remove `skipIsItemContainerCheck` by adding overloads, then there will be 4 overloads of `ProcessPathItems`, which IMHO might be cumbersome.

The optional parameter should never be used for public APIs or even internal APIs that may be called from other friend assemblies because the default value is baked into the caller, and that means if we change the default value, the caller assembly needs to be re-compile to get the new default value.
However, it's fine for private methods, as long as it doesn't affect readability much.

iSazonov:
@daxian-dbw I asked because adding second optional parameter reduced readability. Do we have to create overloads in this case or can we just replace its with one method with all the parameters?

Windos:
I'm happy to put the work in regardless of which direction is decided on.

daxian-dbw:
@iSazonov I think you made the right call. Adding the second optional parameter would worsen the readability. I think the current two overloads look good.

adityapatwardhan:
@iSazonov @daxian-dbw can you update your review

iSazonov:
@daxian-dbw Is `uint depth` parameter on best place?

adityapatwardhan (approved):
LGTM

adityapatwardhan:
@daxian-dbw  Can you update your review?

daxian-dbw (approved):

iSazonov (approved):

adityapatwardhan:
@Windos Thanks for your contribution!

Issue #4940, opened 2017-09-28
PowerShell exits opening runspace created with SSHConnectionInfo


SteveL-MSFT:
Doesn't repro on Windows, but repros on Linux and Mac

Steps to reproduce
------------------

```powershell
 $sshConnectionInfo = [System.Management.Automation.Runspaces.SSHConnectionInfo]::new(
 "UserName",
 "localhost",
 "NoValidKeyFilePath",
 22)
 $rs = [runspacefactory]::CreateRunspace($sshConnectionInfo)
 $rs.Open()
```

Expected behavior
-----------------

```none
error message
```

Actual behavior
---------------

```none
powershell exits due to SIGKILL
```

Environment data
----------------

```powershell
Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.7
OS                             Linux
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

PaulHigin:
Fixed with #4993

PR #4993, opened 2017-10-03
Fix for Linux platform PowerShell exit on error during SSH remoting connection

PaulHigin:
This fixes addresses issue #4940.

The problem is that the connection clean up code in SSHClientSessionTransportManager would try to terminate a process with id == 0, which is the initial value for the _sshProcessId field when no SSH process has been created yet.  For both Linux and Windows platforms process id of zero is a special process and should not be terminated.  Fortunately those processes are not terminated by the existing code but on Linux platforms the PowerShell parent process exits immediately.

The fix is to add checks that the SSH process has been created and is valid before terminating it during connection cleanup.

### DIFF
--- a/src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
+++ b/src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
@@ -1488,18 +1488,23 @@ namespace System.Management.Automation.Remoting.Client
             var stdErrReader = _stdErrReader;
             if (stdErrReader != null) { stdErrReader.Dispose(); }
 
-            try
+            var sshProcessId = _sshProcessId;
+            _sshProcessId = 0;
+            if (sshProcessId != 0)
             {
-                var sshProcess = System.Diagnostics.Process.GetProcessById(_sshProcessId);
-                if ((sshProcess != null) && !sshProcess.HasExited)
+                try
                 {
-                    sshProcess.Kill();
+                    var sshProcess = System.Diagnostics.Process.GetProcessById(sshProcessId);
+                    if ((sshProcess != null) && (sshProcess.Handle != IntPtr.Zero) && !sshProcess.HasExited)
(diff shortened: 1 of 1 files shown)


dantraMSFT (approved):
LGTM

src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
```diff
@@ -1488,18 +1488,23 @@ private void CloseConnection()
             var stdErrReader = _stdErrReader;
             if (stdErrReader != null) { stdErrReader.Dispose(); }
 
-            try
+            var sshProcessId = _sshProcessId;
```
SteveL-MSFT: Can't you use `_sshProcessId` directly instead of creating a new variable?

src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
```diff
@@ -1488,18 +1488,23 @@ private void CloseConnection()
...
+            _sshProcessId = 0;
```
SteveL-MSFT: Perhaps initialize it to 0 in the constructor instead of resetting it here?

SteveL-MSFT (changes requested):

src/System.Management.Automation/engine/remoting/fanin/OutOfProcTransportManager.cs
PaulHigin: CloseConnection can be called from multiple places (state machine, error handling) and this is just to prevent going through this code path multiple times.  I am not concerned about race conditions since Process type is thread safe ... I just want to prevent running the code path after the process has been terminated.

I'll add a comment to clarify.

SteveL-MSFT (approved):

mirichmo (approved):

PR #4994, opened 2017-10-03
Fix Linux packages

daxian-dbw:
Part of the work for #3961

`dotnet-runtime-2.0.0` Linux packages for Debian distributions depend on different lib packages on different distros, for example, `libicu52` on `ubuntu.14.04`, `libicu55` on `ubuntu.16.04`, `libicu57` on `ubuntu.17.04`, `libicu52` on `debian 8`, and `libicu57` on `debian 9`. A similar issue also happens to `libssl` -- `libssl1.0.2` on `debian 9` and `libssl1.0.0` on the rest debian distros.

In order to use a single DEB package for all distros, we need to depend on `-dev` packages, like `libicu-dev` and `libssl-dev`. However, dev packages are not meant for production software, and dotnet core has separate DEB package for each Debian distro, so I think we should do that too.

So for Debian distros, we have a separate package for each distro. For Red Hat family, we have a single RPM package, just like the `dotnet-runtime-2.0.0` package.

Even though we have separate packages for Debian distros, we now can create them all from one single build. Given that AppImage current needs to be created on `ubuntu.14.04`, we can generate all pacakges from `ubuntu.14.04`.
For RPM package, we should generate it from `CentOS7`.

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -131,9 +131,13 @@ function Get-EnvironmentInformation
         $LinuxInfo = Get-Content /etc/os-release -Raw | ConvertFrom-StringData
 
         $environment += @{'LinuxInfo' = $LinuxInfo}
+        $environment += @{'IsDebian' = $LinuxInfo.ID -match 'debian'}
+        $environment += @{'IsDebian8' = $Environment.IsDebian -and $LinuxInfo.VERSION_ID -match '8'}
+        $environment += @{'IsDebian9' = $Environment.IsDebian -and $LinuxInfo.VERSION_ID -match '9'}
         $environment += @{'IsUbuntu' = $LinuxInfo.ID -match 'ubuntu'}
         $environment += @{'IsUbuntu14' = $Environment.IsUbuntu -and $LinuxInfo.VERSION_ID -match '14.04'}
         $environment += @{'IsUbuntu16' = $Environment.IsUbuntu -and $LinuxInfo.VERSION_ID -match '16.04'}
+        $environment += @{'IsUbuntu17' = $Environment.IsUbuntu -and $LinuxInfo.VERSION_ID -match '17.04'}
         $environment += @{'IsCentOS' = $LinuxInfo.ID -match 'centos' -and $LinuxInfo.VERSION_ID -match '7'}
         $environment += @{'IsFedora' = $LinuxInfo.ID -match 'fedora' -and $LinuxInfo.VERSION_ID -ge 24}
         $environment += @{'IsOpenSUSE' = $LinuxInfo.ID -match 'opensuse'}
--- a/tools/appimage.sh
+++ b/tools/appimage.sh
@@ -35,7 +35,7 @@
 APP=powershell
 
 # Generate status file for use by apt-get; assuming that the recipe uses no newer
-# ingredients than what would require more recent dependencies than what we assume 
+# ingredients than what would require more recent dependencies than what we assume
 # to be part of the base system
 generate_status()
 {
@@ -60,7 +60,7 @@ delete_blacklisted()
       rm -f "${FOUND}"
     fi
   done
-  
+
   # Do not bundle developer stuff
   rm -rf usr/include || true
   rm -rf usr/lib/cmake || true
@@ -409,7 +409,7 @@ OPTIONS="-o Debug::NoLocking=1
 -o APT::Install-Suggests=0
 "
 
-cp ../powershell_*_amd64.deb .
+cp ../powershell_*ubuntu.14.04_amd64.deb .
 
 # Add local repository so that we can install deb files
 # that were downloaded outside of a repository
--- a/tools/packaging/packaging.psm1
+++ b/tools/packaging/packaging.psm1
@@ -1,6 +1,7 @@
 $Environment = Get-EnvironmentInformation
 
 $packagingStrings = Import-PowerShellDataFile "$PSScriptRoot\packaging.strings.psd1"
+$DebianDistributions = @("ubuntu.14.04", "ubuntu.16.04", "ubuntu.17.04", "debian.8", "debian.9")
 
 function Start-PSPackage {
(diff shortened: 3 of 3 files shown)


daxian-dbw:
@TravisEz13 @adityapatwardhan The changes in `Start-PSPackage` looks huge, but actually is not that much. It shows a big chunk of changes because the existing script needs to be moved to `End { ... }` block due to the introduction of the dynamic parameter `-Distribution`. This dynamic parameter is enabled only if `-Type deb` is specified, so that we can indicate which Debian distro the package is targeting.

daxian-dbw:
I have tested all packages on all supported distros except for Red Hat Enterprise Linux 7. They are:
`Ubuntu.14.04`, `Ubuntu.16.04`, `Ubuntu.17.04`, `CentOS7`, `openSUSE 42,2`, `Fedora 25`, `Fedora 26`.

tools/packaging/packaging.psm1
```diff
@@ -246,255 +260,305 @@ function New-UnixPackage {
...
+                "libgcc1",
+                "libgssapi-krb5-2",
+                "liblttng-ust0",
+                "libstdc++6",
+                "libunwind8",
+                "libuuid1",
+                "zlib1g"
+            )
+
+            if ($DebDistro -eq "ubuntu.14.04") {
```
TravisEz13: would this be more readable as a switch statement?

build.psm1
```diff
@@ -131,9 +131,13 @@ function Get-EnvironmentInformation
         $LinuxInfo = Get-Content /etc/os-release -Raw | ConvertFrom-StringData
 
         $environment += @{'LinuxInfo' = $LinuxInfo}
+        $environment += @{'IsDebian' = $LinuxInfo.ID -match 'debian'}
+        $environment += @{'IsDebian8' = $Environment.IsDebian -and $LinuxInfo.VERSION_ID -match '8'}
+        $environment += @{'IsDebian9' = $Environment.IsDebian -and $LinuxInfo.VERSION_ID -match '9'}
         $environment += @{'IsUbuntu' = $LinuxInfo.ID -match 'ubuntu'}
         $environment += @{'IsUbuntu14' = $Environment.IsUbuntu -and $LinuxInfo.VERSION_ID -match '14.04'}
         $environment += @{'IsUbuntu16' = $Environment.IsUbuntu -and $LinuxInfo.VERSION_ID -match '16.04'}
+        $environment += @{'IsUbuntu17' = $Environment.IsUbuntu -and $LinuxInfo.VERSION_ID -match '17.04'}
```
TravisEz13: build.psm1 LGTM

tools/appimage.sh
```diff
@@ -409,7 +409,7 @@ OPTIONS="-o Debug::NoLocking=1
 -o APT::Install-Suggests=0
 "
 
-cp ../powershell_*_amd64.deb .
+cp ../powershell_*ubuntu.14.04_amd64.deb .
```
TravisEz13: appimage.sh LGTM

TravisEz13 (approved):
One minor comment

tools/packaging/packaging.psm1
daxian-dbw: possible. I will use `switch` instead.

tools/packaging/packaging.psm1
```diff
@@ -246,255 +260,301 @@ function New-UnixPackage {
...
-        }
-        "osxpkg" {
-            if (!$Environment.IsMacOS) {
-                throw ($ErrorMessage -f "macOS")
-            }
+    DynamicParam {
+        if ($Type -eq "deb") {
+            # Add a dynamic parameter '-Distribution' when the specified package type is 'deb'.
+            # The '-Distribution' parameter can be used to indicate which Debian distro this pacakge is targeting.
+            $ParameterAttr = New-Object "System.Management.Automation.ParameterAttribute"
```
adityapatwardhan: use ```[System.Management.Automation.ParameterAttribute]::new()``` instead

tools/packaging/packaging.psm1
```diff
@@ -246,255 +260,301 @@ function New-UnixPackage {
...
+            $ValidateSetAttr = New-Object "System.Management.Automation.ValidateSetAttribute" -ArgumentList $Script:DebianDistributions
```
adityapatwardhan: Use ```[System.Management.Automation.ValidateSetAttribute]::new($DebianDistributions)``` instead

tools/packaging/packaging.psm1
```diff
@@ -246,255 +260,301 @@ function New-UnixPackage {
...
+            $Attributes = New-Object "System.Collections.ObjectModel.Collection``1[System.Attribute]"
```
adityapatwardhan: same as above.

tools/packaging/packaging.psm1
```diff
@@ -246,255 +260,301 @@ function New-UnixPackage {
...
+            $Attributes.Add($ParameterAttr) > $null
+            $Attributes.Add($ValidateSetAttr) > $null
+
+            $Parameter = New-Object "System.Management.Automation.RuntimeDefinedParameter" -ArgumentList ("Distribution", [string], $Attributes)
```
adityapatwardhan: Same as above

tools/packaging/packaging.psm1
```diff
@@ -246,255 +260,301 @@ function New-UnixPackage {
...
+            $Dict = New-Object "System.Management.Automation.RuntimeDefinedParameterDictionary"
```
adityapatwardhan: Same as above.

adityapatwardhan (changes requested):

tools/packaging/packaging.psm1
daxian-dbw: We use `New-Object` in build.psm1, and thus I used `New-Object` here to keep consistent.

adityapatwardhan (approved):

daxian-dbw:
@TravisEz13 @adityapatwardhan Thank you for the review!

Issue #4987, opened 2017-10-03
Remove ImportSystemModules command line parameter


iSazonov:
We should remove ImportSystemModules command line parameter because PowerShell Core don't care  about backward compatibility with Windows PowerShell 2.0.

iSazonov:
@mklement0 Could you please help - should we remove anything else here?

SteveL-MSFT:
@iSazonov I have a PR I'm already working on to remove `-sta`, `-mta`, FileConsole code, font changing code, and removing `#if CORECLR`.  I can also remove ImportSystemModules in the same PR.

PR #4995, opened 2017-10-03
Clean up console code

SteveL-MSFT:
These are features we haven't been supporting already (for PowerShell Core) and no plans to support as they exist for legacy reasons only for Windows PowerShell:

- removed sta/mta switches and code
- removed psconsolefile switch and code
- removed importsystemmodules switch and code
- removed font changing code
- merged `#if CORECLR` code

Fix https://github.com/PowerShell/PowerShell/issues/4987
Fix https://github.com/PowerShell/PowerShell/issues/4350


src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleShell.cs
```diff
@@ -25,18 +24,6 @@ public static int Start(string bannerText, string helpText, string[] args)
         {
             return Start(null, bannerText, helpText, null, args);
```
iSazonov: Do we use first parameter? Can we remove it?

SteveL-MSFT: The first param is RunspaceConfiguration.  It's being removed as part of https://github.com/PowerShell/PowerShell/pull/4942

iSazonov: Thanks! 

Closed.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -2937,23 +2893,7 @@ private class ConsoleHostStartupException : Exception
...
 
-#if CORECLR
         internal Lazy<TextReader> ConsoleIn { get; } = new Lazy<TextReader>(() => Console.In);
-#else
-        internal Lazy<TextReader> ConsoleIn { get; } = new Lazy<TextReader>(() =>
-        {
-            // This is a workaround for a full clr issue that causes a hang when calling PowerShell from ruby.
-            // They use named pipes instead of anonymous pipes, and for some reason that triggers a hang
-            // reading from Console.In.
-            var inputHandle = ConsoleControl.NativeMethods.GetStdHandle(-10);
```
PaulHigin: It seems like we should also remove NativeMethods.GetStdHandle() declaration since it is only used here.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfaceSecurity.cs
```diff
@@ -145,11 +145,6 @@ class ConsoleHostUserInterface : System.Management.Automation.Host.PSHostUserInt
 
         private IntPtr GetMainWindowHandle()
```
PaulHigin: Since we now always prompt using console we can get rid of this method since it will never be used.

PaulHigin: PromptForCredential() can be simplified since it now always prompts from console.  Also we can remove the HostUtilities.CredUIPromptForCredential() method since it also will no longer be used.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfaceSecurity.cs
```diff
@@ -163,42 +158,13 @@ private IntPtr GetMainWindowHandle()
             }
 
             return mainWindowHandle;
-#endif
         }
 
         // Determines whether we should prompt using the Console prompting
         // APIs
         private bool PromptUsingConsole()
```
PaulHigin: We can remove this method and simply the conditional code in the two places it is called.

PaulHigin (changes requested):

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleControl.cs
```diff
@@ -2830,20 +2727,7 @@ internal static void SetConsoleTextAttribute(ConsoleHandle consoleHandle, WORD a
...
             const uint OEM_CHARSET = 255;
-            // Suppress the PreFAST warning about not using Marshal.GetLastWin32Error() to
-            // get the error code.
-#pragma warning disable 56523
-#if CORECLR // TranslateCharsetInfo exists in an extension API set 'ext-ms-win-gdi-font-l1-1-1.dll', which is not available in NanoServer.
             return OEM_CHARSET;
-#else
-            CHARSETINFO csi;
-            const DWORD TCI_SRCCODEPAGE = 2;
-            if (!NativeMethods.TranslateCharsetInfo((IntPtr)codePage, out csi, TCI_SRCCODEPAGE))
```
PaulHigin: We should remove all unused NativeMethods declarations (TranslateCharsetInfo, GetConsoleWindow, etc.).

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleControl.cs
```diff
@@ -2830,20 +2727,7 @@ internal static void SetConsoleTextAttribute(ConsoleHandle consoleHandle, WORD a
         private static uint CodePageToCharSet(uint codePage)
         {
             const uint OEM_CHARSET = 255;
```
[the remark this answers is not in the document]
SteveL-MSFT: Makes sense, will change

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
SteveL-MSFT: Will remove

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfaceSecurity.cs
SteveL-MSFT: Agree, will remove

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfaceSecurity.cs
SteveL-MSFT: Will change

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleControl.cs
SteveL-MSFT: Will remove

PaulHigin (approved):
LGTM

iSazonov:
@SteveL-MSFT Is the PR ready to merge?

SteveL-MSFT:
@iSazonov since this is a breaking change (from Windows PowerShell), let me get an ack from @PowerShell/powershell-committee

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleControl.cs
```diff
@@ -1893,22 +1893,7 @@ private static bool ReadConsoleOutputCJKSmall
                     }
                     else
                     {
-                        int charLength = LengthInBufferCells(contents[r, c].Character);
-                        if (charLength == 2)
```
lzybkr: As I understand this code - somebody is screen scraping, there is a wide character (2 cells), and the font in use can't render the character, so conhost used 1 cell.
This seems like an extreme corner case, but the code seems correct to me and I don't any value in removing the code.
Also - the PR description should be improved - there is no link to an issue providing more detail.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleControl.cs
```diff
@@ -2944,103 +2783,9 @@ private static int LengthInBufferCellsFE(char c, ref HWND hwnd, ref HDC hDC, ref
...
-        /// <summary>
-        /// From IsConsoleFullWidth in \windows\core\ntcon\server\dbcs.c
-        /// </summary>
-        /// <param name="c"></param>
-        /// <param name="codePage"></param>
-        /// <returns></returns>
-        [SuppressMessage("Microsoft.Usage", "CA1806:DoNotIgnoreMethodResults",
-            MessageId = "Microsoft.PowerShell.ConsoleControl+NativeMethods.ReleaseDC(System.IntPtr,System.IntPtr)")]
-        private static int LengthInBufferCells(char c, uint codePage)
-        {
```
lzybkr: I think we still need LengthInBufferCells - but if you want to make a simplifying assumption when you can't know the font, you can use this [code](https://github.com/lzybkr/PSReadLine/blob/8edcf64663d7b7de3d9ef23435473e6529954b1e/PSReadLine/Render.cs#L512) from PSReadline.

SteveL-MSFT: I'll just use the code from PSReadline.  I ended up removing all this code as the fundamental assumption in the code is that we are using a specific codePage so all the calculations ended up returning a width of 1 anyways as all the other supporting code only applied to Windows PowerShell.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleControl.cs
SteveL-MSFT: I'll update the PR description as I did not originally intend for this PR to be so extensive.  I simply started down the path of removing the -mta/-sta code since we weren't using it and then noticed the ConsoleFile code which wasn't being used, then the Font code... ugh...

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -1682,32 +1657,19 @@ private void DoCreateRunspace(string initialCommand, bool skipProfiles, bool sta
...
-            // On NanoPS: always 'false' = default('false'=MTA) + NO possibility of overload through cmdline parameter '-mta'
-            // ThreadOptions should match on FullPS and NanoPS for corresponding ApartmentStates.
-            if (staMode)
-            {
-                // we can't change ApartmentStates on CoreCLR
-#if !CORECLR
-                runspace.ApartmentState = ApartmentState.STA;
-#endif
-                runspace.ThreadOptions = PSThreadOptions.ReuseThread;
-            }
```
lzybkr: Is this removal intentional? I think if that's not set, we create a new thread every time you execute a new command line.

SteveL-MSFT: Will put back the ReuseThread line

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
lzybkr: You should verify the hang doesn't repro in Core. It looks like I did not add a test (probably because the simplest repro required installing Ruby on Windows), but with all the changes in the CLR, it seems prudent to verify we still don't need this code before removing.

---
In reply to: [143062595](https://github.com/PowerShell/PowerShell/pull/4995#discussion_r143062595) [](ancestors = 143062595)

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostRawUserInterface.cs
```diff
@@ -1405,7 +1405,7 @@ int LengthInBufferCells(string s, int offset)
         public override
         int LengthInBufferCells(char c)
         {
-            return ConsoleControl.LengthInBufferCells(c);
+            return 1;
         }
```
lzybkr: This is wrong, plain and simple.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfacePrompt.cs
```diff
@@ -322,24 +322,12 @@ public override
                 // mind is lost. This can lead to a malicious
                 // server prompting the user for a request
                 // which can appear to come from locally.
-                if (!PromptUsingConsole() && desc.ModifiedByRemotingProtocol)
```
lzybkr: Deleting this code makes the comment above very confusing.
It also implies you are introducing a potential security risk.

lzybkr (changes requested):
There are several distinct changes in this PR, some of which don't feel like "clean up".

I think the commits should be cleaned up with better messages (reset and `git add -p`) and when merging, we should not squash this PR.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
SteveL-MSFT: Actually, this line was only within the `staMode` which was not enabled for the user to set.  So currently, we use the default which is to use a new thread for each pipeline.  Do we want to change this behavior?

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
SteveL-MSFT: This code wasn't being used since it's in the `#else` statement.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostRawUserInterface.cs
SteveL-MSFT: WIll fix!

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfacePrompt.cs
SteveL-MSFT: Will remove the comment since it only applied to the GUI prompt.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
lzybkr: `staMode = true` was the default even if the user couldn't set it, right?
At any rate, I do think we want to reuse the thread.

SteveL-MSFT: No, we defaulted to MTA in PSCore

lzybkr: I see. We defaulted to STA on Windows for many years, so reusing the thread seems preferable.

SteveL-MSFT:
@lzybkr cleaned up the commits to be distinct groups of work

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -1515,10 +1514,6 @@ private uint Run(string bannerText, string helpText, bool isPrestartWarned, stri
 
                 _runspaceRef.Runspace.Close();
                 _runspaceRef = null;
-                if (staMode) // don't recycle the Runspace in STA mode
-                {
-                    ShouldEndSession = true;
```
lzybkr: If I understand correctly - we default to STA and remove the option to use MTA, correct?

Assuming so, this line should probably stay, but I don't quite understand why it's here.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
lzybkr: @SteveL-MSFT - are you putting it back?

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
SteveL-MSFT: No, we defaulted to MTA and removing option to use STA

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleControl.cs
```diff
@@ -3321,6 +3050,39 @@ internal static void SetConsoleCursorInfo(ConsoleHandle consoleHandle, CONSOLE_C
 
 #endregion helper
 
+#region 
+
+        private static int LengthInBufferCells(char c)
+        {
+            if (c < 256)
```
lzybkr: This condition only applies to PSReadline.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
SteveL-MSFT: I need to verify something on Windows (was using a Mac for the changes) before I put it back.  With my other uber PR to remove apartmentstate all-up, the WSMan provider didn't work correctly.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
lzybkr: @SteveL-MSFT - I know. That doesn't change my feedback. The code was originally added for Full CLR, but with the significant work that went into Core CLR 2.0, I think we need to revisit the issue in case the code is needed in Core now.

If you remove this code, you'll lose the context, and if it's still an issue, investigation will be much more difficult because nobody will remember the change I made that introduced the problem in the first place.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
SteveL-MSFT: Looks good on my Windows box, will run through the tests and push the change.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleControl.cs
SteveL-MSFT: will remove

SteveL-MSFT:
@lzybkr I believe your feedback has been addressed, can you take another look?  Thanks

SteveL-MSFT:
@PowerShell/powershell-committee discussed this and was fine with all the changes except for removing the STA code.  Resolution is to leave the STA code commented out in case we have valid use cases on Windows on the future requiring support.

SteveL-MSFT:
I believe all feedback has been addressed.  @lzybkr can you take another look when you get a chance?  Thanks.

lzybkr:
>         {

This code is not called anymore, and is a slight variant of the version you added from PSReadLine below (which my version should be fine, it's more recent and has better comments.) :)


---
Refers to: src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleControl.cs:2736 in 6ca5d51. [](commit_id = 6ca5d517829587785df88ca7c9326adef577c11d, deletion_comment = False)

SteveL-MSFT:
@lzybkr removed that method

lzybkr (approved):

iSazonov:
@SteveL-MSFT Is the PR ready to merge?

SteveL-MSFT:
@iSazonov yes, please!

iSazonov:
It seems here we should preserve commits by create a merge commit?

SteveL-MSFT:
@iSazonov yes, I believe that is the case

Issue #4990, opened 2017-10-03
Commands fail when cd to a file share on windows


roberocity:
When working on a file share simple commands fail with Cannot find path error message.

Steps to reproduce
------------------

```powershell
cd \\someactual\fileshare\path

ls

rm *.zip

cd ..\another_path
```

Expected behavior
-----------------

```none
Changes working directory to file share path.

Lists the contents of the directory.

Removes any .zip file

Changes directory to another path on the share.
```

Actual behavior
---------------

```none
Changes working directory to file share path.

ls : Cannot find path '\someactual\fileshare\path' because it does not exist.
At line:1 char:1
+ ls
+ ~~
    + CategoryInfo          : ObjectNotFound: (\amcd12f02\devgp\releases\1.dev:String) [Get-ChildItem], ItemNotFoundException
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.GetChildItemCommand

Same Cannot find path error for all commands.
```

Environment data
----------------

```powershell
> $PSVersionTable


Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.7
OS                             Microsoft Windows 6.1.7601 S
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

```

KexyBiscuit:
It does not repro on Windows PowerShell 5.1, but it does repro on PowerShell 6.0.0-beta.7.

SteveL-MSFT:
This doesn't repro for me on Win10 and PSCore6 beta.7.

```powershell
PowerShell v6.0.0-beta.7
Copyright (C) Microsoft Corporation. All rights reserved.

PS C:\> cd \\share\scratch\steve
PS Microsoft.PowerShell.Core\FileSystem::\\share\scratch\steve>
```

SteveL-MSFT:
Does the current logged on user have permission to the share and the folder?

roberocity:
Yes, the user has permission to access the share. The user can run utility from a power shell prompt that  reads the files in the directory, manipulates, and saves them again. That works as expected, but the user can't then cd to another folder on the share. 

OS is Windows 7 Enterprise.

SteveL-MSFT:
[...]

PR #4998, opened 2017-10-03
enable using filesystem from a UNC location

SteveL-MSFT:
The problem is while validating the path, it splits `\\server\share\folder` by the backslash so the root is just `\` and the path ends up being `\server\share\folder` which isn't valid.  Fix is on Windows to add extra slash if the root is `\`

Fix https://github.com/PowerShell/PowerShell/issues/4990

### DIFF
--- a/src/System.Management.Automation/namespaces/NavigationProviderBase.cs
+++ b/src/System.Management.Automation/namespaces/NavigationProviderBase.cs
@@ -426,9 +426,10 @@ namespace System.Management.Automation.Provider
                     StringBuilder builder = new StringBuilder(parent, parent.Length + child.Length + 1);
 #if !UNIX
                     // On Windows, if we have a single backslash for the parent, it's a UNC path so we need to add another backslash back
-                    if (String.Compare(parent, StringLiterals.DefaultPathSeparatorString, StringComparison.Ordinal) == 0)
+                    if (this.ProviderInfo.FullName.Equals(@"Microsoft.PowerShell.Core\FileSystem", StringComparison.OrdinalIgnoreCase) &&
+                        String.Compare(parent, StringLiterals.DefaultPathSeparatorString, StringComparison.Ordinal) == 0)
                     {
-                        builder.Append('\\');
+                        builder.Append(StringLiterals.DefaultPathSeparator);
                     }
 #endif
 
--- a/src/System.Management.Automation/namespaces/NavigationProviderBase.cs
+++ b/src/System.Management.Automation/namespaces/NavigationProviderBase.cs
@@ -424,14 +424,6 @@ namespace System.Management.Automation.Provider
                     // Joins the paths
 
                     StringBuilder builder = new StringBuilder(parent, parent.Length + child.Length + 1);
-#if !UNIX
-                    // On Windows, if we have a single backslash for the parent, it's a UNC path so we need to add another backslash back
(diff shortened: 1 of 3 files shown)


src/System.Management.Automation/namespaces/NavigationProviderBase.cs
```diff
@@ -424,6 +424,13 @@ protected string MakePath(string parent, string child, bool childIsLeaf)
                     // Joins the paths
 
                     StringBuilder builder = new StringBuilder(parent, parent.Length + child.Length + 1);
+#if !UNIX
+                    // On Windows, if we have a single backslash for the parent, it's a UNC path so we need to add another backslash back
+                    if (String.Compare(parent, StringLiterals.DefaultPathSeparatorString, StringComparison.Ordinal) == 0)
+                    {
+                        builder.Append('\\');
```
anmenaga: Probably better to use DefaultPathSeparatorString instead of char constant.

anmenaga (changes requested):

SteveL-MSFT:
I don't like the change being outside of the filesystemprovider, let me see if I can rework this.

daxian-dbw:
@SteveL-MSFT yeah, it would be the best if we can fix the `GetParentPath` to take into account UNC paths.

SteveL-MSFT:
@daxian-dbw `GetParentPath` gets called repeatedly to validate the entire path from leaf to root.  I think putting the fix in NormalizePath() may be better.

test/powershell/Modules/Microsoft.PowerShell.Management/FileSystem.Tests.ps1
```diff
@@ -1301,3 +1301,19 @@ Describe "Extended FileSystem Path/Location Cmdlet Provider Tests" -Tags "Featur
         }
     }
 }
+
+Describe "UNC paths" -Tags 'CI' {
+    It "Can Get-ChildItems from a UNC location" -Skip:(!$IsWindows) {
```
iSazonov: Should we check Set-Location too?

iSazonov (commented):
I unlike the fix - it seems a performance is bad.
I do `cd \\localhost\c$` and discover that we call `NormalizePath` 6-9 times for the same path!

test/powershell/Modules/Microsoft.PowerShell.Management/FileSystem.Tests.ps1
SteveL-MSFT: Sure

SteveL-MSFT:
@iSazonov I've noticed that the providers get called a lot.  Perhaps GetParentPath() is the right place even thought that gets called a lot as well, but we only need to make it a UNC path again if it's at the root.

iSazonov:
I wonder why we have to call NormalizePath method so many times on the same path? Maybe we can fix this?

SteveL-MSFT:
@iSazonov I created https://github.com/PowerShell/PowerShell/issues/5002 as the perf work item

adityapatwardhan:
@anmenaga  Please update your review.

adityapatwardhan:
@anmenaga  ping...

anmenaga (approved):
LGTM.
Too many paths. :)

adityapatwardhan (approved):
LGTM

PR #5001, opened 2017-10-04
Cleanup Json cmdlets

iSazonov:
Related #4357.

Cleanup Json cmdlets from FullCLR code.
Remove ImportJsonDotNetModule.
Update Newton.Json version 10.0.2 -> 10.0.3


src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
```diff
@@ -205,157 +180,5 @@ private static ICollection<object> PopulateFromJArray(JArray list, out ErrorReco
...
-                {
-                    result.Properties.Add(new PSNoteProperty(entry.Key, entry.Value));
-                }
-            }
-
-            return result;
-        }
-#endif
     }
 }
```
SteveL-MSFT: Can you add a newline since you're already updating this file?

dantraMSFT (approved):
This looks good to me.
Is it safe to assume that ImportJsonDotNetModule hasn't been needed for a while?

SteveL-MSFT (approved):
LGTM

iSazonov:
>Is it safe to assume that ImportJsonDotNetModule hasn't been needed for a while?

I asked in Gitter and only @vors commented:
```
Ilya @iSazonov Sep 07 07:16 
I wonder why we explicitly load dll Newtonsoft.Json, Version=7.0.0.0 and why version is 7.0.0?

Sergei Vorobev @vors Sep 07 07:17 
@iSazonov it’s historical reason of the convertfrom-json port for nano server as far as I remember
version is whatever was latest available at the moment and went thru all the testing scrunities
hardcoded version also can be an attemp to allow modules to load a newer versions and not affect the core cmdlets
```
Also I guess it could have been because of the problems with loading dlls in the PowerShell Core early port phases.

I tried to list loaded assemblies and don't see Newton.Json 7.0 - only 10 version was loaded.

@vors @daxian-dbw Could you please comment too? Is it safe?

daxian-dbw:
Yes, it's safe to remove it.
Back in NanoServer time, the only way to ship Newton.json.dll is to wrap it into a module (Json.Net) and load that module when the Json cmdlet is in use. Now we don't even have the Json.Net module.

@iSazonov Can you please change the last commit message to add `[Feature]` to trigger the full test run? Some Json cmdlet tests are marked with `-Tags Feature`.

daxian-dbw:
There is one known failure. It failed due to a race condition between the test and the test service, not related to this PR. I will merge this PR.
```
[-] Validate Invoke-RestMethod error with environment proxy set - 'http_proxy' 5.29s
   Expected: {System.Threading.Tasks.TaskCanceledException,Microsoft.PowerShell.Commands.InvokeRestMethodCommand}
   But was:  {}
   at line: 1407 in /Users/travis/build/PowerShell/PowerShell/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
   1407:         $result.Error.FullyQualifiedErrorId | Should Be "System.Threading.Tasks.TaskCanceledException,Microsoft.PowerShell.Commands.InvokeRestMethodCommand"
```

Issue #3616, opened 2017-04-21
Implement Enable-SSHRemoting cmdlet


PaulHigin:
We need an Enable-SSHRemoting cmdlet, similar to existing Enable-PSSession cmdlet for WinRM based remoting, to set up SSH remoting on supported platforms.  This would make using SSH remoting easier to use and also allow automated testing of SSH remoting.

https://github.com/PowerShell/PowerShell-RFC/blob/master/1-Draft/RFC0012-Enable-SSH-Remoting.md

RichardSiddaway:
Agreed

Configuring SSH for remoting is tsk that is better automated if at all possible

joeyaiello:
Per conversation with @SteveL-MSFT, we might want to push this out depending on how the Win32-OpenSSH work lands. 

It might even make sense to push this work out to after the OpenSSH work lands to make sure that we don't end up having to do breaking changes if/when the configuration mechanisms for Win32-OpenSSH change.

SteveL-MSFT:
I think what we may end up doing is just having a setup script in the demo folder and we can blog about it to inform users of it's existence who don't work out of GitHub

PaulHigin:
I am working on a set up script right now as a prototype for the Cmdlet.  I don't think we should worry about Win32-OpenSSH breaking PowerShell remoting since it is intended to be closely compatible with other platform versions of SSH.

SteveL-MSFT:
I think we can publish this in PowerShellGallery initially and target inclusion in the package for 6.1.0

soul4soul:
Two years later I'm still hoping this is implemented. I just had to write PS scripts to do this while provisioning Windows 7 and Windows 10 PCs. For windows 10 I followed several articles on docs.microsoft.com [[1]](https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-windows?view=powershell-6#how-to-create-a-remoting-endpoint)[[2]](https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse). For Windows 7 I used chocolatey.

For configuring the keys I followed the guide at https://docs.microsoft.com/en-us/windows-server/administration/openssh/openssh_keymanagement. It'd be nice to see `Repair-AuthorizedKeyPermission` fixed with this effort too.

SteveL-MSFT:
@PaulHigin has a working version, just need to publish to PSGallery.  Won't be in PS7, but can be in 7.1.

[...]

PR #5009, opened 2017-10-04
Fix for impliict remoting test

PaulHigin:
Fix test setup that was using New-PSSession instead of the helper function New-RemoteSession that works with AppVeyor.  Re-enabled the pending test since it now works.  Also added check to ensure WinRM remoting is enabled and an endpoint is available for PowerShell 6.0.

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
@@ -14,7 +14,6 @@ if ($IsWindows)
         # -SkipNetworkProfileCheck is used in case Docker or another application
         # has created a publich virtual network profile on the system
         Enable-PSRemoting -SkipNetworkProfileCheck
-        $endpointCreated = $true
     }
 }
 
@@ -2109,7 +2108,7 @@ Describe "Import-PSSession on Restricted Session" -tags "Feature","RequireAdminO
         }
         else
         {
-            $configName = "restricted_102F"
+            $configName = "restricted_" + (Get-RandomFileName)
             New-PSSessionConfigurationFile -Path $TestDrive\restricted.pssc -SessionType RestrictedRemoteServer
             Register-PSSessionConfiguration -Path $TestDrive\restricted.pssc -Name $configName -Force
             $session = New-RemoteSession -ConfigurationName $configName
@@ -2132,8 +2131,15 @@ Describe "Import-PSSession on Restricted Session" -tags "Feature","RequireAdminO
     It "Verifies that Import-PSSession works on a restricted session" {
 
         $errorVariable = $null
-        $module = Import-PSSession -Session $session -AllowClobber -ErrorVariable $errorVariable -CommandName Get-Help
+        try
+        {
+            $module = Import-PSSession -Session $session -AllowClobber -ErrorVariable $errorVariable -CommandName Get-Help
+        }
+        finally
+        {
+            if ($module -ne $null) { Remove-Module $module -Force -ErrorAction SilentlyContinue }
+        }
+
         $errorVariable | Should BeNullOrEmpty
-        if ($module -ne $null) { Remove-Module $module -Force -ErrorAction SilentlyContinue }
     }
 }
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
@@ -1,4 +1,24 @@
-﻿Describe "Implicit remoting and CIM cmdlets with AllSigned and Restricted policy" -tags "Feature" {
+﻿#
+# Ensure WinRM remoting is enabled
+#
+if ($IsWindows)
+{
+    # Check to see if a default endpoint has been created
+    $endpointName = "PowerShell.$($psversiontable.GitCommitId)"
+    $matchedEndpoint = Get-PSSessionConfiguration $endpointName -ErrorAction SilentlyContinue
+
+    if ($matchedEndpoint -eq $null)
+    {
+        # An endpoint for this instance of PowerShell does not exist.
+        #
+        # -SkipNetworkProfileCheck is used in case Docker or another application
+        # has created a publich virtual network profile on the system
+        Enable-PSRemoting -SkipNetworkProfileCheck
+        $endpointCreated = $true
+    }
+}
+
+Describe "Implicit remoting and CIM cmdlets with AllSigned and Restricted policy" -tags "Feature" {
 
     BeforeAll {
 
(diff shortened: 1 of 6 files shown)


test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -1,4 +1,24 @@
-﻿Describe "Implicit remoting and CIM cmdlets with AllSigned and Restricted policy" -tags "Feature" {
+﻿#
+# Ensure WinRM remoting is enabled
+#
+if ($IsWindows)
```
SteveL-MSFT: You should have this in `BeforeAll`

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -2089,9 +2109,10 @@ Describe "Import-PSSession on Restricted Session" -tags "Feature","RequireAdminO
         }
         else
         {
+            $configName = "restricted_102F"
```
SteveL-MSFT: If you need it unique, why not use a guid?

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -2104,17 +2125,15 @@ Describe "Import-PSSession on Restricted Session" -tags "Feature","RequireAdminO
...
-    # Need a way to created restricted endpoint based on a different endpoint other than microsoft.powershell which points
-    # to Windows PowerShell 5.1
-    It "Verifies that Import-PSSession works on a restricted session" -Pending {
+    It "Verifies that Import-PSSession works on a restricted session" {
 
         $errorVariable = $null
-        Import-PSSession -Session $session -AllowClobber -ErrorVariable $errorVariable
+        $module = Import-PSSession -Session $session -AllowClobber -ErrorVariable $errorVariable -CommandName Get-Help
         $errorVariable | Should BeNullOrEmpty
+        if ($module -ne $null) { Remove-Module $module -Force -ErrorAction SilentlyContinue }
```
SteveL-MSFT: Since the `Remove-Module` is part of cleanup and not the test, convention is to have the test in a try{} and the cleanup in a finally{}.  No need to check if "$module -ne $null", in general, we just call `Remove-Module -Force -ErrorAction SilentlyContinue`

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
PaulHigin: I believe BeforeAll blocks must be inside a Describe block.  But this file has multiple Describe blocks and I want to enable PSRemoting only once.  This is the only way I know how to do this, but let me know if there is an alternative method.

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
PaulHigin: Fixed

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
PaulHigin: I feel the null check is needed since the Cmdlet will throw an exception on a null parameter.  Otherwise fixed.

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
SteveL-MSFT: I see, perhaps then, it might be better to create a helper function called `Enable-PSCoreRemoting` that wraps this and have each BeforeAll call that?

PaulHigin: It is not necessary to do this check for each remoting test ... the remoting tests are slow enough as they are.  This check should really only be done once per test run.  Do we have any set up code before we start CI or Feature tests?  That might be a better place to ensure remoting is enabled instead of having this check in each remoting test file.

SteveL-MSFT: The check itself shouldn't be that slow, right?  If the endpoint is created, it won't create it again.

PaulHigin: Get-PSSessionConfiguration is slow and in any case should not need to be done before each remoting test.  If the state of remoting changes during a test run then we need to fix the test that altered the state.  Note that this is the same pattern we use in other tests (see PSSessionConfiguration.Tests.ps1)

SteveL-MSFT: I'm ok with this.  Thanks.

SteveL-MSFT (approved):

SteveL-MSFT (changes requested):
It looks like the AppVeyor failure is due to the test not being run as elevated.  You will have to put it within the `Describe` block with a `RequiresAdminOnWindows` tag so that it gets run elevated.  Perhaps the solution here is to cache if remoting is enabled.

PaulHigin:
I didn't think of that.  Ok, will do.

PaulHigin:
I'll use the same patter as with PSSessionConfiguration.Tests.ps1

test/powershell/Modules/Microsoft.PowerShell.Utility/Implicit.Remoting.Tests.ps1
```diff
@@ -1,2120 +1,2105 @@
...
-        if ($null -ne $session) { Remove-PSSession $session -ErrorAction SilentlyContinue }
-    }
-
+﻿#
+# Skip all tests on non-windows and non-PowerShellCore and non-elevated platforms.
+#
+$originalDefaultParameterValues = $PSDefaultParameterValues.Clone()
+$originalWarningPreference = $WarningPreference
+$WarningPreference = "SilentlyContinue"
+$skipTest = ! ($IsWindows -and $IsCoreCLR -and (Test-IsElevated))
```
SteveL-MSFT: Most of these tests should work over PSRP/SSH, right?  Is there an issue already to enable our remoting tests over SSH?

PaulHigin: Yes #3904.  We may be able to selectively run existing remoting tests on WinRM or SSH after we have a way to automatically install and configure SSH on various platforms (#3616).

SteveL-MSFT (approved):

daxian-dbw:
There is one test failure:
```
 [-] Validate Invoke-WebRequest error with -Proxy option set - 'http_proxy' 5.33s
   Expected: {System.Threading.Tasks.TaskCanceledException,Microsoft.PowerShell.Commands.InvokeWebRequestCommand}
   But was:  {}
   at line: 558 in /Users/travis/build/PowerShell/PowerShell/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
   558:         $result.Error.FullyQualifiedErrorId | Should Be "System.Threading.Tasks.TaskCanceledException,Microsoft.PowerShell.Commands.InvokeWebRequestCommand"
```

@SteveL-MSFT Is this failure also due to a race condition between the test and the test service?  I think it's not related to this PR.

PaulHigin:
No, it is not related to this PR.  I am trying to rebase to see if a new version of the Invoke-WebRequest tests fixes the issue.

SteveL-MSFT:
@daxian-dbw yes, that is the race condition between the web service waiting on a delay and the test waiting on a timeout

Issue #3603, opened 2017-04-20
Suggestion: implement an exponentiation (power-of) operator


mklement0:
Currently, to get exponentiation (`$x` to the power of `$n`), you must use the .NET framework:

E.g., to calculate 2 to the power of 8:

    > [Math]::Pow(2, 8)
    256

It would be nice to have a dedicated operator, such as `**`, as in other languages (e.g., Python, Ruby, JavaScript, Perl):

     > 2 ** 8
     256

lzybkr:
I'm disinclined to support this operator. My thinking:

* C# doesn't have it, so does PowerShell really need it?
* An operator is potentially confusing - different languages use different tokens (is it `**` or `^`)
* The operator isn't used all that often, especially in typical PowerShell scripts
* `[Math]::Pow(2, 8)` is really readable.

mklement0:
Thanks for the feedback.

> C# doesn't have it, so does PowerShell really need it?
> The operator isn't used all that often, especially in typical PowerShell scripts

I can't speak to the real-world need, only that I keep missing it personally. Perhaps others can comment.

Frankly, I find the absence of an exponentiation operator from C# puzzling; [this blog post](https://blogs.msdn.microsoft.com/csharpfaq/2004/03/07/why-doesnt-c-have-a-power-operator/) from 2004 explains that it wasn't included because it is "fairly rare", which wasn't too well-received in the comments.

Aside from that, PowerShell already has operators that C# doesn't have, and it is a sufficiently different beast so that I wouldn't rule it out just based on that.

Perhaps PowerShell can lead by example and down the road we'll see an operator added to C# as well? :)

> An operator is potentially confusing - different languages use different tokens (is it ** or ^)

`^` indeed invites confusion, because it has [different semantics in some languages](https://en.wikipedia.org/wiki/Exponentiation#In_programming_languages).

`**`, by contrast, while not used by all languages, seems to have no other meaning, and, as stated, is used in several popular languages: Python, Ruby, JavaScript, Perl

> `[Math]::Pow(2, 8)` is really readable.

It is readable, but neither easy to _remember_ nor to _type_.

To me, a dedicated exponentiation operator seems like a natural complement to the arithmetic operators.

HumanEquivalentUnit:
Suggestion that this operator could be in the form:

    2 -pow 8

[...]

PR #5010, opened 2017-10-04
Implement exponent operator

Gimly:
Here's a first working draft of implementation for adding a Power operator, as discussed in issue #3603.

I decided to go with the ** option, as it's, I think, the most common power operator in the different languages that has it. For now, I've implemented only the operator working on numeric values, as a binary operator.

I have tried to follow the same type of implementation as how the multiply operator. It's just a draft and I'm willing to improve it until it's perfect and can me merged.

### DIFF
--- a/src/System.Management.Automation/engine/parser/Compiler.cs
+++ b/src/System.Management.Automation/engine/parser/Compiler.cs
@@ -4911,6 +4911,13 @@ namespace System.Management.Automation.Language
                     }
                     binder = PSBinaryOperationBinder.Get(ExpressionType.Multiply);
                     return DynamicExpression.Dynamic(binder, typeof(object), lhs, rhs);
+                case TokenKind.Power:
+                    if(lhs.Type == typeof(double) && rhs.Type == typeof(double))
+                    {
+                        return Expression.Power(lhs, rhs);
+                    }
+                    binder = PSBinaryOperationBinder.Get(ExpressionType.Power);
+                    return DynamicExpression.Dynamic(binder, typeof(object), lhs, rhs);
                 case TokenKind.Divide:
                     if (lhs.Type == typeof(double) && rhs.Type == typeof(double))
                     {
--- a/src/System.Management.Automation/engine/parser/token.cs
+++ b/src/System.Management.Automation/engine/parser/token.cs
@@ -414,6 +414,9 @@ namespace System.Management.Automation.Language
         /// <summary>The PS class base class and implemented interfaces operator ':'. Also used in base class ctor calls.</summary>
         Colon = 99,
 
+        /// <summary>The power of operator '**'.</summary>
+        PowerOf = 100,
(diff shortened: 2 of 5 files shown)


lzybkr (commented):
General feedback:

* We'll want committee approval for this.
* We could use `^` to be more like VB and some other languages.
* We probably don't need to return anything but `double`
* Precedence for exponentiation is usually higher than multiply
* There are no tests.

Arguments against taking this change:

* C# and C++ don't have this operator, so why would PowerShell
* Increased complexity in teaching for an infrequent operation, `[Math]::Pow()` is terse and descriptive.

iSazonov:
`[Math]` contains many functions - I don't see the point of starting the process of moving them into PowerShell language.

RichardSiddaway:
Agreed that this doesn't seem worth the effort. if the exponential operator is being added why not the square root and trigonometric operators - SIN, COS, TAN etc.  This functionality is readily available through [math] so why add it to PowerShell

Gimly:
While I agree that we already have access relatively easily to pow with the `[Math]::Pow()`, it makes the code much less readable when doing mathematical operations.

```PowerShell
$cSquared = [Math]::Pow($a, 2.0) + [Math]::Pow($b, 2.0)
```
is, in my opinion, much less readable than
```PowerShell
$cSquared = $a^2.0 + $b^2.0
```
Not sure that the fact that C# or C++ doesn't have the operator is that good of an argument, I'm sure a lot of people are missing it in those languages as well. I definitely did at multiple occasion when writing C# code.

But, I also agree that it might just bring more demand for other mathematical operators and I'm not sure if it's really PowerShell's main goal. In the end, the amount of projects who are doing advanced mathematics with PowerShell is probably very low.

Anyway, I did those changes as a small exercise to discover PowerShell's code ans also because the issue was marked with the Hacktoberfest label 😄. It was very instructive for me, so no hurt feelings if you decide that in the end it's not interesting to add this operator. I understood when I started that it was going to create philosophical questioning about the use of the operator (it had already started in the issue).

If you do decide that you'd like to have the operator, I'll be happy to update my code and fix all the issues that @lzybkr gave me in the feedback.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and decided not to add additional math operators.  Some of the concerns:
  - prefer to have one canonical way `Math.Pow()` rather than allowing multiple ways of doing the same thing
  - adding this operator incurs the cost of spending more time reviewing the semantics and setting precedent for other possible operators
  - `Math.Pow()` seems sufficient for the limited expected usage

We thank @Gimly for taking the time working on this, but perhaps pick up a different `Hacktoberfest` issue :)

Gimly:
@SteveL-MSFT thanks, I will definitely try my luck on a different Hacktoberfest issue, hopefully one a bit less controversial 😄.

PR #5021, opened 2017-10-05
Fix powershell to update the PATH environment variable only if PATH exists

### LINKED ISSUE #5019: Failure when being run with no PATH variable
CentOS 7.2, Powershell Core 6 Beta 8.

I'm attempting to call just a basic Powershell command (/usr/bin/powershell -Command "& {Write-Output 'Hello, world'}") in a script that is run by a software agent. No matter how the simple command (or PS1) file is passed, the error results similarly:

> FailFast: Object reference not set to an instance of an object.
> 
>    at System.Environment.FailFast(System.String, System.Exception)
>    at System.Environment.FailFast(System.String, System.Exception)
>    at Microsoft.PowerShell.UnmanagedPSEntry.Start(System.String, System.String[], Int32)
>    at Microsoft.PowerShell.ManagedPSEntry.Main(System.String[])
> /tmp/f1cbae6e-1431-4ae9-b63c-d40c007de59e/task.sh: line 3:  2366 Aborted                 /usr/bin/powershell -Command "& {Write-Output 'Hello, world'}"
> Error while executing script: Process exited with an error: 134 (Exit value: 134)

Prepending this command with strace -v yields the following output.
https://www.dropbox.com/s/4upjmfrf6x2zn8h/ps-coretask.stdlog?dl=0

After getting some help on the powershell-help channel on Slack, a review of the output as linked above showed no environmental variables being passed (HOME, USER, or PATH). This script is executed as root via an agent. Even passing the command as HOME=/root /usr/bin/powershell -Command "& {Write-Output 'Hello, world'}" still produces failure with the same message as shown above.

However, when manually logging into the system as root and executing either a PS1 script or this simple command, the result is always successful. An example of a successful strace -v dump from such an attempt can be found here. (Both of the strace dumps are too large for Pastebin and would flood this thread.)
https://www.dropbox.com/s/bfsxqfrndrd0yvp/strace-good-dump.txt?dl=0

Please let me know what other forensics I can provide that would be of assistance in isolating this issue and possibly determining a viable workaround.

SteveL-MSFT:
Linux service calling powershell may not have PATH set
only use PATH if it's not null
also fixed a test case using -TestCases without unique test name

Fix https://github.com/PowerShell/PowerShell/issues/5019

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
@@ -136,10 +136,13 @@ namespace Microsoft.PowerShell
 
             // put PSHOME in front of PATH so that calling `powershell` within `powershell` always starts the same running version
             string path = Environment.GetEnvironmentVariable("PATH");
-            string pshome = Utils.DefaultPowerShellAppBase;
-            if (!path.Contains(pshome))
+            if (path != null)
             {
-                Environment.SetEnvironmentVariable("PATH", pshome + Path.PathSeparator + path);
+                string pshome = Utils.DefaultPowerShellAppBase;
+                if (!path.Contains(pshome))
+                {
+                    Environment.SetEnvironmentVariable("PATH", pshome + Path.PathSeparator + path);
+                }
             }
 
             try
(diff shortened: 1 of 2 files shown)


anmenaga (approved):
LGTM

chipzoller:
There still seems to be an issue with this after being rolled into beta 9. For example, if PATH is not set, a test script fails when calling $myvar = hostname on Linux.

"hostname : The term 'hostname' is not recognized as the name of a cmdlet, funct
ion, script file, or operable program."

If PATH is set, the script executes correctly.

SteveL-MSFT:
PATH needs to be set for PowerShell to find exe's not in $PSHOME or current working directory, try setting $myvar to full path to `hostname`

Issue #5023, opened 2017-10-05
Fix or remove scripts that install powershell on openSUSE


daxian-dbw:
> **Note:** When installing the SDK, SUSE and OpenSUSE may report that nothing provides libcurl. libcurl should already be installed on supported versions of both distros. Run zypper search libcurl to confirm. The error will present 2 'solutions'. Choose 'Solution 2' to continue installing .NET Core.

A manual step is required for install powershell on openSUSE (same as `dotnet-sdk-2.0.0`), so the current scripts that automatically set up powershell on OpenSUSE don't work.
Two scripts are affected:
- tools/download.sh
   It will continue to install the beta.6 powershell openSUSE package.
- tools/installpsh-suse.sh
   It's already broken since beta.7, because we stopped publishing suse.42.1 package in beta.7 because openSUSE 42.1 has reached the end of the lifecycle.

So we need to update those two scripts or remove them.

daxian-dbw:
A related issue: #5023

DarwinJS:
@daxian-dbw - I would like to make an installer for amazon linux based on this script.
I see some zypper packages in here and amazon linux uses yum.

It *seems* like I should just be able to change that section to yum because the packages are still RPMs right?

Also, besides distro detection logic, can you think of anything else that would need adaptation?

daxian-dbw:
For Amazon Linux, you can use [community/amazonlinux/Dockerfile](https://github.com/PowerShell/PowerShell/blob/master/docker/community/amazonlinux/Dockerfile) as an example.
The RPM package doesn't work on Amazon Linux because Amazon Linux doesn't have the package 'openssl-libs'. See #5027 for more information.

PR #5024, opened 2017-10-05
Update docs for beta.8 release

daxian-dbw:
A manual step is required for install powershell on openSUSE (same as `dotnet-sdk-2.0.0`), so the current script that automatically set up powershell on OpenSUSE doesn't work. Filed #5023 to track this.
Two scripts are affected:
- tools/download.sh
   It will continue to install the beta.6 powershell openSUSE package.
- tools/installpsh-suse.sh
   It's already broken since beta.7, because we stopped publishing suse.42.1 package in beta.7 due its end of lifecycle.

### DIFF
--- a/docker/release/fedora24/Dockerfile
+++ b/docker/release/fedora24/Dockerfile
@@ -3,9 +3,9 @@ FROM fedora:24
 # TODO: Until a release of PowerShell for Fedora 24 is available,
 # this Dockerfile installs the CentOS 7 version for compatibility.
 
-ARG POWERSHELL_VERSION=6.0.0-beta.7
-ARG POWERSHELL_RELEASE=v6.0.0-beta.7
-ARG POWERSHELL_PACKAGE=powershell-6.0.0_beta.7-1.el7.x86_64.rpm
+ARG POWERSHELL_VERSION=6.0.0-beta.8
+ARG POWERSHELL_RELEASE=v6.0.0-beta.8
+ARG POWERSHELL_PACKAGE=powershell-6.0.0_beta.8-1.rhel.7.x86_64.rpm
 ARG IMAGE_NAME=microsoft/powershell:fedora24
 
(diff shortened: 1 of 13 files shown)


docker/release/fedora24/Dockerfile
```diff
@@ -3,9 +3,9 @@ FROM fedora:24
 # TODO: Until a release of PowerShell for Fedora 24 is available,
 # this Dockerfile installs the CentOS 7 version for compatibility.
 
-ARG POWERSHELL_VERSION=6.0.0-beta.7
-ARG POWERSHELL_RELEASE=v6.0.0-beta.7
-ARG POWERSHELL_PACKAGE=powershell-6.0.0_beta.7-1.el7.x86_64.rpm
+ARG POWERSHELL_VERSION=6.0.0-beta.8
+ARG POWERSHELL_RELEASE=v6.0.0-beta.8
+ARG POWERSHELL_PACKAGE=powershell-6.0.0_beta.8-1.rhel.7.x86_64.rpm
```
adityapatwardhan: According to Linux.md we support fedora 25 and 26. Should this file be called /fedora25/Dockerfile?

docker/release/fedora24/Dockerfile
```diff
@@ -3,9 +3,9 @@ FROM fedora:24
...
 ARG IMAGE_NAME=microsoft/powershell:fedora24
```
adityapatwardhan: Should this be fedora25?

adityapatwardhan (changes requested):

docker/release/fedora24/Dockerfile
daxian-dbw: fedora 24 has reached the end of lifecycle. I will remove this file and create new docker files for 25 and 26 in a separate PR.

adityapatwardhan: Sounds good.

adityapatwardhan (approved):

PR #5025, opened 2017-10-05
Replace CRLF with LF for some files in assets

daxian-dbw:
Those files are shown as modified on Linux. Replace CRLF with LF.


SteveL-MSFT:
@daxian-dbw for License.rtf, I think we should retain the CRLF EOLs since it shows up in the msi (not sure how it handles just LF).  I created a PR to not have git try to convert it https://github.com/PowerShell/PowerShell/pull/5020

daxian-dbw:
OK, I will revert the changes to .rtf.

SteveL-MSFT (approved):

PR #5026, opened 2017-10-05
Add char range overload to DotDot operator

IISResetMe:
This PR attempts to allow the DotDot operator (..) to generate
character ranges, in turn allowing shorthand syntax like:

    $AtoZ = -join('A'..'Z')

Since PowerShell has no shorthand syntax for char literals, we
test whether the left-hand-side is of type String, and subsequently
attempts to convert both operands to a char (in accordance with 
PowerShell operator overload behavior in general).

NOTE: 
1) This breaks a number of currently supported range operations, such
as implicit numerical conversion of strings:
'1'..'100'

2) This implementation is fairly naive, making no attempts to assert whether the input characters (or any codepoints in between) fall outside UCS-2 (ie. surrogates). Attempts to expand ranges bound by surrogate pairs are expected to fail and not covered by this implementation (I will leave this an exercise for someone more experienced with unicode hackery than myself)

I was unable to locate existing tests for the integer range function, please advise on where to add those.


src/System.Management.Automation/engine/runtime/Operations/NumericOps.cs
```diff
@@ -863,5 +863,34 @@ internal static object CompareIne(char lhs, char rhs)
             char secondAsUpper = char.ToUpperInvariant(rhs);
             return firstAsUpper != secondAsUpper ? Boxed.True : Boxed.False;
         }
+
+        internal static object[] Range(char start, char end)
```
lzybkr: What do you think about returning `char[]` instead?

src/System.Management.Automation/engine/runtime/Operations/NumericOps.cs
```diff
@@ -863,5 +863,34 @@ internal static object CompareIne(char lhs, char rhs)
...
+        {
+            int lower = (int)start;
+            int upper = (int)end;
+
+            if (lower == upper)
```
lzybkr: I think this special case is correctly handled by the code below and it happens rarely, so maybe we should remove it. (And I see you copied the code from the `int` version, so you could remove it there too.)

lzybkr (changes requested):
The code changes look fine - but we need tests.

src/System.Management.Automation/engine/runtime/Operations/NumericOps.cs
IISResetMe: Mainly did it this way to keep it similar to the existing code in Compiler.cs. I'm inclined to make a single Range method that takes the operands as Expression's, de-duplicate the loop that generates the actual sequence and then cast the entire thing to char[], int[] or long[] based on the lhs operand before returning. Will update the code later today

src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -4901,6 +4903,11 @@ public object VisitBinaryExpression(BinaryExpressionAst binaryExpressionAst)
                     return Expression.Call(CachedReflectionInfo.TypeOps_AsOperator, lhs.Cast(typeof(object)), rhs.Convert(typeof(Type)));
 
                 case TokenKind.DotDot:
+                    if(lhs.Type == typeof(string)){
```
iSazonov: Will this work:
```
'A'..'Z' | % { ... }
```

src/System.Management.Automation/engine/runtime/Operations/NumericOps.cs
```diff
@@ -863,5 +863,34 @@ internal static object CompareIne(char lhs, char rhs)
             char secondAsUpper = char.ToUpperInvariant(rhs);
             return firstAsUpper != secondAsUpper ? Boxed.True : Boxed.False;
         }
+
+        internal static object[] Range(char start, char end)
+        {
+            int lower = (int)start;
+            int upper = (int)end;
```
iSazonov: I wonder - is this ASCII Arithmetic work correctly for unicode?

src/System.Management.Automation/engine/parser/Compiler.cs
IISResetMe: Yes, this will produce the sequence A (0x41) through Z (0x5A)

src/System.Management.Automation/engine/runtime/Operations/NumericOps.cs
lzybkr: It won't work with surrogate pairs because those can't be represented in a `char`, but I don't think that is an interesting scenario.

iSazonov: Since we added `u{} should we detect surrogate pairs and write an error?

lzybkr: I guess we should error on invalid characters (half of the pair). We could also support surrogate pair ranges if we didn't convert the string to char immediately, instead analyzing the string more thoroughly. But I'm not sure I'd block this PR on either of those.

iSazonov: I guess it will work with traditional languages in common scenarios but before I saw issues from China and Japan mans - if the new operator does not matter to them I agree with you.

IISResetMe:
Oops, rebased powershell:master into the branch, looks like all parent commits are in the PR now, how do I clean it up? 😞

markekraus:
@IISResetMe did you only have the 2 commits? if so you could do this (assuming `PowerShell` is the name of your remote branch for the actual PowerShell repo). Make sure you stash or commit any workspace changes before doing this as they will be lost. I'm not a git expert, and maybe there is a better way. So maybe wait until someone more seasoned either "thumbs up" this or tells you just how wrong I am :)


```bash
git fetch
git checkout -b 20171015fix PowerShell/master
git cherry-pick 6e6dfffc2d0150b504585597ad4df997a174867a
git cherry-pick 6d6f20d28157ee49071f018051ac7b35be57bbb0
git checkout -b 20171015backup overload-DotDot-CharRange-1
git checkout overload-DotDot-CharRange-1
git reset --hard 20171015fix
git push --force
```

This will create a new "20171015fix" branch at the current master on this repo, then cherry-picks your 2 commits, then backs up your existing branch to `20171015backup`, and then hard reset your branch to the "20171015fix" branch. This will both "rebase" your branch on to master and add your commits with out the merge mess.  This will likely mess up any comment history here, but I guess it's already some what messed up.

IISResetMe (commented):
These were the most basic tests I could think of for the existing Range operator. All examples I could find in the about_* help files pass these.

IISResetMe:
Thanks @markekraus, that seems to have done the trick.

TravisEz13:
I agree with @lzybkr.  We really need tests.

TravisEz13:
Can you make sure `RangeOperator.Tests.ps1` is UTF8 without BOM?  It shows up as binary.

IISResetMe:
@TravisEz13 Done, https://github.com/IISResetMe/PowerShell/blob/34bc59046698d0de41bb5403a6105c7070b38a46/test/powershell/Language/Operators/RangeOperator.Tests.ps1 is UTF-8 (with updated tests) now

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
+Describe "Range Operator" -Tags CI {
+    Context "Range integer operations" {
+        It "Range operator generates arrays of integers" {
+            $Range = 5..8
+            $Range.count | Should Be 4
+            $Range[0] | Should BeOfType [int]
```
iSazonov: Please add tests for values - `$Range[0] | Should Be 5`

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+            $Range[1] | Should BeOfType [int]
+            $Range[2] | Should BeOfType [int]
+            $Range[3] | Should BeOfType [int]
+        }
+
+        It "Range operator accepts negative integer values" {
+            {
+                -8..-5
+            } |Should Not Throw
```
iSazonov: Please remove the should - next line do the same and Pester can catch a throw there correctly.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+
+            $Range = -8..-5            
+            $Range.count | Should Be 4
+            $Range[0] |Should Be -8
```
iSazonov: Formatting - please add space after |

Please add type check as in previous test.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+            $Range[1] |Should Be -7
+            $Range[2] |Should Be -6
+            $Range[3] |Should Be -5
+        }
+
+        It "Range operator support single-item sequences" {
+            {
+                0..0
+            } |Should Not Throw
```
iSazonov: Please remove.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+
+            $Range = 0..0
+            $Range.count | Should Be 1
+            $Range[0] | Should BeOfType [int]
+            $Range[0] | Should Be 0
+        }
+
+        It "Range operator works in ascending and descending order" {
```
iSazonov: Previous tests already test "ascending". I think we could test here only descending order for positive and negative numbers. But I believe we should replace some tests with one - "-2 .. 2" and "2 .. -2" - so we check transition through zero too.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+
+            $Range = 4..3
+            $Range.count | Should Be 2
+            $Range[0] | Should Be 4
+            $Range[1] | Should Be 3
+        }
+    }
+
+    Context "Character expansion" {
+
```
iSazonov: Please remove extra line.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+        It "Range operator generates an array of [char] from single-character string operands" {
```
iSazonov: I think we can remove the test and add type checks in next test.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+            $CharRange = 'b'..'a'
+            $CharRange.count | Should Be 2
+            $CharRange[0] | Should Be ([char]'b')
+            $CharRange[1] | Should Be ([char]'a')
+        }
+
+        It "Range operator works with 16-bit unicode characters" {
+            {
+                "$([char]0x0110)".."$([char]0x0114)"
+            } | Should Not Throw
```
iSazonov: Please remove.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+            $UnicodeRange = "$([char]0x0110)".."$([char]0x0114)"
```
iSazonov: Please use new syntax `u{0x0110}

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+            $UnicodeRange.count | Should Be 5
+            $UnicodeRange.Where({$_ -is [char]}).count | Should Be 5
```
iSazonov: Please check values too as above for numbers.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    Context "Range operator operand types" {
+
```
iSazonov: Please remove extra line.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+        It "Range operator works on [int]" {
```
iSazonov: Please remove the test - we do the checks above already.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+            $Range.Where({$_ -is [int]}).count | Should Be 10
+        }
+
+        It "Range operator works on [long]" {
+            {
+                ([long]1)..([long]10)
+            } | Should Not Throw
+            $Range = ([long]1)..([long]10)
+            $Range.count | Should Be 10
+            $Range.Where({$_ -is [int]}).count | Should Be 10
```
iSazonov: Is it really [int]? I think we should use values > int.MaxValue.
And please check values too.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+        }
+        
+        It "Range operator works on [bigint]" {
+            {
+                ([bigint]1)..([bigint]10)
+            } | Should Not Throw
+            $Range = ([bigint]1)..([bigint]10)
+            $Range.count | Should Be 10
+            $Range.Where({$_ -is [int]}).count | Should Be 10
```
iSazonov: Is it really [int]? I think we should use values > long.MaxValue.
And please check values too.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        It "Range operator works on [decimal]" {
+            {
+                1.2d..2.9d
+            } | Should Not Throw
+            $Range = 1.1d..9.9d
```
iSazonov: Why so much? I think enough `$Range = 1.1d..3.9d`

test/powershell/Language/Operators/RangeOperator.Tests.ps1
```diff
@@ -0,0 +1,135 @@
...
+            $Range[3] | Should Be 4
+            $Range[4] | Should Be 5
+            $Range[5] | Should Be 6
+            $Range[6] | Should Be 7
+            $Range[7] | Should Be 8
+            $Range[8] | Should Be 9
+            $Range[9] | Should Be 10
+        }
+    }
+}
```
iSazonov: Please add new line at EOF.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
IISResetMe: Current implementation doesn't support values > int.MaxValue, intended to add this in a separate branch/PR.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
IISResetMe: The existing numeric range operator only supports int regardless of input type

test/powershell/Language/Operators/RangeOperator.Tests.ps1
iSazonov: Thanks for clarify. Please remove the confuse tests.

TravisEz13:
@lzybkr Can you update your review?

TravisEz13:
@iSazonov Can you indicate if you think this is ready or not (feel free to say if someone else signs off)?

src/System.Management.Automation/engine/runtime/Operations/NumericOps.cs
iSazonov: @IISResetMe Is this addresed?

src/System.Management.Automation/engine/runtime/Operations/NumericOps.cs
iSazonov: @IISResetMe Could you please address the comment in code or add relevant comments in the PR description for future documentation?

test/powershell/Language/Operators/RangeOperator.Tests.ps1
iSazonov: @IISResetMe Please address the comment about the tests for [long] and [bigint].
We should remove its - they don't make sense.

test/powershell/Language/Operators/RangeOperator.Tests.ps1
iSazonov: Closed.

iSazonov:
@TravisEz13 I updated three comment above - after they have been resolved we'll ready to merge.

src/System.Management.Automation/engine/runtime/Operations/NumericOps.cs
IISResetMe: @iSazonov I don't feel confident enough in my unicode-fu to attempt to implement the appropriate handling in code for this at the moment, I've updated the PR description to reflect this deficiency instead.

src/System.Management.Automation/engine/runtime/Operations/NumericOps.cs
IISResetMe: I've removed the redundant branch in both overloads now. https://github.com/PowerShell/PowerShell/pull/5026/commits/0406cb37f1556b8497325236672819b54da23e53

test/powershell/Language/Operators/RangeOperator.Tests.ps1
IISResetMe: You're right, they don't make too much sense at the moment. I've removed them now

IISResetMe:
@iSazonov I believe I've addressed all outstanding comments :-)

src/System.Management.Automation/engine/runtime/Operations/NumericOps.cs
iSazonov: Thanks. 

Closed.

src/System.Management.Automation/engine/runtime/Operations/NumericOps.cs
iSazonov: @IISResetMe Could you please address the comment?

iSazonov:
@IISResetMe Many thanks! We are near to merge. Please address @lzybkr comment above.

src/System.Management.Automation/engine/runtime/Operations/NumericOps.cs
IISResetMe: I like the idea, but for now I think we should keep operator output consistent (ie. return `object[]` with individual items of the expected type).

There are are number of builtin operators that return `object[]` in this fashion (`@()`, `-split`, any comparison operator in array/filter mode etc.), changing the behavior should be a concerted effort across all of them IMO

IISResetMe:
@iSazonov Done

iSazonov:
@lzybkr Could you please take another look if you have a bit free time?

lzybkr (approved):

iSazonov (approved):

src/System.Management.Automation/engine/parser/Compiler.cs
kborowinski: Apparently it *does not work* on the Powershell Core 6.0.0 GA
```powershell
> 'A'..'Z' | % {$_ }
Cannot convert value "A" to type "System.Int32". Error: "Input string was not in a correct format."
At line:1 char:1
+ 'A'..'Z' | % {$_ }
+ ~~~~~~~~~~~~~~~~~~
+ FullyQualifiedErrorId : InvalidCastFromStringToInteger
```

markekraus: @kborowinski there is an open issue for that #5519

kborowinski: @markekraus: sorry, I've missed that issue

PR #5035, opened 2017-10-06
Fix Delay Test Race Condition

markekraus:
Race condition where WebListener is answering fast enough for the timeout not to trigger. Address CI issues in #4937 expands work from #5005 to other tests.

example: https://travis-ci.org/PowerShell/PowerShell/jobs/284145590

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -527,8 +527,8 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
     It "Invoke-WebRequest validate timeout option" {
 
-        $uri = Get-WebListenerUrl -Test 'Delay' -TestValue '6'
-        $command = "Invoke-WebRequest -Uri '$uri' -TimeoutSec 5"
+        $uri = Get-WebListenerUrl -Test 'Delay' -TestValue '5'
+        $command = "Invoke-WebRequest -Uri '$uri' -TimeoutSec 2"
 
         $result = ExecuteWebCommand -command $command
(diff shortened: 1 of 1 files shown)


test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -551,7 +551,7 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
     It "Validate Invoke-WebRequest error with -Proxy option set - '<name>'" -TestCases $testCase {
         param($proxy_address, $name, $protocol)
 
-        $uri = Get-WebListenerUrl -Test 'Delay' -TestValue '5' -Https:$($protocol -eq 'https')
+        $uri = Get-WebListenerUrl -Test 'Delay' -TestValue '6' -Https:$($protocol -eq 'https')
         $command = "Invoke-WebRequest -Uri '$uri' -TimeoutSec 5 -Proxy '${proxy_address}' -SkipCertificateCheck"
```
SteveL-MSFT: Since it appears that this test expects a timeout to happen, should we make timeoutsec less?  like 2 seconds?

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @SteveL-MSFT  I'm good with that. I will modify your previous change as well so they are all on the same settings. Are you thinking the `-TestValue '5'` and the `-TimeoutSec 2`?

SteveL-MSFT: I think delay of 5 and timeout of 2 makes sense

SteveL-MSFT (approved):

Issue #2983, opened 2017-01-09
Complete the SemanticVersion work 


HemantMahawar:
- [ ] Support SemVer in #Requires (PowerShell version (-Version) as well as module version (-Modules))
- [ ] Allows cmdlet parameters to consume SemVer instead of System.Version (argument transformation maybe?)
  - [ ] Users can convert Semver to System.Version using that attribute before passing to the cmdlet parameter
  - [ ] Cmdlet author can mark the attribute on the parameter and users can pass Semver string

=======================
**User Experience:**
- On disk, module versions presents - 1.0.0, 1.0.0-alpha17, 1.0.1 
- Result:

  | Command  | Expected Output | Comments |
  | -------------| ------------------ | -------------|
  | Import-Module -Version 1.0.2 | Error |  Version is alias to min version |
  | Import-Module -MinimumVersion 1.0.0 | Module with version 1.0.1 should be loaded |   |
  | Import-Module -Version 1.0.0-alpha15 | Module with version 1.0.1 should be loaded |  |
  | Import-Module -Name | Module with version 1.0.1 should be loaded | Latest version of module should be loaded |
  | Import-Module -RequiredVersion 1.0.0 | Module with version 1.0.0 should be loaded |  |

Note that 1.0.0 is higher than 1.0.0-alpha17, so if 1.0.1 doesn't exist, then `Import-Module` would load 1.0.0.

SteveL-MSFT:
We should wait for corefx support of semver (https://github.com/dotnet/corefx/issues/13526) rather than continue to have our implementation

iSazonov:
There is Set-StrictMode. Is it out of the enhancement?
```powershell
 gcm -ParameterName Version

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Cmdlet          Import-Module                                      7.0.0.0    Microsoft.PowerShell.Core
Cmdlet          Set-StrictMode                                     7.0.0.0    Microsoft.PowerShell.Core
```

Do we agree with breaking change that 1.0.0.1 is not SemVer?

SteveL-MSFT:
Moving out as I think there's a chance .NET 5 will have semver type

tnieto88:
At what point does PowerShell stop waiting for .NET and use NuGet version to support both Semver and traditional versions in the same type?

tnieto88:
[...]

PR #5037, opened 2017-10-06
Make SemanticVersion compatible with SemVer 2.0

### LINKED ISSUE #5004: SMA SemanticVersion not SemVer 2 compliant
First: System.Management.Automation.SemanticVersion is not SemVer v2 compliant

Second: PowerShell releases are using SemVer v2 syntax

Therefore: `$PSVersionTable.PSVersion` is truncated and thus *broken*, as pre-release versions cannot be compared to each other (the only useful assertion possible is that the PSVersion is `-lt "6.0.0"`)

You should be using (an implementation like) Nuget.Core's Nuget.SemanticVersion which handles `-prerelease` and `+patch` as well as dotted versions like your `6.0.0-beta.7`

Note that both the tests below return TRUE if $SemVer is Nuget.SemanticVersion:

Steps to reproduce
------------------

```powershell
$SemVer = [System.Management.Automation.SemanticVersion]
("6.0.0-beta.10" -as $SemVer) -gt ("6.0.0-beta.7" -as $SemVer)
("6.0.0+fixes" -as $SemVer) -gt ("6.0.0-beta" -as $SemVer)
```

Expected behavior
-----------------

```none
True
True
```

Actual behavior
---------------

```none
False
False
```

Environment data
----------------

```powershell
> $PSVersionTable
```

iSazonov:
Fix #5004 

### Fix description

Made compatible with SymVer 2.0 http://semver.org/ (p.10)
Fixed comparisons
Refactored tests 
Added new tests

### Additional information
Useful sample https://github.com/maxhauser/semver/blob/master/src/Semver/SemVersion.cs
Related #2983

### DIFF
--- a/src/System.Management.Automation/FormatAndOutput/DefaultFormatters/DotNetTypes_format_ps1xml.cs
+++ b/src/System.Management.Automation/FormatAndOutput/DefaultFormatters/DotNetTypes_format_ps1xml.cs
@@ -44,6 +44,14 @@ namespace System.Management.Automation.Runspaces
                 ViewsOf_System_Version());
 
             yield return new ExtendedTypeDefinition(
+                "System.Version#IncludeLabel",
+                ViewsOf_System_Version_With_Label());
+
+            yield return new ExtendedTypeDefinition(
+                "System.Management.Automation.SemanticVersion",
+                ViewsOf_Semantic_Version_With_Label());
+
+            yield return new ExtendedTypeDefinition(
                 "System.Drawing.Printing.PrintDocument",
                 ViewsOf_System_Drawing_Printing_PrintDocument());
 
@@ -474,6 +482,46 @@ namespace System.Management.Automation.Runspaces
                         .AddPropertyColumn("Minor")
                         .AddPropertyColumn("Build")
                         .AddPropertyColumn("Revision")
+                    .EndRowDefinition()
+                .EndTable());
+        }
+
+        private static IEnumerable<FormatViewDefinition> ViewsOf_System_Version_With_Label()
+        {
+            yield return new FormatViewDefinition("System.Version",
+                TableControl.Create()
+                    .AddHeader(width: 6)
+                    .AddHeader(width: 6)
+                    .AddHeader(width: 6)
+                    .AddHeader(width: 8)
+                    .AddHeader(width: 26)
+                    .AddHeader(width: 27)
+                    .StartRowDefinition()
+                        .AddPropertyColumn("Major")
+                        .AddPropertyColumn("Minor")
+                        .AddPropertyColumn("Build")
+                        .AddPropertyColumn("Revision")
+                        .AddPropertyColumn("PSSemanticVersionPreLabel")
+                        .AddPropertyColumn("PSSemanticVersionBuildLabel")
+                    .EndRowDefinition()
+                .EndTable());
+        }
+
+        private static IEnumerable<FormatViewDefinition> ViewsOf_Semantic_Version_With_Label()
+        {
+            yield return new FormatViewDefinition("System.Management.Automation.SemanticVersion",
+                TableControl.Create()
+                    .AddHeader(width: 6)
+                    .AddHeader(width: 6)
+                    .AddHeader(width: 6)
+                    .AddHeader(width: 9)
+                    .AddHeader(width: 11)
+                    .StartRowDefinition()
+                        .AddPropertyColumn("Major")
+                        .AddPropertyColumn("Minor")
+                        .AddPropertyColumn("Patch")
+                        .AddPropertyColumn("PreLabel")
+                        .AddPropertyColumn("BuildLabel")
                     .EndRowDefinition()
                 .EndTable());
         }
--- a/src/System.Management.Automation/FormatAndOutput/DefaultFormatters/DotNetTypes_format_ps1xml.cs
+++ b/src/System.Management.Automation/FormatAndOutput/DefaultFormatters/DotNetTypes_format_ps1xml.cs
@@ -501,8 +501,8 @@ namespace System.Management.Automation.Runspaces
                         .AddPropertyColumn("Minor")
                         .AddPropertyColumn("Build")
                         .AddPropertyColumn("Revision")
-                        .AddPropertyColumn("PSSemanticVersionPreLabel")
-                        .AddPropertyColumn("PSSemanticVersionBuildLabel")
+                        .AddPropertyColumn("PSSemVerPreReleaseLabel")
+                        .AddPropertyColumn("PSSemVerBuildLabel")
                     .EndRowDefinition()
                 .EndTable());
         }
@@ -520,7 +520,7 @@ namespace System.Management.Automation.Runspaces
                         .AddPropertyColumn("Major")
                         .AddPropertyColumn("Minor")
                         .AddPropertyColumn("Patch")
-                        .AddPropertyColumn("PreLabel")
+                        .AddPropertyColumn("PreReleaseLabel")
                         .AddPropertyColumn("BuildLabel")
                     .EndRowDefinition()
                 .EndTable());
--- a/src/System.Management.Automation/engine/PSVersionInfo.cs
+++ b/src/System.Management.Automation/engine/PSVersionInfo.cs
@@ -780,8 +780,20 @@ namespace System.Management.Automation
         /// </summary>
         public static int Compare(SemanticVersion versionA, SemanticVersion versionB)
         {
-            if (versionA == null) return -1;
-            return versionA.CompareTo(versionB);
+            if (versionA != null) {
+                if (versionB != null)
+                {
+                    return versionA.CompareTo(versionB);
+                }
+                return 1;
+            }
+
+            if (versionB != null)
+            {
+                return -1;
+            }
+
+            return 0;
         }
 
         /// <summary>
@@ -849,7 +861,7 @@ namespace System.Management.Automation
         /// </summary>
         public override int GetHashCode()
         {
-            return versionString.GetHashCode();
+            return this.ToString().GetHashCode();
         }
 
         /// <summary>
@@ -907,7 +919,15 @@ namespace System.Management.Automation
 
         private static int ComparePreLabel(string preLabel1, string preLabel2)
         {
-            // Symver 2.0 standard p.9 - Pre-release versions have a lower precedence than the associated normal version.
+            // Symver 2.0 standard p.9
+            // Pre-release versions have a lower precedence than the associated normal version.
+            // Comparing each dot separated identifier from left to right
+            // until a difference is found as follows:
+            //     identifiers consisting of only digits are compared numerically
+            //     and identifiers with letters or hyphens are compared lexically in ASCII sort order.
+            // Numeric identifiers always have lower precedence than non-numeric identifiers.
+            // A larger set of pre-release fields has a higher precedence than a smaller set,
+            // if all of the preceding identifiers are equal.
             if (String.IsNullOrEmpty(preLabel1)) { return String.IsNullOrEmpty(preLabel2) ? 0 : 1; }
             if (String.IsNullOrEmpty(preLabel2)) { return -1; }
 
--- a/src/System.Management.Automation/engine/PSVersionInfo.cs
+++ b/src/System.Management.Automation/engine/PSVersionInfo.cs
@@ -389,8 +389,8 @@ namespace System.Management.Automation
         private const string VersionSansRegEx = @"^(?<major>\d+)(\.(?<minor>\d+))?(\.(?<patch>\d+))?$";
         private const string LabelRegEx = @"^((?<preLabel>[0-9A-Za-z][0-9A-Za-z\-\.]*))?(\+(?<buildLabel>[0-9A-Za-z][0-9A-Za-z\-\.]*))?$";
         private const string LabelUnitRegEx = @"^[0-9A-Za-z][0-9A-Za-z\-\.]*$";
-        private const string PreLabelPropertyName = "PSSemanticVersionPreLabel";
-        private const string BuildLabelPropertyName = "PSSemanticVersionBuildLabel";
+        private const string PreLabelPropertyName = "PSSemVerPreReleaseLabel";
+        private const string BuildLabelPropertyName = "PSSemVerBuildLabel";
         private const string TypeNameForVersionWithLabel = "System.Version#IncludeLabel";
         private string versionString;
 
@@ -417,20 +417,20 @@ namespace System.Management.Automation
         /// <param name="major">The major version</param>
         /// <param name="minor">The minor version</param>
         /// <param name="patch">The patch version</param>
-        /// <param name="preLabel">The preLabel for the version</param>
-        /// <param name="buildLabel">The buildLabel for the version</param>
+        /// <param name="preReleaseLabel">The pre-release label for the version</param>
+        /// <param name="buildLabel">The build metadata for the version</param>
         /// <exception cref="FormatException">
-        /// If <paramref name="preLabel"/> don't match 'LabelUnitRegEx'.
+        /// If <paramref name="preReleaseLabel"/> don't match 'LabelUnitRegEx'.
         /// If <paramref name="buildLabel"/> don't match 'LabelUnitRegEx'.
         /// </exception>
-        public SemanticVersion(int major, int minor, int patch, string preLabel, string buildLabel)
+        public SemanticVersion(int major, int minor, int patch, string preReleaseLabel, string buildLabel)
             : this(major, minor, patch)
         {
-            if (!string.IsNullOrEmpty(preLabel))
+            if (!string.IsNullOrEmpty(preReleaseLabel))
             {
-                if (!Regex.IsMatch(preLabel, LabelUnitRegEx)) throw new FormatException(nameof(preLabel));
+                if (!Regex.IsMatch(preReleaseLabel, LabelUnitRegEx)) throw new FormatException(nameof(preReleaseLabel));
 
-                PreReleaseLabel = preLabel;
+                PreReleaseLabel = preReleaseLabel;
             }
 
             if (!string.IsNullOrEmpty(buildLabel))
@@ -665,7 +665,7 @@ namespace System.Management.Automation
 
             if (dashIndex > plusIndex)
             {
-                // 'preLabel' can contains dashes.
+                // 'PreReleaseLabel' can contains dashes.
                 if (plusIndex == -1)
                 {
                     // No buildLabel: buildLabel == null
@@ -675,7 +675,7 @@ namespace System.Management.Automation
                 }
                 else
                 {
-                    // No preLabel: preLabel == null
+                    // No PreReleaseLabel: preLabel == null
                     // Format is 'major.minor.patch+BuildLabel'
                     buildLabel = version.Substring(plusIndex+1);
                     versionSansLabel = version.Substring(0, plusIndex);
@@ -705,7 +705,7 @@ namespace System.Management.Automation
                 (plusIndex != - 1 && String.IsNullOrEmpty(buildLabel)) ||
                 String.IsNullOrEmpty(versionSansLabel))
             {
-                // We have dash and no preLabel or
+                // We have dash and no preReleaseLabel  or
                 // we have plus and no buildLabel or
                 // we have no main version part (versionSansLabel==null)
                 result.SetFailure(ParseFailureKind.FormatException);
@@ -780,12 +780,9 @@ namespace System.Management.Automation
         /// </summary>
         public static int Compare(SemanticVersion versionA, SemanticVersion versionB)
         {
-            if (versionA != null) {
-                if (versionB != null)
-                {
-                    return versionA.CompareTo(versionB);
-                }
-                return 1;
+            if (versionA != null)
+            {
+                return versionA.CompareTo(versionB);
             }
 
             if (versionB != null)
--- a/src/System.Management.Automation/engine/PSVersionInfo.cs
+++ b/src/System.Management.Automation/engine/PSVersionInfo.cs
@@ -386,9 +386,9 @@ namespace System.Management.Automation
     /// </summary>
     public sealed class SemanticVersion : IComparable, IComparable<SemanticVersion>, IEquatable<SemanticVersion>
     {
-        private const string s_VersionSansRegEx = @"^(?<major>\d+)(\.(?<minor>\d+))?(\.(?<patch>\d+))?$";
-        private const string s_LabelRegEx = @"^((?<preLabel>[0-9A-Za-z][0-9A-Za-z\-\.]*))?(\+(?<buildLabel>[0-9A-Za-z][0-9A-Za-z\-\.]*))?$";
-        private const string s_LabelUnitRegEx = @"^[0-9A-Za-z][0-9A-Za-z\-\.]*?$";
+        private const string VersionSansRegEx = @"^(?<major>\d+)(\.(?<minor>\d+))?(\.(?<patch>\d+))?$";
+        private const string LabelRegEx = @"^((?<preLabel>[0-9A-Za-z][0-9A-Za-z\-\.]*))?(\+(?<buildLabel>[0-9A-Za-z][0-9A-Za-z\-\.]*))?$";
+        private const string LabelUnitRegEx = @"^[0-9A-Za-z][0-9A-Za-z\-\.]*?$";
         private const string PreLabelPropertyName = "PSSemanticVersionPreLabel";
         private const string BuildLabelPropertyName = "PSSemanticVersionBuildLabel";
         private const string TypeNameForVersionWithLabel = "System.Version#IncludeLabel";
@@ -428,14 +428,14 @@ namespace System.Management.Automation
         {
             if (!string.IsNullOrEmpty(preLabel))
             {
-                if (!Regex.IsMatch(preLabel, s_LabelUnitRegEx)) throw new FormatException(nameof(preLabel));
+                if (!Regex.IsMatch(preLabel, LabelUnitRegEx)) throw new FormatException(nameof(preLabel));
 
                 PreLabel = preLabel;
             }
 
             if (!string.IsNullOrEmpty(buildLabel))
             {
-                if (!Regex.IsMatch(buildLabel, s_LabelUnitRegEx)) throw new FormatException(nameof(buildLabel));
+                if (!Regex.IsMatch(buildLabel, LabelUnitRegEx)) throw new FormatException(nameof(buildLabel));
 
                 BuildLabel = buildLabel;
             }
@@ -460,7 +460,7 @@ namespace System.Management.Automation
             // 2) 'label' starts with letter or digit.
             if (!string.IsNullOrEmpty(label))
             {
-                var match = Regex.Match(label, s_LabelRegEx);
+                var match = Regex.Match(label, LabelRegEx);
                 if (!match.Success) throw new FormatException(nameof(label));
 
                 PreLabel = match.Groups["preLabel"].Value;
@@ -713,7 +713,7 @@ namespace System.Management.Automation
                 return false;
             }
 
-            var match = Regex.Match(versionSansLabel, s_VersionSansRegEx);
+            var match = Regex.Match(versionSansLabel, VersionSansRegEx);
             if (!match.Success)
             {
                 result.SetFailure(ParseFailureKind.FormatException);
@@ -738,8 +738,8 @@ namespace System.Management.Automation
                 return false;
             }
 
-            if (preLabel != null && !Regex.IsMatch(preLabel, s_LabelUnitRegEx) ||
-               (buildLabel != null && !Regex.IsMatch(buildLabel, s_LabelUnitRegEx)))
+            if (preLabel != null && !Regex.IsMatch(preLabel, LabelUnitRegEx) ||
+               (buildLabel != null && !Regex.IsMatch(buildLabel, LabelUnitRegEx)))
             {
                 result.SetFailure(ParseFailureKind.FormatException);
                 return false;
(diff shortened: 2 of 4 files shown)


daxian-dbw:
@JamesWTruher @PaulHigin Can you please review this PR?

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -571,77 +659,127 @@ public static bool TryParse(string version, out SemanticVersion result)
...
+
+            string versionSansLabel = null;
+            var major=0;
+            var minor=0;
+            var patch=0;
+            string preLabel = null;
+            string buildLabel = null;
+
             var dashIndex = version.IndexOf('-');
+            var plusIndex = version.IndexOf('+');
```
PaulHigin: This parse code would be easier to read if it included comments with examples of expected version string.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -571,77 +659,127 @@ public static bool TryParse(string version, out SemanticVersion result)
...
-                return false;
+                if (plusIndex == -1)
+                {
+                    // No buildLabel - buildLabel == null;
+                    preLabel = version.Substring(dashIndex+1);
+                    versionSansLabel = version.Substring(0, dashIndex);
+                }
+                else
+                {
+                    // No preLabel - preLabel == null;
```
PaulHigin: Can you add a comment that a hyphen character ('-') is allowed as part of the buildLabel (metadata) per the spec?  BTW why do you use "buildLabel" instead of "metadata" or "metadataLable" as referred to by the spec?

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -571,77 +659,127 @@ public static bool TryParse(string version, out SemanticVersion result)
...
-            if (!Int32.TryParse(component, NumberStyles.Integer, CultureInfo.InvariantCulture, out parsedComponent))
+            if (match.Groups["patch"].Success && !int.TryParse(match.Groups["patch"].Value, out patch))
             {
-                result.SetFailure(ParseFailureKind.FormatException, component);
+                result.SetFailure(ParseFailureKind.FormatException);
                 return false;
             }
 
-            if (parsedComponent < 0)
+            if (preLabel != null && !Regex.IsMatch(preLabel, s_LabelUnitRegEx) ||
```
PaulHigin: I wonder if we should do this preLabel, buildLabel regex validation in the SemanticVersion constructor.  That way we can prevent malformed SemanticVersion objects from being created.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -748,7 +882,6 @@ public override int GetHashCode()
         /// </summary>
         public static bool operator <(SemanticVersion v1, SemanticVersion v2)
         {
-            if ((object)v1 == null) throw PSTraceSource.NewArgumentException(nameof(v1));
```
PaulHigin: I feel we still need this check, otherwise v1.CompareTo will throw null reference exception.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -757,7 +890,6 @@ public override int GetHashCode()
         /// </summary>
         public static bool operator <=(SemanticVersion v1, SemanticVersion v2)
         {
-            if ((object)v1 == null) throw PSTraceSource.NewArgumentException(nameof(v1));
```
PaulHigin: I feel we still need this check, otherwise v1.CompareTo will throw null reference exception.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -766,15 +898,50 @@ public override int GetHashCode()
         /// </summary>
         public static bool operator >(SemanticVersion v1, SemanticVersion v2)
         {
-            return (v2 < v1);
+            return (v1.CompareTo(v2) > 0);
```
PaulHigin: We should check v1 for null value.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -766,15 +898,50 @@ public override int GetHashCode()
...
         }
 
         /// <summary>
         /// Overloaded &gt;= operator
         /// </summary>
         public static bool operator >=(SemanticVersion v1, SemanticVersion v2)
         {
-            return (v2 <= v1);
+            return (v1.CompareTo(v2) >= 0);
```
PaulHigin: We should check v1 for null value.

PaulHigin (changes requested):
I am not quite finished with the review.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -384,12 +386,18 @@ IEnumerator IEnumerable.GetEnumerator()
     /// </summary>
     public sealed class SemanticVersion : IComparable, IComparable<SemanticVersion>, IEquatable<SemanticVersion>
     {
+        private static readonly string s_VersionSansRegEx = @"^(?<major>\d+)(\.(?<minor>\d+))?(\.(?<patch>\d+))?$";
```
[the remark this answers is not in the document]
iSazonov: Fixed.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -398,8 +406,47 @@ public SemanticVersion(string version)
...
+            PreLabel = v.PreLabel;
+            BuildLabel = v.BuildLabel;
+        }
+
+        /// <summary>
+        /// Construct a SemanticVersion.
+        /// </summary>
+        /// <param name="major">The major version</param>
+        /// <param name="minor">The minor version</param>
+        /// <param name="patch">The minor version</param>
```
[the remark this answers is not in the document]
iSazonov: Fixed.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -398,8 +406,47 @@ public SemanticVersion(string version)
...
+        {
+            if (!string.IsNullOrEmpty(preLabel))
+            {
+                if (preLabel.StartsWith('-') || preLabel.Contains("+")) throw new FormatException(nameof(preLabel));
+
+                PreLabel = preLabel;
+            }
+            else
+            {
+                PreLabel = null;
```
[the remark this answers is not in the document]
iSazonov: I only thought about symmetry and better readability.

Removed.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -410,17 +457,27 @@ public SemanticVersion(string version)
         /// <param name="patch">The minor version</param>
         /// <param name="label">The label for the version</param>
         /// <exception cref="PSArgumentException">
-        /// If <paramref name="major"/>, <paramref name="minor"/>, or <paramref name="patch"/> is less than 0.
-        /// </exception>
-        /// <exception cref="PSArgumentNullException">
-        /// If <paramref name="label"/> is null or an empty string.
+        /// <exception cref="FormatException">
+        /// If <paramref name="label"/> starts with '-'.
```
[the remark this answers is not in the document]
iSazonov: Complex RegEx is a puzzle for me too. I found some sample RegEx strings for SymVer but after testing I concluded that they all have bugs. 
So I split the SymVer regex string on some string (s_VersionSansRegEx, s_LabelRegEx, s_LabelUnitRegEx). This slightly increased the code but helped to avoid bugs.

Comment added.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -521,22 +602,29 @@ public SemanticVersion(Version version)
         public int Patch { get; }
 
         /// <summary>
-        /// The last component in a SemanticVersion - may be null if not specified.
+        ///
         /// </summary>
-        public string Label { get; }
+        //public string Label { get; }
```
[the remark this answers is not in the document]
iSazonov: Removed.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -521,22 +602,29 @@ public SemanticVersion(Version version)
...
+
+        public string PreLabel { get; }
```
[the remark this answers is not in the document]
iSazonov: Comments added.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Comment added.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Comments added and edited.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: I wonder too why I checked only first char in constructors :-) I wonted to be fast. We should use RegEx.

Fixed.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Fixed by new method Compare()

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -680,16 +822,14 @@ public int CompareTo(SemanticVersion value)
             if (Patch != value.Patch)
                 return Patch > value.Patch ? 1 : -1;
 
-            if (Label == null)
```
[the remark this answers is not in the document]
PaulHigin: Actually this PreLabel null check here can be removed since it is duplicated in ComparePreLabel().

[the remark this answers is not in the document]
PaulHigin: But please add the comment in ComparePreLabel() method.

PaulHigin (changes requested):

src/System.Management.Automation/engine/PSVersionInfo.cs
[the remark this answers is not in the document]
iSazonov: Duplicate checks was removed.
Comment was added.

iSazonov:
I've rebased to resolve a merge conflict.

daxian-dbw:
@anmenaga Can you please also review this PR? Thanks!

PaulHigin (approved):

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -571,77 +647,142 @@ public static bool TryParse(string version, out SemanticVersion result)
...
-                : StringUtil.Format("{0}.{1}.{2}-{3}", Major, Minor, Patch, Label);
+            return versionString;
+        }
+
+        /// <summary>
+        /// Implement Compare.
+        /// </summary>
+        public static int Compare(SemanticVersion versionA, SemanticVersion versionB)
+        {
+            if (versionA == null) return -1;
```
daxian-dbw: what about versionA and versionB are both null?

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -705,21 +839,18 @@ public override bool Equals(object obj)
...
         /// Override <see cref="object.GetHashCode()"/>
         /// </summary>
         public override int GetHashCode()
         {
-            return Utils.CombineHashCodes(
-                Major.GetHashCode(),
-                Minor.GetHashCode(),
-                Patch.GetHashCode(),
-                Label == null ? 0 : Label.GetHashCode());
+            return versionString.GetHashCode();
```
daxian-dbw: `versionString` could be null when running this method.
Besides, if we use the hashcode of  `versionString` for the SemanticVersion, we would have a hashcode collision, would that be a problem? Why not continue to use `Utils.CombineHashCodes`?

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -748,33 +879,67 @@ public override int GetHashCode()
...
+                var isNumber1 = Int32.TryParse(ac, out number1);
+                var isNumber2 = Int32.TryParse(bc, out number2);
+
+                if (isNumber1 && isNumber2)
+                {
+                    if (number1 != number2) { return number1 < number2 ? -1 : 1; }
+                }
+                else
+                {
+                    if (isNumber1) { return -1; }
```
daxian-dbw: It would be great if you can add comment here saying that "Numeric identifiers always have lower precedence than non-numeric identifiers. "

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -384,12 +386,18 @@ IEnumerator IEnumerable.GetEnumerator()
     /// </summary>
     public sealed class SemanticVersion : IComparable, IComparable<SemanticVersion>, IEquatable<SemanticVersion>
     {
+        private const string s_VersionSansRegEx = @"^(?<major>\d+)(\.(?<minor>\d+))?(\.(?<patch>\d+))?$";
+        private const string s_LabelRegEx = @"^((?<preLabel>[0-9A-Za-z][0-9A-Za-z\-\.]*))?(\+(?<buildLabel>[0-9A-Za-z][0-9A-Za-z\-\.]*))?$";
+        private const string s_LabelUnitRegEx = @"^[0-9A-Za-z][0-9A-Za-z\-\.]*?$";
```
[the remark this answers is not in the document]
iSazonov: `s_` was removed.
`?` was removed - I think it is bug in any case.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -398,8 +406,39 @@ public SemanticVersion(string version)
...
+        /// <summary>
+        /// Construct a SemanticVersion.
+        /// </summary>
+        /// <param name="major">The major version</param>
+        /// <param name="minor">The minor version</param>
+        /// <param name="patch">The patch version</param>
+        /// <param name="preLabel">The preLabel for the version</param>
+        /// <param name="buildLabel">The buildLabel for the version</param>
+        /// <exception cref="FormatException">
+        /// If <paramref name="preLabel"/> starts with '-' ot contains '+'.
```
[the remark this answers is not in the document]
iSazonov: I changed the comments because the code was changed previously.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -521,22 +591,28 @@ public SemanticVersion(Version version)
         public int Patch { get; }
 
         /// <summary>
-        /// The last component in a SemanticVersion - may be null if not specified.
+        /// PreLabel position in the SymVer string 'major.minor.patch-PreLabel+BuildLabel'.
+        /// </summary>
+
```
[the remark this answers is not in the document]
iSazonov: Done.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -521,22 +591,28 @@ public SemanticVersion(Version version)
...
+        public string PreLabel { get; }
```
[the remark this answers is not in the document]
iSazonov: Replaced with `PreReleaseLabel` - it looks more readable.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Good catch!

Fixed.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: null - fixed.

versionString is relatively short string - why you believe that CoreFX implementation of `String.GetHashCode()` so bad that we should make more complex hashcode?

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: I copy-paste all text from the standard.

src/System.Management.Automation/engine/PSVersionInfo.cs
daxian-dbw: What I mean is the hashcode of a SemanticsVersion instance would be the same of the hashcode of a string (the versionString), so there is a collision in terms that 2 different objects have the same hashcode.

src/System.Management.Automation/FormatAndOutput/DefaultFormatters/DotNetTypes_format_ps1xml.cs
```diff
@@ -478,6 +486,46 @@ private static IEnumerable<FormatViewDefinition> ViewsOf_System_Version()
...
+                    .AddHeader(width: 6)
+                    .AddHeader(width: 6)
+                    .AddHeader(width: 6)
+                    .AddHeader(width: 9)
+                    .AddHeader(width: 11)
+                    .StartRowDefinition()
+                        .AddPropertyColumn("Major")
+                        .AddPropertyColumn("Minor")
+                        .AddPropertyColumn("Patch")
+                        .AddPropertyColumn("PreLabel")
```
daxian-dbw: The property name has changed to 'PrereleaseLabel'.

src/System.Management.Automation/FormatAndOutput/DefaultFormatters/DotNetTypes_format_ps1xml.cs
```diff
@@ -478,6 +486,46 @@ private static IEnumerable<FormatViewDefinition> ViewsOf_System_Version()
...
+                    .AddHeader(width: 6)
+                    .AddHeader(width: 8)
+                    .AddHeader(width: 26)
+                    .AddHeader(width: 27)
+                    .StartRowDefinition()
+                        .AddPropertyColumn("Major")
+                        .AddPropertyColumn("Minor")
+                        .AddPropertyColumn("Build")
+                        .AddPropertyColumn("Revision")
+                        .AddPropertyColumn("PSSemanticVersionPreLabel")
```
daxian-dbw: Would it make sense to change this property name to `PSSemanticVersionPrereleaseLabel`?

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -398,8 +406,39 @@ public SemanticVersion(string version)
...
+        /// <param name="major">The major version</param>
+        /// <param name="minor">The minor version</param>
+        /// <param name="patch">The patch version</param>
+        /// <param name="preLabel">The preLabel for the version</param>
+        /// <param name="buildLabel">The buildLabel for the version</param>
+        /// <exception cref="FormatException">
+        /// If <paramref name="preLabel"/> don't match 'LabelUnitRegEx'.
+        /// If <paramref name="buildLabel"/> don't match 'LabelUnitRegEx'.
+        /// </exception>
+        public SemanticVersion(int major, int minor, int patch, string preLabel, string buildLabel)
```
daxian-dbw: Shall we rename `preLabel` to `prereleaseLabel` too? It's the parameter of a public method.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -384,12 +386,18 @@ IEnumerator IEnumerable.GetEnumerator()
     /// </summary>
     public sealed class SemanticVersion : IComparable, IComparable<SemanticVersion>, IEquatable<SemanticVersion>
     {
+        private const string VersionSansRegEx = @"^(?<major>\d+)(\.(?<minor>\d+))?(\.(?<patch>\d+))?$";
+        private const string LabelRegEx = @"^((?<preLabel>[0-9A-Za-z][0-9A-Za-z\-\.]*))?(\+(?<buildLabel>[0-9A-Za-z][0-9A-Za-z\-\.]*))?$";
+        private const string LabelUnitRegEx = @"^[0-9A-Za-z][0-9A-Za-z\-\.]*$";
+        private const string PreLabelPropertyName = "PSSemanticVersionPreLabel";
```
daxian-dbw: Same here. We probably should use `PSSemanticVersionPrereleaseLabel`, since it's for a public property.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -571,77 +646,154 @@ public static bool TryParse(string version, out SemanticVersion result)
...
+
+            return versionString;
+        }
+
+        /// <summary>
+        /// Implement Compare.
+        /// </summary>
+        public static int Compare(SemanticVersion versionA, SemanticVersion versionB)
+        {
+            if (versionA != null) {
```
daxian-dbw: Minor comment: it's better to follow the same open curly brace style in this file.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -571,77 +646,154 @@ public static bool TryParse(string version, out SemanticVersion result)
...
+                if (versionB != null)
+                {
+                    return versionA.CompareTo(versionB);
+                }
```
daxian-dbw: The instance method `CompareTo` handles null value in it, so this `versionB != null` check is not needed. It can just be:
```
if (versionA != null)
{
    return versionA.CompareTo(versionB);
}

if (versionB != null)
{
    return -1;
}

return 0;
```

daxian-dbw:
I still have a concern about the `GetHashCode()` method. See my new comment at https://github.com/PowerShell/PowerShell/pull/5037#discussion_r149441925

daxian-dbw:
@SteveL-MSFT This PR changes the public APIs, including:
- Add a public constructions `public SemanticVersion(int major, int minor, int patch, string preLabel, string buildLabel)`
- Remove public property `Label` and add two public properties `PreReleaseLabel` and `BuildLabel`.
- Add a public method `public static int Compare(SemanticVersion versionA, SemanticVersion versionB)`

Do you think it needs the committee review?

src/System.Management.Automation/engine/Types_Ps1Xml.cs
```diff
@@ -1982,6 +1982,22 @@ public static IEnumerable<TypeData> Get()
             var td251 = new TypeData(@"Deserialized.System.Management.Automation.DebuggerCommandResults", true);
             td251.TargetTypeForDeserialization = typeof(Microsoft.PowerShell.DeserializingTypeConverter);
             yield return td251;
+
+            var td252 = new TypeData(@"System.Version#IncludeLabel", true);
+            td252.Members.Add("ToString",
+                new ScriptMethodData(@"ToString", GetScriptBlock(@"
+          $suffix = """"
+          if (![String]::IsNullOrEmpty($this.PSSemanticVersionPreLabel))
```
daxian-dbw: This needs to be changed if the property name is changed.

SteveL-MSFT:
@daxian-dbw the general guidance is that for small api changes, the code review process is sufficient.  However, any maintainer can request Committee review if needed or if the contributors for the code review can't come to an agreement.

src/System.Management.Automation/FormatAndOutput/DefaultFormatters/DotNetTypes_format_ps1xml.cs
iSazonov: Looks very long 😕

iSazonov:
SemanticVersion API is new in PowerShell Core so until RTM we can change it free if we need.

daxian-dbw:
I'm fine with the API changes in this PR.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: In the [docs](https://msdn.microsoft.com/en-us/library/system.object.gethashcode%28v=vs.110%29.aspx) I don't found that the hash must be different for different objects of different types. It seems we follow the docs - it will be work well with Dictionary and Hashtable types.

src/System.Management.Automation/FormatAndOutput/DefaultFormatters/DotNetTypes_format_ps1xml.cs
iSazonov: Maybe ` PSSemVerPreReleaseLabel `?

daxian-dbw: That looks good to me, especially `semver` is the type accelerator for `SemanticVersion`.

iSazonov: Fixed.

src/System.Management.Automation/FormatAndOutput/DefaultFormatters/DotNetTypes_format_ps1xml.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Renamed.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/PSVersionInfo.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/PSVersionInfo.cs
daxian-dbw: Maybe I'm just overthinking it. Let's keep it as is.

.vscode/tasks.json
```diff
@@ -1,6 +1,6 @@
 {
     "version": "0.1.0",
-    "command": "pwsh",
+    "command": "powershell",
```
daxian-dbw: Please revert this change.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -530,10 +530,10 @@ public SemanticVersion(Version version)
             Minor = version.Minor;
             Patch = version.Build == -1 ? 0 : version.Build;
             var psobj = new PSObject(version);
-            var preLabelNote = psobj.Properties[PreLabelPropertyName];
-            if (preLabelNote != null)
+            var PreReleaseLabelNote = psobj.Properties[PreLabelPropertyName];
```
daxian-dbw: A local variable name should start with a lower-case letter.
In fact, I think you can continue to use `preLabel` for any non-public members.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -387,10 +387,10 @@ IEnumerator IEnumerable.GetEnumerator()
     public sealed class SemanticVersion : IComparable, IComparable<SemanticVersion>, IEquatable<SemanticVersion>
     {
         private const string VersionSansRegEx = @"^(?<major>\d+)(\.(?<minor>\d+))?(\.(?<patch>\d+))?$";
-        private const string LabelRegEx = @"^((?<preLabel>[0-9A-Za-z][0-9A-Za-z\-\.]*))?(\+(?<buildLabel>[0-9A-Za-z][0-9A-Za-z\-\.]*))?$";
+        private const string LabelRegEx = @"^((?<PreReleaseLabel>[0-9A-Za-z][0-9A-Za-z\-\.]*))?(\+(?<buildLabel>[0-9A-Za-z][0-9A-Za-z\-\.]*))?$";
```
daxian-dbw: Same here, I think it's OK to use `preLabel` here.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -656,7 +656,7 @@ private static bool TryParseVersion(string version, ref VersionResult result)
             var major=0;
             var minor=0;
             var patch=0;
-            string preLabel = null;
+            string PreReleaseLabel = null;
```
daxian-dbw: Same here. Local variable should start with lower case letter.

src/System.Management.Automation/engine/PSVersionInfo.cs
```diff
@@ -917,7 +914,7 @@ public override int GetHashCode()
             return (Compare(v1, v2) >= 0);
         }
 
-        private static int ComparePreLabel(string preLabel1, string preLabel2)
+        private static int ComparePreLabel(string PreReleaseLabel1, string PreReleaseLabel2)
```
daxian-dbw: Same here. parameter names.

daxian-dbw (commented):
It looks to me you just simply replaced `preLabel` with `PreReleaseLabel` and didn't care about the local variables and parameter names. Local variable names and parameter names should start with a lower case letter. I will fix it for you.

daxian-dbw (approved):

daxian-dbw:
The test failure in AppVeyor is a known issue and will be fixed by #5379 
Travis CI passed, but it's not updated here.

iSazonov:
😕  Sorry, I forgot to set VS Code case-sensitive flag to search and replace.

daxian-dbw:
The macOS CI was shown as passed yesterday after clicking the Travis CI "Details" link. I have no idea why it's shown as canceled today. Now we seem to have a problem with macOS CI -- it takes forever to start.

Given that I know the Travis CI previously passed, I will merge this PR.

Issue #5031, opened 2017-10-06
v6-beta.8: Potential regression of Remove-Item with wildcards and missing items


nightroman:
In all previous PowerShell versions (from v2 to v6-beta.7) `Remove-Item`
ignores missing items if they are specified by wildcards, i.e.

```powershell
Remove-Item [z] -Force -Recurse
Remove-Item z.* -Force -Recurse
```

work fine if `z` or `z.*` is missing.

In v6-beta.8 the above commands fail.

Is this by design or a regression?

```
PS> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.0.0-beta.8
PSEdition                      Core
GitCommitId                    v6.0.0-beta.8
OS                             Microsoft Windows 10.0.10586
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

nightroman:
The change https://github.com/PowerShell/PowerShell/pull/4866 says

> This means that using -literalpath with an asterisk and the filesystem provider will now return an error rather than returning quietly.

This is probably fine for `-literalpath`, using wildcards with it looks like a user error.

But with `-path` wildcards are rather expected and may point to missing items.

nightroman:
NB: new failure `remove-item -LiteralPath z.*` (which looks reasonable) is not consistent with `get-item -LiteralPath z.*` which still ignores wildcards.

SteveL-MSFT:
@nightroman can you open a separate issue for `Get-Item -LiteralPath`?  `-Path` behavior shouldn't have changed, will look into this.

nightroman:
> can you open a separate issue for Get-Item -LiteralPath?

https://github.com/PowerShell/PowerShell/issues/5057

PR #5038, opened 2017-10-06
fix detection of whether `-LiteralPath` was used to suppress wildcardexpansion for navigation cmdlets

SteveL-MSFT:
Fix https://github.com/PowerShell/PowerShell/issues/5031


adityapatwardhan (approved):
LGTM

adityapatwardhan:
@SteveL-MSFT The test 'Remove-Item -LiteralPath should succeed for file with asterisk in name' is failing. Please fix.

SteveL-MSFT:
@adityapatwardhan I only ran it on my Mac, I need to skip it on Windows since asterisk isn't allowed in filename

adityapatwardhan:
@SteveL-MSFT  Please add the test as described in the issue #5031. That should work on Windows as well.

SteveL-MSFT:
@adityapatwardhan added more test variations for the wildcard not finding files scenario

Issue #1755, opened 2016-08-12
ConvertFrom-Json fails to parse project.lock.json


SteveL-MSFT:
```powershell
$json = gc project.lock.json -Raw
ConvertFrom-Json $json
```

Expected: PowerShell should be able to parser the json data.
Actual: A useless error message which does not tell me what went wrong

```
ConvertFrom-Json : Cannot process argument because the value of argument "name" is not valid. Change the value of the
"name" argument and run the operation again.
At line:1 char:1
- ConvertFrom-Json $json
- ~~~~~~~~~~~~~~~~~~~~~~
  - CategoryInfo          : InvalidArgument: (:) [ConvertFrom-Json], PSArgumentException
  - FullyQualifiedErrorId : Argument,Microsoft.PowerShell.Commands.ConvertFromJsonCommand
```

kittholland:
Bug is due to empty name for json dictionary entry:

Example from the project.lock.json:

  "projectFileDependencyGroups": {
    "": [
      "System.Management.Automation >= 6.0.0-*"
    ],

thrown from line 4474 in MshMemberInfo.cs 

                    if (String.IsNullOrEmpty(name))
                    {
                        throw PSTraceSource.NewArgumentException("name");
                    }

"Cannot process argument because the value of argument \"name\" is not valid. Change the value of the \"name\" argument and run the operation again."

Reproducible with the following simple json:

{
	"": "Test"
}

SteveL-MSFT:
@kittholland thanks for looking into this, best as I can tell per http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf, an empty string as the name is allowed and works in javascript.  I guess the next step is to determine if the issue is in newtonsoft json parser.

kittholland:
From what I could see in the debugger, newtonsoft was OK with it, but converting the Jobject into a PSObject failed because the name was empty.

SteveL-MSFT:
Ok, that makes sense (on why it failed).  A HashTable will accept an empty string key, so it seems we should have PSObject support it as well.  @kittholland are you going to work on a fix?

kittholland:
I have not looked deeply into what would be required, but I can give it a try.

joeyaiello:
We should also handle the `Invoke-WebRequest`/`Invoke-RestMethod` scenario where you're able to post a null JSON key as a string or as an object.

joeyaiello:
[...]

Issue #3623, opened 2017-04-22
Add option to ConvertFrom-Json as a Hashtable


kittholland:
Currently ConvertFrom-Json returns a PSCustomObject. It would be nice in some cases if you could choose to return a Hashtable instead, perhaps using an -As parameter. Hashtables are often much faster for processing Json document databases compared to having to iterate through object properties.

This would also be a possible way to address #1755 .

BrucePay:
Can you provide a concrete example where hashtables are superior to an object tree?

powercode:
Apart from object size and performance, it may matter with case sensitivity. I have run into issues where I couldn't use the PowerShell cmdlets because the json contained 'Property' and 'property'.

kittholland:
Yes, the case sensitivity is one area, empty property names are another area.

On the performance side, I would suggest that adding, removing and enumerating properties are all slower/more difficult on a pscustomobject than a hashtable.

```
$hashTable = @{
    "1" = 1
    "2" = 2
}

$object = [pscustomobject]$hashTable
```

```
Measure-command {
    Foreach($number in 3..10000)
    {
        $hashTable.Add("$number", $number)
    }
}



Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 0
Milliseconds      : 63
Ticks             : 631145
TotalDays         : 7.30491898148148E-07
TotalHours        : 1.75318055555556E-05
TotalMinutes      : 0.00105190833333333
TotalSeconds      : 0.0631145
TotalMilliseconds : 63.1145
```

```
Measure-Command {
    Foreach($number in 3..10000)
    {
        $object | Add-Member -MemberType NoteProperty -Name "$number" -Value $number
    }
}


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 1
Milliseconds      : 76
Ticks             : 10769741
TotalDays         : 1.24649780092593E-05
TotalHours        : 0.000299159472222222
TotalMinutes      : 0.0179495683333333
TotalSeconds      : 1.0769741
TotalMilliseconds : 1076.9741

```

KevinMarquette:
I like the idea of an -AsHashtable option or something similar. Whenever I enumerate all the property names and then walk them, it is very inefficient.  Even more so when I am recursively walking the object to turn it into a nested hashtable. 

There is also not a simple way to test for a property on an object. Sometimes you can't just test for `$null` because the value could be `$null`. The existence of a property is different than a `$null` property value.

[...]

PR #5043, opened 2017-10-06
Add -AsHashtable switch to ConvertFrom-Json for issue #3623

bergmeister:
Address https://github.com/PowerShell/PowerShell/issues/3623.

Implementation symmetric to traditional code path that returns a `PSCustomObject`. Code is shared on the top level but 2 low level methods were difficult to share, therefore 2 separate but similar methods were created for HashTables.
All existing tests related to this change were adapted to also test against this new switch and were slightly improved. 

Update (@TravisEz13):
`-AsHashtable` is an existing pattern used in `Group-Object`

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
@@ -171,7 +171,7 @@ namespace Microsoft.PowerShell.Commands
             try
             {
                 ErrorRecord error;
-                obj = JsonObject.ConvertFromJson(json, false, out error);
+                obj = JsonObject.ConvertFromJson(json, out error);
 
                 if (error != null)
                 {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
@@ -171,7 +171,7 @@ namespace Microsoft.PowerShell.Commands
             try
             {
                 ErrorRecord error;
-                obj = JsonObject.ConvertFromJson(json, out error);
+                obj = JsonObject.ConvertFromJson(json, false, out error);
 
                 if (error != null)
                 {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertFromJsonCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertFromJsonCommand.cs
@@ -34,7 +34,6 @@ namespace Microsoft.PowerShell.Commands
         /// <summary>
         /// Returned data structure is a Hashtable instead a CustomPSObject.
         /// </summary>
-        [Parameter(Mandatory = false)]
         public SwitchParameter AsHashtable { get; set; }
 
         #endregion parameters
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertFromJsonCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertFromJsonCommand.cs
@@ -34,6 +34,7 @@ namespace Microsoft.PowerShell.Commands
         /// <summary>
         /// Returned data structure is a Hashtable instead a CustomPSObject.
         /// </summary>
+        [Parameter()]
         public SwitchParameter AsHashtable { get; set; }
 
         #endregion parameters
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertFromJsonCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertFromJsonCommand.cs
@@ -20,16 +20,22 @@ namespace Microsoft.PowerShell.Commands
         #region parameters
 
         /// <summary>
-        /// gets or sets the InputString property
+        /// Gets or sets the InputString property.
         /// </summary>
         [Parameter(Mandatory = true, Position = 0, ValueFromPipeline = true)]
         [AllowEmptyString]
         public string InputObject { get; set; }
 
         /// <summary>
-        /// inputObjectBuffer buffers all InputObjet contents available in the pipeline.
+        /// InputObjectBuffer buffers all InputObject contents available in the pipeline.
         /// </summary>
         private List<string> _inputObjectBuffer = new List<string>();
+
+        /// <summary>
+        /// Returned data structure is a Hashtable instead a CustomPSObject.
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        public SwitchParameter AsHashtable { get; set; }
 
         #endregion parameters
 
@@ -44,7 +50,7 @@ namespace Microsoft.PowerShell.Commands
         }
 
         /// <summary>
-        /// the main execution method for the convertfrom-json command
+        /// The main execution method for the ConvertFrom-Json command.
         /// </summary>
         protected override void EndProcessing()
         {
@@ -95,7 +101,7 @@ namespace Microsoft.PowerShell.Commands
         private bool ConvertFromJsonHelper(string input)
         {
             ErrorRecord error = null;
-            object result = JsonObject.ConvertFromJson(input, out error);
+            object result = JsonObject.ConvertFromJson(input, AsHashtable.IsPresent, out error);
 
             if (error != null)
             {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
@@ -13,6 +13,7 @@ using Newtonsoft.Json.Linq;
 using System.Collections;
 using System.Collections.ObjectModel;
 using System.IO;
+using System.Linq;
 using System.Management.Automation.Internal;
 using System.Reflection;
 
@@ -133,14 +134,30 @@ namespace Microsoft.PowerShell.Commands
                     return null;
                 }
 
+                // Case sensitive duplicates should normally not occur since JsonConvert.DeserializeObject
+                // does not throw when encountering duplicates and just uses the last entry.
+                if (result.Properties.Any(psPropertyInfo => psPropertyInfo.Name.Equals(entry.Key, StringComparison.InvariantCulture)))
+                {
+                    string errorMsg = string.Format(CultureInfo.InvariantCulture,
+                                                    WebCmdletStrings.DuplicateKeysInJsonString, entry.Key);
+                    error = new ErrorRecord(
+                        new InvalidOperationException(errorMsg),
+                        "DuplicateKeysInJsonString",
+                        ErrorCategory.InvalidOperation,
+                        null);
+                    return null;
+                }
+
+                // Compare case insensitive to tell the user to use the -AsHashTable option instead.
+                // This is because PSObject cannot have keys with different casing.
                 PSPropertyInfo property = result.Properties[entry.Key];
                 if (property != null)
                 {
                     string errorMsg = string.Format(CultureInfo.InvariantCulture,
-                                                    WebCmdletStrings.DuplicateKeysInJsonString, property.Name, entry.Key);
+                                                    WebCmdletStrings.KeysWithDifferentCasingInJsonString, property.Name, entry.Key);
                     error = new ErrorRecord(
                         new InvalidOperationException(errorMsg),
-                        "DuplicateKeysInJsonString",
+                        "KeysWithDifferentCasingInJsonString",
                         ErrorCategory.InvalidOperation,
                         null);
                     return null;
@@ -228,10 +245,12 @@ namespace Microsoft.PowerShell.Commands
             Hashtable result = new Hashtable();
             foreach (var entry in entries)
             {
+                // Case sensitive duplicates should normally not occur since JsonConvert.DeserializeObject
+                // does not throw when encountering duplicates and just uses the last entry.
                 if (result.ContainsKey(entry.Key))
                 {
                     string errorMsg = string.Format(CultureInfo.InvariantCulture,
-                                                    WebCmdletStrings.DuplicateKeysInJsonString, entry.Key, entry.Key);
+                                                    WebCmdletStrings.DuplicateKeysInJsonString, entry.Key);
                     error = new ErrorRecord(
                         new InvalidOperationException(errorMsg),
                         "DuplicateKeysInJsonString",
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
@@ -121,6 +121,18 @@ namespace Microsoft.PowerShell.Commands
             PSObject result = new PSObject();
             foreach (var entry in entries)
             {
+                if (string.IsNullOrEmpty(entry.Key))
+                {
+                    string errorMsg = string.Format(CultureInfo.InvariantCulture,
+                                                    WebCmdletStrings.EmptyKeyInJsonString);
+                    error = new ErrorRecord(
+                        new InvalidOperationException(errorMsg),
+                        "EmptyKeyInJsonString",
+                        ErrorCategory.InvalidOperation,
+                        null);
+                    return null;
+                }
+
                 PSPropertyInfo property = result.Properties[entry.Key];
                 if (property != null)
                 {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
@@ -27,6 +27,18 @@ namespace Microsoft.PowerShell.Commands
         private const int maxDepthAllowed = 100;
 
         /// <summary>
+        /// Convert a Json string back to an object of type PSObject.
+        /// </summary>
+        /// <param name="input"></param>
+        /// <param name="error"></param>
+        /// <returns>A PSObject.</returns>
+        [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly")]
+        public static object ConvertFromJson(string input, out ErrorRecord error)
+        {
+            return ConvertFromJson(input, false, out error);
+        }
+
+        /// <summary>
         /// Convert a Json string back to an object of type PSObject or Hashtable depending on parameter <paramref name="returnHashTable"/>.
         /// </summary>
         /// <param name="input"></param>
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
@@ -10,6 +10,7 @@ using System.Management.Automation;
 using System.Text.RegularExpressions;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using System.Collections;
 using System.Collections.ObjectModel;
 using System.IO;
 using System.Management.Automation.Internal;
@@ -26,13 +27,14 @@ namespace Microsoft.PowerShell.Commands
         private const int maxDepthAllowed = 100;
 
         /// <summary>
-        /// Convert a Json string back to an object
+        /// Convert a Json string back to an object of type PSObject or Hashtable depending on parameter <paramref name="returnHashTable"/>.
         /// </summary>
         /// <param name="input"></param>
+        /// <param name="returnHashTable"></param>
         /// <param name="error"></param>
-        /// <returns></returns>
+        /// <returns>A PSObject or a Hashtable if the <paramref name="returnHashTable"/> parameter is true.</returns>
         [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly")]
-        public static object ConvertFromJson(string input, out ErrorRecord error)
+        public static object ConvertFromJson(string input, bool returnHashTable, out ErrorRecord error)
         {
             if (input == null)
             {
@@ -65,7 +67,14 @@ namespace Microsoft.PowerShell.Commands
                 var dictionary = obj as JObject;
                 if (dictionary != null)
                 {
-                    obj = PopulateFromJDictionary(dictionary, out error);
+                    if (returnHashTable)
+                    {
+                        obj = PopulateHashTableFromJDictionary(dictionary, out error);
+                    }
+                    else
+                    {
+                        obj = PopulateFromJDictionary(dictionary, out error);
+                    }
                 }
                 else
                 {
@@ -73,7 +82,14 @@ namespace Microsoft.PowerShell.Commands
                     var list = obj as JArray;
                     if (list != null)
                     {
-                        obj = PopulateFromJArray(list, out error);
+                        if (returnHashTable)
+                        {
+                            obj = PopulateHashTableFromJArray(list, out error);
+                        }
+                        else
+                        {
+                            obj = PopulateFromJArray(list, out error);
+                        }
                     }
                 }
             }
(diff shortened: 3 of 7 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
```diff
@@ -26,13 +27,14 @@ public static class JsonObject
...
+        /// Convert a Json string back to an object of type PSObject or Hashtable depending on parameter <paramref name="returnHashTable"/>.
         /// </summary>
         /// <param name="input"></param>
+        /// <param name="returnHashTable"></param>
         /// <param name="error"></param>
-        /// <returns></returns>
+        /// <returns>A PSObject or a Hashtable if the <paramref name="returnHashTable"/> parameter is true.</returns>
         [SuppressMessage("Microsoft.Naming", "CA1704:IdentifiersShouldBeSpelledCorrectly")]
-        public static object ConvertFromJson(string input, out ErrorRecord error)
+        public static object ConvertFromJson(string input, bool returnHashTable, out ErrorRecord error)
```
iSazonov: We shouldn't change a public API. Please add new overload.

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
```diff
@@ -1,46 +1,33 @@
...
-        @{ name = "empty"; str = "" }
-        @{ name = "spaces"; str = "  " }
-        @{ name = "object"; str = "{a:1}" }
+        @{ name = "empty";  str = "";      ReturnHashTable = $true  }
+        @{ name = "spaces"; str = "  ";    ReturnHashTable = $true  }
+        @{ name = "object"; str = "{a:1}"; ReturnHashTable = $true  }
+        @{ name = "empty";  str = "";      ReturnHashTable = $false }
+        @{ name = "spaces"; str = "  ";    ReturnHashTable = $false }
+        @{ name = "object"; str = "{a:1}"; ReturnHashTable = $false }
     )
```
iSazonov: Please move to BeforeAll block.

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
```diff
@@ -1,46 +1,33 @@
...
 
-    It 'no error for valid string - <name>' -TestCase $validStrings {
-        param ($str)
+    It 'no error for valid string - <name> when ReturnHashTable is <ReturnHashTable>' -TestCase $validStrings {
```
iSazonov: Maybe 'no error for valid string - <name> with -ReturnHashTable:`$\<ReturnHashTable>'

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertFromJsonCommand.cs
```diff
@@ -20,17 +20,23 @@ public class ConvertFromJsonCommand : Cmdlet
...
         /// <summary>
-        /// inputObjectBuffer buffers all InputObjet contents available in the pipeline.
+        /// InputObjectBuffer buffers all InputObject contents available in the pipeline.
         /// </summary>
         private List<string> _inputObjectBuffer = new List<string>();
 
+        /// <summary>
+        /// Returned data structure is a Hashtable instead a CustomPSObject.
+        /// </summary>
+        [Parameter(Mandatory = false)]
```
iSazonov: It is default. Please remove.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
```diff
@@ -65,15 +67,29 @@ public static object ConvertFromJson(string input, out ErrorRecord error)
...
-                        obj = PopulateFromJArray(list, out error);
+                        if (returnHashTable)
+                        {
+                            obj = PopulateHashTableFromJArray(list, out error);
+                        }
+                        else
+                        {
+                            obj = PopulateFromJArray(list, out error);
+                        }
                     }
```
iSazonov: Please add 'else' with Diagnostics.Assert (if the object is not JObject or JArray).

iSazonov (commented):
@bergmeister Thanks for great contribution!

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
bergmeister: Thanks. I did not know PowerShell also exposes some of its classes. I'll re-add the old signature API  that will then internally call the new overload.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
bergmeister: Because of the cast to `JArray` a few lines above, `obj` cannot be something else. I initially put your suggestion in but even the compiler tells me that:
`error CS0184: The given expression is never of the provided ('JObject') type`

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
bergmeister: I am not sure what you exactly mean by that. I removed the redundant line about `$TestCasesForReturnHashTableParameter that was a leftover.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertFromJsonCommand.cs
bergmeister: What do you mean by that? Can you elaborate please (what is 'it')?

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
bergmeister: I changed to match your suggestion whilst still displaying the name of the string.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertFromJsonCommand.cs
TravisEz13: `[Parameter(Mandatory = false)]` is default

bergmeister: Ok. I see, thanks.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
iSazonov: I meant:
```c#
                        Dbg.Diagnostics.Assert(
                            false,
                            "Only JObject or JArray is supported.");

```

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
iSazonov: Please see [sample](https://github.com/PowerShell/PowerShell/blob/309bf603f9aff9009dce7e725d42b2d4f99f58c9/test/powershell/Modules/Microsoft.PowerShell.Management/Clear-Item.Tests.ps1#L2) 
We should place ` $TestCasesForReturnHashTableParameter ` and `$validStrings ` in the BeforeAll block.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
```diff
@@ -171,7 +171,7 @@ private bool TryConvertToJson(string json, out object obj, ref Exception exRef)
             try
             {
                 ErrorRecord error;
-                obj = JsonObject.ConvertFromJson(json, out error);
+                obj = JsonObject.ConvertFromJson(json, false, out error);
```
iSazonov: Please revert the change - we have the overload.

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
```diff
@@ -1,46 +1,31 @@
...
         $errRecord | Should BeNullOrEmpty
     }
 
     $invalidStrings = @(
-        @{ name = "plain text"; str = "plaintext" }
-        @{ name = "part"; str = '{"a" :' }
+        @{ name = "plain text"; str = "plaintext"; ReturnHashTable = $true  }
+        @{ name = "part";       str = '{"a" :';    ReturnHashTable = $true  }
+        @{ name = "plain text"; str = "plaintext"; ReturnHashTable = $false }
+        @{ name = "part";       str = '{"a" :';    ReturnHashTable = $false }
```
iSazonov: I see we have a check for `input == null` but haven't the test - please add new test with `str = $null`.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
bergmeister: Yes, you're right.

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
bergmeister: OK. Added

TravisEz13:
`-AsHashtable` is an existing pattern used in `Group-Object`

iSazonov:
We have discussion for this https://github.com/PowerShell/PowerShell/issues/3623

TravisEz13:
@dantraMSFT Can you review this?

SteveL-MSFT:
As part of this PR, if we can detect that the json has an property named empty string for the pscustomobject path and provide an error message like:

```markdown
The provided json includes a property who name is an empty string, this is only supported using the `-AsHashTable` switch.
```

Then we can resolve https://github.com/PowerShell/PowerShell/issues/1755 as fixed.

src/Microsoft.PowerShell.Commands.Utility/resources/WebCmdletStrings.resx
```diff
@@ -126,9 +126,12 @@
   <data name="CredentialConflict" xml:space="preserve">
     <value>The cmdlet cannot run because the following conflicting parameters are specified: Credential and UseDefaultCredentials. Specify either Credential or UseDefaultCredentials, then retry.</value>
   </data>
-  <data name="DirecotryPathSpecified" xml:space="preserve">
+  <data name="DirectoryPathSpecified" xml:space="preserve">
     <value>Path '{0}' resolves to a directory. Specify a path including a file name, and then retry the command.</value>
   </data>
+  <data name="EmptyKeyInJsonString" xml:space="preserve">
+    <value>The provided json includes a property whose name is an empty string, this is only supported using the -AsHashTable switch.</value>
```
SteveL-MSFT: Should probably capitalize json as `JSON`

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/resources/WebCmdletStrings.resx
bergmeister: OK. Fixed it.

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
```diff
@@ -1,46 +1,47 @@
...
+        @{ name = "object"; str = "{a:1}"; ReturnHashTable = $true  }
+        @{ name = "null";   str = $null;   ReturnHashTable = $false }
+        @{ name = "empty";  str = "";      ReturnHashTable = $false }
+        @{ name = "spaces"; str = "  ";    ReturnHashTable = $false }
+        @{ name = "object"; str = "{a:1}"; ReturnHashTable = $false }
     )
 
-    It 'no error for valid string - <name>' -TestCase $validStrings {
-        param ($str)
+    It 'no error for valid string ''<name>'' with -ReturnHashTable:$<ReturnHashTable>' -TestCase $validStrings {
```
iSazonov: Please start test descriptions with capital.

We use `$validStrings` one time. Please use our common template:
```powershell
It 'no error for valid string ''<name>'' with -ReturnHashTable:$<ReturnHashTable>' -TestCase @(
 @{ name = "null";   str = $null;   ReturnHashTable = $true  }
 ...
) {
...
}
```

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
```diff
@@ -1,46 +1,47 @@
...
-        @{ name = "part"; str = '{"a" :' }
+        @{ name = "plain text"; str = "plaintext"; ReturnHashTable = $true  }
+        @{ name = "part";       str = '{"a" :';    ReturnHashTable = $true  }
+        @{ name = "plain text"; str = "plaintext"; ReturnHashTable = $false }
+        @{ name = "part";       str = '{"a" :';    ReturnHashTable = $false }
     )
 
-    It 'throw ArgumentException for invalid string - <name>' -TestCase $invalidStrings  {
-        param ($str)
+    It 'throw ArgumentException for invalid string ''<name>'' with -ReturnHashTable:$<ReturnHashTable>' -TestCase $invalidStrings  {
```
iSazonov: The same.

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
```diff
@@ -1,46 +1,47 @@
...
+        param ($str, $ReturnHashTable)
         $errRecord = $null
-        { [Microsoft.PowerShell.Commands.JsonObject]::ConvertFromJson($str, [ref]$errRecord) } | ShouldThrow 'ArgumentException'
+        { [Microsoft.PowerShell.Commands.JsonObject]::ConvertFromJson($str, $ReturnHashTable, [ref]$errRecord) } | ShouldBeErrorId "ArgumentException"
+    }
+
+    $jsonWithEmptyKey = '{"": "Value"}'
```
iSazonov: We should put variables in `BeforeAll` block or in `It` block.

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
```diff
@@ -1,46 +1,47 @@
...
+    It 'throw InvalidOperationException when json contains empty key name' {
+        $errorRecord = $null
+        [Microsoft.PowerShell.Commands.JsonObject]::ConvertFromJson($jsonWithEmptyKey, [ref]$errorRecord)
+        $errorRecord.Exception.GetType() | Should Be System.InvalidOperationException
```
iSazonov: Usually we don't check an exception types. Please remove.
(For type check we use Should BeOfType)

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
```diff
@@ -1,46 +1,47 @@
...
+        $errorRecord.FullyQualifiedErrorId | Should Be EmptyKeyInJsonString
```
iSazonov: Please put `EmptyKeyInJsonString` in quotas. It is more readable.

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertFrom-Json.Tests.ps1
```diff
@@ -1,16 +1,33 @@
 Describe 'ConvertFrom-Json' -tags "CI" {
-    It 'can convert a single-line object' {
-        ('{"a" : "1"}' | ConvertFrom-Json).a | Should Be 1
+    
+    $testCasesWithAndWithoutAsHashtableSwitch = @(
```
iSazonov: Please put this in `BeforeAll` block.

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertFrom-Json.Tests.ps1
```diff
@@ -1,16 +1,33 @@
...
+        @{ AsHashtable = $true  }
+        @{ AsHashtable = $false }
+    )
+
+    It 'can convert a single-line object with AsHashtable switch set to <AsHashtable>' -TestCase $testCasesWithAndWithoutAsHashtableSwitch {
```
iSazonov: Please start test descriptions with capital.

iSazonov (commented):
Do we address the [comment](https://github.com/PowerShell/PowerShell/issues/3623#issuecomment-296537316) about case sensitivity?

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
```diff
@@ -1,6 +1,11 @@
 ﻿Describe 'Unit tests for JsonObject' -tags "CI" {
 
-    $validStrings = @(
+    BeforeAll {
+        $jsonWithEmptyKey = '{"": "Value"}'
+    }
+
+
```
iSazonov: Please remove extra line.

iSazonov:
@SteveL-MSFT @TravisEz13 @dantraMSFT Could you please review?

Also should we address the [comment](https://github.com/PowerShell/PowerShell/issues/3623#issuecomment-296537316) from @powercode 
>Apart from object size and performance, it may matter with case sensitivity. I have run into issues where I couldn't use the PowerShell cmdlets because the json contained 'Property' and 'property'.

iSazonov:
@bergmeister Thanks for your contribution! LGTM with above question about  case sensitivity.

SteveL-MSFT:
@iSazonov @powercode regarding case sensitivity, I think we should treat it the same as the empty key.  [Hashtable] allows two keys differing in case, so if we detect that, we should just tell the user to use `-AsHashTable`.

bergmeister:
@SteveL-MSFT @iSazonov @powercode I added a specific error message to point the user to the new `-AsHashTable` option.
As part of this I also found out that actually no error is thrown in the case of duplicate keys with the same casing. This is due to the behaviour of Newtonsoft.Json that just uses the last item if there are multiple items with the same key of the same casing. The PowerShell code currently caters for it should this behaviour change in a newer version of Newtonsoft.Json. This should definitely be discussed in another issue if you want to make a breaking change to throw when encountering case matching duplicate keys.

src/Microsoft.PowerShell.Commands.Utility/resources/WebCmdletStrings.resx
```diff
@@ -133,7 +133,10 @@
     <value>The provided JSON includes a property whose name is an empty string, this is only supported using the -AsHashTable switch.</value>
   </data>
   <data name="DuplicateKeysInJsonString" xml:space="preserve">
-    <value>Cannot convert the JSON string because a dictionary that was converted from the string contains the duplicated keys '{0}' and '{1}'.</value>
+    <value>Cannot convert the JSON string because a dictionary that was converted from the string contains the duplicated key '{0}'.</value>
+  </data>
+  <data name="KeysWithDifferentCasingInJsonString" xml:space="preserve">
+    <value>Cannot convert the JSON string because a PSObject does not support keys with different casing. Please use the -AsHashTable switch instead. The key that was attempted to be added to the exisiting key '{0}' was '{1}'.</value>
```
iSazonov: Oh, very long message!

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
```diff
@@ -1,46 +1,63 @@
...
+        It 'Throw InvalidOperationException when json contains key with different casing' {
+            $errorRecord = $null
+            [Microsoft.PowerShell.Commands.JsonObject]::ConvertFromJson($jsonContainingKeysWithDifferentCasing, [ref]$errorRecord)
+            $errorRecord.FullyQualifiedErrorId | Should Be 'KeysWithDifferentCasingInJsonString'
+        }
+
+        It 'Not throw when json contains key (same casing) when ReturnHashTable is true' {
+            $errorRecord = $null
+            $result = [Microsoft.PowerShell.Commands.JsonObject]::ConvertFromJson($jsonContainingKeysWithDifferentCasing, $true, [ref]$errorRecord)
+            $result | Should Not Be $null
```
iSazonov: I think it makes sense to check the hashtable values.

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
```diff
@@ -1,46 +1,63 @@
...
+        @{ name = "object"; str = "{a:1}"; ReturnHashTable = $true  }
+        @{ name = "null";   str = $null;   ReturnHashTable = $false }
+        @{ name = "empty";  str = "";      ReturnHashTable = $false }
+        @{ name = "spaces"; str = "  ";    ReturnHashTable = $false }
+        @{ name = "object"; str = "{a:1}"; ReturnHashTable = $false }
+    ) {
+        param ($str, $ReturnHashTable)
         $errRecord = $null
-        [Microsoft.PowerShell.Commands.JsonObject]::ConvertFromJson($str, [ref]$errRecord)
+        [Microsoft.PowerShell.Commands.JsonObject]::ConvertFromJson($str, $ReturnHashTable, [ref]$errRecord)
```
SteveL-MSFT: Please validate the returned object

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
```diff
@@ -1,46 +1,63 @@
...
+        It 'Throw InvalidOperationException when json contains empty key name' {
+            $errorRecord = $null
+            [Microsoft.PowerShell.Commands.JsonObject]::ConvertFromJson($jsonWithEmptyKey, [ref]$errorRecord)
+            $errorRecord.FullyQualifiedErrorId | Should Be 'EmptyKeyInJsonString'
+        }
+
+        It 'Not throw when json contains empty key name when ReturnHashTable is true' {
+            $errorRecord = $null
+            $result = [Microsoft.PowerShell.Commands.JsonObject]::ConvertFromJson($jsonWithEmptyKey, $true, [ref]$errorRecord)
+            $result | Should Not Be $null
```
SteveL-MSFT: Please validate the returned object

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
SteveL-MSFT: Please validate the returned object

iSazonov (commented):
@bergmeister Thanks for your contribution! I believe we are near to merge.
Feel free to open new Issue to discuss the discovered problem.

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/resources/WebCmdletStrings.resx
bergmeister: I know. I shortened it a bit but I think the last sentence still adds value.

bergmeister:
@iSazonov @SteveL-MSFT I added more validation to the positive test cases and shortened the error message a bit. I opened issue #5199 for the problem with Newtonsoft.Json for keys with matching cases, which is unrelated to this PR.

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
```diff
@@ -1,46 +1,68 @@
...
+            $result.Count | Should Be 1
+            $result.$('') | Should Be 'Value'
```
SteveL-MSFT: You should simplify this to just: $result."" | Should Be 'Value'

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
SteveL-MSFT: You should capture the output as $result and validate it is what you expect

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
SteveL-MSFT: This one is resolved.  Thanks.

test/powershell/Modules/Microsoft.PowerShell.Utility/JsonObject.Tests.ps1
SteveL-MSFT: Resolved

iSazonov (approved):

SteveL-MSFT (approved):

bergmeister:
@SteveL-MSFT I see that the documentation [here](https://github.com/PowerShell/PowerShell-Docs/blob/a1520688fb6b3e25ec51096d2611de9b54d08c2d/reference/6/Microsoft.PowerShell.Utility/ConvertFrom-Json.md) has not been updated yet for the new `-AsHashTable` switch (or is it already done in another branch?). Am I supposed to do that or shall I simply open an issue in the docs repo?

iSazonov:
@bergmeister Feel free to open an Issue in docs repo or push PR with the fix.

SteveL-MSFT:
@bergmeister preferably submit a PR to powershell-docs repo.  The `Documentation Needed` label here means you don't need to open an issue in the powershell-docs repo.  But my team won't be reviewing and adding docs until sometime after RC is out.  If you submit a PR for the doc update, please reference this PR so we can remove the `Documentation Needed` label.  Thanks!

PR #5051, opened 2017-10-07
Add location history for Set-Location to enable 'cd -' scenario (issue #2188)

### LINKED ISSUE #2188: Add support for 'cd -'
Sorely missing the ability of being able to go to previous directory without any previous action.

bergmeister:
## PR Summary

Closes #2188
This makes it possible to go back to the last directory using 'cd -', which is exactly the same syntax as on Linux shells. 
In contrast to the original proposal, it does not set the environment variable `OLDPWD` but instead uses a bounded stack to store the history in a similar way how Push/Pop-Location do.
A localised error message is given if there is no location history.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [x] Issue filed - Issue link: https://github.com/PowerShell/PowerShell-Docs/issues/2476
    - **OR**


PetSerAl:
Should it be some per-`Runspace` storage rather than per-`Process` environment variable? For example: PowerShell ISE allow to open multiple PowerShell tabs, which all share same process. Should it be one shared last location for all tabs or each tab should have its own last location?

bergmeister:
@PetSerAl Interesting thought and I did not know that the ISE tabs share the same process. The reason why it is an environment variable and why its name is `OLDPWD` is because I want to mimic the exact same behaviour as the native `cd -` function does on Unix systems. As far as I understand, PowerShell uses the native `cd` command on Linux/Mac but on Windows uses an alias for `Set-Location`. I am not sure if it is worth the effort of having a Windows specific implementation given the odd/rare use case that someone is cd-ing between different ISE tabs and wants to use `cd -` in a tab other than the current. Also note that the ISE is going to be replaced by VisualStudio Code at some point as far as I have heard. What do other people think?

PetSerAl:
As far as I understand, `cd` can not be native application even for Unix-like shells. That means if you start new process `cd` it change its own working directory and exit, then working directory of parent process (shell) will be unaffected.

Also PowerShell ISE is just an example of PowerShell host, which use more than one `Runspace` per `Process`. PowerShell Remote host can be configured to use shared process as well. `Enter-PSHostProcess` create addition `Runspace` in target process. There are modules, which achieve multithreading by using `Runspace`s. Simply: there are no rule "only one `Runspace` per `Process`".

bergmeister:
OK. I see what you mean. My assumption was that `cd -` is only being used interactively and in the same shell. If people were to start using `cd -` in multi-threaded modules then one could argue that this is bad practice and explicit paths should be specified. Would it be a possible workaround to use both environment and 'normal' ([automatic](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-5.1)) variables when setting the value and when getting the value checking only the 'normal' variable? The idea behind still setting the environment variable is consistency on Unix with mixed usage of `cd` and Set-Location.
Let's see what the PowerShell team thinks about this issue.

PetSerAl:
> My assumption was that cd - is only being used interactively and in the same shell.

Even when `cd -` will be used in only one `Runspace`, other `Runspace`s can do normal `cd something`, which affect subsequent `cd -` in that *only one `Runspace`*.

rkeithhill:
I agree with @PetSerAl.  See [PoshRSJobs](https://github.com/proxb/PoshRSJob) for an example of how users can run multiple runspaces in the same process.  Rather than make this an environment variable just make it a session (runspace) variable e.g. `PSOldLocation` or something like that.

bergmeister:
Ok, fair enough. But even executing a function that changes location would affect `cd -` as well. Even if we made it a runspace variable, I am wondering if we need to apply scoping to it to support the scenario `cd $location; Run-PowerShellFunctionThatChangesLocationInternally; cd -`?

iSazonov:
I agree with @PetSerAl too - we should take into account many runspaces.

Main question - Do we really need this in scripts?
```powershell
Set-Location -
cd -
```
looks very poorly read. Using Push-Location Pop-Location pair looks more readable and understandable.
I think this question is similar to the question of using aliases in an interactive session and banning its in scripts. I'd rather disable `cd -` in scripts immediately.

Next question - Should we keep a history and implement some step back by "cd-; cd-; cd -"?

bergmeister:
Ok. If you say runspace variable, do you mean we should declare it as an automatic variable or is there a subtle difference? How would one ban its usage in scripts? About the history: the idea is to keep it simple at the moment and another PR could improve it to use a history stack (that is the point about agile development: makes something minimal with value that works and then iteratively enhance it)

rkeithhill:
> Should we keep a history and implement some step back by "cd-; cd-; cd -"?

As an example of this, see the PSCX package.  It implements a cd stack that you can backwards navigate with `cd -` and forwards navigate with `cd +` e.g.:
```
[master ≡ +0 ~1 -0!] ~\GitHub\rkeithhill\PowerShellEditorServices
12:63ms> cd

     # Directory Stack:
   --- ----------------
     0 C:\Users\Keith\GitHub\rkeithhill\PowerShellEditorServices
     1 C:\Users\Keith\GitHub\dahlbyk\posh-git
     2 C:\Users\Keith\GitHub\rkeithhill\PowerShell
->   3 C:\Users\Keith\GitHub\rkeithhill\PowerShellEditorServices

[master ≡ +0 ~1 -0!] ~\GitHub\rkeithhill\PowerShellEditorServices
13:58ms> cd -
[master ↑8] ~\GitHub\rkeithhill\PowerShell
14:73ms> cd -0
[master ≡ +0 ~1 -0!] ~\GitHub\rkeithhill\PowerShellEditorServices
15:74ms> cd +
[rkeithhill/settings-prototype ≡] ~\GitHub\dahlbyk\posh-git
```
It also handles two cases that I particularly like:
```
# cd to the directory containing the file
16:47ms> cd $profile
~\Documents\WindowsPowerShell
```
and 
```
# cd to dir with spaces in path without requiring quoting - something CMD supports
17:5ms> cd c:\program files
C:\program files
```
> How would one ban its usage in scripts?

You don't.  That isn't something I'd worry about.  However there are other CD type packages out there.  Some allow you to fuzzy match when CD'ing to directories.

iSazonov:
I think we should easily implement the history based on "Stack" from the class.
`IsInteractive` can help to block the feature in scripts. We discussed `IsInteractive` in another context (shebang and profile loading). We definitely need it.
@lzybkr Could you please comment?

lzybkr:
I'd prefer a stack as well, though it needs to be bounded. You could start with this [code](https://github.com/lzybkr/PSReadLine/blob/master/PSReadLine/HistoryQueue.cs) from PSReadline.

I don't think it's necessary to prevent the use of `cd -` in a script - I don't expect it would be used much, and it might occasionally be useful.

src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
```diff
@@ -1054,6 +1071,29 @@ protected override void ProcessRecord()
             }
         } // ProcessRecord
 
+        private string ReplacePathWithLastLocationIfApplicable(string path)
+        {
+            Environment.SetEnvironmentVariable("DEBUG_PATH", Path);
```
lzybkr: I don't think we want this.

src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
```diff
@@ -915,7 +920,10 @@ protected override void ProcessRecord()
                             Path = SessionState.Internal.GetSingleProvider(Commands.FileSystemProvider.ProviderName).Home;
                         }
 
+                        Path = ReplacePathWithLastLocationIfApplicable(Path);
+                        var initialPath = SessionState.Path.CurrentLocation.Path;
                         result = SessionState.Path.SetLocation(Path, CmdletProviderContext);
+                        Environment.SetEnvironmentVariable(environmentVariableNameForPreviousLocation, initialPath);
```
lzybkr: We shouldn't use environment variables, instead we should expose a new variable (`PSOldPWD`?) that is an instance of a bounded stack.

We could consider a user configurable stack size, but we can leave that for later (not this PR).

I also wonder if this logic should be in `SessionState.Path.SetLocation` - that could be useful for alternate implementations of `Set-Location` that call the api.

src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
```diff
@@ -959,7 +967,10 @@ protected override void ProcessRecord()
                     try
                     {
                         // Change the default location stack
+                        StackName = ReplacePathWithLastLocationIfApplicable(StackName);
+                        var initialPath = SessionState.Path.CurrentLocation.Path;
                         result = SessionState.Path.SetDefaultLocationStack(StackName);
+                        Environment.SetEnvironmentVariable(environmentVariableNameForPreviousLocation, initialPath);
```
lzybkr: The previous comments apply here, including moving the logic to `SessionState.Path.SetDefaultLocationStack`.

src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
```diff
@@ -977,13 +988,16 @@ protected override void ProcessRecord()
                 case relationshipSet:
                     string relationshipPath = null;
                     try
-                    {
+                    {   
+                        relationshipPath = ReplacePathWithLastLocationIfApplicable(relationshipPath);
+                        var initialPath = SessionState.Path.CurrentLocation.Path;
```
lzybkr: You can revert the changes under `RELATIONSHIP_SUPPORTED` - the code is dead, and I've submitted a PR to delete it.

src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
```diff
@@ -996,7 +1010,10 @@ protected override void ProcessRecord()
 
                     try
                     {
-                        result = SessionState.Path.SetLocation (relationshipPath, CmdletProviderContext);
+                        relationshipPath = ReplacePathWithLastLocationIfApplicable(relationshipPath);
+                        var initialPath = SessionState.Path.CurrentLocation;
+                        result = SessionState.Path.SetLocation(relationshipPath, CmdletProviderContext);
```
lzybkr: The previous comments apply here, including moving the logic to `SessionState.Path.SetLocation`. In fact, I think if you did that, the change here wouldn't be necessary.

src/System.Management.Automation/resources/LocationStrings.resx
```diff
@@ -0,0 +1,123 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
lzybkr: I think it's reasonable to put the new string in `SessionStateStrings` instead of this new file.

lzybkr (changes requested):

src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
bergmeister: True. Well spotted. This was a development debugging leftover, sorry.

src/System.Management.Automation/resources/LocationStrings.resx
bergmeister: Ok.

src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
bergmeister: Ok, will do.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Location.Tests.ps1
```diff
@@ -55,4 +55,42 @@ Describe "Set-Location" -Tags "CI" {
...
+                $initialLocation = Get-Location
+                Set-Location ([System.IO.Path]::GetTempPath())
+                Set-Location -
+                (Get-Location).Path | Should Be ($initialLocation).Path
+            }
+
+            It 'Should go to last location when specifying minus as a path using alias' {
+                $initialLocation = Get-Location
+                Set-Location ([System.IO.Path]::GetTempPath())
+                cd -
```
iSazonov: Please remove the test - we shouldn't tests aliases.

src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
```diff
@@ -916,6 +916,10 @@ protected override void ProcessRecord()
                         }
 
                         result = SessionState.Path.SetLocation(Path, CmdletProviderContext);
+#if UNIX
+                        // Be consistent with most Unix shells that set the environment variable 'OLDPWD' when changing location
+                        Environment.SetEnvironmentVariable("OLDPWD", Path);
```
iSazonov: I don't see the point - we don't use it anywhere.

iSazonov (commented):
I don't like side effects - performance decreases, memory consumption increases.

src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
lzybkr: Agreed - I also don't see the point. Maybe if `ps1` files ran in a new process, then it might make sense, but they don't.

And if setting the environment variable was useful, we should explain why it is UNIX only.

src/System.Management.Automation/engine/SessionStateLocationAPIs.cs
```diff
@@ -231,6 +231,15 @@ internal PathInfo SetLocation(string path, CmdletProviderContext context)
...
             string providerId = null;
 
+
+            const string locationStackHistoryId = "SetLocationHistoryStack";
+            // Replace path with last working directory when '-' was passed.
+            if (originalPath.Equals("-"))            
+            {
+                return PopLocation(locationStackHistoryId);
+            }
+            PushCurrentLocation(locationStackHistoryId);
```
lzybkr: Nice - so simple. I do have a concern about an unbounded stack though. This stack will not be popped often, so I think there should be some limit.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Location.Tests.ps1
```diff
@@ -55,4 +55,42 @@ Describe "Set-Location" -Tags "CI" {
...
+                (Get-Location).Path | Should Be ($initialLocation).Path
+            }
+    
+            It 'Should go back to original location when specifying minus as a path twice' {
+                $tempPath = ([System.IO.Path]::GetTempPath())
+                Set-Location $tempLocation
+                $tempLocation = Get-Location
+                Set-Location -
+                Set-Location -
+                (Get-Location).Path | Should Be ($tempLocation).Path
```
lzybkr: This test doesn't make sense to me - there is one push, but two pops. How can this pass?

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Location.Tests.ps1
```diff
@@ -55,4 +55,42 @@ Describe "Set-Location" -Tags "CI" {
...
+            }
+
+            It 'Should fail if there is no last location' {
```
lzybkr: I'm also surprised about this test - it seems like you need a way to clear the stack before running `cd -` to make sure it reliably raises the exception.

src/Microsoft.PowerShell.Commands.Management/commands/management/Navigation.cs
bergmeister: Ok, I've removed it. I found out later that on Linux `cd` is also an alias for `Set-Location` (my assumption was that native Linux commands are not overridden by pwsh but this seems to be one of the exceptions)

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Location.Tests.ps1
bergmeister: ok

src/System.Management.Automation/engine/SessionStateLocationAPIs.cs
bergmeister: OK, I've implemented now a bounded stack based on a linked list.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Location.Tests.ps1
bergmeister: I have re-written all tests but the reason why it passed was because `Set-Location` also pushed to the stack (this was a bug that is fixed now)

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Location.Tests.ps1
bergmeister: I rewrote this test since it does not make sense any more (it was based on the first environment variable based implementation)

iSazonov (commented):
I wonder why not enhance  the existing API?

bergmeister:
@iSazonov If I understand correctly, changing the existing API to use a bounded stack would be a breaking change for pushd/popd. Enhancing the existing API should just be a matter of changing the type from `Stack` to the new `BoundedStack`. Or did you mean something else?

iSazonov:
@bergmeister We should _enhance_ the API not _change_. Actually you will use new constructor to set limits, existing code will keep current behavior without limits. This will allow us enhance our *-Location cmdlets and allow customers to use the new API possibilities.

src/System.Management.Automation/engine/SessionStateLocationAPIs.cs
```diff
@@ -231,10 +232,36 @@ internal PathInfo SetLocation(string path, CmdletProviderContext context)
             ProviderInfo provider = null;
             string providerId = null;
 
+            // Replace path with last working directory when '-' was passed.
+            if (_workingLocationHistoryStack == null)
+            {
+                const int locationHistoryLimit = 1000;
```
lzybkr: I think 1000 is at least 1, maybe 2 orders of magnitude too large.
Many people will never use this feature, and those that do will struggle to unwind the stack > [7](https://en.wikipedia.org/wiki/The_Magical_Number_Seven,_Plus_or_Minus_Two) anyway, instead using some other tool like [ZLocation](https://github.com/vors/ZLocation) or something like [this](https://github.com/lzybkr/PSReadLine/blob/d3e05af42dc550a31a44ac7f8b464e4b47094788/PSReadLine/SamplePSReadlineProfile.ps1#L504).

src/System.Management.Automation/engine/SessionStateLocationAPIs.cs
```diff
@@ -231,10 +232,36 @@ internal PathInfo SetLocation(string path, CmdletProviderContext context)
...
+                _workingLocationHistoryStack = new BoundedStack<PathInfo>(locationHistoryLimit);
```
lzybkr: Why here? It's always needed, so you can create in the constructor or where the field is declared.

lzybkr (dismissed):
I just had a couple of small issues with the initialization, otherwise looks pretty good.

Would it be useful to also support `+` to go forwards? Or does `pushd -; popd` work? If the later, It's maybe enough, but we should add tests.

rkeithhill:
+1 for `cd +` to move forward through the stack.

src/System.Management.Automation/engine/SessionStateLocationAPIs.cs
bergmeister: Ok. Shall we go for 50 then? This number also probably needs to go into the docs afterwards.

src/System.Management.Automation/engine/SessionStateLocationAPIs.cs
bergmeister: In terms of class design I would definitely agree but given that this partial class is spread across a lot of files, I thought it might be better to rather place it nearby its usage. I can of course move it into the constructor if that's your preference.

lzybkr: If you initialize on the field, it's in the same file and relatively close to the usage.

As for preference - when I see the `if` test, I ask myself - how can this be `null`? I think most people don't do this though, they just assume it's a possible condition and move on, but I dig deeper. It doesn't feels like initialization because it's not obvious - e.g. deferred initialization in a property getter. So I investigate - and I shouldn't need to do that.

Also - the `if` test felt unnecessary (and hence, not good for performance, not that it matters here) - another reason to change it.

src/System.Management.Automation/engine/SessionStateLocationAPIs.cs
lzybkr: I was thinking 10, maybe 15.

bergmeister:
I will make the requested changes next Monday since I am currently on travel. The `cd +` scenario via popd would not work at the moment because `popd` and `pushd` still use their default stacks, but should this feature not rather be another PR?

src/System.Management.Automation/engine/SessionStateLocationAPIs.cs
bergmeister: Ok. I'll do it on the property then, which sounds like reasonable compromise.

src/System.Management.Automation/engine/SessionStateLocationAPIs.cs
bergmeister: I can and will put in any number you'd like me to but I don't understand why we need to do penny pinching for a bunch of strings that are usually just around 100 char long. In terms of usage I can imagine some people including me wanting to go back further (I typically step out of directories one at a time in PS because I just repeatedly press 'up' and 'enter' after the first `cd ..`)

rkeithhill: I go over 10 frequently enough with the PSCX impl of this.  I was down with the `50` suggestion but at least make it 20.  :-)

bergmeister: Maybe Jeffrey should post a survey about this on Twitter (JK). I think it is an overkill but if we really need a low default then we could offer the user to override the default value by setting an environment variable or automatic variable?

src/System.Management.Automation/engine/SessionStateLocationAPIs.cs
lzybkr: Uh, sorry, that's not what I was looking for.

Deferred initialization is done when something is expensive and not always needed. If something is always needed (should never be null), then it should be initialized in the constructor.

src/System.Management.Automation/engine/SessionStateLocationAPIs.cs
lzybkr: An configuration setting is probably overkill.

You're storing a PathInfo, not a string. It's not immediately obvious to me that there are no NoteProperties also being stored - another small concern for the storage costs.

At any rate, I see assumptions far too often that something costs nothing and we find out far later that the assumption was wrong. Or it might have been right at the time, but something else crept in and made the other thing more expensive.

I try to be conservative because I tend to be the person finding those sorts of issues - the folks that make these assumptions tend to never see the consequences.

I'm fine with 20.

lzybkr:
I guess the real question about `cd +` is how much that would change this implementation. It could be another PR - though you'll change `BoundedStack` quite a bit, it won't be a stack anymore.

Personally, I think `pushd -` is sufficient though - did you try it?

bergmeister:
@lzybkr I made your requested changes to initialize in the constructor and changed the history limit to 20.

TravisEz13:
@lzybkr  Can you update your review?

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1520,4 +1520,44 @@ public static void SetTestHook(string property, object value)
             }
         }
     }
+
+    /// <summary>
+    /// An bounded stack based on a linked list.
+    /// </summary>
+    public class BoundedStack<T> : LinkedList<T>
```
lzybkr: @PowerShell/powershell-committee needs to approve a new public api like this.

bergmeister: Should we then rather think of making it internal (if possible)?

bergmeister: I made it internal then.

bergmeister:
NB: The test failure in Travis was `Update-Help for module 'PackageManagement'`. I guess this is a sporadic failure?

iSazonov:
I see the test failure sporadicly.

bergmeister:
@TravisEz13 Is there a reason why this PR keeps getting stale? I have always addressed the issues on my side and then everytime everything looks good, the PR becomes dead.

src/System.Management.Automation/engine/SessionStateLocationAPIs.cs
```diff
@@ -230,10 +231,31 @@ internal PathInfo SetLocation(string path, CmdletProviderContext context)
...
+                    var previousLocation =  _workingLocationHistoryStack.Pop();
+                    path = previousLocation.Path;
+                    pushNextLocation = false;
+                }
+                else
+                {
+                    throw new InvalidOperationException(SessionStateStrings.SetContentToLastLocationWhenHistoryIsEmpty);
+                }
+            }
+            if (pushNextLocation)
```
iSazonov: Please add new line.

iSazonov (commented):
Sorry we forgot the PR.
If you want to address some CodeFactor issues please push new PR.

rkeithhill:
I'd like to see this make it in for 6.1.  I was just setting up a fellow dev to do cross-platform development using WSL/Ubuntu.  We have a build.ps1 script that bootstraps and builds the product on Ubuntu.  After installing pwsh on his Ubuntu instance, he started it up, did a few things, tried `cd -` and was disappointed to see that it didn't work.  So he bounced back to bash and is relying on the fact that the build.ps1 script is shebanged/chmod+x.  Although he doesn't get parameter completion using it like that.  :-(

iSazonov (approved):

bergmeister:
I added a documentation issue here since this PR seems to get serious now: https://github.com/PowerShell/PowerShell-Docs/issues/2476

bergmeister:
@iSazonov There were test failures due it being the full test suite:
image
But the problem seems to be either sporadic or due to the test suite itself falling over:
image

bergmeister:
Those tests seemed to be sporadic and the new build contains 'only' 1 other sporadic failure

iSazonov:
@bergmeister Fis is ready #6943.

bergmeister:
@iSazonov Any updates on this so that it goes into the next preview for getting some testing exposure before 6.2 goes RTM after that?

iSazonov:
@bergmeister I don't understand. Please reword your question.

bergmeister:
@isazonov. Nevermind, I just saw that the next preview is already in the making.
Are there any interests to merge this before 6.1 gets released as RTM?

iSazonov:
@bergmeister We are waiting MSFT team sign.

src/System.Management.Automation/engine/SessionStateLocationAPIs.cs
```diff
@@ -230,10 +231,31 @@ internal PathInfo SetLocation(string path, CmdletProviderContext context)
             ProviderInfo provider = null;
             string providerId = null;
 
+            // Replace path with last working directory when '-' was passed.
+            bool pushNextLocation = true;
+            if (originalPath.Equals("-", StringComparison.OrdinalIgnoreCase))
+            {
+                if (_workingLocationHistoryStack.Count > 0)
```
rjmholt: I would structure this `if` as:
```csharp
if (_workingLocationHistoryStack.Count <= 0)
{
    throw new InvalidOperationException(...);
}

// Golden path behaviour
```

Just because there are some pretty deep `if`/`else` trees in the codebase and they get hard to reason about quickly.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1520,4 +1520,44 @@ public static void SetTestHook(string property, object value)
...
+
+        /// <summary>
+        /// Push item.
+        /// </summary>
+        /// <param name="item"></param>
+        internal void Push(T item)
+        {
+            this.AddFirst(item);
+
+            if(this.Count > _capacity)
```
rjmholt: Maybe include braces for the `if` and drop the `this` from property references. That seems to be the style convention of the file.

rjmholt (approved):
This looks good to me. Thanks @bergmeister!

iSazonov:
@daxian-dbw @adityapatwardhan Have you any thoughts before we merge?

src/System.Management.Automation/engine/SessionState.cs
```diff
@@ -65,6 +66,9 @@ internal SessionStateInternal(SessionStateInternal parent, bool linkToGlobal, Ex
             // is used for the pushd and popd commands
 
             _workingLocationStack = new Dictionary<String, Stack<PathInfo>>(StringComparer.OrdinalIgnoreCase);
+            // conservative choice to limit the memory impact in case of a regression
+            const int locationHistoryLimit = 20;
+            _workingLocationHistoryStack = new BoundedStack<PathInfo>(locationHistoryLimit);
```
BrucePay: This variable name is easily confused with _workingLocationStack and _workingLocationHistoryStack. I would suggest some more meaningful like _SetLocationHistory. (I don't think it needs it's type in it's name, and if we do decide to implement `cd +` it won't be a stack anymore.)

src/System.Management.Automation/engine/SessionState.cs
```diff
@@ -65,6 +66,9 @@ internal SessionStateInternal(SessionStateInternal parent, bool linkToGlobal, Ex
             // is used for the pushd and popd commands
 
             _workingLocationStack = new Dictionary<String, Stack<PathInfo>>(StringComparer.OrdinalIgnoreCase);
+            // conservative choice to limit the memory impact in case of a regression
```
BrucePay: I would put a space before the comment. I'd also add a comment saying that this is for the Set-Location history to distinguish it from the earlier "pushd and popd" comment.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1520,4 +1520,44 @@ public static void SetTestHook(string property, object value)
...
+        }
+
+        /// <summary>
+        /// Pop item.
+        /// </summary>
+        /// <returns></returns>
+        internal T Pop()
+        {
+            var item = this.First.Value;
+            this.RemoveFirst();
```
BrucePay: Should have a specific error message here instead of letting the base class throw it's own message if the linked list is empty. If we change the implementation to be a circular buffer, then the exception change too.

src/System.Management.Automation/engine/SessionStateLocationAPIs.cs
bergmeister: I agree, this is neater.

BrucePay (changes requested):
In general it looks good. Just a few things I think need to change.

src/System.Management.Automation/engine/Utils.cs
bergmeister: I agree, normally, it is me who comments on that on other people's code.
Alt Text

bergmeister: Yep, I agree. Fixed.

bergmeister:
@rjmholt @BrucePay I addressed all your comments

BrucePay (approved):

src/System.Management.Automation/engine/SessionState.cs
iSazonov: @bergmeister Please replace `_SetLocationHistory` -> `_setLocationHistory` - it is our name convention for static variables.

iSazonov:
@bergmeister Thanks for your contribution!

iSazonov:
@daxian-dbw Please clarify what is the "Revert..." commit?

bergmeister:
I saw this as well and cannot track it back, the PR is still in master (which is good), so I assume this commit either happened on a branch or the revert got reverted afterwards

rjmholt:
I think when you use GitHub's UI to revert a PR, it generates a new commit that reverts **all** the commits since that PR (which is not what we want -- reverting the single commit in the commandline is the better way). That commit is generated on another branch and not merged in, but still references all the other commits and shows up everywhere, which manages to confuse everyone.

II'm guessing @daxian-dbw initially tried to use the revert button, and has since overridden it in the commandline. So this merge hasn't been reverted, but a revert commit for it was generated and never used.

I know because I clicked the "revert" button once and was answering questions about it for a week.

daxian-dbw:
The confusing revert commit comes from https://github.com/PowerShell/PowerShell/pull/7182/commits.
Github automatically generates some empty commits when I revert #7151

PR #5052, opened 2017-10-07
Add Authentication Parameter to Web Cmdlets for Basic and OAuth

### LINKED ISSUE #4274: Invoke-WebRequest should directly support Basic authentication
We are working with more and more "Linux" based servers where we get instructions to run `curl` to say get back content for a `.npmrc` file from a product called Artifactory.  This product uses Basic auth to accept credentials.

The instructions they give are:
```bash
curl -u<USERNAME>:<PASSWORD> http://artifactory.it.keysight.com:8081/artifactory/api/npm/auth
```

To get this to work in PowerShell using Invoke-WebRequest, this is the best I could come up with:
```posh
$userpassB64 = [byte[]][char[]]"hillr:OpenSesame" | ConvertTo-Base64 -NoLineBreak
$res = Invoke-WebRequest http://artifactory.my.company.com:8081/artifactory/api/npm/auth `
       -Headers @{Authorization = "Basic $userpassB64"}
```
First, `ConvertTo-Base64` is a PSCX command (we really need B64 encode/decode in PS Core).  Second, Basic auth should be built-in like it is in CURL. 

This what I'd like to see from Invoke-WebRequest:
```posh
$res = Invoke-WebRequest http://artifactory.my.company.com:8081/artifactory/api/npm/auth `
       -UseBasicAuth -Credential (Get-Credential)
```
This `-UseBasicAuth` switch should also support the `-UseDefaultCredentials` parameter.  However the trick is whether or not to include the `Domain\` in the `username` field.  That option, when coupled with `-UseBasicAuth` "probably" should not include the domain.  In our scenario, Artifactory will fail to authenticate if the domain is specified.  Either way, the user can always fall back to explicitly providing the username with `-Credential (Get-Credential)`.

FYI, the curl docs on `-u`:
> -u, --user <user:password> 
Specify the user name and password to use for server authentication. Overrides -n, --netrc and --netrc-optional. 
If you simply specify the user name, curl will prompt for a password. 
The user name and passwords are split up on the first colon, which makes it impossible to use a colon in the user name with this option. The password can, still. 
When using Kerberos V5 with a Windows based server you should include the Windows domain name in the user name, in order for the server to successfully obtain a Kerberos Ticket. If you don't then the initial authentication handshake may fail. 
When using NTLM, the user name can be specified simply as the user name, without the domain, if there is a single domain and forest in your setup for example. 
To specify the domain name use either Down-Level Logon Name or UPN (User Principal Name) formats. For example, EXAMPLE\user and user@example.com respectively. 
If you use a Windows SSPI-enabled curl binary and perform Kerberos V5, Negotiate, NTLM or Digest authentication then you can tell curl to select the user name and password from your environment by specifying a single colon with this option: "-u :". 
If this option is used several times, the last one will be used.

markekraus:
Closes #4274

- Adds an `-Authentication` parameter to `Invoke-RestMethod` and `Invoke-WebRequest`
- Adds an `-Token` parameter to `Invoke-RestMethod` and `Invoke-WebRequest`
- Adds an `-AllowUnencryptedAuthentication` parameter to `Invoke-RestMethod` and `Invoke-WebRequest`
- Adds tests for various `-Authorization` uses
-  `-Authentication` Parameter has 3 options: `Basic`, `OAuth`, and `Bearer`
- `Basic` requires `-Credential` and provides RFC-7617 Basic Authorization credentials to the remote server
- `OAuth` and `Bearer` require the `-Token` which is a SecureString containing the bearer token to send to the remote server
- If any authentication is provided for any transport scheme other than HTTPS, the request will result in an error. A user may use the `-AllowUnencryptedAuthentication` switch to bypass this behavior and send their secrets unencrypted at their own risk.
- `-Authentication` does not work with `-UseDefaultCredentials` and will result in an error.

The existing behavior with `-Credential` is left untouched. When *not* supplying `-Authentication`, A user will *not* receive an error when using `-Credential` over unencrypted connections.

Code design choice is meant to accommodate more Authentication types in the future.

# Documentation Needed
The 3 new parameters will need to be added to the `Invoke-RestMethod` and `Invoke-WebRequest` documentation along with examples. Syntax will need to be updated.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
@@ -89,7 +89,7 @@ namespace Microsoft.PowerShell.Commands
         #region Authorization and Credentials
 
         /// <summary>
-        /// gets or sets the AllowUnencryptedAuthentication property
+        /// Gets or sets the AllowUnencryptedAuthentication property
         /// </summary>
         [Parameter]
         public virtual SwitchParameter AllowUnencryptedAuthentication { get; set; }
@@ -138,7 +138,7 @@ namespace Microsoft.PowerShell.Commands
         public virtual SwitchParameter SkipCertificateCheck { get; set; }
 
         /// <summary>
-        /// gets or sets the Token property. Token is required by Authentication OAuth and Bearer.
+        /// Gets or sets the Token property. Token is required by Authentication OAuth and Bearer.
         /// </summary>
         [Parameter]
         public virtual SecureString Token { get; set; }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
@@ -63,32 +63,32 @@ namespace Microsoft.PowerShell.Commands
         #region Authorization and Credentials
 
         /// <summary>
-        /// gets or sets the AllowUnencryptedAuthorization property
+        /// gets or sets the AllowUnencryptedAuthentication property
         /// </summary>
         [Parameter]
-        public virtual SwitchParameter AllowUnencryptedAuthorization { get; set; }
+        public virtual SwitchParameter AllowUnencryptedAuthentication { get; set; }
 
         /// <summary>
-        /// Gets or sets the Authorization property used to determin the Authorization method for the web session. 
-        /// Authorization does not work with UseDefaultCredentials.
-        /// Authorization over unencrypted sessions requires AllowUnencryptedAuthorization.
+        /// Gets or sets the Authentication property used to determin the Authentication method for the web session. 
+        /// Authentication does not work with UseDefaultCredentials.
+        /// Authentication over unencrypted sessions requires AllowUnencryptedAuthentication.
         /// Basic: Requires Credential
         /// OAuth/Bearer: Requires Token
         /// </summary>
         [Parameter]
         [ValidateSet("Basic","Bearer","OAuth")]
-        public virtual string Authorization 
+        public virtual string Authentication 
         { 
             get
             {
-                return _authorization;
+                return _authentication;
             } 
             set
             {
-                _authorization = value.ToLower();
+                _authentication = value.ToLower();
             }
         }
-        private string _authorization;
+        private string _authentication;
         
         /// <summary>
         /// gets or sets the Credential property
@@ -124,7 +124,7 @@ namespace Microsoft.PowerShell.Commands
         public virtual SwitchParameter SkipCertificateCheck { get; set; }
 
         /// <summary>
-        /// gets or sets the Token property. Token is required by Authorization OAuth and Bearer.
+        /// gets or sets the Token property. Token is required by Authentication OAuth and Bearer.
         /// </summary>
         [Parameter]
         public virtual SecureString Token { get; set; }
@@ -309,35 +309,35 @@ namespace Microsoft.PowerShell.Commands
                 ThrowTerminatingError(error);
             }
 
-            // Authorization
-            if (UseDefaultCredentials && (null != Authorization))
+            // Authentication
+            if (UseDefaultCredentials && (null != Authentication))
             {
-                ErrorRecord error = GetValidationError(WebCmdletStrings.AuthorizationConflict,
-                                                       "WebCmdletAuthorizationConflictException");
+                ErrorRecord error = GetValidationError(WebCmdletStrings.AuthenticationConflict,
+                                                       "WebCmdletAuthenticationConflictException");
                 ThrowTerminatingError(error);
             }
-            if ((null != Authorization) && (null != Token) && (null != Credential))
+            if ((null != Authentication) && (null != Token) && (null != Credential))
             {
-                ErrorRecord error = GetValidationError(WebCmdletStrings.AuthorizationTokenConflict,
-                                                       "WebCmdletAuthorizationTokenConflictException");
+                ErrorRecord error = GetValidationError(WebCmdletStrings.AuthenticationTokenConflict,
+                                                       "WebCmdletAuthenticationTokenConflictException");
                 ThrowTerminatingError(error);
             }
-            if ((Authorization == "basic") && (null == Credential))
+            if ((Authentication == "basic") && (null == Credential))
             {
-                ErrorRecord error = GetValidationError(WebCmdletStrings.AuthorizationCredentialNotSupplied,
-                                                       "WebCmdletAuthorizationCredentialNotSuppliedException");
+                ErrorRecord error = GetValidationError(WebCmdletStrings.AuthenticationCredentialNotSupplied,
+                                                       "WebCmdletAuthenticationCredentialNotSuppliedException");
                 ThrowTerminatingError(error);
             }
-            if ((Authorization == "oauth" || Authorization == "bearer") && (null == Token))
+            if ((Authentication == "oauth" || Authentication == "bearer") && (null == Token))
             {
-                ErrorRecord error = GetValidationError(WebCmdletStrings.AuthorizationTokenNotSupplied,
-                                                       "WebCmdletAuthorizationTokenNotSuppliedException");
+                ErrorRecord error = GetValidationError(WebCmdletStrings.AuthenticationTokenNotSupplied,
+                                                       "WebCmdletAuthenticationTokenNotSuppliedException");
                 ThrowTerminatingError(error);
             }
-            if (!AllowUnencryptedAuthorization && (null != Authorization) && (Uri.Scheme != "https"))
+            if (!AllowUnencryptedAuthentication && (null != Authentication) && (Uri.Scheme != "https"))
             {
-                ErrorRecord error = GetValidationError(WebCmdletStrings.AllowUnencryptedAuthorizationRequired,
-                                                       "WebCmdletAllowUnencryptedAuthorizationRequiredException");
+                ErrorRecord error = GetValidationError(WebCmdletStrings.AllowUnencryptedAuthenticationRequired,
+                                                       "WebCmdletAllowUnencryptedAuthenticationRequiredException");
                 ThrowTerminatingError(error);
             }
 
@@ -456,7 +456,7 @@ namespace Microsoft.PowerShell.Commands
             //
             // handle credentials
             //
-            if (null != Credential && null == Authorization)
+            if (null != Credential && null == Authentication)
             {
                 // get the relevant NetworkCredential
                 NetworkCredential netCred = Credential.GetNetworkCredential();
@@ -465,9 +465,9 @@ namespace Microsoft.PowerShell.Commands
                 // supplying a credential overrides the UseDefaultCredentials setting
                 WebSession.UseDefaultCredentials = false;
             }
-            else if ((null != Credential || null!= Token) && null != Authorization)
+            else if ((null != Credential || null!= Token) && null != Authentication)
             {
-                ProcessAuthorization();
+                ProcessAuthentication();
             }
             else if (UseDefaultCredentials)
             {
@@ -749,19 +749,19 @@ namespace Microsoft.PowerShell.Commands
             return String.Format("Bearer {0}", new NetworkCredential(String.Empty, Token).Password);
         }
 
-        private void ProcessAuthorization()
+        private void ProcessAuthentication()
         {
-            if(Authorization == "basic")
+            if(Authentication == "basic")
             {
                 WebSession.Headers["Authorization"] = GetBasicAuthorizationHeader();
             }
-            else if (Authorization == "bearer" || Authorization == "oauth")
+            else if (Authentication == "bearer" || Authentication == "oauth")
             {
                 WebSession.Headers["Authorization"] = GetBearerAuthorizationHeader();
             }
             else
             {
-                Diagnostics.Assert(false, String.Format("Unrecognized Authorization value: {0}", Authorization));
+                Diagnostics.Assert(false, String.Format("Unrecognized Authentication value: {0}", Authentication));
             }
         }
 
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
@@ -102,7 +102,7 @@ namespace Microsoft.PowerShell.Commands
         /// OAuth/Bearer: Requires Token
         /// </summary>
         [Parameter]
-        public virtual WebAuthenticationType Authentication { get; set; }
+        public virtual WebAuthenticationType Authentication { get; set; } = WebAuthenticationType.None;
         
         /// <summary>
         /// gets or sets the Credential property
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
@@ -472,10 +472,6 @@ namespace Microsoft.PowerShell.Commands
             //
             if (null != Credential && Authentication == WebAuthenticationType.None)
             {
-                if (Uri.Scheme != "https" && !AllowUnencryptedAuthentication)
-                {
-                    WriteWarning(WebCmdletStrings.UnencryptedCredentialWarning);
-                }
                 // get the relevant NetworkCredential
                 NetworkCredential netCred = Credential.GetNetworkCredential();
                 WebSession.Credentials = netCred;
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
@@ -4,7 +4,6 @@ Copyright (c) Microsoft Corporation.  All rights reserved.
 
 using System;
 using System.Collections.ObjectModel;
-using System.Diagnostics;
 using System.Management.Automation;
 using System.Net;
 using System.IO;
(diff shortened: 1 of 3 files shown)


iSazonov:
>The existing behavior with  -Credential  is left untouched. When not supplying  -Authorization , A user will not receive an error when using  -Credential  over unencrypted connections.

We could write non-terminating error. I think we should.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -62,6 +64,30 @@ public abstract partial class WebRequestPSCmdlet : PSCmdlet
...
+            get
+            {
+                return _authorization;
+            } 
+            set
+            {
+                _authorization = value.ToLower();
+            }
+        }
+        private string _authorization;
```
iSazonov: I believe we should use autoproperty and new public enum type for "Basic","Bearer","OAuth" - 'AuthorizationHeaderEnum' .

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -94,6 +120,12 @@ public abstract partial class WebRequestPSCmdlet : PSCmdlet
         [Parameter]
         public virtual SwitchParameter SkipCertificateCheck { get; set; }
 
+        /// <summary>
+        /// gets or sets the Token property
```
iSazonov: A more detailed description is welcomed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -62,6 +64,30 @@ public abstract partial class WebRequestPSCmdlet : PSCmdlet
         #region Authorization and Credentials
 
         /// <summary>
+        /// gets or sets the AllowUnencryptedAuthorization property
+        /// </summary>
+        [Parameter]
+        public virtual SwitchParameter AllowUnencryptedAuthorization { get; set; }
+
+        /// <summary>
+        /// Gets or sets the Authorization property used to determin the Authorization method for the web session.!--
```
iSazonov: A more detailed description is welcomed.
Please add about correlation with other parameters from PR description.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1286,6 +1286,117 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
...
+            }
+            $Response = Invoke-WebRequest @params
+            $result = $response.Content | ConvertFrom-Json
+
+            $result.Headers.Authorization | Should BeExactly "Basic dGVzdHVzZXI6dGVzdHBhc3N3b3Jk"
+        }
+
+        It "Verifies Invoke-WebRequest -Authorization <authorization>" -TestCases @(
+            @{ authorization = "bearer"}
+            @{ authorization = "OAuth"}
```
iSazonov: Formatting (spaces).
We use it twice - please move to BeforeAll. Below too.

iSazonov (commented):
@markekraus Great work!

markekraus:
> We could write non-terminating error or just terminating error w/o explicitly `-AllowUnencrypted`.

I considered that but I wasn't sure if that should be done in this PR or something separate. This PR is adding new features only without altering existing behavior. It's still not clear to me what is too much for one PR and what is not.

> Why 'Authorization' ? ~Right term is 'Authentication'.~ It's confusing. I think better 'AuthorizationHeaderType' and '-AllowUnencryptedAuthentication'.

I would take that up with IETF. They are the crazy ones that decided that Authorization is the correct term here regardless of what the dictionary begs to differ. 😆 

I think `-Authorization` is fine. I asked for feedback in the issue and there seemed to be an agreement on it being ok. The originally proposed parameter was `-Authentication`. To most people the words are interchangeable. 'AuthorizationHeaderType' or 'AuthorizationType'  is a bit verbose and redundant in my opinion.

'-AllowUnencryptedAuthentication' How about '-AllowUnencryptedSecrets'? Assuming there may be more uses outside of the 'Authorization'. I wouldn't want to have `-Authorization` and `-AllowUnencryptedAuthentication` mixing. so if one is `Authorization` so should be the other unless we decide on something completely different.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
markekraus: autoproperty? Please let me know what this is.

As for the new Public Enum. I was following suit with `-TransferEncoding` which doesn't have a public enum. I'm open to creating one, but I assume this requires committee approval?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -666,6 +734,34 @@ private bool IsCustomMethodSet()
...
+            {
+                WebSession.Headers["Authorization"] = GetBasicAuthorizationHeader();
+            }
+            else if (Authorization == "bearer" || Authorization == "oauth")
+            {
+                WebSession.Headers["Authorization"] = GetBearerAuthorizationHeader();
+            }
+            else
+            {
+                Debug.Assert(false, String.Format("Unrecognized Authorization value: {0}", Authorization));
```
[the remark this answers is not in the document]
markekraus: Will fix. Any idea why VS Code gives me this `'Diagnostics' is inaccessible due to its protection level [Microsoft.PowerShell.Commands.Utility]`?

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
markekraus: fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
markekraus: fixed.

markekraus:
Travis CI fail due to #5062

iSazonov:
I believe that the use of `Authorization` term on the client side is a crude distortion of its classical definition of AAA. @SteveL-MSFT pointed out this in the Issue discussion. When I started with the PR, I came here in a bit of confusion about what we wanted to do on the client side?! Turns out we need classical _authentication_ by means of setting `HTTP authorization header`. We have to be more precise in using long-known terms to prevent users from being misled.
So I think 'AuthorizationHeaderType' is good. IntelliSense help us with long parameters. But '-AuthHeaderType' is also good. Perhaps `-AllowUnencryptedAuth` too.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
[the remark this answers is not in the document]
iSazonov: Because the 'Diagnostics' is defined in another assembly SDK but we have `InternalsVisibleTo` in AssemblyInfo.cs - VS Code is still stuped here.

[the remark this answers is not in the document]
markekraus: Thanks for explaining. This is fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
iSazonov: I meant https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/auto-implemented-properties
Yes, public API should be approved. I believe it is right place to use it.

markekraus:
What is the consensus on the parameter names? leave them? change them?

rkeithhill:
I think adding "HeaderType" is more of an impl detail then the average user will care about.  And while the header is named Authorization, in articles and papers the overall process is referred to as `Authentication` as in Basic Authentication, Digest Authentication, etc.  Perhaps you just call the parameter `-Auth` and let the user interpret it as they see fit.  :-)

SteveL-MSFT:
To be consistent with other cmdlets, I think we should call it `-Authentication`.  People can use `-Auth` as it will resolve correctly.

markekraus:
Ok. I have switch Authorization with Authentication,

@SteveL-MSFT @iSazonov can I get some guidance on the public enum for the `-Authentication` parameter? Should we do this? What namespace do I use? I assume file location is in the `WebCmdlet` folder?

iSazonov:
I think an enum is good because we rid from many string constants and string comparitions, our code becomes more readable. I only worry about extensibility.
It seems `namespace Microsoft.PowerShell.Commands` is good.
` WebCmdletAuthenticationType` - we could use the type not only in web cmdlets.

markekraus:
@iSazonov 
> we could use the type not only in web cmdlets.

Then we would probably need to include logic that limits the subsets allowed by the authentication parameter. For example, if this enum were used in another cmdlet and it was extended to include `Shiboleth` for support in that cmdlet but the web cmdlets lack logic to handle that.

iSazonov:
It's overkill. I mean, this is a public type and users can use it. Although it may be better to mark it as obsolete and document that it is for internal use only.

markekraus:
@iSazonov I guess it is overkill. I looked a how other enums are being implemented for parameter options and have tried to follow their conventions. I chose the name `WebAuthenticationType` to make it a bit more generic and to give it some clarity vs other Auth enums used in various other cmdlets.

@SteveL-MSFT is a comittee review required for the new public enum or is your blessing good enough? 

@iSazonov and @SteveL-MSFT the other outstanding question in this PR is to add a warning to the legacy `-Credential` usage when being sent over non-HTTPS schemes. Should this be done in this PR or in a separate PR?

iSazonov:
You can add one commit for the warning :-)

SteveL-MSFT:
@markekraus I'll see if I can get a quick committee review on this

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -472,6 +472,10 @@ internal virtual void PrepareSession()
             //
             if (null != Credential && Authentication == WebAuthenticationType.None)
             {
+                if (Uri.Scheme != "https" && !AllowUnencryptedAuthentication)
+                {
+                    WriteWarning(WebCmdletStrings.UnencryptedCredentialWarning);
```
iSazonov: Maybe `WriteError()`? It seems we never use `WriteWarning()`.

SteveL-MSFT:
@PowerShell/powershell-committee brief discussion is that adding an enum used as a parameter type doesn't require committee approval, code review feedback is sufficient.  The current name and namespace seems appropriate to me.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
rkeithhill: Yup, this is what `ConvertTo-SecureString` does if -AsPlainText is used without -Force:
```
2:8ms> $ss = ConvertTo-SecureString "mypassword" -AsPlainText
ConvertTo-SecureString : The system cannot protect plain text input. To suppress this warning and convert the plain
text to a SecureString, reissue the command specifying the Force parameter. For more information ,type: get-help
ConvertTo-SecureString.
At line:1 char:7
+ $ss = ConvertTo-SecureString "mypassword" -AsPlainText
+       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [ConvertTo-SecureString], ArgumentException
    + FullyQualifiedErrorId : ImportSecureString_ForceRequired,Microsoft.PowerShell.Commands.ConvertToSecureStringComm
   and
```

markekraus: @iSazonov I think an error would be a breaking change here. Recall this piece of code is not part of the new `-Authentication` parameter, but part of the legacy `-Credential` usage. If we want this an error here, I will revert and that can be done in a separate PR. If we are going to break it, the better way, IMO, is to remove the legacy usage and have it be completely replaced with the new usage and removing the legacy `-Credential` usage has some cleanup associated with it.

iSazonov: It seems the PR is near to merge. I think we shouldn't block it - please revert and open tracking Issue to discuss "warning", "removing" and "force".

rkeithhill: Maybe error error with the new -Auth Basic (no -AllowUnen) scenario but not the existing -Credential (no -AllowUnen)?

markekraus: @rkeithhill  that is already part of this PR 😄 the new method is terminating errors when you tery to use non-HTTPS and can only be suppressed with `-AllowUnencryptedAuthentication`. 

This piece of code that this comment thread is on was just to add some more whining to the legacy `-Credential` usage which has always casually let you send credentials over HTTP with out so much as a peep.

@iSazonov #5112 for the this. I have reverted that commit (with the exception of the minor spelling fix I the other tests in this PR)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -62,6 +89,22 @@ public abstract partial class WebRequestPSCmdlet : PSCmdlet
...
+
+        /// <summary>
+        /// Gets or sets the Authentication property used to determin the Authentication method for the web session. 
+        /// Authentication does not work with UseDefaultCredentials.
+        /// Authentication over unencrypted sessions requires AllowUnencryptedAuthentication.
+        /// Basic: Requires Credential
+        /// OAuth/Bearer: Requires Token
+        /// </summary>
+        [Parameter]
+        public virtual WebAuthenticationType Authentication { get; set; }
```
iSazonov: Should we assign default `WebAuthenticationType.None`?

markekraus: I don't think ther is a need, The enum is never null so its creation here automatically sets it to `None` since it is the first enum value... at least.. if i understand everything I read correctly. If it wasn't automatically None, all of the many tests would be failing. Unless setting it here provides some kind of emit somewhere. I didin't see any other enum parameters setting a default value.

iSazonov: If internal logic is based on `None` we should initialize explicitly to `None` (Can we sort the enum or add new element on first position?). Another way is to being based on Authentication.IsPresent.

markekraus: fixed.

TravisEz13 (approved):

TravisEz13:
@markekraus Thanks for your contribution.  I'll allow 24 hours for the code owners to review.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -62,6 +89,22 @@ public abstract partial class WebRequestPSCmdlet : PSCmdlet
         #region Authorization and Credentials
 
         /// <summary>
+        /// gets or sets the AllowUnencryptedAuthentication property
```
iSazonov: Could you please replace "gets" -> "Gets"?
Although the old comments start with a small letter, the new ones are already capital. In this case, it's better to follow our rules and use the capital letter.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -94,6 +137,12 @@ public abstract partial class WebRequestPSCmdlet : PSCmdlet
         [Parameter]
         public virtual SwitchParameter SkipCertificateCheck { get; set; }
 
+        /// <summary>
+        /// gets or sets the Token property. Token is required by Authentication OAuth and Bearer.
```
iSazonov: The same "get" -> "Gets".

src/Microsoft.PowerShell.Commands.Utility/resources/WebCmdletStrings.resx
```diff
@@ -120,6 +120,21 @@
   <data name="AccessDenied" xml:space="preserve">
     <value>Access to the path '{0}' is denied.</value>
   </data>
+  <data name="AllowUnencryptedAuthenticationRequired" xml:space="preserve">
+    <value>The cmdlet cannot protect plain text secrets sent over unencrypted connections. To supress this warning and send plain text secrets over unencrypted networks, reissue the command specifying the AllowUnencryptedAuthentication parameter.</value>
```
iSazonov: I think it is more readable to put parameter names in single quotes `AllowUnencryptedAuthentication` ...
We use the format in other Resx files.
Please edit messages below too.

markekraus: There is a mix, from what I have seen, of quoted, unquoted and Hyphen prepended in the project. Since there is no clear set standard, I went with the style used for the Web Cmdlets. Please look at the errors in this Resx. I would rather have a separate PR to address the errors styles as whole for these commands than have a few that do not match in this PR.

iSazonov: I agree. Could you please open a tracking issue (or PR)?

iSazonov (approved):

src/Microsoft.PowerShell.Commands.Utility/resources/WebCmdletStrings.resx
markekraus: I could.. but I think another Issue on standardizing how parameters are displayed in errors within the project is probably warranted first. I'd hate to go through the trouble of changing these error and then have the current way be later selected as the standard (I would hope it's not, because I agree it's not very readable, but its possible). thoughts?

markekraus: #5140

iSazonov:
😄 I've been dragged the PR.

PR #5055, opened 2017-10-07
Add better error message for empty and null -UFormat arg

### LINKED ISSUE #5047: Get-Date -UFormat '' throws an unhelpful exception for an empty string argument
An empty string argument in `Get-Date -UFormat ''` throws `IndexOutOfRangeException`, it would be more helpful if the error said what was wrong and directed attention to the UFormat argument. "Missing or Invalid UFormat String" or words to that effect.

Steps to reproduce
------------------

```powershell
PS /test> get-date -uformat "$accidentallyEmptyFormatStringVariable"
get-date : Index was outside the bounds of the array.
At line:1 char:1
+ get-date -uformat ''
+ ~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Get-Date], IndexOutOfRangeException
    + FullyQualifiedErrorId : System.IndexOutOfRangeException,Microsoft.PowerShell.Commands.GetDateCommand

```

Expected behavior
-----------------

```none
A message directing attention to the -UFormat string parameter being empty.
```

Actual behavior
---------------

```none
A message about some internal problem.
```

Environment data
----------------

```powershell
> $PSVersionTable
PSVersion                      6.0.0-beta.8
GitCommitId                    v6.0.0-beta.8
OS                             Linux 2.6.32-042stab120.6 #1 SMP Thu Oct 27 16:59:03 MSK 2016

```

djweber:
Previously, when passing an empty format string to -UFormat, a somewhat unhelpful
error message would appear. This commit fixes #5047 by adding a more descriptive error to better
guide the user

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
@@ -199,6 +199,7 @@ namespace Microsoft.PowerShell.Commands
         /// Unix format string
         /// </summary>
         [Parameter(ParameterSetName = "UFormat")]
+        [ValidateNotNullOrEmpty]
         public string UFormat { get; set; }
 
 
@@ -331,7 +332,6 @@ namespace Microsoft.PowerShell.Commands
             DateTime epoch = DateTime.Parse("January 1, 1970", System.Globalization.CultureInfo.InvariantCulture);
             int offset = 0;
             StringBuilder sb = new StringBuilder();
-
 
             // folks may include the "+" as part of the format string
             if (UFormat[0] == '+')
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Date.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Date.Tests.ps1
@@ -37,6 +37,14 @@ Describe "Get-Date DRT Unit Tests" -Tags "CI" {
         Get-date -Date 1/1/0030 -uformat %S%T%u%U%V%w%W%x%X%y%Y%% | Should be "0000:00:002012001/01/3000:00:00300030%"
     }
 
+    It "Passing '<name>' to -uformat produces a descriptive error" -TestCases @(
+        @{ name = "`$null"      ; value = $null }
+        @{ name = "empty string"; value = "" }
+    ) {
+        param($value)
+        { Get-date -Date 1/1/1970 -uformat $value -ErrorAction Stop } | ShouldBeErrorId "ParameterArgumentValidationError,Microsoft.PowerShell.Commands.GetDateCommand"
+    }
+
     It "Get-date works with pipeline input" {
         $x = new-object System.Management.Automation.PSObject
         $x | add-member NoteProperty Date ([DateTime]::Now)


src/Microsoft.PowerShell.Commands.Utility/resources/GetDateStrings.resx
```diff
@@ -0,0 +1,123 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="InvalidUFormatString" xml:space="preserve">
+    <value>Date format string passed to -UFormat cannot be empty</value>
```
SteveL-MSFT: Should have sentence end with period.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Date.Tests.ps1
```diff
@@ -37,6 +37,10 @@ Describe "Get-Date DRT Unit Tests" -Tags "CI" {
         Get-date -Date 1/1/0030 -uformat %S%T%u%U%V%w%W%x%X%y%Y%% | Should be "0000:00:002012001/01/3000:00:00300030%"
     }
 
+    It "using -uformat with an empty format string produces a descriptive error" {
+        { Get-date -Date 1/1/1970 -uformat '' -ErrorAction Stop } | ShouldBeErrorId "InvalidUFormatString,Microsoft.PowerShell.Commands.GetDateCommand"
```
SteveL-MSFT: Perhaps have two `-TestCases` where one is an empty string and the other is just whitespace

SteveL-MSFT (changes requested):
Thanks for the contribution!  Just a few minor comments

DdWr:
> Thanks for the contribution! Just a few minor comments
> Show outdated src/Microsoft.PowerShell.Commands.Utility/resources/GetDateStrings.resx
> Show outdated test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Date.Tests.ps1

Thanks for the feedback! I made the changes you requested.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
```diff
@@ -279,6 +279,18 @@ protected override void ProcessRecord()
 
             if (UFormat != null)
             {
+                if (UFormat.Trim().Length == 0)
```
iSazonov: What about `null`? And I wonder why not use standard `[ValidateNotNullOrEmpty()]`?

DdWr: It seems that `[ValidateNotNullOrEmpty()]` does not take whitespace-only strings into account, so it looks like that has to be checked manually. As for `null`, it looks like the way the current code is written, if `$null` is passed to `-uformat`, it will just fall through the if-else completely and just print the date with a default format. Do we want to keep this behavior, or do we want to use [ValidateNotNull()]?

iSazonov: " " - is allowed format string.
Disabling Null is breaking change - @SteveL-MSFT What is your thoughts?

DdWr:
> " " - is allowed format string.
> Disabling Null is breaking change - @SteveL-MSFT What is your thoughts?

Just to clarify, I haven't changed the behavior of `$null` being passed in. I did not know that `" "` was a valid format string -- I assumed that this was also considered empty.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
DdWr: Thanks for the feedback. Just to clarify, I haven't changed the behavior of $null being passed in. I did not know that " " was a valid format string -- I assumed that this was also considered empty.

SteveL-MSFT: Since `$null` is the same as if `-UFormat` wasn't provided, we probably don't want to break that behavior as someone may be relying on it.  `-UFormat` enables custom formatting, so any valid string should work.  I suppose it's not different if they provided `" "` or `"apple"` which should assume they know what they are doing.  So it seems that in this case, the only fix is on [this line](https://github.com/DdWr/PowerShell/blob/d7fc318c63e9192148fb100070b13deb9ea0266d/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs#L349) if UFormat.Length == 0, just return.

iSazonov: We can use `ValidateLength`

DdWr: @SteveL-MSFT @iSazonov Thank you for the responses. Based off this discussion, can we clarify the desired behavior? Based off of what has been said, this is the behavior it seems that we want here:

**Current**
* If `$null` is passed in, fall through and just print the date
* If `'  '`, `apple`, or any unconventional date format of length > 0 is passed in, keep the current behavior and show the formatted result.

**Proposed addition**
* Add a length check to [this line](https://github.com/DdWr/PowerShell/blob/d7fc318c63e9192148fb100070b13deb9ea0266d/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs#L349), which will cause it to match the above behavior and format the date to an empty string.
* Use `ValidateLength`, which will throw an error if an empty string is passed in.

iSazonov: I believe ` ValidateNotNullOrEmpty ` would best solution if @SteveL-MSFT'd agreed. This breaking change seems incredible.

DdWr: Updated PR with proposed changes.

iSazonov:
CI MacOs temporary failed - I restart it.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs
iSazonov: @SteveL-MSFT I think the uformat is static constant in most scenarios. If someone dynamically forms it, it's not likely that it will leave it uninitialized (null). So I believe this is a breaking change  only formally and we can accept it.

SteveL-MSFT: I agree that most likely if $null gets passed, it's probably a problem in someone's script.  Let me get a quick ack from @PowerShell/powershell-committee before we take this breaking change.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and agree that using the ValidateNotNullOrEmpty attribute is the right fix and having it error with $null is the right behavior although breaking it seems low risk of impact.

iSazonov:
@DdWr Please correct your commit - it contains alien code.

DdWr:
@iSazonov Fixed

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Date.Tests.ps1
```diff
@@ -37,6 +37,10 @@ Describe "Get-Date DRT Unit Tests" -Tags "CI" {
         Get-date -Date 1/1/0030 -uformat %S%T%u%U%V%w%W%x%X%y%Y%% | Should be "0000:00:002012001/01/3000:00:00300030%"
     }
 
+    It "using -uformat with an empty format string produces a descriptive error" {
+        { Get-date -Date 1/1/1970 -uformat '' -ErrorAction Stop } | ShouldBeErrorId "ParameterArgumentValidationError,Microsoft.PowerShell.Commands.GetDateCommand"
```
iSazonov: Please add test for $null.
Use `TestCases`.

DdWr: @iSazonov PR updated

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Date.Tests.ps1
```diff
@@ -37,6 +37,14 @@ Describe "Get-Date DRT Unit Tests" -Tags "CI" {
         Get-date -Date 1/1/0030 -uformat %S%T%u%U%V%w%W%x%X%y%Y%% | Should be "0000:00:002012001/01/3000:00:00300030%"
     }
 
+    It "Passing <name> to -uformat produces a descriptive error" -TestCases @(
```
SteveL-MSFT: I would suggest putting single quotes around it to make it more readable in the test output:

```none
It "Passing '<name>' to ...
```

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Date.Tests.ps1
```diff
@@ -37,6 +37,14 @@ Describe "Get-Date DRT Unit Tests" -Tags "CI" {
...
+        @{ name = "null"; value = $null }
```
SteveL-MSFT: Please change `"null"` to `"$null"` (use backtick to escape the dollar sign).  Also add some whitespace to align `value` with test case below to make it easier to read:

```powershell
@{ name = "`$null"      ; value = $null }
@{ name = "empty string"; value = "" }
```

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Date.Tests.ps1
```diff
@@ -37,6 +37,14 @@ Describe "Get-Date DRT Unit Tests" -Tags "CI" {
...
+        @{ name = "an empty string"; value = ""}
```
SteveL-MSFT: Suggesting removing `an` and adding whitespace after the empty string (see above)

SteveL-MSFT (changes requested):
Thanks for the contribution.  Just a few minor comments.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Date.Tests.ps1
DdWr: @SteveL-MSFT Done

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Date.Tests.ps1
DdWr: @SteveL-MSFT Done. Thank you for the feedback -- it will be helpful moving forward.

SteveL-MSFT (approved):
LGTM!

iSazonov (approved):

iSazonov:
@DdWr Please don't rebase your commits and keep a _history_. Sometimes we have to rebase to resolve merge conflitcs. Ask maintainers If you are not sure.

PR #5058, opened 2017-10-08
Replace HttpBin.org/response-headers Tests with WebListener

markekraus:
* Adds `/ResponseHeaders/` Test to WebListener
* Adds `-Query` parameter to `Get-WebListenerUrl` which accepts a dictionary where keys are wuery string fields and values are query string values
* Replaces http://httpbin.org/response-headers tests with WebListener

Reference #2504 

(note: the tests as they were intended before were actually not accurate. http://httpbin.org/response-headers returns an `Content-Type: application/json` in addition to `Content-Type: ` from the supplied header. The implementation in this PR does allow for a blank/missing `Content-Type` to be returned. It is possible this may reveal errors that were not visible before.)

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -1190,3 +1190,8 @@ tests.zip
 v5.0
 v6.0.
 #endregion
+
+#region test/tools/WebListener/README.md Overrides
+ - test/tools/WebListener/README.md
+ResponseHeaders
+#endregion
--- a/test/tools/WebListener/Controllers/ResponseHeadersController.cs
+++ b/test/tools/WebListener/Controllers/ResponseHeadersController.cs
@@ -40,6 +40,7 @@ namespace mvc.Controllers
             }
             return JsonConvert.SerializeObject(headers);
         }
+        
         public IActionResult Error()
         {
             return View(new ErrorViewModel { RequestId = Activity.Current?.Id ?? HttpContext.TraceIdentifier });
--- a/test/tools/WebListener/Controllers/ResponseHeadersController.cs
+++ b/test/tools/WebListener/Controllers/ResponseHeadersController.cs
@@ -1,0 +1,48 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Extensions;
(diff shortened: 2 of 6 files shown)


.spelling
```diff
@@ -1190,3 +1190,8 @@ tests.zip
 v5.0
 v6.0.
 #endregion
+
+#region test/tools/WebListener/README.md Overrides
+ - test/tools/WebListener/README.md
+ResponseHeaders
+#endregion
```
iSazonov: Are they really all used?

test/tools/WebListener/Controllers/ResponseHeadersController.cs
```diff
@@ -0,0 +1,48 @@
...
+                    }, HttpContext);
+                }
+                else
+                {
+                    Response.Headers.TryAdd(key, Request.Query[key]);
+                }
+            }
+            return JsonConvert.SerializeObject(headers);
+        }
+        public IActionResult Error()
```
iSazonov: Maybe add newline.

iSazonov (approved):
LGTM with two minor comments.

.spelling
markekraus: I'm not sure what you are asking. But, I added a file level override because I had been putting all of the overrides for this in the Global dictionary.  That was probably a poor decision, so I'm adding the new section and will do a separate PR after this is merged to move the ones from my previous PRs from global to the file override.

test/tools/WebListener/Controllers/ResponseHeadersController.cs
markekraus: fixed.

.spelling
iSazonov: Thanks. Clear.

Closed.

markekraus:
Travis CI fail due to #5062

iSazonov:
@markekraus MSFT maintainers address such issues very fast. Be patient :-)

markekraus:
@iSazonov they do :smile:. But the community can also address them if an issue is filed. Besides, I prefer to have an issue that explains why my PRs are failing externally. Makes it easier to track when the external issue has been resolved to move forward. @SteveL-MSFT would likele create an issue for this when he came online anyway. I'm no no hurry, just trying to be helpful and thorough.

SteveL-MSFT:
@markekraus looking at the Mac issue, restarted your build for now

markekraus:
@TravisEz13 Is there something outstanding on this PR I need to do?

TravisEz13 (approved):

PR #5061, opened 2017-10-09
Update testing-guidelines.md to remind users about submodules

djweber:
The bit about submodules in the README is easy to miss or forget if you're reading deep into the docs. It might be helpful to have a reminder here if someone is having trouble with the Pester module not being found.

### DIFF
--- a/docs/testing-guidelines/testing-guidelines.md
+++ b/docs/testing-guidelines/testing-guidelines.md
@@ -100,6 +100,12 @@ Two helper functions are part of the build.psm1 module to help with that:
 
 Our CI system runs these as well; there should be no difference between running these on your dev system, versus in CI.
 
+Make sure that the git submodules have been loaded into your project before running `Start-PSPester`, or it will fail to run. If you did not clone the project with the `--recursive` flag, you can load the submodules by running: 
+
+```
+git submodule update --init
+```
+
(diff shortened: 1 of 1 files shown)


iSazonov:
We already have this in 4 files https://github.com/PowerShell/PowerShell/search?l=Markdown&q=git+submodule+&type=&utf8=%E2%9C%93

markekraus:
@iSazonov I think this one trips up enough people that it should also be included here. I have answered several DM's about testing fails due to this exact issue. It's not obvious, so we might as well be repetitive about it.

iSazonov:
I think "dry" principle is good for docs too. We plan remove submodules and already remove some.

iSazonov:
Maybe it is better to add a warning in `Start-PSPester`.

markekraus:
@iSazonov Either a warning in Start-PSPester (if it's possible to check for the sub modules there, then even better) or a note in the documentation. Just so long as submodules are more closely linked to testing due to the dependencies.

DdWr:
@iSazonov @markekraus If we choose to add a warning to `Start-PSPester`, I can help with that.

TravisEz13:
restarted macOS ci

DdWr:
Sent up a PR for the build script warning: #5069

PR #5066, opened 2017-10-09
Remove dead code

lzybkr:
Code guarded by the following defines is ancient and has never been
used, so removing.

* RELATIONSHIP_SUPPORTED
* SUPPORTS_IMULTIVALUEPROPERTYCMDLETPROVIDER


adityapatwardhan (approved):
LGTM

SteveL-MSFT:
@lzybkr are these used in Windows PowerShell (and thus a breaking change)?

anmenaga:
@SteveL-MSFT I've ran a search on WindowsPS code; I don't see these macros defined anywhere.

lzybkr:
Right, this code has never been used, probably hasn't been compiled in over 10 years.
I pushed a couple more changes (sorry, force pushed) to remove a little bit more unused dead code (well, the file I removed is probably being compiled, but not used apparently, removed because I saw `#if IA64` and was going to kill that, realized I'd never seen that code before, and did a build without it.)

daxian-dbw:
A side question, the `PSMI` code at https://github.com/PowerShell/PowerShell/tree/master/src/System.Management.Automation/engine/PSMI is the prototype code @Indhukrishna worked on about 4 years ago to do powershell remoting over WMI remoting. Do we still need that code?

@Indhukrishna can you please weigh in?

SteveL-MSFT (approved):

daxian-dbw:
I synced with @Indhukrishna offline and submitted PR #5075 to remove `PSMI` code.

PR #5067, opened 2017-10-09
Add warning to Start-PSPester if Pester module not found

djweber:
Currently, if a user does not clone with the --recursive flag or run
git submodule update --init, Start-PSPester will fail to run due to
the Pester module missing. While the errorhints at the Pester module not
being found, there is no suggested way of fixing the issue presented to
the user.

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -848,6 +848,10 @@ function Start-PSPester {
         [switch]$IncludeFailingTest
     )
 
+    if (-not (Test-Path $Pester)) {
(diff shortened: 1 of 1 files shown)


build.psm1
```diff
@@ -848,6 +848,10 @@ function Start-PSPester {
         [switch]$IncludeFailingTest
     )
 
+    if (-not (Test-Path $Pester)) {
```
TravisEz13: you would need to run `git submodule status` and read the status

DdWr:
Ancestor of #5069. Had an issue with a deleted submodule, so I closed this and made a new PR.

PR #5069, opened 2017-10-09
Terminate Start-PSPester if Pester module not found

djweber:
Currently, if a user does not clone with the `--recursive` flag or run `git submodule update --init`, `Start-PSPester` will fail to run due to the Pester module missing. While the error hints at the Pester module not being found, there is no suggested way of fixing the issue presented to the user.  This change causes a warning to appear at the beginning of the execution of `Start-PSPester` if the Pester module cannot be found.

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -848,6 +848,16 @@ function Start-PSPester {
         [switch]$IncludeFailingTest
     )
 
+    if (-not (Get-Module -ListAvailable -Name $Pester -ErrorAction SilentlyContinue)) 
+    {
+        Write-Warning @"
+Pester module not found.
+Make sure that the proper git submodules are installed by running:
+    git submodule update --init
+"@
+        return;
+    }
+
     if ($IncludeFailingTest.IsPresent)
     {
         $Path += "$PSScriptRoot/tools/failingTests"


build.psm1
```diff
@@ -848,6 +848,10 @@ function Start-PSPester {
         [switch]$IncludeFailingTest
     )
 
+    if (-not (Test-Path $Pester)) {
+        Write-Warning "Pester module not found. Make sure that the proper git submodules are installed by running:`r`n`r`ngit submodule update --init`r`n "
```
adityapatwardhan: Should this be an error instead? We would not be able to run tests without Pester.

DdWr: @adityapatwardhan I could change this to use `Write-Error, but now I'm wondering if we should use `throw` instead -- after all, there is no point in continuing at all if Pester can't be found.

adityapatwardhan: Actually if we follow the pattern from: https://github.com/DdWr/PowerShell/blob/3fdd9f1b639df805f895c56cf574f2039ec15a76/build.psm1#L415

We should do Write-Warning and then return.

DdWr: @adityapatwardhan Changes made.

build.psm1
```diff
@@ -848,6 +848,17 @@ function Start-PSPester {
...
     )
 
+    if (-not (Test-Path $Pester)) {
+        Write-Warning @"
+Pester module not found. Make sure that the proper git submodules are installed by running:
+
+git submodule update --init 
+
+
+"@
```
daxian-dbw: A minor comment about the message -- how about this:
```
Write-Warning @"
Pester module not found.
Make sure that the proper git submodules are installed by running:
    git submodule update --init
"@
```
It looks like this:
image

build.psm1
```diff
@@ -848,6 +848,15 @@ function Start-PSPester {
         [switch]$IncludeFailingTest
     )
 
+    if (-not (Get-Module -ListAvailable -Name $Pester -ErrorAction SilentlyContinue)) {
```
DdWr: @daxian-dbw I made the module check a little stricter here than simply checking for the path.

build.psm1
```diff
@@ -848,6 +848,16 @@ function Start-PSPester {
         [switch]$IncludeFailingTest
     )
 
+    if (-not (Get-Module -ListAvailable -Name $Pester -ErrorAction SilentlyContinue))
```
adityapatwardhan: I actually prefer the `Test-Path` instead. With `-ListAvailable` the Pester version that comes installed with Windows 10 will be found. There are difference in those versions which might cause false errors in running tests.

adityapatwardhan (changes requested):

build.psm1
DdWr: @adityapatwardhan Since the full path to the project's Pester module is being passed in here, I believe this would not be an issue:

screen shot 2017-10-10 at 2 22 11 pm

However, if we only passed in `Pester`, I could see that being a cause for concern:

screen shot 2017-10-10 at 2 24 13 pm

adityapatwardhan: Makes sense. Misread it being Pester instead of $Pester.

adityapatwardhan (approved):

daxian-dbw (approved):
Thanks for your contribution!

PR #5072, opened 2017-10-09
Update instructions for building from source

kylesferrazza:
Internals.md updated to use the new System.Management.Automation dll required for TypeCatalogGen.
Also, build.sh no longer exists, so I added the custom msbuild target to the instructions.


TravisEz13:
@iSazonov Can you review?

iSazonov:
LGTM.

Out of the PR. 
@daxian-dbw About "lack of  GetAssemblies()" - it seems .Net Core 2.0 has it. Can we remove TypeCatalogGen? I don't found a tracking Issue.

daxian-dbw:
@iSazonov we cannot remove `TypeCatalogGen`. When you use a type like `[System.Net.Http.HttpClient]`, the type catalog generated by `TypeCatalogGen` kicks in to figure out which assembly should be loaded. The type catalog also plays a role when you are doing a tab completion on a type like `[httpcl<tab>`. The reason is that `GetAssemblies()` can only give you the loaded assemblies, so without the type catalog, you need to make sure `System.Net.Http.dll` is loaded before using any types from it.

daxian-dbw (approved):

iSazonov:
@daxian-dbw Thanks for clarify! 
This comment in the code is misleading and must be corrected.

daxian-dbw:
@iSazonov Agreed. I will send out a PR shortly to correct it.

PR #5073, opened 2017-10-09
Add black icns file for Mac.

thezim:


SteveL-MSFT (approved):
Looked at it on a Mac and looks good.  Thanks for the contribution!

PR #5080, opened 2017-10-10
Unify cmdlets with parameter 'Encoding' to be of type System.Text.Encoding

JamesWTruher:
This unifies file encoding across the inbox cmdlets to be UTF-8 without a BOM for all platforms. This is a breaking change as cmdlets on windows have a number of different encodings. This  supports better interoperability with tradition Linux shells as we are using the same encoding.

Use `ArgumentCompletionsAttribute` for the parameter to accept the currently used strings.
Register encoding providers on all platforms. This seems safe enough, and there are defintely more encodings available on CoreFX after this is done, which will be useful for some of the web cmdlet scenarios

Validate that files are created with UTF-8 encoding without BOM
Update tests to validate Encoding parameter to new type and create new tests for
parameter type validation.

*Breaking Change* The `-Encoding Byte` has been removed from the filesystem provider cmdlets. A new parameter -Byte is now the signal that a byte stream is required as input or output will be a stream of bytes.


src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1210,77 +1211,26 @@ internal static FileSystemCmdletProviderEncoding GetEncoding(string path)
...
-
-                default:
-                    break;
-            }
-
-            return result;
-        } // GetEncodingFromEnum
 
+        // BigEndianUTF32 encoding is possible, but requires creation
+        internal static Encoding BigEndianUTF32Encoding = new UTF32Encoding(true,true);
```
adityapatwardhan: Use named parameters instead of `true, true`

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -7658,27 +7613,15 @@ public class FileSystemContentDynamicParametersBase
...
         /// Gets the Byte Encoding status of the StreamType parameter.  Returns true
         /// if the stream was opened with "Byte" encoding, false otherwise.
         /// </summary>
         public bool UsingByteEncoding
         {
-            get
-            {
-                return Encoding == FileSystemCmdletProviderEncoding.Byte;
-            } // get
+            get { return _usingByteEncoding; }
```
adityapatwardhan: Please use automatic property and remove _usingByteEncoding.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -7688,9 +7631,14 @@ public bool WasStreamTypeSpecified
         {
             get
             {
-                return (Encoding != FileSystemCmdletProviderEncoding.String);
-            } // get
+                return _wasStreamTypeSpecified;
```
adityapatwardhan: Same as above.

src/System.Management.Automation/utils/ClrFacade.cs
```diff
@@ -109,16 +104,17 @@ internal static Encoding GetDefaultEncoding()
 
         /// <summary>
         /// Facade for getting OEM encoding
+        /// OEM encodings work on all platforms, or rather codepage 437 is available on both Windows
```
adityapatwardhan: Probably the comment intended to be ` codepage 437 is available on both Windows and Non-Windows`

src/System.Management.Automation/utils/PathUtils.cs
```diff
@@ -515,5 +542,104 @@ internal static Encoding Convert(Cmdlet cmdlet, string encoding)
...
             return null;
         }
+
+        // We need a way to provide an encoding and also
+        // a way to notify that the user wants "byte" encoding
+        // which is just a stream
+        internal class ByteEncoding : System.Text.Encoding
+        {
+            // the encoding for this is not bigendian and not BOM
+            public Encoding ActualEncoding = new UnicodeEncoding(false, false);
```
adityapatwardhan: Use named parameters

src/System.Management.Automation/utils/PathUtils.cs
```diff
@@ -515,5 +542,104 @@ internal static Encoding Convert(Cmdlet cmdlet, string encoding)
...
+                    string.Equals(encodingName, EncodingConversion.Utf8NoBom, StringComparison.OrdinalIgnoreCase) ||
+                    string.Equals(encodingName, EncodingConversion.Utf7, StringComparison.OrdinalIgnoreCase) ||
+                    string.Equals(encodingName, EncodingConversion.Utf32, StringComparison.OrdinalIgnoreCase) ||
+                    string.Equals(encodingName, EncodingConversion.Ascii, StringComparison.OrdinalIgnoreCase) ||
+                    string.Equals(encodingName, EncodingConversion.Default, StringComparison.OrdinalIgnoreCase) ||
+                    string.Equals(encodingName, EncodingConversion.OEM, StringComparison.OrdinalIgnoreCase))
+                {
+                    // the encodingName is guaranteed to be valid, so it is safe to pass null to method
+                    // Convert(Cmdlet cmdlet, string encoding) as the value of 'cmdlet'.
+                    return EncodingConversion.Convert(null, encodingName);
```
adityapatwardhan: use named parameter.

test/powershell/Modules/Microsoft.PowerShell.Core/TestGetCommand.Tests.ps1
```diff
@@ -211,7 +211,7 @@
         $paramName = "Encoding"
         $results = get-command -verb get -noun content -Encoding Unicode
```
adityapatwardhan: Now that you have touched the file. Can you fix the casing for the Get-Command cmdlets. 
get-command -> Get-Command 
-verb -> -Verb
-noun -> -Noun

test/powershell/engine/Basic/Encoding.Tests.ps1
```diff
@@ -0,0 +1,74 @@
+Describe "File encoding tests" -tag CI {
+
+    Context "ParameterType for parameter 'Encoding' should be 'Encoding'" {
+        BeforeAll {
+            $testCases = get-command -module Microsoft.PowerShell.* |
```
adityapatwardhan: Please fix Get-Command casing

test/powershell/engine/Basic/Encoding.Tests.ps1
```diff
@@ -0,0 +1,74 @@
...
+                # Redirection
+                @{ Command = { $testStr > $outputFile } ; Expected = $ExpectedWithNewline ; Operator = "be" }
+                )
+            function Get-FileBytes ( $path ) {
+                [io.file]::ReadAllBytes($path)
+            }
+        }
+
+        AfterEach { 
+            if ( test-path $outputFile ) {
```
adityapatwardhan: Casing for Test-Path, Remove-Item -Force

test/powershell/engine/Basic/Encoding.Tests.ps1
```diff
@@ -0,0 +1,74 @@
...
+
+        It "<command> produces correct content '<Expected>'" -testcases $simpleTestCases {
+            param ( $Command, $parameters, $Expected, $Operator)
+            & $command @parameters
+            $bytes = Get-FileBytes $outputFile
+            $bytes -join "-" | should ${Operator} ($Expected -join "-")
+        }
+
+        It "Export-CSV creates file with UTF-8 encoding without BOM" {
+            [pscustomobject]@{ Key = $testStr } | export-csv $outputFile
```
adityapatwardhan: Casing Export-Csv

test/powershell/engine/Basic/Encoding.Tests.ps1
```diff
@@ -0,0 +1,74 @@
...
+            $bytes = Get-FileBytes $outputFile
+            $bytes[0,1,2] -join "-" | should not be ($utf8Preamble -join "-")
+            $bytes -join "-" | should match ($utf8bytes -join "-")
+        }
+
+        It "Export-CliXml creates file with UTF-8 encoding without BOM" {
+            [pscustomobject]@{ Key = $testStr } | export-clixml $outputFile
```
adityapatwardhan: Casing Export-CliXml

test/powershell/engine/Basic/Encoding.Tests.ps1
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        It "Appends correctly on non-Windows systems" -skip:$IsWindows {
+            bash -c "echo '${testStr}' > $outputFile"
+            ${testStr} >> $outputFile
+            $bytes = Get-FileBytes $outputFile
+            $Expected = $( $ExpectedWithNewline; $ExpectedWithNewline )
+            $bytes -join "-" | should be ($Expected -join "-")
+        }
+
```
adityapatwardhan: Extra line?

test/powershell/engine/Basic/Encoding.Tests.ps1
```diff
@@ -0,0 +1,74 @@
...
+            if ( test-path $outputFile ) {
+                remove-item -force $outputFile
+            }
+        }
+
+        It "<command> produces correct content '<Expected>'" -testcases $simpleTestCases {
+            param ( $Command, $parameters, $Expected, $Operator)
+            & $command @parameters
+            $bytes = Get-FileBytes $outputFile
+            $bytes -join "-" | should ${Operator} ($Expected -join "-")
```
adityapatwardhan: If the operator is always `Be` we can directly use it here and remove it from `$simpleTestCases`

adityapatwardhan: Should this be `BeExactly`?

test/powershell/engine/Basic/Encoding.Tests.ps1
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        It "Export-CSV creates file with UTF-8 encoding without BOM" {
+            [pscustomobject]@{ Key = $testStr } | export-csv $outputFile
+            $bytes = Get-FileBytes $outputFile
+            $bytes[0,1,2] -join "-" | should not be ($utf8Preamble -join "-")
+            $bytes -join "-" | should match ($utf8bytes -join "-")
```
adityapatwardhan: Should this be `MatchExactly`

test/powershell/engine/Basic/Encoding.Tests.ps1
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        It "Export-CliXml creates file with UTF-8 encoding without BOM" {
+            [pscustomobject]@{ Key = $testStr } | export-clixml $outputFile
+            $bytes = Get-FileBytes $outputFile
+            $bytes[0,1,2] -join "-" | should not be ($utf8Preamble -join "-")
+            $bytes -join "-" | should match ($utf8bytes -join "-")
```
adityapatwardhan: Should this be `MatchExactly`

test/powershell/engine/Basic/Encoding.Tests.ps1
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        It "Appends correctly on non-Windows systems" -skip:$IsWindows {
+            bash -c "echo '${testStr}' > $outputFile"
+            ${testStr} >> $outputFile
+            $bytes = Get-FileBytes $outputFile
+            $Expected = $( $ExpectedWithNewline; $ExpectedWithNewline )
+            $bytes -join "-" | should be ($Expected -join "-")
```
adityapatwardhan: Should this be `BeExactly`

test/powershell/engine/Basic/Encoding.Tests.ps1
```diff
@@ -0,0 +1,74 @@
...
+
+    Context "ParameterType for parameter 'Encoding' should be 'Encoding'" {
+        BeforeAll {
+            $testCases = get-command -module Microsoft.PowerShell.* | 
+                Where-Object { $_.Parameters -and $_.Parameters['Encoding'] } |
+                ForEach-Object { @{ Command = $_ } }
+        }
+        It "Encoding parameter of command '<Command>' is type 'Encoding'" -testcase $testCases {
+            param ( $command )
+            $command.Parameters['Encoding'].ParameterType.FullName | Should Be "System.Text.Encoding"
```
adityapatwardhan: Should be `BeExactly`?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Send-MailMessage.cs
```diff
@@ -523,4 +523,4 @@ public override object Transform(EngineIntrinsics engineIntrinsics, object input
     }
 
     #endregion
-}
```
adityapatwardhan: Missed adding the change for the Encoding parameter?

adityapatwardhan (changes requested):

dantraMSFT:
>     internal static class EncodingConversion

I think this class should be in its own CS file; it's not a path utility and discoverability is problematic. 

---
Refers to: src/System.Management.Automation/utils/PathUtils.cs:446 in 6ca1339. [](commit_id = 6ca13399470454bdc9a7087e2f1d95b494bd3531, deletion_comment = False)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -206,8 +206,27 @@ public SwitchParameter NoClobber
...
+        [ArgumentCompleter(typeof(EncodingArgumentCompleter))]
+        public Encoding Encoding
+        {
+            get
+            {
+                return _encoding;
+            }
+            set
+            {
+                if ( value == EncodingConversion.byteEncoding )
```
dantraMSFT: I believe you want to use Object.ReferenceEquals or override Equals in your ByteEncoding class and handle equality there; otherwise, you're relying on Encoding.Equals which may not do what you expect.

dantraMSFT: These set methods need a comment as to why the comparison to EncodingConversion.byteEncoding is needed to provide a guide when creating new cmdlets that provide an Encoding parameter.

src/System.Management.Automation/utils/PathUtils.cs
```diff
@@ -515,5 +542,104 @@ internal static Encoding Convert(Cmdlet cmdlet, string encoding)
 
             return null;
         }
+
+        // We need a way to provide an encoding and also
+        // a way to notify that the user wants "byte" encoding
+        // which is just a stream
+        internal class ByteEncoding : System.Text.Encoding
```
dantraMSFT: Should this class and the EncodingConversion.byteEncoding moved here as a public field?

In other words, is the problem you're solving with this custom instance be something third-party cmdlets will need to solve as well?

src/System.Management.Automation/utils/ClrFacade.cs
```diff
@@ -93,14 +93,9 @@ internal static Encoding GetDefaultEncoding()
...
-#if UNIX        // PowerShell Core on Unix
-                s_defaultEncoding = new UTF8Encoding(false);
-#else           // PowerShell Core on Windows
+                // load all available encodings
                 EncodingRegisterProvider();
-
-                uint currentAnsiCp = NativeMethods.GetACP();
-                s_defaultEncoding = Encoding.GetEncoding((int)currentAnsiCp);
-#endif
+                s_defaultEncoding = new UTF8Encoding(false);
```
iSazonov: I don't like the change. Previously (discussion #3248, PR #3467) I fixed this for Windows because we're still very sensitive to the OEM encoding. We have tons applications in enterprises that doesn't work if we don't set the correct OEM encoding on the system. If MSFT wants to cut it off, do it first in Windows 10 kernel and get feedback.

test/powershell/Language/Parser/RedirectionOperator.Tests.ps1
```diff
@@ -29,13 +29,14 @@ Describe "Redirection operator now supports encoding changes" -Tags "CI" {
...
+    It "If encoding is unset, redirection should be UTF8 without bom" {
         $asciiString > TESTDRIVE:\file.txt
         $bytes = get-content -encoding byte TESTDRIVE:\file.txt
-        # create the expected
-        $BOM = [text.encoding]::unicode.GetPreamble()
-        $TXT = [text.encoding]::unicode.GetBytes($asciiString)
-        $CR  = [text.encoding]::unicode.GetBytes($asciiCR)
+        # create the expected - utf8 encoding without a bom
+        $encoding = [text.utf8encoding]::new($false)
+        $BOM = $encoding.GetPreamble()
```
dantraMSFT: This is very misleading; the test is verifying with a BOM yet the test is requesting the BOM.  If the Encoding class has a bug or was created incorrectly, the test will still pass.  In other words, if the array is empty, as expected, the test passes. If the array is non-empty, the test will still pass.

I suggest removing $BOM completely and not relying on the implicit empty array.

test/powershell/Language/Parser/RedirectionOperator.Tests.ps1
```diff
@@ -29,13 +29,14 @@ Describe "Redirection operator now supports encoding changes" -Tags "CI" {
...
-    It "If encoding is unset, redirection should be Unicode" {
+    It "If encoding is unset, redirection should be UTF8 without bom" {
         $asciiString > TESTDRIVE:\file.txt
         $bytes = get-content -encoding byte TESTDRIVE:\file.txt
-        # create the expected
-        $BOM = [text.encoding]::unicode.GetPreamble()
-        $TXT = [text.encoding]::unicode.GetBytes($asciiString)
-        $CR  = [text.encoding]::unicode.GetBytes($asciiCR)
+        # create the expected - utf8 encoding without a bom
+        $encoding = [text.utf8encoding]::new($false)
```
dantraMSFT: Suggest matching the casing of the type [Text.UTF8Encoding].

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
```diff
@@ -99,13 +99,12 @@ baz
 "@
         $expected = 'foo'
         $tailCount = 3
-        [Microsoft.PowerShell.Commands.FileSystemCmdletProviderEncoding] $encoding = $EncodingName
 
         $testPath = Join-Path -Path $TestDrive -ChildPath 'TailWithEncoding.txt'
-        $content | Set-Content -Path $testPath -Encoding $encoding
+        $content | Set-Content -Path $testPath -Encoding $encodingName
```
dantraMSFT: Should this test cover the ByteEncoding path?  It doesn't appear to.

dantraMSFT (changes requested):

src/System.Management.Automation/utils/PathUtils.cs
```diff
@@ -459,7 +478,7 @@ internal static Encoding Convert(Cmdlet cmdlet, string encoding)
         {
             if (string.IsNullOrEmpty(encoding))
             {
-                // no parameter passed, default to Unicode (OS preferred)
+                // no parameter passed, default to  (OS preferred)
```
iSazonov: Typo?

src/System.Management.Automation/utils/PathUtils.cs
```diff
@@ -445,10 +451,23 @@ internal static class EncodingConversion
...
         internal const string Utf7 = "utf7";
         internal const string Utf32 = "utf32";
         internal const string Default = "default";
         internal const string OEM = "oem";
+        internal const string Byte = "byte"; // This will return unicode, but 'byte' is special as it uses unicode encoding without a BOM
+        internal static ByteEncoding byteEncoding = new ByteEncoding();
+        // We will provide a partial list for tab completion we will convert to an encoding.
+        // If the user has an encoding object, we will cheerfully use it, but we can't use
+        // it as a validation set, as a user may have an specialty encoding which we have
+        // no idea about.
```
iSazonov: I think we could use `ArgumentCompletionsAttribute` #4835

src/System.Management.Automation/utils/PathUtils.cs
```diff
@@ -500,9 +528,8 @@ internal static Encoding Convert(Cmdlet cmdlet, string encoding)
             // error condition: unknown encoding value
             string validEncodingValues = string.Join(
                 ", ",
-                new string[] { Unknown, String, Unicode, BigEndianUnicode, Ascii, Utf8, Utf7, Utf32, Default, OEM });
-            string msg = StringUtil.Format(PathUtilsStrings.OutFile_WriteToFileEncodingUnknown,
-                encoding, validEncodingValues);
+                new string[] { Unknown, String, Unicode, BigEndianUnicode, Byte, Ascii, Utf8, Utf8Bom, Utf8NoBom, Utf7, Utf32, Default, OEM });
```
iSazonov: Can we reuse `TabCompletionResults` ?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -206,8 +206,27 @@ public SwitchParameter NoClobber
         /// Encoding optional flag
         /// </summary>
         [Parameter()]
-        [ValidateSetAttribute(new string[] { "Unicode", "UTF7", "UTF8", "ASCII", "UTF32", "BigEndianUnicode", "Default", "OEM" })]
-        public string Encoding { get; set; }
+        [ArgumentToEncodingTransformationAttribute()]
+        [ArgumentCompleter(typeof(EncodingArgumentCompleter))]
```
iSazonov: I think we could use `ArgumentCompletionsAttribute` #4835

src/System.Management.Automation/engine/Utils.cs
JamesWTruher: fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
JamesWTruher: fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
JamesWTruher: fixed

src/System.Management.Automation/utils/ClrFacade.cs
JamesWTruher: indeed so, fixed

src/System.Management.Automation/utils/PathUtils.cs
JamesWTruher: fixed

test/powershell/Modules/Microsoft.PowerShell.Core/TestGetCommand.Tests.ps1
JamesWTruher: done

test/powershell/engine/Basic/Encoding.Tests.ps1
JamesWTruher: done

test/powershell/engine/Basic/Encoding.Tests.ps1
JamesWTruher: it could be - `BeExactly` is a case-sensitive string compare, but the case of the response is not the point, only that it is the proper fullname.
I'm in the file so I'll update it, but we should be cautious about being overly pedantic.

test/powershell/engine/Basic/Encoding.Tests.ps1
JamesWTruher: fixed

test/powershell/engine/Basic/Encoding.Tests.ps1
JamesWTruher: fixed

test/powershell/engine/Basic/Encoding.Tests.ps1
JamesWTruher: not needed (as they're numbers) - `MatchExactly` isn't the same as `Be`, it's just a case-sensitive match `-cmatch` I just need to be sure that my string is in the output somewhere

test/powershell/engine/Basic/Encoding.Tests.ps1
JamesWTruher: fixed

test/powershell/engine/Basic/Encoding.Tests.ps1
JamesWTruher: not needed - for the same reason as above, i'm just making sure that the magic bytes are in the file somewhere

test/powershell/engine/Basic/Encoding.Tests.ps1
JamesWTruher: nope - they're numbers so case doesn't provide value

test/powershell/engine/Basic/Encoding.Tests.ps1
JamesWTruher: removed

src/System.Management.Automation/utils/PathUtils.cs
JamesWTruher: indeed so - fixed

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
lzybkr: Maybe add a helper, something like:
```c#
EncodingConversion.SetEncodingParameterHelper(ref _encoding, value);
```
Or:
```c#
_encoding = EncodingConversion.EncodingParameterHelper(value);
```


---
In reply to: [144073559](https://github.com/PowerShell/PowerShell/pull/5080#discussion_r144073559) [](ancestors = 144073559)

src/System.Management.Automation/utils/PathUtils.cs
```diff
@@ -470,6 +489,9 @@ internal static Encoding Convert(Cmdlet cmdlet, string encoding)
             if (string.Equals(encoding, String, StringComparison.OrdinalIgnoreCase))
                 return System.Text.Encoding.Unicode;
 
+            if (string.Equals(encoding, Byte, StringComparison.OrdinalIgnoreCase))
+                return byteEncoding;
```
lzybkr: If we aren't using a dictionary for fast lookup, we can at least order the tests from most likely (guessing of course) to least likely.

e.g., is Unknown really the most likely case? I hope not.

src/System.Management.Automation/utils/PathUtils.cs
```diff
@@ -515,5 +542,104 @@ internal static Encoding Convert(Cmdlet cmdlet, string encoding)
...
+
+        // We need a way to provide an encoding and also
+        // a way to notify that the user wants "byte" encoding
+        // which is just a stream
+        internal class ByteEncoding : System.Text.Encoding
+        {
+            // the encoding for this is not bigendian and not BOM
+            public Encoding ActualEncoding = new UnicodeEncoding(false, false);
+            public override unsafe int GetByteCount(char* c, int i1) { throw new NotImplementedException(); }
+            public override int GetByteCount(char[] c, int i1, int i2) { throw new NotImplementedException(); }
```
lzybkr: I believe it's not hard for someone to get an instance of this encoding, so maybe we should be implementing all of these methods, assuming it's straightforward to do so (and mostly it seems like it is.)

src/System.Management.Automation/utils/PathUtils.cs
JamesWTruher: reordered

src/System.Management.Automation/utils/PathUtils.cs
JamesWTruher: done

src/System.Management.Automation/utils/PathUtils.cs
JamesWTruher: this is really a dodge for backward compatibility. `-Encoding Byte` should have never existed, but it does and I don't want to have this be a difference between 6 and 5.x. This _byte_ encoding is really needed for the *-content cmdlets where the parameter is used as a flag indicating that the user is looking for a stream of bytes, I would rather not make this public

test/powershell/Language/Parser/RedirectionOperator.Tests.ps1
JamesWTruher: done

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Content.Tests.ps1
JamesWTruher: covered by the tests in engine/Basic

test/powershell/Language/Parser/RedirectionOperator.Tests.ps1
JamesWTruher: I fixed up the cases of everything while I was there

src/System.Management.Automation/utils/ClrFacade.cs
JamesWTruher: @iSazonov where are tests which provide validation for the changes you made in PR3467? The discussion in #3248 doesn't seem to have any concrete examples of misbehavior, which I would really like to understand. As for your last, I would rather get feedback on breaking behavior outside of the Windows ship cycles.

src/System.Management.Automation/utils/PathUtils.cs
SteveL-MSFT: Since we're already making this big breaking change, I think we should fix this by not having `ByteEncoding` as `Byte` is an artificial "encoding".  The `*-Content` cmdlets should have a different parameter to output or receive a byte stream.

src/System.Management.Automation/utils/ClrFacade.cs
SteveL-MSFT: @iSazonov, @PowerShell/powershell-committee has already decided to go with UTF8 (No BOM) as default.  Can you give me a specific example where this is known to break?  If OEM is needed, perhaps we could add a setting in the powershell json file but still default to utf8

SteveL-MSFT (changes requested):

src/System.Management.Automation/utils/PathUtils.cs
iSazonov: It would more clear to users to have -Byte or -Binary parameter.
@mklement0 @rkeithhill What do you think?

SteveL-MSFT: I kind of prefer `-AsByteArray` myself.

cc @HemantMahawar @joeyaiello

src/System.Management.Automation/utils/ClrFacade.cs
iSazonov: PR #3467#3467 haven't tests because (1) we did not have a full set of encoding tests in the time (2) we waited the encoding RFC (3) we only restore Windows PowerShell behavior by removing a breaking change.
About concrete examples. I speak about business applications. I have no idea how I can demonstrate it concrete. I understand that you hardly ever changed the OEM encoding on your computers. But ask yourself why are customers forced to do so? If MSTF wants to see it and understand it, it needs TAP with non-english enterprises. As I said earlier, the simple answer is that we have many business applications that only work with a specific OEM encoding and read/write files in OEM. PowerShell also works in _this environment_! The #3467 was caused by the fact that the scripts were no longer working and we needed to re-encode the files. The value of PowerShell Core disappears completely.

We can look at it on the other side. A year ago there was a categorical position that we should be backward compatible and that we should avoid making breaking changes as much as possible. The position was later changed so we want to facilitate the adaptation of UNIX users. This led to the creation of many incompatible changes. The current change from this series. These two principles have proved to be contradictory. We need to decide which one's in charge.

If the first means that we want to allow multiple Windows users to use PowerShell on Unix without much effort, and to get feedback to start the process of evolutionary convergence of platforms.

The second actually means that we're creating a brand new product that cannot be version 6, but must be PowerShell Core 1.0. It's fair to say that Windows users will have to migrate to a new product instead of using it directly. In this case, I agree that we can be ahead of the Windows 10 system and even its Notepad and use Utf8 by default like all other changes. Only then will it not be expected that the community will begin to use PowerShell Core extensively before version 3.
The second way is very dangerous. Microsoft can break the Windows PowerShell community and not build a new one-there are already well-known and popular alternatives (PHP, JS, Perl) that have proven their reliability and performance and have already formed a significant community. I've seen some examples of Microsoft already quickly destroying communities. The recovery took several years. It's sad. If it were my business, I would prefer to keep today's real customers, not future abstracts, and make the first step similar WSL like Windows PowerShell for Unix to capture a reliable bridgehead. Only then the community would be prepared to accept any breaking changes as understandable and necessary. Sorry for many words :-)

SteveL-MSFT: It seems again that a viable option is to have a setting for powershell.runtimeconfig.json to have this be OEM for compat reasons.

src/System.Management.Automation/utils/PathUtils.cs
mklement0: Purely from a usability perspective, I don't have a problem with `-Encoding Byte`, because even though it is technically not a _text_ encoding, the `-Encoding` parameter can be conceptualized as being about how to _interpret_ the data on reading and writing, and `-Encoding Byte` (a byte stream) is one way of (non-)interpreting .

This keeps all considerations regarding encoding/non-encoding focused on a single parameter, which simplifies things (it doesn't introduce another parameter that you have to learn about and remember), both in terms of use and documentation.

src/System.Management.Automation/utils/ClrFacade.cs
mklement0: I haven't followed this closely, but what happened to the `$PSDefaultEncoding = 'WindowsLegacy'` proposal from the RFC?

mklement0: And, just to clarify: for backward compatibility, it's both OEM _and "ANSI"_ code pages that must be considered.

src/System.Management.Automation/utils/PathUtils.cs
```diff
@@ -445,10 +451,23 @@ internal static class EncodingConversion
         internal const string BigEndianUnicode = "bigendianunicode";
         internal const string Ascii = "ascii";
         internal const string Utf8 = "utf8";
+        internal const string Utf8NoBom = "utf8NoBOM";
+        internal const string Utf8Bom = "utf8BOM";
         internal const string Utf7 = "utf7";
         internal const string Utf32 = "utf32";
         internal const string Default = "default";
         internal const string OEM = "oem";
+        internal const string Byte = "byte"; // This will return unicode, but 'byte' is special as it uses unicode encoding without a BOM
```
mklement0: Not knowing the inner workings: How will this return _Unicode_? With `Get-Content`, you get a _byte_ array, and on output  (`Set-Content` / `Add-Content`) you're forced to provide a byte array _as input_ too (however, you constructed that).

src/System.Management.Automation/utils/ClrFacade.cs
SteveL-MSFT: @mklement0 the simplified version we scoped for 6.0.0 does not include `WindowsLegacy`

src/System.Management.Automation/utils/PathUtils.cs
iSazonov: How can we interpret `Set-Content -Encoding 'Ru-ru' -Byte`?

JamesWTruher:
@adityapatwardhan @dantraMSFT @SteveL-MSFT I've believe I've made updates that address your blocking issues

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -7477,73 +7477,6 @@ public static Hashtable Invoke(System.Management.Automation.PowerShell ps, FileS
         }
     }
 
-    /// <summary>
-    /// Defines the values that can be supplied as the encoding parameter in the
-    /// FileSystemContentDynamicParametersBase class.
-    /// </summary>
-    public enum FileSystemCmdletProviderEncoding
```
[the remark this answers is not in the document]
SteveL-MSFT: Jim, can you check if this API is in PowerShell Standard 3.0?  This enum isn't used at all now, right?  So leaving it in with Obsolete doesn't help.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and has agreed on:

1. name the parameter `-AsByteArray`
2. output warning if `-Encoding` is used with `-AsByteArray` that Encoding is not being used

mklement0:
A message from the COSP (Crying Over Spilt Milk) department:

> output warning if -Encoding is used with -AsByteArray that Encoding is not being used

This results in an awkward user experience, because users will expect incompatible parameters to be in different parameter sets, with attempts to combine them resulting in an _error_, not a warning.

A step backward in usability, and a breaking change to boot - and no gain that I can discern.

I also agree with @iSazonov that forgoing compatibility with the legacy encoding behavior on Windows will dismay many Windows users, especially those whose "ANSI" / OEM code pages aren't based on the Latin alphabet.

If the idea is to focus on the Unix crowd first, I see trouble as well: broken quoting, broken CLI, lack of `&&` and `||`, ...

SteveL-MSFT:
@mklement one of the considerations the committee acknowledged is the complexity of using multiple parametersets to make parameters mutually exclusive so we were ok with using a warning as a simplification until we can revisit the recalled RFC on making it easier to both define as well as understand mutually exclusive parameters

adityapatwardhan (approved):
LGTM

src/System.Management.Automation/utils/ClrFacade.cs
SteveL-MSFT: @JamesWTruher can you open an issue to cover @iSazonov 's concern?

SteveL-MSFT (approved):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -206,8 +206,9 @@ public SwitchParameter NoClobber
...
+        public Encoding Encoding = ClrFacade.GetDefaultEncoding();
```
daxian-dbw: This is a public field. I think we need it to be a property.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -571,8 +572,9 @@ public SwitchParameter UseCulture
         /// Encoding optional flag
         /// </summary>
         [Parameter()]
-        [ValidateSetAttribute(new[] { "Unicode", "UTF7", "UTF8", "ASCII", "UTF32", "BigEndianUnicode", "Default", "OEM" })]
-        public string Encoding { get; set; }
+        [ArgumentToEncodingTransformationAttribute()]
+        [ArgumentCompleter(typeof(EncodingArgumentCompleter))]
+        public Encoding Encoding = ClrFacade.GetDefaultEncoding();
```
daxian-dbw: Same here, should make it a property.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/format-hex/Format-Hex.cs
```diff
@@ -46,14 +46,9 @@ public sealed class FormatHex : PSCmdlet
...
-            EncodingConversion.Unicode,
-            EncodingConversion.BigEndianUnicode,
-            EncodingConversion.Utf8,
-            EncodingConversion.Utf7,
-            EncodingConversion.Utf32,
-            EncodingConversion.Ascii})]
-        public string Encoding { get; set; } = "Ascii";
+        [ArgumentToEncodingTransformationAttribute()]
+        [ArgumentCompleter(typeof(EncodingArgumentCompleter))]
+        public Encoding Encoding = ClrFacade.GetDefaultEncoding();
```
daxian-dbw: Same here.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/out-file/Out-File.cs
```diff
@@ -72,25 +73,9 @@ public string LiteralPath
         /// </summary>
         ///
         [Parameter(Position = 1)]
-        [ValidateNotNullOrEmpty]
```
daxian-dbw: I think we still need this attribute, so that user cannot pass in a null value.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -206,8 +206,9 @@ public SwitchParameter NoClobber
         /// Encoding optional flag
         /// </summary>
         [Parameter()]
-        [ValidateSetAttribute(new string[] { "Unicode", "UTF7", "UTF8", "ASCII", "UTF32", "BigEndianUnicode", "Default", "OEM" })]
-        public string Encoding { get; set; }
+        [ArgumentToEncodingTransformationAttribute()]
```
daxian-dbw: I think we need to add `AttributeNotNullOrEmpty` attribute. Before this change, $null value would be rejected because the validate attribute will fail. After this change, user is able to pass in a null value.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -571,8 +572,9 @@ public SwitchParameter UseCulture
         /// Encoding optional flag
         /// </summary>
         [Parameter()]
-        [ValidateSetAttribute(new[] { "Unicode", "UTF7", "UTF8", "ASCII", "UTF32", "BigEndianUnicode", "Default", "OEM" })]
-        public string Encoding { get; set; }
+        [ArgumentToEncodingTransformationAttribute()]
+        [ArgumentCompleter(typeof(EncodingArgumentCompleter))]
```
daxian-dbw: We need a `ValidateNotNullOrEmtpy` attribute here.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/format-hex/Format-Hex.cs
```diff
@@ -46,14 +46,9 @@ public sealed class FormatHex : PSCmdlet
...
-        [ValidateSetAttribute(new string[] {
-            EncodingConversion.Unicode,
-            EncodingConversion.BigEndianUnicode,
-            EncodingConversion.Utf8,
-            EncodingConversion.Utf7,
-            EncodingConversion.Utf32,
-            EncodingConversion.Ascii})]
-        public string Encoding { get; set; } = "Ascii";
+        [ArgumentToEncodingTransformationAttribute()]
+        [ArgumentCompleter(typeof(EncodingArgumentCompleter))]
```
daxian-dbw: Need `ValidateNullOrEmpty` here.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ImplicitRemotingCommands.cs
```diff
@@ -77,19 +77,9 @@ public SwitchParameter Force
...
-            }
-            set
-            {
-                _encoding = EncodingConversion.Convert(this, value);
-            }
-        }
-        private Encoding _encoding = System.Text.Encoding.UTF8;
+        [ArgumentToEncodingTransformationAttribute()]
+        [ArgumentCompleter(typeof(EncodingArgumentCompleter))]
+        public Encoding Encoding = ClrFacade.GetDefaultEncoding();
```
daxian-dbw: Need to be a property.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ImplicitRemotingCommands.cs
```diff
@@ -77,19 +77,9 @@ public SwitchParameter Force
...
-            {
-                return _encoding.GetType().Name;
-            }
-            set
-            {
-                _encoding = EncodingConversion.Convert(this, value);
-            }
-        }
-        private Encoding _encoding = System.Text.Encoding.UTF8;
+        [ArgumentToEncodingTransformationAttribute()]
```
daxian-dbw: Need `ValidateNullOrEmpty`.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
```diff
@@ -1201,19 +1202,9 @@ public SwitchParameter AllMatches
...
-            EncodingConversion.Ascii,
-            EncodingConversion.BigEndianUnicode,
-            EncodingConversion.Default,
-            EncodingConversion.OEM })]
-        public string Encoding { get; set; }
-
-        private System.Text.Encoding _textEncoding;
+        [ArgumentToEncodingTransformationAttribute()]
+        [ArgumentCompleter(typeof(EncodingArgumentCompleter))]
+        public Encoding Encoding = ClrFacade.GetDefaultEncoding();
```
daxian-dbw: Same comments here.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/Send-MailMessage.cs
```diff
@@ -89,20 +89,14 @@ public SwitchParameter BodyAsHtml
...
-            get { return _encoding; }
-            set
-            {
-                _encoding = value;
-            }
-        }
-        private Encoding _encoding = new ASCIIEncoding();
+        [ArgumentCompleter(typeof(EncodingArgumentCompleter))]
+        [ArgumentToEncodingTransformationAttribute()]
+        public Encoding Encoding { get; set; } = new ASCIIEncoding();
```
daxian-dbw: It's better to use `Encoding.ASCII`, so that we don't create a new object.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/XmlCommands.cs
```diff
@@ -108,8 +108,9 @@ public SwitchParameter NoClobber
         /// </summary>
         ///
         [Parameter]
-        [ValidateSetAttribute(new string[] { "Unicode", "UTF7", "UTF8", "ASCII", "UTF32", "BigEndianUnicode", "Default", "OEM" })]
-        public string Encoding { get; set; } = "Unicode";
+        [ArgumentToEncodingTransformationAttribute()]
+        [ArgumentCompleter(typeof(EncodingArgumentCompleter))]
+        public Encoding Encoding = ClrFacade.GetDefaultEncoding();
```
daxian-dbw: Should be a property.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/XmlCommands.cs
```diff
@@ -108,8 +108,9 @@ public SwitchParameter NoClobber
         /// </summary>
         ///
         [Parameter]
-        [ValidateSetAttribute(new string[] { "Unicode", "UTF7", "UTF8", "ASCII", "UTF32", "BigEndianUnicode", "Default", "OEM" })]
-        public string Encoding { get; set; } = "Unicode";
+        [ArgumentToEncodingTransformationAttribute()]
+        [ArgumentCompleter(typeof(EncodingArgumentCompleter))]
```
daxian-dbw: need `validatenotnulloremtpy`

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1210,77 +1211,26 @@ internal static FileSystemCmdletProviderEncoding GetEncoding(string path)
             string initialBytesAsAscii = System.Text.Encoding.ASCII.GetString(initialBytes, 0, bytesRead);
             if (initialBytesAsAscii.IndexOfAny(nonPrintableCharacters) >= 0)
             {
-                return FileSystemCmdletProviderEncoding.Byte;
+                return Encoding.Unicode;
```
daxian-dbw: Returning `Encoding.Unicode` doesn't seem right. It looks we should deal with bytes in this case.

daxian-dbw (changes requested):
I'm now looking at `FileSystemProvider.cs`, but want to share my feedback early.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
dantraMSFT: You have a few cases where you declare the parameter as a field instead of a property and it doesn't appear that you enforce non-null.  In this case, the null Encoding is passed to new StreamReader which throws an ArgumentNullException.   

In this particular case, the parameter is not marked as mandatory and an exception should not be raised if not provided.

src/System.Management.Automation/utils/EncodingUtils.cs
```diff
@@ -0,0 +1,136 @@
...
+    /// To make it easier to specify -Encoding parameter, we add an ArgumentTransformationAttribute here.
+    /// When the input data is of type string and is valid to be converted to System.Text.Encoding, we do
+    /// the conversion and return the converted value. Otherwise, we just return the input data.
+    /// </summary>
+    internal sealed class ArgumentToEncodingTransformationAttribute : ArgumentTransformationAttribute
+    {
+        public override object Transform(EngineIntrinsics engineIntrinsics, object inputData)
+        {
+            string encodingName;
+            if (LanguagePrimitives.TryConvertTo<string>(inputData, out encodingName))
```
daxian-dbw: If the `inputData` is already an Encoding object, shouldn't we just let it pass by?

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -7647,7 +7590,28 @@ public class FileSystemContentDynamicParametersBase
         /// reading data from the file.
         /// </summary>
         [Parameter]
-        public FileSystemCmdletProviderEncoding Encoding { get; set; } = FileSystemCmdletProviderEncoding.String;
+        [ArgumentToEncodingTransformationAttribute()]
+        [ArgumentCompleter(typeof(EncodingArgumentCompleter))]
```
daxian-dbw: I think we also need a `ValidateNullOrEmpty` here.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
```diff
@@ -1434,7 +1415,7 @@ private bool ProcessFile(string filename)
 
                 using (FileStream fs = new FileStream(filename, FileMode.Open, FileAccess.Read, FileShare.ReadWrite))
                 {
-                    using (StreamReader sr = new StreamReader(fs, _textEncoding))
+                    using (StreamReader sr = new StreamReader(fs, Encoding))
```
dantraMSFT: There's an ArgumentNullException path here if Encoding is null.

src/System.Management.Automation/resources/FileSystemProviderStrings.resx
```diff
@@ -243,6 +243,9 @@
   <data name="TargetCannotContainDeviceName" xml:space="preserve">
     <value>Cannot process path '{0}' because the target represents a reserved device name.</value>
   </data>
+  <data name="EncodingNotUsed" xml:space="preserve">
+    <value>Encoding not used when -AsByteStream specified.</value>
```
daxian-dbw: Minor comment: maybe change to `... when '-AsByteStream' is specified`?

dantraMSFT (dismissed):
The change from property to field for some of the Encoding parameters needs to be reviewed for null references. MatchString.cs is one case where the parameter is a field, it is not mandatory, not marked as ValidateNotNull and is passed to an underlying StreamReader unchecked.

src/System.Management.Automation/utils/EncodingUtils.cs
daxian-dbw: And do we really need to convert `inputData` to `string` here? Why not just use `string encodingName = inputData as string`?

src/System.Management.Automation/utils/EncodingUtils.cs
```diff
@@ -0,0 +1,136 @@
...
+            System.Collections.IDictionary fakeBoundParameters)
+        {
+            List<CompletionResult> encodings = new List<CompletionResult>();
+            foreach(string encoding in EncodingConversion.TabCompletionResults)
+            {
+                if (string.IsNullOrEmpty(wordToComplete))
+                {
+                    encodings.Add(new CompletionResult(encoding, encoding, CompletionResultType.Text, encoding));
+                }
+                else if (encoding.IndexOf(wordToComplete, StringComparison.InvariantCultureIgnoreCase) == 0)
```
daxian-dbw: I think `wordToComplete` could have wildcards in it. For example, `Get-Content -Encoding *ianUnicode<tab>` ==> `Get-Content -Encoding BigEndianUnicode`

src/System.Management.Automation/utils/PathUtils.cs
```diff
@@ -188,6 +188,12 @@ internal static void ReportFileOpenFailure(Cmdlet cmdlet, string filePath, Excep
             cmdlet.ThrowTerminatingError(errorRecord);
         }
 
+        internal static StreamReader OpenStreamReader(PSCmdlet command, string filePath, Encoding encoding, bool isLiteralPath)
+        {
+            FileStream fileStream = OpenFileStream(filePath, command, isLiteralPath);
+            return new StreamReader(fileStream, encoding);
+        }
+
         internal static StreamReader OpenStreamReader(PSCmdlet command, string filePath, string encoding, bool isLiteralPath)
```
daxian-dbw: Is this method still used anywhere? It looks to me not used anymore. If that's the case, then it's better to remove it.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/Utils.cs
JamesWTruher: There is no byte encoding, and FileSystemCmdletProviderEncoding.Byte wound up being converted to a unicode encoding in GetEncodingFromEnum (now removed) so there's actually no change here.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
JamesWTruher: I've added `[ValidateNotNullOrEmpty]` to all of these.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
JamesWTruher: fixed with `[ValidateNotNullOrEmpty]` attribute

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -6614,12 +6614,17 @@ public IContentReader GetContentReader(string path)
                         delimiter = dynParams.Delimiter;
 
                     // Get the stream type
-                    usingByteEncoding = dynParams.UsingByteEncoding;
+                    usingByteEncoding = dynParams.AsByteStream;
                     streamTypeSpecified = dynParams.WasStreamTypeSpecified;
 
+                    if (usingByteEncoding && streamTypeSpecified)
+                    {
+                        WriteWarning(StringUtil.Format(FileSystemProviderStrings.EncodingNotUsed));
```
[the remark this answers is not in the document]
JamesWTruher: fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -6765,12 +6770,17 @@ public IContentWriter GetContentWriter(string path)
 
                 if (dynParams != null)
                 {
-                    usingByteEncoding = dynParams.UsingByteEncoding;
+                    usingByteEncoding = dynParams.AsByteStream;
                     streamTypeSpecified = dynParams.WasStreamTypeSpecified;
 
+                    if (usingByteEncoding && streamTypeSpecified)
+                    {
+                        WriteWarning(StringUtil.Format(FileSystemProviderStrings.EncodingNotUsed));
```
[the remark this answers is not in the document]
JamesWTruher: fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -7658,39 +7622,10 @@ public class FileSystemContentDynamicParametersBase
...
         /// if the stream was opened with a user-specified encoding, false otherwise.
         /// </summary>
-        public bool WasStreamTypeSpecified
-        {
-            get
-            {
-                return (Encoding != FileSystemCmdletProviderEncoding.String);
-            } // get
-        } // WasStreamTypeSpecified
+        public bool WasStreamTypeSpecified { get; set; }
```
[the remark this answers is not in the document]
JamesWTruher: done

src/System.Management.Automation/resources/FileSystemProviderStrings.resx
JamesWTruher: done

src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
JamesWTruher: I've fixed up all the parameters to use [ValidateNotNullOrEmpty]

src/System.Management.Automation/utils/EncodingUtils.cs
JamesWTruher: I can add that

src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
JamesWTruher: fixed with attribute

src/System.Management.Automation/utils/PathUtils.cs
JamesWTruher: removed

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
JamesWTruher: fixed

src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/format-hex/Format-Hex.cs
JamesWTruher: fixed

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ImplicitRemotingCommands.cs
JamesWTruher: fixed

src/Microsoft.PowerShell.Commands.Utility/commands/utility/XmlCommands.cs
JamesWTruher: fixed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
JamesWTruher: fixed

src/System.Management.Automation/utils/ClrFacade.cs
JamesWTruher: @mklement0 WindowsLegacy was pulled waiting for customer need - opened https://github.com/PowerShell/PowerShell/issues/5204 to track.

src/System.Management.Automation/utils/EncodingUtils.cs
```diff
@@ -0,0 +1,139 @@
...
+            Encoding foundEncoding;
+
+            if ( inputData is Encoding )
+            {
+                return inputData;
+            }
+            if (EncodingConversion.encodingMap.TryGetValue(encodingName, out foundEncoding))
+            {
+                return foundEncoding;
+            }
```
daxian-dbw: It looks to me we only care about convert a stirng to an encoding. I suggest change the body to the following
```
string encodingName = inputData as String;
Encoding foundEncoding;
if (encodingName != null && EncodingConversion.encodingMap.TryGetValue(encodingName, out foundEncoding))
{
    return foundEncoding;
}
return inputData.
```

src/System.Management.Automation/utils/EncodingUtils.cs
```diff
@@ -0,0 +1,139 @@
...
+            return inputData;
+        }
+
+    }
+
+    internal class EncodingArgumentCompleter : IArgumentCompleter
```
daxian-dbw: This is not necessary. You should use [`ArgumentCompletionsAttribute`](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/CommandCompletion/ExtensibleCompletion.cs#L172) instead for the parameters. Here is an example of using it: https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetDateCommand.cs#L210

daxian-dbw (changes requested):

src/System.Management.Automation/utils/EncodingUtils.cs
JamesWTruher: done

src/System.Management.Automation/utils/EncodingUtils.cs
JamesWTruher: fixed

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -212,8 +212,20 @@ public SwitchParameter NoClobber
...
+            EncodingConversion.Ascii,
+            EncodingConversion.BigEndianUnicode,
+            EncodingConversion.OEM,
+            EncodingConversion.Unicode,
+            EncodingConversion.Utf7,
+            EncodingConversion.Utf8,
+            EncodingConversion.Utf8Bom,
+            EncodingConversion.Utf8NoBom,
+            EncodingConversion.Utf32
+            )]
```
daxian-dbw: How about use `ArgumentCompletions(EncodingConversion.TabCompletionResults)` here? It would be easy for future updates too.

daxian-dbw (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
daxian-dbw: Same comments for other instances of `ArgumentCompletions` usage.

JamesWTruher: that doesn't work, because arrays aren't constant

daxian-dbw: OK, then we are good here. #Closed

daxian-dbw (approved):

daxian-dbw:
@dantraMSFT Can you please take another look and update your review?

daxian-dbw:
I'm getting this PR merged so that it can be validated by today's daily builds.
Actually, this PR should have the last commit marked with '[Feature]' tag to trigger full build run.

PR #5089, opened 2017-10-11
Add compiler switches replace dangerous function with safer ones.

### LINKED ISSUE #4975: Remove use for strlen from native code
Remove and or replace the use for strlen from native code

TravisEz13:
This should fix #4975


src/libpsl-native/CMakeLists.txt
```diff
@@ -3,7 +3,8 @@ project(PSL-NATIVE)
 
 # Can't use add_compile_options with 2.8.11
 set(CMAKE_BUILD_TYPE "Release")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Werror -fstack-protector-strong -fpie")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Werror -fstack-protector-strong -fpie -DFORTIFY_SOURCE=2")
```
SteveL-MSFT: Based on the documentation, FORTIFY_SOURCE=2 isn't enabled unless you also use `-O2` optimizations

SteveL-MSFT (changes requested):

src/libpsl-native/CMakeLists.txt
TravisEz13: Resolved.

src/libpsl-native/CMakeLists.txt
```diff
@@ -3,7 +3,16 @@ project(PSL-NATIVE)
 
 # Can't use add_compile_options with 2.8.11
 set(CMAKE_BUILD_TYPE "Release")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Werror -fstack-protector-strong -fpie")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Werror -fstack-protector-strong -fpie -DFORTIFY_SOURCE=2 -O2")
+
+if (${CMAKE_SYSTEM_NAME} MATCHES "Windows")
+    set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -Wl,-z,relro,-z,now")
```
daxian-dbw: Will `CMAKE_SYSTEM_NAME` ever be set to 'Windows'?
Can you please share some resources where I can find what each of those options is for?

daxian-dbw (changes requested):

src/libpsl-native/CMakeLists.txt
mirichmo: This component isn't built for Windows. The Windows flags should go in ...\src\powershell-native\coreclr_defs.cmake.

SteveL-MSFT: This Windows section should be removed

TravisEz13: Removed the windows case.

The docs are here:
https://cmake.org/cmake/help/v3.4/variable/CMAKE_HOST_SYSTEM_NAME.html#variable:CMAKE_HOST_SYSTEM_NAME
and 
https://cmake.org/cmake/help/v3.4/variable/CMAKE_SYSTEM_NAME.html

BTW, we use this when compiling for ARM, `set(CMAKE_SYSTEM_NAME Linux)`.

SteveL-MSFT (approved):

src/libpsl-native/CMakeLists.txt
daxian-dbw: @TravisEz13 Thanks. I'm actually interested in the options like `-Wl,-z,relro,-z,now`, and would like to know what they mean. I quickly search them bug didn't find anything very useful.

It will be great if you can provide some pointers. And better, put the link in `CMakeLists.txt` as comments.

mirichmo: What about the other aspect of my comment? Are there corresponding flags that we can or should use with MSBuild when compiling pwrshplugin.dll?

TravisEz13: I have another work item to run the scan that would detect these type of issues.  I just merged a change to the build to give me the data to run the scan.  I should have it done Monday.

TravisEz13:
@daxian-dbw Can you update your review?

daxian-dbw (approved):
It would be great if you add comments about some links that has information about the options used here, like `-Wl,-z,relro,-z,now`

TravisEz13:
I'll likely be making more changes.  I'll include the detail where the tools give me them.

Issue #1621, opened 2016-08-03
Importing modules by name is case sensitive


jdhitsolutions:
## Steps to reproduce

Start a PowerShell session.
Remove-Module PSReadline
Import-Module PSReadline
## Expected behavior

Expect to be able to re-import PSReadline. This works with other modules.
## Actual behavior

Error reporting module can't be found.

<img width="362" alt="psreadlineerror" src="https://cloud.githubusercontent.com/assets/12089920/17369373/17b028fe-5966-11e6-9c62-f8ef8cefec58.png">
## Environment data

``` powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      5.1.10032.0
PSEdition                      PowerShellCore
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   3.0.0.0
GitCommitId                    v6.0.0-alpha.7
CLRVersion
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

jpsnover:
I believe that as a general principle, PowerShell should be as case-preserving/case insensitive as possible while not breaking the underlying OS.

So in the case of the Unix File system manipulation and environmental variables - these directly affect the underlying OS and integration with other tools so we have to be case sensitive.

In the case of PowerShell variable names - there is no interaction between them and the underlying OS so we should be case insensitive.

So then issue with Module Names is that they are a tweener.  On the one hand there is a strong mapping to an underlying OS - a foldername.  On the other hand, theyy refer to a pure PowerShell concept - a PowerShell Module.   So you could argue it either way.

To me, the thing that pushes this decisively to the Case Insensitive solution is the effect the issue has on Ecosystem sharing.  If module names are case sensitive in Unix, then Tom might publish Module TEST and Mary might publish Module test.  But then if a Windows users tries to use both these modules, they are going to collide and they fail.   

I'm going to propose an acid test we need to apply:  Does case sensitivity break sharing artifacts between Windows and Linux?    If the answer is yes, then the prima facia answer should be that we should be case insensitive.

Ergo - Module names should be case insensitive. 

Does that sound right?

ForNeVeR:
[...]

PR #5097, opened 2017-10-12
Enable import-module to be case insensitive

SteveL-MSFT:
Apparently macOS is case insensitive

Fix https://github.com/PowerShell/PowerShell/issues/1621

### DIFF
--- a/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
@@ -309,7 +309,7 @@ namespace Microsoft.PowerShell.Commands
                             {
                                 qualifiedPath = Path.Combine(qualifiedPath, fileBaseName);
                             }
-                            else if (Directory.Exists(qualifiedPath))
+                            else if (Utils.NativeDirectoryExists(qualifiedPath))
                             {
                                 // if it points to a directory, add the basename back onto the path...
                                 qualifiedPath = Path.Combine(qualifiedPath, Path.GetFileName(fileBaseName));
--- a/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
@@ -296,14 +296,12 @@ namespace Microsoft.PowerShell.Commands
 #if UNIX
                 foreach (string folder in Directory.EnumerateDirectories(path))
                 {
-                    string fileName = Path.GetFileName(folder);
-                    if (String.Compare(fileName, name, StringComparison.OrdinalIgnoreCase) == 0)
+                    string moduleName = Path.GetFileName(folder);
+                    if (String.Compare(moduleName, fileBaseName, StringComparison.OrdinalIgnoreCase) == 0)
                     {
-                        string qualifiedPath = folder;
-                        name = fileName;
-#else
-                        string qualifiedPath = Path.Combine(path, fileBaseName);
+                        fileBaseName = moduleName;
 #endif
+                        string qualifiedPath = Path.Combine(path, fileBaseName);
                         module = LoadUsingMultiVersionModuleBase(qualifiedPath, manifestProcessingFlags, options, out found);
                         if (!found)
                         {
--- a/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
@@ -309,11 +309,7 @@ namespace Microsoft.PowerShell.Commands
                         {
                             if (name.IndexOfAny(Utils.Separators.Directory) == -1)
                             {
-#if UNIX
-                                qualifiedPath = Path.Combine(folder, fileName);
-#else
                                 qualifiedPath = Path.Combine(qualifiedPath, fileBaseName);
-#endif
                             }
                             else if (Directory.Exists(qualifiedPath))
                             {
--- a/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
@@ -293,67 +293,49 @@ namespace Microsoft.PowerShell.Commands
             // Now search using the module path...
             foreach (string path in modulePath)
             {
-                // a name takes the form of .../moduleDir/moduleName.<ext>
-                // if only one name has been specified, then the moduleDir
-                // and moduleName are identical and repeated...
-                // Also append th ename if the path currently points at a directory
-                string qualifiedPath = Path.Combine(path, fileBaseName);
-
-                // Load the latest valid version if it is a multi-version module directory
-                module = LoadUsingMultiVersionModuleBase(qualifiedPath, manifestProcessingFlags, options, out found);
-
-                if (!found)
+#if UNIX
+                foreach (string folder in Directory.EnumerateDirectories(path))
                 {
-                    if (name.IndexOfAny(Utils.Separators.Directory) == -1)
+                    string fileName = Path.GetFileName(folder);
+                    if (String.Compare(fileName, name, StringComparison.OrdinalIgnoreCase) == 0)
                     {
-                        qualifiedPath = Path.Combine(qualifiedPath, fileBaseName);
-                    }
-                    else if (Directory.Exists(qualifiedPath))
-                    {
-                        // if it points to a directory, add the basename back onto the path...
-                        qualifiedPath = Path.Combine(qualifiedPath, Path.GetFileName(fileBaseName));
-                    }
-
-                    module = LoadUsingExtensions(parentModule, name, qualifiedPath, extension, null, this.BasePrefix, ss, options, manifestProcessingFlags, out found);
-                }
-
-                if (found)
-                {
-                    break;
-                }
-            }
-
-#if UNIX    // case-sensitive filesystems are predominately used only on Unix
-            if (!found)
-            {
-                // try a case-insensitive search for the module folder
-                // based on PSGet, we can assume the module file casing matches the folder
-                foreach (string path in modulePath)
-                {
-                    foreach (string folder in Directory.EnumerateDirectories(path))
-                    {
-                        string fileName = Path.GetFileName(folder);
-                        if (String.Compare(fileName, name, StringComparison.OrdinalIgnoreCase) == 0)
+                        string qualifiedPath = folder;
+                        name = fileName;
+#else
+                        string qualifiedPath = Path.Combine(path, fileBaseName);
+#endif
+                        module = LoadUsingMultiVersionModuleBase(qualifiedPath, manifestProcessingFlags, options, out found);
+                        if (!found)
                         {
-                            module = LoadUsingMultiVersionModuleBase(folder, manifestProcessingFlags, options, out found);
-                            if (!found)
+                            if (name.IndexOfAny(Utils.Separators.Directory) == -1)
                             {
-                                string qualifiedPath = Path.Combine(folder, fileName);
-                                module = LoadUsingExtensions(parentModule, fileName, qualifiedPath, extension, null, this.BasePrefix, ss, options, manifestProcessingFlags, out found);
+#if UNIX
+                                qualifiedPath = Path.Combine(folder, fileName);
+#else
+                                qualifiedPath = Path.Combine(qualifiedPath, fileBaseName);
+#endif
                             }
+                            else if (Directory.Exists(qualifiedPath))
+                            {
+                                // if it points to a directory, add the basename back onto the path...
+                                qualifiedPath = Path.Combine(qualifiedPath, Path.GetFileName(fileBaseName));
+                            }
+
+                            module = LoadUsingExtensions(parentModule, name, qualifiedPath, extension, null, this.BasePrefix, ss, options, manifestProcessingFlags, out found);
                         }
-                        if (found)
-                        {
-                            break;
-                        }
+#if UNIX
                     }
                     if (found)
                     {
                         break;
                     }
                 }
-            }
+                if (found)
+                {
+                    break;
+                }
 #endif
+            }
 
             if (found)
             {
--- a/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
@@ -323,32 +323,23 @@ namespace Microsoft.PowerShell.Commands
                 }
             }
 
+#if UNIX    // case-sensitive filesystems are predominately used only on Unix
             if (!found)
             {
                 // try a case-insensitive search for the module folder
+                // based on PSGet, we can assume the module file casing matches the folder
                 foreach (string path in modulePath)
                 {
                     foreach (string folder in Directory.EnumerateDirectories(path))
                     {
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -323,6 +323,46 @@ internal List<WildcardPattern> MatchAll
                 }
             }
 
+            if (!found)
```
iSazonov: If it is only for Unix maybe use "#if UNIX"?

adityapatwardhan: The cyclomatic complexity of this function is pretty high. Can this be refactored into a smaller function?

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -124,3 +124,56 @@ namespace ModuleCmdlets
...
+    }
+
+    It "Import-Module can import a module using different casing using '<modulePath>' and manifest:<manifest>" -TestCases @(
+        @{modulePath="TESTMODULE/1.1"; manifest=$true},
+        @{modulePath="TESTMODULE/1.1"; manifest=$false},
+        @{modulePath="TESTMODULE"    ; manifest=$true},
+        @{modulePath="TESTMODULE"    ; manifest=$false}
+        ) {
+        param ($modulePath, $manifest)
+        New-Item -ItemType Directory -Path "$modulesPath/$modulePath" -Force
```
adityapatwardhan: Please redirect to null.
New-Item -ItemType Directory -Path "$modulesPath/$modulePath" -Force > $null

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -124,3 +124,56 @@ namespace ModuleCmdlets
...
+        if ($manifest) {
+            New-ModuleManifest -Path "$modulesPath/$modulePath/TESTMODULE.psd1" -RootModule "TESTMODULE.psm1" -ModuleVersion 1.1
+        }
+        Set-Content -Path "$modulesPath/$modulePath/TESTMODULE.psm1" -Value "function mytest { 'hello' }"
+        Import-Module testMODULE
```
adityapatwardhan: Maybe `-Force` to ensure you are not loading a cached version.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -124,3 +124,56 @@ namespace ModuleCmdlets
...
+        $m = Get-Module TESTmodule
+        $m | Should BeOfType "System.Management.Automation.PSModuleInfo"
+        $m.Name | Should BeExactly "TESTMODULE"
+        mytest | Should BeExactly "hello"
+        Remove-Module TestModule
+        Get-Module tESTmODULE | Should BeNullOrEmpty
+    }
+
+    It "Import-Module will import exact casing if available" {
+        New-Item -ItemType Directory -Path "$modulesPath\Test" -Force
```
adityapatwardhan: `> $null`

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -124,3 +124,56 @@ namespace ModuleCmdlets
...
+        Set-Content -Path "$modulesPath\Test\Test.psm1" -Value "function casetest { 'first' }"
+        New-Item -ItemType Directory -Path "$modulesPath\tEST" -Force
```
adityapatwardhan: `> $null`

adityapatwardhan (changes requested):

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
SteveL-MSFT: it's technically any filesystem that's case sensitive and technically (although unlikely), it seems you can make NTFS case sensitive.  for practical cases, it's probably fine to #ifdef this and if a customer has an issue on Windows, we can revisit

SteveL-MSFT: I could simplify it a bit if we can make the assumption that the psd1 or psm1 file always has the same casing as the folder name.  It's not clear to me if that is a safe assumption or not.  The current code is conservative in that you could have: "...\Modules\FOO\foo.psd1"

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
SteveL-MSFT: will add

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
SteveL-MSFT: will add

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
SteveL-MSFT: will add

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
SteveL-MSFT: will add

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
SteveL-MSFT: actually, as part of this test, I'm removing the module with a different casing, so I don't want to -force loading

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
SteveL-MSFT: I'll check with the PSGet folks

SteveL-MSFT: per PSGet, it is safe to assume the folder and file names have the same casing.  So I'll simplify this code.

SteveL-MSFT: fixed and added comment

SteveL-MSFT:
@adityapatwardhan feedback addressed

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -323,6 +323,38 @@ internal List<WildcardPattern> MatchAll
                 }
             }
 
+#if UNIX    // case-sensitive filesystems are predominately used only on Unix
+            if (!found)
+            {
+                // try a case-insensitive search for the module folder
+                // based on PSGet, we can assume the module file casing matches the folder
+                foreach (string path in modulePath)
```
iSazonov: In line 294 the same cycle? Can we remove it and replace with the new one on all platforms?

SteveL-MSFT: We still need both as the first one will find an exact match and fall through to this if it doesn't find it.  Also the above loop also handles a file path to the module which I think should be case sensitive since it's a literal file path.  The lower code only handles the module name which we can treat as case insensitive.  The code is similar, but I don't think it's worth combining which will make it more complex.

iSazonov: Why we should use different logic for module folder names and module file names?

SteveL-MSFT: If someone uses: `Import-module /users/steve/dev/MyModule/MyModule.psd1` then that should be treated as a case-sensitive path.  If they use `Import-module mymodule`, then that's case-insensitive as that's a module name and not a filepath.

iSazonov: Ah, I meant folder and file names not paths.

SteveL-MSFT: I don't see an easy way to have a single loop between the two without incurring a cost of enumerating the folders for the purpose of case-insensitive comparison in the "normal" case where either the case matches or the filesystem is case-insensitive.  If there's two modules "Foo" and "foo", and the user does `import-module foo` we want the first loop to match `foo` and not `Foo` as that is an exact match.

iSazonov: We have the same problem in #4699 - can you review the idea to use here?

SteveL-MSFT: In the tab completion case, I think you have to cover more cases as `ab` might match `ABc`, `abc`, or `aBcdefg`, etc...  so creating a case-insenstive match pattern makes sense.  In this case, we have a specific name (wildcard is not accepted) and just need a case-insensitive match if we haven't already matched the exact casing (which is preferred).  And we don't need to return all matches, the first one wins.

If we want just the first match to win even if there is an exact match, it would be easy to merge the two loops.  However, I think we want to get an exact match if it exists, then fallback on case-sensitive search.  For example, if the user typed `ipmo AAA` which exists in the last modulepath, but `aaa` exists in the first, I would have to enumerate all module folder names in every modulepath part every time just to make sure `AAA` wasn't found after my first case-insensitive match.

iSazonov: We could have solved this situation on the other hand. Do we want _all_ the modules to be ported? If so, we must document this and exclude the case-sensitivity in this code.

SteveL-MSFT: That's a good point on portable scripts being able to rely on module names being case insensitive.  It would make sense to just say module names are case insensitive everywhere and have it enforced by install-module and publish-module.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -293,34 +293,48 @@ internal List<WildcardPattern> MatchAll
...
+                        module = LoadUsingMultiVersionModuleBase(qualifiedPath, manifestProcessingFlags, options, out found);
+                        if (!found)
+                        {
+                            if (name.IndexOfAny(Utils.Separators.Directory) == -1)
+                            {
+#if UNIX
+                                qualifiedPath = Path.Combine(folder, fileName);
+#else
+                                qualifiedPath = Path.Combine(qualifiedPath, fileBaseName);
+#endif
```
iSazonov: Perhaps we could exclude `#if` here and leave one line.

SteveL-MSFT: I think this line isn't needed as I have qualifiedPath = folder above, was probably thinking this, but didn't put it together.  Will fix.

iSazonov:
Should we add negative tests too (to reflect the situations we have decided to exclude)?  What errors should appear?
Also why we skip Windows in new test? If we want to have all modules portable all test should run on all platforms identically.

SteveL-MSFT:
@iSazonov test was originally skipped due to the previous code for an exact match since it wasn't possible to create two modules that differ only by casing on Windows.  With the new change, no reason not to run it on Windows.  I also noticed a bug in the test where I wasn't removing the test module between runs.

iSazonov:
I don't see still that tests reflect to "StringComparison.OrdinalIgnoreCase".

SteveL-MSFT:
@iSazonov the new test creates a module called TESTMODULE and imports it using testMODULE, this is where the case-insensitive comparison occurs.  In the UNIX code path, if it enumerates the module folders in the module path and sees if it matches the name of the given module.  Not clear to me if you were expecting a different test?

iSazonov:
@SteveL-MSFT My thoughts was that we should do `Import-Module testMODULE` and `Import-Module TESTMODULE` and check that we load the same module. Although I'm not sure it's necessary.

SteveL-MSFT:
@iSazonov I don't think that's necessary

iSazonov (approved):

adityapatwardhan (approved):
LGTM

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -293,34 +293,42 @@ internal List<WildcardPattern> MatchAll
...
+                        string qualifiedPath = Path.Combine(path, fileBaseName);
+                        module = LoadUsingMultiVersionModuleBase(qualifiedPath, manifestProcessingFlags, options, out found);
+                        if (!found)
+                        {
+                            if (name.IndexOfAny(Utils.Separators.Directory) == -1)
+                            {
+                                qualifiedPath = Path.Combine(qualifiedPath, fileBaseName);
+                            }
+                            else if (Directory.Exists(qualifiedPath))
```
adityapatwardhan: Please use Utils.NativeDirectoryExists()
https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/Utils.cs#L912

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -293,34 +293,42 @@ internal List<WildcardPattern> MatchAll
...
-                // and moduleName are identical and repeated...
-                // Also append th ename if the path currently points at a directory
-                string qualifiedPath = Path.Combine(path, fileBaseName);
-
-                // Load the latest valid version if it is a multi-version module directory
-                module = LoadUsingMultiVersionModuleBase(qualifiedPath, manifestProcessingFlags, options, out found);
-
-                if (!found)
+#if UNIX
+                foreach (string folder in Directory.EnumerateDirectories(path))
```
adityapatwardhan: Please use Utils.NativeEnumerateDirectory()
https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/Utils.cs#L926

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
adityapatwardhan: Ok

adityapatwardhan (changes requested):

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
SteveL-MSFT: Will change

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
SteveL-MSFT: Will change

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
SteveL-MSFT: This method appears to be only implemented for Win32.  Since this path is only for UNIX, I'll leave it as-is.

adityapatwardhan: ok

adityapatwardhan (approved):

Issue #4214, opened 2017-07-11
Create pwsh as a shorter name for calling powershell.exe


dragonwolf83:
The muscle memory from switching cmd to powershell is a pain. It is made worse by the fact that it is much longer than typing out cmd or bash. Even with Windows Search in Start, it often picks up the ISE if you have been using it awhile, so typing out the full name is often needed.

A better name would be posh.exe or other valid extension. It is as long as bash and just one character longer than cmd. Posh has been the accepted abbreviation in the community for all things PowerShell so there is some familiarity there. I don't believe there are any conflicts with other apps that this would cause, but that would need some digging.

This could be achieved in two ways.
1. Create a an official shortcut or symbolic link named posh that would reference powershell.exe
2. Rename powershell.exe to posh.exe.

The former maintains backwards compatibility and allows the user to choose between posh or powershell.exe. Existing scripts not impacted. 

The latter could help give more freedom to issues like  #4199 and other POSIX proposals to how the shell is started. However, it is a breaking change and would need to be reviewed for the impact and the branding implication. 

I think making an official posh shortcut is the best option with minimal impact. New documentation could give preference to posh as the way to enter PowerShell.

joeyaiello:
I would like a short alias too, but [this](https://www.commandlinux.com/man-page/man1/posh.1.html) is why we didn't get go with `posh` when we first went out (we talked about it at length). [Appears](https://www.howtoinstall.co/en/ubuntu/xenial/posh) to be still an up-to-date package too. 

Open to other suggestions, but most of the obvious *sh ones have been taken. After the only wget/curl alias thing, I'd really like to avoid colliding with other packages out there, even if they're less common.

rkeithhill:
I'm not too crazy about this whole idea but how about `pwsh`?

RichardSiddaway:
OK - forget that - I just realised ps is an alias for get-process. that would be too confusing

joeyaiello:
I'm getting strong deja vu right now. 😉

rkeithhill:
There's a part of me that appreciates PowerShell's readability (when not using aliases) so running `powershell` works fine for me.

dragonwolf83:
`ps` is an existing command on Linux as well. 

Some other ideas are: `pscmd`, `psc`, `monad`, `psh`, `pshell`

There is definitely diminishing returns with these names though on whether they really save anything or just add unnecessary confusion if they are even available.

iSazonov:
[...]

Issue #1744, opened 2016-08-12
Further changes/improvement needed to module path calculation


daxian-dbw:
While working on the sxs module path, I found some method names are intuitive, for example, `GetDscModulePath()` actually returns the `Program Files` shared module path, `GetSystemWideModulePath()` actually returns the pshome module path. Those method need to be renamed.

- [x] Refactor method names related to module path calculation
- [x] The code that detects if a module path is PSHome module path derived from parent process needs to be smarter, not only check `powershell` file in the parent folder, but maybe check `powershell.deps.json` file as well, and also the short folder name of the path should be `Modules`. The current detection logic casued a test to fail in Travis CI, see #1792
- [x] Add more SxS module path tests
- [ ] ~~Think about the name `inbox` of `Platform.IsInbox` property. Do we need a better name?~~ `Platform.IsInbox` is an internal property, and it's well commented to clarify its meaning. We will keep this name unless there is a big concern about it.

daxian-dbw:
The issue is resolved by #2265

PR #5101, opened 2017-10-12
Rename powershell.exe to pwsh.exe

SteveL-MSFT:
renamed powershell.exe to pwsh.exe
updated powershell core code dependencies, console host references, and tests
updated vscode json files
updated docker files
updated logic in moduleintrinsics.cs to account for SxS where you are starting pwsh from powershell.exe

Fix https://github.com/PowerShell/PowerShell/issues/4214

### DIFF
--- a/assets/Product.wxs
+++ b/assets/Product.wxs
@@ -113,7 +113,7 @@
             </Component>
             <!-- add ourselvs to %PATH% so pwsh.exe can be started from Windows PowerShell or cmd.exe -->
             <Component Id="SetPath" Guid="{9dbb7763-7baf-48e7-b025-3bdedcb0632f}">
-              <Environment Id="PATH" Action="set" Name="PATH" Part="last" Permanent="yes" System="yes" Value="[$(var.ProductVersionWithName)]"/>
+              <Environment Id="PATH" Action="set" Name="PATH" Part="last" Permanent="no" System="yes" Value="[$(var.ProductVersionWithName)]"/>
             </Component>
           </Directory>
         </Directory>
--- a/assets/Product.wxs
+++ b/assets/Product.wxs
@@ -31,7 +31,7 @@
     <WixVariable Id="WixUIInfoIco" Value="assets\WixUIInfoIco.bmp" />
     <Property Id="WIXUI_EXITDIALOGOPTIONALCHECKBOXTEXT" Value="Open $(env.ProductName)" />
     <!-- Default value of Checkbox of starting PowerShell after installation -->
-    <Property Id="WixShellExecTarget" Value="[$(var.ProductVersionWithName)]PowerShell.exe"/>
+    <Property Id="WixShellExecTarget" Value="[$(var.ProductVersionWithName)]pwsh.exe"/>
     <CustomAction Id="LaunchApplication" BinaryKey="WixCA" DllEntry="WixShellExec" Impersonate="yes" />
     <UI>
       <Dialog Id="MyExitDialog" Width="370" Height="270" Title="!(loc.ExitDialog_Title)">
@@ -73,6 +73,8 @@
       <ComponentGroupRef Id="$(var.ProductVersionWithName)"/>
       <ComponentRef Id="ProductVersionFolder"/>
       <ComponentRef Id="ApplicationProgramsMenuShortcut"/>
+      <ComponentRef Id="RegistryEntries"/>
+      <ComponentRef Id="SetPath"/>
     </Feature>
     <!-- We need to show EULA, and provide option to customize download location -->
     <Property Id="WIXUI_INSTALLDIR" Value="INSTALLFOLDER" />
@@ -103,13 +105,23 @@
             <Component Id="ProductVersionFolder" Guid="{e1a7f05e-0cd6-4227-80a8-e4fb311f045c}">
               <CreateFolder/>
             </Component>
+            <!-- register ourselves in application registry so can be started using just Win+R `pwsh.exe` -->
+            <Component Id="RegistryEntries" Guid="{402e52f7-baf8-489d-af21-f756a6ca3530}">
+                <RegistryKey Root="HKCU" Key="Software\Microsoft\Windows\CurrentVersion\App Paths\pwsh.exe" ForceCreateOnInstall="yes" ForceDeleteOnUninstall="yes">
+                    <RegistryValue Type="string" Value="[$(var.ProductVersionWithName)]pwsh.exe"/>
+                </RegistryKey>
+            </Component>
+            <!-- add ourselvs to %PATH% so pwsh.exe can be started from Windows PowerShell or cmd.exe -->
+            <Component Id="SetPath" Guid="{9dbb7763-7baf-48e7-b025-3bdedcb0632f}">
+              <Environment Id="PATH" Action="set" Name="PATH" Part="last" Permanent="yes" System="yes" Value="[$(var.ProductVersionWithName)]"/>
+            </Component>
           </Directory>
         </Directory>
       </Directory>
       <Directory Id="ProgramMenuFolder">
         <Directory Id="ApplicationProgramsFolder" Name="$(var.ProductSemanticVersionWithName)">
           <Component Id="ApplicationProgramsMenuShortcut" Guid="{A77507A7-F970-4618-AC30-20AFE36EE2EB}">
-            <Shortcut Id="PowerShell_ProgramsMenuShortcut" Name="$(var.ProductSemanticVersionWithName)" Description="$(var.ProductSemanticVersionWithName)" Target="[$(var.ProductVersionWithName)]PowerShell.exe" WorkingDirectory="$(var.ProductVersionWithName)"
+            <Shortcut Id="PowerShell_ProgramsMenuShortcut" Name="$(var.ProductSemanticVersionWithName)" Description="$(var.ProductSemanticVersionWithName)" Target="[$(var.ProductVersionWithName)]pwsh.exe" WorkingDirectory="$(var.ProductVersionWithName)"
               Icon = "PowerShellExe.ico" />
             <RemoveFolder Id="ApplicationProgramsFolder" On="uninstall"/>
             <RegistryValue Root="HKCU" Key="Software\Microsoft\$(var.ProductSemanticVersionWithName)\ProgramsMenuShortcut" Name="installed" Type="integer" Value="1" KeyPath="yes"/>
--- a/demos/SSHRemoting/README.md
+++ b/demos/SSHRemoting/README.md
@@ -46,9 +46,9 @@ In addition you will need to enable password authentication and optionally key b
     ```none
     PasswordAuthentication yes
     ```
-    - Add a PowerShell subsystem entry, replace `c:/program files/powershell/6.0.0/powershell.exe` with the correct path to the version you want to use
+    - Add a PowerShell subsystem entry, replace `c:/program files/powershell/6.0.0/pwsh.exe` with the correct path to the version you want to use
     ```none
-    Subsystem    powershell c:/program files/powershell/6.0.0/powershell.exe -sshs -NoLogo -NoProfile
+    Subsystem    powershell c:/program files/powershell/6.0.0/pwsh.exe -sshs -NoLogo -NoProfile
     ```
     - Optionally enable key authentication
     ```none
@@ -173,7 +173,7 @@ Microsoft Windows [Version 10.0.10586]
 #
 # Windows to Windows
 #
-C:\Users\PSUser\Documents>"C:\Program Files\PowerShell\6.0.0.17\powershell.exe"
+C:\Users\PSUser\Documents>pwsh.exe
 PowerShell
 Copyright (C) Microsoft Corporation. All rights reserved.
 
--- a/src/Microsoft.PowerShell.ScheduledJob/ScheduledJobDefinition.cs
+++ b/src/Microsoft.PowerShell.ScheduledJob/ScheduledJobDefinition.cs
@@ -42,7 +42,7 @@ namespace Microsoft.PowerShell.ScheduledJob
         private bool _isDisposed;
 
         // Task Action strings.
-        private const string TaskExecutionPath = @"powershell.exe";
+        private const string TaskExecutionPath = @"pwsh.exe";
         private const string TaskArguments = @"-NoLogo -NonInteractive -WindowStyle Hidden -Command ""Import-Module PSScheduledJob; $jobDef = [Microsoft.PowerShell.ScheduledJob.ScheduledJobDefinition]::LoadFromStore('{0}', '{1}'); $jobDef.Run()""";
         private static object LockObject = new object();
         private static int CurrentId = 0;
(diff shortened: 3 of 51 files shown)


test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -418,7 +418,7 @@ Describe "TabCompletion" -Tags CI {
                 @{ inputStr = 'gmo -list PackageM'; expected = 'PackageManagement'; setup = $null }
                 @{ inputStr = 'gcm -Module PackageManagement Find-Pac'; expected = 'Find-Package'; setup = $null }
                 @{ inputStr = 'ipmo PackageM'; expected = 'PackageManagement'; setup = $null }
-                @{ inputStr = 'Get-Process powersh'; expected = 'powershell'; setup = $null }
+                @{ inputStr = 'Get-Process pws'; expected = 'pwsh'; setup = $null }
```
daxian-dbw: A typo -- `pws`

daxian-dbw (changes requested):

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
SteveL-MSFT: This is for the tab completion test so `pws` completes to `pwsh` since there isn't necessarily a `powershell` process anymore and `pw` might resolve to something else

demos/SSHRemoting/README.md
```diff
@@ -173,7 +173,7 @@ Microsoft Windows [Version 10.0.10586]
 #
 # Windows to Windows
 #
-C:\Users\PSUser\Documents>"C:\Program Files\PowerShell\6.0.0.17\powershell.exe"
+C:\Users\PSUser\Documents>pwsh.exe
```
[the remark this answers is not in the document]
SteveL-MSFT: We should update the msi installer to add it to %PATH%, I wanted to remove it since it references a specific version path.  I'll do the msi update separate from this PR.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
daxian-dbw: Ah, yes, my mistake.

daxian-dbw (approved):

SteveL-MSFT:
rebased to fix merge conflict

src/Microsoft.PowerShell.ScheduledJob/ScheduledJobDefinition.cs
```diff
@@ -42,7 +42,7 @@ public sealed class ScheduledJobDefinition : ISerializable, IDisposable
         private bool _isDisposed;
 
         // Task Action strings.
-        private const string TaskExecutionPath = @"powershell.exe";
+        private const string TaskExecutionPath = @"pwsh.exe";
```
PaulHigin: This code assumes a discoverable pwsh.exe native command.  Does this fix include adding pwsh.exe to the system path environment variable?  If so what version?  I wonder if we should allow the user to select a powershell.exe or pwsh.exe path to use for scheduled jobs (maybe in a separate issue/pr).

daxian-dbw (approved):

src/Microsoft.PowerShell.ScheduledJob/ScheduledJobDefinition.cs
PaulHigin: Maybe we should use Utils.DefaultPowerShellAppBase here as we do with PowerShellProcessInstance.cs to ensure the scheduled job is using the correct path.
The problem with scheduled jobs is that they can run indefinitely on a machine and if pwsh.exe is uninstalled or installed to a new location then they will break.  Of course we have a similar problem with credentials expiring.

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -3166,7 +3166,7 @@ private void CreateContainerProcessInternal()
                     // Windows Server container (i.e., RuntimeId is empty) uses named pipe transport for now.
                     //
                     cmd = string.Format(System.Globalization.CultureInfo.InvariantCulture,
-                        @"{{""CommandLine"": ""powershell.exe {0} -NoLogo {1}"",""RestrictedToken"": {2}}}",
+                        @"{{""CommandLine"": ""pwsh.exe {0} -NoLogo {1}"",""RestrictedToken"": {2}}}",
```
PaulHigin: Same issue here in that we rely on a "built-in" pwsh.exe command that is discoverable.  But we have the same issue with ssh.exe used for SSH based remoting.

PaulHigin: Actually as I think about it, we don't want to change this here.  This is a command run on a target VM and relies on the built-in Windows PowerShell.  pwsh.exe will not be available on a new VM.

PaulHigin (changes requested):

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
SteveL-MSFT: So this command is run within the VM expecting Windows PowerShell?

PaulHigin: Yes

src/Microsoft.PowerShell.ScheduledJob/ScheduledJobDefinition.cs
SteveL-MSFT: I have a separate PR coming for the msi installer to append to %PATH% which would also solve the Win+R->pwsh usage

If PSCore6 is uninstalled, the full path won't help either.

SteveL-MSFT: Actually, I'll do the %PATH% msi change as part of this PR as it's more useful that pwsh.exe differs form powershell.exe and I can put it at the end of the path rather than taking over the front

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
SteveL-MSFT: Ok, will change back.

PaulHigin (approved):
LGTM

SteveL-MSFT:
@daxian-dbw can you review again since last commit?

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
```diff
@@ -791,8 +793,9 @@ private static string RemoveSxSPsHomeModulePath(string currentProcessModulePath,
                 {
                     string parentDir = Path.GetDirectoryName(trimedPath);
                     string psExePath = Path.Combine(parentDir, powershellExeName);
+                    string oldExePath = Path.Combine(parentDir, oldPowershellExeName);
                     string psDepsPath = Path.Combine(parentDir, "powershell.deps.json");
```
daxian-dbw: Once you change it to `pwsh` in csproj, I guess the .deps.json file will be changed to `pwsh.deps.json`.
But maybe you want to keep the check for `powershell.deps.json` for now just like the check for old "powershell.exe".

assets/Product.wxs
```diff
@@ -103,13 +105,23 @@
...
             </Component>
+            <!-- register ourselves in application registry so can be started using just Win+R `pwsh.exe` -->
+            <Component Id="RegistryEntries" Guid="{402e52f7-baf8-489d-af21-f756a6ca3530}">
+                <RegistryKey Root="HKCU" Key="Software\Microsoft\Windows\CurrentVersion\App Paths\pwsh.exe" ForceCreateOnInstall="yes" ForceDeleteOnUninstall="yes">
+                    <RegistryValue Type="string" Value="[$(var.ProductVersionWithName)]pwsh.exe"/>
+                </RegistryKey>
+            </Component>
+            <!-- add ourselvs to %PATH% so pwsh.exe can be started from Windows PowerShell or cmd.exe -->
+            <Component Id="SetPath" Guid="{9dbb7763-7baf-48e7-b025-3bdedcb0632f}">
+              <Environment Id="PATH" Action="set" Name="PATH" Part="last" Permanent="yes" System="yes" Value="[$(var.ProductVersionWithName)]"/>
```
daxian-dbw: Do we remove the path when uninstalling the package?

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
```diff
@@ -769,8 +769,10 @@ private static string RemoveSxSPsHomeModulePath(string currentProcessModulePath,
         {
 #if UNIX
             const string powershellExeName = "pwsh";
+            const string oldPowershellExeName = "powershell";
 #else
             const string powershellExeName = "pwsh.exe";
+            const string oldPowershellExeName = "powershell.exe";
```
daxian-dbw: Shall we remove the old exe name when we reach RC/RTM?

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -3166,7 +3166,7 @@ private void CreateContainerProcessInternal()
                     // Windows Server container (i.e., RuntimeId is empty) uses named pipe transport for now.
                     //
                     cmd = string.Format(System.Globalization.CultureInfo.InvariantCulture,
-                        @"{{""CommandLine"": ""pwsh.exe {0} -NoLogo {1}"",""RestrictedToken"": {2}}}",
+                        @"{{""CommandLine"": ""powershell.exe {0} -NoLogo {1}"",""RestrictedToken"": {2}}}",
```
daxian-dbw: A comment explaining why `powershell.exe` is used here will be helpful when looking at this code in future.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
SteveL-MSFT: I think this is still needed if we start pwsh from Windows PowerShell

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
SteveL-MSFT: Yeah, it did get changed to `pwsh.deps.json`, will add `oldDepsPath`

SteveL-MSFT: Do we really need the deps.json check?

daxian-dbw: I remember we didn't have the .deps.json check initially and it caused a build break and the fix was to add this deps.json check. Here is the related issue #1744 and #1792

Not sure if that requirement is applicable now.

daxian-dbw:
@SteveL-MSFT One more issue: when creating Unix packages, symbolic link is created to point to `powershell` in `$pshome` (see the following places). You need to update the symbolic link file name and the target file name.

Besides, are you planning to change the package name? (I assume not).

https://github.com/PowerShell/PowerShell/blob/master/tools/packaging/packaging.psm1#L459
https://github.com/PowerShell/PowerShell/blob/master/tools/packaging/packaging.psm1#L471
https://github.com/PowerShell/PowerShell/blob/master/tools/packaging/packaging.psm1#L587

SteveL-MSFT:
@daxian-dbw will update the symlink, the package name is not changing

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
SteveL-MSFT: I'll keep it for now

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
daxian-dbw: I think the "start pwsh from windows powershell" case is handled in [`NeedToClearProcessModulePath`](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs#L725), and `RemoveSxSPsHomeModulePath` only handles powershell core.

SteveL-MSFT: Because of the .deps.json file check, this doesn't apply to Windows PowerShell anyways.  We can remove at RC/GA

daxian-dbw: Then I will open an issue on this.

assets/Product.wxs
SteveL-MSFT: Will fix

test/docker/networktest/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # escape=`
 FROM microsoft/windowsservercore
 
-SHELL ["powershell.exe","-command"]
+SHELL ["pwsh.exe","-command"]
```
TravisEz13: Please don't break docker... there are tests in `docker/tests`. All tests can be can be run from windows, just run them once with the docker daemon set to windows mode and once set to Linux mode.

TravisEz13: This is before PSCore has been installed.  So this won't work.

TravisEz13: Sorry, just notice this is a test file, not a production docker file.

daxian-dbw (changes requested):
I did some more updates to the packaging.psm1 regarding the name change.

I think we also need to update the docker files, including those in `docker/release` and `tools\releaseBuild\Images`. But I'm OK if you want to do that in a separate PR.

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
```diff
@@ -791,8 +796,10 @@ private static string RemoveSxSPsHomeModulePath(string currentProcessModulePath,
                 {
                     string parentDir = Path.GetDirectoryName(trimedPath);
                     string psExePath = Path.Combine(parentDir, powershellExeName);
-                    string psDepsPath = Path.Combine(parentDir, "powershell.deps.json");
-                    if (File.Exists(psExePath) && File.Exists(psDepsPath))
+                    string oldExePath = Path.Combine(parentDir, oldPowershellExeName);
+                    string psDepsPath = Path.Combine(parentDir, powershellDepsName);
+                    string oldDepsPath = Path.Combine(parentDir, oldPowershellDepsName);
+                    if ((File.Exists(psExePath) && File.Exists(psDepsPath)) || (File.Exists(oldDepsPath) && File.Exists(oldDepsPath)))
```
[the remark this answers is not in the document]
SteveL-MSFT: Will fix

SteveL-MSFT:
Let's do the docker updates as separate PR

daxian-dbw (approved):

tmknight:
Shouldn't this be an alias vs a complete rename?  Please point me to the discussion where the decision to rename took place.  Thanks

markekraus:
@tmknight here is where the PowerShell comitee's decision was announced: https://github.com/PowerShell/PowerShell/issues/4214#issuecomment-335989245

PR #5106, opened 2017-10-13
Use InvariantCulture for implicit argument conversion

### LINKED ISSUE #3348: PowerShell use different logic, when converting passed value to parameter type, for compiled and script cmdlets
Steps to reproduce
------------------
```powershell
PS> Add-Type -TypeDefinition @'
     using System;
     [Cmdlet("Test", "Compiled")]
     public class TestCompiled : PSCmdlet {
         [Parameter(Position = 0)]
         public DateTime DT;
         protected override void EndProcessing() {
             WriteObject(DT);
         }
     }
'@ -PassThru | Select-Object -First 1 -ExpandProperty Assembly | Import-Module
PS> function Test-Script {
     param(
         [DateTime]$DT
     )
     $DT
 }
PS> @(
     [CultureInfo]::CurrentCulture = 'ru-RU'
     Test-Compiled 1/11/1111
     Test-Script   1/11/1111
     [CultureInfo]::CurrentCulture = 'en-US'
 )

Wednesday, November 1, 1111 12:00:00 AM
Wednesday, January 11, 1111 12:00:00 AM
```

Expected behavior
-----------------
Same dates in both cases.

Actual behavior
---------------
Strings get parsed with different culture in mind.

Environment data
----------------
```powershell
> $PSVersionTable

GitCommitId                    v6.0.0-alpha.17-18-gedecf7e04d65f7183cd1d278daa5ebed2970813f
CLRVersion
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0, 5.0, 5.1.10032.0, 6.0.0}
```

My thoughts about the problem
----------------
I think the problem happens due to PowerShell automatically adds `ArgumentTypeConverterAttribute` to typed parameters for script cmdlets:
```powershell
PS> (Get-Command Test-Script).Parameters.DT.Attributes | % GetType | % FullName
System.Management.Automation.ParameterAttribute
System.Management.Automation.ArgumentTypeConverterAttribute
```
but not doing the same for compiled cmdlets:
```powershell
PS> (Get-Command Test-Compiled).Parameters.DT.Attributes | % GetType | % FullName
System.Management.Automation.ParameterAttribute
```
That attribute make conversion use `InvariantCulture`, while compiled cmdlets use `CurrentCulture` for conversion.

daxian-dbw:
Fix #3348

PowerShell uses `InvariantCulture` for explicit type conversion. The implicit argument conversion for the script function/cmdlet also uses `InvariantCulture`. However, the implicit parameter argument type conversion for the C# cmdlet uses `CurrentCulture`. Change it to use 'InvariantCulture' too.

### DIFF
--- a/test/powershell/Language/Parser/Conversions.Tests.ps1
+++ b/test/powershell/Language/Parser/Conversions.Tests.ps1
@@ -21,6 +21,19 @@ Describe 'conversion syntax' -Tags "CI" {
         ([System.Management.Automation.ActionPreference]"Stop").ToString() | Should Be "Stop"
     }
 
+    It "language conversion uses 'InvariantCulture' by default" {
+        $dateString = "1/11/1111"
+        $expected = [System.Management.Automation.LanguagePrimitives]::ConvertTo($dateString, [datetime], [cultureinfo]::InvariantCulture)
+        $result = try {
+            $oldCulture = [CultureInfo]::CurrentCulture
+            [CultureInfo]::CurrentCulture = 'ru-RU'
+            [datetime] $dateString
+        } finally {
+            [CultureInfo]::CurrentCulture = $oldCulture
+        }
+        $result | Should Be $expected
+    }
+
     Context "Cast object[] to more narrow generic collection" {
         BeforeAll {
             $testCases1 = @(
--- a/test/powershell/Language/Parser/Conversions.Tests.ps1
+++ b/test/powershell/Language/Parser/Conversions.Tests.ps1
@@ -24,8 +24,8 @@ Describe 'conversion syntax' -Tags "CI" {
     It "language conversion uses 'InvariantCulture' by default" {
         $dateString = "1/11/1111"
         $expected = [System.Management.Automation.LanguagePrimitives]::ConvertTo($dateString, [datetime], [cultureinfo]::InvariantCulture)
+        $oldCulture = [CultureInfo]::CurrentCulture
         $result = try {
-            $oldCulture = [CultureInfo]::CurrentCulture
             [CultureInfo]::CurrentCulture = 'ru-RU'
             [datetime] $dateString
         } finally {
--- a/test/powershell/engine/ParameterBinding/ParameterBinding.Tests.ps1
+++ b/test/powershell/engine/ParameterBinding/ParameterBinding.Tests.ps1
@@ -301,9 +301,9 @@
         $ruCastResult = [System.Management.Automation.LanguagePrimitives]::ConvertTo("1/11/1111", [datetime], $ruCulture)
 
         # Implict parameter argument conversion should also use 'InvariantCulture'
+        $oldCulture = [CultureInfo]::CurrentCulture
         $csharpCmdlet, $scriptFunction, $scriptCmdlet = try {
-            $oldCulture = [CultureInfo]::CurrentCulture
-            [CultureInfo]::CurrentCulture = 'ru-RU'
+            [CultureInfo]::CurrentCulture = $ruCulture
             Get-Date -Date 1/11/1111
             Test-Function  1/11/1111
             Test-ScriptCmdlet 1/11/1111
--- a/test/powershell/engine/ParameterBinding/ParameterBinding.Tests.ps1
+++ b/test/powershell/engine/ParameterBinding/ParameterBinding.Tests.ps1
@@ -293,20 +293,23 @@
             $date
         }
 
-        # Explict type cast use 'InvariantCulture' by default
-        $expected = [datetime] "1/11/1111"
+        # Explict type cast using 'InvariantCulture'
+        $expected = [System.Management.Automation.LanguagePrimitives]::ConvertTo("1/11/1111", [datetime], [cultureinfo]::InvariantCulture)
 
         # Cast result with ru-RU culture
         $ruCulture = [cultureinfo]::GetCultureInfo("ru-RU")
         $ruCastResult = [System.Management.Automation.LanguagePrimitives]::ConvertTo("1/11/1111", [datetime], $ruCulture)
 
         # Implict parameter argument conversion should also use 'InvariantCulture'
-        $csharpCmdlet, $scriptFunction, $scriptCmdlet = @(
+        $csharpCmdlet, $scriptFunction, $scriptCmdlet = try {
+            $oldCulture = [CultureInfo]::CurrentCulture
             [CultureInfo]::CurrentCulture = 'ru-RU'
             Get-Date -Date 1/11/1111
             Test-Function  1/11/1111
             Test-ScriptCmdlet 1/11/1111
-        )
+        } finally {
+            [CultureInfo]::CurrentCulture = $oldCulture
+        }
 
         $expected | Should Not Be $ruCastResult
         $csharpCmdlet | Should Be $expected
--- a/test/powershell/engine/ParameterBinding/ParameterBinding.Tests.ps1
+++ b/test/powershell/engine/ParameterBinding/ParameterBinding.Tests.ps1
@@ -281,6 +281,39 @@
         }
     }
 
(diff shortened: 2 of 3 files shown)


test/powershell/engine/ParameterBinding/ParameterBinding.Tests.ps1
```diff
@@ -281,6 +281,39 @@
...
+        function Test-Function {
+            param([DateTime] $date)
+            $date
+        }
+
+        function Test-ScriptCmdlet {
+            [CmdletBinding()]
+            param([DateTime] $date)
+            $date
+        }
```
iSazonov: Should we add C# cmdlet?

test/powershell/engine/ParameterBinding/ParameterBinding.Tests.ps1
```diff
@@ -281,6 +281,39 @@
...
+
+        # Explict type cast use 'InvariantCulture' by default
+        $expected = [datetime] "1/11/1111"
```
iSazonov: Why we think that is 'InvariantCulture' ? I think we should have explicit constructor and explicit check that "Explict type cast use 'InvariantCulture' by default". Also it is another test then "Verify implicit argument conversion uses InvariantCulture".

test/powershell/engine/ParameterBinding/ParameterBinding.Tests.ps1
daxian-dbw: We can use `Get-Date -Date` instead, so no need to add a C# cmdlet.
```
PS:162> gcm Get-Date -Syntax
Get-Date [[-Date] <datetime>] ...
```

daxian-dbw:
There are some test failures, I will investigate whether they are test or code issues.

iSazonov:
I see all test failures is culture-specific - the PR is really breaking change! :-)

test/powershell/engine/ParameterBinding/ParameterBinding.Tests.ps1
lzybkr: Using invariant culture for a cast was a choice for consistency at the language level, casting a literal would always produce the same result in any culture.

If invariant culture conversions aren't appropriate, then one simply uses the appropriate code (constructor or whatever) and specifies how to do the conversion.

iSazonov: @lzybkr Thanks for useful comment! But my question was just about the test - what if we break internally and replace InvariantCulture with other culture accidentally? So the line and comment will be wrong.

lzybkr: Ah, I see, I was thinking it was a test for the conversion. I agree then, we should not rely on the default when checking results.

If we don't already, we should have a test that the cast is culture invariant.

daxian-dbw: I will change this line to `LangaugePrimitives.ConvertTo` explicitly with `InvariantCulture`. And also will add a test to validate if the explicit conversion in powershell language is using 'InvariantCulture'.

daxian-dbw:
The early failures were because I didn't set the culture back after setting it to 'ru-RU'. It's now fixed.

test/powershell/engine/ParameterBinding/ParameterBinding.Tests.ps1
```diff
@@ -293,20 +293,23 @@
...
 
         # Cast result with ru-RU culture
         $ruCulture = [cultureinfo]::GetCultureInfo("ru-RU")
         $ruCastResult = [System.Management.Automation.LanguagePrimitives]::ConvertTo("1/11/1111", [datetime], $ruCulture)
 
         # Implict parameter argument conversion should also use 'InvariantCulture'
-        $csharpCmdlet, $scriptFunction, $scriptCmdlet = @(
+        $csharpCmdlet, $scriptFunction, $scriptCmdlet = try {
+            $oldCulture = [CultureInfo]::CurrentCulture
             [CultureInfo]::CurrentCulture = 'ru-RU'
```
iSazonov: We use the 'ru-Ru' string in line 300 too - maybe put it in a variable?

iSazonov:
If we have so many culture-sensitive test should we fix them too?

test/powershell/Language/Parser/Conversions.Tests.ps1
```diff
@@ -21,6 +21,19 @@ Describe 'conversion syntax' -Tags "CI" {
         ([System.Management.Automation.ActionPreference]"Stop").ToString() | Should Be "Stop"
     }
 
+    It "language conversion uses 'InvariantCulture' by default" {
+        $dateString = "1/11/1111"
+        $expected = [System.Management.Automation.LanguagePrimitives]::ConvertTo($dateString, [datetime], [cultureinfo]::InvariantCulture)
+        $result = try {
+            $oldCulture = [CultureInfo]::CurrentCulture
```
PetSerAl: Should it be outside of `try` block?

test/powershell/engine/ParameterBinding/ParameterBinding.Tests.ps1
```diff
@@ -293,20 +293,23 @@
...
+        $expected = [System.Management.Automation.LanguagePrimitives]::ConvertTo("1/11/1111", [datetime], [cultureinfo]::InvariantCulture)
 
         # Cast result with ru-RU culture
         $ruCulture = [cultureinfo]::GetCultureInfo("ru-RU")
         $ruCastResult = [System.Management.Automation.LanguagePrimitives]::ConvertTo("1/11/1111", [datetime], $ruCulture)
 
         # Implict parameter argument conversion should also use 'InvariantCulture'
-        $csharpCmdlet, $scriptFunction, $scriptCmdlet = @(
+        $csharpCmdlet, $scriptFunction, $scriptCmdlet = try {
+            $oldCulture = [CultureInfo]::CurrentCulture
```
PetSerAl: Should it be outside of `try` block?

PetSerAl:
@iSazonov
>If we have so many culture-sensitive test should we fix them too?

IMO, tests which valid only in certain culture, like this one:
```none
 [-] using -format produces the correct output 41ms
   String lengths are both 9. Strings differ at index 0.
   Expected: {Jan-01-20}
   But was:  {янв-01-20}
   -----------^
   at line: 21 in D:\PowerShell\test\powershell\Modules\Microsoft.PowerShell.Utility\Get-Date.Tests.ps1
   21:         Get-date -Date:"Jan 1, 2020"  -Format:"MMM-dd-yy" | Should be "Jan-01-20"
```
should explicitly set that culture for themselves.

test/powershell/engine/ParameterBinding/ParameterBinding.Tests.ps1
daxian-dbw: Fixed.

test/powershell/Language/Parser/Conversions.Tests.ps1
daxian-dbw: Fixed.

test/powershell/engine/ParameterBinding/ParameterBinding.Tests.ps1
daxian-dbw: Fixed.

daxian-dbw:
Some tests call `ToString()` in them, which makes them culture dependent. Maybe it's not really necessary to call the `ToString` method in the first place.

iSazonov (approved):

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and due to the potential for breaking many existing modules, it's too risky to make this change so that c# and script cmdlets are alike for this case.  We should keep the current behavior.

daxian-dbw:
Given the review decision of powershell committee, I'm closing this PR.

PR #2038, opened 2016-08-23
Fix Parameter Binder bug with Advanced Functions and ValueFromRemainingArguments

dlwyatt:
Resolves issues #2035 

In the current code, two calls to `BindParameter` are potentially made, and the assumption was that if a caller had specified `Set-ClusterOwnerNode foo,bar` instead of `Set-ClusterOwnerNode foo bar`, the first call would fail.

This was not the case with Advanced Functions, so the logic which was supposed to cover that case in `HandleRemainingArguments()` wasn't being triggered.  This update unwraps a single-element list first, rather than requiring a failed call to `BindParameter` before that happens.

As a result, `Write-Object` no longer needs its own foreach loop to compensate for being sent a `List<Object>` from `HandleRemainingArguments()`.

PR #5109, opened 2017-10-13
More value from remaining arguments fixes

dlwyatt:
Expands on a previous PR based on conversation here:  https://github.com/PowerShell/PowerShell/pull/2038#discussion_r144460107

Unwrapping of ValueFromRemainingArguments was being performed only for object[] arrays (which covers the most common PowerShell binding scenarios), but could be skipped if a collection of any other type were passed to the parameter.

### DIFF
--- a/src/System.Management.Automation/engine/CmdletParameterBinderController.cs
+++ b/src/System.Management.Automation/engine/CmdletParameterBinderController.cs
@@ -1691,7 +1691,7 @@ namespace System.Management.Automation
                         //     Set-ClusterOwnerNode foo bar
                         //     Set-ClusterOwnerNode foo,bar
                         // we unwrap our List, but only if there is a single argument of type object[].
-                        if (valueFromRemainingArguments.Count == 1 && valueFromRemainingArguments[0] is object[])
+                        if (valueFromRemainingArguments.Count == 1 && null != LanguagePrimitives.GetEnumerable(valueFromRemainingArguments[0]))
                         {
                             cpi.SetArgumentValue(UnboundArguments[0].ArgumentExtent, valueFromRemainingArguments[0]);
                         }
--- a/src/System.Management.Automation/engine/CmdletParameterBinderController.cs
+++ b/src/System.Management.Automation/engine/CmdletParameterBinderController.cs
@@ -1691,7 +1691,7 @@ namespace System.Management.Automation
                         //     Set-ClusterOwnerNode foo bar
                         //     Set-ClusterOwnerNode foo,bar
                         // we unwrap our List, but only if there is a single argument which is a collection.
-                        if (valueFromRemainingArguments.Count == 1 && null != LanguagePrimitives.GetEnumerable(valueFromRemainingArguments[0]))
+                        if (valueFromRemainingArguments.Count == 1 && null != valueFromRemainingArguments[0] && LanguagePrimitives.IsTypeEnumerable(valueFromRemainingArguments[0].GetType()))
                         {
                             cpi.SetArgumentValue(UnboundArguments[0].ArgumentExtent, valueFromRemainingArguments[0]);
                         }
--- a/src/System.Management.Automation/engine/CmdletParameterBinderController.cs
+++ b/src/System.Management.Automation/engine/CmdletParameterBinderController.cs
@@ -1691,7 +1691,7 @@ namespace System.Management.Automation
                         //     Set-ClusterOwnerNode foo bar
                         //     Set-ClusterOwnerNode foo,bar
                         // we unwrap our List, but only if there is a single argument which is a collection.
-                        if (valueFromRemainingArguments.Count == 1 && null != valueFromRemainingArguments[0] && LanguagePrimitives.IsTypeEnumerable(valueFromRemainingArguments[0].GetType()))
+                        if (valueFromRemainingArguments.Count == 1 && LanguagePrimitives.IsObjectEnumerable(valueFromRemainingArguments[0]))
                         {
                             cpi.SetArgumentValue(UnboundArguments[0].ArgumentExtent, valueFromRemainingArguments[0]);
                         }
--- a/src/System.Management.Automation/engine/CmdletParameterBinderController.cs
+++ b/src/System.Management.Automation/engine/CmdletParameterBinderController.cs
@@ -1690,7 +1690,7 @@ namespace System.Management.Automation
                         //     Set-ClusterOwnerNode -Owners foo,bar
                         //     Set-ClusterOwnerNode foo bar
                         //     Set-ClusterOwnerNode foo,bar
-                        // we unwrap our List, but only if there is a single argument of type object[].
+                        // we unwrap our List, but only if there is a single argument which is a collection.
                         if (valueFromRemainingArguments.Count == 1 && null != LanguagePrimitives.GetEnumerable(valueFromRemainingArguments[0]))
                         {
                             cpi.SetArgumentValue(UnboundArguments[0].ArgumentExtent, valueFromRemainingArguments[0]);
--- a/src/System.Management.Automation/engine/LanguagePrimitives.cs
+++ b/src/System.Management.Automation/engine/LanguagePrimitives.cs
@@ -445,22 +445,6 @@ namespace System.Management.Automation
             return (getEnumerable != LanguagePrimitives.ReturnNullEnumerable);
         }
 
-        private static object GetBaseObject(object obj)
-        {
-            if (obj == null)
-            {
-                return null;
-            }
-
-            var psobj = obj as PSObject;
-            if (null != psobj)
-            {
-                obj = psobj.BaseObject;
-            }
-
-            return obj;
-        }
-
         /// <summary>
         /// Returns True if the language considers obj to be IEnumerable
         /// </summary>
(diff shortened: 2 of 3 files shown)


src/System.Management.Automation/engine/CmdletParameterBinderController.cs
```diff
@@ -1690,8 +1690,8 @@ private void HandleRemainingArguments()
                         //     Set-ClusterOwnerNode -Owners foo,bar
                         //     Set-ClusterOwnerNode foo bar
                         //     Set-ClusterOwnerNode foo,bar
-                        // we unwrap our List, but only if there is a single argument of type object[].
-                        if (valueFromRemainingArguments.Count == 1 && valueFromRemainingArguments[0] is object[])
+                        // we unwrap our List, but only if there is a single argument which is a collection.
+                        if (valueFromRemainingArguments.Count == 1 && null != LanguagePrimitives.GetEnumerable(valueFromRemainingArguments[0]))
```
daxian-dbw: I suggest using `LanguagePrimitives.IsTypeEnumerable(Type)` for 2 reasons:
1. `LanguagePrimitives.GetEnumerable` unwrap PSObject, which might strip off ETS properties associated with the PSObject.
2. `LanguagePrimitives.GetEnumerable` actually calls the `GetEnumerable` delegate which may throw an exception or cause a long delay (e.g. fetching object from Azure).

iSazonov:
@dlwyatt Please add test(s) to cover the change.

src/System.Management.Automation/engine/CmdletParameterBinderController.cs
dlwyatt: Will do

dlwyatt:
Actually, I don't like how that was changed.  Sec for another commit.

daxian-dbw:
How about changing `IsTypeEnumerable` to return false when the passed-in type is null? Then the change could become
```
if (valueFromRemainingArguments.Count == 1 && LanguagePrimitives.IsTypeEnumerable(valueFromRemainingArguments[0]?.GetType()))

```

dlwyatt:
That's part of what I'm doing, but I also want the code from GetEnumerable that checks for PSObject / BaseObject to run.

dlwyatt:
I added a new LanguagePrimitives.IsObjectEnumerable method to do this, and refactored some duplicated code to a new internal method.  Let me know if this is acceptable, since it's making a bit of new public API in LanguagePrimitives.

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -472,6 +466,33 @@ public static IEnumerable GetEnumerable(object obj)
...
+        /// <param name="obj">
+        /// IEnumerable or IEnumerable-like object
+        /// </param>
+        [SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", MessageId = "obj", Justification = "Since V1 code is already shipped, excluding this message.")]
+        public static IEnumerable GetEnumerable(object obj)
+        {
+            Type objectType = GetBaseObjectType(obj);
+            if (objectType == null) { return null; }
             GetEnumerableDelegate getEnumerable = GetOrCalculateEnumerable(objectType);
             return getEnumerable(obj);
```
daxian-dbw: This is wrong. Before the change, the `obj` passed in `getEnumerable` is the base object if the original one is PSObject, but now it will be the PSObject in that case.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -440,18 +440,12 @@ private static void InitializeGetEnumerableCache()
...
 
-        /// <summary>
-        /// Retrieves the IEnumerable of obj or null if the language does not consider obj to be IEnumerable
-        /// </summary>
-        /// <param name="obj">
-        /// IEnumerable or IEnumerable-like object
-        /// </param>
-        [SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", MessageId = "obj", Justification = "Since V1 code is already shipped, excluding this message.")]
-        public static IEnumerable GetEnumerable(object obj)
+        internal static Type GetBaseObjectType(object obj)
```
daxian-dbw: It's recommended to always start from `private`, since currently nothing outside the type depends on it.

daxian-dbw:
As for the new public API `LanguagePrimitives.IsObjectEnumerable(object)`, it could be useful. I will let @PowerShell/powershell-committee review it.

src/System.Management.Automation/engine/LanguagePrimitives.cs
dlwyatt: Oops, you're right.  Fixing now.

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -440,39 +440,52 @@ private static void InitializeGetEnumerableCache()
...
+        /// <summary>
+        /// Retrieves the IEnumerable of obj or null if the language does not consider obj to be IEnumerable
+        /// </summary>
+        /// <param name="obj">
+        /// IEnumerable or IEnumerable-like object
+        /// </param>
+        [SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", MessageId = "obj", Justification = "Since V1 code is already shipped, excluding this message.")]
+        public static IEnumerable GetEnumerable(object obj)
+        {
+            obj = GetBaseObject(obj);
```
daxian-dbw: How about replace `GetBaseObject(object)` with [`PSObject.Base(object)`](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/MshObject.cs#L989)? It seems a good fit here.

dlwyatt: I just knew there was already going to be a method for this somewhere in the code.  :)  I searched for .BaseObject and didn't find it; Base() uses the field name instead.

daxian-dbw (approved):

daxian-dbw:
Need @PowerShell/powershell-committee to review the new public API `public static bool IsObjectEnumerable(object obj)` in `LanguagePrimitives`.

Just noticed the PR #5123 was submitted to always unwrap single argument for `ValueFromRemainingArguments`. @dlwyatt thoughts?

dlwyatt:
I don't think I like that other PR.  It's changed Write-Output's parameter from an array type to a single PSObject (which may be undesirable by itself); ValueFromRemainingArguments should really only be used on array parameters, and the change to the parameter binder is to try to shoehorn it into use on a non-array type.

dlwyatt:
The issue #5122 is interesting, though.  I'm not sure whether to consider it a bug or not; we knew that the ValueFromRemainingArguments fix was a breaking change.

dlwyatt:
I suppose it doesn't matter if the change in #5123 is made to the binder, though.  If you assume that ValueFromRemainingArguments should only be used for array type parameters, then it's no harm done; the non-collection argument will be converted back into an array anyway.

If ValueFromRemainingArguments should be usable on non-array parameters, then the change in #5123 has value.  However, I can't think of any case where that would work other than [object] and [psobject].  Otherwise it's just going to fail when you do pass in multiple arguments to that parameter.  I've only used it like "params" in C#, and that only works for arrays.

PetSerAl:
@dlwyatt `ValueFromRemainingArguments` was never prohibited for non-array types. All you need is to specify a conversion:
```
PS> class MyClass {
>>>     $a; $b; $c
>>>     MyClass([Object]$Object) {
>>>         $this.a, $this.b, $this.c = $Object
>>>     }
>>> }
PS> function f { param( [Parameter(ValueFromRemainingArguments)][MyClass] $Class ) $Class }
PS> f 1 2 3

a b c
- - -
1 2 3


PS> f (f 1 2 3)

a       b c
-       - -
MyClass
```
So, if array pass thru, when passed as single argument, why `MyClass` should not?

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and on the topic of a single element array being a scalar, our decision is that it is simpler for the developer to always expect an array rather than having code handle both an array and a scalar, so it should be an array

daxian-dbw:
Nothing stops a user from applying `ValueFromRemainingArguments` on non-array type parameters, but I don't think that's the right way to use the attribute. I agree with @dlwyatt that `ValueFromRemainingArguments` should be similar to `params` in C#, which means when it gets to the point to call `BindParameter`, the provided argument should always be a collection.

daxian-dbw:
@lzybkr Could you please weigh in? Your thoughts would be helpful.

lzybkr:
There are a couple ways to think about this.

In C#, it just doesn't make sense for `params` to not be an array - otherwise you need annoying code testing the type of the argument.

In PowerShell, that's somewhat less of an issue, the pipeline doesn't care, and since V3, many other places handle a single instance an array of objects in a somewhat similar way.

On the other hand, we do far too **little** checking on the use of attributes and if they make sense. I'm all for adding more errors on inappropriate use of an an attribute.

One more random thought - not necessarily a good idea, but the only case that truly needs to be a breaking change if we want an error - if the parameter type is explicitly specified and is not `object` or `PSObject` or an array of anything. For singular `object` or `PSObject`, the cmdlet must handle arrays anyway or it's broken, so there is little risk in always creating an array.  I suspect there is very little code out there that explicitly uses a parameter type constraint that isn't `object`, `PSObject`, or an array with `ValueFromRemainArguments`.

daxian-dbw:
Given the review from powershell committee, I will merge this PR. But please feel free to open an issue and continue the discussion.

Issue #2247, opened 2016-09-13
-Verbose Changes  (downgrades) the Error Behaviour


timker:
-Verbose changes (downgrades) the Error Action from a Stop to Continue. This is undocumented and confusing.
## Steps to reproduce

Run this script

```powershell
$ErrorActionPreference = "Stop"
$VerbosePreference = "continue"

#safely creating a file for first run
if(-Not (Test-Path test.txt))
{
    new-item test.txt  -Verbose 
}

#create the same file
new-item test.txt  -Verbose
Write-output 'This should not output because ErrorActionPreference is set to STOP'


new-item test.txt 
Write-output 'This should never output because of the error above, and it doesn't'
```
## Expected behavior

'This should not output because ErrorActionPreference is set to STOP', should not output
## Actual behavior

'This should not output because ErrorActionPreference is set to STOP' is output
## Bug Location

MshCommandRuntime.cs
    internal ActionPreference ErrorAction
     if (Verbose)
                    return ActionPreference.Continue;
doesn't will downgrade a Stop
## Proposed Solution

```c#
                if (!_isErrorActionPreferenceCached)
                {
                    bool defaultUsed = false;
                    _errorAction = Context.GetEnumPreference<ActionPreference>(SpecialVariables.ErrorActionPreferenceVarPath, _errorAction, out defaultUsed);

                    //Verbose should upgrade communication preferences, but not downgrade ErrorAction behavioural preferences    
                    if (Verbose)
                    {
                        // don't downgrade preferences 
                        if (!(_errorAction == ActionPreference.Stop || _errorAction == ActionPreference.Suspend))
                        {
                            _errorAction = ActionPreference.Continue;
                        }
                    }

                    _isErrorActionPreferenceCached = true;
                }
```
## Environment data

```
Major Minor Patch Label
----- ----- ----- -----
   6     0     0 alpha
```

joeyaiello:
This is a breaking change that we'd need to review, but if we're going to take it, there's the code, and we should do it sooner rather than later.

SteveL-MSFT:
@PowerShell/powershell-committee it appears `-Debug` also has the same issue.  Committee agrees that we should fix this so that `-ErrorAction` works correctly and not overridden by `-Debug` or `-Verbose`, but @lzybkr would like to think about the implications of this change

mintsoft:
We just ran into this bug too; +1 for changing it to respect the `$ErrorActionPreference`

[...]

PR #5113, opened 2017-10-14
-Verbose should not override $ErrorActionPreference

SteveL-MSFT:
Fix https://github.com/PowerShell/PowerShell/issues/2247

### DIFF
--- a/test/powershell/Language/Scripting/ActionPreference.Tests.ps1
+++ b/test/powershell/Language/Scripting/ActionPreference.Tests.ps1
@@ -127,11 +127,22 @@
             $num | Should Be 2
         }
 
-        It '-Verbose does not take precedence over $ErrorActionPreference' {
+        It '<switch> does not take precedence over $ErrorActionPreference' -TestCases @(
+            @{switch="-Verbose"},
+            @{switch="-Debug"}
+        ) {
+            param($switch)
             $ErrorActionPreference = "SilentlyContinue"
-            New-Item -ItemType File -Path "$testdrive\test.txt"
-            { New-Item -ItemType File -Path "$testdrive\test.txt" -Verbose } | Should Not Throw
+            $params = @{
+                ItemType = "File";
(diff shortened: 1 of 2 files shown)


anmenaga (changes requested):
This change is only for Verbose;
however, original issue says that it was approved for both: Verbose and Debug : "Committee agrees that we should fix this so that -ErrorAction works correctly and not overridden by -Debug or -Verbose".
Should this change be done for Debug as well?

SteveL-MSFT:
@anmenaga that's a fair point about `-Debug`.  I'll update this.

SteveL-MSFT:
@anmenaga can you review again?

test/powershell/Language/Scripting/ActionPreference.Tests.ps1
```diff
@@ -126,4 +126,23 @@
...
+            @{switch="-Debug"}
+        ) {
+            param($switch)
+            $ErrorActionPreference = "SilentlyContinue"
+            $params = @{
+                ItemType = "File";
+                Path = "$testdrive\test.txt";
+                Confirm = $false
+            }
+            New-Item @params
```
adityapatwardhan: ` > $null`

adityapatwardhan (changes requested):

test/powershell/Language/Scripting/ActionPreference.Tests.ps1
SteveL-MSFT: will add

anmenaga (approved):

SteveL-MSFT:
@adityapatwardhan can you review again?

adityapatwardhan (approved):

mintsoft:
Is this likely to make it into the final for Powershell 6?

markekraus:
@mintsoft This was merged to master so it will be in the release unless it needs to be reverted for some reason.

PR #5114, opened 2017-10-14
Fetch resource string correctly

daxian-dbw:
I found `Add-Member` throws error with an empty message. This PR fixes it.

**Before fix**
```
PS:2> Add-Member -InputObject $object "ABC" "Value1"
[C:\arena\pscore]
PS:3> Add-Member -InputObject $object "ABC" "Value1"
Add-Member :
At line:1 char:1
+ Add-Member -InputObject $object "ABC" "Value1"
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (1 2:PSObject) [Add-Member], InvalidOperationException
    + FullyQualifiedErrorId : MemberAlreadyExists,Microsoft.PowerShell.Commands.AddMemberCommand
```

**After fix**
```
PS:8> Add-Member -InputObject $object "ABC" "Value1"
Add-Member : Cannot add a member with the name "ABC" because a member with that name already exists. To overwrite the
member anyway, add the Force parameter to your command.
At line:1 char:1
+ Add-Member -InputObject $object "ABC" "Value1"
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (1 2:PSObject) [Add-Member], InvalidOperationException
    + FullyQualifiedErrorId : MemberAlreadyExists,Microsoft.PowerShell.Commands.AddMemberCommand
```
Also updated a resource string use in `ConvertTo-Json` to directly use the C# property.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
@@ -381,10 +381,8 @@ namespace Microsoft.PowerShell.Commands
 
         private ErrorRecord NewError()
         {
-            ErrorDetails details = new ErrorDetails(this.GetType().GetTypeInfo().Assembly,
-                "WebCmdletStrings", "JsonStringInBadFormat");
             ErrorRecord errorRecord = new ErrorRecord(
-                new InvalidOperationException(details.Message),
+                new InvalidOperationException(WebCmdletStrings.JsonStringInBadFormat),
                 "JsonStringInBadFormat",
                 ErrorCategory.InvalidOperation,
                 InputObject);
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddMember.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddMember.cs
@@ -514,7 +514,7 @@ namespace Microsoft.PowerShell.Commands
         private ErrorRecord NewError(string errorId, string resourceId, object targetObject, params object[] args)
         {
             ErrorDetails details = new ErrorDetails(this.GetType().GetTypeInfo().Assembly,
-                "AddMember", resourceId, args);
+                "Microsoft.PowerShell.Commands.Utility.resources.AddMember", resourceId, args);
             ErrorRecord errorRecord = new ErrorRecord(
                 new InvalidOperationException(details.Message),
                 errorId,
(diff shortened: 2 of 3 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -381,10 +381,8 @@ private void AddSpaces(int numberOfSpacesToReturn, StringBuilder result)
 
         private ErrorRecord NewError()
         {
-            ErrorDetails details = new ErrorDetails(this.GetType().GetTypeInfo().Assembly,
-                "WebCmdletStrings", "JsonStringInBadFormat");
             ErrorRecord errorRecord = new ErrorRecord(
-                new InvalidOperationException(details.Message),
+                new InvalidOperationException(WebCmdletStrings.JsonStringInBadFormat),
```
iSazonov: Is the change for the PR? I don't see a description and a test.

daxian-dbw: If the resource file name and the resource string id is known/fixed, there is no point to use `ResourceManager` to retrieve the error message (which is what `ErrorDetails` does). Instead, we can just use the generated C# property.
The existence of the resource string is guaranteed by the build, so I didn't add a test for it.

iSazonov: I meant that the PR description mention only `Add-Member` cmdlet but not `ConvertTo-Json`.

daxian-dbw: Updated the description to include the `ConvertTo-Json` change.

PaulHigin: The error string for this resource (WebCmdletStrings.JsonStringInBadFormat) doesn't seem right:
"The converted JSON string is in bad format."
I think this would be a better error message:
"The JSON string could not be converted because the format is incorrect."

PaulHigin (approved):
Please also update the WebCmdletStrings.JsonStringInBadFormat error string.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
daxian-dbw: This code is from `ConverTo-Json` cmdlet, which takes arbitrary objects and converts them to JSON string. The error message is used when it detects the JSON string we converted to is in bad format.

iSazonov (approved):

adityapatwardhan (approved):

PR #5118, opened 2017-10-14
Add global.json to pick correct SDK version

### LINKED ISSUE #5015: Need global.json to set the desired SDK version
I have .NET Core `2.0.2-vspre-006949` installed as well as `2.0.0`.  When I try to build with Start-PSBuild, I get this error:

```
WARNING: The currently installed .NET Command Line Tools is not the required version.

Installed version: 2.0.2-vspre-006949
Required version: 2.0.0

Fix steps:

1. Remove the installed version from:
    - on windows '$env:LOCALAPPDATA\Microsoft\dotnet'
    - on macOS and linux '$env:HOME/.dotnet'
2. Run Start-PSBootstrap or Install-Dotnet
3. Start-PSBuild -Clean
```
FWIW I had just recently run `Start-PSBootstrap` and restarted the console.  And when I got the above error I was running `Start-PSBuild -Clean`.

But in this case, the issue is easily fixed by putting a `global.json` file in the root of the repo with these contents:
```
{
  "sdk": {
    "version": "2.0.0"
  }
}
```

Environment data
----------------
I'm building on Windows PowerShell 5.1 on Windows 10 CU.

```powershell
> $PSVersionTable

PSVersion                      5.1.15063.608
BuildVersion                   10.0.15063.608
```

rkeithhill:
Fix #5015 

This sets the version of the dotnet SDK that should be used to compile the project.  I happen to have newer version of the SDK (2.0.2) on my machine and without this file, the build attempts to use the newer version of the SDK which doesn't work.

BTW if you have 2.0.2 installed, without this file when you run `Start-PSBuild` you get this warning and the build stops - even if you have 2.0.0 installed:
```
5> Start-PSBuild
WARNING: The currently installed .NET Command Line Tools is not the required version.

Installed version: 2.0.2
Required version: 2.0.0

Fix steps:

1. Remove the installed version from:
    - on windows '$env:LOCALAPPDATA\Microsoft\dotnet'
    - on macOS and linux '$env:HOME/.dotnet'
2. Run Start-PSBootstrap or Install-Dotnet
3. Start-PSBuild -Clean
```


global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "2.0.0"
+  }
+}
```
markekraus: Please add a newline at the end of the file.

markekraus (changes requested):

global.json
rkeithhill: Done.  BTW the original file was simply the output of running:
```
dotnet new global.json --sdk-version 2.0.0
```

markekraus (approved):

iSazonov:
@daxian-dbw Could you please review this bit change? It looks very good.

rkeithhill:
Any update on this?  I'm constantly having to create this file on every new branch I create. Otherwise, I can't build.  Plus this allows the PowerShell team to specify *which* version of the SDK is requires.  This will reduce contributor issues when their build fails with the message above.  And when the team is ready to move to a later version of the SKD like 2.0.2, the global.json can be updated to reflect that change.

SteveL-MSFT (approved):

daxian-dbw:
@rkeithhill Thanks for this PR! Sorry that we had it slipped off our radar.

PR #5121, opened 2017-10-14
Fix MatchInfoContext Clone implementation

### LINKED ISSUE #3994: Clone() implemented improperly for MatchString.cs?
It looks like the Clone functions in MatchString.cs often reference the clone when they meant to reference "this" (explicitly or implicitly). For example, https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs#L65 checks the newly-created object to see if it has a PreContext, when it probably should check the current object.

It looks like this mistake occurs in both Clone() functions.

Also, where is the definition of the Match class?

dee-see:
Fix #3994 

The old implementation always set the new object's properties to null.

The original issue mentioned that the other `Clone` implementation in MatchString.cs had the same bug, but in fact it's correct because of the use of `MemberwiseClone`.

I didn't much unit testing of C# code going on in the project (or maybe I missed it?). Should I add a test and if so, where?

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/MatchString.cs
@@ -61,12 +61,13 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         public object Clone()
         {
-            MatchInfoContext clone = new MatchInfoContext();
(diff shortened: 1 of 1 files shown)


lzybkr (approved):
Code change is fine.

A test would be nice, ideally failing if a new property is introduced but not cloned. That might be hard, so just checking the number of properties and a good comment might be sufficient.

PR #5124, opened 2017-10-15
Make testablescript.ps1 binary

PetSerAl:
As file `testablescript.ps1` used in hash testing, it should be checked-out without any line ending correction.

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/assets/.gitattributes
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/assets/.gitattributes
@@ -1,1 +1,3 @@
+# testablescript.ps1 used in Get-FileHash tests.
+# It need to be checked-out without any conversion.
 /testablescript.ps1 binary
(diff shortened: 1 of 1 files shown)


SteveL-MSFT (approved):
Good idea!

lzybkr:
Why not generate the file instead of committing a binary file?

PetSerAl:
@lzybkr Given that file is 14 bytes long, it would mean to embed file in tests directly. It could be done this way, but I do not see any benefits in changing it now.

iSazonov:
Also we would remove \r\n at EOF. Although I agree that it's better to generate this file, it makes the tests more understandable and more reliable.

lzybkr:
There are benefits to self contained tests, 14 bytes means it's a very readable test too since the embedded file is so short.

If we were talking about a longer file, then I might feel differently.

iSazonov:
Longer asserts files we have is for certificates - it seems no problems with them.

iSazonov:
@PetSerAl Plase add comment why the Issue was closed. Thanks!

PetSerAl:
@iSazonov It looks like @lzybkr prefer to include `testablescript.ps1` in tests directly, rather then keeping it as separate file in repository. That way this PR would be unnecessary.

PR #5125, opened 2017-10-15
Enhance -split operator with RightToLeft splitting

IISResetMe:
This pull request adds RightToLeft support to the binary -split operator, as requested in #4765 

I've added tests for negative integer input for the `Max-substrings` operand to work with both regex patterns and scriptblocks.


test/powershell/Language/Operators/SplitOperator.Tests.ps1
```diff
@@ -30,13 +30,18 @@ Describe "Split Operator" -Tags CI {
...
+            $res[2] | Should Be "d"
 
             $res = "a b c d" -split " ", -1
-            $res.count | Should Be 4
-            $res[0] | Should Be "a"
-            $res[1] | Should Be "b"
-            $res[2] | Should Be "c"
-            $res[3] | Should Be "d"
+            $res.count | Should Be 1
+            $res[0] | Should Be "a b c d"
```
iSazonov: I believe we want to get "d" as a last element.
/cc @mklement0

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -681,24 +696,30 @@ private static object SplitWithPredicate(ExecutionContext context, IScriptExtent
...
                 // Add any remainder, if we're under the limit.
                 if (buf.Length > 0 &&
-                    (limit <= 0 || split.Count < limit))
+                    (limit == 0 || split.Count < limit))
                 {
                     split.Add(buf.ToString());
                 }
 
+                if(rightToLeft)
+                    split.Reverse();
```
iSazonov: Can we exclude `reverse`?

test/powershell/Language/Operators/SplitOperator.Tests.ps1
IISResetMe: While that might seem intuitive, it would make the `Max-substrings` operand asymmetrical. 

The guiding principal behind the change is that the sign (`+/-`) signifies only the direction of the split operation, whereas the absolute value determines the maximum resulting number of substrings, with 0 preserving the meaning of "unlimited"

mklement0: Actually, I wasn't thinking of _direction_, but _what end_ to apply the split to - elements returned should always reflect the input order.

In this case, due to `-1`, the _last 1_ element should be returned, preceded by the remaining _prefix_, so we should get:

```none
$rest, $result = 'a b c d' -split '', -1
$rest | Should Be 'a b c'
$result.Count | Should be 1
$result[0] | Should Be 'd'
```

src/System.Management.Automation/engine/lang/parserutils.cs
IISResetMe: Remove the `List<string>.Reverse()` call for performance reasons or...? 

Removing it would funtionally result in the substrings for each input being returned in the wrong order, ie:

```
PS> 'a b c d' -split {$_ -eq ' '},-3
d
c
a b
```

Rather than the expected:

```
PS> 'a b c d' -split {$_ -eq ' '},-3
a b
c
d
```

I see two possible alternatives: 
1. Change all invocations of `split.Add()` to `split.Insert()` conditional on `rightToLeft`
2. Implement two completely separate routines for negative and positive `limit` values, duplicating the code slightly (but keeping performance overhead at a minimum)

test/powershell/Language/Operators/SplitOperator.Tests.ps1
IISResetMe: Whether you call it *direction* or not doesn't change the fact that by splitting from the *end* rather than the *start* we are in fact modifying the direction of the operation :-) 

Wouldn't your proposed result be confusing given the name of the operand (`Max-substrings` rather than `Number-Of-Times-To-Split` or similar)?

mklement0: _Direction_ is ambiguous, as that could be interpreted as wanting to also reverse the order in which the resulting elements are returned (e.g., `'a b c d' -split ' ', -2` resulting in `'a b', 'd, 'c'`, which we don't want), so I wanted to clarify.

To me, `Max-Substrings` makes sense: a negative value means: return `Max-Substrings` elements _from the end_ (in input order).

mklement0: Re `Max-Substrings`: I see your point now. With my proposal, you actually get `Max-SubStrings + 1` elements back - this is the asymmetry you mention.

mklement0: I agree that your approach is better for consistency with how `-split` currently works.

Given that what was approved was my approach, however, we should get approval again: cc @SteveL-MSFT

IISResetMe: If you think about how negative array indices are treated in PowerShell, it makes perfect sense, and as https://github.com/PowerShell/PowerShell/pull/5125/commits/1d59be0f17e3b4e4b6c5d57daaf882b364f2939f shows, shifting to your original suggestion is trivial, but I would like to keep it symmetric (again, given that the second operand is not an index but a boundary value)

mklement0: Understood and agreed.

I indeed originally had array-index logic in mind, but neglected to consider that the current `-split` behavior uses the-number-of-elements-I-want-plus-one-for-the-rest logic.

Just to spell it out: with your approach (which we should go with), both `1` and `-1` are no-ops - both request no splitting at all, and which end to apply it to is then moot.

SteveL-MSFT: I agree that the symmetric behavior makes sense and updating the documentation with good examples will clear up any confusion. cc @lzybkr

iSazonov: @mklement0 Could you please modify your Issue description to satisfy the behavior?

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: Yes, my thoughts was about performance. I don't know what case is better. I guess we would use only one routine. At a minimum, we have to start with this.

IISResetMe: Splitting it into 2 loops would mean getting rid of at least one non-local variable assignment per input item and a number of branch points, the intent (splitting from the end) would also be clearer. Any thoughts @lzybkr ?

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -637,6 +637,12 @@ private static object SplitWithPredicate(ExecutionContext context, IScriptExtent
             {
                 List<string> split = new List<String>();
```
iSazonov: It seems we could move this to line 653 as
```c#
List<string> split = new List<String>(limit); 
```

IISResetMe: I'd keep it without specifying capacity up front, because you'd still need to support unbounded capacity in the special case of (`limit == 0`) and the (very common) case of _no `Max-substrings` specified_.

PetSerAl: @iSazonov What if someone put `1000000` as `limit`? I do not think you should pre-allocate so much memory.

test/powershell/Language/Operators/SplitOperator.Tests.ps1
mklement0: @iSazonov: Thanks for reminding me - done.

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: I skipped ` limit == 0 `
Closed.

src/System.Management.Automation/engine/lang/parserutils.cs
lzybkr: `List.Reverse` is not expensive, it swaps in place, see the [implementation](http://source.dot.net/#System.Private.CoreLib/src/System/Array.cs,b242f3290263f0f2) (List is really an Array).

Insert would be more expensive, shifting everything on each insert instead of swapping once like in Reverse.

src/System.Management.Automation/engine/lang/parserutils.cs
powercode: Why shouldn't we specify the limit in the ctor? 
I'd gladly have powershell allocate the list once to the size I had given than have the list being reallocated and copied many times when the value is larger?
If someone specifies `1000000` they will get a list of roughly 8MB, which would soon be garbage collected, but with a single allocation, instead of the 19 we would get with the 0 default capacity.

IISResetMe: Fair point. We could also allocate the list with capacity `Min(limit, item.Length +2)` (unless there's some negative-width split pattern I can't think of)

iSazonov: I agree - it is good. Reallocation is much worse.

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: We could do reverse copy in next line in `ExtendList`.
By the way, what are we doing this recopy?

src/System.Management.Automation/engine/lang/parserutils.cs
IISResetMe: Forget about that `Min(limit, item.Length +2)` comment above, I just realized we're talking about regex with capture support, including `'abcde' -split '((?=.)(.)(?<=.))'` similarly terrifying things.

I've added the suggested changes + comments for the `List.Reverse()` call in https://github.com/PowerShell/PowerShell/pull/5125/commits/c423d8d6c998dd01ca9e7e2c54752d635bd7dd78

lzybkr: I still like your thought to use `Min(limit, item.Length +2)`. It covers the common case sufficiently and protects against complete stupidity.

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: Up.

src/System.Management.Automation/engine/lang/parserutils.cs
IISResetMe: Just had a look at the [`List<T>` source](https://referencesource.microsoft.com/#mscorlib/system/collections/generic/list.cs,51), the default capacity reallocation behavior (multiply by 2) actually fits our needs perfectly if we do `Min(limit, item.Length + 2)` - matches worst case with no captures, and each nested capture-group added will at most require 1 extra re-allocation.

iSazonov: [CoreFX link on EnsureCapacity](https://github.com/dotnet/corefx/blob/bffef76f6af208e2042a2f27bc081ee908bb390b/src/System.Collections/src/System/Collections/Generic/List.cs#L433)

TravisEz13:
@iSazonov Can you update your review?   
@PetSerAl and @powercode  Please feel free to comment if you think if this is ready to be merged or not.  And, thanks for your feedback.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -681,24 +715,37 @@ private static object SplitWithPredicate(ExecutionContext context, IScriptExtent
...
                     split.Add(buf.ToString());
                 }
 
+                if(rightToLeft)
+                {
+                    // We want to preserve the order from the
+                    // original string (always output chunks left-to-right)
+                    split.Reverse();
+                }
                 ExtendList(results, split);
```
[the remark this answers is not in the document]
IISResetMe: We _can't_ copy in reverse order without doing `split.Insert(0,item)` for each chunk. As @lzybkr noted earlier `List<T>.Insert()` is much more costly (it shifts the underlying array by copying every time) than simply adding in the order of the cursor and calling `Reverse()` before adding.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -644,10 +650,31 @@ private static object SplitWithPredicate(ExecutionContext context, IScriptExtent
...
+                    split = new List<String>();
+                }
+                else
+                {
+                    // Limit was specified by the user
+                    // instantiate list with maximum needed capacity
+                    split = new List<String>(limit);
+                }
+                
                 StringBuilder buf = new StringBuilder();
```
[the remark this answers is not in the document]
IISResetMe: Required capacity for the final list of substrings is easy to predict with a reasonable degree of confidence based on length alone. 

Required capacity for the individual substrings is near-impossible to predict, as we know nothing about the character distribution up front.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -644,10 +650,31 @@ private static object SplitWithPredicate(ExecutionContext context, IScriptExtent
                     results.Add(item);
                     continue;
                 }
-
+                
+                List<string> split; 
+                if(limit == 0)
+                {
+                    split = new List<String>();
```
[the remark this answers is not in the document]
IISResetMe: Reviewed the code again, realized that the scenario I was trying to predict above doesn't actually apply to SplitWithPredicate(), worst case is `item.Length + 1`. I've added this to the constructor call when `limit == 0`.

src/System.Management.Automation/engine/lang/parserutils.cs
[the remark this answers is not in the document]
iSazonov: I meant:
```c#
        private static void ExtendListReverse<T>(IList<T> list, IList<T> items)
        {
            for(int i=items.Count-1; i >= 0 ; i++ )
            {
                list.Add(items[i]);
            }
        }
```

src/System.Management.Automation/engine/lang/parserutils.cs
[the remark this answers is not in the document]
iSazonov: Closed.

src/System.Management.Automation/engine/lang/parserutils.cs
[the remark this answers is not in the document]
iSazonov: My thoughts was about `StringBuilder buf = new StringBuilder(item.Length); `.
Another thought is that we recreate StringBuilder in cycle. I think we can allocate it before the external cycle `foreach (string item in content)` and reuse in the line with buf.Clear();

src/System.Management.Automation/engine/lang/parserutils.cs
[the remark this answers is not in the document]
powercode: LGTM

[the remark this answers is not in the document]
IISResetMe: @iSazonov Ahh, that makes sense (I completely misunderstood your request). While testing, I realized that I inadvertently broke the operation for subsequent strings when working on arrays due to the [modification of the `limit` param](https://github.com/PowerShell/PowerShell/pull/5125/files/5cf38f49940b3c31ad51088e2962186389c45939#diff-ebfc1e729c1362fdfc74174d6e5bafaeR643), resulting in:

```
PS C:\> 'a b c','d e f' -split {$_ -eq ' '},-2
a b
c
d
e f
```

Will update test cases and copy `Abs(limit)` to a local var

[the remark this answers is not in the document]
iSazonov: Please add the new test.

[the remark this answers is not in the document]
IISResetMe: Test cases added, `limit` bug fixed. 

I decided against implementing `ExtendListReverse()` for two reasons:
1. It's the only current use case
2. The name `ExtendListReverse()` seems quite ambiguous

IISResetMe:
travis build on linux seems to have failed for reasons unrelated to the code, how to re-run without commiting?

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -635,19 +635,49 @@ private static object SplitWithPredicate(ExecutionContext context, IScriptExtent
...
+                    split = new List<String>(item.Length + 1);
+                }
+                else
+                {
+                    // Limit was specified by the user
+                    // instantiate list with maximum needed capacity
+                    split = new List<String>(maxSubstrings);
+                }
+                
                 StringBuilder buf = new StringBuilder();
```
iSazonov: We add a char by char in the `buf`. Doesn't make sense to recreate the `buf` after every separator.
Please move out of the cycle to the line 635 `StringBuilder buf = new StringBuilder(256); ` (default 16 is very short for typical strings) and replace the line 691 `buf = new StringBuilder();` on `buf.Clear();`.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -681,24 +718,37 @@ private static object SplitWithPredicate(ExecutionContext context, IScriptExtent
...
                         {
                             split.Add("");
                         }
                     }
                     else
                     {
-                        buf.Append(item[strIndex]);
+                        if(rightToLeft)
+                        {
+                            buf.Insert(0, item[strIndex]);
```
iSazonov: We add a char by char here so this is exactly the case when we should avoid `Insert`.
Please here remove `if(rightToLeft) ` and do only `buf.Append(item[strIndex]);` and add in the line 690
```
if(rightToLeft)
{
    buf.Reverse();
}
split.Add(buf.ToString());
buf.Clear();

File: src/System.Management.Automation/engine/lang/parserutils.cs (hunk 4 above)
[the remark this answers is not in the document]
Comment (iSazonov) [reply src/System.Management.Automation/engine/lang/parserutils.cs]: If we're starting to optimize, let's finish it.
We can use the name  `ReverseExtendList`.

I know it's a long time PR. Thank you for your patience. I think after my last comments, we're close to merge..

File: src/System.Management.Automation/engine/lang/parserutils.cs (hunk 8 above)
Comment (IISResetMe) [reply src/System.Management.Automation/engine/lang/parserutils.cs]: I wanted to do that when we first dicussed the same issue with the `List<string>`, but unfortunately there's no such thing as `StringBuilder.Reverse()` :-\

File: src/System.Management.Automation/engine/lang/parserutils.cs (hunk 7 above)
Comment (IISResetMe) [reply src/System.Management.Automation/engine/lang/parserutils.cs]: This is fixed now. `StringBuilder buf` declaration moved outside loop, constructor calls replaced with `.Clear()`

File: src/System.Management.Automation/engine/lang/parserutils.cs [hunk 9]
```diff
@@ -635,19 +635,49 @@ private static object SplitWithPredicate(ExecutionContext context, IScriptExtent
             List<string> results = new List<string>();
             foreach (string item in content)
             {
-                List<string> split = new List<String>();
+                int maxSubstrings = System.Math.Abs(limit);
```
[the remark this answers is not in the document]
IISResetMe: Moved `rightToLeft` assignment outside loop, reverted to `limit = Abs(limit)`

IISResetMe:
`buf` re-allocation, `maxSubstrings`/`limit` and whitespace issues have all been addressed now. 

I still disagree on `Reverse()` + `ExtendList()` v `ReverseExtendList()` - won't add any more clarity to the code and doesn't seem justifiable with this single very specific use case.

---

I'm still trying to work out an elegant way of doing `buf.Reverse()` - thought about an extension method, like:

```
private static StringBuilder Reverse(this StringBuilder chunk)
{
    string str = chunk.ToString();
    if (str.Length < 2) 
        return chunk;
    char[] charArray = str.ToCharArray();
    Array.Reverse(charArray);

    return new StringBuilder(new string(charArray));
}
```

But I'm not sure that's the best way to go about it for multiple reasons:

1. Given I won't have access to the underlying array, I'll have to materialize the string before reversing it
2. StringBuilder ctor doesn't take `char[]`, so need to construct a new string again after reversing (lesser concern) 
3. Unlike existing methods (`Append*()`, `Insert()`, `Remove()` etc) the extension method won't actually change state of existing `chunk`, only return a new (reversed) `StringBuilder`

Any thoughts, recommendations?

src/System.Management.Automation/engine/lang/parserutils.cs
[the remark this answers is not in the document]
IISResetMe: No worries, I want to get it right, this is a nice learning experience 👍 

Please see my latest comment re: `Reverse` of both the list and string buffer

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: I think we can replace  `StringBuilder` with `List<Char>`.

powercode:
StringBuilder has an ```System.Text.StringBuilder Append(char[] value, int startIndex, int charCount)``` overload.
If you create the stringbuilder with value.Length as intialCapacity and use that append overload you get away from the allocation of the string

iSazonov:
I think in the context we can easy replace  StringBuilder  with  List<Char> .

IISResetMe:
Nice one @powercode, hadn't considered that. I think @iSazonov is right though. There's no inherent value in using `StringBuilder` over `List<char>` in the current implementation, and `List<char>` gives us a fairly efficient `Reverse()` option for free

IISResetMe:
`StringBuilder` replaced with `List<char>`, and I've removed the `Insert()` call in favor of `Reverse()` prior to adding the chunk to the list.

TravisEz13:
I restarted travis-ci but I think you might need to rebase.

IISResetMe:
Any quick advice on how to do it correctly @TravisEz13 ? Last time I tried [it went horribly wrong](https://github.com/PowerShell/PowerShell/pull/5026#issuecomment-336746060)

TravisEz13:
Honestly, I just run `git rebase <upstreambranch> <targetbranch>` and follow the instruction.  A good conflict solver (merge tool) is very important.  Run `git status` if you need to know what you are supposed to do next.

The end of this article has a link to some good tools [GitHowTo - 30. Resolving Conflicts](http://githowto.com/resolving_conflicts ).
[Instructions to setup BC4, my favorite merge tool, but it's a paid product.](http://www.scootersoftware.com/support.php?zz=kb_vcs#gitwindows)

IISResetMe:
Alright, I rebased like you suggested, then did a pull and a push (as there were no merge conflicts), but it still dragged all the intermediate commits from the master branch into my PR :-\ @TravisEz13

iSazonov:
I use 
```
git chechout <branch>
git pull --rebase PowerShell master
```
I think you need remove the merge commit and try rebase again.

IISResetMe:
@TravisEz13 @iSazonov Tried to revert the last merge locally with 

    git revert c8eb3407ac839e5191b99c3a719a7629d9fdd746 -m 2

but this just adds a new commit reverting all the changes of the others at once, super messy. How can I "reset" the branch (`--reset hard [latest PR-speficic commit]` doesn't seem to work because of the previous rebase/fast-forwarding)? 

Maybe it's easier to just make a diff patch of the relevant files, create a new branch and submit a new PR?

Apologies for the git n00bness

SteveL-MSFT:
@IISResetMe the easier thing is probably to copy off the files you changed, reset this branch to before your first commit, copy over your changed files, then rebase off upstream/master, and push those changes.

If you want to do it the `git` way, you should `git rebase -i HEAD~56` (for your 56 commits), drop all the commits that aren't yours.  As it rebases, you'll hit lots of merge conflicts.  You'll have to fix them, `git add <file>` then `git rebase --continue` until it's clean.  Then `git fetch upstream` and `git rebase upstream/master` and resolve any merge conflicts.

iSazonov:
You can remove last commits by

git reset --hard ooooooo

IISResetMe:
Talked to @SteveL-MSFT offline, and decided to start over and push the changes to a new branch with the following commits:

https://github.com/IISResetMe/PowerShell/commit/81873b30f061056710a7417d7aafa4f091f6e1f9 (tests)
https://github.com/IISResetMe/PowerShell/commit/5e5e3e2f3c643cef73cb21968c58f4f3fc0f5a6c (SplitWithPattern())
https://github.com/IISResetMe/PowerShell/commit/cf47352d71d0e78d39fc634a2d1e659c1030a6ef (SplitWithPredicate())

Only concern is that we lose a bit of context from this PR, so wanted to check with you @TravisEz13 @iSazonov before submitting a new PR off the new branch

iSazonov:
Ok, only please add a comment in new PR that it overload the PR.

IISResetMe:
Re-submitted as https://github.com/PowerShell/PowerShell/pull/5270, please review. Thanks

PR #5126, opened 2017-10-15
Add a new parameter -Depth to Split-Path command

kvprasoon:
This new parameter will allow users to mention the depth of the split operation.
This will reduce the usage of multiple Split-Path command to get the parent of parent or leaf of parent.

#Default value is 0

### DIFF
--- a/src/System.Management.Automation/engine/PathInterfaces.cs
+++ b/src/System.Management.Automation/engine/PathInterfaces.cs
@@ -1122,7 +1122,7 @@ namespace System.Management.Automation
         /// <exception cref="ProviderInvocationException">
         /// If the provider threw an exception.
         /// </exception>
-        public string ParseParent(string path, int depth, string root)
+        public string ParseParent(string path, string root)
         {
             Dbg.Diagnostics.Assert(
                 _sessionState != null,
@@ -1130,7 +1130,7 @@ namespace System.Management.Automation
 
             // Parameter validation is done in the session state object
 
-            return _sessionState.GetParentPath(path, depth, root);
+            return _sessionState.GetParentPath(path, root);
         } // GetParentPath
 
         /// <summary>
--- a/src/System.Management.Automation/engine/PathInterfaces.cs
+++ b/src/System.Management.Automation/engine/PathInterfaces.cs
@@ -1122,7 +1122,7 @@ namespace System.Management.Automation
         /// <exception cref="ProviderInvocationException">
         /// If the provider threw an exception.
         /// </exception>
-        public string ParseParent(string path, string root)
+        public string ParseParent(string path, int depth, string root)
         {
             Dbg.Diagnostics.Assert(
                 _sessionState != null,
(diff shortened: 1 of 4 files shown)


src/System.Management.Automation/engine/PathInterfaces.cs
```diff
@@ -1122,15 +1122,15 @@ internal string Combine(string parent, string child, CmdletProviderContext conte
         /// <exception cref="ProviderInvocationException">
         /// If the provider threw an exception.
         /// </exception>
-        public string ParseParent(string path, string root)
+        public string ParseParent(string path, int depth, string root)
```
iSazonov: We shouldn't change a public API. Please revert the change and add new overload.

src/System.Management.Automation/engine/SessionStateNavigation.cs
```diff
@@ -384,7 +386,7 @@ private string RemoveQualifier(string path, ProviderInfo provider, out string qu
                 "Caller should validate context before calling this method");
 
             CmdletProvider providerInstance = GetProviderInstance(provider);
-            return GetParentPath(providerInstance, path, root, context);
+            return GetParentPath(providerInstance,depth , path, root, context);
```
iSazonov: Formatting.

src/System.Management.Automation/namespaces/NavigationProviderBase.cs
```diff
@@ -492,7 +493,7 @@ protected string MakePath(string parent, string child, bool childIsLeaf)
         /// for the provider namespace. For example, the file system provider should look
         /// for the last "\" and return everything to the left of the "\".
         /// </remarks>
-        protected virtual string GetParentPath(string path, string root)
+        protected virtual string GetParentPath(string path,int depth, string root)
```
iSazonov: Formatting.
And maybe add new overload too.

iSazonov (commented):
@kvprasoon Please add tests.

src/System.Management.Automation/namespaces/NavigationProviderBase.cs
kvprasoon: Sorry, new to C#.
So I need to create a new overload for ParseParent() has to be created and need to format the invocation lines ?

iSazonov: @kvprasoon Please see docs https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/member-overloading
> DO make only the longest overload virtual (if extensibility is required). Shorter overloads should simply call through to a longer overload. 

So `GetParentPath(string path, string root)` should call `virtual GetParentPath(string path,int depth, string root)`

src/System.Management.Automation/engine/PathInterfaces.cs
TravisEz13: Otherwise, this is a breaking change...

iSazonov:
@kvprasoon  It seems you revert all changes.

kvprasoon:
Yes, I've reverted and will be making changes as per the guidelines .

iSazonov:
@kvprasoon If you're experiencing any problems feel free to ask!

Issue #3996, opened 2017-06-13
Inconsistent handling of native command stderr 


BurtHarris:
I discovered recently that the handling of `stderr` from native commands isn't being dealt with consistently between Windows PowerShell and Windows PowerShell ISE.     

Steps to reproduce
------------------

```powershell
$errorActionPreference="stop"; try { cmd /c nosuchexe } catch { 'dang!' }
```

Actual ISE behavior (expected)
-----------------

```none
dang!
```
Native commands writing to `stderr` while `$errorActionPreference="stop"` generate a catchable `NativeCommandError` which encapsulates the original error message in a `ErrorRecord`.   It is easy to capture, analyze, and reword error messages in context.

Actual CLI behavior
---------------

```none
'nosuchexe' is not recognized as an internal or external command,
operable program or batch file.
```

No `NativeCommandError` is generated, no matter how `$errorActionPreference` is set.  The `stderr` from the native command is piped straight through to the default process `stderr`.   

This behavior is a serious loss of functionality, leading some to believe that PowerShell can't distinguish between `stderr` and `stdout`.   Lots of community confusion generated because of the discrepency.   See https://stackoverflow.com/questions/44488202/powershell-streaming-output/44491420#44491420, among others.

Analysis
---------
The difference isn't just the ISE, I think it is related to PowerShell remote runspaces!   The full info on the caught error record (in the ISE) demonstrates the underlying exception is a RemoteException like this:

```powershell
PS C:\Users\Burt_> $errorActionPreference="stop"; try { cmd /c nosuchexe } catch { Format-List * -force -InputObject $_ }



PSMessageDetails      : 
Exception             : System.Management.Automation.RemoteException: 'nosuchexe' is not recognized as an internal or external 
                        command,
TargetObject          : 'nosuchexe' is not recognized as an internal or external command,
CategoryInfo          : NotSpecified: ('nosuchexe' is ...ternal command,:String) [], RemoteException
FullyQualifiedErrorId : NativeCommandError
ErrorDetails          : 
InvocationInfo        : System.Management.Automation.InvocationInfo
ScriptStackTrace      : at <ScriptBlock>, <No file>: line 1
PipelineIterationInfo : {}
```

Environment data
----------------
Both powershell.exe and powershell_ise.exe report the same version information:

```powershell
Name                           Value
----                           -----
PSVersion                      5.1.15063.296
PSEdition                      Desktop
[...]

PR #5127, opened 2017-10-15
StdErr/ErrorRecord should not be modified on output

SteveL-MSFT:
ErrorRecord output goes through ComplexWriter which splits lines based on console window width

Fix is to add a flag to FormatEntryData to know that it's an ErrorRecord and don't constrain
the output to window width

Existing Write-Error.Tests.ps1 was not formatted correctly, so added some indentation to make it easier to read while modifying the file.

Fix https://github.com/PowerShell/PowerShell/issues/3813

### DIFF
--- a/src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs
+++ b/src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs
@@ -565,7 +565,14 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             {
                 ComplexWriter complexWriter = new ComplexWriter();
 
-                complexWriter.Initialize(_lo, _lo.ColumnNumber);
+                if (fed.isErrorRecord)
+                {
+                    complexWriter.Initialize(_lo, int.MaxValue);
+                }
+                else
+                {
+                    complexWriter.Initialize(_lo, _lo.ColumnNumber);
+                }
                 complexWriter.WriteObject(cve.formatValueList);
 
                 return;
--- a/src/System.Management.Automation/commands/utility/FormatAndOutput/common/DefaultFormatters/PowerShellCore_format_ps1xml.cs
+++ b/src/System.Management.Automation/commands/utility/FormatAndOutput/common/DefaultFormatters/PowerShellCore_format_ps1xml.cs
@@ -789,7 +789,6 @@ namespace System.Management.Automation.Runspaces
                                         }
 
                                         $indent = 4
-                                        $width = $host.UI.RawUI.BufferSize.Width - $indent - 2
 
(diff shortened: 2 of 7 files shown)


lzybkr:
I think there are 2 issues here:

* Native command errors should be treated like a simple string - maybe we can add to `Microsoft.PowerShell.Commands.Internal.Format.DefaultScalarTypes`.
* Script errors (or any type really) need a way to say - let wrapping for this property happen downstream or not at all.

I wonder if fixing the first issue (assuming my suggestion works) is sufficient for now while we explore a less hacky fix for the second issue.

SteveL-MSFT:
@lzybkr currently, native command errors get wrapped in an ErrorRecord making this more complicated than it should be.  Are you suggesting adding ErrorRecord to DefaultScalarTypes so that it should (hypothetically) just be ToString()?

lzybkr:
Add `ErrorRecord` to `DefaultScalarTypes` is probably insufficient - we'd find the view and apply that before looking at `DefaultScalarTypes`.

It feels wrong to wrap those lines in an `ErrorRecord`. Maybe we should just stop doing that.

iSazonov:
Is `Out-Default` a single entry point for all the cases? If so we could add new parameter to turn on/off wrap. `Out-String` and `Out-File` already has `Width` parameter - for redirection operator we could call `Out-File -Width [int]::MaxValue`.

SteveL-MSFT:
@lzybkr Adding System.Management.Automation.ErrorRecord to DefaultScalarTypes is not sufficient as in FormatViewManager.cs:531, view is not null so it doesn't go through the DefaultScalarTypes check path.

SteveL-MSFT:
@lzybkr do you have other concerns with the current fix as-is?  I think the wrapping of stderr is covered by the other existing issues such as https://github.com/PowerShell/PowerShell/issues/3996

lzybkr:
My concern is - this change affects more than native commands, and if we stop wrapping stderr in an ErrorRecord, this PR isn't necessary.

lzybkr:
And I should add - it's too targeted - it doesn't solve the more general problem of wrapping when the output is to a file.

SteveL-MSFT:
This solves the specific problem of stderr output being broken when output is redirected to file (stdout already didn't have an issue) as well as any errorrecord having linebreaks inserted into the output.  I think we can have the general issue fix as a separate PR.

lzybkr:
I have two objections:

* Wrapping stderr in an ErrorRecord introduces undesirable side effects like this.
* Formatting wraps nearly all types when output to file - which we also don't want.

We could fix this issue by not wrapping stderr and not touch formatting, or you could generalize the fix for formatting and not add the knowledge of this specific type.

From my point of view, you're adding dead code to the formatting engine - I can see this code living well beyond the other issues getting addressed.

SteveL-MSFT:
I've re-titled the original issue to be more general.  I think the formatting wrapping is a separate issue which I'll open.

PR #2820, opened 2016-12-01
Empty key name converting to PSObject

kittholland:
In issue #1755 we determined that you could not create a PSObject with a
key name of an empty string. This is supported in JSON and in
Hashtables.

Simple change was to lower evaluations of name field from
'IsNullOrEmpty' to '== null'.

Simple test to confirm is:
[pscustomobject]@{'' = 'Test'}

Simple JSON to test which results in ths same error:
'{"": "Test"}' | ConvertFrom-Json

After these changes, both of these cases work and I am able to
succesfully Convert project.lock.json from JSON and back to JSON with no
data loss.

There are many references to the areas of code I have updated and I do not believe I can foresee all consequences however I have ran the full test suite locally and have not found any indication these changes are harmful, while I have confirmed that it addresses the issue we are seeing.

I would also at least like to submit the above two cases as tests for this issue, but I am uncertain where those tests should go.

PR #5129, opened 2017-10-16
allow pscustomobject to have property with empty string name

SteveL-MSFT:
Needed to allow convertfrom-json to support json objects that have empty string property names which apparently is used pretty often

Fix https://github.com/PowerShell/PowerShell/issues/1755


PetSerAl:
How about `ForEach-Object -MemberName ''` or `Add-Member -Name ''`?

SteveL-MSFT:
@PetSerAl great point, will update: ForEach-Object and *-Member cmdlets

SteveL-MSFT:
For now, I'm only going to add support for `Get-Member` and not `Add-Member` as we don't want to promote creation of members with empty string names.

lzybkr:
We considered this before and rejected it, see https://github.com/PowerShell/PowerShell/pull/2820

SteveL-MSFT:
Completely forgot about that discussion almost a year ago.  This is unfortunately not going to make it into 6.0.0 and we'll have to revisit in 6.1.0

PR #5145, opened 2017-10-17
Update packaging to only package PowerShell binaries when packaging symbols

TravisEz13:
disallow all other package types for symbols, as this package is no longer intended to be installed.

I run compliance tools on all binaries in this packages and we are getting more issues from the dependencies than from our product binaries.

### DIFF
--- a/tools/packaging/packaging.psm1
+++ b/tools/packaging/packaging.psm1
@@ -284,6 +284,7 @@ function Start-PSPackage {
 
         if($IncludeSymbols.IsPresent)
         {
+            # Source is a temporary folder when -IncludeSymbols is present.  So, we should remove it.
             Remove-Item -Path $Source -Recurse -Force -ErrorAction SilentlyContinue
         }
     }
--- a/tools/packaging/packaging.psm1
+++ b/tools/packaging/packaging.psm1
@@ -58,7 +58,7 @@ function Start-PSPackage {
             log 'setting IncludeSymbols'
             $IncludeSymbols = $PSBoundParameters['IncludeSymbols']
         }
-        log "$($IncludeSymbols.IsPresent):$IncludeSymbols"
+
         # Runtime and Configuration settings required by the package
         ($Runtime, $Configuration) = if ($WindowsRuntime) {
             $WindowsRuntime, "Release"
--- a/tools/packaging/packaging.psm1
+++ b/tools/packaging/packaging.psm1
@@ -30,8 +30,6 @@ function Start-PSPackage {
 
         [Switch] $Force,
 
-        [Switch] $IncludeSymbols,
-
         [Switch] $SkipReleaseChecks
     )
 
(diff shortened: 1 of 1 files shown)


TravisEz13:
https://github.com/PowerShell/PowerShell/pull/5145/files?w=1 make this a lot easier to review.

tools/packaging/packaging.psm1
```diff
@@ -30,226 +30,261 @@ function Start-PSPackage {
...
-            $params += '-CrossGen'
+        $crossGenCorrect = $false
+        if ($Type -eq "deb-arm") {
+            # crossgen doesn't support arm32 yet
+            $crossGenCorrect = $true
+        }
+        elseif(-not $IncludeSymbols.IsPresent -and $Script:Options.CrossGen) {
+            $crossGenCorrect = $true
+        }
+        elseif ($IncludeSymbols.IsPresent) {
```
daxian-dbw: So now it's OK to specify `-IncludeSymbols` while build with `-Crossgen`?

tools/packaging/packaging.psm1
```diff
@@ -30,226 +30,261 @@ function Start-PSPackage {
...
     }
 
-    log "Packaging RID: '$Runtime'; Packaging Configuration: '$Configuration'"
+    End {
+        $IncludeSymbols = $null
+        if ($PSBoundParameters.ContainsKey('IncludeSymbols')) {
+            log 'setting IncludeSymbols'
+            $IncludeSymbols = $PSBoundParameters['IncludeSymbols']
+        }
+        log "$($IncludeSymbols.IsPresent):$IncludeSymbols"
```
daxian-dbw: When `-IncludeSymbols` is not specified, this log will print just a colon.

tools/packaging/packaging.psm1
```diff
@@ -318,6 +353,19 @@ function New-TarballPackage {
     }
 }
 
+function New-TempFolder
+{
+    $tempPath = [System.IO.Path]::GetTempPath()
+
+    $tempFolder = Join-Path -Path $tempPath -ChildPath ([System.IO.Path]::GetRandomFileName())
```
daxian-dbw: How about replace these 2 lines with `[System.IO.Path]::GetTempFileName()`?

daxian-dbw (changes requested):

tools/packaging/packaging.psm1
```diff
@@ -30,226 +30,261 @@ function Start-PSPackage {
...
-                Version = $Version
-                Force = $Force
-            }
 
-            if ($PSCmdlet.ShouldProcess("Create $_ Package")) {
-                New-UnixPackage @Arguments
-            }
+        if($IncludeSymbols.IsPresent)
+        {
+            Remove-Item -Path $Source -Recurse -Force
```
adityapatwardhan: Maybe -ErrorAction SilentlyContinue?

adityapatwardhan (approved):
With minor comment

tools/packaging/packaging.psm1
TravisEz13: Yes, the non-publish builds always keep the symbols, unaltered.

tools/packaging/packaging.psm1
TravisEz13: I'll remove... this was troubleshooting...

tools/packaging/packaging.psm1
TravisEz13: This API creates the file.  I want a folder.

tools/packaging/packaging.psm1
TravisEz13: added

adityapatwardhan:
@daxian-dbw can you update your review?

tools/packaging/packaging.psm1
daxian-dbw: I know, but you can use that name to create a folder. The two lines of script shown here are doing the same thing.

daxian-dbw: I know, but you can use that name to create a folder. The two lines of script shown here are doing the same thing.

tools/packaging/packaging.psm1
```diff
@@ -30,226 +30,261 @@ function Start-PSPackage {
...
-                Version = $Version
-                Force = $Force
-            }
 
-            if ($PSCmdlet.ShouldProcess("Create $_ Package")) {
-                New-UnixPackage @Arguments
-            }
+        if($IncludeSymbols.IsPresent)
+        {
+            Remove-Item -Path $Source -Recurse -Force -ErrorAction SilentlyContinue
```
daxian-dbw: It would be great if you can put a comment here saying that "$Source is point to a temporary folder when -IncludeSymbols is present".

daxian-dbw (approved):
Approved with 2 comments.

tools/packaging/packaging.psm1
TravisEz13: I would have to delete the file before I can create a folder with the name.

daxian-dbw: I see your point. Yeah, please ignore this comment.

daxian-dbw (approved):
LGTM

Issue #4976, opened 2017-10-02
PowerShell on Linux package installation should set up a symbolic link to resolve libgssapi_krb5 to support client-side psrp over kerberos.


dantraMSFT:
Steps to reproduce
------------------
Install powershell on Ubuntu 16.04  with the latest build of the psrp client (1.3.0).  Domain join the machine/vm and attempt to connect to a windows machine using -Authentication Kerberos.

Expected behavior
-----------------
The connection succeeds.

Actual behavior
---------------
The connection fails with an access denied.

the problem occurs due to a PSRP's dynamic loading of libgssapi_krb5.  Since powershell itself has no explicit dependency, the binary cannot be resolved. Additionally, the while the dependency refers to libgssapi_krb5.so, current versions are .so.2 and .so.2.2.  The solution is to create a symbolic link to the installed binary as libgssapi_krb5.so to ensure it is resolved.

DHowett-MSFT:
This sounds like it should be on the maintainer of `libgssapi_krb5`. Specifically, it is the library owner's job to provide the non-soname symlink without the version number.

Unless we're shipping that in the PowerShell package, of course.

dantraMSFT:
@DHowett-MSFT You're probably right but I don't see how that addresses the existing installations.  Am I missing something?

dantraMSFT:
This is no longer an issue with OMI 1.4.0 and moving psrp to that version. As part of fixing https://github.com/PowerShell/psl-omi-provider/issues/104, I'll also be moving powershell's dependency to 1.4.0.

PR #5151, opened 2017-10-17
Create symbolic link to libgssapi_kbr5.so at build time in the publish directory

dantraMSFT:
Fixes https://github.com/PowerShell/PowerShell/issues/4976

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -545,7 +545,7 @@ Fix steps:
 
     if ($Environment.IsUbuntu -or $Environment.IsDebian)
     {
-        # libpsl-native.so v1.3 now depends on libgssapi_krb5 for Kerberos support.
+        # libpsrpclient.so v1.3 now depends on libgssapi_krb5 for Kerberos support.
         # Ubuntu 16.04 and Debian 8.* do not have this file or symbolic link which prevents
         # tests from running.  Create the symbolic link to work around it.
         # See tools\packaging.psm1 for similar behavior at packaging time.
--- a/build.psm1
+++ b/build.psm1
@@ -543,9 +543,10 @@ Fix steps:
         }
     }
 
-    if ($Environment.IsUbuntu)
+    if ($Environment.IsUbuntu -or $Environment.IsDebian)
     {
         # psrp 1.3* now depends on libgssapi_krb5 for Kerberos support.
+        # selected Ubuntu and Debian systems do not have a link for it.
         $null = New-Item -Force -ItemType SymbolicLink -Target "/usr/lib/x86_64-linux-gnu/libgssapi_krb5.so.2" -Path "$publishPath/libgssapi_krb5.so" -ErrorAction Stop
     }
 
--- a/build.psm1
+++ b/build.psm1
@@ -545,7 +545,7 @@ Fix steps:
 
     if ($Environment.IsUbuntu -or $Environment.IsDebian)
     {
-        # psrp 1.3* now depends on libgssapi_krb5 for Kerberos support.
+        # libpsl-native.so v1.3 now depends on libgssapi_krb5 for Kerberos support.
         # Ubuntu 16.04 and Debian 8.* do not have this file or symbolic link which prevents
         # tests from running.  Create the symbolic link to work around it.
         # See tools\packaging.psm1 for similar behavior at packaging time.
--- a/build.psm1
+++ b/build.psm1
@@ -543,6 +543,12 @@ Fix steps:
         }
     }
 
+    if ($Environment.IsUbuntu)
+    {
+        # psrp 1.3* now depends on libgssapi_krb5 for Kerberos support.
+        $null = New-Item -Force -ItemType SymbolicLink -Target "/usr/lib/x86_64-linux-gnu/libgssapi_krb5.so.2" -Path "$publishPath/libgssapi_krb5.so" -ErrorAction Stop
+    }
+
     # download modules from powershell gallery.
     #   - PowerShellGet, PackageManagement, Microsoft.PowerShell.Archive
     if($PSModuleRestore)
--- a/build.psm1
+++ b/build.psm1
@@ -546,7 +546,10 @@ Fix steps:
     if ($Environment.IsUbuntu -or $Environment.IsDebian)
     {
         # psrp 1.3* now depends on libgssapi_krb5 for Kerberos support.
-        # selected Ubuntu and Debian systems do not have a link for it.
+        # Ubuntu 16.04 and Debian 8.* do not have this file or symbolic link which prevents
+        # tests from running.  Create the symbolic link to work around it.
+        # See tools\packaging.psm1 for similar behavior at packaging time.
+        # NOTE: PSRP client version 1.4 'should' remove the need for this workaround.
         $null = New-Item -Force -ItemType SymbolicLink -Target "/usr/lib/x86_64-linux-gnu/libgssapi_krb5.so.2" -Path "$publishPath/libgssapi_krb5.so" -ErrorAction Stop
     }
 
--- a/test/powershell/Installer/UbuntuInstalled.Tests.ps1
+++ b/test/powershell/Installer/UbuntuInstalled.Tests.ps1
@@ -1,0 +1,20 @@
+Describe 'Ubuntu Post-Install Tests' -Tag 'CI' {
+    BeforeAll {
+        $skip = $IsWindows
+
+        if (-not $skip)
+        {
+            $LinuxInfo = Get-Content /etc/os-release -Raw | ConvertFrom-StringData
+            $IsUbuntu = $LinuxInfo.ID -match 'ubuntu'
+            $skip = -not $IsUbuntu
+        }
+    }
+
+    It "libgssapi_krb5.so symbolic link exists in the $PSHOME directory" -skip:$skip {
+        $expectedPath = Join-Path -Path $PSHOME -ChildPath 'libgssapi_krb5.so'
+
+        Test-Path -path $expectedPath | Should Be $true
+        $item = Get-Item -Path $expectedPath
+        $item.LinkType | Should Be 'SymbolicLink'
+    }
+}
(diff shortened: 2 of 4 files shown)


build.psm1
```diff
@@ -539,6 +539,12 @@ Fix steps:
         }
     }
 
+    if ($Environment.IsUbuntu)
+    {
+        # psrp 1.3* now depends on libgssapi_krb5 for Kerberos support.
+        $null = New-Item -Force -ItemType SymbolicLink -Target "/usr/lib/x86_64-linux-gnu/libgssapi_krb5.so.2" -Path "$publishPath/libgssapi_krb5.so" -ErrorAction Stop
```
iSazonov: `libgssapi_krb5.so.2` is already link. It seems a right place for the fix is PSRP client repo - it should depend on `libgssapi_krb5.so.2` not `libgssapi_krb5.so`.

dantraMSFT: At this point, I don't have a better fix...

I raised that question with the authors and for other platforms,  libgssapi_krb5.so or an explicit link is created when the component is installed. On Ubuntu, this is not the case.
The PSRP omi provider install already handles this. 

For client only, (i.e., PowerShell only) the binaries are delivered in a nuget package and the link is not created.

iSazonov: I see also above the lines with IsRedHatFamily for libssl.so.10.
But I still don't understand how such fixes work if today we create _single package_ for all Linux.

dantraMSFT: @iSazonov : We don't create a single package for all Linux; we create multiple packages.  This symlink is created at build time for any Ubuntu release but, in practice it is only required for 16.04. If we merge all Ubuntu into a single package, the logic will continue to work.
If we do somehow figure out how to create a single package for all Linux, this model will not work; we'll need to patch at post install time.  From what I've seen to date; that doesn't appear viable in the near term.

iSazonov: Thanks for clarify.
Closed.

iSazonov (approved):

build.psm1
```diff
@@ -539,6 +539,12 @@ Fix steps:
         }
     }
 
+    if ($Environment.IsUbuntu)
+    {
```
daxian-dbw: Is this symbolic link required on all Ubuntu versions? We support 14.04 16.04 and 17.04
Is it also required on `Debian 8/9`? We support `Debian 8/9`

build.psm1
daxian-dbw: The two symbolic links about `/lib64/libssl.so.10` are also created in `packaging.psm1`, see https://github.com/PowerShell/PowerShell/blob/master/tools/packaging/packaging.psm1#L466

Do you need to create this one there too? It seems to me we don't need to create in packaging.psm1, but not sure on that.

daxian-dbw (changes requested):

build.psm1
dantraMSFT: I don't know why their being created in packaging.psm1.  My understanding is creating them at build time in the publish directory ensures they are in the package and will be created at install time.

daxian-dbw: Maybe the two symbolic links about `/lib64/libssl.so.10` created in `packaging.psm1` should be removed?
When testing Linux PSRP, can you please create a powershell package and use the package to test? We will know for sure that it works end-to-end.

dantraMSFT: It turns out that the nuget package does have the symbolic link but the deb package did not. In the process of creating the staging directory, the actual file was copied. I've updated packaging.psm1 to address this.

daxian-dbw: > It turns out that the nuget package does have the symbolic link

Which nuget package are you talking about?

> I've updated packaging.psm1 to address this

Maybe you forget to push the commit that has this change?

dantraMSFT: Sorry, I didn't see this comment.  You're right, I hadn't pushed the change. It's there now.

dantraMSFT: @daxian-dbw  Do you have any addition comments?

build.psm1
TravisEz13: I don't see an answer to this question

dantraMSFT: @daxian-dbw I was not aware libgssapi_krb5 has an issue on Debian - is that true or are you refering to openssl?  Just want to clarify.

dantraMSFT: I checked with the OMI team.  Debian 9 does not need it. They don't have Debian 8 in their testing so it's unknown.

dantraMSFT: Debian 8 also needs the symbolic link.  I'll update the PR.

dantraMSFT: @daxian-dbw and @TravisEz13 - I've updated the PR to create the symbolic link on Debian.

tools/packaging/packaging.psm1
```diff
@@ -476,6 +476,15 @@ function New-UnixPackage {
...
                 $packagingStrings.UbuntuAfterRemoveScript -f "$Link/pwsh" | Out-File -FilePath $AfterRemoveScript -Encoding ascii
+                # Needed for Ubuntu 16.04 and Debian 8.*
+                # NOTE: Staging is ending up with the actual file instead of the symbolic link.
+                # Remove the file and create the link.
+                if (Test-Path -Path "$Staging/libgssapi_krb5.so")
+                {
+                    Remove-Item -Path "$Staging/libgssapi_krb5.so" -Force -ErrorAction SilentlyContinue
+                }
+                Write-Verbose -Message "Creating symbolic link $Staging/libgssapi_krb5.so"
+                New-Item -Force -ItemType SymbolicLink -Target "/usr/lib/x86_64-linux-gnu/libgssapi_krb5.so.2" -Path "$Staging/libgssapi_krb5.so" -ErrorAction Stop >$null
```
daxian-dbw: With "New-Item -Force", the existing "libgssapi_krb5.so" file will be overwritten with the symbolic link. So no need to remove it.

tools/packaging/packaging.psm1
```diff
@@ -476,6 +476,15 @@ function New-UnixPackage {
                 $AfterRemoveScript = [io.path]::GetTempFileName()
                 $packagingStrings.UbuntuAfterInstallScript -f "$Link/pwsh" | Out-File -FilePath $AfterInstallScript -Encoding ascii
                 $packagingStrings.UbuntuAfterRemoveScript -f "$Link/pwsh" | Out-File -FilePath $AfterRemoveScript -Encoding ascii
+                # Needed for Ubuntu 16.04 and Debian 8.*
+                # NOTE: Staging is ending up with the actual file instead of the symbolic link.
+                # Remove the file and create the link.
```
daxian-dbw: Need to update the comment to explain what binary is requiring this symbolic creation. Please see the comment for creating symlink for "libssl.so" above as an example.

test/powershell/Installer/UbuntuInstalled.Tests.ps1
```diff
@@ -0,0 +1,20 @@
+Describe 'Ubuntu Post-Install Tests' -Tag 'CI' {
+    BeforeAll {
+        $skip = $IsWindows
+
+        if (-not $skip)
+        {
+            $LinuxInfo = Get-Content /etc/os-release -Raw | ConvertFrom-StringData
+            $IsUbuntu = $LinuxInfo.ID -match 'ubuntu'
```
daxian-dbw: Need to also include debian now since we always create the symlink.

build.psm1
```diff
@@ -543,6 +543,13 @@ Fix steps:
         }
     }
 
+    if ($Environment.IsUbuntu -or $Environment.IsDebian)
+    {
+        # psrp 1.3* now depends on libgssapi_krb5 for Kerberos support.
+        # selected Ubuntu and Debian systems do not have a link for it.
```
daxian-dbw: > selected Ubuntu

what is the "selected" ubuntu and debian system? Do you mean all ubuntu/debian distros we officially support? Can you please update the comment to make it a bit clear?

daxian-dbw (changes requested):

tools/packaging/packaging.psm1
dantraMSFT: Fixed.

tools/packaging/packaging.psm1
dantraMSFT: I updated the comments in both packaging.psm1 and build.psm1.

tools/packaging/packaging.psm1
dantraMSFT: Fixed

tools/packaging/packaging.psm1
dantraMSFT: fixed

build.psm1
```diff
@@ -545,7 +545,7 @@ Fix steps:
 
     if ($Environment.IsUbuntu -or $Environment.IsDebian)
     {
-        # psrp 1.3* now depends on libgssapi_krb5 for Kerberos support.
+        # libpsl-native.so v1.3 now depends on libgssapi_krb5 for Kerberos support.
```
daxian-dbw: `libpsl-native.so` is the binary powershell builds, not related to PSRP.

dantraMSFT: fixed

test/powershell/Installer/UbuntuInstalled.Tests.ps1
daxian-dbw: @dantraMSFT Could you please enable this test on Debian as well. Jim runs tests on all supported distros before release.

daxian-dbw:
Look mostly good except for a comment in the test.

test/powershell/Installer/UbuntuInstalled.Tests.ps1
dantraMSFT: Fixed. I also renamed the test file to be linux generic.

dantraMSFT:
My testing this weekend of a 1.4.0 build shows that this fix is no longer needed on the client side.  I'm closing the PR and will close the issue.

Issue #5084, opened 2017-10-11
Hosted PS core in ASP.Net core The system cannot find the file specified


Qowy:
ASP.NET core 2.0 app.

.csproj:
```
    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.0" />
    <PackageReference Include="Microsoft.EntityFrameworkCore.Tools" Version="2.0.0" PrivateAssets="All" />
    <PackageReference Include="Microsoft.PowerShell.Commands.Diagnostics" Version="6.0.0-beta.8" />
    <PackageReference Include="Microsoft.PowerShell.SDK" Version="6.0.0-beta.8" />
    <PackageReference Include="Microsoft.VisualStudio.Web.CodeGeneration.Design" Version="2.0.0" PrivateAssets="All" />
    <PackageReference Include="Microsoft.WSMan.Management" Version="6.0.0-beta.8" />
```

controller code:
```
    using (PowerShell ps = PowerShell.Create())
            {
                var result = ps.AddScript(@"New-Item -Path 'C:\Programming\' -Type Directory -Name TestFolder").Invoke();
            }
```

Results in an exeption:

> System.Management.Automation.RuntimeException: The following errors occurred when updating the assembly list for the runspace: 
> Could not load file or assembly 'C:\Programming\Intranet\Intranet\Microsoft.PowerShell.Commands'. The system cannot find the file specified.
> Could not load file or assembly 'C:\Programming\Intranet\Intranet\Microsoft.PowerShell'. The system cannot find the file specified.
> Could not load file or assembly 'C:\Programming\Intranet\Intranet\Microsoft.PowerShell.Commands'. The system cannot find the file specified.
> Could not load file or assembly 'C:\Programming\Intranet\Intranet\Microsoft.PowerShell.Commands'. The system cannot find the file specified.
> Could not load file or assembly 'C:\Programming\Intranet\Intranet\Microsoft.PowerShell'. The system cannot find the file specified.
> Could not load file or assembly 'C:\Programming\Intranet\Intranet\Microsoft.WSMan'. The system cannot find the file specified..
> 
>    at System.Management.Automation.ExecutionContext.UpdateAssemblyCache()
> 
>    at System.Management.Automation.Runspaces.RunspaceConfigurationEntryCollection`1.Update(Boolean force)
> 
>    at System.Management.Automation.Runspaces.RunspaceConfiguration.Bind(ExecutionContext executionContext)
> 
[...]

PR #5161, opened 2017-10-19
Update PS should try to load the assembly from file path first before…

chunqingchen:
fix issue #3325

After we merge #4196, the fix will break some our partner's working code. Although the working code often works with an invalid assembly path.

This fix is to follow the old logic to minimize the breaking change.

1. we first try to load the assembly from the file path
2. if it fails, we try to load the assembly from the assembly name
3. if 2 passes, we ignore the exception and return the assembly. otherwise we throw the exception.

### DIFF
--- a/src/System.Management.Automation/engine/ExecutionContext.cs
+++ b/src/System.Management.Automation/engine/ExecutionContext.cs
@@ -1196,21 +1196,19 @@ namespace System.Management.Automation
             }
         }
 
-
         [SuppressMessage("Microsoft.Reliability", "CA2001:AvoidCallingProblematicMethods", MessageId = "System.Reflection.Assembly.LoadWithPartialName")]
         [SuppressMessage("Microsoft.Reliability", "CA2001:AvoidCallingProblematicMethods", MessageId = "System.Reflection.Assembly.LoadFrom")]
         internal static Assembly LoadAssembly(string name, string filename, out Exception error)
         {
             // First we try to load the assembly based on the filename
-
             Assembly loadedAssembly = null;
             error = null;
-
             if (!String.IsNullOrEmpty(filename))
             {
                 try
                 {
                     loadedAssembly = Assembly.LoadFrom(filename);
+                    return loadedAssembly;
                 }
                 catch (FileNotFoundException fileNotFound)
                 {
@@ -1219,19 +1217,22 @@ namespace System.Management.Automation
                 catch (FileLoadException fileLoadException)
                 {
                     error = fileLoadException;
+                    return null;
                 }
                 catch (BadImageFormatException badImage)
                 {
                     error = badImage;
+                    return null;
                 }
                 catch (SecurityException securityException)
                 {
                     error = securityException;
+                    return null;
                 }
             }
-            
-            // if loading from filename fails, we will try to load from assembly name again to follow the previous code logic, this avoid breaking partner's existed code.
-            if (loadedAssembly == null && !String.IsNullOrEmpty(name))
+
+            // Then we try to load the assembly based on the given name
+            if (!String.IsNullOrEmpty(name))
             {
                 string fixedName = null;
                 // Remove the '.dll' if it's there...
@@ -1247,45 +1248,33 @@ namespace System.Management.Automation
                 {
                     loadedAssembly = Assembly.Load(new AssemblyName(assemblyString));
                 }
-                // if the filename exists and error has been created, we would try NOT to override the error.
                 catch (FileNotFoundException fileNotFound)
                 {
-                    if (error != null)
-                    {
-                        error = fileNotFound;
-                    }
+                    error = fileNotFound;
                 }
                 catch (FileLoadException fileLoadException)
                 {
-                    if (error != null)
-                    {
-                        error = fileLoadException;
-                    }
+                    error = fileLoadException;
                     // this is a legitimate error on CoreCLR for a newly emited with Add-Type assemblies
                     // they cannot be loaded by name, but we are only interested in importing them by path
                 }
                 catch (BadImageFormatException badImage)
                 {
-                    if (error != null)
-                    {
-                        error = badImage;
-                    }
+                    error = badImage;
+                    return null;
                 }
                 catch (SecurityException securityException)
                 {
-                    if (error != null)
-                    {
-                        error = securityException;
-                    }
+                    error = securityException;
+                    return null;
                 }
             }
 
-            // the error is coming from loading assembly from filename, since we load the assembly from assembly name later, ignore the exception.
+            // If the assembly is loaded, we ignore error as it may come from the filepath loading.        
             if (loadedAssembly != null)
             {
                 error = null;
             }
-            // We either return the loaded Assembly, or return null.
             return loadedAssembly;
         }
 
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1229,7 +1229,9 @@ internal static Assembly LoadAssembly(string name, string filename, out Exceptio
                     error = securityException;
                 }
             }
-            else if (!String.IsNullOrEmpty(name))
+
```
daxian-dbw: How about we check if the file exists before trying `Assembly.LoadFrom`? So maybe just change `if (!String.IsNullOrEmpty(filename))` to `if (File.Exists(filename))`?

daxian-dbw (changes requested):

chunqingchen:
@daxian-dbw your comments are resolved

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1258,14 +1261,20 @@ internal static Assembly LoadAssembly(string name, string filename, out Exceptio
                 catch (BadImageFormatException badImage)
                 {
                     error = badImage;
+                    return null;
```
daxian-dbw: The `return null` at this line and line 1269 are not needed.

src/System.Management.Automation/engine/ExecutionContext.cs
```diff
@@ -1258,14 +1261,20 @@ internal static Assembly LoadAssembly(string name, string filename, out Exceptio
                 catch (BadImageFormatException badImage)
```
daxian-dbw: The comments in the `catch (FileLoadException fileLoadException)` block right above this line seems not needed anymore, since we now try loading from path first.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -122,5 +122,14 @@ namespace ModuleCmdlets
     $results[0] | Should Be $path
     $results[1] | Should BeExactly "BinaryModuleCmdlet1 exported by the ModuleCmdlets module."
     }
+
+    It "PS should try to load the assembly from assembly name if file path doesn't exist" -skip:(!$IsWindows) {
```
daxian-dbw: Why is this test skipped on Unix plats?

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -122,5 +122,14 @@ namespace ModuleCmdlets
...
+
+        New-ModuleManifest -Path $TESTDRIVE\test.psd1 -NestedModules \NOExistedPath\System.Management.Automation.dll
+        $module = Import-Module $TESTDRIVE\test.psd1 -PassThru
+        $module.NestedModules | Should Not BeNullOrEmpty
+        $assemblyLocation = [psobject].Assembly.location
+        $module.NestedModules.ImplementingAssembly.Location | Should Be $assemblyLocation
```
daxian-dbw: Please unload the module after the test.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/ExecutionContext.cs
chunqingchen: resolved

src/System.Management.Automation/engine/ExecutionContext.cs
chunqingchen: resolved

chunqingchen:
@daxian-dbw your comment is resolved

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -122,5 +122,29 @@ namespace ModuleCmdlets
...
+    It "PS should try to load the assembly from assembly name if file path doesn't exist" {
+
+        if ($IsWindows)
+        {
+            New-ModuleManifest -Path $TESTDRIVE\test.psd1 -NestedModules \NOExistedPath\System.Management.Automation.dll
+        }
+        else
+        {
+            New-ModuleManifest -Path $TESTDRIVE/test.psd1 -NestedModules /NOExistedPath/System.Management.Automation.dll
+        }
```
daxian-dbw: Can you use `Join-Path` to avoid the `if ($IsWindows) else` blocks? For example,
```
$psdFile = Join-Path $TESTDRIVE test.psd1
$nestedModule = Join-Path NOExistedPath System.Management.Automation.dll
New-ModuleManifest -Path $psdFile -NestedModules $nestedModule 
```

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -122,5 +122,29 @@ namespace ModuleCmdlets
...
+        try
+        {
+            $module = Import-Module $TESTDRIVE\test.psd1 -PassThru
+            $module.NestedModules | Should Not BeNullOrEmpty
+            $assemblyLocation = [psobject].Assembly.location
+            $module.NestedModules.ImplementingAssembly.Location | Should Be $assemblyLocation
+        }
+        finally
+        {
+            Remove-Module  $TESTDRIVE\test.psd1 -ErrorAction SilentlyContinue
```
daxian-dbw: `Remove-Module $module -ErrorAction SilentlyContinue` would do

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -122,5 +122,29 @@ namespace ModuleCmdlets
...
+        {
+            New-ModuleManifest -Path $TESTDRIVE\test.psd1 -NestedModules \NOExistedPath\System.Management.Automation.dll
+        }
+        else
+        {
+            New-ModuleManifest -Path $TESTDRIVE/test.psd1 -NestedModules /NOExistedPath/System.Management.Automation.dll
+        }
+        try
+        {
+            $module = Import-Module $TESTDRIVE\test.psd1 -PassThru
```
daxian-dbw: Once you have `$psdFil` replace the `$TESTDRIVE\test.psd1` with `$psdFil`.

daxian-dbw (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
chunqingchen: resolved

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
chunqingchen: resolved

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
chunqingchen: resolved

chunqingchen:
@daxian-dbw your comment is resolved

daxian-dbw (approved):
LGTM. Can you please add `[Feature]` tag to the commit message of your last commit so that we can validate the fix with a full test run?

daxian-dbw:
@chunqingchen I have assigned #5084 to you so you can verify whether this PR fixes that issue.

PR #5164, opened 2017-10-19
Make -NoTypeInformation Default on Export-Csv and ConvertTo-Csv

### LINKED ISSUE #5131: -NoTypeInformation should be default on Export-Csv
In most situations, the default behavior of `Export-Csv` is undesirable. Its usage as an object Import and Export is limited to begin with. What it is more commonly used for (in my experience, anyway) is to produce CSV files either to be  parsed by other automated processes outside the PowerShell ecosystem or to be used in Excel as human parsable reports. The initial line with the type information is undesirable in those situations. 

This is a common "gotcha" for PowerShell novices. It violates the Principle of least astonishment.

I'm curious how many profiles have this `$PSDefaultParameterValues['Export-CSV:NoTypeInformation'] = $true`.  I know I have that in my work profile and at the top of so many scripts.

It's a minor annoyance and a breaking change, but I just don't think the default behavior is actually useful.

My tentative proposal is to leave `-NoTypeInformation` in for backwards compatibility but make it default and add a `-IncludeTypeInformation` switch to revert to legacy behavior per request.

This issue is for discussion, yays, nays, and possible alternatives for a proposed solution.

markekraus:
closes #5131

* Sets `-NoTypeInformation` as the default behavior for `Export-Csv` and `ConvertTo-Csv`
* Hides the `-NoTypeInformation` parameter switch
* Adds `-IncludeTypeInformation` switch to `Export-Csv` and `ConvertTo-Csv` to enable legacy behavior
* Provides a terminating error when both `-NoTypeInformation` and `-IncludeTypeInformation` are supplied
* adds tests for the new behavior
* fixes existing tests to align with new behavior

Breaking change was approved by committee in #5131 

The new behavior will need to be documented.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
@@ -96,7 +96,7 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         protected override void BeginProcessing()
         {
-            if (this.MyInvocation.BoundParameters.ContainsKey("IncludeTypeInformation") && this.MyInvocation.BoundParameters.ContainsKey("NoTypeInformation"))
+            if (this.MyInvocation.BoundParameters.ContainsKey(nameof(IncludeTypeInformation)) && this.MyInvocation.BoundParameters.ContainsKey(nameof(NoTypeInformation)))
             {
                 InvalidOperationException exception = new InvalidOperationException(CsvCommandStrings.CannotSpecifyIncludeTypeInformationAndNoTypeInformation);
                 ErrorRecord errorRecord = new ErrorRecord(exception, "CannotSpecifyIncludeTypeInformationAndNoTypeInformation", ErrorCategory.InvalidData, null);
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
@@ -67,9 +67,29 @@ namespace Microsoft.PowerShell.Commands
         }
 
         /// <summary>
-        /// NoTypeInformation : should the #TYPE line be generated
+        /// IncludeTypeInformation : The #TYPE line should be generated. Default is false. Cannot specify with NoTypeInformation.
         /// </summary>
         [Parameter]
+        [Alias("ITI")]
+        public SwitchParameter IncludeTypeInformation
+        {
+            get
+            {
+                return _includeTypeInformation;
+            }
+            set
+            {
+                _includeTypeInformation = value;
+                _includeTypeInformationIsSet = true;
+            }
+        }
+        private bool _includeTypeInformation;
+        private bool _includeTypeInformationIsSet;
+
+        /// <summary>
+        /// NoTypeInformation : The #TYPE line should not be generated. Default is true. Cannot specify with IncludeTypeInformation.
+        /// </summary>
+        [Parameter(DontShow = true)]
         [Alias("NTI")]
         public SwitchParameter NoTypeInformation
         {
@@ -80,9 +100,11 @@ namespace Microsoft.PowerShell.Commands
             set
             {
                 _noTypeInformation = value;
+                _noTypeInformationIsSet = true;
             }
         }
-        private bool _noTypeInformation;
+        private bool _noTypeInformation = true;
+        private bool _noTypeInformationIsSet;
 
         #endregion Command Line Parameters
 
@@ -100,6 +122,16 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         protected override void BeginProcessing()
         {
+            if (_noTypeInformationIsSet && _includeTypeInformationIsSet)
+            {
+                InvalidOperationException exception = new InvalidOperationException(CsvCommandStrings.CannotSpecifyIncludeTypeInformationAndNoTypeInformation);
+                ErrorRecord errorRecord = new ErrorRecord(exception, "CannotSpecifyIncludeTypeInformationAndNoTypeInformation", ErrorCategory.InvalidData, null);
+                this.ThrowTerminatingError(errorRecord);
+            }
+            if (_includeTypeInformationIsSet)
+            {
+                _noTypeInformation = !_includeTypeInformation;
+            }
             _delimiter = ImportExportCSVHelper.SetDelimiter(this, ParameterSetName, _delimiter, UseCulture);
         }
     }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
@@ -71,40 +71,14 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         [Parameter]
         [Alias("ITI")]
-        public SwitchParameter IncludeTypeInformation
-        {
-            get
-            {
-                return _includeTypeInformation;
-            }
-            set
-            {
-                _includeTypeInformation = value;
-                _includeTypeInformationIsSet = true;
-            }
-        }
-        private bool _includeTypeInformation;
-        private bool _includeTypeInformationIsSet;
+        public SwitchParameter IncludeTypeInformation { get; set; }
 
         /// <summary>
         /// NoTypeInformation : The #TYPE line should not be generated. Default is true. Cannot specify with IncludeTypeInformation.
         /// </summary>
         [Parameter(DontShow = true)]
         [Alias("NTI")]
-        public SwitchParameter NoTypeInformation
-        {
-            get
-            {
-                return _noTypeInformation;
-            }
-            set
-            {
-                _noTypeInformation = value;
-                _noTypeInformationIsSet = true;
-            }
-        }
-        private bool _noTypeInformation = true;
-        private bool _noTypeInformationIsSet;
+        public SwitchParameter NoTypeInformation { get; set; } = true;
 
         #endregion Command Line Parameters
 
@@ -122,15 +96,15 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         protected override void BeginProcessing()
         {
-            if (_noTypeInformationIsSet && _includeTypeInformationIsSet)
+            if (this.MyInvocation.BoundParameters.ContainsKey("IncludeTypeInformation") && this.MyInvocation.BoundParameters.ContainsKey("NoTypeInformation"))
             {
                 InvalidOperationException exception = new InvalidOperationException(CsvCommandStrings.CannotSpecifyIncludeTypeInformationAndNoTypeInformation);
                 ErrorRecord errorRecord = new ErrorRecord(exception, "CannotSpecifyIncludeTypeInformationAndNoTypeInformation", ErrorCategory.InvalidData, null);
                 this.ThrowTerminatingError(errorRecord);
             }
-            if (_includeTypeInformationIsSet)
+            if (this.MyInvocation.BoundParameters.ContainsKey("IncludeTypeInformation"))
             {
-                _noTypeInformation = !_includeTypeInformation;
+                NoTypeInformation = !IncludeTypeInformation;
             }
             _delimiter = ImportExportCSVHelper.SetDelimiter(this, ParameterSetName, _delimiter, UseCulture);
         }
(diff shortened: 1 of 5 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -67,9 +67,29 @@ public char Delimiter
...
+                return _includeTypeInformation;
+            }
+            set
+            {
+                _includeTypeInformation = value;
+                _includeTypeInformationIsSet = true;
+            }
+        }
+        private bool _includeTypeInformation;
+        private bool _includeTypeInformationIsSet;
```
iSazonov: In ProcessRecord() we use only `if (NoTypeInformation == false)`.
So I believe we should use auto-implemented property for `IncludeTypeInformation` and `NoTypeInformation` and check `IncludeTypeInformation.IsPresent`.

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Csv.Tests.ps1
```diff
@@ -45,27 +45,44 @@ Describe "ConvertTo-Csv" -Tags "CI" {
     }
 
     It "Should output an array of objects" {
-        $result = $testObject | ConvertTo-Csv
+        $result = $testObject | ConvertTo-Csv -IncludeTypeInformation
```
iSazonov: I believe we shouldn't add `-IncludeTypeInformation` in tests where we don't check ""#TYPE ".

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Csv.Tests.ps1
```diff
@@ -45,27 +45,44 @@ Describe "ConvertTo-Csv" -Tags "CI" {
...
-	$result = $testObject | ConvertTo-Csv
+	$result = $testObject | ConvertTo-Csv -IncludeTypeInformation
 	$result[2] | Should Match "`"Hello`""
 	$result[2] | Should Match "`"World`""
     }
 
+    It "Includes type information when -IncludeTypeInformation is supplied" {
+        $result = $testObject | ConvertTo-Csv -IncludeTypeInformation
+
+        ($result -split ([Environment]::NewLine))[0] | Should BeExactly "#TYPE System.Management.Automation.PSCustomObject"
```
iSazonov: I think we need add reverse test to check that ""#TYPE " is absent if `-IncludeTypeInformation` is absent.
Also that `-NoTypeInformation` do nothing.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
markekraus: I did testing with that.. and... it resulted in conflicts when using @splatting and supplying both. :( I ultimately went with this method because it was the most reliable way to detect if both were supplied.

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Csv.Tests.ps1
markekraus: Ah, I miss understood what this one was testing. I will fix it.

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Csv.Tests.ps1
markekraus: sure I'll add those and do the same for the `Export-Csv` tests

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
iSazonov: As for @splatting - maybe it is a bug in @splatting ?
I think we should remove ` NoTypeInformation` everywhere in code and leave only its declaration - the parameter is dummy and we should ignore it in code at all.

markekraus: It's not a bug in splatting. Its the nature of how switch params work. `IsPresent` is does not mean `was supplied` but more `user supplied $true`

```powershell
Function Test-SwitchParam {
    param(
        [Switch]$Switch
    )
    
    "Switch: {0}" -f $Switch
    "Switch.IsPresent: {0}" -f $Switch.IsPresent
    "  "
}

'Test-SwitchParam'
Test-SwitchParam
'Test-SwitchParam -Switch'
Test-SwitchParam -Switch
'Test-SwitchParam -Switch:$true'
Test-SwitchParam -Switch:$true
'Test-SwitchParam -Switch:$false'
Test-SwitchParam -Switch:$false
'$Params = @{Switch = $true}; Test-SwitchParam @Params'
$Params = @{Switch = $true}; Test-SwitchParam @Params
'$Params = @{Switch = $false}; Test-SwitchParam @Params'
$Params = @{Switch = $false}; Test-SwitchParam @Params
```

```none
Test-SwitchParam
Switch: False
Switch.IsPresent: False
  
Test-SwitchParam -Switch
Switch: True
Switch.IsPresent: True
  
Test-SwitchParam -Switch:$true
Switch: True
Switch.IsPresent: True
  
Test-SwitchParam -Switch:$false
Switch: False
Switch.IsPresent: False
  
$Params = @{Switch = $true}; Test-SwitchParam @Params
Switch: True
Switch.IsPresent: True
  
$Params = @{Switch = $false}; Test-SwitchParam @Params
Switch: False
Switch.IsPresent: False
```

> I think we should remove `NoTypeInformation` everywhere in code and leave only its declaration - the parameter is dummy and we should ignore it in code at all.

I'm torn on that. It's still an active parameter and setting it still has impact and will continue to do so. It's just hidden in favor of the new switch to make the new default behavior more clear. The internal logic would all need to be reworked. I didn't see value in switching all the logic to use `IncludeTypeInformation` when all that needs to be done is the upfront logic of detecting if both were supplied and setting `NoTypeInformation` based on the presence and setting of `-IncludeTypeInformation`. The logic would still need to be there, so why disturb the rest of the code for it?

iSazonov: I see - somebody can use ` -NoTypeInformation:$false`.
However, I think we should make the code simpler and more readable. Currently  `-IncludeTypeInformation:$false` don't work. I vote for removing many private variables and using `IsPresent`.

markekraus: > Currently -IncludeTypeInformation:$false don't work

It does currently work, that's not new functionality. The decision by the committee was to maintain that functionality. This is how we preserve that functionality while adding a new parameter that is basically a reverse switch for the existing one. You can't move to `IsPresent` without breaking current functionality.

markekraus: @iSazonov I think I can make us both happy using `this.MyInvocation.BoundParameters.ContainsKey()`. I'll work on it.

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Csv.Tests.ps1
markekraus: fixed

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Csv.Tests.ps1
markekraus: fixed

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
markekraus: fixed

iSazonov: After nice walking and good dinner I was ready to agree with you 😄

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -100,6 +96,16 @@ public virtual void WriteCsvLine(string line)
         /// </summary>
         protected override void BeginProcessing()
         {
+            if (this.MyInvocation.BoundParameters.ContainsKey("IncludeTypeInformation") && this.MyInvocation.BoundParameters.ContainsKey("NoTypeInformation"))
```
iSazonov: Interesting, can we replace the _string_ names with strong typed names?

What about different parameter sets?

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Csv.Tests.ps1
```diff
@@ -50,22 +50,47 @@ Describe "ConvertTo-Csv" -Tags "CI" {
     }
 
     It "Should return the type of data in the first element of the output array" {
-	$result = $testObject | ConvertTo-Csv
+	$result = $testObject | ConvertTo-Csv -IncludeTypeInformation
 
 	$result[0] | Should Be "#TYPE System.Management.Automation.PSCustomObject"
     }
 
     It "Should return the column info in the second element of the output array" {
```
iSazonov: We could remove the test - it is dup test in line 28.
Or it is better move the checks and next test to the test in line 28 - there we can check a type, a header and data.

It is minor comment.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
markekraus: >  can we replace the string names with strong typed names?

I'm not sure I understand what you are asking. `IncludeTypeInformation.GetType().Name` instead of `"IncludeTypeInformation"`?

> What about different parameter sets?

parameter sets wont work without creating a bunch more. The parameter sets are currently being used for culture vs delimiter. It would get messy and harder to maintain than a logic check, IMO.

SteveL-MSFT: In the future, we need to revisit how to make defining and reading mutual exclusion of parameters easier.  Using multiple parametersets today can be quite complex to get right.  For now, having a runtime check is ok.

markekraus: @SteveL-MSFT I created #5175 to track/discuss mutually exclusive parameters form a general position.

iSazonov: I meant something like `nameof(param1)`.

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Csv.Tests.ps1
markekraus: @iSazonov Which test do you want to remove? the one you comment is on (lin 58) is not a duplicate of any of the tests. Line 28 is a delimiter test which needs to remain as its a different logic gate (parameter set).

the test on line 52 is now redundant with the the test on line 84, though.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
markekraus: @iSazonov thanks. fixed.

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Csv.Tests.ps1
iSazonov: It seems we could try to refactor the tests but not in the PR.
So skip the minor comment.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -96,7 +96,7 @@ public virtual void WriteCsvLine(string line)
         /// </summary>
         protected override void BeginProcessing()
         {
-            if (this.MyInvocation.BoundParameters.ContainsKey("IncludeTypeInformation") && this.MyInvocation.BoundParameters.ContainsKey("NoTypeInformation"))
+            if (this.MyInvocation.BoundParameters.ContainsKey(nameof(IncludeTypeInformation)) && this.MyInvocation.BoundParameters.ContainsKey(nameof(NoTypeInformation)))
```
iSazonov: Can we use different parameter sets to exclude the code?

markekraus: @iSazonov we already discussed that here 

https://github.com/PowerShell/PowerShell/pull/5164#discussion_r145791740

The parameter sets are currently being used for delimiter vs culture. That would require we add 2 more parameter sets. It would make the code harder to maintain. The parameter sets would be confusing to the user because we are hiding `-NoTypeInformation` with `DontShow=true`. @SteveL-MSFT  agreed that the runtime check was fine for now until we could address the larger general problem of mutex parameters. which prompted issue #5175

iSazonov: Sorry, I skipped this.

iSazonov (approved):

SteveL-MSFT (approved):

PR #5172, opened 2017-10-19
Update appimage.sh to reflect the new name 'pwsh'

daxian-dbw:
This is to fix the Travis CI build.

### DIFF
--- a/tools/appimage.sh
+++ b/tools/appimage.sh
@@ -35,8 +35,8 @@
 APP=powershell
 
 # Generate status file for use by apt-get; assuming that the recipe uses no newer
-# ingredients than what would require more recent dependencies than what we assume
-# to be part of the base system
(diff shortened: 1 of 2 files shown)


tools/appimage.sh
```diff
@@ -385,10 +385,6 @@ cd ./$APP/
 # but not using the host system's information about what is
 # installed on the system but our own assumptions instead
 
-mkdir -p ./tmp/archives/
-mkdir -p ./tmp/lists/partial
-touch tmp/pkgcache.bin tmp/srcpkgcache.bin
```
daxian-dbw: In case you wonder why they are removed -- same operations are done in the `generate_status` function called below.

TravisEz13 (approved):

adityapatwardhan (approved):
LGTM

daxian-dbw:
Thanks for the review!

PR #5173, opened 2017-10-19
Update the man help file used in packaging

daxian-dbw:
Update the man help file to reflect the name change.
I only changed the executable name. I think we should continue to use `PowerShell` in the rest of file

After this change, you will get `No manual entry for powershell` when running `man powershell`.
This is how it looks like when running `man pwsh`
image


SteveL-MSFT (approved):

.gitignore
```diff
@@ -50,9 +50,6 @@ DELETE_ME_TO_DISABLE_CONSOLEHOST_TELEMETRY
 # default location for produced nuget packages
 /nuget-artifacts
 
-# generated man files
-/assets/powershell.1*
-
 # resgen output
```
adityapatwardhan: Do we need /assets/pwsh.1* here?

daxian-dbw: We don't need it. `New-UnixPackage` command will clean up the generated `pwsh.1.gz` file.

adityapatwardhan (approved):
LGTM

iSazonov:
We should fix:
```
 powershell /h

Usage: powershell[.exe] [[-File] <filePath> [args]]
                        [-Command { - | <script-block> [-args <arg-array>]
                                      | <string> [<CommandParameters>] } ]
                        [-ConfigurationName <string>] [-EncodedCommand <Base64EncodedCommand>]
                        [-ExecutionPolicy <ExecutionPolicy>] [-InputFormat {Text | XML}]
                        [-Interactive] [-NoExit] [-NoLogo] [-NonInteractive] [-NoProfile]
                        [-OutputFormat {Text | XML}] [-Version] [-WindowStyle <style>]

       powershell[.exe] -h | -Help | -? | /?
```

daxian-dbw:
Good catch. I will submit a PR shortly.

Issue #5166, opened 2017-10-19
Please Start Versioning PowerShell.exe (So The latest version can be found by Get-Command)


DarwinJS:
Can't do the following from openssh installer to find the latest powershell core.

get-command "c:\program files\PowerShell\*\PowerShell.exe"

So the weird thing is to do the same thing for *any other default shell candidate* from third parties - this would likely work - to, for instance, find the latest bash.exe

I can do a work around by using the versioned folder for now - but that only handles the default install location where the methodology of a versioned folder is followed by convention.  If someone drops it in another folder or I try get-command against the file unpathed - I won't find a version.

There are obviously a lot of other reasons this should be done - just giving light to a very practical and current scenario.

iSazonov:
It seems dup #4836 - we can discuss there.

DarwinJS:
It is not really a duplicate - getting a version number to compile into the PE header should be a couple compiler settings.  Additionally, you could have an exe emit it's version without coding the PE header.

Both ways should be probably enabled, but importantly, the PE header can be examined without running the EXE.

SteveL-MSFT:
#4836 is different.  I think I can take care of this one and the icon one at the same time.

iSazonov:
The issue for Windows only. If we'll think about portability it is better to be based on SMA.dll to get a version by Get-Command.

SteveL-MSFT:
@DarwinJS what @iSazonov has found is that SMA.dll has file version information already with PSCore6 even on non-Windows.  Since that is a core dll that will always be there with PowerShell, you can rely on it.

DarwinJS:
I appreciate the tip.  It would be great if it could be right in the EXEs like you all did with openssh.

For instance, I was working on the ability for the openssh package to probe for which exe to use for the DefaultShell.  By the time I got it done, I realized I would need to make sure they can only choose an EXE and should not be allowed to have any wildcards in the file name so that they are always going for a know file name and don't pickup unintention EXEs.

Just wanted to give light to a use case where having it actually in the exe is not only what is generally expected on Windows - in this case it avoids custom code to go find the version for an EXE by looking in a DLL.

Is it somehow more challenging to do it for powershell than it was for openssh?

SteveL-MSFT:
@DarwinJS I have a PR opened for pwsh.exe on Windows, @iSazonov's comment is that checking the version of pwsh.exe only works on Windows while checking version of SMA.dll works on all platforms.

[...]

Issue #5034, opened 2017-10-06
powershell.exe does not contain all VersionInfo properties.


thezim:
Steps to reproduce
------------------

```powershell
Get-Item .\powershell.exe | Format-List
```

Expected behavior
-----------------

The following contain values.

```none
InternalName:
OriginalFilename:
FileVersion:
FileDescription:
Product:
ProductVersion:
```

Actual behavior
---------------

```none
Name           : powershell.exe
Length         : 78336
CreationTime   : 10/4/2017 5:03:40 PM
LastWriteTime  : 10/4/2017 5:03:40 PM
LastAccessTime : 10/4/2017 5:03:40 PM
Mode           : -a----
LinkType       :
Target         : {}
VersionInfo    : File:
                 C:\Development\Apps\PowerShell\powershell.exe
                 InternalName:
                 OriginalFilename:
                 FileVersion:
                 FileDescription:
                 Product:
                 ProductVersion:
                 Debug:            False
                 Patched:          False
                 PreRelease:       False
                 PrivateBuild:     False
                 SpecialBuild:     False
                 Language:
```

Environment data
----------------

```powershell
Name                           Value
----                           -----
PSVersion                      6.0.0-beta.8
PSEdition                      Core
GitCommitId                    v6.0.0-beta.8
OS                             Microsoft Windows 6.1.7601 S
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

SteveL-MSFT:
I believe this is due to dotnet.exe build doesn't insert those resource values.  Related https://github.com/PowerShell/PowerShell/issues/2883

Issue #2883, opened 2016-12-14
Please compile PowerShell icon into powershell.exe for Windows binary


DarwinJS:
This is likely related to #2881 - perhaps the reason that it was necessary to create a separate component for the icon?

Would be nice to have it compiled right in so that normal windows icon behaviors are observed (shortcut creation, visual identification in file system, etc).

iSazonov:
It seems this is still not supported dotnet/cli#3267

SteveL-MSFT:
See discussion in #5013 for some of the findings

PR #5178, opened 2017-10-20
use rcedit to embed icon and version information into pwsh.exe

SteveL-MSFT:
pwsh.exe today doesn't contain file version information and the icon is only associated with the shortcut file and not the exe

Fix is to use rcedit to embed:
- icon
- product version
- file version
- product name
- copyright

Fix https://github.com/PowerShell/PowerShell/issues/2883
Fix https://github.com/PowerShell/PowerShell/issues/5166
Fix https://github.com/PowerShell/PowerShell/issues/5034

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -1862,6 +1862,18 @@ function New-MSIPackage
         [Switch] $Force
     )
 
+    ## need RCEdit to modify the binaries embedded resources
+    if (-not (Test-Path "~/.rcedit/rcedit-x64.exe"))
+    {
+        $rceditUrl = "https://github.com/electron/rcedit/releases/download/v1.0.0/rcedit-x64.exe"
+        New-Item -Path "~/.rcedit" -Type Directory -Force > $null
+        Invoke-WebRequest -OutFile "~/.rcedit/rcedit-x64.exe" -Uri $rceditUrl
+    }
+
+    Start-NativeExecution { & "~/.rcedit/rcedit-x64.exe" (Get-PSOutput) --set-icon "$AssetsPath\Powershell_black.ico" `
+        --set-file-version $ProductVersion --set-product-version $ProductVersion --set-version-string "ProductName" "PowerShell Core 6" `
+        --set-version-string "LegalCopyright" "(C) Microsoft Corporation.  All Rights Reserved." } | Write-Verbose
+
     ## AppVeyor base image might update the version for Wix. Hence, we should
     ## not hard code version numbers.
     $wixToolsetBinPath = "${env:ProgramFiles(x86)}\WiX Toolset *\bin"
--- a/build.psm1
+++ b/build.psm1
@@ -1442,6 +1442,15 @@ function Start-PSBootstrap {
 
         # Install Windows dependencies if `-Package` or `-BuildWindowsNative` is specified
         if ($Environment.IsWindows) {
+            ## need RCEdit to modify the binaries embedded resources
+            if (-not (Test-Path "~/.rcedit/rcedit-x64.exe"))
+            {
+                log "Install RCEdit for modifying exe resources"
+                $rceditUrl = "https://github.com/electron/rcedit/releases/download/v1.0.0/rcedit-x64.exe"
+                New-Item -Path "~/.rcedit" -Type Directory -Force > $null
+                Invoke-WebRequest -OutFile "~/.rcedit/rcedit-x64.exe" -Uri $rceditUrl
+            }
+
             if ($BuildWindowsNative) {
                 log "Install Windows dependencies for building PSRP plugin"
 
@@ -1865,9 +1874,7 @@ function New-MSIPackage
     ## need RCEdit to modify the binaries embedded resources
     if (-not (Test-Path "~/.rcedit/rcedit-x64.exe"))
     {
-        $rceditUrl = "https://github.com/electron/rcedit/releases/download/v1.0.0/rcedit-x64.exe"
-        New-Item -Path "~/.rcedit" -Type Directory -Force > $null
-        Invoke-WebRequest -OutFile "~/.rcedit/rcedit-x64.exe" -Uri $rceditUrl
+        throw "RCEdit is required to modify pwsh.exe resources, please run 'Start-PSBootStrap' to install"
     }
 
     Start-NativeExecution { & "~/.rcedit/rcedit-x64.exe" (Get-PSOutput) --set-icon "$AssetsPath\Powershell_black.ico" `


build.psm1
```diff
@@ -1862,6 +1862,18 @@ function New-MSIPackage
         [Switch] $Force
     )
 
+    ## need RCEdit to modify the binaries embedded resources
+    if (-not (Test-Path "~/.rcedit/rcedit-x64.exe"))
```
iSazonov: What about Unix?

SteveL-MSFT: I'm not as familiar on non-Windows with regards to viewing resources (which Explorer does easily), is this something I can do easily with Ubuntu?  If so, I can move this to the build step rather than packaging.

SteveL-MSFT: Also rcedit only works on Windows

iSazonov: `Get-Item <file> | fl *` show ProductVersion on WSL.

thezim:
This will fix #5034 as well.

iSazonov:
It seems CoreFX use [System.Reflection.Metadata](https://www.nuget.org/packages/System.Reflection.Metadata/) to read FileVersionInfo. Perhaps we can use the API write meta data too.
[CoreFX implementation](https://github.com/dotnet/corefx/tree/master/src/System.Reflection.Metadata/src/System/Reflection/Metadata)

build.psm1
SteveL-MSFT: @iSazonov not working for me with WSL:

```powershell
PS /mnt/c/Windows/System32/WindowsPowerShell/v1.0> get-item ./powershell.exe | fl *


PSPath            : Microsoft.PowerShell.Core\FileSystem::/mnt/c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe
PSParentPath      : Microsoft.PowerShell.Core\FileSystem::/mnt/c/Windows/System32/WindowsPowerShell/v1.0
PSChildName       : powershell.exe
PSDrive           : /
PSProvider        : Microsoft.PowerShell.Core\FileSystem
PSIsContainer     : False
Mode              : --r--l
VersionInfo       : File:             /mnt/c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe
                    InternalName:
                    OriginalFilename:
                    FileVersion:
                    FileDescription:
                    Product:
                    ProductVersion:
```

SteveL-MSFT: It looks like PowerShell is simply using the FileVersionInfo class which doesn't return anything on Linux:
```
PS /mnt/c/Windows/System32/WindowsPowerShell/v1.0> [System.Diagnostics.FileVersionInfo]::GetVersionInfo("/mnt/c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe")

ProductVersion   FileVersion      FileName
--------------   -----------      --------
                                  /mnt/c/Windows/System32/WindowsPowerShell/v1.0/powershell.exe
```
but on Windows:
```
PS C:\Windows\System32\WindowsPowerShell\v1.0> [System.Diagnostics.FileVersionInfo]::GetVersionInfo("C:\Windows\system32
\WindowsPowerShell\v1.0\powershell.exe")

ProductVersion   FileVersion      FileName
--------------   -----------      --------
10.0.17019.1000  10.0.17019.10... C:\Windows\system32\WindowsPowerShell\v1.0\powershell.exe
```

SteveL-MSFT: Linux ELF binaries probably don't have a resource section to stuff this info into.

iSazonov: @SteveL-MSFT Why you test the Windows FullCRL exe file? In my test I used SMA.dll from PowerShell Core.

SteveL-MSFT:
@iSazonov I think we can move to that later particularly if we expose cmdlets for it

iSazonov:
@SteveL-MSFT Do you want to exclude Unix from the PR?

build.psm1
SteveL-MSFT: I used Windows PowerShell.exe as I knew that had that info populated.  I see that SMA.dll from PSCore6 does have that info as does powershell.exe from PSCore6.  So dotnet build is adding that, but the executable doesn't have it.

SteveL-MSFT: This might be because the executable is an actual ELF binary while the dlls are using the same format as Windows making them portable.

SteveL-MSFT:
@iSazonov using the `file` command, the executable is a native executable whereas the .dll assemblies are PE32+ format which supports the embedded resources.  I don't think we can support this for non-Windows executables.

iSazonov:
@SteveL-MSFT Thanks for clarify. I see Unix haven't a standard for this.

build.psm1
```diff
@@ -1862,6 +1862,18 @@ function New-MSIPackage
...
+    {
+        $rceditUrl = "https://github.com/electron/rcedit/releases/download/v1.0.0/rcedit-x64.exe"
```
TravisEz13: Can we install this via chocolatey or nuget?  Based on my understanding of our requirements for release builds, this will have to be rewritten sooner rather than later.

TravisEz13: And move it to Start-PSBootstrap?

TravisEz13 (changes requested):

build.psm1
iSazonov: I don't found nuget package.

SteveL-MSFT: It's not on chocolatey or nuget currently.  I could move the install to start-psbootstrap.

TravisEz13 (approved):

build.psm1
TravisEz13: It's under MIT license.  Nothing keeps us from packaging it our self... and chocolatey, doesn't even package it, it just encapsulates the installation code.

TravisEz13: But this can wait until we have to do it.

build.psm1
```diff
@@ -1442,6 +1442,15 @@ function Start-PSBootstrap {
 
         # Install Windows dependencies if `-Package` or `-BuildWindowsNative` is specified
         if ($Environment.IsWindows) {
+            ## need RCEdit to modify the binaries embedded resources
+            if (-not (Test-Path "~/.rcedit/rcedit-x64.exe"))
```
iSazonov: What about Windows 32-bit?

SteveL-MSFT: Yeah, will add 32-bit support

SteveL-MSFT: Actually, I don't think any change is needed here.  I believe we build x86 running as x64, so we can still use the x64 rcedit to modify the x86 binary.

iSazonov: Clear. Thanks.

Closed.

PR #5183, opened 2017-10-20
Add code to send WebHook for Travis-CI Daily Build

TravisEz13:
Example output in Teams:
image

### DIFF
--- a/tools/travis.ps1
+++ b/tools/travis.ps1
@@ -12,6 +12,7 @@ function Send-DailyWebHook
         )
 
     # Only send web hook if the environment variable is present
+    # Varible should be set in Travis-CI.org settings
     if ($env:WebHookUrl)
     {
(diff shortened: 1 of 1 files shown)


tools/travis.ps1
```diff
@@ -236,6 +272,13 @@ else
             catch {
                 Write-Warning "Could not update status badge: $_"
             }
+            try {
+                Send-DailyWebHook -result $result
+            }
+            catch {
+                Write-Warning "Could not send webhook: $_"
+
```
daxian-dbw: Remove the extra line?

daxian-dbw (approved):
Looks good except for a minor comment.
A question: where do we set the `$webhook` environment variable?

TravisEz13:
The WebhookURL environment variable will be set in travis-ci.org itself.  the same way as the secrets to upload artifacts.

Issue #4971, opened 2017-10-01
Native globbing on Unix platforms should pass syntactically-invalid-as-globbing-patterns unquoted literals through as-is


mklement0:
Related: #3931

An _unquoted_ token that is not a syntactically valid globbing pattern (wildcard expression) should be considered a _literal_ to be _passed as-is_ - not an _error_.

Note that, by default, `bash` retains whatever unquoted token you specify as-is, if it (a) happens not to be a valid globbing pattern or (b) happens not to match any files.
(Setting `shopt -s nullglob`, however, would result in syntactically valid non-matching globs evaluating to the _empty string_).

Steps to reproduce (Unix platforms)
------------------

```powershell
/bin/echo []
```

Expected behavior
-----------------

```none
[]
```

Actual behavior
---------------

```none
The specified wildcard character pattern is not valid: []
...
```

Environment data
----------------

```powershell
PowerShell Core v6.0.0-beta.7 on macOS 10.12.6
PowerShell Core v6.0.0-beta.7 on Ubuntu 16.04.3 LTS
PowerShell Core v6.0.0-beta.7 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
```

SteveL-MSFT:
It seems the downside to making this change is if the user intended to use a valid pattern but typed it wrong

mklement0:
Implementing implicit globbing for external utilities on Unix platforms is an attempt to _emulate_ the behavior of the _typical_ shell behavior there.

Defining _typical_ is tricky, as it depends on whether it refers to the default shell for _interactive_ use - which is often `bash` - vs. running _scripts_ via the POSIX-features-only `/bin/sh` shell (which is `bash` in  its POSIX-compatible mode on some platforms (e.g., macOS), and `dash` on others (e.g, Ubuntu), and possibly a different implementation on others).

Either way, the following - run on macOS 10.12.6 - shows that both `bash` and `dash` interpret unquoted `[]` as a _literal_ by default:

```sh 
$ shall -c '/bin/echo []
✓ sh (bash variant)                       [0.01s]
  []

✓ dash                                    [0.00s]
  []

✓ bash                                    [0.00s]
  []

✗ zsh                                     [0.02s]
  zsh:1: no matches found: []

✓ ksh                                     [0.02s]
  []

```

[`shall`](https://www.npmjs.com/package/shall) is a Unix utility of mine that runs a shell command against multiple POSIX-like shells.

[...]

PR #5188, opened 2017-10-21
Stop globbing of quoted arguments to native commands

### LINKED ISSUE #3931: globbing for native commands is too agressive

lzybkr:
Fix globbing so that quoted args are not expanded.

Also Fix some minor issues with exceptions being raised when resolving
the path - falling back to no glob.

The bulk of the change is in the first commit, which passes the `Ast` or `null` instead of the `IScriptExtent` when constructing the parameters and arguments that are used by the parameter binder.

Fix #3931,#4971

### DIFF
--- a/src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
+++ b/src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
@@ -2,13 +2,18 @@
 Copyright (c) Microsoft Corporation. All rights reserved.
 --********************************************************************/
 
+using Microsoft.PowerShell.Commands;
 using System.Collections;
 using System.Collections.ObjectModel;
+using System.IO;
+using System.Linq;
 using System.Management.Automation.Internal;
 using System.Text;
 
 namespace System.Management.Automation
 {
+    using Language;
+
     /// <summary>
     /// The parameter binder for native commands.
     /// </summary>
@@ -79,7 +84,7 @@ namespace System.Management.Automation
                 if (parameter.ParameterNameSpecified)
                 {
                     Diagnostics.Assert(parameter.ParameterText.IndexOf(' ') == -1, "Parameters cannot have whitespace");
-                    _arguments.Append(parameter.ParameterText);
+                    PossiblyGlobArg(parameter.ParameterText, usedQuotes: false);
 
                     if (parameter.SpaceAfterParameter)
                     {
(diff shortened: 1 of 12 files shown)


rkeithhill:
Must ... have ... this ... fix.  :-)

daxian-dbw:
@lzybkr I was out in training last Friday. Will get this done today.

src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
```diff
@@ -207,105 +227,103 @@ private void appendOneNativeArgument(ExecutionContext context, object obj, char
...
-                                    _arguments.Append(arg);
-                                }
+                    // Expand paths, but only from the file system.
+                    if (paths?.Count > 0 && paths.All(p => p.BaseObject is FileSystemInfo))
+                    {
+                        var sep = "";
+                        foreach (var path in paths)
+                        {
+                            _arguments.Append(sep);
+                            sep = " ";
```
daxian-dbw: Is `sep = " "` really needed here? It seems we just need `var sep = " "` right before the `foreach`

daxian-dbw (dismissed):

src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
lzybkr: Well, it's more tidy - it avoids putting 2 spaces between the previous token - either the command name or an argument.

daxian-dbw: Thanks for clarifying #Closed.

src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
```diff
@@ -207,105 +227,103 @@ private void appendOneNativeArgument(ExecutionContext context, object obj, char
...
-#else
-                            _arguments.Append(arg);
-#endif
+                            argExpanded = true;
                         }
                     }
                 }
-            } while (list != null);
+            }
+            else
```
daxian-dbw: Now we will expand `~` even if `arg` is not a bare word. Do we want to expand `~` unconditionally?

daxian-dbw (commented):
LGTM, but have one more comment about expanding `~`.

src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
daxian-dbw: Another question: it seems to me that after this change we don't glob for variables. Is it intentional?
For example, before the change
```
PS> $var = '*.txt'
PS> /bin/echo $var                                                                               
a.txt b.txt c.txt
```
with the change
```
PS> /bin/echo $var                                                                                
*.txt
```

lzybkr: Thanks for catching, I missed this when reviewing Bruce's changes, but you're right, we don't want to expand `~` unconditionally. I'll fix it.

lzybkr: And I see bash also expands variables unless quoted, so we should probably match that behavior too.

lzybkr: @daxian-dbw - updated. If you sign off, I'll merge - I want to squash the last 2 commits and rebase 2 commits instead squashing the 3 commits.

daxian-dbw (approved):
LGTM. @lzybkr please feel free to merge the PR.

vors:
🎉

PR #5193, opened 2017-10-21
don't insert linebreaks to output (except for tables)

SteveL-MSFT:
don't use current console window width to add linebreaks to output (including debug, error, and verbose)
explicitly change errorrecord formatter to not add linebreaks
set width to int.maxvalue if not specified, in some special cases if width is int.maxvalue, set to console.windowwidth
so that you don't get int.maxvalue padding (like tables)

Related: https://github.com/PowerShell/PowerShell/issues/3813

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterface.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterface.cs
@@ -1148,44 +1148,6 @@ namespace Microsoft.PowerShell
             }
         }
 
-        /// <summary>
-        ///
-        /// Writes text, wrapping a "word" boundaries when needed.
-        ///
-        /// </summary>
-        /// <param name="fg">
-        ///
-        /// Foreground color to output the text.
-        ///
-        /// </param>
-        /// <param name="bg">
-        ///
-        /// Background color to output the text.
-        ///
-        /// </param>
-        /// <param name="text">
-        ///
-        /// Text to be emitted.
-        ///
-        /// </param>
-
-        internal void WriteWrappedLine(ConsoleColor fg, ConsoleColor bg, string text)
-        {
-            WriteLine(fg, bg, WrapToCurrentWindowWidth(text));
-        }
-
-
-
-        // unused for now.
-        //internal
-        //void
-        //WriteWrappedLine(string text)
-        //{
-        //    WriteWrappedLine(RawUI.ForegroundColor, RawUI.BackgroundColor, text);
-        //}
-
-
-
         internal string WrapToCurrentWindowWidth(string text)
         {
             StringBuilder sb = new StringBuilder();
@@ -1207,11 +1169,7 @@ namespace Microsoft.PowerShell
             return sb.ToString();
         }
 
-
-
-        #endregion Word Wrapping
-
-
+#endregion Word Wrapping
 
         /// <summary>
         ///
@@ -1248,7 +1206,7 @@ namespace Microsoft.PowerShell
             else
             {
                 // NTRAID#Windows OS Bugs-1061752-2004/12/15-sburns should read a skin setting here...
-                WriteWrappedLine(
+                WriteLine(
                     DebugForegroundColor,
                     DebugBackgroundColor,
                     StringUtil.Format(ConsoleHostUserInterfaceStrings.DebugFormatString, message));
@@ -1309,7 +1267,7 @@ namespace Microsoft.PowerShell
             }
             else
             {
-                WriteWrappedLine(
+                WriteLine(
                     VerboseForegroundColor,
                     VerboseBackgroundColor,
                     StringUtil.Format(ConsoleHostUserInterfaceStrings.VerboseFormatString, message));
@@ -1352,7 +1310,7 @@ namespace Microsoft.PowerShell
             }
             else
             {
-                WriteWrappedLine(
+                WriteLine(
                     WarningForegroundColor,
                     WarningBackgroundColor,
                     StringUtil.Format(ConsoleHostUserInterfaceStrings.WarningFormatString, message));
--- a/src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs
+++ b/src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs
@@ -921,7 +921,11 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 }
 
                 int columnsOnTheScreen = this.InnerCommand._lo.ColumnNumber;
-                // for tables, we want to make sure we don't end up adding padding to int.MaxValue and default to 120 columns instead
+                // Tables need to use spaces for padding to maintain table look even if console window is resized.
+                // For all other output, we use int.MaxValue if the user didn't explicitly specify a width.
+                // If we detect that int.MaxValue is used, first we try to get the current console window width.
+                // However, if we can't read that (for example, implicit remoting has no console window), we default
+                // to something reasonable: 120 columns.
                 if (columnsOnTheScreen == int.MaxValue)
                 {
                     try
--- a/src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs
+++ b/src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs
@@ -921,6 +921,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 }
 
                 int columnsOnTheScreen = this.InnerCommand._lo.ColumnNumber;
+                // for tables, we want to make sure we don't end up adding padding to int.MaxValue and default to 120 columns instead
                 if (columnsOnTheScreen == int.MaxValue)
                 {
                     try
--- a/src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs
+++ b/src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs
@@ -565,7 +565,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             {
                 ComplexWriter complexWriter = new ComplexWriter();
 
-                complexWriter.Initialize(_lo, _lo.ColumnNumber);
+                complexWriter.Initialize(_lo, int.MaxValue);
                 complexWriter.WriteObject(cve.formatValueList);
 
                 return;
@@ -921,6 +921,10 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 }
 
                 int columnsOnTheScreen = this.InnerCommand._lo.ColumnNumber;
+                if (columnsOnTheScreen == int.MaxValue)
+                {
+                    columnsOnTheScreen = Console.WindowWidth;
+                }
 
                 int columns = this.CurrentTableHeaderInfo.tableColumnInfoList.Count;
                 if (columns == 0)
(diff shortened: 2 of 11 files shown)


test/powershell/Language/Scripting/NativeExecution/NativeStreams.Tests.ps1
```diff
@@ -49,14 +49,25 @@ Describe "Native streams behavior with PowerShell" -Tags 'CI' {
...
+
+        It 'does not get truncated or split when redirected' {
+            $longtext = "0123456789"
+            while ($longtext.Length -lt [console]::WindowWidth) {
+                $longtext += $longtext
+            }
+            pwsh -c "& { [Console]::Error.WriteLine('$longtext') }" 2>&1 > $testdrive\error.txt
+            $e = Get-Content -Path $testdrive\error.txt
+            $e.Count | Should BeExactly 1
+            $e | Should Match $longtext
```
iSazonov: Maybe use "BeExactly"? It seems better for testing a formatting.

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Debug.Tests.ps1
```diff
@@ -0,0 +1,17 @@
+Describe "Write-Debug tests" -Tags "CI" {
+    It "Should not have added line breaks" {
+        $text = "0123456789"
+        while ($text.Length -lt [Console]::WindowWidth) {
+            $text += $text
+        }
+        $origDebugPref = $DebugPreference
+        $DebugPreference = "Continue"
+        try {
+            $out = Write-Debug $text *>&1
```
iSazonov: Maybe use exactly "2>&1"

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Debug.Tests.ps1
```diff
@@ -0,0 +1,17 @@
...
+            $out | Should Match $text
```
iSazonov: Maybe use "$out | Should BeExactly $longtext"?

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Debug.Tests.ps1
```diff
@@ -0,0 +1,17 @@
...
+        }
+        finally {
+            $DebugPreference = $origDebugPref
+        }
+    }
+}
```
iSazonov: Please add new line at EOF.

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Error.Tests.ps1
```diff
@@ -74,40 +74,49 @@ Describe "Write-Error DRT Unit Tests" -Tags "CI" {
...
-	$error[0]| Should Be $theError
+    It "ErrorRecord should not be truncated" {
+        $longtext = "0123456789"
+        while ($longtext.Length -lt [console]::WindowWidth) {
+            $longtext += $longtext
+        }
+        pwsh -c Write-Error -Message $longtext 2>&1 > $testdrive\error.txt
+        $e = Get-Content -Path $testdrive\error.txt
+        $e.Count | Should BeExactly 4
+        $e[0] | Should Match $longtext
```
iSazonov: Can we use more strong match pattern? I'm again worried about false positives.

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Verbose.Tests.ps1
```diff
@@ -26,4 +26,20 @@ Describe "Write-Verbose" -Tags "CI" {
...
+
+    It "Should not have added line breaks" {
+        $text = "0123456789"
+        while ($text.Length -lt [Console]::WindowWidth) {
+            $text += $text
+        }
+        $origVerbosePref = $VerbosePreference
+        $VerbosePreference = "continue"
+        try {
+            $out = Write-Verbose $text *>&1
```
iSazonov: The same "2>&1".

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Verbose.Tests.ps1
```diff
@@ -26,4 +26,20 @@ Describe "Write-Verbose" -Tags "CI" {
...
+            $out | Should Match $text
```
iSazonov: The same about "BeExactly".

src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs
```diff
@@ -921,6 +921,17 @@ internal override void Initialize()
                 }
 
                 int columnsOnTheScreen = this.InnerCommand._lo.ColumnNumber;
+                if (columnsOnTheScreen == int.MaxValue)
```
iSazonov: Could we please add comments about the special cases?

test/powershell/Language/Scripting/NativeExecution/NativeStreams.Tests.ps1
SteveL-MSFT: The problem currently is that redirection writes with a BOM, so this would fail

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Debug.Tests.ps1
SteveL-MSFT: Debug isn’t stream number 2 and it seems you can only redirect debug if you redirect all streams

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Debug.Tests.ps1
SteveL-MSFT: Same problem as above in that redirection currently adds a BOM

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Debug.Tests.ps1
SteveL-MSFT: Will do

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Error.Tests.ps1
SteveL-MSFT: Write-Error adds extra info prefixed to the text.  The important bits for this test is the long text not getting choppedby a linebreak

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Verbose.Tests.ps1
SteveL-MSFT: Verbose isn’t stream 2, I believe it’s 3, but redirection only worked when using all streams

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Verbose.Tests.ps1
SteveL-MSFT: Same as above, BOM gets added

test/powershell/Language/Scripting/NativeExecution/NativeStreams.Tests.ps1
iSazonov: We read with `Get-Content -Path $testdrive\error.txt` - I believe there is not BOM if we don't use `-Byte`.

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Debug.Tests.ps1
iSazonov: It seems works:
```powershell
PS> $a=Write-Verbose "ext " -Verbose 4>&1
PS> $a
VERBOSE: ext
PS> $b=Write-Debug "qwerty " -Verbose 5>&1
PS> $b
PS> $b=Write-Debug "qwerty " -Debug 5>&1
Confirm
Continue with this operation?
[Y] Yes  [A] Yes to All  [H] Halt Command  [S] Suspend  [?] Help (default is "Y"): y
PS C:\Users\sie> $b

DEBUG: qwerty
```

src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs
SteveL-MSFT: Yes, will add

test/powershell/Language/Scripting/NativeExecution/NativeStreams.Tests.ps1
SteveL-MSFT: Good point about Get-Content, I'll change it

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Debug.Tests.ps1
SteveL-MSFT: Maybe I did something wrong when I tried it.  Will update.

test/powershell/Modules/Microsoft.PowerShell.Utility/Write-Error.Tests.ps1
iSazonov: I see - no false positives can be.

Closed.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterface.cs
```diff
@@ -1248,7 +1206,7 @@ public override void WriteDebugLine(string message)
             else
             {
                 // NTRAID#Windows OS Bugs-1061752-2004/12/15-sburns should read a skin setting here...
```
iSazonov: Is the comment still actual?

src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs
```diff
@@ -921,6 +921,18 @@ internal override void Initialize()
                 }
 
                 int columnsOnTheScreen = this.InnerCommand._lo.ColumnNumber;
+                // for tables, we want to make sure we don't end up adding padding to int.MaxValue and default to 120 columns instead
```
iSazonov: Can we rephrase it for easy reading? 

And why 120? Can we use current windows width?

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterface.cs
SteveL-MSFT: I believe the comment is saying that if we decide to support color themes (aka skins), this is one place to put the code for Write-Debug.  Did a quick search for "read a skin" and found four places for Debug, Verbose, Warning, and ProgressPane.

src/System.Management.Automation/commands/utility/FormatAndOutput/common/BaseOutputtingCommand.cs
SteveL-MSFT: Let me try to add a bit more detail to make it easier to understand.  The code tries to use window width, but if it can't be read (like implicit remoting where there's no console window), we have to default to something.  120 was the original default without my changes.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterface.cs
iSazonov: Thanks for clarify.

Closed.

lzybkr (approved):
I tried out these changes - it looks like a big improvement for Format-List and no change for Format-Table or Format-Wide and error records aren't wrapped, so that's great.

Maybe in some future work we can detect when the output is to a file and be smarter about Format-Table and Format-Wide.

daxian-dbw:
Given that @iSazonov's comments have been addressed, I will merge this PR.

PR #5195, opened 2017-10-21
Display full help with 'help' function

### LINKED ISSUE #4962: Before 6.0.0 ships can we make Get-Help -full the default?
On Linux, the man pages display their full content by default.  I believe PowerShell should do the same.  The number of times I use get-help (or the alias man) without the `-Full` parameter can be counted on 1 hand (actually probably 0 hands).  I've always disliked the switch that was made to seriously abbreviate PowerShell help topics.  It was too dumbed down. 

So here's a chance to do the right thing and make the default output the *full* output.  I think that is what both Linux and macOS users would expect.  

A "Did you look at the help for the examples on how to use that command", 
B "Yes, I didn't see any."
A "Oh, didn't you use the -Full parameter on help?"
B "Huh?  You mean the help command didn't show me **all** the help?"
A "Nope."
B "WTF?"

rkeithhill:
Display full help on Windows systems where user has overriden pager with $env:PAGER.

Fix #4962


rkeithhill:
Any chance of something like this making it in before 6.0.0 final?  It would be nice to offer full help to Linux users.  And to Windows  users who elect to set $env:PAGER to a better pager app.  :-)

adityapatwardhan:
@rkeithhill  Sorry for the delay. I will review this today.

rkeithhill:
No problem.  You all had a big push to get beta.9 out.

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -4253,6 +4253,12 @@ .FORWARDHELPCATEGORY Cmdlet
     [switch]
     ${ShowWindow})
 
+    # On non-Window systems always display the full help topic.
+    # On Windows, only display the full help topic if the user has specified a PAGER app.
+    if (($psCmdlet.ParameterSetName -eq 'AllUsersView') -and !$Full -and (!$IsWindows -or (Test-Path env:PAGER))) {
+        $PSBoundParameters['Full'] = $true
+    }
```
adityapatwardhan: Can we add a test which check the view that gets selected. It should have `#FullView`
```PowerShell
(Get-Help Get-Command -Full).psobject.typenames
```

adityapatwardhan (changes requested):

src/System.Management.Automation/engine/InitialSessionState.cs
rkeithhill: Yes, I can add that.

rkeithhill: BTW, to be clear, this tweak affects only the `help` function on Linux/macOS/Windows as well as the `man` alias on Windows.  That function pipes through a pager so we don't get rich objects out - just text.  That is why I'm using `-cmatch` to look for certain words that indicate it is full help.

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -300,5 +300,83 @@ Describe "Get-Help should find pattern help files" -Tags "CI" {
...
+        }
+    }
+
+    AfterAll {
+        if ($origPager) {
+            New-Item Env:\PAGER -Value $origPager > $null
+        }
+    }
+
+    It "help should not return full view without -Full switch" -Skip:(!$IsWindows) {
```
adityapatwardhan: Since you already have and after all the `Skip` should be done over there instead of repeating it in every `It`.

https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/WritingPesterTests.md#skipping-tests-in-bulk

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -300,5 +300,83 @@ Describe "Get-Help should find pattern help files" -Tags "CI" {
...
+    AfterAll {
+        if ($origPager) {
+            Set-Item Env:\PAGER -Value $origPager
+        }
+        else {
+            Remove-Item Env:\PAGER
+        }
+    }
+
+    It "help should return full view without -Full switch" -Skip:(!$IsWindows) {
```
adityapatwardhan: Same comment as above.

adityapatwardhan (changes requested):

adityapatwardhan (approved):
LGTM

rkeithhill:
I think this is ready to go.  The feedback from the @PowerShell/powershell-committee review was implemented in the last commit.

PR #5206, opened 2017-10-24
Fix AppImage package name to be lower case 'powershell-..'

daxian-dbw:
Fix AppImage package name to be lower case 'powershell-..'
This makes the package name consistent with all the other Linux package names.


TravisEz13 (approved):

adityapatwardhan (approved):

daxian-dbw:
Double checked the existing AppImage package. It prompts for desktop integration too, but the icon shown up in GNOME search is not the powershell icon. Opened the issue #5207

PR #5209, opened 2017-10-24
Build: Make powershell build in Visual Studio

powercode:
#3400 

Adding `TargetFramework` property to csprojs. This needs to be in each project to work around a visual studio issue where it doesn't react to TargetFramework set in an imported file.
Updating sln to reflect the change of project types to .net core
removing obsolete resources that makes the build fail.

At least I think the resources are obsolete :) 
Hope reviewers can verify.


iSazonov:
I think we should add `TargetFramework` property to `PowerShell.Common.props` not every csproj.

powercode:
The whole point of having it in every project is to work around the issue that visual studio issue doesn't understand TargetFramework from imported files.

powercode:
The property already is in `PowerShell.Common.props`.

iSazonov:
I wonder how and why the trick works.

powercode:
My guess is that they explicitly look for the property in the file, but using the full logic for evalutating it's value.

powercode:
By doing this, we can at least keep the value consistent, and updatable in one place, while still being able to use Visual Studio. 

I agree it isn't ideal, and it should be removed if VS fixes the issue. 
I haven't checked where this can be reported.  @PowerShellTeam anyone got any good connections in devdiv?

iSazonov:
It would be good to have a comment from VS team that is best way in the case.

powercode:
And maybe a comment in the csproj files why that strange construct is used.

Bhaal22:
This issue looks to be related the issue experienced here: 

https://github.com/dotnet/project-system/issues/1358

bergmeister:
I have tried your fork on my local machine. I ran `Start-PSBootStrap`, installed the .Net Core SDK 2.0.0 (since this is the state of your fork) and ran a `dotnet restore` on the solution. Then trying to build in VS gives me loads of errors:
image

powercode:
@bergmeister That is wierd. That reference is removed in c8e4192

bergmeister:
@powercode Sorry, I was caught out by cloning your fork but being on the wrong branch. I can see the files now in solution explorer but building only seems to work if the repo is pre-build via command line via `Start-PSBuild` (just running `Start-ResGen` is not sufficient). I am using using VS 15.5.2 on Win10.

bergmeister:
@powercode : can you please downstream changes from master to your fork since GitHub does not allow to have [multiple forks](https://stackoverflow.com/a/40367495/1810304) so that people can start using it already. What timeline do we expect until this is merged into master @TravisEz13 ?
Downstreaming can be done as follows:
````powershell
cd $PowerShellCheckout
git checkout vs-csproj
git remote add upstream https://github.com/PowerShell/PowerShell.git
git pull upstream master
git push
````

**Comment by TravisEz13** [comment]
@bergmeister , @daxian-dbw  is assigned to this PR.
I've restarted Travis-CI linux due to markdown test failures due to out of date code.  hopefully github will remerge on it's own.

**Comment by powercode** [comment]
Rebased on master.

File: PowerShell-Win.sln [hunk 1]
```diff
@@ -101,44 +101,44 @@ Global
 		{131A8527-92D7-468F-822D-5354229A865C}.Linux|Any CPU.Build.0 = Release|Any CPU
 		{131A8527-92D7-468F-822D-5354229A865C}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{131A8527-92D7-468F-822D-5354229A865C}.Release|Any CPU.Build.0 = Release|Any CPU
-		{439A24FC-8E0A-48B6-8227-44C297311F49}.CodeCoverage|Any CPU.ActiveCfg = CodeCoverage|Any CPU
-		{439A24FC-8E0A-48B6-8227-44C297311F49}.CodeCoverage|Any CPU.Build.0 = CodeCoverage|Any CPU
+		{439A24FC-8E0A-48B6-8227-44C297311F49}.CodeCoverage|Any CPU.ActiveCfg = Release|Any CPU
```
iSazonov: Could you please clarify why we need only "Release"?

iSazonov:
@markekraus Do you use Visual Studio? Could you review the PR?

markekraus:
@iSazonov I have Visual Studio, but I'm not familiar with it (I use VS Code). You are better off asking for a review @rkeithhill and/or @bergmeister. I believe they are both familiar with Visual Studio.

bergmeister:
I can take a look again since I have to use VS nearly the whole day but my review from a user's perspective is already [here](https://github.com/PowerShell/PowerShell/pull/5209#issuecomment-352091250)
But it definitely makes sense if others chip in as well.

iSazonov:
@daxian-dbw Could you please continue with the PR?

daxian-dbw:
We need to actually use the changes for a few days to get a feel. I suggest all active maintainers to try out the changes. /cc @PowerShell/powershell-maintainers 
Also, for anyone who want/like this PR, please also try out the changes and help bring up issues. The more eyes the merrier :)

PowerShell-Win.sln
powercode: That may have been unintentional. Seems to me, that our platform/config system is not really well designed, but it is not that important here. To begin with, this is just so we can start iterating faster on the code from Visual Studio. It is not affecting the build of the product.

iSazonov:
1. Step

git clean -fdX
run Build in VS

========== Rebuild All: 0 succeeded, 13 failed, 0 skipped ==========

I found related https://github.com/dotnet/project-system/issues/2129

2. Step

dotenet restore
run Build in VS

We see two errors "Invalid Resx file" with xsd.
src\System.Management.Automation\resources\CmdletizationCoreResources.resx
src\Microsoft.PowerShell.Commands.Management\resources\CmdletizationResources.resx


3.Step

After fix this Resx files
run Build in VS

visual studio cannot start debugging because the debug target is missing exe

powercode:
Until we move resource generation into a msbuild task, we will have to complete a cmdline build first.

I set up a publishing rule to publish a build before debugging, and manually configured the debug target.
Debugging is generally not something you share, since you often start with different arguments.

iSazonov:
I already push PR #3870 for TypeGen but it don't approved a long time. I already patched our temp utility for ResGen in PR #3918 and ready to push PR to run it by MSBuild.

TravisEz13:
I created a new PR based on this one.  #6546   I've verified it works and the fix is simpler.  I based it on your fix.  I think we should probably close this one.  Thanks, Travis . Thanks for your help.

PowerShell-Win.sln
```diff
@@ -3,31 +3,33 @@ Microsoft Visual Studio Solution File, Format Version 12.00
...
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.WSMan.Management", "src\Microsoft.WSMan.Management\Microsoft.WSMan.Management.csproj", "{8F63D134-E413-4181-936D-D82F3F5F1D85}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.PowerShell.Security", "src\Microsoft.PowerShell.Security\Microsoft.PowerShell.Security.csproj", "{C4F81816-C87A-4ABF-8A37-24AC16A0A6CF}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.PowerShell.Security", "src\Microsoft.PowerShell.Security\Microsoft.PowerShell.Security.csproj", "{C4F81816-C87A-4ABF-8A37-24AC16A0A6CF}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.WSMan.Runtime", "src\Microsoft.WSMan.Runtime\Microsoft.WSMan.Runtime.csproj", "{D9CCCB67-4EBE-4854-AB52-C0129DC5BAE4}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.WSMan.Runtime", "src\Microsoft.WSMan.Runtime\Microsoft.WSMan.Runtime.csproj", "{D9CCCB67-4EBE-4854-AB52-C0129DC5BAE4}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "powershell-unix", "src\powershell-unix\powershell-unix.csproj", "{73EA0BE6-C0C5-4B56-A5AA-DADA4C01D690}"
```
daxian-dbw: Want to change this one as well?

PowerShell-Win.sln
```diff
@@ -3,31 +3,33 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
```
daxian-dbw: Can we have a comment here pointing to  https://github.com/dotnet/project-system/blob/master/docs/opening-with-new-project-system.md#project-type-guids ?

src/Microsoft.PowerShell.Commands.Management/resources/CmdletizationResources.resx
```diff
@@ -124,10 +124,6 @@
 {1} is a placeholder for a server name.  Example: "localhost".</comment>
   </data>
   <assembly alias="System.Windows.Forms" name="System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" />
```
daxian-dbw: I thought this file can be removed, not?

PR #5218, opened 2017-10-24
Rename to unix for installation instructions.

thezim:
If we are going to group macOS with Linux installation instructions we should use the general name Unix.


TravisEz13:
@joeyaiello Can you please review?

@thezim   CI fail because you didn't update `.spelling` .  https://travis-ci.org/PowerShell/PowerShell/jobs/292313955#L3205

thezim:
@TravisEz13 Done.

TravisEz13:
@thezim Sorry, I didn't follow up.  Can you resolve the merge conflicts? 
 @SteveL-MSFT  or @joeyaiello  can you review?

SteveL-MSFT:
My main concern is that Linux isn't Unix, but I don't want to get into a philosophy/religious discussion here (but my impression is that Linux hardcore fans care about the distinction).  If the concern is grouping macOS with Linux, then perhaps we should split the docs keeping Linux just the OSs using the Linux kernel and a separately set for macOS?

thezim:
@SteveL-MSFT I agree, that was my intension, macOS != Linux. I'll close this PR an start a new one with macOS as its own set of documents.

PR #5221, opened 2017-10-25
[wip] Parallel test execution for start-pspester

SteveL-MSFT:
Enable running tests in parallel using runspaces with new `-UseRunspacePool` switch.

Still TODO: support UnElevate and merging NUnit logs to a single log.


iSazonov:
Make it sense on slow CI VMs?

SteveL-MSFT:
@iSazonov that's part of the experiment I'm trying to figure out.  I think even with 1 virtual core, there is enough wait time that having a second thread running may help.

test/powershell/Modules/Microsoft.PowerShell.Management/FileSystem.Tests.ps1
```diff
@@ -1389,3 +1350,44 @@ Describe "UNC paths" -Tags 'CI' {
...
+            "Test timeout waiting for $cmdLine" | Set-Content -Path $errFile
+
+            runas.exe /trustlevel:0x20000 "$powershell -nop -c try { $cmdline -ErrorAction Stop } catch { `$_.FullyQualifiedErrorId | Out-File $errFile }; New-Item -Type File -Path $doneFile"
+            Wait-FileToBePresent -File $doneFile -TimeoutInSeconds 15 -IntervalInMilliseconds 100
+
+            $err = Get-Content $errFile
+            $err | Should Be $expectedError
+        }
+    }
+}
```
TravisEz13: missing newline

SteveL-MSFT: will add

build.psm1
```diff
@@ -914,9 +964,28 @@ Make sure that the proper git submodules are installed by running:
         $outputBufferFilePath = [System.IO.Path]::GetTempFileName()
     }
 
-    $Command += "Invoke-Pester "
+    $testPaths = @{}
+    if ($UseRunspacePool)
```
adityapatwardhan: We need to figure out how to make TestHooks session local v/s global as they are currently, before this gets merged.

build.psm1
```diff
@@ -914,9 +964,28 @@ Make sure that the proper git submodules are installed by running:
...
+        $testScriptPaths = @()
+        foreach ($subPath in $Path)
+        {
+            $testScriptPaths += (Get-ChildItem -Recurse -Path $subPath -Filter *.?ests.ps1 | Select-Object -Property DirectoryName -Unique).DirectoryName
+        }
+        # Remove any child paths included in parents
+        $prevPath = (Get-Random).ToString()
+        foreach ($testPath in $testScriptPaths)
+        {
+            if (!$testPath.Contains($prevPath))
```
adityapatwardhan: Can we filter for Remoting folder here? Maybe move the filtering to a function so we can easily add more folders to be filtered if necessary.

build.psm1
```diff
@@ -932,141 +1001,222 @@ Make sure that the proper git submodules are installed by running:
...
                     {
-                        Start-NativeExecution -sb {& $powershell -noprofile -c $Command} | ForEach-Object { Write-Host $_}
+                        $lines | Write-Host
+                    }
+                    if ($lines | Where-Object { $_ -eq '__UNELEVATED_TESTS_THE_END__'})
+                    {
+                        break
+                    }
+
+                    $count = ($lines | measure-object).Count
```
adityapatwardhan: Fix casing on `measure-object`

adityapatwardhan (changes requested):

build.psm1
SteveL-MSFT: One option, that I don't particularly like, is to put all the test hook tests in the same folder so they are all run serially

SteveL-MSFT:
Closing for now as no work is happening for this.

PR #5222, opened 2017-10-25
Update install/build scripts and dockerfiles due to the executable name change

daxian-dbw:
Need to update install script, otherwise, Travis-CI will fail.

**Note that:** the `installpsh-osx.sh` depends on homebrew to install macOS package, so this script will fail until the new macOS package is uploaded to homebrew, which depends on the package being available on Github release page.


adityapatwardhan (approved):
LGTM

Issue #5158, opened 2017-10-18
Add-Type ReadAllText perf improvement


PaulHigin:
Add-Type uses File.ReadAllText() API in a loop for reading in multiple source files.  This can be non-performant for multiple reasons:
a. ReadAllText() uses a small buffer that increases IO calls.
b. It performs StringBuilder.ToString() internally which allocates a string that is not used, making possible LOH allocation and adding GC pressure.

Consider using a different pattern:
```
StringBuilder sourceCode = new StringBuilder();

foreach (string file in paths)
{
    foreach (string line in File.ReadAllLines(file))
    {
        sourceCode.AppendLine(line);
    }
}

String result = sourceCode.ToString();
```

SaiKai:
I will take a stab at this.

iSazonov:
I don't see using `StringBuilder` in CoreFX implementation https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/File.cs#L376

iSazonov:
[ReadAllLines](https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/File.cs#L499) put string in `List` then we put in `StringBuilder` - I don't see benefits.
I found [tests](http://cc.davelozinski.com/c-sharp/fastest-way-to-read-text-files). ReadAllText looks best compared to ReadAllLines if we want to read the entire file in a single string.

lzybkr:
if `ReadAllText` is inefficient, we should be fixing that, not working around it.

The use case here though is a little different, we're reading multiple files into a single buffer.

With some unsafe code, one could preallocate a `char[]` buffer big enough to hold all of the files, then use `StreamReader.Read(char[] buffer, int index, int count)` - this would minimize the number of large allocations to just 2 - the `char[]` and then turning that into a string.

I am curious where this came up though - if this was a perf problem, why not just compile the code with C# ahead of time?

PaulHigin:
This was reported internally based on ETW perf traces showing a lot of LOH allocations from Add-Type/ReadAllText.  But I think you are right, calling Add-Type so often is unexpected and sounds like a possible design/implementation error.

iSazonov:
I'm surprised too that we compile a string instead of files. We could use `CompileAssemblyFromFile`

lzybkr:
Good suggestion - it might not have been an option with CodeDom, but now that we're using Roslyn, there obviously should be better options.

[...]

PR #5243, opened 2017-10-26
Fix performance issues in Add-Type

iSazonov:
Related #5158.

### Fix description

Commit 1. Minor optimizations in `OutputError()`.

Commit 2. Before the fix we read sources files twice in base.EndProcessing() and in this.EndProcessing() - it is excluded. Also now we read source files in StringBuilder to exclude large reallocations. Added one test.

### Additional considerations
I expect the fix remove performance issues reported in #5158. So we can close the Issue by the PR and open new Issue to discuss refactoring the code to use ` CompileAssemblyFromFile()`. Using Roslyn can open paths to enhance the Add-Type cmdlet's capabilities but requires a lot of work.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
@@ -763,6 +763,14 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         protected override void EndProcessing()
         {
+            // Prevent code compilation in ConstrainedLanguage mode
+            if (SessionState.LanguageMode == PSLanguageMode.ConstrainedLanguage)
+            {
+                ThrowTerminatingError(
+                    new ErrorRecord(
+                        new PSNotSupportedException(AddTypeStrings.CannotDefineNewType), "CannotDefineNewType", ErrorCategory.PermissionDenied, null));
+            }
+
             // Generate an error if they've specified an output
             // assembly type without an output assembly
             if (String.IsNullOrEmpty(outputAssembly) && outputTypeSpecified)
@@ -778,36 +786,6 @@ namespace Microsoft.PowerShell.Commands
 
                 ThrowTerminatingError(errorRecord);
                 return;
-            }
-
-            PopulateSource();
-        }
-
-        internal void PopulateSource()
-        {
-            // Prevent code compilation in ConstrainedLanguage mode
-            if (SessionState.LanguageMode == PSLanguageMode.ConstrainedLanguage)
-            {
-                ThrowTerminatingError(
-                    new ErrorRecord(
-                        new PSNotSupportedException(AddTypeStrings.CannotDefineNewType), "CannotDefineNewType", ErrorCategory.PermissionDenied, null));
-            }
-
-            // Load the source if they want to load from a file
-            if (String.Equals(ParameterSetName, "FromPath", StringComparison.OrdinalIgnoreCase) ||
-                String.Equals(ParameterSetName, "FromLiteralPath", StringComparison.OrdinalIgnoreCase)
-                )
-            {
-                sourceCode = "";
-                foreach (string file in paths)
-                {
-                    sourceCode += System.IO.File.ReadAllText(file) + "\n";
-                }
-            }
-
-            if (String.Equals(ParameterSetName, "FromMember", StringComparison.OrdinalIgnoreCase))
-            {
-                sourceCode = GenerateTypeSource(typeNamespace, Name, sourceCode, language);
             }
         }
 
@@ -932,13 +910,23 @@ namespace Microsoft.PowerShell.Commands
             {
                 // Load the source if they want to load from a file
                 if (String.Equals(ParameterSetName, "FromPath", StringComparison.OrdinalIgnoreCase) ||
-                    String.Equals(ParameterSetName, "FromLiteralPath", StringComparison.OrdinalIgnoreCase))
+                    String.Equals(ParameterSetName, "FromLiteralPath", StringComparison.OrdinalIgnoreCase)
+                   )
                 {
-                    this.sourceCode = "";
+                    StringBuilder sb = new StringBuilder(paths.Length);
+                    sourceCode = "";
+
                     foreach (string file in paths)
                     {
-                        this.sourceCode += System.IO.File.ReadAllText(file) + "\n";
+                        sb.Append(System.IO.File.ReadAllText(file));
+                        sb.Append("\n");
                     }
+
+                    sourceCode = sb.ToString();
+                }
+                else if (String.Equals(ParameterSetName, "FromMember", StringComparison.OrdinalIgnoreCase))
+                {
+                    sourceCode = GenerateTypeSource(typeNamespace, Name, sourceCode, language);
                 }
 
                 CompileSourceToAssembly(this.sourceCode);
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
@@ -913,15 +913,44 @@ namespace Microsoft.PowerShell.Commands
                     String.Equals(ParameterSetName, "FromLiteralPath", StringComparison.OrdinalIgnoreCase)
                    )
                 {
-                    StringBuilder sb = new StringBuilder(paths.Length);
-
-                    foreach (string file in paths)
+                    if (paths.Length == 1)
                     {
-                        sb.Append(System.IO.File.ReadAllText(file));
-                        sb.Append("\n");
+                        sourceCode = File.ReadAllText(paths[0]);
                     }
+                    else
+                    {
 
-                    sourceCode = sb.ToString();
+                        long initLength = 0;
+
+                        foreach (string file in paths)
+                        {
+                            FileInfo f = new FileInfo(file);
+                            initLength = f.Length;
+                        }
+
+                        if (initLength < int.MaxValue)
+                        {
+                            StringBuilder sb = new StringBuilder((int)initLength);
+
+                            foreach (string file in paths)
+                            {
+                                foreach (string line in File.ReadAllLines(file))
+                                {
+                                    sb.AppendLine(line);
+                                }
+                            }
+
+                            sourceCode = sb.ToString();
+                        }
+                        else
+                        {
+                            sourceCode = "";
+                            foreach (string file in paths)
+                            {
+                                sourceCode += System.IO.File.ReadAllText(file) + "\n";
+                            }
+                        }
+                    }
                 }
                 else if (String.Equals(ParameterSetName, "FromMember", StringComparison.OrdinalIgnoreCase))
                 {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
@@ -914,7 +914,6 @@ namespace Microsoft.PowerShell.Commands
                    )
                 {
                     StringBuilder sb = new StringBuilder(paths.Length);
-                    sourceCode = "";
 
                     foreach (string file in paths)
                     {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
@@ -920,17 +920,10 @@ namespace Microsoft.PowerShell.Commands
                     else
                     {
 
-                        long initLength = 0;
-
                         // We replace 'ReadAllText' with 'StringBuilder' and 'ReadAllLines'
                         // to avoide temporary LOH allocations.
-                        foreach (string file in paths)
-                        {
-                            FileInfo f = new FileInfo(file);
-                            initLength += f.Length;
-                        }
 
-                        StringBuilder sb = new StringBuilder((int)initLength);
+                        StringBuilder sb = new StringBuilder(8192);
 
                         foreach (string file in paths)
                         {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
@@ -853,12 +853,12 @@ namespace Microsoft.PowerShell.Commands
             {
                 if (!String.IsNullOrEmpty(error.FileName))
                 {
-                    actualSource = System.IO.File.ReadAllText(error.FileName).Split(Utils.Separators.Newline);
+                    actualSource = System.IO.File.ReadAllLines(error.FileName);
                 }
             }
 
             string errorText = StringUtil.Format(AddTypeStrings.CompilationErrorFormat,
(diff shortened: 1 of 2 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
```diff
@@ -932,13 +910,23 @@ protected override void EndProcessing()
             {
                 // Load the source if they want to load from a file
                 if (String.Equals(ParameterSetName, "FromPath", StringComparison.OrdinalIgnoreCase) ||
-                    String.Equals(ParameterSetName, "FromLiteralPath", StringComparison.OrdinalIgnoreCase))
+                    String.Equals(ParameterSetName, "FromLiteralPath", StringComparison.OrdinalIgnoreCase)
+                   )
                 {
-                    this.sourceCode = "";
+                    StringBuilder sb = new StringBuilder(paths.Length);
+                    sourceCode = "";
```
PaulHigin: sourceCode doesn't need to be initialized since it is always assigned after loop.  Also we normally name class fields with beginning underscore: "_sourceCode".

PaulHigin (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
iSazonov: I'll remove the line.

`sourceCode` is used in many places in the file. If I rename it will be very noisy and unrelated with the performance fix - I think it is better to leave it as is.

PaulHigin (approved):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
```diff
@@ -932,13 +910,22 @@ protected override void EndProcessing()
...
+                    String.Equals(ParameterSetName, "FromLiteralPath", StringComparison.OrdinalIgnoreCase)
+                   )
                 {
-                    this.sourceCode = "";
+                    StringBuilder sb = new StringBuilder(paths.Length);
+
                     foreach (string file in paths)
                     {
-                        this.sourceCode += System.IO.File.ReadAllText(file) + "\n";
+                        sb.Append(System.IO.File.ReadAllText(file));
```
daxian-dbw: Some feedback:
1. `ReadAllText` returns a large string in the end, which may be allocated in LOH if the file size is larger than 85kb. The allocation of this large string will be wasted since we are appending it to the StringBuilder.
If using `ReadAllLines`, then only small strings will be created, which can be quickly garbage collected. So it probably would be better if we go with
```
foreach (string file in paths)
{
    foreach (string line in File.ReadAllLines(file))
    {
        sb.AppendLine(line);
    }
}
sourceCode = sb.ToString();
```
2. When there is only one file specified, which might be the case for the most time, this is still not performant. It could be better if we can treat the one-file case specially. Like:
```
if (paths.Count == 1)
{
    sourceCode = File.ReadAllText(file);
}
else
{
    foreach (string file in paths)
    {
        foreach (string line in File.ReadAllLines(file))
        {
            sb.AppendLine(line);
        }
    }
    sourceCode = sb.ToString();
}
```

daxian-dbw (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
```diff
@@ -932,13 +910,22 @@ protected override void EndProcessing()
             {
                 // Load the source if they want to load from a file
                 if (String.Equals(ParameterSetName, "FromPath", StringComparison.OrdinalIgnoreCase) ||
-                    String.Equals(ParameterSetName, "FromLiteralPath", StringComparison.OrdinalIgnoreCase))
+                    String.Equals(ParameterSetName, "FromLiteralPath", StringComparison.OrdinalIgnoreCase)
+                   )
                 {
-                    this.sourceCode = "";
+                    StringBuilder sb = new StringBuilder(paths.Length);
```
daxian-dbw: Use `paths.Length` as the initial capacity is not very helpful. Maybe use a larger estimate?

iSazonov: Oops 😄

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
iSazonov: My thoughts was that we should get `CompileAssemblyFromFile` and there is no point in doing a lot of optimization.  But we can get it for a while, so I'll add the optimization. Only it will work for small files - `StringBuilder.MaxCapacity` is `int.MaxValue`.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
```diff
@@ -932,14 +910,52 @@ protected override void EndProcessing()
...
+                    }
+                    else
                     {
-                        this.sourceCode += System.IO.File.ReadAllText(file) + "\n";
+
+                        long initLength = 0;
+
+                        foreach (string file in paths)
+                        {
+                            FileInfo f = new FileInfo(file);
```
PaulHigin: I wonder what the cost is for obtaining file information.  I assume the file doesn't need to be opened?  Is the cost a reasonable trade off to StringBuilder automatic allocation?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
```diff
@@ -932,14 +910,52 @@ protected override void EndProcessing()
...
+                            initLength = f.Length;
+                        }
+
+                        if (initLength < int.MaxValue)
+                        {
+                            StringBuilder sb = new StringBuilder((int)initLength);
+
+                            foreach (string file in paths)
```
PaulHigin: Please add a comment that this pattern is used to minimize potential temporary LOH allocations.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
```diff
@@ -932,14 +910,52 @@ protected override void EndProcessing()
...
+                                {
+                                    sb.AppendLine(line);
+                                }
+                            }
+
+                            sourceCode = sb.ToString();
+                        }
+                        else
+                        {
+                            sourceCode = "";
```
PaulHigin: It was my understanding that ReadAllText() uses StringBuilder internally.  Do we know that it correctly handles strings larger than MaxInt characters?

PaulHigin (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
iSazonov: My bad. I removed the code.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
iSazonov: Comment added.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
iSazonov: I couldn't find out how it was implemented in CoreFX. 😕 
In any case, if we want to handle a lot of files, we need to refactor the cmdlet so that we don't read all the files in one string, but compile file by file. I plan to do it later

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
daxian-dbw: I don't understand your comment about `CompileAssemblyFromFile` (is it a class?). Could you please elaborate it a bit more? I quickly search about `CSharpSyntaxTree.ParseText`, and didn't find a good way to make it consume multiple files all together directly from files. That API takes a `SourceText` type, which has a static method `Create` that takes a file path, but not multiple files. But I'm sure there should be ways to do it, otherwise, how would a compiler to work 😄.

For some background, the perf issue of Add-Type was reported by Azure Profiler team. They find this exact code causes a lot LOH allocations in some services and if we treat the `path.length == 1` case specially, we can save, as quoted, "5.39 million samples here, or around 1% of total PowerShell.exe CPU cost". It was about the windows powershell, but this piece of code is same for powershell core.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
```diff
@@ -932,14 +910,43 @@ protected override void EndProcessing()
...
+
+                        long initLength = 0;
+
+                        // We replace 'ReadAllText' with 'StringBuilder' and 'ReadAllLines'
+                        // to avoide temporary LOH allocations.
+                        foreach (string file in paths)
+                        {
+                            FileInfo f = new FileInfo(file);
+                            initLength += f.Length;
+                        }
```
daxian-dbw: Calculate the total size might not be worth it. How about we just use a relatively large initial length? Say 8192 (8k)?
To be honest, I think if a user feed a file with more than 85k in size to Add-Type, they are using it wrongly (that Azure service might use `Add-Type` wrongly).

daxian-dbw (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
iSazonov: Could you ask Azure profile team about this? I guess they can have real things.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
iSazonov: >didn't find a good way to make it consume multiple files all together directly from files

I didn't find too. So I decided to postpone this to a later PR.
Really `CompileAssemblyFromFile` is in CodeDom. There isn't such method In CodeAnalysis.
[I found](https://stackoverflow.com/questions/31175881/unable-to-get-method-syntaxtree-parsefile-in-new-nuget-of-roslyn) using `SourceText.From`.
Maybe @lzybkr know a right way.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
daxian-dbw: Do you mean asking about the azure service might misuse Add-Type? @PaulHigin already replied them and brought up this to them.

iSazonov: I meant what is typical file size they see in their traces?

daxian-dbw: If you mean the initial capacity for StringBuilder, I think there is no accurate answer for it as it's probably a case by case basis. The default initial capacity is 16 bytes, so I think as long as we are using a relatively large number, it should be OK. (it's micro optimization that may be premature anyway)

PaulHigin: I dislike always pre-allocating StringBuilder to a large size.  It may be over allocated for the majority of cases.  But if we do this we should definitely ensure we pre-allocate at a size less than a LOH block.  I like the idea of pre-allocating intelligently but this means computing the file sizes and I am concerned that may have a high cost (I seem to remember a .Net bug  where FileInfo resulted in opening files on some platforms).

I am inclined to just leave it at the .Net default allocation.  To me the big perf win was removing the duplicate file read / string allocation in the base class.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
```diff
@@ -932,14 +910,43 @@ protected override void EndProcessing()
             {
                 // Load the source if they want to load from a file
                 if (String.Equals(ParameterSetName, "FromPath", StringComparison.OrdinalIgnoreCase) ||
-                    String.Equals(ParameterSetName, "FromLiteralPath", StringComparison.OrdinalIgnoreCase))
+                    String.Equals(ParameterSetName, "FromLiteralPath", StringComparison.OrdinalIgnoreCase)
+                   )
                 {
-                    this.sourceCode = "";
-                    foreach (string file in paths)
+                    if (paths.Length == 1)
```
PaulHigin: Also I don't think we need to special case this for a single file.  My understanding is that ReadAllText() uses StringBuilder internally in a similar way so there is no gain in using it over our StringBuilder use.

PaulHigin (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
iSazonov: .Net default (16 bytes) is very small for _file_ size. I think I agree that 8k is best in the case. Also we plan remove the code at all in follow PR so I ready put any value in the time :-)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
iSazonov: We plan remove the code at all in follow PR. We can keep it as is today, but I'll have that thought in follow PR.

PaulHigin (approved):

daxian-dbw (approved):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
daxian-dbw: > My understanding is that ReadAllText() uses StringBuilder internally in a similar way so there is no gain in using it over our StringBuilder use.

The gain we get in specializing `paths.length == 1` is to save a big string list that holds all lines and the big string array that is returned. See https://referencesource.microsoft.com/#mscorlib/system/io/file.cs,1018

After inspecting the API code, I start to suspect whether the `ReadAllLines()` API will actually save us from the LOH allocation because of the string list and the string array it converts to. It may double the LOH allocation ... Anyways, it's good that @iSazonov is going to eliminate the "reading text from file" code and instead try to compile the files directly.

daxian-dbw:
@iSazonov Thanks for spending effort redesign the code, looking forward to the new PR.
As for porting to windows, one thing we all need to know is that in Windows powershell, add-type goes through completely different code path (it uses CodeDom in full .NET). So your new PR most likely won't be ported back, but this micro-optimization change may be ported back.

PaulHigin:
@daxian-dbw, @iSazonov I just noticed that the AddTypeCommandBase abstract class is public.  So removing the duplicate sourceCode read from there could potentially be a breaking change if someone was relying on that functionality.  I feel it is an acceptable risk for Core 6 since it is unlikely anyone has a dependency on it.  But for Windows (back porting) it could adversely affect a customer.

iSazonov:
Good catch! We can avoid breaking change if we move AddTypeCommandBase.EndProcessing code in AddTypeCommand.EndProcessing. Makes a new PR?

PaulHigin:
@iSazonov Yes, moving the EndProcessing code into the base class seems like the right thing to do.

PR #5248, opened 2017-10-26
Sync tags in release build so that shallow cloning can be enabled.

TravisEz13:

### DIFF
--- a/tools/releaseBuild/Images/GenericLinuxFiles/PowerShellPackage.ps1
+++ b/tools/releaseBuild/Images/GenericLinuxFiles/PowerShellPackage.ps1
@@ -27,6 +27,9 @@ try {
     Set-Location $location
     Import-Module "$location/build.psm1"
     Import-Module "$location/tools/packaging"
+    
+    Write-Verbose "Sync'ing Tags..." -verbose
+    Sync-PSTags
 
     Start-PSBootstrap -Package -NoSudo
     Start-PSBuild -Crossgen -PSModuleRestore @releaseTagParam
--- a/tools/releaseBuild/Images/microsoft_powershell_windowsservercore/PowerShellPackage.ps1
+++ b/tools/releaseBuild/Images/microsoft_powershell_windowsservercore/PowerShellPackage.ps1
@@ -52,6 +52,10 @@ try{
     Import-Module "$location\build.psm1" -Force
     Import-Module "$location\tools\packaging" -Force
     $env:platform = $null
+
+    Write-Verbose "Sync'ing Tags..." -verbose
+    Sync-PSTags
+
     Write-Verbose "Bootstrapping powershell build..." -verbose
     Start-PSBootstrap -Force -Package
 


daxian-dbw:
The change looks good, but can you explain a bit about the shallow cloning?

adityapatwardhan (approved):

TravisEz13:
Shallow cloning allows cloning only the last few commits not the entire history of the repo.  It's standard in travis-ci.  We set it to a 1000 in travis-ci to virtually have all the history.  I can update this to turn on shallow cloning in CI as well.  It should speed up the clone and reduce the size of he repo on the build machine.  But the repo size is very important in the release build where we are copying it all over the place.

daxian-dbw:
`git clone --depth 1 --shallow-submodules` sounds interesting and will save footprint for sure.
However, I quickly tried the shallow clone and find `git describe --abbrev=60 --long` doesn't work. Here is what I did:
1. git clone --depth 1 --shallow-submodules https://github.com/PowerShell/PowerShell.git
2. cp .\fix\PowerShell\ .\ppp\ -Recurse
3. cd .\ppp\PowerShell
4. sync tags as `Sync-PSTags` does
5. run `git describe --abbrev=60 --long` or `git --git-dir="C:\arena\tmp\ppp\PowerShell\.git" describe --abbrev=60 --long`

I got the following when running step 5:
```
fatal: No tags can describe '4df703616ae7f3fe2e11484702dbc84b89c21a52'.
Try --always, or create some tags.
```
I verified that `git tag` does return all tags. Have you verified if this change would work?

TravisEz13:
With VSTS, I can only build from branches out of the PowerShell repo (without setting up a new build.)  This change itself is safe enough.  The shallow clone change is the dangerous part.  It would be in the build definition and can be changed back easily.  Since VSTS doesn't give me the exact command they use to do a shallow clone.  I don't see any real way of testing it.

daxian-dbw (approved):

daxian-dbw:
> this PR makes the release system work more like the CI system.

As far as I understanding, currently we clone recursively and then copy the repository folder to the image during `docker build`. So the tags should already be available as the recursive clone will got them.
So it seems to me if we don't enable the shallow clone, this change would become unnecessary.

adityapatwardhan:
@daxian-dbw  Are you still signed off? Is your last comment blocking?

daxian-dbw:
@adityapatwardhan It's not a blocking comment. Just a thought that if the shallow clone doesn't work, maybe this change should be reverted 😄

TravisEz13:
True, it might not be necessary, but I would like the scripts to be more similar to the workflow in CI.  The more the scripts have diverged the more likely a perceived small change in one is to cause them to behave differently.  Say, we add behavior that later steps depend on to sync-pstags.  It should still let me enable shallow cloning with depth 1000 (20% of our history) like in travis-ci.  The size reduction now is very small to none but will grow with time.

PR #5249, opened 2017-10-26
Make sure web hook gets called

TravisEz13:
if something goes wrong during the build, we might exit and not update the badge or fire the webhook.

Enable the `after_success` and `after_failure` stages of the travis-ci build so we can have a better chance of getting these done.

### DIFF
--- a/tools/travis.ps1
+++ b/tools/travis.ps1
@@ -271,6 +271,7 @@ elseif($Stage -eq 'Build')
                 Start-NativeExecution -sb {dotnet nuget push $package --api-key $env:NUGET_KEY --source "$env:NUGET_URL/api/v2/package"} -IgnoreExitcode
             }            
         }
+    }
 
     # if the tests did not pass, throw the reason why
     if ( $result -eq "FAIL" ) {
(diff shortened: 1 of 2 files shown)


tools/travis.ps1
```diff
@@ -270,32 +272,44 @@ else
...
+}
+elseif($Stage -in 'Failure', 'Success')
+{
+    $result = 'PASS'
+    if($Stage -eq 'Failure')
+    {
+        $result = 'FAIL'
+    }
+
+    if ((-not $isPr) -and $cronBuild) {
```
daxian-dbw: Is `-not $isPr` necessary? `$cronBuild` is true only if `$env:TRAVIS_EVENT_TYPE -eq 'cron'`.

daxian-dbw (changes requested):
I like this arrangement. It's cleaner. I left one comment.

adityapatwardhan (approved):
LGTM other than Dongbo's comment

tools/travis.ps1
TravisEz13: resolved.

daxian-dbw (approved):

Issue #4008, opened 2017-06-14
Stress tests needed


SteveL-MSFT:
- [ ] Publish RFC for stress plan
- [ ] Code coverage targets for stress tests
- [ ] Reuse existing Pester tests
- [ ] May need targeted tests for concurrency (runspaces)

anmenaga:
Ran old stress tests that were used for WindowsPS; results on PSCore were much better than what they used to be on WindowsPS.
Moving this to 6.1

PR #5258, opened 2017-10-27
Porting random hang fixes

PaulHigin:
These are two fixes for random hangs found internally.

The first fix (job.cs) is for data stream Close calls on PSDataCollection objects because the method also raises public events.  If an exception is thrown on the event handler then the thread never completes job clean up causing a hang.  Fix is to catch all thrown exceptions and report.

The second (invokecommandcommand.cs) fixes a hang scenario where a remote job goes into disconnected state while simultaneously handling an exception.  Fix is to use try/finally pattern to ensure Invoke-Command can always complete with jobs in the disconnected state.

### DIFF
--- a/src/System.Management.Automation/engine/remoting/client/Job.cs
+++ b/src/System.Management.Automation/engine/remoting/client/Job.cs
@@ -1542,14 +1542,25 @@ namespace System.Management.Automation
         /// </summary>
         internal void CloseAllStreams()
         {
-            if (_resultsOwner) _results.Complete();
-            if (_outputOwner) _output.Complete();
-            if (_errorOwner) _error.Complete();
-            if (_progressOwner) _progress.Complete();
-            if (_verboseOwner) _verbose.Complete();
-            if (_warningOwner) _warning.Complete();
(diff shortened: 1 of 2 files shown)


dantraMSFT (approved):

adityapatwardhan:
@PaulHigin Is it possible to add tests for this?

PaulHigin:
@adityapatwardhan No, not safely.  The code closes all opened runspaces when an "interactive host" runspaces closes, which unfortunately includes Pester runspaces.

PaulHigin:
BTW I have created an Issue with dotnet CLR about this (https://github.com/dotnet/coreclr/issues/14889).

PaulHigin:
@adityapatwardhan  Whoops, I though this was referring to the recent PowerShell hang on exit issue.  But again, the answer is not to tests since this is fixing rare random hangs.  I might be able to add test hooks, but I don't know if it is worth it.

SteveL-MSFT:
We should just take this PR into account when we develop reliability specific tests as part of https://github.com/PowerShell/PowerShell/issues/4008

PR #5263, opened 2017-10-28
fix sparse-checkout list

JamesWTruher:
Run powershell.exe in OpenCover since it will be in the path.
If there's an error in Start-CodeCoverageRun be sure to log as much as possible

### DIFF
--- a/test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
+++ b/test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
@@ -251,19 +251,19 @@ try
         # operations relative to where the test location is.
         # some tests rely on source files being available in $outputBaseFolder/test
         Push-Location $outputBaseFolder
+
         # clean up partial repo clone before starting
-        if ( Test-Path "$outputBaseFolder/.git" )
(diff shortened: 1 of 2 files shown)


test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
```diff
@@ -275,7 +278,7 @@ try
 
         Write-LogPassThru -Message "pulling sparse repo"
         "src" | Out-File -Encoding ascii .git\info\sparse-checkout -Force
-        "assets" | Out-File -Encoding ascii .git\info\sparse-checkout -Append
+        "/assets" | Out-File -Encoding ascii .git\info\sparse-checkout -Append
```
adityapatwardhan: Can you add a comment why '/' is required for assets and not for src.

adityapatwardhan (changes requested):

test/tools/CodeCoverageAutomation/Start-CodeCoverageRun.ps1
JamesWTruher: i added it for /src as that seems to work and is more clear

adityapatwardhan (approved):

PR #5265, opened 2017-10-28
[Feature] Add Multiple Link Header Support

### LINKED ISSUE #5257: Link header pagination only parsing first rel
Steps to reproduce
------------------

```pwsh
$uri_base = 'https://MYDEVORG.okta.com'
$apikey = 'MYAPIKEY'
$uri = $uri_base + '/api/v1/users/?filter=status eq "STAGED"&limit=200'
$method = 'GET'
$headers = @{'Accept'='application/json';'Content-Type'='application/json';'Authorization'="SSWS $apikey"}

PS /> $json = Invoke-RestMethod -Uri $uri -Method $method -Headers $headers -FollowRelLink
PS /> $json.Count
200
PS /> $webreq= Invoke-WebRequest -Uri $uri -Method $method -Headers $headers
PS /> $webreq.RelationLink

Key  Value
---  -----
self https://MYDEVORG.okta.com/api/v1/users?limit=200&filter=status+eq+%22STAGED%22


PS /> $webreq.Headers.Link
<https://MYDEVORG.okta.com/api/v1/users?limit=200&filter=status+eq+%22STAGED%22>; rel="self"
<https://MYDEVORG.okta.com/api/v1/users?after=OKTAUSERID&limit=200&filter=status+eq+%22STAGED%22>; rel="next"
PS />
```

Expected behavior
-----------------

Invoke-RestMethod should follow the next rel link and return additional results.
Invoke-WebRequest RelationLink dictionary should contain both "self" and "next".

Actual behavior
---------------

Only the first value from $webreq.Headers.Link is parsed into RelationLink.

Environment data
----------------

```
lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04.3 LTS
Release:        16.04
Codename:       xenial
```

```pwsh
> $PSVersionTable

OS                             Linux 4.4.0-43-Microsoft #1-Microsoft Wed Dec 31 14:42:53 PST 2014
```

References
----------------

Related commit: https://github.com/PowerShell/PowerShell/commit/40446c80a705f71260a9258ebc6477982789e82e
Related RFC: https://github.com/PowerShell/PowerShell-RFC/blob/master/2-Draft-Accepted/RFC0021-Link-header-based-pagination-for-WebCmdlets.md

markekraus:
fixes #5257 

Adds support for the following response Header scenario:

```none
Link: <https://MYDEVORG.okta.com/api/v1/users?limit=200&filter=status+eq+%22STAGED%22>; rel="self"
Link: <https://MYDEVORG.okta.com/api/v1/users?after=OKTAID&limit=200&filter=status+eq+%22STAGED%22>; rel="next"
```

CoreFX treats multiple header responses as separate array elements. Unless each array element is parsed, only the first one would be accepted. The current implimentation not only returns an incomplete `RelationLink` dictionary, but also breaks `-FollowRelLink` when a sever responds with multiple  Link headers. This PR corrects both the pagination and dictionary.

The test can currently only be done on Windows as we currently lack a native server that supports returning multiple headers with the same name (see #4639). The test would only result in a false pass on Linux and macOS.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
@@ -847,17 +847,20 @@ namespace Microsoft.PowerShell.Commands
             IEnumerable<string> links;
             if (response.Headers.TryGetValues("Link", out links))
             {
-                foreach(string link in links.FirstOrDefault().Split(","))
+                foreach (string linkHeader in links)
                 {
-                    Match match = Regex.Match(link, pattern);
-                    if (match.Success)
+                    foreach (string link in linkHeader.Split(","))
                     {
-                        string url = match.Groups["url"].Value;
-                        string rel = match.Groups["rel"].Value;
-                        if (url != String.Empty && rel != String.Empty && !_relationLink.ContainsKey(rel))
+                        Match match = Regex.Match(link, pattern);
+                        if (match.Success)
                         {
-                            Uri absoluteUri = new Uri(requestUri, url);
-                            _relationLink.Add(rel, absoluteUri.AbsoluteUri.ToString());
+                            string url = match.Groups["url"].Value;
+                            string rel = match.Groups["rel"].Value;
+                            if (url != String.Empty && rel != String.Empty && !_relationLink.ContainsKey(rel))
+                            {
+                                Uri absoluteUri = new Uri(requestUri, url);
+                                _relationLink.Add(rel, absoluteUri.AbsoluteUri.ToString());
+                            }
                         }
                     }
                 }
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -803,6 +803,24 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         $result.Output.RelationLink["last"] | Should BeExactly "http://localhost:8080/PowerShell?test=linkheader&maxlinks=5&linknumber=5"
     }
 
+    # Test pending support for multiple header capable server on Linux/macOS ses issue #4639
+    It "Validate Invoke-WebRequest returns valid RelationLink property with absolute uris if Multiple Link Headers are present" -Pending:$(!$IsWindows){
+        $headers = @{
+            Link =
+                '<http://localhost:8080/PowerShell?test=linkheader&maxlinks=5&linknumber=1>; rel="self"',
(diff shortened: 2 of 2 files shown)


test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -803,6 +803,24 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         $result.Output.RelationLink["last"] | Should BeExactly "http://localhost:8080/PowerShell?test=linkheader&maxlinks=5&linknumber=5"
     }
 
+    # Test pending support for multiple header capable server on Linux/macOS ses issue #4639
```
iSazonov: Typo - ses.

iSazonov (commented):
Leave a comment

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/WebRequestPSCmdlet.CoreClr.cs
```diff
@@ -847,17 +847,20 @@ internal void ParseLinkHeader(HttpResponseMessage response, System.Uri requestUr
...
-                        string rel = match.Groups["rel"].Value;
-                        if (url != String.Empty && rel != String.Empty && !_relationLink.ContainsKey(rel))
+                        Match match = Regex.Match(link, pattern);
+                        if (match.Success)
                         {
-                            Uri absoluteUri = new Uri(requestUri, url);
-                            _relationLink.Add(rel, absoluteUri.AbsoluteUri.ToString());
+                            string url = match.Groups["url"].Value;
+                            string rel = match.Groups["rel"].Value;
+                            if (url != String.Empty && rel != String.Empty && !_relationLink.ContainsKey(rel))
```
iSazonov: I don't know if `match.Success` is true enough to exclude an exception. It seems safer to check `String.IsNullOrEmpty()` for `url` and `rel`.

markekraus: > I don't know if match.Success

In this instance, it probably is. We are assured a string for both link and pattern so long as we don't mess with the pattern `Regex.Match` will never except and it true it the 2 indexes will exist in matches.

I agree on the `String.IsNullOrEmpty()` but the point of this PR is to address the Multiple Link Header issue, not to change this logic. Should that be done in a separate PR?

iSazonov: [Doc](https://msdn.microsoft.com/en-us/library/hchbwda2(v=vs.110).aspx) says that there is no null.

Closed.

PaulHigin (approved):

PR #5268, opened 2017-10-29
Remove AllScope from most default aliases

lzybkr:
To speed up scope creation, I removed AllScope from most default aliases.

This results in a 15-20% speedup for:
```powershell
function foo {}
for ($i = 0; $i -lt 100kb; $i++) { & { foo } }
```
I left AllScope of a few frequently used aliases because it does make command lookup faster. If we introduce something like dynamic sites for command lookup, then we could probably remove the rest
of the AllScope aliases.

This is a low risk breaking change. One can ask for aliases at a particular scope:
```powershell
Get-Alias -Scope 1 nsn
````
This could now fail if the scope number doesn't correspond to global scope.

### CODE REVIEW

**Comment by iSazonov** [comment]
Does it make sense to leave these aliases readonly? I think we talked about this earlier.

**Comment by lzybkr** [comment]
That's 2 questions.

I don't feel strongly about using `ReadOnly`. I think it's unnecessary, but also don't see the harm.

As for using the `Alias` attribute - I thought there was an unrelated reason to not use the attribute - but I don't recall. If it was simply the lack of ability to specify scope options, the attribute could be enhanced to support that.

**Comment by iSazonov** [comment]
If we remove ReadOnly here we can safely move to `Alias()` attribute.
I found the @daxian-dbw  related [comment(s)](https://github.com/PowerShell/PowerShell/pull/3595#issuecomment-296239341) and [comment](https://github.com/PowerShell/PowerShell/pull/3595#issuecomment-300950380).

After PR 3595 I tried to enhance Alias attribute in [the my branch](https://github.com/iSazonov/PowerShell/tree/add-options-to-alias-attribute). I couldn't implement AllScope and froze the work. But ReadOnly works. If we don't remove ReadOnly here after deletion of so much AllScope, maybe it makes sense to finish the work for ReadOnly here or I can push new PR later if you approve.

**Comment by SteveL-MSFT** [comment]
If we remove AllScope from all the aliases, what perf impact would it be to the commonly used ones?  Also, it seems like we should have an internal overload of SessionStateAliasEntry constructor rather than passing empty string everytime for the description.

File: src/System.Management.Automation/engine/InitialSessionState.cs [hunk 1]
```diff
@@ -4552,326 +4552,181 @@ internal static SessionStateAliasEntry[] BuiltInAliases
...
+                    new SessionStateAliasEntry("fc", "Format-Custom", "", ReadOnly),
+                    new SessionStateAliasEntry("kill", "Stop-Process"),
+                    new SessionStateAliasEntry("pwd", "Get-Location"),
+                    new SessionStateAliasEntry("type", "Get-Content"),
                     // Native commands we keep because the functions act correctly on Linux
-                    new SessionStateAliasEntry("clear",
-                        "Clear-Host",      "", ScopedItemOptions.AllScope),
+                    new SessionStateAliasEntry("clear", "Clear-Host"),
 //#if !CORECLR is used to disable aliases for cmdlets which are not available on OneCore
 #if !CORECLR
```
SteveL-MSFT: Should we just remove the !CORECLR section?

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -4552,326 +4552,181 @@ internal static SessionStateAliasEntry[] BuiltInAliases
...
-                        "Stop-Job",  "", ScopedItemOptions.AllScope),
-                    new SessionStateAliasEntry("wjb",
-                        "Wait-Job",  "", ScopedItemOptions.AllScope),
+                    new SessionStateAliasEntry("gjb", "Get-Job"),
+                    new SessionStateAliasEntry("rcjb", "Receive-Job"),
+                    new SessionStateAliasEntry("rjb", "Remove-Job"),
+                    new SessionStateAliasEntry("sajb", "Start-Job"),
+                    new SessionStateAliasEntry("spjb", "Stop-Job"),
+                    new SessionStateAliasEntry("wjb", "Wait-Job"),
 #if !CORECLR
```
SteveL-MSFT: Should we remove the !CORECLR section?

SteveL-MSFT (changes requested):

src/System.Management.Automation/engine/InitialSessionState.cs
lzybkr: I considered that - but maybe a couple still make sense - ise e.g. - and would `Out-Printer` come back?

src/System.Management.Automation/engine/InitialSessionState.cs
lzybkr: I thought most of these cmdlets might be ported to CORECLR, so left it.

lzybkr:
I considered adding the overload, but if we're fine with removing `ReadOnly`, we won't even need that.

lzybkr:
As for perf impact on the common commands - it depends a lot, but command lookup isn't exactly cheap - there is a 25% difference in perf between in the two examples below - in the first, command lookup happens just once, in the second, it happens on every invocation.

```powershell
function foo {}
measure-command { & {
    $c = Get-Command foo
    for ($i = 0; $i -lt 100kb; $i++) { & $c }
} }
measure-command { & {
    for ($i = 0; $i -lt 100kb; $i++) {foo }
} }
```

src/System.Management.Automation/engine/InitialSessionState.cs
SteveL-MSFT: It's possible, so I guess it's best to leave it for now.

src/System.Management.Automation/engine/InitialSessionState.cs
SteveL-MSFT: same as above

SteveL-MSFT (approved):

Issue #4986, opened 2017-10-03
Enhance PowerShell full command line help


iSazonov:
- [ ] #5780 We should show the full/short command line help in Bash style if a parameter is wrong.
- [x] #4989 We should reformat the help in Bash style:
    - sort parameter list in alphabetical order.
    - replace long header on  powershell[.exe] [options] 
- [x] #4989 We should add more examples in line of the help.
- [x] #5275  Add "To get help type Help or Help followed by the command name, such as: Help Get-Process" in PowerShell banner bottom.

The Issue consider command line only  powershell -Help  (in bash or cmd.exe).

markekraus:
@iSazonov this is for the PowerShell binary, correct?

> We should show the full command line help in Bash style if a parameter is wrong.

Most Linux commands show usage and how to obtain help (`-h`, `--help`, `man`, etc.) on incorrect parameters unless their usage is very simple. Full help leads to the previous command going off console. command history is not always a given in the Linux world. I would prefer short syntax help and how to obtain the full help when I use the wrong parameters.

iSazonov:
@markekraus The Issue is for discussion to collect ideas. :-)

I like short help on error too. We mean an interactive session where typos is a common matter - full help will going to annoy.

markekraus:
@iSazonov  by "We mean an interactive session where typos is a common matter"
Do you mean an interactive session in PowerShell?

example:

```
c:\> Get-Service -Nmae Get-Command
Get-Service : A parameter cannot be found that matches parameter name 'Nmae'.
At line:1 char:13
+ Get-Service -Nmae Get-Command
+             ~~~~~

SYNTAX
    Get-Service [[-Name] <String[]>] [-ComputerName <String[]>] [-DependentServices] [-Exclude <String[]>] [-Include <String[]>] [-RequiredServices] [<CommonParameters>]

    Get-Service [-ComputerName <String[]>] [-DependentServices] [-Exclude <String[]>] [-Include <String[]>] [-RequiredServices] -DisplayName <String[]> [<CommonParameters>]

    Get-Service [-ComputerName <String[]>] [-DependentServices] [-Exclude <String[]>] [-Include <String[]>] [-InputObject <ServiceController[]>] [-RequiredServices] [<CommonParameters>]

For more help on Get-Service run:
Get-Help -Name Get-Service -Full
```

iSazonov:
No, the Issue consider command line only `powershell -h`.

markekraus:
[...]

PR #5275, opened 2017-10-30
Add help strings in PowerShell banner

iSazonov:
Related #4986.

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
+++ b/src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
@@ -121,8 +121,8 @@
     <value>PowerShell {0}
 Copyright (c) Microsoft Corporation. All rights reserved.
 
-To get help type Help or Help followed by a command name, such as: Help Get-Process
-To get PowerShell Online Help open the link: https://aka.ms/pscore6-docs</value>
+Type 'Help' to get help.
+Type 'Start-Process https://aka.ms/pscore6-docs' to get Online Help.</value>
   </data>
   <data name="ShellHelp" xml:space="preserve">
     <value>Usage: pwsh[.exe] [[-File] &lt;filePath&gt; [args]]
--- a/src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
+++ b/src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
@@ -119,7 +119,10 @@
   </resheader>
   <data name="ShellBannerNonWindowsPowerShell" xml:space="preserve">
     <value>PowerShell {0}
-Copyright (c) Microsoft Corporation. All rights reserved.</value>
+Copyright (c) Microsoft Corporation. All rights reserved.
+
+To get help type Help or Help followed by a command name, such as: Help Get-Process
+To get PowerShell Online Help open the link: https://aka.ms/pscore6-docs</value>
   </data>
   <data name="ShellHelp" xml:space="preserve">
     <value>Usage: pwsh[.exe] [[-File] &lt;filePath&gt; [args]]
--- a/src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
+++ b/src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
@@ -121,8 +121,8 @@
     <value>PowerShell {0}
 Copyright (c) Microsoft Corporation. All rights reserved.
 
-Type 'Help' to get help.
-Type 'Start-Process https://aka.ms/pscore6-docs' to get Online Help.</value>
+https://aka.ms/pscore6-docs
+Type 'help' to get help.</value>
   </data>
   <data name="ShellHelp" xml:space="preserve">
     <value>Usage: pwsh[.exe] [[-File] &lt;filePath&gt; [args]]


src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
```diff
@@ -119,7 +119,10 @@
   </resheader>
   <data name="ShellBannerNonWindowsPowerShell" xml:space="preserve">
     <value>PowerShell {0}
-Copyright (c) Microsoft Corporation. All rights reserved.</value>
+Copyright (c) Microsoft Corporation. All rights reserved.
+
+To get help type Help or Help followed by a command name, such as: Help Get-Process
```
markekraus: Perhaps instead of giving an example just recommend typing `Help`. the default help topic is about using the help system.

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
```diff
@@ -119,7 +119,10 @@
...
+To get PowerShell Online Help open the link: https://aka.ms/pscore6-docs</value>
```
markekraus: As much as I hate to add another new line, I very much would like to have URLs on their own line. That makes them easier to grab and for limited width consoles it reduces the risk of it being line wrapped. 

```none
For PowerShell Help online, visit:
https://aka.ms/pscore6-docs
```

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
SteveL-MSFT: I agree that we can simply say:

```
Type `help` to get help.
```

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
SteveL-MSFT: I would prefer this link to be in the help text since the link isn't clickable

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
iSazonov: What about:
```
Type 'Help' to get help.
Type 'Start-Process https://aka.ms/pscore6-docs' to get Online Help. 
```

markekraus: `Start-Process` makes no sense on linux with a URL.
The link is clickable in KiTTY, so that kind of depends on the console app. 
I see benefit in the link being displayed in the banner, but only if the URL is on a line by itself. Though, having it in the default help topic is a good alternative. As a user, I am not a fan of apps with long wordy banners.

iSazonov: I don't like long banners too but the idea is to follow Bash-style and to be web/Azure-friendly.
I expect `Start-Process` works on Linux by mean of XDS. I don't know about MacOs.

markekraus: > I expect Start-Process works on Linux

Assuming you are in a Linux GUI, maybe. it most certainly doesn't work in any of the ~20 Linux VMs I have `pwsh` installed on. It definitely wont work raspberry pi. It is better to assume a user is not in a GUI with Core, IMO. 

Providing just the link in an easy to read, select, copy and with the least risk of line-wrapping is the best bet if we are to provide it in the banner at all, IMO. Might not even need an explanation as the URL is decently self-explanatory:

```none

https://aka.ms/pscore6-docs
Type 'Help' to get help.
```

Bash doesn't have any enforced banners from bash itself, I don't think. If we follow bash style, it will show nothing unless the user/system admin deems it necessary. So I'm not sure what you mean by bash-style. Can you provide an example? (and lets not go by python's REPL, that thing is painful)

iSazonov: Sorry, skip my comment about Bash banner - I means whole original Issue in wrong context.

rkeithhill: Why upcase the `H` i.e. why not `Type 'help' to get help.`?  That feels more *nixy.  :-)

Also, someone should consider updating the `Get-Help` topic because when you execute `help` (or `man` on Windows) and get this topic that talks all about `Get-Help`, it can be a bit confusing.  I think the LONG DESCRIPTION should mention the fact that:

> `help` exists as a function wrapper to the Get-Help command that pages the output to the console.  On Windows, `man` is an alias for the `help` function.  On Windows, the default pager is more.com.  On other systems, it is less.  You can select a different pager utility with the `$env:PAGER` environment variable.  For example, if you have less on your Windows system you
can set $env:PAGER to the path to less and the help function will use less to page Get-Help output.

SteveL-MSFT:
I would be fine with just:

```none

https://aka.ms/pscore6-docs
Type 'help' to get help.
```

with the lowercase `h`

cc @joeyaiello @HemantMahawar

markekraus (approved):

SteveL-MSFT (approved):

PR #5287, opened 2017-10-31
Conversions from PSMethod to Delegate

powercode:
Underpinnings to make calling of Extension methods /Linq
easier from PowerShell.

Enables the following that previously had to be done via
reflection.

```powershell
class M {
    static [int] DoubleStrLen([string] $value) { return 2 * $value.Length }
    
    static [long] AggregateString([string[]] $values, [func[string, int]] $selector) {
        [long] $res = 0
        foreach($s in $values){
            $res += $selector.Invoke($s)
        }
        return $res
    }        
}

[M]::AggregateString((gci).Name, [M]::DoubleStrLen)
```

Each PSMethod is created as with a unique type for the combinations of
method signatures in the MethodInfos it represents.

PSMethod<T> where T is a MethodGroup<>, potentially recursive in the
last template argument.

This way, we can determine by just looking at the type of a PSMethod
if there exists a conversion from the PSMethod to a delegate.


test/powershell/Language/Parser/Conversions.Tests.ps1
```diff
@@ -31,27 +31,30 @@ Describe 'conversion syntax' -Tags "CI" {
                 ## would become `, [List[int]]@(1,2)`, which is more confusing than `$result = [List[int]]@(1,2)`.
                 ## This is why the current form of `$result = [List[int]]@(1,2)` is used intentionally here.
```
powercode: VSCode formats this.
Is it OK to keep this formatting?
Or is there a setting that aligns one line hashtables?

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -2642,8 +2643,326 @@ public override string TypeNameOfValue
...
+                case 62: return typeof(MethodGroup<,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,>).MakeGenericType(types[0], types[1], types[2], types[3], types[4], types[5], types[6], types[7], types[8], types[9], types[10], types[11], types[12], types[13], types[14], types[15], types[16], types[17], types[18], types[19], types[20], types[21], types[22], types[23], types[24], types[25], types[26], types[27], types[28], types[29], types[30], CreateMethodGroup(types, 31, 31));
+                case 63: return typeof(MethodGroup<,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,>).MakeGenericType(types[0], types[1], types[2], types[3], types[4], types[5], types[6], types[7], types[8], types[9], types[10], types[11], types[12], types[13], types[14], types[15], types[16], types[17], types[18], types[19], types[20], types[21], types[22], types[23], types[24], types[25], types[26], types[27], types[28], types[29], types[30], CreateMethodGroup(types, 31, 32));
+
+            }
+            throw Assert.Unreachable;
+        }
     }
 
+
+
```
iSazonov: Please remove extra line.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -2642,8 +2643,326 @@ public override string TypeNameOfValue
...
+    class PSOutParameter<T> { private PSOutParameter() { }}
+
+    abstract class PSGenericType
+    {
+
```
iSazonov: Please remove extra line.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -2642,8 +2643,326 @@ public override string TypeNameOfValue
...
+    internal class MethodGroup<T1, T2, T3, T4, T5, T6, T7, T8> : MethodGroup { }
+    internal class MethodGroup<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16> : MethodGroup { }
+    internal class MethodGroup<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21, T22, T23, T24, T25, T26, T27, T28, T29, T30, T31, T32> : MethodGroup { }
+
+    class Unit
+    {
+        private Unit() { }
+    }
+
+
```
iSazonov: Please remove extra line.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -2642,8 +2643,326 @@ public override string TypeNameOfValue
...
+                case 13: return typeof(PSGenericType13);
+                case 14: return typeof(PSGenericType14);
+                case 15: return typeof(PSGenericType15);
+                case 16: return typeof(PSGenericType16);
+            }
+            throw Assert.Unreachable;
+        }
+    }
+
+    class PSGenericType0 : PSGenericType { private PSGenericType0() { } }
```
iSazonov: Private class? Please use explicit modifer.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -2642,8 +2643,326 @@ public override string TypeNameOfValue
...
+
+    internal abstract class MethodGroup { }
+    internal class MethodGroup<T1> : MethodGroup { }
+    internal class MethodGroup<T1, T2> : MethodGroup { }
+    internal class MethodGroup<T1, T2, T3, T4> : MethodGroup { }
+    internal class MethodGroup<T1, T2, T3, T4, T5, T6, T7, T8> : MethodGroup { }
+    internal class MethodGroup<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16> : MethodGroup { }
+    internal class MethodGroup<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21, T22, T23, T24, T25, T26, T27, T28, T29, T30, T31, T32> : MethodGroup { }
+
+    class Unit
```
iSazonov: Private class?

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -2642,8 +2643,326 @@ public override string TypeNameOfValue
...
+            _current = -1;
+        }
+
+        public bool MoveNext()
+        {
+            _current++;
+            return MoveNext(_t, _current);
+        }
+
+        bool MoveNext(Type type, int index)
```
iSazonov: Public? Private?

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -2642,8 +2643,326 @@ public override string TypeNameOfValue
...
+            if (index >= genericTypeArguments.Length)
+            {
+                Current = null;
+                return false;
+            }
+            Current = t;
+            return true;
+        }
+
+
```
iSazonov: Please remove extra line.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -2642,8 +2643,326 @@ public override string TypeNameOfValue
...
+        /// <summary>
+        /// Helper factory function since we cannot bind a delegate to a ConstructorInfo.
+        /// </summary>
+        internal static PSMethod<T> Create(string name, Adapter adapter, object baseObject, object adapterData,
+            bool isSpecial, bool isHidden)
+        {
+            return new PSMethod<T>(name, adapter, baseObject, adapterData, isSpecial, isHidden);
+        }
+
```
iSazonov: Please remove extra line.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -2642,8 +2643,326 @@ public override string TypeNameOfValue
...
+    }
+
+
```
iSazonov: Please remove extra line.

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -3212,6 +3212,56 @@ private static PSConverter<bool> CreateNumericToBoolConverter(Type fromType)
...
+                        continue;
+                    }
+                    if (comparator.SignatureMatches(candidate.ReturnType, candidate.GetParameters()))
+                    {
+                        return maybeType != null
+                            ? candidate.CreateDelegate(resultType)
+                            : candidate.CreateDelegate(resultType, psMethod.instance);
+                    }
+                }
+                var msg = $"No matching overload found from {psMethod} to {resultType}";
```
lzybkr: Strings that appear in error messages need to be in a resource for localization.

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -4739,9 +4791,87 @@ internal static object ThrowInvalidConversionException(object valueToConvert, Ty
...
+            {
+                return ReturnTypeMatches(returnType) && ParameterTypeMatches(arguments);
+            }
+
+            private bool ReturnTypeMatches(Type returnType)
+            {
+                return _returnType == typeof(void) ?  returnType == typeof(Unit) : _returnType.IsAssignableFrom(returnType);
+            }
+
+            private bool ParameterTypeMatches(ParameterInfo[] arguments)
```
lzybkr: ParameterType**s**Matches?

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -2642,8 +2643,326 @@ public override string TypeNameOfValue
...
         /// </summary>
         internal bool IsSpecial { get; private set; }
+
+
+        internal static PSMethod Create(string name, DotNetAdapter dotNetInstanceAdapter, object baseObject, DotNetAdapter.MethodCacheEntry method)
+        {
+            return Create(name, dotNetInstanceAdapter, baseObject, method, false, false);
+        }
+
+        internal static PSMethod Create(string name, DotNetAdapter dotNetInstanceAdapter, object baseObject, DotNetAdapter.MethodCacheEntry method, bool isSpecial, bool isHidden)
```
lzybkr: I'm considering imposing a limit on line length :) - working on a laptop this week and GitHub isn't letting me scroll to see the rest of this line.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -2642,8 +2643,326 @@ public override string TypeNameOfValue
...
+                case 62: return typeof(MethodGroup<,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,>).MakeGenericType(types[0], types[1], types[2], types[3], types[4], types[5], types[6], types[7], types[8], types[9], types[10], types[11], types[12], types[13], types[14], types[15], types[16], types[17], types[18], types[19], types[20], types[21], types[22], types[23], types[24], types[25], types[26], types[27], types[28], types[29], types[30], CreateMethodGroup(types, 31, 31));
+                case 63: return typeof(MethodGroup<,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,>).MakeGenericType(types[0], types[1], types[2], types[3], types[4], types[5], types[6], types[7], types[8], types[9], types[10], types[11], types[12], types[13], types[14], types[15], types[16], types[17], types[18], types[19], types[20], types[21], types[22], types[23], types[24], types[25], types[26], types[27], types[28], types[29], types[30], CreateMethodGroup(types, 31, 32));
+
+            }
+            throw Assert.Unreachable;
```
lzybkr: You should have a real error or at least NotSupported exception.
And FYI - I've seen real code with > 300 parameters - code from a company I'd like to name but won't. :)

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -2642,8 +2643,326 @@ public override string TypeNameOfValue
...
+                case 9: return typeof(PSGenericType9);
+                case 10: return typeof(PSGenericType10);
+                case 11: return typeof(PSGenericType11);
+                case 12: return typeof(PSGenericType12);
+                case 13: return typeof(PSGenericType13);
+                case 14: return typeof(PSGenericType14);
+                case 15: return typeof(PSGenericType15);
+                case 16: return typeof(PSGenericType16);
+            }
+            throw Assert.Unreachable;
```
lzybkr: Real error here - and must resist urge to think of template metaprogramming to solve this problem.

test/powershell/Language/Parser/Conversions.Tests.ps1
```diff
@@ -31,27 +31,30 @@ Describe 'conversion syntax' -Tags "CI" {
...
 
-                @{ Command = {$result = [Collections.Generic.List[int]]@(1)};       CollectionType = 'List`1'; ElementType = "Int32";  Elements = @(1) }
-                @{ Command = {$result = [Collections.Generic.List[int]]@(1,2)};     CollectionType = 'List`1'; ElementType = "Int32";  Elements = @(1,2) }
-                @{ Command = {$result = [Collections.Generic.List[int]]"4"};        CollectionType = 'List`1'; ElementType = "Int32";  Elements = @(4) }
-                @{ Command = {$result = [Collections.Generic.List[int]]@("4","5")}; CollectionType = 'List`1'; ElementType = "Int32";  Elements = @(4,5) }
+                @{ Command = {$result = [Collections.Generic.List[int]]@(1)}; CollectionType = 'List`1'; ElementType = "Int32"; Elements = @(1) }
```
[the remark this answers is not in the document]
lzybkr: I find `git add -p` very useful to add just my changes and not the formatting changes.

lzybkr (changes requested):
Looks good - just a couple of small complaints.

This must have been a lot of fun to implement - nice work.

src/System.Management.Automation/engine/MshMemberInfo.cs
powercode: 4k 27" monitor may have desensitized me to the issue :)

src/System.Management.Automation/engine/MshMemberInfo.cs
powercode: Can this be a problem? I don't really like having these arbitrary constants as limitations.
Is that 300 generic type parameters?

I took some of the limitations from the choises already made for Func/Action, but 63 was just a bit higher than the 58, which was the hightest number of overloads any type in the assemblies loaded by Windows Powershell used.

I am very open to suggestions here.

src/System.Management.Automation/engine/MshMemberInfo.cs
powercode: yeah - templates != generics. Or do you have some tricks up your sleave that I don't know about?

Hardcoded up to a limit, and dynamic type generation after that?

src/System.Management.Automation/engine/MshMemberInfo.cs
lzybkr: Oh, not type parameters, just normal. But I think what you have is fine, just need a better error.

src/System.Management.Automation/engine/CoreAdapter.cs
```diff
@@ -2353,6 +2353,10 @@ internal DotNetAdapter(bool isStatic)
         internal class MethodCacheEntry
         {
             internal MethodInformation[] methodInformationStructures;
+            /// <summary>
+            /// cache delegate to the ctor of PSMethod&lt;&gt; with a template parameter derived from the methodInformationsStructures
```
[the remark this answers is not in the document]
powercode: `PSMethod<>` is the notation for the open generic type.

src/System.Management.Automation/engine/MshMemberInfo.cs
lzybkr: Yeah, but hard coded is fine. MutableTuple has an interesting approach if you want to explore it - basically you have:
PSGenericType<T>
PSGenericType<T1, T2>
PSGenericType<T1, T2, T3, T4>
PSGenericType<T1, T2, T3, T4, T5, T6, T7, T8>
etc.

And if your list is too long, you encode the last parameter as another PSGenericType<...>.
If you don't have a power of 2, you use placeholder types like you've done with `Unit`.

src/System.Management.Automation/engine/MshMemberInfo.cs
powercode: private is not allowed as a modifier. It is implicit.

powercode:
I would still like a discussion about what's the right thing to do methods with more than 63 overloads and or more than 16 generic parameters.

These are for sure edge cases, but it would be nice to at least handle them gracefully

lzybkr:
@powercode - if you want full generality, as long as `MethodGroup` is an internal type, you can have a special case where, if the last type parameter is `MethodGroup`, then you have essentially a linked list, and the type parameters to that last `MethodGroup` is the continuation of the list.

To be concrete, say we only support 1 or 2 methods in `MethodGroup`, but you have 3 overloads. You would encode that as:

```
MethodGroup[
    Func[int, int],
    MethodGroup[
        Func[string, string],
        Func[long, long]]]
```

Using this trick, you could use a single generic type to represent all generic type parameters, the length of the list would encode T0, T1, etc.

```
PSGenericType[unit] # T0
PSGenericType[PSGenericType[unit]] # T1
PSGenericType[PSGenericType[PSGenericType[unit]]] # T2
# etc.
```

src/System.Management.Automation/engine/MshMemberInfo.cs
powercode: I actually already do that to an extent with MethodGroup. Maybe I could rewrite it to be more general.

powercode:
@lzybkr I kinda already to that:
```
case 7: return typeof(MethodGroup<,,,>).MakeGenericType(types[0], types[1], types[2], CreateMethodGroup(types, 3, 4));
```

powercode:
I just learned that I need to have a representation of Span<T> too. That special class cannot be used as a generic argument. https://github.com/dotnet/corefxlab/blob/master/docs/specs/span.md

powercode:
@lzybkr  I will explore the `PSGenericType[unit]` route in the weekend. I liked the idea better than the hardcoded types. Or maybe a combination, that if more than `n` parameters, use the template

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -2642,8 +2643,403 @@ public override string TypeNameOfValue
...
+    internal class MethodGroup<T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, T11, T12, T13, T14, T15, T16, T17, T18, T19, T20, T21, T22, T23, T24, T25, T26, T27, T28, T29, T30, T31, T32> : MethodGroup { }
+
+    class Unit
+    {
+        private Unit() { }
+    }
+
+    internal struct PSMethodSignatureEnumerator : IEnumerator<Type>
+    {
+        private int _current;
```
lzybkr: `_current` is not the backing field for `Current`, you should rename `_current` to avoid any confusion.

src/System.Management.Automation/engine/MshMemberInfo.cs
lzybkr: Is there a good reason for this private method? It's only called in one place.

src/System.Management.Automation/engine/MshMemberInfo.cs
powercode: good point

lzybkr (changes requested):
I just had one naming nit that should be addressed.

lzybkr (approved):

src/System.Management.Automation/resources/ExtendedTypeSystem.resx
```diff
@@ -370,4 +370,7 @@ PS&gt; [System.Collections.Generic.Comparer``1]::get_Default()
   <data name="InvalidCastExceptionNonCoreType" xml:space="preserve">
     <value>Cannot convert value to type "{0}". Only core types are supported in this language mode.</value>
   </data>
-</root>
+  <data name="PSMethodToDelegateNoMatchingOverLoad" xml:space="preserve">
+    <value>No matching overload found from {0} to {1}</value>
+  </data>
+</root>
```
iSazonov: Please add new line at EOF.

test/powershell/Language/Parser/Conversions.Tests.ps1
```diff
@@ -127,3 +127,274 @@ namespace TestTypeResolution {
...
+        [Func[timespan, timespan, timespan, timespan, timespan, timespan]] $f69 = [M]::Foo
+        $f69.Invoke($timeSpan, [Timespan]::Zero, [Timespan]::Zero, [Timespan]::Zero, [Timespan]::Zero) | Should BE $timeSpan
+        [Func[timespan, timespan, timespan, timespan, timespan, timespan, timespan]] $f70 = [M]::Foo
+        $f70.Invoke($timeSpan, [Timespan]::Zero, [Timespan]::Zero, [Timespan]::Zero, [Timespan]::Zero, [Timespan]::Zero) | Should BE $timeSpan
+        [Func[timespan, timespan, timespan, timespan, timespan, timespan, timespan, timespan]] $f71 = [M]::Foo
+        $f71.Invoke($timeSpan, [Timespan]::Zero, [Timespan]::Zero, [Timespan]::Zero, [Timespan]::Zero, [Timespan]::Zero, [Timespan]::Zero) | Should BE $timeSpan
+        [Func[timespan, timespan, timespan, timespan, timespan, timespan, timespan, timespan, timespan]] $f72 = [M]::Foo
+        $f72.Invoke($timeSpan, [Timespan]::Zero, [Timespan]::Zero, [Timespan]::Zero, [Timespan]::Zero, [Timespan]::Zero, [Timespan]::Zero, [Timespan]::Zero) | Should BE $timeSpan
+    }
+
```
iSazonov: Please remove extra line.

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -3212,6 +3212,56 @@ private static PSConverter<bool> CreateNumericToBoolConverter(Type fromType)
                 valueToConvert.ToString(), resultType.ToString(), exception.Message);
         }
 
+
```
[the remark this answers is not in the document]
iSazonov: Up.

iSazonov (approved):
LGTM with 3 minor comment.

daxian-dbw (commented):
Sorry for this late review. I will address the comments myself and also start a feature test run.

daxian-dbw:
A few further improvement points
--------------------------------------
1. `DelegateArgsComparator.ParameterTypesMatches` doesn't consider the Variance in delegate parameter types. A method signature `int CallMe(object input)` should be able to be converted to `Func<string, int>`.
2. `PSMethod.MatchesPSMethodProjectedType` doesn't handle the case where both `targetType` and `projectedSourceType` are the same `ByRef` type or `Pointer` type.
3. I think the 'Unit' type name should be changed to 'VOID'. Use type `Unit` to represent `void` is no intuitive -- I misread it to `uint` when first looking at the code.
4. I don't get the point of having `ReplaceGenericTypeArgumentsWithMarkerTypes`. After making a generic method using our synthetic reference/value types, the generated `MethodInfo` signature won't match any real delegates except in one condition -- the original generic method doesn't use the generic type argument for any of its parameters or return type -- I think it's rare for this condition to be met for a generic method.

Two questions about the design
----------------------------------------------
1. It's eye-opening for me to see how you can track the signature information using the generic type argument of `PSMethod<>` 😮 But why do you choose the current design over adding a `List<Type>` field/property to `PSMethod` to keep all the `Func<...>` delegate types? 
With a `List<Type>` field/property, you don't need the algorithm to create `MethodGroup<...>` or to enumerate the `Func<...>` type from the `MethodGroup<...>` type, and we won't need to make that many generic types, and the code would be easier to understand too.
2. I agree that we need pseudo types to represent `ByRef` type (ref/out), `Pointer` type `void` type and even `TypeReference`, but why do we need `PSEnum<>` to wrap enum types?

lzybkr:
@daxian-dbw - The use of `PSMethod<>` allows for faster conversions.

If you have a conversion from `PSMethod` to `Action`, `Action<int>`, `Action<int,int>`, etc., then it requires a more expensive check to determine the conversion validity.

If it is type based, then there is no conversion from `PSMethod` to `Action<int>`, but there is for `PSMethod<MethodGroup<Action<int>>>` to `Action<int>`.

Put another way, the `FigureConversion` logic essentially relies on types. If you look closely, you'll see that value dependent conversions do exist, but the design only accommodates them because I couldn't see a better way to support that rarely used scenario.

daxian-dbw:
@lzybkr That makes perfect sense now. I forgot about the conversion cache.

- Without `PSMethod<>` type, the cache is useless for this kind of conversion -- we will end up doing the validity check on every `PSMethod -> Delegate` conversion.
- With the `PSMethod<>` type, no duplicate validity check is needed. For invalid 
`PSMethod<> -> Delegate` conversions, the conversion cache will have a `ConvertNoConversion` for the `FromType, ToType` tuples, powershell won't need to do the check again.

Can you please also shed some light on the `PSEnum<>` question?

daxian-dbw:
The failed tests in macOS build are known test failures. It's tracked by #5590.

lzybkr:
@powercode  will have to address `PSEnum<>` - he just took my basic idea and worked out all the details - and I do mean all, I might have skipped `Pointer` because you can't really use it from PowerShell.

daxian-dbw:
@powercode I will merge this PR, but let's continue the discussion if you have time 😄

powercode:
Sorry all for my horrible response time. I have to revisit the code and look at it. Some of the special cases, like `pointer`, never shows up from PowerShell, but I ran into them by enumerating all methods in all loaded assemblies and tried to make sure I could create `PSMethod<>`s for them.

iSazonov:
Should we document the feature?

PR #5291, opened 2017-11-01
Enable macOS launcher

### LINKED ISSUE #5262: Remove the workaround to enable macOS launcher
The change was reverted to unblock PRs being blocked in packaging. 

https://github.com/PowerShell/PowerShell/pull/5259

Further investigation should be done to re-enable macOS launcher.

thezim:
Notable change this time around.

* Values set by script are "unset" in `Info.plist` by default to prevent conflicts if a fresh clone.
* After packaging `CFBundleIdentifier` is set to random value to prevent it from being picked up by installer. An empty value could not be used because native defaults doesn't allow it.
* The magic here is getting macOS to recognize the change in the plist file. `plutil` is enough to make this happen.

Fixes #5262

### DIFF
--- a/tools/packaging/packaging.psm1
+++ b/tools/packaging/packaging.psm1
@@ -648,7 +648,7 @@ function New-UnixPackage {
                 # Set values in plist.
                 $plist = "$macosapp/Contents/Info.plist"
                 Start-NativeExecution {
-                    defaults write $plist CFBundleIdentifier $Name
+                    defaults write $plist CFBundleIdentifier com.microsoft.powershell
                     defaults write $plist CFBundleVersion $Version
                     defaults write $plist CFBundleShortVersionString $Version
                     defaults write $plist CFBundleGetInfoString $Version
--- a/tools/packaging/packaging.psm1
+++ b/tools/packaging/packaging.psm1
@@ -684,6 +684,28 @@ function New-UnixPackage {
             }
         } finally {
             if ($Environment.IsMacOS) {
+                if($pscmdlet.ShouldProcess("Cleanup macOS launcher"))
+                {
+                    # This is needed to prevent installer from picking up
+                    # the launcher app in the build structure and updating
+                    # it which locks out subsequent package builds due to
+                    # increase permissions.
+                    $macosapp = "$PSScriptRoot/macos/launcher/ROOT/Applications/Powershell.app"
+                    $plist = "$macosapp/Contents/Info.plist"
+                    $tempguid = (New-Guid).Guid
+                    Start-NativeExecution {
+                        defaults write $plist CFBundleIdentifier $tempguid
+                        plutil -convert xml1 $plist
+                    }
+
+                    # Restore default permissions.
+                    $shellscript = "$macosapp/Contents/MacOS/PowerShell.sh"
+                    Start-NativeExecution {
+                        chmod 644 $shellscript
+                        chmod 644 $plist
+                    }
+                }
+
                 # this is continuation of a fpm hack for a weird bug
                 if (Test-Path $hack_dest) {
                     Write-Warning "Move $hack_dest to $symlink_dest (fpm utime bug)"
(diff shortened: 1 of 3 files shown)


TravisEz13:
Can you go to https://travis-ci.org/profile/thezim?offset=0 and enable your repo then do another push?  This will trigger a build with packaging.

thezim:
@TravisEz13 Done.

TravisEz13:
Verified packaged correctly here:  https://travis-ci.org/thezim/PowerShell/builds/296027725

tools/packaging/packaging.psm1
```diff
@@ -629,6 +629,49 @@ function New-UnixPackage {
...
+                # Convert to XML plist, needed because defaults native
+                # app auto converts it to binary format when it modify
+                # the plist file.
+                Start-NativeExecution {
+                    plutil -convert xml1 $plist
+                }
+
+                # Set permissions.
+                Start-NativeExecution {
+                    find $macosapp | xargs chmod 755
```
TravisEz13: Can you undo this when it's finished?

TravisEz13 (changes requested):
Can we not leave files with altered permission laying around?

tools/packaging/packaging.psm1
thezim: Sure thing, good catch, commit added. I realized that 755 wasn't needed for everything so I only set what was absolutely needed and then restore to a default state post fpm.

TravisEz13 (approved):

TravisEz13:
packaging build is still running here:  https://travis-ci.org/thezim/PowerShell/jobs/296567451

TravisEz13:
relaunched AppVeyor due to myget failure..

tools/packaging/packaging.psm1
```diff
@@ -665,6 +713,27 @@ function New-UnixPackage {
                 $newPackagePath = Join-Path $createdPackage.DirectoryName $newPackageName
                 $createdPackage = Rename-Item $createdPackage.FullName $newPackagePath -PassThru -Force:$Force
             }
+            if($pscmdlet.ShouldProcess("Cleanup macOS launcher"))
```
daxian-dbw: Should this cleanup code be moved to the `if ($Environment.IsMacOS)` block in the `finally` above?

thezim: I would agree. Since `fpm` is in a `try`, if it fails the next fatal error would be [here](https://github.com/thezim/PowerShell/blob/12516d8f7c6f6012c5b695ef4b2acfe89f6379de/tools/packaging/packaging.psm1#L703) leaving it in a bad state. I'll go ahead and move it.

As an aside I'm not unsure if the `try` should even be there for the `fpm` call. I would think you would want the build point of failure reported at the `fpm` call and not a dependancy later in the code.

TravisEz13:
Packaging build passed here:  https://travis-ci.org/thezim/PowerShell/jobs/297030200

TravisEz13:
@PowerShell/powershell-maintainers  @joeyaiello @SteveL-MSFT  Since this previously regressed after working, I would consider this at least moderately risky.  Do we want to take this for 6.0.0?
I think we should.  I believe we have addressed the issues (thanks @thezim) and we shipped the last beta with this feature.

thezim:
We should ensure if #5323 merges first that we include the UTI update to `CFBundleIdentifier` in the plist in this PR.

TravisEz13:
#5323 was merged.  I updated the title so we don't accidentally merge until the PR is updated.

thezim:
UTI updated. Packaging build [passed](https://travis-ci.org/thezim/PowerShell/builds/297421012).

daxian-dbw:
@TravisEz13 I agree that we should take this fix for 6.0.0, since the beta.9 macOS package already had the macOS launcher change.

daxian-dbw (approved):

adityapatwardhan (approved):
I agree we should take this change.

PR #5292, opened 2017-11-01
Fix dynamic class assembly name

lzybkr:
Using the assembly name to hint at the source of the classes was
problematic in multiple ways.

This change stores the actual filename in an attribute on the assembly.

So for a given type, one can get the assembly this way:

[SomeType].Assembly.GetCustomAttributes() |
    ? { $_ -is [System.Management.Automation.DynamicClassImplementationAssemblyAttribute] } |
    % { $_.ScriptFile }

### DIFF
--- a/src/System.Management.Automation/engine/parser/PSType.cs
+++ b/src/System.Management.Automation/engine/parser/PSType.cs
@@ -19,7 +19,6 @@ namespace System.Management.Automation.Language
         internal const string DynamicClassAssemblyName = "PowerShell Class Assembly";
 
         private static int s_globalCounter = 0;
-        private static readonly object[] s_emptyArgArray = Utils.EmptyArray<object>();
         private static readonly CustomAttributeBuilder s_hiddenCustomAttributeBuilder =
             new CustomAttributeBuilder(typeof(HiddenAttribute).GetConstructor(Type.EmptyTypes), Utils.EmptyArray<object>());
 
@@ -1115,8 +1114,6 @@ namespace System.Management.Automation.Language
             var propertyInfo = new PropertyInfo[] {
                 typeof(DynamicClassImplementationAssemblyAttribute).GetProperty(nameof(DynamicClassImplementationAssemblyAttribute.ScriptFile)) };
             var propertyArgs = new object[] { scriptFile };
-            var fieldInfoList = Utils.EmptyArray<FieldInfo>();
-            var fieldArgs = Utils.EmptyArray<object>();
 
             yield return new CustomAttributeBuilder(ctor, emptyArgs,
                 propertyInfo, propertyArgs, Utils.EmptyArray<FieldInfo>(), emptyArgs);
--- a/src/System.Management.Automation/engine/parser/PSType.cs
+++ b/src/System.Management.Automation/engine/parser/PSType.cs
@@ -16,10 +16,12 @@ namespace System.Management.Automation.Language
 {
     internal class TypeDefiner
     {
+        internal const string DynamicClassAssemblyName = "PowerShell Class Assembly";
+
         private static int s_globalCounter = 0;
         private static readonly object[] s_emptyArgArray = Utils.EmptyArray<object>();
         private static readonly CustomAttributeBuilder s_hiddenCustomAttributeBuilder =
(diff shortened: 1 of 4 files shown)


src/System.Management.Automation/engine/parser/PSType.cs
```diff
@@ -1100,30 +1102,42 @@ internal void DefineEnum()
...
+            if (string.IsNullOrEmpty(scriptFile)) {
+                yield return new CustomAttributeBuilder(ctor, emptyArgs);
+                yield break;
+            }
+
+            var propertyInfo = new PropertyInfo[] {
+                typeof(DynamicClassImplementationAssemblyAttribute).GetProperty(nameof(DynamicClassImplementationAssemblyAttribute.ScriptFile)) };
+            var propertyArgs = new object[] { scriptFile };
+            var fieldInfoList = Utils.EmptyArray<FieldInfo>();
+            var fieldArgs = Utils.EmptyArray<object>();
```
daxian-dbw: `fieldInfoList` and `fieldArgs` are not used anywhere.

src/System.Management.Automation/engine/parser/PSType.cs
```diff
@@ -1100,30 +1102,42 @@ internal void DefineEnum()
             }
         }
 
-        private static IEnumerable<CustomAttributeBuilder> GetAssemblyAttributeBuilders()
+        private static IEnumerable<CustomAttributeBuilder> GetAssemblyAttributeBuilders(string scriptFile)
         {
-            yield return new CustomAttributeBuilder(typeof(DynamicClassImplementationAssemblyAttribute).GetConstructor(Type.EmptyTypes), s_emptyArgArray);
+            var ctor = typeof(DynamicClassImplementationAssemblyAttribute).GetConstructor(Type.EmptyTypes);
+            var emptyArgs = Utils.EmptyArray<object>();
```
daxian-dbw: You call `Utils.EmptyArray<object>()` multiple times, why not just use `s_emptyArgArray`?

test/powershell/Language/Classes/Scripting.Classes.BasicParsing.Tests.ps1
```diff
@@ -688,6 +688,14 @@ Describe 'Type building' -Tags "CI" {
             ++$a::a | Should Be 2
         }
     }
+
+    It 'should get the script from a class type' {
+        class C {}
+
+        $a = [C].Assembly.GetCustomAttributes($false).Where{
+            $_ -is [System.Management.Automation.DynamicClassImplementationAssemblyAttribute]}
+        $a.ScriptFile | Should BeExactly (& { $MyInvocation.ScriptName })
```
daxian-dbw: Maybe replace `(& { $MyInvocation.ScriptName })` with `$PSCommandPath`?

daxian-dbw (changes requested):

src/System.Management.Automation/engine/parser/PSType.cs
lzybkr: I'll remove s_emptyArgArray - it's probably not better to store 2 references to the same array, `Utils.EmptyArray` already has a static reference.

src/System.Management.Automation/engine/parser/PSType.cs
lzybkr: Thanks, cut and paste I guess.

test/powershell/Language/Classes/Scripting.Classes.BasicParsing.Tests.ps1
lzybkr: Uh, sure. My excuse - I've never used it before and was too lazy to look for it. :)

SteveL-MSFT:
This is great, but the usage seems too complex to remember.  Can we use ETS to just add a ScriptFile property making it more discoverable?

lzybkr:
@SteveL-MSFT - I've only seen 1 request for this ability - from the author of ISESteroids - so ease of use doesn't seem too critical. I suppose we could use a `NoteProperty` on the assembly instance, but that is harder to access from C#.

daxian-dbw (approved):

SteveL-MSFT:
@lzybkr we can wait for customer request before adding more.  Thanks

iSazonov (approved):
Leave a comment

Issue #5260, opened 2017-10-27
.NET Core tools version check in build.psm1 blocks the build


anmenaga:
Condition in build.psm1 seems to be looking for exact match with "2.0.0", while the tools are of "2.0.2". This blocks the build.

```none
$dotnetCLIRequiredVersion = "2.0.0"
...
    # Verify if the dotnet in-use is the required version
    $dotnetCLIInstalledVersion = (dotnet --version)
    If ($dotnetCLIInstalledVersion -ne $dotnetCLIRequiredVersion) {
        Write-Warning @"
The currently installed .NET Command Line Tools is not the required version.
...
```
Expected behavior:
---------------
Build succeeds.

Actual behavior
---------------
Build stops at warning:

![versions](https://user-images.githubusercontent.com/11860095/32125069-1b1bee6c-bb1f-11e7-813e-953a89f0702e.png)

Steps to reproduce
------------------

```
git clone --recursive https://github.com/PowerShell/PowerShell.git
Import-Module ./build.psm1
Start-PSBootstrap
Start-PSBuild
```
Environment data
----------------

```powershell
Name                           Value
----                           -----
PSVersion                      5.1.17025.1001
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.17025.1001
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

rkeithhill:
Do you have a `global.json` file in the root of the repo?  It's contents should specify that PowerShell currently uses the 2.0.0 version of the SDK.  The gist is that the presence of this file should cause `dotnet --version` to return `2.0.0` **if** that command is run within the repo dir structure.  From that code snippet, I can't tell which dir the command is run from.  I suspect the dir isn't being set to the repo path which would cause this problem.

anmenaga:
```
PS C:\github\PowerShell> dir global.json


    Directory: C:\github\PowerShell


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----       10/27/2017   1:34 PM             47 global.json


PS C:\github\PowerShell> gc .\global.json
{
  "sdk": {
    "version": "2.0.0"
  }
}
PS C:\github\PowerShell> dotnet --version
2.0.2
PS C:\github\PowerShell>
```

rkeithhill:
[...]

PR #5295, opened 2017-11-01
Update powershell to use 2.0.4-servicing dotnet core runtime

### LINKED ISSUE #5157: Move to latest dotnetcore 2.0.x servicing build after beta.9
We should move to latest validated dotnetcore 2.0.x servicing build to ensure nothing got broken

daxian-dbw:
Use [`Microsoft.NETCore.App 2.0.4-servicing-25831-01`](https://dotnet.myget.org/feed/dotnet-core/package/nuget/Microsoft.NETCore.App/2.0.4-servicing-25831-01) to build powershell.
Also update to dotnet core SDK 2.0.2.

The CoreCLR issue https://github.com/dotnet/coreclr/issues/9745 is fixed in the 2.0.4-servicing packages.
PowerShell will use .NET Core 2.0.4 servicing for the GA build, so we move to the 2.0.4 preview bits to validate dotnet core packages early.

Fix https://github.com/PowerShell/PowerShell/issues/5157


build.psm1
```diff
@@ -3,7 +3,7 @@
 $script:TestModulePathSeparator = [System.IO.Path]::PathSeparator
 
 $dotnetCLIChannel = "release"
-$dotnetCLIRequiredVersion = "2.0.0"
+$dotnetCLIRequiredVersion = "2.0.2"
```
anmenaga: @daxian-dbw While you are at it... maybe it is a good idea to change the required version check from "exact match" to "only major+minor version should match" in build.psm1:414 . See #5260 or more details.

PowerShell.Common.props
```diff
@@ -94,7 +94,7 @@
     <Copyright>(c) Microsoft Corporation. All rights reserved.</Copyright>
 
     <TargetFramework>netcoreapp2.0</TargetFramework>
-    <RuntimeFrameworkVersion>2.0.0</RuntimeFrameworkVersion>
+    <RuntimeFrameworkVersion>2.0.4-servicing-25831-01</RuntimeFrameworkVersion>
```
anmenaga: The package that was tested was "25824-03".
This one, in this change, is just newer?

anmenaga (changes requested):

PowerShell.Common.props
daxian-dbw: Yes, it's a week newer. Produced on 10/31

build.psm1
daxian-dbw: @anmenaga As far as I know, this issue has been addressed by the PR #5118. Please see the original issue #5015.
`global.json` defines which version of SDK we want to use, so it will work even if you have 2.0.2 installed.

anmenaga (approved):

PowerShell.Common.props
iSazonov: I see 2.0.4 here and 2.0.2 in Build.psm1 and global.json - could you please clarify the difference?

daxian-dbw: The 2.0.2 in global.json indicates the version of the .NET Core SDK, see [global.json reference](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json). The `2.0.4-servicing` here is the version of dotnet core runtime the build will target.

`dotnet --info` shows 2 versions: .NET Command Line Tools version and the default .NET Core framework version. Currently the 2.0.2 .NET Core SDK still target .NET Core runtime 2.0.0 by default.
```
.NET Command Line Tools (2.0.2)

Product Information:
 Version:            2.0.2
 Commit SHA-1 hash:  a04b4bf512
...
Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0
  Build    : e8b8861ac7faf042c87a5c2f9f2d04c98b69f28d
```

adityapatwardhan (approved):
LGTM

PR #5301, opened 2017-11-02
Rework passing array literal to native commands

lzybkr:
Now that the Ast is passed through to the NativeCommandParameterBinder, there is a cleaner fix to ensuring we pass the correct text to a command that uses PowerShell array syntax.

### DIFF
--- a/src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
+++ b/src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
@@ -366,19 +366,14 @@ namespace System.Management.Automation
             var afterPrev = prev.Extent.EndOffset;
             var beforeNext = next.Extent.StartOffset - 1;
 
-            if (afterPrev == beforeNext)
-                return ",";
+            if (afterPrev == beforeNext) return ",";
 
             var arrayText = arrayExtent.Text;
             afterPrev -= arrayExtent.StartOffset;
             beforeNext -= arrayExtent.StartOffset;
 
-            if (arrayText[afterPrev] == ',')
-                return ", ";
-
-            if (arrayText[beforeNext] == ',')
-                return " ,";
-
+            if (arrayText[afterPrev] == ',') return ", ";
+            if (arrayText[beforeNext] == ',') return " ,";
             return " , ";
         }
 
--- a/src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
+++ b/src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
@@ -105,28 +105,27 @@ namespace System.Management.Automation
 
                     if (argValue != AutomationNull.Value && argValue != UnboundParameter.Value)
                     {
-                        // ArrayIsSingleArgumentForNativeCommand is true when a comma is used in the
-                        // command line, e.g.
+                        // ArrayLiteralAst is used to reconstruct the correct argument, e.g.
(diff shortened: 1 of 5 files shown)


test/powershell/Language/Scripting/NativeExecution/NativeCommandArguments.Tests.ps1
```diff
@@ -43,4 +43,21 @@ Describe "Native Command Arguments" -tags "CI" {
             $lines[$i] | Should Be "Arg $i is <$($expected[$i])>"
         }
     }
+
+    It "Should handle PowerShell arrays with or without spaces correctly: <arguments>" -TestCases @(
+        @{arguments = "a 1,2"; expected = "a", "1,2"}
```
vors: let's add something boring, like `a,1,2` here as well

lzybkr: There is no difference between 2 and 3 elements - it's just spaces around the argument(s) that matter.

src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
```diff
@@ -158,14 +157,19 @@ internal String Arguments
...
         {
             IEnumerator list = LanguagePrimitives.GetEnumerator(obj);
-            bool needSeparator = false;
 
+            Diagnostics.Assert(argArrayAst == null
+                || obj is object[] && ((object[])obj).Length == argArrayAst.Elements.Count,
+                "array argument and ArrayLiteralAst differ in number of elements");
+
+            int currentElement = -1;
+            string separator = "";
```
iSazonov: Should we use `String.Empty`?

src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
```diff
@@ -349,6 +352,35 @@ internal static bool NeedQuotes(string stringToCheck)
...
+
+            var next = arrayLiteralAst.Elements[index];
+            var prev = arrayLiteralAst.Elements[index - 1];
+
+            var arrayExtent = arrayLiteralAst.Extent;
+            var afterPrev = prev.Extent.EndOffset;
+            var beforeNext = next.Extent.StartOffset - 1;
+
+            if (afterPrev == beforeNext)
+                return ",";
```
iSazonov: Could you please use one line formatting as above in line 357?
Below too in 376 and 379.

test/powershell/Language/Scripting/NativeExecution/NativeCommandArguments.Tests.ps1
iSazonov: Should we add `1,2 3`?

iSazonov (commented):
Leave a comment

test/powershell/Language/Scripting/NativeExecution/NativeCommandArguments.Tests.ps1
lzybkr: I really didn't need `a 1,2`, `1,2` would have been sufficient, but I had a bug in the test with array vs. scalar and the first thought I had was to add `a ` to force array. I'm fixing by using `@()` instead to force array.

src/System.Management.Automation/engine/NativeCommandParameterBinder.cs
lzybkr: No reason to imo.

daxian-dbw (approved):
LGTM.

PR #5302, opened 2017-11-02
Speed up check for suspicious content

lzybkr:
This replaces the code that scanned a script block for suspicious strings.
The previous implementation:
* Tokenized input (generating many strings for garbage collection)
* Use multiple threads

This approach is based on Rubin-Karp and does not allocate any memory
other than a small array to hold the running hash values.

I tested the new and old approaches on 2200 files in the PowerShell repo.
The old code ran in about 1.8-2.1s (ignoring time spent reading files)
The new code runs in about 0.6s and is more stable due to no garbage.

### DIFF
--- a/src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
+++ b/src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
@@ -13,6 +13,7 @@ using System.Management.Automation.Runspaces;
 using System.Management.Automation.Tracing;
 using System.Security.Cryptography.X509Certificates;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Runtime.Serialization;
 using System.Text;
 using System.Threading.Tasks;
@@ -1492,44 +1493,10 @@ namespace System.Management.Automation
 
         // Quick check for script blocks that may have suspicious content. If this
         // is true, we log them to the event log despite event log settings.
-        //
-        // Performance notes:
-        //
-        // For the current number of search terms, the this approach is about as high
-        // performance as we can get. It adds about 1ms to the invocation of a script
-        // block (we don't do this at parse time).
-        // The manual tokenization is much faster than either Regex.Split
-        // or a series of String.Split calls. Lookups in the HashSet are much faster
-        // than a ton of calls to String.IndexOf (which .NET implements in native code).
-        //
-        // If we were to expand this set of keywords much farther, it would make sense
-        // to look into implementing the Aho-Corasick algorithm (the one used by antimalware
-        // engines), but Aho-Corasick is slower than the current approach for relatively
-        // small match sets.
         internal static string CheckSuspiciousContent(Ast scriptBlockAst)
         {
-            // Split the script block text into an array of elements that have
-            // a-Z A-Z dash.
-            string scriptBlockText = scriptBlockAst.Extent.Text;
-            IEnumerable<string> elements = TokenizeWordElements(scriptBlockText);
-
-            // First check for plain-text signatures
-            ParallelOptions parallelOptions = new ParallelOptions();
-            string foundSignature = null;
-
-            Parallel.ForEach(elements, parallelOptions, (element, loopState) =>
-            {
-                if (foundSignature == null)
-                {
-                    if (s_signatures.Contains(element))
-                    {
-                        foundSignature = element;
-                        loopState.Break();
-                    }
-                }
-            });
-
-            if (!String.IsNullOrEmpty(foundSignature))
+            var foundSignature = SuspiciousContentChecker.Match(scriptBlockAst.Extent.Text);
+            if (foundSignature != null)
             {
                 return foundSignature;
             }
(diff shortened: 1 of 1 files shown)


src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
```diff
@@ -1556,106 +1523,285 @@ internal static string CheckSuspiciousContent(Ast scriptBlockAst)
...
-            // Suspicious methods / properties on "Type"
-            "GetConstructor", "GetConstructors", "GetDefaultMembers", "GetEvent", "GetEvents", "GetField",
-            "GetFields", "GetInterface", "GetInterfaceMap", "GetInterfaces", "GetMember", "GetMembers",
-            "GetMethod", "GetMethods", "GetNestedType", "GetNestedTypes", "GetProperties", "GetProperty",
-            "InvokeMember", "MakeArrayType", "MakeByRefType", "MakeGenericType", "MakePointerType",
-            "DeclaringMethod", "DeclaringType", "ReflectedType", "TypeHandle", "TypeInitializer",
-            "UnderlyingSystemType",
+                    // If the character isn't in any of our patterns,
+                    // don't bother hashing and reset the running length.
+                    if (!((h >= 'a' && h <= 'z') || h == '-'))
```
iSazonov: We could use here "else if".

lzybkr: Nice - I manually inlined the first `if` and didn't notice.

src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
```diff
@@ -1556,106 +1523,284 @@ internal static string CheckSuspiciousContent(Ast scriptBlockAst)
...
+            /// Instead, we can rely on knowing we need the hashes for
+            /// shorter strings anyway, so we reuse their values in
+            /// computing the hash for the longer patterns. This lets us
+            /// use a much simpler hash as well - we can avoid the use of
+            /// mod.
+            /// </summary>
+            /// <returns>The string matching the hash, or null.</returns>
+            public static string Match(string text)
+            {
+                // The longest pattern is 29 characters.
```
PaulHigin: I assume the 29 character limit was chosen to accommodate the current set of suspicious patterns.  But what if a new pattern is added that is greater than 29 characters?  It seems the generator method (HashNewPattern) should check length and throw an error if greater than current longest pattern.

src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
```diff
@@ -1556,106 +1523,284 @@ internal static string CheckSuspiciousContent(Ast scriptBlockAst)
...
+                // The values in the array are the computed hashes of length
+                // index-1 (so runningHash[0] holds the hash for length 1).
+                var runningHash = new uint[29];
+
+                int longestPossiblePattern = 0;
+                for (int i = 0; i < text.Length; i++)
+                {
+                    uint h = text[i];
```
PaulHigin: Please change 'h' variable to 'c' to be consistent with HasNewPattern which uses 'c' as character variable.

src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
```diff
@@ -1556,106 +1523,284 @@ internal static string CheckSuspiciousContent(Ast scriptBlockAst)
...
+                        // If the character isn't in any of our patterns,
+                        // don't bother hashing and reset the running length.
+                        longestPossiblePattern = 0;
+                        continue;
+                    }
 
-            // Doing things with System.Runtime.InteropServices
-            "InteropServices", "Marshal", "AllocHGlobal", "PtrToStructure", "StructureToPtr",
-            "FreeHGlobal", "IntPtr",
+                    for (int j = Math.Min(i, runningHash.Length) - 1; j > 0; j--)
```
PaulHigin: I must be missing something because I don't understand this loop.  It looks like computing the next hash (based on HashNewPattern below) simply takes the previous hash, multiples by LCG, and adds next character.  Why are previous hashes re-computed (j-1, j=(i-1) -> 1)?

PaulHigin (changes requested):

src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
lzybkr: Consistency is over rated. Here `h` is not a character, but the hash value.

src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
lzybkr: I'm not sure how to improve the comment. I'll try again I guess.

Briefly - omitting many other steps:

Iteration n: compute hash on `Emi` (len 3)
Iteration n+1: we compute hash on `Emit` (len 4) using hash from previous iteration (`j-1`)
Iteration n+1: compute hash on `mit` (len 3) (overwriting previous iteration, so that's why we go from longest to shortest.)

src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
lzybkr: Sure, I'll add a check.

src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
PaulHigin: uint h = text[i];    'h' is not a character from the text to be tested?  I though the hashes were stored in 'runningHash'?

lzybkr: Yes, it's both, but I don't need 2 local variables.

lzybkr: And hashes are stored in `runningHash`, but `h` is being used to compute the hash. One could say it is partially hashed. In previous iterations of the code, I had intermediate steps, so there were multiple assignments to `h` as part of hashing.

src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
PaulHigin: Ok, I understand the progression now and the loop makes sense.  This makes the algorithm much more clear (at least to me).  Can you add this to your comment?

PaulHigin (approved):
Just some minor further comments.  Otherwise LGTM.
With your tests it looks like false positives are very rare.

PR #5306, opened 2017-11-02
Fix breaking change in Add-Type

iSazonov:
Remove a breaking change we made in #5243
Because AddtypeCommandBase.EndProcessing() is a _public_ method we revert changes in it.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
@@ -760,17 +760,10 @@ namespace Microsoft.PowerShell.Commands
         /// <summary>
         /// Perform common error checks.
         /// Populate source code.
+        /// We only keep the code for backward compatibility.
         /// </summary>
         protected override void EndProcessing()
         {
-            // Prevent code compilation in ConstrainedLanguage mode
-            if (SessionState.LanguageMode == PSLanguageMode.ConstrainedLanguage)
-            {
-                ThrowTerminatingError(
-                    new ErrorRecord(
-                        new PSNotSupportedException(AddTypeStrings.CannotDefineNewType), "CannotDefineNewType", ErrorCategory.PermissionDenied, null));
-            }
-
             // Generate an error if they've specified an output
             // assembly type without an output assembly
             if (String.IsNullOrEmpty(outputAssembly) && outputTypeSpecified)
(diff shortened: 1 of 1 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
```diff
@@ -787,6 +780,37 @@ protected override void EndProcessing()
...
+                    new ErrorRecord(
+                        new PSNotSupportedException(AddTypeStrings.CannotDefineNewType), "CannotDefineNewType", ErrorCategory.PermissionDenied, null));
+            }
+
+            // Load the source if they want to load from a file
+            if (String.Equals(ParameterSetName, "FromPath", StringComparison.OrdinalIgnoreCase) ||
+                String.Equals(ParameterSetName, "FromLiteralPath", StringComparison.OrdinalIgnoreCase)
+                )
+            {
+                sourceCode = "";
```
PaulHigin: Don't we want to use the updated StringBuilderCode here to get the perf benefit?  Also there is duplicate code that could be handled with base class helper methods.  But I believe you are looking to refactor this code so it may not be worth changing now just to be refactored later.

PaulHigin (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
iSazonov: I believe the method is a candidate to remove. I've tried add [Obsolete] but it don't work. I think we could document the method and perhaps  the class as deprecated - I'll wonder if anybody use it. Another argument for removing this code is that the code is the remainder of the CodeDom.
So I split the code (no helper methods) to easily remove the deprecated code later.

PaulHigin: Ok.

PaulHigin (approved):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
daxian-dbw: I think the breaking changes to `AddTypeCommandBase` is most likely safe. I searched on Github and don't see the type `AddTypeCommandBase` gets used except AddType.
Given that, would it be OK that we just keep the breaking change made in #5243?

iSazonov: I don't see the need for `AddTypeCommandBase`. If the team approves, I'll be happy to remove that.

daxian-dbw: I'm merging this PR. We can loop in the powershell committee in your subsequent add-type refactoring PR (or the issue for that work).

iSazonov: The request is in #5158.

PR #5309, opened 2017-11-02
 Update 'installpsh-suse.sh' and remove 'download.sh'

### LINKED ISSUE #5023: Fix or remove scripts that install powershell on openSUSE

daxian-dbw:
Fix #5023
Update `installpsh-suse.sh` to work with the `.tar.gz` binary archive.
Remove `download.sh` as it's not used anywhere now. (checked with PowerShellGet and OneGet, they are using a local copy of `download.sh`).


tools/installpsh-suse.sh
```diff
@@ -154,10 +161,27 @@ if [[ ! -r "$package" ]]; then
...
+$SUDO ln -s /opt/microsoft/powershell/$release/pwsh $pwshlink
+
+## Add the symbolic link path to /etc/shells
+if [ ! -f /etc/shells ] ; then
+    echo $pwshlink | $SUDO tee /etc/shells ;
+else
+    grep -q "^${pwshlink}$" /etc/shells || echo $pwshlink | $SUDO tee --append /etc/shells > /dev/null ;
+fi
+
+## Remove the package
```
anmenaga: Maybe better to change this comment to "Remove the downloaded package file"

tools/installpsh-suse.sh
```diff
@@ -154,10 +161,27 @@ if [[ ! -r "$package" ]]; then
...
+rm -f $package
 
-powershell -noprofile -c '"Congratulations! PowerShell is installed at $PSHOME"'
+pwsh -noprofile -c '"Congratulations! PowerShell is installed at $PSHOME"'
```
anmenaga: Maybe a good idea to indicate the new short 'pwsh' name in the message; for example "$PSHOME/pwsh" instead of "$PSHOME" - this way complete command line for launch can be copy pasted.

anmenaga (changes requested):

tools/installpsh-suse.sh
daxian-dbw: Sure, will fix it.

tools/installpsh-suse.sh
daxian-dbw: Will do.

daxian-dbw: `pwsh` will be in PATH, so you don't need the complete command. But we should indicate that run `pwsh` to start a powershell session.

anmenaga (approved):

TravisEz13 (approved):

DarwinJS:
There could be a lot of user automation depending on download.sh - so I think it would be better to redirect download.sh to install-powershell.sh.  Or at a minimum have it emit a  depreciation message.

TravisEz13:
I think a redirect to `install-powershell.sh` would be acceptable.

daxian-dbw:
@DarwinJS Thanks for bringing this up. I did a quick search of `"master/tools/download.sh"` on Github (here is the [result](https://github.com/search?p=1&q=%22master%2Ftools%2Fdownload.sh%22&type=Code&utf8=%E2%9C%93)). Some of the occurrences are from forks of powershell repository. For the rest, I have opened issues in each of those repositories to ask them to replace `download.sh` with `install-powershell.sh`.

DarwinJS:
@daxian-dbw - on my last comment I forgot to say thanks a lot - I struggled trying to fix this with regular package manager commands and had to move on to other things.
(Sidenote: check out http://install-powershell.sh and the forwarded domain http://pwsh.io)

My intent in building install-powershell.sh was to allow the super easy installation of powershell core to add jet fuel to adoption.

This includes:

1. Installing it on one's workstation to get a feel for developing with it (hence the -includeide switch) and
2. Installing it on regular servers (hence default behavior mimics download.sh)
3. Using the install scripts offline (hence install-powershell.sh checks locally for sub-scripts).

FYI: I didn't even learn it was used to install for CI testing until much later.

I was also thinking that there was probably many who took a direct dependency on download.sh to get rid of all the hassles of figuring out their platform specific install procedure and then keeping up to date.

Although I don't think this project should feel a contractual obligation to not break that, I also feel that managing it with the sensitivity that people probably do take a dependency on it helps fuel adoption.

So rather than retire it because it seems redundant, I would advise changing its contents to:

```
bash <(curl -s https://raw.githubusercontent.com/PowerShell/PowerShell/master/tools/install-powershell.sh)
```

daxian-dbw:
@DarwinJS I chatted with @TravisEz13 offline and we agree to bring back `download.sh` and make it redirect to `install-powershell.sh`.

DarwinJS:
Awesome :)

daxian-dbw:
@DarwinJS Please see the PR #5386. Thanks!

PR #5144, opened 2017-10-17
Redirect ETW logging to Syslog on Linux

dantraMSFT:
NOTE: Tests are pending...

This PR is divided into the following areas: 
1: Add/Rename the PowerShell/Windows ETW manifest to the repo and change both the provider id (guid) and name.  See https://github.com/PowerShell/PowerShell/issues/4939.
The manifest is at tools/resxgen/PowerShell-Core-Instrumentation.man

2: Generate a resx file containing the string resources needed for logging ETW events to syslog.
This is accomplished by tools\resxgen\resxgen.psm1 and resxgen.ps1.  The tool generates two files
* A resx file containing string resources for each message string from the manifest.  This is generated into the System.Management.Automation\gen directory.
* A C# class (EventResource) that provides the mapping between the integer event id and the associated string resource name. The file is generated into the System.Management.Automation\CoreCLR directory with a compile-time condition of UNIX
NOTE: The EventResource.cs class generated by resgen is explicitly ignored in the csproj file; it is not used.

3: Native shims for the syslog apis to enable pInvoke.
The shim implementations are libpsl-native\src\nativesyslog.cpp.  Native_SysLog, Native_OpenLog, and Native_CloseLog are the target pInvoke.
NOTE: See https://github.com/PowerShell/PowerShell/pull/5149.

4: SMA\utils\tracing\PSSysLogProvider.cs 
Implements the abstract LogProvider class and is the syslog equivalent of PSEtwLogProvider.  The class contains a number of logical methods for logging lifecycle, health, and normal events.  

5: SMA\utils\tracing\SysLogProvider.cs 
This is the Syslog equivalent of ETW's log provider class and implements a Log method versus ETW's EventWrite.  It is also responsible for resolving event ids to resource names and peforming the Syslog call.  There is a large comment block in the class XML doc that describes the types of log output it produces.

6: SMA\utils\tracing\PSEtwLog.cs 
PowerShell's current implementation is tightly coupled to this class; with code calling it directly for all events. To simplify integration of syslog, I updated the class to create an instance of PSSysLogProvider on Linux and removed the Linux-specific stub file.

7: SMA\engine\PropertyAccessor.cs
This class provides a wrapper around PowerShellProperties.json and has been extended to have Unix-specific assessors for configuring logging.  Note that the file is expected to be in the $PSHOME directory to ensure SxS.

Currently, there are four configuration properties:

[...]

Issue #3766, opened 2017-05-11
PowerShell logging needs to go to syslog on non-Windows


SteveL-MSFT:
PowerShell currently has ETW logging on Windows, but no logging on non-Windows

iSazonov:
Is there no plans to port ETW on Unix?

SteveL-MSFT:
Perhaps consider using [OpenTracing ](https://github.com/opentracing/opentracing-csharp)

lzybkr:
I believe .Net settled on [LTTng](http://lttng.org/) for logging, so it's worth a look too.

daxian-dbw:
Closed via #5144 
Tests of the syslog work need to be added, which is tracked by #5312

PR #5310, opened 2017-11-02
Use os_log APIs to log PowerShell events on OSX

dantraMSFT:
This change replaces calls to syslog with os_log on OSX.  syslog calls on 10.12 was not producing log output and the functions do not report errors.

See https://github.com/PowerShell/PowerShell/issues/3766.

NOTE: PR https://github.com/PowerShell/PowerShell/pull/5144 depends on this PR.

### DIFF
--- a/src/libpsl-native/src/nativesyslog.cpp
+++ b/src/libpsl-native/src/nativesyslog.cpp
@@ -2,18 +2,65 @@
 //! @brief Provides wrappers around the syslog apis to support exporting
 //! for PInvoke calls by powershell.
 //! These functions are intended only for PowerShell internal use.
+//! To view log output in real time
+//! On Linux
+//!    tail -f /var/log/syslog | grep powershell 
+//!  On OSX
+//!    sudo log stream
+//!  NOTE: replace powershell with the LogIdentity value when overriding in configuration
 #include <syslog.h>
 #include <nativesyslog.h>
 
+#if defined(__APPLE__) && defined(__MACH__)
+
+#include <os/log.h>
+#include <stdio.h>
+static os_log_t _log = NULL;
+
+// This format string ensures the message string for the log statements
+// are visible. Using just %s marks them as private and the do not show
+// up with log stream.
+#define MESSAGE_FORMAT "%{public}s"
+
+// The submodule name to pass to os_log_create.
+// The passed in ident value will be used for the category.
+// see man os_log_create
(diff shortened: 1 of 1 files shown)


src/libpsl-native/src/nativesyslog.cpp
```diff
@@ -2,18 +2,65 @@
...
+        case LOG_CRIT:
+            os_log_fault(_log, MESSAGE_FORMAT, message);
+            break;
+        
+        case LOG_ERR:
+            os_log_error(_log, MESSAGE_FORMAT, message);
+            break;
+
+        case LOG_DEBUG:
+            os_log_debug(_log, MESSAGE_FORMAT, message);
```
daxian-dbw: For those different log types, do we need to check whether the type is enabled before calling the log method? What would happen in case the target type is not enabled?

dantraMSFT: The check occurs in SysLogProvider.cs.  The filters are passed into its constructor by PSSysLogProvider.  Both Linux and Mac use the PowerShellProperties.json to define the filters (see JsonConfigFileAccessor in PropertyAccessor.cs)

daxian-dbw: I don't see the API [`os_log_type_enabled`](https://developer.apple.com/documentation/os/1643749-os_log_type_enabled?language=objc) exposed in this .cpp file, how does `SysLogProvider.cs` check if a log type is enabled? Does it enable all log types when creating the log object when calling `os_log_create`?

dantraMSFT: It's not needed.  Operational events are logged at the default level; which is enabled by default so we'll see the same output we would see in the event log.

https://developer.apple.com/documentation/os/os_log

I'd prefer to filter without interop calls at the risk that someone will change the configuration for powershell at the native layer and we make unneeded interop calls.  We can address this later, if it becomes an issue.

dantraMSFT:
@adityapatwardhan Would you mind taking a look at this?  My syslog PR is blocked until this gets into main and the nuget  package is published.

Thanks.

src/libpsl-native/src/nativesyslog.cpp
daxian-dbw: Thanks for the clarification. #Closed

daxian-dbw (approved):

PR #5321, opened 2017-11-03
Revert refactoring changes that broke remoting to Windows PowerShell 5.1

### LINKED ISSUE #4615: Remoting host callbacks incompatible between PowerShell 5.1 and 6.0
This is a remoting incompatibility introduced in PowerShell 6.0.  Remoting supports host callbacks to the client and the callback information includes type information and values for passed parameter arguments.  For example a PromptForChoice callback passes a ChoiceDescription type argument, which includes fields "label" and "message".  But PowerShell 6.0 changed the field names to "_label" and "_message".  When decoding the host callback message these fields names do not exist on the client and the argument decode fails with a null reference exception:(https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/remoting/common/WireDataFormat/RemoteHostEncoder.cs#L95).

We need to go through all supported host callbacks and ensure all types have not modified field names.  There are only a few callbacks supported so this shouldn't be too bad.

Steps to reproduce
------------------

```powershell
PS [PowerShellCore 6.0] > Enter-PSSession -cn localhost -config microsoft.powershell
[localhost (PowerShell 5.1)]: C:\> Install-Module Razzle
```

Actual behavior
---------------

```none
Null reference exception in remote decoder when handling PromptForChoice callback.
```

SteveL-MSFT:
Call backs depend on private members to have the same name across versions of PowerShell.  Specifically fixes prompts and clear.

@PaulHigin are there other callbacks I need to fix?  I noticed that ctrl+c doesn't work from PSCore6 to Windows PowerShell 5.1.

Fix https://github.com/PowerShell/PowerShell/issues/4615


lzybkr (approved):
Sigh - it's silly that the protocol requires the names of the private fields. I suppose it's just an expedient implementation detail - just use reflection, but it's so brittle.

SteveL-MSFT:
@lzybkr I agree now that I've learned how this aspect of the protocol works

src/System.Management.Automation/engine/hostifaces/ChoiceDescription.cs
```diff
@@ -19,6 +19,10 @@ namespace System.Management.Automation.Host
     public sealed
     class ChoiceDescription
     {
+        // DO NOT REMOVE OR RENAME THESE FIELDS - it will break remoting
```
PaulHigin: I think the warning messages should say "it will break remoting compatibility with Windows PowerShell".

PaulHigin (changes requested):

src/System.Management.Automation/engine/hostifaces/ChoiceDescription.cs
PaulHigin: This is incomplete.  To fully fix this incompatibility we need to revert field names for all encoded class/structs.  Here is the supported list:
https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/remoting/common/WireDataFormat/RemoteHostEncoder.cs#L41

SteveL-MSFT: I reused an existing comment.  I can update all occurrences of that comment to be more specific.  Thanks for the pointer to the list of types, I'll fix those as well.

TravisEz13:
@SteveL-MSFT One of the commits is not linked to your github account.

SteveL-MSFT:
@TravisEz13 fixed

src/System.Management.Automation/engine/hostifaces/ChoiceDescription.cs
```diff
@@ -19,6 +19,10 @@ namespace System.Management.Automation.Host
     public sealed
     class ChoiceDescription
     {
+        // DO NOT REMOVE OR RENAME THESE FIELDS - it will break remoting compatibility with Windows PowerShell compatibility with Windows PowerShell
```
mirichmo: What about putting these fields within a region to make the grouping more explicit?

SteveL-MSFT: Good idea.  Will make an update.

PaulHigin (approved):

SteveL-MSFT:
@TravisEz13 can you merge?

PR #5324, opened 2017-11-03
Remove $IsCoreClr

### LINKED ISSUE #4904: Missing documentation - What's the purpose of $IsCoreCLR?
This is not an issue but I can't find any information about what's the purpose and usage for this variable:
$IsCoreCLR.

This is not event included in the PowerShell Help documentation.
It's excluded from the "About_Automatic_Variables" help doc.

So, all other $Is* are self-explanatory!

Steps to reproduce
------------------

```powershell
Get-Variable Is*

**IsCoreCLR                      True**
IsLinux                        False
IsMacOS                        False
IsWindows                      True

Get-Help About_Automatic_Variables

```
Help document has no information about any of the $Is* variables.

$Psver
Environment data
----------------

```powershell
PS C:\Program Files\PowerShell\6.0.0-beta.7> $PSVersionTable

OS                             Microsoft Windows 10.0.16294

```

iSazonov:
Fix #4904.

### Fix description
1. `$isCoreCLR` is removed from C# code.
2. `$isCoreCLR = $true` is added in tests because these tests are very different and each one requires a separate conclusion.

### DIFF
--- a/src/System.Management.Automation/CoreCLR/CorePsPlatform.cs
+++ b/src/System.Management.Automation/CoreCLR/CorePsPlatform.cs
@@ -52,17 +52,6 @@ namespace System.Management.Automation
         }
 
         /// <summary>
-        /// True if PowerShell was built targeting .NET Core.
-        /// </summary>
-        public static bool IsCoreCLR
-        {
-            get
-            {
-                return true;
-            }
-        }
-
-        /// <summary>
         /// True if the underlying system is NanoServer.
         /// </summary>
         public static bool IsNanoServer
--- a/src/System.Management.Automation/engine/InitialSessionState.cs
+++ b/src/System.Management.Automation/engine/InitialSessionState.cs
@@ -4534,12 +4534,6 @@ end
                 Platform.IsWindows,
                 String.Empty,
                 ScopedItemOptions.ReadOnly | ScopedItemOptions.AllScope),
-
-            new SessionStateVariableEntry(
-                SpecialVariables.IsCoreCLR,
-                Platform.IsCoreCLR,
-                String.Empty,
-                ScopedItemOptions.ReadOnly | ScopedItemOptions.AllScope),
             #endregion
         };
 
--- a/src/System.Management.Automation/engine/SpecialVariables.cs
+++ b/src/System.Management.Automation/engine/SpecialVariables.cs
@@ -166,9 +166,6 @@ namespace System.Management.Automation
         internal const string IsWindows = "IsWindows";
         internal static VariablePath IsWindowsPath = new VariablePath("IsWindows");
 
-        internal const string IsCoreCLR = "IsCoreCLR";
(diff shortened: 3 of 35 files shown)


iSazonov:
PowerShellGet depends heavily on $IsCoreCLR - its tests failed.

Should we close the PR until PowerShellGet uses $IsCoreCLR?

daxian-dbw:
I opened the issue https://github.com/PowerShell/PowerShellGet/issues/187 on PowerShellGet repo and also submitted the PR https://github.com/PowerShell/PowerShellGet/pull/188 to update the IsCoreCLR check in their scripts. After the PR gets merged, we still need them to publish a new version of PowerShellGet module that we can use in PowerShell Core.

I also found that they are still running test on powershell-beta.5. I opened https://github.com/PowerShell/PowerShellGet/issues/189 to ask them to update CIs to run on beta.9 powershell core.

SteveL-MSFT (changes requested):
Rather than pasting this everywhere.  `$IsCoreCLR` should be replaced with `$PSVersionTable.PSEdition -eq 'Core'`.  For tests that skip on !$IsCoreCLR since we have made the decision to not have a .Net Framework build of PSCore6, we should probably just remove those tests if they really are not applicable to PSCore6.

iSazonov:
@SteveL-MSFT I wonder ` $PSVersionTable.PSEdition -eq 'Core'` is just as dummy as the `$IsCoreCLR = $true` - we don't need to check anything, we only need skip them and need a conclusion on every test.
I didn't remove these tests because I was worried that the situation was like removing `#if CoreCLR` from C# codes - maybe some features we porting later.

SteveL-MSFT:
@iSazonov if the concern is that these tests might be valuable in the future as we port more stuff (or corefx gets richer and we can just enable), then it seems that we should rename usage of $IsCoreClr to something like $NotCurrentlyPorted = $true.

daxian-dbw:
FYI, https://github.com/PowerShell/PowerShellGet/pull/188 has been merged. We still need them to publish  a new module (question asked in https://github.com/PowerShell/PowerShellGet/issues/187).

iSazonov:
The new module will published in weeks - can we wait?

SteveL-MSFT:
At this time, I'm thinking we should just leave `$IsCoreClr` as-is and focus our energy on something else.  I wouldn't block RC for this and wouldn't want to take this breaking change after RC either.

SteveL-MSFT:
At this point, I don't think it's worth additional effort on this.  `$IsCoreClr` is staying.

iSazonov:
@SteveL-MSFT Is this delayed or canceled at all? Should I keep iSazonov:remove-iscoreclr branch?

SteveL-MSFT:
@iSazonov I appreciate the work you've done already, but it seems that your time may be spent on other issues.  I would consider this closed and you can delete your branch.

PR #5327, opened 2017-11-03
Replace pwsh with powershell in VS Code tasks.json for Windows

iSazonov:
### Problem
It is impossible to build and debug from VS Code on Windows - VS Code cannot find pwsh.exe.
### Fix
Use powershell.exe on Windows to build.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -5,6 +5,9 @@
     "showOutput": "always",
     "suppressTaskName": true,
     "args": [ "-command" ],
+    "windows": {
+        "command": "powershell"
+    },
 
     "tasks": [
         {


SteveL-MSFT:
`pwsh` should be in the PATH with beta.9

iSazonov:
What if I don't install Beta.9?

SteveL-MSFT:
But isn't this file only used with PSCore6?  Eventually we will only support the final release which will be called pwsh.

iSazonov:
By default Windows don't have pwsh.exe installed. Currently we can not build PowerShell Core 6.0 on Windows from VS Code until we install PowerShell Core 6.0 >=Beta.9 but can from command line. On Unix we have to install PowerShell Core 6.0 in any case because we deprecate install.sh. 
So question is - Shall we continue support build PowerShell Core with Windows PowerShell or no? If we continue build from command line why we don't support to build from VS Code?

iSazonov:
We should document if PowerShell Core 6.0 is mandatory prerequest on Windows now or better add in BootStrap (check latest version?).

SteveL-MSFT:
I like the suggestion to have it in `Start-PSBootStrap`.  Created issue https://github.com/PowerShell/PowerShell/issues/5367

I'll submit a doc update.

iSazonov:
Ok, I close the PR because we'll have `pwsh.exe` with `Start-PSBootStrap`.

PR #5338, opened 2017-11-04
Fix Single Value JSON null in Invoke-RestMethod

### LINKED ISSUE #5320: Invoke-RestMethod doesn't return useful info when no data is returned.
Steps to reproduce
------------------
You will need to [generate](https://github.com/settings/tokens) an API token and grant the notifications privilege.

```powershell
PS /Users/greg> $apitoken = "yourapitoken"
PS /Users/greg> $headers = @{ Authorization = "token $apitoken" }
PS /Users/greg> $resp = Invoke-RestMethod -Uri https://api.github.com/notifications -Headers $headers
PS /Users/greg> $resp -eq $null                                                                                                     
PS /Users/greg>
```

What is returned is an `Object`.
```powershell
PS /Users/greg> $resp -is [System.Object]                                                                                           
True
PS /Users/greg> 
```

Should be no different than this.

```powershell
PS /Users/greg> $obj = New-Object System.Object                                                                                     
PS /Users/greg> $obj -eq $null                                                                                                      
False
PS /Users/greg> 
```

Expected behavior
-----------------
```powershell
PS /Users/greg> $resp -eq $null
True
PS /Users/greg>
```

Actual behavior
---------------
```powershell
PS /Users/greg> $resp -eq $null
PS /Users/greg>
```

Environment data
----------------
```powershell
OS                             Darwin 17.2.0 Darwin Kernel Version 17.2.0: Fri Sep 29 18:27:05 PDT 2017; root:xnu-4570.20.62~3/R...
```

This occurs for Windows PowerShell versions too.

markekraus:
closes #5320 

When an API returns just `null`, `Invoke-RestMethod` was serializing this as the string `"null"` instead of `$null`. Singe value literals are valid JSON according to [rfc7159](https://tools.ietf.org/html/rfc7159) section 2:

```none
   A JSON text is a serialized value.  Note that certain previous
   specifications of JSON constrained a JSON text to be an object or an
   array.  Implementations that generate only objects or arrays where a
   JSON text is called for will be interoperable in the sense that all
   implementations will accept these as conforming JSON texts.
```

Invoke-RestMethod is already properly managing `false`, `true`, numbers, string, arrays, and objects.

This PR fixes the logic in `Invoke-RestMethod` to properly serialize a valid single value JSON `null` literal as `$null`.

This is a breaking change in the sense that users who are using `Invoke-RestMethod` against endpoints that returned single value `null` likely have logic in place to work around the current broken behavior.This change will break that logic.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
@@ -5,6 +5,7 @@ Copyright (c) Microsoft Corporation. All rights reserved.
 using System;
 using System.Management.Automation;
 using System.IO;
+using System.Text.RegularExpressions;
 using System.Xml;
 
 namespace Microsoft.PowerShell.Commands
@@ -168,28 +169,39 @@ namespace Microsoft.PowerShell.Commands
 
         private bool TryConvertToJson(string json, out object obj, ref Exception exRef)
         {
+            bool converted;
             try
             {
                 ErrorRecord error;
                 obj = JsonObject.ConvertFromJson(json, out error);
+                converted = true;
+
+                if (null == obj)
+                {
+                    string pattern = @"^\s*null\s*$";
+                    Match matches = Regex.Match(json, pattern);
+                    converted = matches.Success;
+                }
 
                 if (error != null)
                 {
                     exRef = error.Exception;
-                    obj = null;
+                    converted = false;
                 }
             }
             catch (ArgumentException ex)
             {
                 exRef = ex;
+                converted = false;
                 obj = null;
             }
             catch (InvalidOperationException ex)
             {
                 exRef = ex;
+                converted = false;
                 obj = null;
             }
-            return (null != obj);
+            return converted;
         }
 
         #endregion
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
@@ -5,8 +5,9 @@ Copyright (c) Microsoft Corporation. All rights reserved.
 using System;
 using System.Management.Automation;
 using System.IO;
-using System.Text.RegularExpressions;
 using System.Xml;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 
 namespace Microsoft.PowerShell.Commands
 {
@@ -169,36 +170,43 @@ namespace Microsoft.PowerShell.Commands
 
         private bool TryConvertToJson(string json, out object obj, ref Exception exRef)
         {
-            bool converted;
+            bool converted = false;
             try
             {
                 ErrorRecord error;
                 obj = JsonObject.ConvertFromJson(json, out error);
-                converted = true;
 
                 if (null == obj)
                 {
-                    string pattern = @"^\s*null\s*$";
-                    Match matches = Regex.Match(json, pattern);
-                    converted = matches.Success;
+                    // This ensures that a null returned by ConvertFromJson() is the actual JSON null literal. 
+                    // if not, the ArgumentException will be caught.
+                    JToken.Parse(json);
                 }
 
                 if (error != null)
                 {
                     exRef = error.Exception;
-                    converted = false;
+                    obj = null;
+                }
+                else
+                {
+                    converted = true;
                 }
             }
             catch (ArgumentException ex)
             {
                 exRef = ex;
-                converted = false;
                 obj = null;
             }
             catch (InvalidOperationException ex)
             {
                 exRef = ex;
-                converted = false;
+                obj = null;
+            }
+            catch (JsonException ex)
+            {
+                var msg = string.Format(System.Globalization.CultureInfo.CurrentCulture, WebCmdletStrings.JsonDeserializationFailed, ex.Message);
+                exRef = new ArgumentException(msg, ex);
                 obj = null;
             }
             return converted;
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -2624,9 +2624,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
         It "Invoke-RestMethod Supports a Single Value JSON null and ignores whitespace" {
             $url = '{0}{1}' -f $baseUrl, "            null         "
             Invoke-RestMethod -Uri $url | Should Be $null
-            $url = '{0}{1}' -f $baseUrl, "            
-                                            null         
-                                         "
+            $url = '{0}{1}' -f $baseUrl, "           null         `n"
             Invoke-RestMethod -Uri $url | Should Be $null
         }
     }
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -2613,6 +2613,24 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
 
     }
 
+    Context "Invoke-RestMethod Single Value JSON null support" {
+        BeforeAll {
+            $baseUrl = 'http://localhost:8081/PowerShell?test=response&contenttype=application/json&output='
+        }
+        It "Invoke-RestMethod Supports a Single Value JSON null" {
+            $url = '{0}{1}' -f $baseUrl, 'null'
+            Invoke-RestMethod -Uri $url | Should Be $null
+        }
+        It "Invoke-RestMethod Supports a Single Value JSON null and ignores whitespace" {
+            $url = '{0}{1}' -f $baseUrl, "            null         "
+            Invoke-RestMethod -Uri $url | Should Be $null
+            $url = '{0}{1}' -f $baseUrl, "            
+                                            null         
+                                         "
+            Invoke-RestMethod -Uri $url | Should Be $null
+        }
+    }
+
     BeforeEach {
         if ($env:http_proxy) {
             $savedHttpProxy = $env:http_proxy


src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
```diff
@@ -168,28 +169,39 @@ private bool TryConvertToXml(string xml, out object doc, ref Exception exRef)
 
         private bool TryConvertToJson(string json, out object obj, ref Exception exRef)
         {
+            bool converted;
             try
             {
                 ErrorRecord error;
                 obj = JsonObject.ConvertFromJson(json, out error);
+                converted = true;
```
iSazonov: I wonder why the fix so high level? It seems `ConvertFromJson` or just `PopulateFromJDictionary` is right place to fix.
We should think to return `System.Management.Automation.Internal.AutomationNull`.

Also please review [JsonSerializerSettings](https://www.newtonsoft.com/json/help/html/NullValueHandlingIgnore.htm).

markekraus: @iSazonov `ConvertFromJson()` was already returning null. The problem is that `TryConvertToJson()` was using null to mean that JSON serialization failed. We want to return null and idealy keep what ever `ConvertFromJson()` was already returning. But, `ConvertFromJson()` can return null in situations that don't mean it was JSON `null` (at least I'm pretty sure it does). In that case we should let Invoke-RestMethod continue with i's other serialization tests by returning false. This has to be fixed here because it's a problem with Invoke-RestMethod's logic, not with `ConvertFromJson()`

markekraus: Hmm i realize this is probably very confusing. The an API endpoint returns the following:

```none
HTTP/1.1 200 OK
Content-Length: 4
Content-Type: application/json
Date: Sat, 04 Nov 2017 18:49:54 GMT

null
```
that `null` is the JSON literal for `$null`. `ConvertFromJson()` properly returns `$null` for that, but `Invoke-RestMethod` was interpreting that `$null` as failing to serialize the object. `TryConvertToJson()` was returning false and `Invoke-RestMethod` would move on to other serialization methods until ultimately falling back to a string and returning the string `null`.

`ConvertFromJson()` returns null  for empty strings too, however, in this instance, an empty or whitespace string should be returned as an an empty or whitespace string, not null, because an empty or white space string is not valid JSON.  I guess it could be argued that maybe that behavior of `ConvertFromJson()` should be changed and that it should throw on an empty or whitespace string. But, I don't see that going over well.

In any case, because `ConvertFromJson()` behaves like that I'm double checking it is a true JSON `null`.

iSazonov: We have the [workarround](https://github.com/PowerShell/PowerShell/blob/4bc52d2358222084738157a08907fac32d13bd3a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs#L71). We can remove it - the bug was fixed in Newton.Json 10.0.3. But I think we can keep `JArray.Parse(input);` to throw on any error in Json input string.

markekraus: @iSazonov this has nothing to do with arrays and that code path is not hit in the instace this is addressing. This is about when `null` is NOT in a JSON array or object, but is solitary, by it self, a single value and the only value in the JSON string. this is literally the JSON string this PR addresses:

```json
null
```

This has nothing to do with either of these which already work properly:

```json
[
  null
]
````
```json
{
  "key": null
}
```

Removing that workaround in `ConvertFromJson()` will not affect the logic in this PR.

iSazonov: What about `JToken.Parse()`? I believe it throw on null string.

markekraus: Are you suggesting to use `JToken.Parse()` instead of the regex? I guess we could do that. It does throw on empty and whitespace only strings.

`ConvertFromJson()` does both. It throws and sets the `out Error`. Invalid JSON throws, trouble serializing JSON into PowerShell objects gives errors.

iSazonov: Yes, sorry I walk around :-), I think we should use `JToken.Parse()` - maybe it'll catch other mistakes we don't know yet.

As for `ConvertFromJson()` - I guess it is common convention - no throw if there is `out Error`?

markekraus: > As for `ConvertFromJson()` - I guess it is common convention - no throw if there is out Error?

I'm not sure what you are asking.

Here are the kinds of strings it throws on: `{`, `[`, `[ null null]`
Here is a string it errors on: `{ "": "value"}`

Errors arise from valid JSON that cannot serialized into PowerShell objects while exceptions arise from invalid JSON. The only invalid JSON I'm aware of that doesn't throw an exception is `String.Empty` and a string with just whitespace. An empty string and whitespace-only string are technically not valid JSON.

In any case, when there is an error it returns null. so we can just ensure that the null it is returning is an actual `null` literal with `JToken.Parse()` In `Invoke-RestMethod`.  It might be a bit redundant, but I will run it on a null check only. Otherwise, I'd rather trust everything else that comes from `ConvertFromJson()`. It's just that the existing `Invoke-RestMethod` logic was assuming null was always a failure, when null is a legitimate single value for JSON.

But, this does bring to light some shortcomings of `ConvertFromJson()`. I think you are the better person to decide how serious they are. but if we want/need to change that behavior, we need to do it before 6.0.0 as I doubt we can get away with that kind of breaking change until 7.0.0. What do you think?

markekraus:
@iSazonov I switched to `JToken.Parse()` and re did some of the logic to make it cleaner (less false assignments). 
I borrowed the the exception handling code for  `JToken.Parse()` from `ConvertFfomJson()` for consistency, though, I'm not sure that exception ever actually sees the light of day.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
iSazonov: I really don't like the API but we already have no time to discuss and make changes in depth. 😕 

I'm sure we'll be with CoreFX implementation in 7.0. 😄 So I would document the API as obsolete or make it `internal` - the later allow us to migrate on CoreFX implementation in any time before 7.0.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/InvokeRestMethodCommand.Common.cs
```diff
@@ -169,36 +170,43 @@ private bool TryConvertToXml(string xml, out object doc, ref Exception exRef)
...
-                converted = true;
 
                 if (null == obj)
                 {
-                    string pattern = @"^\s*null\s*$";
-                    Match matches = Regex.Match(json, pattern);
-                    converted = matches.Success;
+                    // This ensures that a null returned by ConvertFromJson() is the actual JSON null literal. 
+                    // if not, the ArgumentException will be caught.
+                    JToken.Parse(json);
```
iSazonov: After I see the change I'm even more confident that we should make this check in [ConvertFromJson](https://github.com/PowerShell/PowerShell/blob/4bc52d2358222084738157a08907fac32d13bd3a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs#L71) - we should remove `if ((Regex.Match(input, @"^\s*\[")).Success)` and replace `JArray.Parse(input);` with `JToken.Parse(json);`. So we'll be closer to JavaScriptSerializer, and we can make it easier to move on one in the future. Also we exclude the same catch on level above.

I review a history of the code and discover that I see ways to clean up and refactor the code. So I agree to merge the PR as is and ready to refactor in days.

markekraus: Awesome. Just note that just like the change here, making that change `ConvertFromJson()` will be a breaking change since empty strings and whitespace-only strings will produce errors where as before they returned `$null`. I think that's appropriate, but the world at large may not. But you are right, we don't have much time. Will you handle the `ConvertFromJson()` refactor?

iSazonov: I'd refactor the code in any case because it provokes troubles.

As for a breaking change I have the thoughts - should we fix the Issue now or leave the behavior as in Windows PowerShell? Will MSFT fix the Issue in Windows PowerShell too and when? Will MSFT fix the Issue in .Net JavaScriptSerializer and when? Should we request the fix in Newton.Json as we did for JArray?
I think to get sync fixes is better then a breaking change in only PowerShell Core.

markekraus: @iSazonov Good questions, but we should track that in a separate issue.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -2332,6 +2332,24 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
...
 
+    Context "Invoke-RestMethod Single Value JSON null support" {
+        BeforeAll {
+            $baseUrl = 'http://localhost:8081/PowerShell?test=response&contenttype=application/json&output='
+        }
+        It "Invoke-RestMethod Supports a Single Value JSON null" {
+            $url = '{0}{1}' -f $baseUrl, 'null'
+            Invoke-RestMethod -Uri $url | Should Be $null
+        }
+        It "Invoke-RestMethod Supports a Single Value JSON null and ignores whitespace" {
```
TravisEz13: Perhaps, add test(s) for the error case where the JSON cannot be parsed?

TravisEz13 (approved):

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: @TravisEz13 if by "error" you mean empty strings or whitspace-only? I can add the tests, but they will just be ensuring a a `String.Empty` returns a `String.Empty` and that a whitespace string returns a whitespace string. When that error condition hits, the cmdlet just goes to the next serialization type (xml) and then when that fails it just returns the string. 

It also looks like I would need  to either add a WebListener controller to another switch option to HttpListener. Currently none of the available tests can return whitespace only responses.  All of them truncate to 0 bytes when the body payload is all whitespace.

TravisEz13: It would be good to have the test to ensure good coverage, but I think this is adequate.

TravisEz13:
@PowerShell/powershell-maintainers @SteveL-MSFT @joeyaiello I recommend we take this for RC.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: I've created #5414 to add coverage for that. I have a backlog of test and cleanup activities to do, I'll start on those after RC is underway.

SteveL-MSFT:
@markekraus @TravisEz13 yes, please take for RC

iSazonov:
Are we ready to merge?

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -2332,6 +2332,24 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
...
+            $url = '{0}{1}' -f $baseUrl, "            null         "
+            Invoke-RestMethod -Uri $url | Should Be $null
+            $url = '{0}{1}' -f $baseUrl, "            
+                                            null
```
SteveL-MSFT: I would prefer to use `` `n`` for newline whitespace than rely on the line endings of the file

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1

TravisEz13 (approved):

SteveL-MSFT (approved):

SteveL-MSFT:
@iSazonov I think this is ready to merge

iSazonov:
Reopen to restart CI.

markekraus:
@iSazonov *sigh. It looks like both Travis CI tests passed, but it has not updated the PR.

iSazonov:
I can not manually merge 😕

SteveL-MSFT:
@markekraus can you rebase and pick up #5249?

markekraus:
rebased.

TravisEz13:
@markekraus  there is a merge conflict now.

markekraus:
rebased.

TravisEz13:
restarted appveyor due to github rate limit failure

PR #5341, opened 2017-11-04
Update Find-Dotnet to find dotnet with compat SDK

### LINKED ISSUE #5260: .NET Core tools version check in build.psm1 blocks the build

rkeithhill:
Fix #5260 - with this fix, folks shouldn't run afoul of the SDK version mismatch check.

Note: the branch name turns out to be a little misleading.  Read further.

The approach is to see if the dotnet in the current PATH has a compatible SDK.  Folks will have a globally installed dotnet if they've installed VS or have installed the .NET Core SDK themselves.  The global dotnet may not have access to the SDK required by PSCore.  That is, the global dotnet cannot see user local dotnet SDK installs.  So if the global dotnet doesn't have the right SDK, we prepend the path to the user's local dotnet dir.

Also, update $dotnetCLIRequiredVersion to read its value from global.json so there is one less source of the truth (for the SDK version).

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -3,7 +3,7 @@
 $script:TestModulePathSeparator = [System.IO.Path]::PathSeparator
 
 $dotnetCLIChannel = "release"
-$dotnetCLIRequiredVersion = "2.0.2"
+$dotnetCLIRequiredVersion = $(Get-Content $PSScriptRoot/global.json | ConvertFrom-Json).Sdk.Version
 
 # Track if tags have been sync'ed
 $tagsUpToDate = $false
@@ -600,7 +600,7 @@ function New-PSOptions {
         [string]$Output,
 
         [switch]$SMAOnly,
-        
+
         [switch]$PSModuleRestore
     )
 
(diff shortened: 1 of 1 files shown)


build.psm1
```diff
@@ -3,7 +3,7 @@
 $script:TestModulePathSeparator = [System.IO.Path]::PathSeparator
 
 $dotnetCLIChannel = "release"
-$dotnetCLIRequiredVersion = "2.0.2"
+$dotnetCLIRequiredVersion = $(Get-Content $PSScriptRoot/global.json | ConvertFrom-Json).Sdk.Version
```
iSazonov: Should we stop if global.json does not contain Sdk version?

rkeithhill: It would be highly unusual for global.json to be missing that field since that's about its only purpose.

bergmeister:
That's great. I found the old behaviour really annoying because every update of `VS Preview` always installed the `.Net Core SDK` (sometimes it even ended up with multiple installations of the same version!)

daxian-dbw (approved):

TravisEz13 (approved):

adityapatwardhan (approved):

PR #5346, opened 2017-11-05
Fix Start-PSBuild on WSL if repository was already built on Windows

### LINKED ISSUE #5194: Building PowerShell Core on WSL generates error
FWIW I do seem to get a valid pwsh built and I can run it.  So that is good.

Steps to reproduce
------------------
1. Make sure you're running WSL Xenial 16.04.
2. Install PowerShell Core for Ubuntu 16.04 in WSL
3. Start PowerShell Core in WSL
4. Cd to your Windows clone dir e.g. `cd /mnt/c/GitHub/PowerShell`
5. ipmo ./build.psm1
6. Start-PSBootstrap
7. Start-PSBuild

BTW this is a darn convenient way to build/test for both Windows and Linux.  No need to push your changes to origin so you can pull them to Linux to test.

Expected behavior
-----------------
pwsh builds - no errors.

Actual behavior
---------------
Get an error:
```none
PowerShell output: /mnt/c/Users/Keith/GitHub/rkeithhill/PowerShell/src/powershell-unix/bin/Linux/netcoreapp2.0/linux-x64/publish/pwsh
Copy-Item : Cannot find path 'C:/Program
Files/dotnet/sdk/NuGetFallbackFolder/microsoft.netcore.app/2.0.0/ref/netcoreapp2.0/Microsoft.CSharp.dll' because it
does not exist.
At /mnt/c/Users/Keith/GitHub/rkeithhill/PowerShell/build.psm1:528 char:9
+         Copy-Item -Path $refAssemblies -Destination $refDestFolder -F ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:/Program File...soft.CSharp.dll:String) [Copy-Item], ItemNotFoundExce
   ption
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.CopyItemCommand
```
Sure enough, the powershell.inc file includes a whole bunch of paths that start with `C:\`.

Environment data
----------------

```powershell
> $PSVersionTable
PSVersion                      6.0.0-beta.8
GitCommitId                    v6.0.0-beta.8
OS                             Linux 4.4.0-43-Microsoft #1-Microsoft Wed Dec 31 14:42:53 PST 2014
```

BTW the root of the problem is that TypeGen dir - `$PSScriptRoot/src/TypeCatalogGen` - is not platform specific.  It probably should be `$PSScriptRoot/src/TypeCatalogGen/$Runtime`.  Easiest work-around for now is `Start-PSBuild -Clean`.

bergmeister:
To fix #5194: Use the first 3 characters of the passed `-Runtime` parameter for the .inc file (the full runtime name could not be used since the separation characters ike e.g. the the hyphen would make dotnet crash when building).

A separate issue is that the runtime does not get determined if is not being passed (which is OK since dotnet takes then care of choosing the right one) but maybe given this scenario would it make sense to determine and set it as a convenience?

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -492,8 +492,8 @@ Fix steps:
     $incFileName = 'powershell.inc'
     if (-not [string]::IsNullOrEmpty($Runtime)) {
         # File name must be different for Windows and Linux to allow build on Windows and WSL.
-        # Append the first 3 characters fo the Runtime only because otherwise dotnet will crash due to the separation characters. The underscore is only for better readability.
-        $incFileName = "$($incFileName)_$($Runtime.Substring(0,3)).inc"
+        # Append the first 3 characters for the Runtime only because otherwise dotnet will crash due to the separation characters. The underscore is only for better readability.
+        $incFileName = "$($incFileName)_$($Runtime).inc"
     }
     if ($TypeGen -or -not (Test-Path "$PSScriptRoot/TypeCatalogGen/$($incFileName)")) {
         log "Run TypeGen (generating CorePsTypeCatalog.cs)"
--- a/build.psm1
+++ b/build.psm1
@@ -489,13 +489,13 @@ Fix steps:
     }
 
     # handle TypeGen
-    $incFileName = 'powershell'
+    $incFileName = 'powershell.inc'
     if (-not [string]::IsNullOrEmpty($Runtime)) {
         # File name must be different for Windows and Linux to allow build on Windows and WSL.
         # Append the first 3 characters fo the Runtime only because otherwise dotnet will crash due to the separation characters. The underscore is only for better readability.
-        $incFileName = "$($incFileName)_$($Runtime.Substring(0,3))"
+        $incFileName = "$($incFileName)_$($Runtime.Substring(0,3)).inc"
     }
-    if ($TypeGen -or -not (Test-Path "$PSScriptRoot/TypeCatalogGen/$($incFileName).inc")) {
+    if ($TypeGen -or -not (Test-Path "$PSScriptRoot/TypeCatalogGen/$($incFileName)")) {
         log "Run TypeGen (generating CorePsTypeCatalog.cs)"
         Start-TypeGen -IncFileName $incFileName
     }
@@ -521,7 +521,7 @@ Fix steps:
     # publish netcoreapp2.0 reference assemblies
     try {
         Push-Location "$PSScriptRoot/src/TypeCatalogGen"
-        $refAssemblies = Get-Content -Path "$($incFileName).inc" | Where-Object { $_ -like "*microsoft.netcore.app*" } | ForEach-Object { $_.TrimEnd(';') }
+        $refAssemblies = Get-Content -Path $incFileName | Where-Object { $_ -like "*microsoft.netcore.app*" } | ForEach-Object { $_.TrimEnd(';') }
         $refDestFolder = Join-Path -Path $publishPath -ChildPath "ref"
 
         if (Test-Path $refDestFolder -PathType Container) {
@@ -1663,7 +1663,7 @@ function Start-TypeGen
 
     Push-Location "$PSScriptRoot/src/Microsoft.PowerShell.SDK"
     try {
-        $ps_inc_file = "$PSScriptRoot/src/TypeCatalogGen/$($incFileName).inc"
+        $ps_inc_file = "$PSScriptRoot/src/TypeCatalogGen/$($incFileName)"
         dotnet msbuild .\Microsoft.PowerShell.SDK.csproj /t:_GetDependencies "/property:DesignTimeBuild=true;_DependencyFile=$ps_inc_file" /nologo
     } finally {
         Pop-Location
@@ -1671,7 +1671,7 @@ function Start-TypeGen
 
     Push-Location "$PSScriptRoot/src/TypeCatalogGen"
     try {
-        dotnet run ../System.Management.Automation/CoreCLR/CorePsTypeCatalog.cs "$($incFileName).inc"
+        dotnet run ../System.Management.Automation/CoreCLR/CorePsTypeCatalog.cs $incFileName
     } finally {
         Pop-Location
     }
--- a/build.psm1
+++ b/build.psm1
@@ -488,13 +488,9 @@ Fix steps:
         Start-ResGen
     }
 
-    # handle TypeGen
-    $incFileName = 'powershell.inc'
-    if (-not [string]::IsNullOrEmpty($Runtime)) {
-        # File name must be different for Windows and Linux to allow build on Windows and WSL.
-        # Append the first 3 characters for the Runtime only because otherwise dotnet will crash due to the separation characters. The underscore is only for better readability.
-        $incFileName = "$($incFileName)_$($Runtime).inc"
-    }
+    # Handle TypeGen
+    # .inc file name must be different for Windows and Linux to allow build on Windows and WSL.
+    $incFileName = "powershell_$($Options.Runtime).inc"
     if ($TypeGen -or -not (Test-Path "$PSScriptRoot/TypeCatalogGen/$($incFileName)")) {
         log "Run TypeGen (generating CorePsTypeCatalog.cs)"
         Start-TypeGen -IncFileName $incFileName
--- a/build.psm1
+++ b/build.psm1
@@ -491,7 +491,9 @@ Fix steps:
     # handle TypeGen
     $incFileName = 'powershell'
     if (-not [string]::IsNullOrEmpty($Runtime)) {
-        $incFileName = "$($incFileName)_$($Runtime.Substring(0,3))" # use the first 3 characters only because otherwise dotnet will crash due to the separation characters
+        # File name must be different for Windows and Linux to allow build on Windows and WSL.
+        # Append the first 3 characters fo the Runtime only because otherwise dotnet will crash due to the separation characters. The underscore is only for better readability.
+        $incFileName = "$($incFileName)_$($Runtime.Substring(0,3))"
     }
     if ($TypeGen -or -not (Test-Path "$PSScriptRoot/TypeCatalogGen/$($incFileName).inc")) {
         log "Run TypeGen (generating CorePsTypeCatalog.cs)"
--- a/build.psm1
+++ b/build.psm1
@@ -489,9 +489,13 @@ Fix steps:
     }
 
     # handle TypeGen
-    if ($TypeGen -or -not (Test-Path "$PSScriptRoot/src/System.Management.Automation/CoreCLR/CorePsTypeCatalog.cs")) {
+    $incFileName = 'powershell'
+    if (-not [string]::IsNullOrEmpty($Runtime)) {
+        $incFileName = "$($incFileName)_$($Runtime.Substring(0,3))" # use the first 3 characters only because otherwise dotnet will crash due to the separation characters
+    }
+    if ($TypeGen -or -not (Test-Path "$PSScriptRoot/TypeCatalogGen/$($incFileName).inc")) {
         log "Run TypeGen (generating CorePsTypeCatalog.cs)"
-        Start-TypeGen
+        Start-TypeGen -IncFileName $incFileName
     }
 
     # Get the folder path where pwsh.exe is located.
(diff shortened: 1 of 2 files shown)


rkeithhill:
This should slot in nicely with my tweak to build.psm1 #5341

rkeithhill:
I think you need to update this .gitignore file to make it handle the platform specific inc files:
```
40:121ms> gc .\src\TypeCatalogGen\.gitignore
powershell.inc
```
I was noticing that Git thought I had a new file with powershell_win.inc.

bergmeister:
Thanks. Well spotted, I updated it to ignore now `powershell*.inc`. I find it a bit odd that there is a special `.gitignore` file in this folder and am wondering if we can move this to the top level `.gitignore` file?

andschwa:
@bergmeister Because this ignore rule doesn't apply to anywhere else in the repository, just keeping it simple 😄

bergmeister:
@andschwa My thinking is rather that if there will never be any other `.inc` files somewhere else, it would be nice to put it into the top one to have a more central place, especially since there is only one entry in it. But this is probably more of a stylistic preference with no 'right' answer depending on how big the central place can become (to me the output of `gci -Recurse .gitignore | % {cat $_}` did not seem to be too long)... But I'll leave it then as it is if that's your preference.

build.psm1
```diff
@@ -489,9 +489,13 @@ Fix steps:
     }
 
     # handle TypeGen
-    if ($TypeGen -or -not (Test-Path "$PSScriptRoot/src/System.Management.Automation/CoreCLR/CorePsTypeCatalog.cs")) {
+    $incFileName = 'powershell'
+    if (-not [string]::IsNullOrEmpty($Runtime)) {
+        $incFileName = "$($incFileName)_$($Runtime.Substring(0,3))" # use the first 3 characters only because otherwise dotnet will crash due to the separation characters
```
TravisEz13: I believe our guidance is to put the comment on a separate line.

TravisEz13: you could replace the seperation character.... not sure it's needed though.

TravisEz13 (approved):
Looks good with minor comments.

@daxian-dbw  and/or @iSazonov  should review as well.

build.psm1
bergmeister: You're right about the comment. I noticed that only after the commit but I will fix that. The separation character is only for readability because to me `powershell_win` or `powerhell_lin` is easier and faster to read than `powershellwin` or `powershelllin`

bergmeister:
NB: The build was read because NuGet has a global outage (see [here](https://twitter.com/nuget/status/927630886509539332))

build.psm1
```diff
@@ -489,9 +489,15 @@ Fix steps:
...
 
     # handle TypeGen
-    if ($TypeGen -or -not (Test-Path "$PSScriptRoot/src/System.Management.Automation/CoreCLR/CorePsTypeCatalog.cs")) {
+    $incFileName = 'powershell'
+    if (-not [string]::IsNullOrEmpty($Runtime)) {
+        # File name must be different for Windows and Linux to allow build on Windows and WSL.
+        # Append the first 3 characters fo the Runtime only because otherwise dotnet will crash due to the separation characters. The underscore is only for better readability.
+        $incFileName = "$($incFileName)_$($Runtime.Substring(0,3))"
+    }
+    if ($TypeGen -or -not (Test-Path "$PSScriptRoot/TypeCatalogGen/$($incFileName).inc")) {
```
iSazonov: We use `$incFileName).inc` so many times so make sense add:
```powershell
$incFileName = ($incFileName).inc
```

iSazonov:
I'd prefer to fix the issue more natively like #3870.

Also I'm not sure that's the only runtime addiction. Should we clean up if our files contains "#if UNIX"?

build.psm1
bergmeister: Ok, I have addressed that.

build.psm1
```diff
@@ -489,9 +489,15 @@ Fix steps:
     }
 
     # handle TypeGen
-    if ($TypeGen -or -not (Test-Path "$PSScriptRoot/src/System.Management.Automation/CoreCLR/CorePsTypeCatalog.cs")) {
+    $incFileName = 'powershell.inc'
+    if (-not [string]::IsNullOrEmpty($Runtime)) {
+        # File name must be different for Windows and Linux to allow build on Windows and WSL.
+        # Append the first 3 characters fo the Runtime only because otherwise dotnet will crash due to the separation characters. The underscore is only for better readability.
+        $incFileName = "$($incFileName)_$($Runtime.Substring(0,3)).inc"
```
daxian-dbw: Maybe just use `powershell_$($Runtime.Substring(0,3)).inc`? Otherwise it would beome `powershell.inc_xxx.inc`

build.psm1
```diff
@@ -489,9 +489,15 @@ Fix steps:
...
+    }
+    if ($TypeGen -or -not (Test-Path "$PSScriptRoot/TypeCatalogGen/$($incFileName)")) {
```
daxian-dbw: `$($incFileName)` can be replaced by `$incFileName`

build.psm1
```diff
@@ -1631,7 +1637,12 @@ function Start-DevPowerShell {
 function Start-TypeGen
 {
     [CmdletBinding()]
-    param()
+    param
+    (
+        [Parameter(Mandatory=$true)]
+        [ValidateNotNullOrEmpty()]
+        $IncFileName
```
daxian-dbw: Maybe we should have 'powershell.inc' as the default value for this parameter, so that `Start-TypeGen` without parameter will continue to work.

build.psm1
```diff
@@ -489,9 +489,15 @@ Fix steps:
     }
 
     # handle TypeGen
-    if ($TypeGen -or -not (Test-Path "$PSScriptRoot/src/System.Management.Automation/CoreCLR/CorePsTypeCatalog.cs")) {
+    $incFileName = 'powershell.inc'
+    if (-not [string]::IsNullOrEmpty($Runtime)) {
```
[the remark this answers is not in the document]
daxian-dbw: The `$Options.Runtime` will always be 'NotNullOrEmpty', so maybe you can have:
```
$incFileName = "powershell_$($Runtime.Substring(0,3)).inc"
```

build.psm1
```diff
@@ -1652,15 +1663,15 @@ function Start-TypeGen
 
     Push-Location "$PSScriptRoot/src/Microsoft.PowerShell.SDK"
     try {
-        $ps_inc_file = "$PSScriptRoot/src/TypeCatalogGen/powershell.inc"
+        $ps_inc_file = "$PSScriptRoot/src/TypeCatalogGen/$($incFileName)"
```
daxian-dbw: `$($incFileName)` can be replaced by `$incFileName`

daxian-dbw (changes requested):

bergmeister:
I'll address the comments next Monday when I'm back from travelling.

build.psm1
bergmeister: No, this is one of the gotchas about PowerShell string interpolation because PowerShell allows underscores for variable names. Therefore PowerShell would then look for a variable named `incFileName_` instead of `incFileName`. As a general rule, I always wrap variables in a string inside `$()` because it also safeguards against cases like calling a property on an object...

build.psm1
bergmeister: I can do that if people actually call `Start-TypeGen`. I always use the `-ResGen` swich on `Start-PSBuild` because the reason why I need update ResGen is because I need to build anyway.

build.psm1
bergmeister: In this case we can do that but what do you think about my comment above where this is not always possible?

build.psm1
```diff
@@ -489,9 +489,15 @@ Fix steps:
     }
 
     # handle TypeGen
-    if ($TypeGen -or -not (Test-Path "$PSScriptRoot/src/System.Management.Automation/CoreCLR/CorePsTypeCatalog.cs")) {
+    $incFileName = 'powershell.inc'
+    if (-not [string]::IsNullOrEmpty($Runtime)) {
+        # File name must be different for Windows and Linux to allow build on Windows and WSL.
+        # Append the first 3 characters fo the Runtime only because otherwise dotnet will crash due to the separation characters. The underscore is only for better readability.
```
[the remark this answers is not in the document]
bergmeister: Thanks for spotting the typo and thanks for coming back about the crash. When I started out I was experimenting with different approaches (one of was to put the inc file into a runtime specific folder) and I actually cannot repro the crash anymore and will therefore use the full `$RunTime` string instead of the first 3 characters.

build.psm1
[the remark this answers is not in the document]
bergmeister: Good idea. Yes, we can use `$Options.Runtime` indeed instead. I was not aware that you already figure it out automatically when it is not being provided. Shall I maybe add this information as a comment to the `$Runtime` parameter?

build.psm1
bergmeister: This change has now been undone since we do not need that logic any more due to the usage of `$Options.Runtime`

bergmeister:
@daxian-dbw Thanks for your useful feedback. I have addressed all issues and left comments there. But basically the 2 remaining questions are:

- Should we be strict about string interpolation (i.e. use `$()`) since there are gotchas as I pointed out?
- Should `Start-TypeGen` have a default value for the new `$IncFileName` parameter or should we automatically determine it in there similarly as in `Start-PSBuild`?

build.psm1
daxian-dbw: It's not commonly needed, but when debugging anything about the type catalog, it's handy to just run `Start-TypeGen` without building the rest of the projects. For that purpose, a simple default value like `powershell.inc` would be sufficient, no need to call `New-Options` to figure out the runtime.

build.psm1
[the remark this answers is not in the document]
daxian-dbw: You mean the `-Runtime` parameter of `Start-PSBuild`? Sure you can add a comment there.

build.psm1
```diff
@@ -488,10 +488,12 @@ Fix steps:
         Start-ResGen
     }
 
-    # handle TypeGen
-    if ($TypeGen -or -not (Test-Path "$PSScriptRoot/src/System.Management.Automation/CoreCLR/CorePsTypeCatalog.cs")) {
+    # Handle TypeGen
+    # .inc file name must be different for Windows and Linux to allow build on Windows and WSL.
+    $incFileName = "powershell_$($Options.Runtime).inc"
+    if ($TypeGen -or -not (Test-Path "$PSScriptRoot/TypeCatalogGen/$($incFileName)")) {
```
daxian-dbw: > $($incFileName)

Subexpression is not needed here, you can just use `$incFileName`

build.psm1
daxian-dbw: In the case of `"$($incFileName)_xxx.inc"`, you can replace it with `"${incFileName}_xxx.inc"`

build.psm1
daxian-dbw: It's not commonly needed, but when debugging anything about the type catalog, it's handy to just run `Start-TypeGen` without building the rest of the projects. For that purpose, a simple default value like `powershell.inc` would be sufficient, no need to call `New-Options` to figure out the runtime.

build.psm1
daxian-dbw: PowerShell support this syntax for variable: `${var name}`, so you can use `"${varname}_xxx.inc"` in that case.

daxian-dbw (changes requested):

build.psm1
daxian-dbw: And a minor comment: the variable name you used in the rest of this function is `incFileName`, maybe it can be changed to `IncFileName`.

daxian-dbw:
@bergmeister For the two questions, I answered them in the corresponding comment threads: https://github.com/PowerShell/PowerShell/pull/5346#discussion_r150988197 and https://github.com/PowerShell/PowerShell/pull/5346#discussion_r150988544

bergmeister:
@daxian-dbw Thanks for the review. I have addressed your comments.

daxian-dbw (approved):
LGTM. Thanks @bergmeister!

TravisEz13 (approved):

Issue #5205, opened 2017-10-23
PowerShell hangs when exiting having used a PowerShell runspace pool/runspace


tylerl0706:
PowerShell Core hangs when exiting and using a PowerShell runspace pool. On Windows PowerShell it exits successfully.

Steps to reproduce
------------------
here's the gist for my code: https://gist.github.com/tylerl0706/2d87a13c6147278bb4e8565585aa9722

Steps:
```powershell
PS > Add-Type -Path "path/to/TestHttpListener/bin/Debug/netstandard2.0/TestHttpListener.dll"
PS > $app = [TestHttpListener.TestHttpListener]::new()
PS > $app.Start(8081)
PS > Invoke-WebRequest -Uri http://localhost:8081 # VERY IMPORTANT
PS > exit
```

Expected behavior
-----------------
PowerShell successfully exists

Actual behavior
---------------
PowerShell hangs and doesn't give the prompt back - this is causing my AppVeyor and Travis builds to fail due to timeout :( 

Environment data
----------------

```powershell
> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.0.0-beta.8
PSEdition                      Core
GitCommitId                    v6.0.0-beta.8
OS                             Darwin 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64
Platform                       Unix
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

tylerl0706:
@vors and I were both able to repro this locally.

SteveL-MSFT:
I think I have a simpler repro without using httplistener

```powershell
PS /Users/steve> $rs = [runspacefactory]::CreateRunspacePool(1,5)                                                                                               
PS /Users/steve> $rs.Open()                                                                                                                                     
PS /Users/steve> $ps = [powershell]::Create()                                                                                                                   
PS /Users/steve> $ps.RunspacePool = $rs                                                                                                                         
PS /Users/steve> $null = $ps.AddScript({1+1})                                                                                                                           
PS /Users/steve> $ps.Invoke()
PS /Users/steve> exit #hangs            
```

PaulHigin:
[...]

PR #5356, opened 2017-11-06
Fix for PowerShell hang on exit

PaulHigin:
This change addresses the issue #5205

Repro steps are:
```powershell
$rs = [runspacefactory]::CreateRunspacePool(1,5)
$rs.Open()

$ps1 = [powershell]::Create()
$ps1.RunspacePool = $rs
$null = $ps1.AddScript({sleep 31; 1+1})
$ps1.BeginInvoke()

$ps2 = [powershell]::Create()
$ps2.RunspacePool = $rs
$null = $ps2.AddScript({sleep 32; 1+1})
$ps2.BeginInvoke()

$ps3 = [powershell]::Create()
$ps3.RunspacePool = $rs
$null = $ps3.AddScript({sleep 33; 1+1})
$ps3.BeginInvoke()

exit #hangs
```
The problem is that the latest version of CoreCLR doesn't seem to be calling finalizers (or the finalizer thread dies for some reason and finalizers are not getting called).  

PowerShell relies on the CLR finalizer to clean up state on exit.  In this case a runspace pool was not closed or disposed and any pipeline worker threads created to run concurrent scripts won't end, causing the hang.  The same thing can happen if any individual runspace is created to run concurrent script and is not closed.

PowerShell really needs to rely on CLR finalizer to clean up because it can be hosted in many ways and it cannot know when a session is being intentionally ended or some helper session is being closed.

I'll contact the DotNet CoreCLR team to see if this is a known issue.

But in the mean time I have a work around to prevent the hang.  The workaround uses existing code that tries to determine when a session is ending and stop transcription.  This code has a bug and I have rewritten it to use a new Runspace.IsInteractiveHost property.  When this is set to true in indicates to the runspace clean up code that the entire PowerShell session is ending and that any remaining active runspaces should be closed (along with pipeline threads).  This way any application hosting PowerShell can use the property to ensure full runspace cleanup on exit.


PaulHigin:
FYI  I have created an Issue about finalizers not being called for the CLR team(https://github.com/dotnet/coreclr/issues/14889).

PaulHigin:
It turns out that this "CLR no longer calling finalizer on shutdown" is by design.  Making it work across multiple platforms is probably much harder than making it work just on Windows.

I am going to think about this some more so don't bother reviewing this just yet.  I will look into using the AppDomain.DomainUnload.Unload event.  I suspect it is no more reliable than the finalizer thread, since it is just another thread for clean up, after all.  But we can use it as notification for process shut down.

PaulHigin:
I have removed the "On Hold" and "Workaround" designations from the title.  After talking with the CLR team, we can no longer rely on CLR events or finalizer calls on exit.  The AppDomain.DomainUnload is not supported.  The AppDomain.ProcessExit event is not helpful since it is only called during application exit which means threads already have to be cleaned up, plus there is a time limit.

So we will need to do our own clean up in this change.  
I have introduced a new Runspace property (*IsInteractiveHost*) to indicate that runspace clean up must be performed when the interactive host runspace closes.  I made this public so that other hosts can use it as needed.

src/System.Management.Automation/engine/hostifaces/Connection.cs
```diff
@@ -631,6 +631,14 @@ public bool RunspaceIsRemote
         }
 
         /// <summary>
+        /// When true this property indicates that this runspace is associated with an interactive session,
+        /// such as a runspace for the PowerShell console.
+        /// A runspace designated as interactive host means that it is the primary runspace, and when it closes
+        /// any other open runspace on the system should be forcibly closed.
+        /// </summary>
+        public bool IsInteractiveHost { get; set; }
```
daxian-dbw: Talked to Paul offline about this property -- how to make sure this property is set only once so that there won't be a misuse of the property. One feedback is to make it internal for now, so that any follow-up work on this matter between RC and GA won't be a breaking change.

daxian-dbw: And the name of the property perhaps can be changed to `IsDefaultRunspace`, without the `Host` in it, so there is less confusion about Runspace vs. Host.

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
```diff
@@ -811,48 +811,40 @@ private void CloseThreadProc()
...
-                            return;
-                        }
-                    }
+            // When closing the interactive host runspace, ensure all other local runspaces are closed.
+            var closeAllOpenRunspaces = this.IsInteractiveHost && haveOpenRunspaces;
 
-                    PSHostUserInterface host = executionContext.EngineHostInterface.UI;
-                    if (host != null)
+            // Stop all transcriptions and unitialize AMSI if we're the last runspace to exit or we are an exiting interactive host.
+            if (this.IsInteractiveHost || !haveOpenRunspaces)
```
daxian-dbw: Perhaps this check should be `if (!haveOpenRunspaces)`. When closing the default runspace, it's possible there are other runspaces still running code, which may be calling into `AmsiUtils`, so calling `AmsiUtils.Uninitialize()` may cause unexpected error.
`AmsiUtils.Uninitialize()` and `StopAllTranscribing()` will called anyway when the last runspace is being closed.

daxian-dbw: It's just my thought, I didn't verify the scenario of calling `AmsiUtils.Uninitialize()` while having another runspace running.

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
```diff
@@ -897,6 +889,18 @@ private void DoCloseHelper()
             //Raise Event
             RaiseRunspaceStateEvents();
 
+            // Close all open runspaces to ensure that any associated pipeline thread is released.
+            if (closeAllOpenRunspaces)
```
daxian-dbw: I'm thinking about ISE here. Each tab is an interactive host with a different Runspace, so which one should be the default Runspace? 
A user can type `ctrl+t` and `ctrl+w` to create a new tab and close an existing tab, so it's very common that the first tab gets closed by `ctrl+w` later. If the first tab's Runspace is set to be the default, then closing that tab would terminate all tabs ...

PaulHigin:
Per offline discussion with Dongbo I will make some changes.  We should not use the term "Default Runspace" because it has a current meaning for a runspace per thread that is available to use for running scriptblocks.  This clean up code is meant to be used only once per process session and is intended to prevent process hang on exit due to dangling threads (because runspaces are not cleaned up).

Therefore this property should be called "PrimaryRunspace" and should be static.  When the PrimaryRunspace is closing it means that the PowerShell session is ending and on exit clean should be performed.  I will make it internal for now but feel it should eventually be public so that other hosters of PowerShell can use it.  The purpose of the property is to do on exit clean up.

I'll submit a new commit with these changes.

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
```diff
@@ -811,48 +811,41 @@ private void CloseThreadProc()
...
-                            return;
-                        }
-                    }
+            // When closing the primary runspace, ensure all other local runspaces are closed.
+            var closeAllOpenRunspaces = isPrimaryRunspace && haveOpenRunspaces;
 
-                    PSHostUserInterface host = executionContext.EngineHostInterface.UI;
-                    if (host != null)
+            // Stop all transcriptions and unitialize AMSI if we're the last runspace to exit or we are exiting the primary runspace.
+            if (isPrimaryRunspace || !haveOpenRunspaces)
```
daxian-dbw: Can you please respond to my previous comment here?
> Perhaps this check should be `if (!haveOpenRunspaces)`. When closing the default runspace, it's possible there are other runspaces still running code, which may be calling into `AmsiUtils`, so calling `AmsiUtils.Uninitialize()` may cause unexpected error.
`AmsiUtils.Uninitialize()` and `StopAllTranscribing()` will called anyway when the last runspace is being closed.

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
```diff
@@ -897,8 +890,19 @@ private void DoCloseHelper()
...
+                {
+                    if (runspace.RunspaceStateInfo.State == RunspaceState.Opened)
+                    {
+                        runspace.Dispose();
+                    }
+                }
+            }
 
+            // Report telemetry if we have no more open runspaces.
+            // TODO: Is this still needed?
```
daxian-dbw: This is needed. The old telemetry code is kept as references for the later rewriting work.

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
```diff
@@ -897,8 +890,19 @@ private void DoCloseHelper()
             //Raise Event
             RaiseRunspaceStateEvents();
 
-            // Report telemetry if we have no more open runspaces.
+            if (closeAllOpenRunspaces)
```
daxian-dbw: This solution won't work well with ISE-like applications. Maybe we should open an issue to track it.
> I'm thinking about ISE here. Each tab is an interactive host with a different Runspace, so which one should be the default Runspace? 
A user can type `ctrl+t` and `ctrl+w` to create a new tab and close an existing tab, so it's very common that the first tab gets closed by `ctrl+w` later. If the first tab's Runspace is set to be the default, then closing that tab would terminate all tabs ...

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
PaulHigin: I'll remove the comment.

src/System.Management.Automation/engine/hostifaces/Connection.cs
```diff
@@ -521,6 +521,35 @@ public static Runspace DefaultRunspace
...
+                {
+                    if (s_primaryRunspace == null)
+                    {
+                        s_primaryRunspace = value;
+                    }
+                    else
+                    {
+                        throw new PSInvalidOperationException(RunspaceStrings.PrimaryRunspaceAlreadySet);
+                    }
+                }
```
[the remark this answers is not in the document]
PaulHigin: A race condition under a lock?  How can that happen?

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
PaulHigin: I agree.  Will just use haveOpenRunspaces.

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
PaulHigin: Yes, if there is no primary runspace then there is a problem with auto clean up.  In this case the static variable should not be used and the application needs to do its own cleanup on exit.  I don't think an issue needs to be created since this is just being used internally right now for PowerShell shell.

src/System.Management.Automation/engine/hostifaces/Connection.cs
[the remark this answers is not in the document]
daxian-dbw: Ah, a lock ... I missed that. I guess I came too early this morning :)

src/System.Management.Automation/engine/hostifaces/LocalConnection.cs
daxian-dbw: >  I don't think an issue needs to be created since this is just being used internally right now for PowerShell shell.

Agreed. #Closed

daxian-dbw (commented):
LGTM. Are you going to add tests for this fix?

daxian-dbw (approved):

daxian-dbw:
Merging this PR. I will add the test to this fix soon.

Issue #1908, opened 2016-08-18
Don’t parse the pipeline as text when it is directed from an EXE to another EXE or file. Keep the bytes as-is.


be5invis:
Currently PowerShell parses STDOUT as string when piping from an EXE, while in some cases it should be preserved as a byte stream, like this scenario:

``` powershell
curl.exe http://whatever/a.png > a.png
```

or

``` powershell
node a.js | gzip -c > out.gz
```

Affected patterns include: `native | native`, `native > file` and (maybe) `cat file | native`.

be5invis:
@vors @lzybkr 
The current `NativeCommandProcessor` breaks:

- LF line endings.
- Non-ASCII text within UTF-8 without BOM header.
- Binary file redirects (like `curl.exe`’s output).
- `>` layouts text into 80 columns by default.

ForNeVeR:
Maybe it'd be better to add another set of pipe operators that wouldn't parse the output as strings?

ForNeVeR:
Or add a cmdlet/operator to call native command and get its raw output (as a byte array / stream?), something like this:

```powershell
$output = ^& curl.exe http://whatever/a.png # $output now is a byte array or stream
$output > C:\Temp\file.png # file.png now is a valid image file

# This should be valid, too:
^& curl.exe http://whatever/a.png > C:\Temp\file.png
```

This opens an opportunity for some additional usage patterns (you can put this raw content to variables, and pipe raw content from native commands to managed cmdlets).

ForNeVeR:
But maybe we could add a special kind of redirection operator (like `2>&1`, `3>&1`, `*>1` we already have), something like this (where `%>1` is a new redirection operator that redirects command "raw output" without processing it as a string):

```powershell
$output = curl.exe http://whatever/a.png %>1
$output > C:\Temp\file.png

# Or even this:
curl.exe http://whatever/a.png %> C:\Temp\file.png # which is just awesome
```

be5invis:
@ForNeVeR My proposal is that:

1. For `native | native`, keep the bytes as-is. This is already purposed by @vors.
2. For `ps | native`, add a set of cmdlets which encodes PS objects into bytes, perhaps `ps | encode-text utf-8 | native`.
3. For `native | ps`, we can use the type system to identify whether a cmdlet accepts “raw input”. For cmdlets like `out-file` or maybe `decode-text`, it will keep the bytes from `native`, and other cmdlets will use the parsed string as its input.

[...]

Issue #4681, opened 2017-08-27
Prepare for BOM-less UTF-8 default character encoding with respect to $OutputEncoding and console code page


mklement0:
BOM-less UTF-8 character encoding [is coming as the default for PowerShell Core on all platforms](https://github.com/PowerShell/PowerShell-RFC/issues/71#issuecomment-306614751).

Two attendant changes are required:

* Preference variable `$OutputEncoding`, which currently defaults to ASCII, must default to `[System.Text.UTF8Encoding]::new()` (UTF-8 with no BOM), or, perhaps preferably, _not_ predefine this variable and default to that encoding (the internally used default) in its absence.

  * `$OutputEncoding` tells PowerShell what character encoding to use when sending output _to_ external utilities.

* Console / terminal character encoding:

  * On _Windows_, `[Console]::InputEncoding` and `[Console]::OutputEncoding` must both be set to `[System.Text.UTF8Encoding]::new()`, which is the equivalent of configuring a console window to use code page `65001` (UTF-8) or executing `chcp 65001` _before_ PowerShell is launched.

    * `[Console]::OutputEncoding` tells PowerShell what encoding to assume when reading output _from_ external utilities.

    * On Windows, the Start Menu shortcut that is created during installation should be preconfigured to open a console window with code page `65001`.

      * Conceivably, PowerShell should _automatically_ switch to the `65001` code page in case it is launched from a console window with a different active code page (such as from `cmd.exe`), though it is worth noting that this change in encoding by default remains in effect until the window is closed (even after exiting PowerShell and returning to `cmd.exe`; perhaps a warning could be issued on startup).

  * On _Unix_ platforms with UTF-8-based locales, which are the norm these days, no action is required.

     * To be determined: How should the rare event of being invoked from a terminal with a different active character encoding be handled? Changing the encoding on the fly, as on Windows, is not guaranteed to work. Perhaps a warning on startup is sufficient.

---

Before the above is implemented, the **interim workaround** to make a console window / terminal use UTF-8 consistently is the following command:

```powershell
$OutputEncoding = [console]::InputEncoding = [console]::OutputEncoding = [System.Text.UTF8Encoding]::new()

```

Environment data
----------------

```powershell
PowerShell Core v6.0.0-beta.6
```

iSazonov:
It is already addressed in https://github.com/PowerShell/PowerShell/pull/4119

mklement0:
[...]

PR #5361, opened 2017-11-07
Change $OutputEncoding to use iso-8859-1 encoding rather than ASCII

JamesWTruher:
This enables a number of useful scenarios.
With this change it is now possible to do the following:

```powershell
get-content -raw -encoding iso8859 /tmp/archive.tgz | gunzip | tar xvf -
```
and expand a compressed tar archive.
It also ensures that the bytes/characters passed to a native executable are not changed by the pipeline.
What is passed into the pipeline is passed to the next native application

This does _not_ address https://github.com/PowerShell/PowerShell/issues/1908, as that requires a larger re-architecture of the pipeline. This PR continues to have a `[environment]::newline` which is added to the native output which should be addressed when 1908 is fixed.

This is marked as a breaking change because the output of the pipeline is not altered as it was previously, anyone relying on ascii encoding (especially for extended ascii sequences) will no longer have that behavior (broken though it was).

### DIFF
--- a/src/System.Management.Automation/engine/InitialSessionState.cs
+++ b/src/System.Management.Automation/engine/InitialSessionState.cs
@@ -4410,7 +4410,7 @@ end
             // Variable which controls the encoding for piping data to a NativeCommand
             new SessionStateVariableEntry(
                 SpecialVariables.OutputEncoding,
-                System.Text.Encoding.ASCII,
+                Utils.iso8859,
                 RunspaceInit.OutputEncodingDescription,
                 ScopedItemOptions.None,
                 new ArgumentTypeConverterAttribute(typeof(System.Text.Encoding))
--- a/src/System.Management.Automation/engine/NativeCommandProcessor.cs
+++ b/src/System.Management.Automation/engine/NativeCommandProcessor.cs
@@ -1797,8 +1797,7 @@ namespace System.Management.Automation
             //Get the encoding for writing to native command. Note we get the Encoding
             //from the current scope so a script or function can use a different encoding
             //than global value.
-            Encoding pipeEncoding = _command.Context.GetVariableValue(SpecialVariables.OutputEncodingVarPath) as System.Text.Encoding ??
-                                    Encoding.ASCII;
+            Encoding pipeEncoding = _command.Context.GetVariableValue(SpecialVariables.OutputEncodingVarPath) as System.Text.Encoding ??  Utils.iso8859;
 
             _streamWriter = new StreamWriter(process.StandardInput.BaseStream, pipeEncoding);
             _streamWriter.AutoFlush = true;
--- a/src/System.Management.Automation/engine/Utils.cs
+++ b/src/System.Management.Automation/engine/Utils.cs
@@ -1239,8 +1239,9 @@ namespace System.Management.Automation
             (char) 21, (char) 22, (char) 23, (char) 24, (char) 25, (char) 26, (char) 28, (char) 29, (char) 30,
             (char) 31, (char) 127, (char) 129, (char) 141, (char) 143, (char) 144, (char) 157 };
 
-        internal static readonly UTF8Encoding utf8NoBom =
-            new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
+        internal static readonly UTF8Encoding utf8NoBom = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
+        // encoding which is essentially a passthru
+        internal static readonly Encoding iso8859 = Encoding.GetEncoding(28591);
 
 #if !CORECLR // TODO:CORECLR - WindowsIdentity.Impersonate() is not available. Use WindowsIdentity.RunImpersonated to replace it.
         /// <summary>
--- a/test/tools/TestExe/TestExe.cs
+++ b/test/tools/TestExe/TestExe.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Text;
 using System.Threading;
 using System.Diagnostics;
 
@@ -17,6 +18,12 @@ namespace TestExe
                         break;
                     case "-createchildprocess":
                         CreateChildProcess(args);
+                        break;
+                    case "-writeoutput":
+                        WriteStandardOutput();
+                        break;
+                    case "-readinput":
+                        ReadStandardInput(args);
                         break;
                     default:
                         Console.WriteLine("Unknown test {0}", args[0]);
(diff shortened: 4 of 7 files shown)


src/System.Management.Automation/engine/NativeCommandProcessor.cs
```diff
@@ -1797,8 +1797,7 @@ internal void Start(Process process, NativeCommandIOFormat inputFormat)
             //Get the encoding for writing to native command. Note we get the Encoding
             //from the current scope so a script or function can use a different encoding
             //than global value.
-            Encoding pipeEncoding = _command.Context.GetVariableValue(SpecialVariables.OutputEncodingVarPath) as System.Text.Encoding ??
-                                    Encoding.ASCII;
+            Encoding pipeEncoding = _command.Context.GetVariableValue(SpecialVariables.OutputEncodingVarPath) as System.Text.Encoding ??  Utils.iso8859;
```
daxian-dbw: Minor comment: there is an extra space before `Utils.iso8859`.

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -4410,7 +4410,7 @@ .ForwardHelpCategory Cmdlet
             // Variable which controls the encoding for piping data to a NativeCommand
```
daxian-dbw: It would be great if you can add more comments about the characteristics of this encoding and mention that's why we use it as the default output encoding.

test/tools/TestExe/TestExe.cs
```diff
@@ -40,6 +47,48 @@ static void EchoArgs(string[] args)
...
+                        ;
+                    }
+                }
+            }
+            if ( encodingToUse == null )
+            {
+                encodingToUse = Encoding.GetEncoding(28591);
+            }
+            Console.InputEncoding = encodingToUse;
+            string pipedText = Console.In.ReadToEnd();
```
daxian-dbw: Maybe we should just spit out the characters read in without having the encoding involved? For example, like this:
```
int ch;
while ((ch = Console.In.Read()) != -1)
{
    Console.WriteLine(ch);
}
```

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -1239,8 +1239,9 @@ internal static Encoding GetEncoding(string path)
             (char) 21, (char) 22, (char) 23, (char) 24, (char) 25, (char) 26, (char) 28, (char) 29, (char) 30,
             (char) 31, (char) 127, (char) 129, (char) 141, (char) 143, (char) 144, (char) 157 };
 
-        internal static readonly UTF8Encoding utf8NoBom =
-            new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
+        internal static readonly UTF8Encoding utf8NoBom = new UTF8Encoding(encoderShouldEmitUTF8Identifier: false);
+        // encoding which is essentially a passthru
+        internal static readonly Encoding iso8859 = Encoding.GetEncoding(28591);
```
SteveL-MSFT: You should use the overload that takes a string instead:

```c#
GetEncoding("iso-8859-1");
```

daxian-dbw (commented):
@JamesWTruher Per our discussion, you are going to add newlines when reading the redirected output from a native command. I don't see that change in this PR. Will that be a separate PR?

test/tools/TestExe/TestExe.cs
```diff
@@ -40,6 +47,48 @@ static void EchoArgs(string[] args)
             }
         }
 
+        static void WriteStandardOutput()
+        {
+            // write a string of characters (bytes), this closest resembles a binary
+            // write the string "test" with an accent over the e
+            byte[] testbytes = new byte[] { 116, 233, 115, 116 };
```
SteveL-MSFT: Why not just write all the bytes from 0 to 255?

test/tools/TestExe/TestExe.cs
```diff
@@ -40,6 +47,48 @@ static void EchoArgs(string[] args)
...
+                int result;
+                if ( int.TryParse(args[1], out result))
+                {
+                    try
+                    {
+                        encodingToUse = Encoding.GetEncoding(result);
+                    }
+                    catch
+                    {
+                        ;
```
SteveL-MSFT: Seems like you should fail this and return non-zero exit code if invalid encoding is provided.  Also suggest using the string overload version instead of the int making the code easier to read.

test/tools/TestExe/TestExe.cs
```diff
@@ -40,6 +47,48 @@ static void EchoArgs(string[] args)
...
+                    }
+                }
+            }
+            if ( encodingToUse == null )
+            {
+                encodingToUse = Encoding.GetEncoding(28591);
```
SteveL-MSFT: Use GetEncoding("iso-8859-1")

SteveL-MSFT (changes requested):

mklement0:
This PR is fundamentally misguided, as discussed [here](https://github.com/PowerShell/PowerShell/issues/4681#issuecomment-341298296).

SteveL-MSFT:
Based on the discussion @mklement0 linked, it seems that for console text output, we should default to UTF-8.  Binary pipeline is something we'll have to defer to later.

mklement0:
@SteveL-MSFT: That's a great summary - thank you.

src/System.Management.Automation/engine/NativeCommandProcessor.cs
JamesWTruher: fixed!

test/tools/TestExe/TestExe.cs
JamesWTruher: I like that much better, thanks
updated

test/tools/TestExe/TestExe.cs
JamesWTruher: no need
[char]233 is all that is needed to show the problem, and is readable from the command line if you run it.

src/System.Management.Automation/engine/Utils.cs
JamesWTruher: fixed

test/tools/TestExe/TestExe.cs
JamesWTruher: i've replaced this with something far simpler (see dongbo's suggestion above)

JamesWTruher:
closing in favor of a new PR which simply changes the output encoding to utf8nobom.

PR #5362, opened 2017-11-07
Don't write an error if file already unblocked

### LINKED ISSUE #5353: Unblock-File should not throw error if file is not blocked
When `unblock-file` can't find the alternate stream for a blocked file, it reports a technically correct, but useless error message.  For files that aren't even blocked (downloaded from untrusted internet zone), we should just quietly ignore the error and return success.

Steps to reproduce
------------------

```powershell
unblock-file $profile
```

Expected behavior
-----------------

```none
no error
```

Actual behavior
---------------

```none
unblock-file : The system cannot find the file specified
At line:1 char:1
+ unblock-file $profile
+ ~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : PermissionDenied: (C:\Users\slee\D...ell_profile.ps1:String) [Unblock-File], Win32Exception
+ FullyQualifiedErrorId : RemoveItemUnauthorizedAccessError,Microsoft.PowerShell.Commands.UnblockFileCommand
```

Environment data
----------------

```powershell
OS                             Microsoft Windows 10.0.17035
```

iSazonov:
Fix #5353

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
@@ -127,6 +127,10 @@ namespace Microsoft.PowerShell.Commands
                         {
                             WriteError(new ErrorRecord(accessException, "RemoveItemUnauthorizedAccessError", ErrorCategory.PermissionDenied, path));
                         }
+                        else
+                        {
+                            WriteVerbose("File stream 'Zone.Identifier': "+accessException.Message);
+                        }
                     }
 
                 }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
@@ -121,9 +121,9 @@ namespace Microsoft.PowerShell.Commands
                     }
                     catch (Win32Exception accessException)
                     {
-                        // -2147467259 - File not found.
-                        // If the block stream not found the 'path' file was not blocked and we successfully return.
-                        if (accessException.HResult != -2147467259)
+                        // NativeErrorCode=2 - File not found.
+                        // If the block stream not found the 'path' was not blocked and we successfully return.
+                        if (accessException.NativeErrorCode != 2)
                         {
                             WriteError(new ErrorRecord(accessException, "RemoveItemUnauthorizedAccessError", ErrorCategory.PermissionDenied, path));
                         }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
@@ -121,7 +121,12 @@ namespace Microsoft.PowerShell.Commands
                     }
                     catch (Win32Exception accessException)
                     {
-                        WriteError(new ErrorRecord(accessException, "RemoveItemUnauthorizedAccessError", ErrorCategory.PermissionDenied, path));
+                        // -2147467259 - File not found.
+                        // If the block stream not found the 'path' file was not blocked and we successfully return.
+                        if (accessException.HResult != -2147467259)
+                        {
+                            WriteError(new ErrorRecord(accessException, "RemoveItemUnauthorizedAccessError", ErrorCategory.PermissionDenied, path));
+                        }
                     }
 
                 }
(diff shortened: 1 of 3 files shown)


lzybkr (approved):
I see there are existing tests for "file does not exist" and they still pass, so this change looks good.

dantraMSFT (approved):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
```diff
@@ -121,7 +121,12 @@ protected override void ProcessRecord()
                     }
                     catch (Win32Exception accessException)
                     {
-                        WriteError(new ErrorRecord(accessException, "RemoveItemUnauthorizedAccessError", ErrorCategory.PermissionDenied, path));
+                        // NativeErrorCode=2 - File not found.
+                        // If the block stream not found the 'path' was not blocked and we successfully return.
+                        if (accessException.NativeErrorCode != 2)
```
SteveL-MSFT: @jpsnover suggests adding a verbose message here if the block stream is not found

iSazonov: Done. I tried to avoid adding a resource string

SteveL-MSFT: I think it would be better to add a resource string saying:

```none
File '{0}' did not contain a 'Zone.Identifier' file stream and was already not blocked.
```

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/UnblockFile.cs
iSazonov: I added a resouce string with {0} at last position - we should take into account long paths.

SteveL-MSFT (approved):

src/Microsoft.PowerShell.Commands.Utility/resources/UtilityCommonStrings.resx
```diff
@@ -174,4 +174,7 @@
   <data name="TypeNotSupported" xml:space="preserve">
     <value>'{0}' is not supported in this system.</value>
   </data>
+  <data name="NoZoneIdentifierFileStream" xml:space="preserve">
+    <value>"No 'Zone.Identifier' file stream. The file was already not blocked: {0}"</value>
```
rkeithhill: Should this be present tense e.g. "The file is already not blocked: ..."?  Maybe ask one of your documentation folks about the wording for this error message?  Sorry but error messages in PowerShell really need help and I'd rather we not contribute to the problem with poor error messages.  Do you have a "doc/grammar" person to review error messages?

markekraus: `The file is already unblocked: `

SteveL-MSFT: @rkeithhill unfortunately at this time, our previous doc person left Msft.  Doc changes aren't breaking, so we can continue to improve them.  I agree reading it again that it should probably be present tense: The file is already not blocked.

@markekraus perhaps I'm a bit pedantic on this one, but `unblocked` implies the file was once blocked which may not have been the case here.

SteveL-MSFT (changes requested):
@iSazonov can you make the requested string change?

src/Microsoft.PowerShell.Commands.Utility/resources/UtilityCommonStrings.resx
markekraus: @SteveL-MSFT You're right, and what Keith wrote is more accurate. but `already not blocked` sounds odd.. maybe `The file is not blocked`?

SteveL-MSFT: @markekraus agree, that sounds better: The file is not blocked.

iSazonov:
Yes, I agree.
Is it good:
"No 'Zone.Identifier' file stream. The file is not blocked: {0}"

SteveL-MSFT:
@iSazonov thinking about it, I think it may be better to exclude that implementation detail as it may confuse some users.  I think just:

```none
The file is not blocked: {0}
```

Is good.

iSazonov:
Done.

SteveL-MSFT (approved):

PR #5363, opened 2017-11-07
Add support LF newlines in Import-Csv

### LINKED ISSUE #3692: Import-Csv does not handle newlines containing just \r (Carriage Return)
Steps to reproduce
------------------
"one,two,three`rthree,three,three" | out-file test.csv 
import-csv test.csv

Expected behavior
-----------------
Expected result:
```
one   two   three
---   ---   -----
three three three
```

Actual behavior
---------------
import-csv : The member "three" is already present.

Environment data
----------------
Tested on windows 8.1 and windows X (10). 
```
```
Steps to fix
----------------
It seems that the ImportCsvHelper method located in my C:\WINDOWS\Microsoft.Net\assembly\GAC_MSIL\Microsoft.PowerShell.Commands.Utility\v4.0_3.0.0.0__31bf3856ad364e35\Microsoft.PowerShell.Commands.Utility.dll is the culprit. 

It looks like the code here:
```
		private bool IsNewLine(char ch)
		{
			bool result = false;
			if (ch == '\n')
			{
				result = true;
			}
			else if (ch == '\r' && this.PeekNextChar('\n'))
			{
				result = true;
			}
			return result;
		}
```

should include 
```
			else if (ch == '\r')
			{
				result = true;
			}
```

and then all will be happy. Thanks!

iSazonov:
Fix #3692.

## Fix description
After the fix Import-Csv support CR (\r), LF (\n), CRLF (\r\n) as line delimiters.

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
@@ -1485,22 +1485,19 @@ namespace Microsoft.PowerShell.Commands
                 }
                 else if (IsNewLine(ch))
                 {
-                    if (ch == '\r')
-                    {
-                        ReadChar();
-                    }
-
                     if (seenBeginQuote)
                     {
                         //newline inside quote are valid
                         current.Append(ch);
-                        if (ch == '\r')
+                        if (ch == '\r' && PeekNextChar('\n'))
                         {
+                            ReadChar();
                             current.Append('\n');
                         }
                     }
                     else
                     {
+                        SkipNewLine();
                         result.Add(current.ToString());
                         current.Remove(0, current.Length);
                         //New line outside quote is end of word and end of record
@@ -1524,19 +1521,17 @@ namespace Microsoft.PowerShell.Commands
         bool
         IsNewLine(char ch)
         {
-            bool newLine = false;
-            if (ch == '\n')
+            return ch == '\r' || ch == '\n';
+        }
+
+        private
+        void
+        SkipNewLine()
+        {
+            if (PeekNextChar('\n'))
             {
-                newLine = true;
+                ReadChar();
             }
-            else if (ch == '\r')
-            {
-                if (PeekNextChar('\n'))
-                {
-                    newLine = true;
-                }
-            }
-            return newLine;
         }
 
         /// <summary>
@@ -1577,10 +1572,7 @@ namespace Microsoft.PowerShell.Commands
                 else if (IsNewLine(ch))
                 {
                     endOfRecord = true;
-                    if (ch == '\r')
-                    {
-                        ReadChar();
-                    }
+                    SkipNewLine();
                     break;
                 }
                 else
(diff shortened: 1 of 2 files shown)


markekraus:
@iSazonov are there existing test cases for all three line ending scenarios?

iSazonov:
Yes, I added new test.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -1524,19 +1521,17 @@ private string
...
-            bool newLine = false;
-            if (ch == '\n')
-            {
-                newLine = true;
-            }
-            else if (ch == '\r')
+            return ch == '\r' || ch == '\n';
+        }
+
+        private
```
lzybkr: We
should
fix
this.

But thanks for being consistent. :)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -1577,10 +1572,7 @@ private string
                 else if (IsNewLine(ch))
                 {
                     endOfRecord = true;
-                    if (ch == '\r')
-                    {
-                        ReadChar();
-                    }
+                    SkipNewLine();
```
lzybkr: If there are two newlines, won't this skip the second?

A better pattern might be to change `IsNewLine` to return a string (`"\n"`, `"\r"`, or `"\r\n"`) or null, and if the newline must be appended, append the result. There should be no need for `SkipNewLine`.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
markekraus: Can we get a good project wide reformatting done right after release? So many times I'm just following the ugly formatting around the code and then have to point that out in reviews as to why my code looks weird or doesn't follow the conventions.

lzybkr: I ran the Roslyn code format tool on the code base before we went public - it didn't fix stuff like this, but we should run it again.

iSazonov: Maybe just after RTM take two _formatting_ weeks to push as many formatting PRs as we can starting with Roslyn code format tool and continue with manual formatting.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
iSazonov: I don't know about skipping empty lines. Formally [RFC4180](https://www.ietf.org/rfc/rfc4180.txt) don't allow empty lines. Although practically it may make sense.

I thought about the pattern ( `IsNewLine`  to return a string ) and decided to begin with `SkipNewLine()` as seems more readable.  Now I'm going to implement `IsNewLine(out newLine)`.

iSazonov: I test with Windows PowerShell - we keep its behaviour and ignore empty lines.

Microsoft Excel imports empty lines as empty rows. We should discuss this in new Issue if we want because it is a breaking change.

lzybkr (approved):

PR #5371, opened 2017-11-07
enable Update-List cmdlet

SteveL-MSFT:
Part of https://github.com/PowerShell/PowerShell/issues/2123


test/powershell/Modules/Microsoft.PowerShell.Utility/Update-List.Tests.ps1
```diff
@@ -0,0 +1,29 @@
+Describe "Update-List tests" -Tags CI {
+    BeforeEach {
+        $list = [System.Collections.Generic.List``1[int]]::new()
```
lzybkr: You don't need ` ``1` - I'm guessing you copied this from an old test that was written before we had better generics support.

adityapatwardhan (approved):
LGTM

lzybkr (approved):
Does anybody still use this cmdlet? I'd be happy to see my one of my first contributions to PowerShell disappear. :)

test/powershell/Modules/Microsoft.PowerShell.Utility/Update-List.Tests.ps1
SteveL-MSFT: Yup, will change.

SteveL-MSFT:
@lzybkr that's a good question about whether we even want this cmdlet.  I've never used it until I wrote tests for it.  @ffeldhaus opened the original issue and had this in the list.  Personally, I'm fine either way.

iSazonov:
@SteveL-MSFT Could you please review WriteError()'s in `update-list.cs` and add tests?

iSazonov:
I found 45 results in GitHub https://github.com/search?q=%22Update-List%22+language%3APowerShell&type=Code&utf8=%E2%9C%93 - somebody still use this cmdlet!!!

SteveL-MSFT:
@iSazonov you have to exclude the .psd1 (which is just copies of Utility.psd1) and any results from PowerShell/* repos or forks of PowerShell/PowerShell.  So there's really less than 20 hits and many of them are for the same factorial script and a number are for a custom function called `Update-List`.  Seems like good justification to just remove this code.

iSazonov:
I agree. I can't think of PowerShell magic scenarios that are impossible without this cmdlet.

SteveL-MSFT:
@PowerShell/powershell-committee agreed that this is not likely a highly used cmdlet and we can defer bringing it back until we hear customer feedback leaving the disabled code where it is for now

iSazonov:
We could reduce size by .Net IL Lynker #5881 - now it is in .Net Core 3.0.

Issue #5367, opened 2017-11-07
Start-PSBootStrap should check and install PSCore6 >= Beta.9


SteveL-MSFT:
The tooling for PSCore6 now expects pwsh(.exe).  Start-PSBootStrap should make a check that pwsh is available and if not, install it onto the system.

SteveL-MSFT:
This should depend on https://github.com/PowerShell/PowerShell/issues/5348

daxian-dbw:
Fixed via #5383

PR #5383, opened 2017-11-08
Add 'install-powershell.ps1' to install powershell core on windows

### LINKED ISSUE #5348: Need install-powershell.ps1 for Windows similar to install-powershell.sh
- Need install-powershell.ps1 for Windows similar to [install-powershell.sh](
https://github.com/PowerShell/PowerShell/blob/f3dd2f041433c274d200720219fe5a80d39aeba3/tools/install-powershell.sh)
- Also provide an option for installing the daily build of PowerShell. This is required for other GitHub repositories like PowerShellGet, OneGet, etc.,

daxian-dbw:
Fix #5348 #5367

Add `install-powershell.ps1` to install powershell core packages on windows.
Update `Start-PSBootStrap` to check and install latest PSCore package.

### DIFF
--- a/tools/install-powershell.ps1
+++ b/tools/install-powershell.ps1
@@ -1,0 +1,107 @@
+<#
+.Synopsis
+    Install PowerShell Core on Windows.
+    The release PowerShell Core package will be installed by default.
+.Parameter Destination
+    The destination path to install PowerShell Core to.
+.Parameter Daily
+    Install PowerShell Core from the daily build.
+.Parameter NotOverwrite
+    Do not overwrite the destination folder if it already exists.
+.Parameter AddToPath
+    Add the absolute destination path to the 'User' scope environment variable 'Path'
+#>
+[CmdletBinding()]
+param(
+    [Parameter()]
+    [ValidateNotNullOrEmpty()]
+    [string] $Destination,
+
+    [Parameter()]
+    [switch] $Daily,
+
+    [Parameter()]
+    [switch] $NotOverwrite,
+
+    [Parameter()]
+    [switch] $AddToPath
+)
+
+Set-StrictMode -Version latest
+$ErrorActionPreference = "Stop"
+
+if (-not $Destination) {
+    $Destination = "~/pscore"
+    if ($Daily) {
+        $Destination = "${Destination}-daily"
+    }
+}
+
+if (Test-Path -Path $Destination) {
+    if ($NotOverwrite) {
+        throw "Destination folder '$Destination' already exist. Use a different path or omit '-NotOverwrite' to overwrite."
+    }
+    Remove-Item -Path $Destination -Recurse -Force
+}
+New-Item -ItemType Directory -Path $Destination -Force > $null
+$Destination = Resolve-Path -Path $Destination | ForEach-Object -MemberName Path
+Write-Verbose "Destination: $Destination" -Verbose
+
+$architecture = Get-CimInstance win32_operatingsystem | ForEach-Object {
+                    switch ($_.OSArchitecture) {
+                        "64-bit" { "x64" }
+                        "32-bit" { "x86" }
+                        default  { throw "Daily package for OS architecture '$_' is not supported." }
+                    }
+                }
+$tempDir = Join-Path ([System.IO.Path]::GetTempPath()) ([System.IO.Path]::GetRandomFileName())
+New-Item -ItemType Directory -Path $tempDir -Force > $null
+try {
+    if ($Daily) {
+        if ($architecture -ne "x64") {
+            throw "The OS architecture is '$architecture'. However, we currently only support daily package for x64 Windows."
+        }
+
+        ## Register source if not yet
+        if (-not (Get-PackageSource -Name powershell-core-daily -ErrorAction SilentlyContinue)) {
+            $packageSource = "https://powershell.myget.org/F/powershell-core-daily"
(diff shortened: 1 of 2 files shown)


daxian-dbw:
The Travis CI was successful too, but it's not refreshed here. Please click "Details" to see the builds.

tools/install-powershell.ps1
```diff
@@ -0,0 +1,114 @@
...
+
+    [Parameter()]
+    [switch] $AddToPath
+)
+
+Set-StrictMode -Version latest
+$ErrorActionPreference = "Stop"
+
+if (-not $Destination) {
+    $Destination = "~/pscore"
```
SteveL-MSFT: Should we default to `~/.pscore` following `.dotnet` convention?

tools/install-powershell.ps1
```diff
@@ -0,0 +1,114 @@
...
+    if ($NotOverwrite) {
+        throw "Destination folder '$Destination' already exist. Use a different path or omit '-NotOverwrite' to overwrite."
+    }
+    Remove-Item -Path $Destination -Recurse -Force
+}
+New-Item -ItemType Directory -Path $Destination -Force > $null
+$Destination = Resolve-Path -Path $Destination | ForEach-Object -MemberName Path
+Write-Verbose "Destination: $Destination" -Verbose
+
+$architecture = Get-CimInstance win32_operatingsystem | ForEach-Object {
```
SteveL-MSFT: This seems overkill to check architecture, I would simply use:

```powershell
switch ($env:PROCESSOR_ARCHITECTURE) {
   ...
}

File: tools/install-powershell.ps1 [hunk 3]
```diff
@@ -0,0 +1,114 @@
...
+        $downloadURL = "https://github.com/PowerShell/PowerShell/releases/download/v${release}/${packageName}"
+        Write-Verbose "About to download package from '$downloadURL'" -Verbose
+
+        $packagePath = Join-Path -Path $tempDir -ChildPath $packageName
+        Invoke-WebRequest -Uri $downloadURL -OutFile $packagePath
+
+        Expand-Archive -Path $packagePath -DestinationPath $Destination
+    }
+
+    if ($AddToPath -and (-not $env:Path.Contains($Destination))) {
```
SteveL-MSFT: Is there a case where the user wouldn't addtopath?  Seems like this would be the primary use case?  Perhaps have `-DoNotAddToPath` instead?

tools/install-powershell.ps1
```diff
@@ -0,0 +1,114 @@
...
+        ## Add to the User scope 'Path' environment variable
+        $userPath = [System.Environment]::GetEnvironmentVariable("Path", "User")
+        $userPath += [System.IO.Path]::PathSeparator + $Destination
```
SteveL-MSFT: Should this be in front?

SteveL-MSFT (changes requested):

tools/install-powershell.ps1
daxian-dbw: I don't have a preference here. But on windows `dotnet-install.ps1` actually installs dotnet in `$env:LOCALAPPDATA\Microsoft\dotnet\`, only the some helping files are placed in `~/.dotnet`.

tools/install-powershell.ps1
daxian-dbw: `$env:PROCESSOR_ARCHITECTURE` shows `x86` from a x86 windows powershell on a x64 machine. That's why I turned to `win32_operatingsystem`.

tools/install-powershell.ps1
daxian-dbw: I think when installing daily build package, a user probably don't want it to be added to Path by default.

tools/install-powershell.ps1
daxian-dbw: OK, will change to add in front.

tools/install-powershell.ps1
SteveL-MSFT: Perhaps we should use `$env:LOCALAPPDATA\Microsoft\powershell`

tools/install-powershell.ps1
SteveL-MSFT: I'm thinking the opposite where for daily self-hosting, I would want it in the path, although it would only need to be added once, so this isn't a big deal.  I'm fine leaving it the way it is.

tools/install-powershell.ps1
daxian-dbw: OK, will change to that path.

tools/install-powershell.ps1
SteveL-MSFT: It seems like someone deliberately running x86 powershell would expect to install x86 pscore.  Not a blocking issue, but seems unnecessary.

tools/install-powershell.ps1
daxian-dbw: OK. I will leave it unchanged.

tools/install-powershell.ps1
daxian-dbw: I will change back to `$env:PROCESSOR_ARCHITECTURE`. It would be rare for someone to run a x86 powershell on a x64 machine anyways. We can fix it If it turns out to be an issue.

SteveL-MSFT (approved):

daxian-dbw:
macOS CI was canceled several times. This PR only affects AppVeyor CI runs, and given that I will merge this PR.
After merging this PR, I will enhance the script to work on Linux/macOS as well.

PR #5387, opened 2017-11-08
add ability to produce tar.gz pkg for Raspbian

SteveL-MSFT:
Although Raspbian supports .deb pkgs, the creation of a deb pkg needs to happen on an actual R-Pi device.  Next best thing for now is to produce a .tar.gz pkg making it easier to deploy PowerShell on R-Pi.

Not sure if any other files need to be updated so we have a `preview` release of R-Pi support with 6.0.0-RC.

### DIFF
--- a/tools/packaging/packaging.psm1
+++ b/tools/packaging/packaging.psm1
@@ -20,7 +20,7 @@ function Start-PSPackage {
         [string]$Name = "powershell",
 
         # Ubuntu, CentOS, Fedora, macOS, and Windows packages are supported
-        [ValidateSet("deb", "osxpkg", "rpm", "msi", "zip", "AppImage", "nupkg", "tar")]
+        [ValidateSet("deb", "osxpkg", "rpm", "msi", "zip", "AppImage", "nupkg", "tar", "tar-arm")]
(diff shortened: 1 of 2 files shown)


tools/packaging/packaging.psm1
```diff
@@ -20,7 +20,7 @@ function Start-PSPackage {
         [string]$Name = "powershell",
 
         # Ubuntu, CentOS, Fedora, macOS, and Windows packages are supported
-        [ValidateSet("deb", "osxpkg", "rpm", "msi", "zip", "AppImage", "nupkg", "tar")]
+        [ValidateSet("deb", "osxpkg", "rpm", "msi", "zip", "AppImage", "nupkg", "tar", "tar-arm")]
```
daxian-dbw: There are comments about `deb-arm` right below the parameter block, please update or remove them if they are not applicable anymore.

SteveL-MSFT: Will remove as you can only build a deb package by running this cmdlet on the actual platform itself.

daxian-dbw (approved):

adityapatwardhan (approved):

PR #5402, opened 2017-11-10
Add error on Legacy Credential over non-HTTPS for Web Cmdlets

### LINKED ISSUE #5112: Web Cmdlets should warn when  legacy -Credential is sent over unencrypted connections
# Problem

In #5052 we are introducing the new `-Authentication` parameter which include a terminating error when any scheme other than `https://` is provided in the URI. The legacy `-Credential` usage currently does not offer any warnings or errors when the secrets are sent over an unencrypted connection. 

This issue is to track and discuss which method to go with.

# Possible Solutions

## Add a warning
This solution would add a warning message (via `WriteWarning()`) that the use could suppress with the `-AllowUnencryptedAuthentication` parameter. This would likely be a non-breaking change that would simply the user politely when they use the legacy `-Credential` and something other than `'https://`

## Add an Error
This is similar to the previous but instead return an error. This could be a terminating or non-terminating error, but either would be a breaking change. A common usage of the web cmdlets is to use `-ErrorAction Stop` in a try/catch and this would introduce new stops for previously working code if users were sending credentials over HTTP before

## Remove the legacy `-Credential` usage.
The new `-Authentication` usage has some duplication of functionality in that it does the same thing on its `Basic` option as the legacy `-Credential` usage. Legacy `-Credential` would only sent the Authorization header when the server present an Authorization request where the new method always sends the Authorization header (as many OAuth systems do not present auth realm).  This would require some discovery and clean up. I think this is the ultimate choice, but probably not a good candidate for 6.0.0 RTM.

markekraus:
closes #5112

* Adds an error when a user tries to use `-Credential` (legacy usage without `-Authentication`) or `-UseDefaultCredentials` over a non-HTTPS URI
* User can Bypass error with `-AllowUnencryptedAuthentication`
* `-UseDefaultCredentials` can only be reliably tested on Windows as support on other platforms depends on a kerberos infrastructure. 
* Adds `/Auth/` tests to WebListener for challenge authentication Basic, Negotiate, and NTLM

Incidentally, this increases test coverage for the web cmdlets as `-Credential` and `-UseDefaultCredentials` were not being tested.

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -1239,5 +1239,6 @@ v6.0.
 
 #region test/tools/WebListener/README.md Overrides
  - test/tools/WebListener/README.md
+NTLM
 ResponseHeaders
 #endregion
--- a/.spelling
+++ b/.spelling
@@ -1239,6 +1239,7 @@ v6.0.
 
 #region test/tools/WebListener/README.md Overrides
  - test/tools/WebListener/README.md
+Auth
 NTLM
 ResponseHeaders
 #endregion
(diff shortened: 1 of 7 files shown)


SteveL-MSFT:
@markekraus you added mock NTLM support in the weblistener but didn't use it?

markekraus:
@SteveL-MSFT whoops. NTLM tests added.

SteveL-MSFT (approved):
LGTM

SteveL-MSFT:
This breaking change was approved by @PowerShell/powershell-committee

Issue #5403, opened 2017-11-10
RCEdit should not occur during packaging


TravisEz13:
running RCEdit removes any signature.  This needs to be done before we package (only in the include Symbols zip)

SteveL-MSFT:
I can move it to `start-psbuild`

PR #5404, opened 2017-11-10
move rcedit step to build rather than packaging

SteveL-MSFT:
Fix https://github.com/PowerShell/PowerShell/issues/5403

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -565,10 +565,13 @@ Fix steps:
             throw "RCEdit is required to modify pwsh.exe resources, please run 'Start-PSBootStrap' to install"
         }
 
-        $ProductVersion = (git --git-dir="$PSScriptRoot/.git" describe) -Replace '^v'
+        if (!$ReleaseTag)
+        {
+            $ReleaseTag = Get-PSCommitId
+        }
 
         Start-NativeExecution { & "~/.rcedit/rcedit-x64.exe" "$($Options.Output)" --set-icon "$PSScriptRoot\assets\Powershell_black.ico" `
-            --set-file-version $ProductVersion --set-product-version $ProductVersion --set-version-string "ProductName" "PowerShell Core 6" `
+            --set-file-version $ReleaseTag --set-product-version $ReleaseTag --set-version-string "ProductName" "PowerShell Core 6" `
             --set-requested-execution-level "asInvoker" --set-version-string "LegalCopyright" "(C) Microsoft Corporation.  All Rights Reserved." } | Write-Verbose
     }
 
--- a/build.psm1
+++ b/build.psm1
@@ -558,6 +558,20 @@ Fix steps:
         }
     }
 
+    if ($Environment.IsWindows) {
+        ## need RCEdit to modify the binaries embedded resources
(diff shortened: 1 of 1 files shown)


build.psm1
```diff
@@ -558,6 +558,20 @@ Fix steps:
...
     }
 
+    if ($Environment.IsWindows) {
+        ## need RCEdit to modify the binaries embedded resources
+        if (-not (Test-Path "~/.rcedit/rcedit-x64.exe"))
+        {
+            throw "RCEdit is required to modify pwsh.exe resources, please run 'Start-PSBootStrap' to install"
+        }
+
+        $ProductVersion = (git --git-dir="$PSScriptRoot/.git" describe) -Replace '^v'
```
TravisEz13: Use release tag parameter to get version if it was passed it.  This is needed for release builds.

TravisEz13: Please use `Get-PSCommitId` instead of rewriting this code `git --git-dir="$PSScriptRoot/.git" describe`

TravisEz13 (changes requested):

build.psm1
```diff
@@ -565,10 +565,13 @@ Fix steps:
             throw "RCEdit is required to modify pwsh.exe resources, please run 'Start-PSBootStrap' to install"
         }
 
-        $ProductVersion = (git --git-dir="$PSScriptRoot/.git" describe) -Replace '^v'
+        if (!$ReleaseTag)
+        {
+            $ReleaseTag = Get-PSCommitId
```
TravisEz13: you still need to remove the v at the beginning.
Get-PSVersion will do that for you, but I don't have a function to do this for the releasetag.

TravisEz13 (changes requested):

build.psm1
```diff
@@ -558,6 +558,28 @@ Fix steps:
...
+        ## need RCEdit to modify the binaries embedded resources
+        if (-not (Test-Path "~/.rcedit/rcedit-x64.exe"))
+        {
+            throw "RCEdit is required to modify pwsh.exe resources, please run 'Start-PSBootStrap' to install"
+        }
+
+        $ReleaseVersion = ""
+        if ($ReleaseTag)
+        {
+            $ReleaseVersion = $ReleaseTag -replace '^v'
```
TravisEz13: $ReleaseTagToUse should already be this value

TravisEz13 (approved):

adityapatwardhan (approved):
LGTM

bergmeister:
Minor but worth pointing out: The bracing style is inconsistent.

SteveL-MSFT:
@bergmeister I think that's a fair point.  Will update for the code I added.

daxian-dbw (approved):

daxian-dbw:
The macOS CI was canceled by Travis-CI ...
Given that changes in this PR don't affect PowerShell on Linux/macOS, I will merge this PR.

daxian-dbw:
@TravisEz13 note that even the step is moved to build phase, it's the `publish\pwsh.exe` that gets modified, so you need to copy the `pwsh.exe` from the publish folder when doing the signing.

PR #5411, opened 2017-11-10
Enhance 'install-powershell.ps1' to work on Linux/MacOS

daxian-dbw:
Enhance 'install-powershell.ps1' to work on Linux/MacOS.

When `-AddToPath` is specified:
- On Windows, add the absolute destination path to the 'User' scope environment variable 'Path';
- On Linux, make the symlink '/usr/bin/pwsh' points to "$Destination/pwsh";
- On MacOS, make the symlink '/usr/local/bin/pwsh' points to "$Destination/pwsh".

### DIFF
--- a/tools/install-powershell.ps1
+++ b/tools/install-powershell.ps1
@@ -33,8 +33,17 @@ param(
 Set-StrictMode -Version latest
 $ErrorActionPreference = "Stop"
 
+$IsLinuxEnv = (Get-Variable -Name "IsLinux" -ErrorAction Ignore) -and $IsLinux
+$IsMacOSEnv = (Get-Variable -Name "IsMacOS" -ErrorAction Ignore) -and $IsMacOS
+$IsWinEnv   = !$IsLinuxEnv -and !$IsMacOSEnv
+
 if (-not $Destination) {
-    $Destination = "$env:LOCALAPPDATA\Microsoft\powershell"
+    $Destination = if ($IsWinEnv) {
+        "$env:LOCALAPPDATA\Microsoft\powershell"
+    } else {
+        "~/.powershell"
+    }
+
     if ($Daily) {
         $Destination = "${Destination}-daily"
     }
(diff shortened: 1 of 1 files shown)


TravisEz13 (approved):

tools/install-powershell.ps1
```diff
@@ -85,25 +107,75 @@ try {
...
-        ## Add to the 'Path' for the current process
-        $env:Path = $Destination + [System.IO.Path]::PathSeparator + $env:Path
-        Write-Verbose "'$Destination' is added to Path" -Verbose
+    if ($AddToPath) {
+        if ($IsWinEnv -and (-not $env:Path.Contains($Destination))) {
+            ## Add to the User scope 'Path' environment variable
+            $userPath = [System.Environment]::GetEnvironmentVariable("Path", "User")
+            $userPath = $Destination + [System.IO.Path]::PathSeparator + $userPath
+            [System.Environment]::SetEnvironmentVariable("Path", $userPath, "User")
+            Write-Verbose "'$Destination' is added to the Path" -Verbose
```
adityapatwardhan: I think, since we have CmdletBinding() we should not force -verbose

tools/install-powershell.ps1
```diff
@@ -85,25 +107,75 @@ try {
...
+
         $downloadURL = "https://github.com/PowerShell/PowerShell/releases/download/v${release}/${packageName}"
         Write-Verbose "About to download package from '$downloadURL'" -Verbose
 
         $packagePath = Join-Path -Path $tempDir -ChildPath $packageName
         Invoke-WebRequest -Uri $downloadURL -OutFile $packagePath
 
-        Expand-Archive -Path $packagePath -DestinationPath $Destination
+        if ($IsWinEnv) {
+            Expand-Archive -Path $packagePath -DestinationPath $Destination
```
adityapatwardhan: Should we delete the compressed package after expansion?

adityapatwardhan (approved):
Some non-blocking comments

tools/install-powershell.ps1
daxian-dbw: The compressed package will be deleted. The compressed package is downloaded to `$tempDir` which will be deleted in the finally block at the end of the script.

tools/install-powershell.ps1
daxian-dbw: These messages are actually intentionally written out so that a user roughly knows what step the script is doing.

daxian-dbw:
Linux job succeeded, but the macOS job was canceled again.
Currently `install-powershell.ps1` is only used in AppVeyor bootstrapping (install latest powershell core if not found), so it won't affect macOS CI run. Given that, I will merge this PR.

Issue #5409, opened 2017-11-10
install-powershell.sh support for amazon linux


DarwinJS:
Create installpsh-amazonlinux.sh to support installation on amazon linux.

Also address:

- incorrect repo urls in comments of each script
- ensure that tarball installs overwrite (reinstall) and upgrades

DarwinJS:
@daxian-dbw - I am also planning to do some cleanup of the install*.sh scripts.  

Is this bit of code already in the package based installs or should it be backported to the other installpsh scripts?  https://github.com/PowerShell/PowerShell/blob/master/tools/installpsh-suse.sh#L174-L179

Are there any improvements in https://github.com/PowerShell/PowerShell/blob/master/tools/installpsh-suse.sh that would be helpful to have in the other installpsh*.sh scripts?

DarwinJS:
@daxian-dbw - I should also ask will installpsh-suse.sh work for upgrading over an older version - did you test it for that?  I am thinking since tarball based installs can't use the package manager to update, it might be nice if this code could be re-run to update.

daxian-dbw:
> Is this bit of code already in the package based installs or should it be backported to the other installpsh*.sh scripts? https://github.com/PowerShell/PowerShell/blob/master/tools/installpsh-suse.sh#L174-L179

This bit of code will be done automatically when installing the DEB/RPM packages.

> Are there any improvements in https://github.com/PowerShell/PowerShell/blob/master/tools/installpsh-suse.sh that would be helpful to have in the other installpsh*.sh scripts

No additional improvement. Just does what the package is doing.

>  should also ask will installpsh-suse.sh work for upgrading over an older version - did you test it for that?

If a newer release is available, then running the script should install the newer package, but it doesn't remove the older powershell core files from `/opt/microsoft/powershell`.
If there is no newer release, then re-running the script might fail as it tries to un-compress the files to the existing folder at [line 167](https://github.com/PowerShell/PowerShell/blob/master/tools/installpsh-suse.sh#L167). I didn't test the re-run scenario.

daxian-dbw:
Close via #5461

PR #5420, opened 2017-11-11
issue #5409 support for amazonlinux, tarball reinstall / upgrade and clean up

DarwinJS:
as discussed in 5409:

- support for amazonlinux
- ensure both tarball installs will reinstall / upgrade
- clean up comments with incorrect urls

### DIFF
--- a/tools/installpsh-amazonlinux.sh
+++ b/tools/installpsh-amazonlinux.sh
@@ -2,8 +2,8 @@
 
 #Companion code for the blog https://cloudywindows.com
 #call this code direction from the web with:
-#bash <(wget -O - https://raw.githubusercontent.com/PowerShell/PowerShell/master/tools/installpsh-redhat.sh) ARGUMENTS
-#bash <(curl -s https://raw.githubusercontent.com/PowerShell/PowerShell/master/tools/installpsh-redhat.sh) <ARGUMENTS>
+#bash <(wget -O - https://raw.githubusercontent.com/PowerShell/PowerShell/master/tools/installpsh-amazonlinux.sh) ARGUMENTS
+#bash <(curl -s https://raw.githubusercontent.com/PowerShell/PowerShell/master/tools/installpsh-amazonlinux.sh) <ARGUMENTS>
 
 #Usage - if you do not have the ability to run scripts directly from the web, 
 #        pull all files in this repo folder and execute, this script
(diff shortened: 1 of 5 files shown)


tools/installpsh-amazonlinux.sh
```diff
@@ -0,0 +1,221 @@
...
+
+if [[ "'$*'" =~ -interactivetesting ]] ; then
+    echo
+    echo "Amazon Linux does not have a desktop manager to support vscode, ignoring -includeide"
+fi
+
+if [[ "$repobased" == true ]] ; then
+  echo "*** NOTE: Run your regular package manager update cycle to update PowerShell Core"
+else
+  echo "*** NOTE: Re-run this script to update PowerShell Core"
```
daxian-dbw: This line should be added to `installpsh-suse.sh` too.

tools/installpsh-amazonlinux.sh
```diff
@@ -0,0 +1,221 @@
...
+echo "Source URL: $downloadurl"
+
+curl -L -o "$package" "$downloadurl"
+
+if [[ ! -r "$package" ]]; then
+    echo "ERROR: $package failed to download! Aborting..." >&2
+    exit 1
+fi
+
+echo "Installing PowerShell to /opt/microsoft/powershell/$release in overwrite mode"
```
daxian-dbw: Maybe add this line to `installpsh-suse.sh` too?

tools/installpsh-amazonlinux.sh
```diff
@@ -0,0 +1,221 @@
...
+        if [ -f /etc/redhat-release ] ; then
+            DistroBasedOn='redhat'
+            DIST=`cat /etc/redhat-release |sed s/\ release.*//`
+            PSUEDONAME=`cat /etc/redhat-release | sed s/.*\(// | sed s/\)//`
+            REV=`cat /etc/redhat-release | sed s/.*release\ // | sed s/\ .*//`
+        elif [ -f /etc/system-release ] ; then
+            DIST=`cat /etc/system-release |sed s/\ release.*//`
+            PSUEDONAME=`cat /etc/system-release | sed s/.*\(// | sed s/\)//`
+            REV=`cat /etc/system-release | sed s/.*release\ // | sed s/\ .*//`
+            if [[ $DIST == *"Amazon Linux"* ]] ; then
```
daxian-dbw: The code to check current OS and `DistroBasedOn` is different from other `installsh-` scripts targeting other Linux distros. Are you going to unify them all?

DarwinJS: Although it could be done, it's not absolutely necessary so I tacked toward minimal changes to anything I wasn't going to unit test before doing the PR.  install-powershell.sh needs it to correctly call installpsh-amazonlinux.sh and then installpsh-amazonlinux.sh needs it to validate the correct installer.  None of the others need the amazonlinux piece - but I can also understand keeping this bit of code the same.  I don't currently have the bandwidth to test if I change the others.

Let me know what you think we should do with this and then I'll do one update and include the other two items you mentioned in this PR.

daxian-dbw: Thanks for the clarification. I'm fine with the current change.
It would be more consistent if this piece of code can be unified, but that can be done in separate PRs.

tools/installpsh-amazonlinux.sh
daxian-dbw: For this comment and the one below, they are about adding the `ehco message` to the same place in `installpsh-suse.sh`. Making this change won't need additional tests as it's just adding a log message. Could you please do it?

DarwinJS: This change is made, squashed and pushed - but this PR is now showing as "closed" ?

daxian-dbw: @DarwinJS The timeline of this PR shows that you closed it ... But I see you already opened a new PR.

PR #5429, opened 2017-11-13
Enable install-powershell.ps1 to work from powershell-daily folder

SteveL-MSFT:
To make it easier to selfhost daily:

1. removed [validate] attribute from parameter so this script can be invoked directly from web.
  This doesn't have any negative impact as the $Destination parameter will have default value if null or empty.
2. moved removal of destination folder later as installing the package requires packagemanagement module and archive module (on Windows)
3. on Windows, because files have open handles when run from existing powershell-daily install, I rename the existing files and copy over the new ones.  User needs to exit and restart pwsh to take effect (similar to macOS/Linux where you have to exit and restart anyways)

### DIFF
--- a/tools/install-powershell.ps1
+++ b/tools/install-powershell.ps1
@@ -49,6 +49,7 @@ if (-not $Destination) {
         $Destination = "${Destination}-daily"
     }
 }
+$Destination = $PSCmdlet.SessionState.Path.GetUnresolvedProviderPathFromPSPath($Destination)
 Write-Verbose "Destination: $Destination" -Verbose
 
 Function Remove-Destination([string] $Destination) {
@@ -70,6 +71,7 @@ Function Remove-Destination([string] $Destination) {
                 }
             }
         } else {
+            # Unix systems don't keep open file handles so you can just move files/folders even if in use
             Move-Item "$Destination" "$Destination.old"
         }
     }
@@ -115,15 +117,7 @@ try {
         Write-Verbose "Daily package found. Name: $packageName; Version: $($package.Version)" -Verbose
 
         Install-Package -InputObject $package -Destination $tempDir -ExcludeVersion > $null
-        Remove-Destination $Destination
-
         $contentPath = [System.IO.Path]::Combine($tempDir, $packageName, "content")
-        if (Test-Path $Destination) {
-            Write-Verbose "Copying files" -Verbose
-            Copy-Item -Recurse -Path "$contentPath\*" -Destination $Destination -ErrorAction Ignore
-        } else {
-            Move-Item -Path $contentPath -Destination $Destination
-        }
     } else {
         $metadata = Invoke-RestMethod https://api.github.com/repos/powershell/powershell/releases/latest
         $release = $metadata.tag_name -replace '^v'
@@ -141,16 +135,24 @@ try {
 
         $packagePath = Join-Path -Path $tempDir -ChildPath $packageName
         Invoke-WebRequest -Uri $downloadURL -OutFile $packagePath
+        $contentPath = Join-Path -Path $tempDir -ChildPath "new"
 
-        New-Item -ItemType Directory -Path "$Destination.new" > $null
+        New-Item -ItemType Directory -Path $contentPath > $null
         if ($IsWinEnv) {
-            Expand-Archive -Path $packagePath -DestinationPath "$Destination.new"
+            Expand-Archive -Path $packagePath -DestinationPath $contentPath
         } else {
-            tar zxf $packagePath -C "$Destination.new"
+            tar zxf $packagePath -C $contentPath
         }
-
-        Remove-Destination $Destination
-        Move-Item -Path "$Destination.new" -Destination $Destination
+    }
+    Remove-Destination $Destination
+    if (Test-Path $Destination) {
+        Write-Verbose "Copying files" -Verbose
+        # only copy files as folders will already exist at $Destination
+        Get-ChildItem -Recurse -Path "$contentPath\*" -File | ForEach-Object {
+            Copy-Item $_ -Destination $Destination
+        }
+    } else {
+        Move-Item -Path $contentPath -Destination $Destination
     }
 
     ## Change the mode of 'pwsh' to 'rwxr-xr-x' to allow execution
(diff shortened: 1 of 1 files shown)


tools/install-powershell.ps1
```diff
@@ -121,15 +148,19 @@ try {
         $packagePath = Join-Path -Path $tempDir -ChildPath $packageName
         Invoke-WebRequest -Uri $downloadURL -OutFile $packagePath
 
+        New-Item -ItemType Directory -Path "$Destination.new" > $null
```
daxian-dbw: Please use the `$tempDir` for anything that is temporary, it's cleaner and also prevent the corner case taht `$Destination.new` already exists.
For example
```
$packageContent = Join-Path -Path $tempDir -ChildPath content
New-Item -ItemType Directory -Path $packageContent > $null
...
if ($IsWinEnv) {
    Expand-Archive ... -DestinationPath $packageContent
} else {
    tar zxf  ... -C $packageContent
}

File: tools/install-powershell.ps1 [hunk 2]
```diff
@@ -121,15 +148,19 @@ try {
...
         if ($IsWinEnv) {
-            Expand-Archive -Path $packagePath -DestinationPath $Destination
+            Expand-Archive -Path $packagePath -DestinationPath "$Destination.new"
         } else {
-            tar zxf $packagePath -C $Destination
+            tar zxf $packagePath -C "$Destination.new"
         }
+
+        Remove-Destination $Destination
+        Move-Item -Path "$Destination.new" -Destination $Destination
```
daxian-dbw: If you can use `Move-Item` here, I think you can do the same in the `daily` code path without having to differentiate between `$Destination` exists or not (line 127).

daxian-dbw (commented):
Still reviewing.

tools/install-powershell.ps1
```diff
@@ -50,16 +55,31 @@ if (-not $Destination) {
         $Destination = "${Destination}-daily"
     }
 }
+Write-Verbose "Destination: $Destination" -Verbose
```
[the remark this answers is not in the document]
SteveL-MSFT: Will update

tools/install-powershell.ps1
SteveL-MSFT: Will update

tools/install-powershell.ps1
SteveL-MSFT: The differentiation is needed when `$Destination` is in use (because `install-powershell` was run from `$Destination`) so I can't use the `Move-Item` optimization.

daxian-dbw: When installing a release package here, the `Destination` could also be `$PSHOME`, right?

SteveL-MSFT: Yes, that's true.  Will fix.

daxian-dbw: I just tried `install-powershell.ps1 -Destination $PSHOME` on Windows, and new files are not moved to $PSHOME after the run.

tools/install-powershell.ps1
```diff
@@ -99,10 +113,17 @@ try {
...
         Write-Verbose "Daily package found. Name: $packageName; Version: $($package.Version)" -Verbose
+
         Install-Package -InputObject $package -Destination $tempDir -ExcludeVersion > $null
+        Remove-Destination $Destination
 
         $contentPath = [System.IO.Path]::Combine($tempDir, $packageName, "content")
-        Copy-Item -Path $contentPath\* -Destination $Destination -Recurse -Force
+        if (Test-Path $Destination) {
+            Write-Verbose "Copying files" -Verbose
+            Copy-Item -Recurse -Path "$contentPath\*" -Destination $Destination -ErrorAction Ignore
```
daxian-dbw: We shouldn't specify `-ErrorAction Ignore`, if it fails to copy a file then the installation fails. And maybe add `-Force`.

tools/install-powershell.ps1
```diff
@@ -99,10 +113,17 @@ try {
...
         $package = Find-Package -Source powershell-core-daily -AllowPrereleaseVersions -Name $packageName
         Write-Verbose "Daily package found. Name: $packageName; Version: $($package.Version)" -Verbose
+
         Install-Package -InputObject $package -Destination $tempDir -ExcludeVersion > $null
+        Remove-Destination $Destination
 
         $contentPath = [System.IO.Path]::Combine($tempDir, $packageName, "content")
-        Copy-Item -Path $contentPath\* -Destination $Destination -Recurse -Force
+        if (Test-Path $Destination) {
+            Write-Verbose "Copying files" -Verbose
```
daxian-dbw: I think this log message should be moved before `if (Test-Path $Destination)` since both `copy-item` and `move-item` would take time. Maybe change the message to `Deploying files ...`

tools/install-powershell.ps1
```diff
@@ -50,16 +49,31 @@ if (-not $Destination) {
...
+Function Remove-Destination([string] $Destination) {
+    if (Test-Path -Path $Destination) {
+        if ($DoNotOverwrite) {
+            throw "Destination folder '$Destination' already exist. Use a different path or omit '-DoNotOverwrite' to overwrite."
+        }
+        Write-Verbose "Removing old installation: $Destination" -Verbose
+        if (Test-Path -Path "$Destination.old") {
+            Remove-Item "$Destination.old" -Recurse -Force
+        }
+        if ($IsWinEnv -and ($Destination -eq $PSHome)) {
```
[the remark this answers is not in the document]
SteveL-MSFT: Sure

tools/install-powershell.ps1
SteveL-MSFT: Move-Item is just updating a single directory pointer so shouldn't take any time

tools/install-powershell.ps1
SteveL-MSFT: The failure is due to it copying folders, I can change this to filter to just files and letting it fail

TravisEz13:
restarted macOS ci

daxian-dbw (approved):

TravisEz13 (approved):

SteveL-MSFT:
Still testing, don't merge yet

SteveL-MSFT:
Working for me.  I think it's ready to go.

PR #5432, opened 2017-11-13
Skip null-element check for collections with a value-type element type

### LINKED ISSUE #5417: Slow function when parameter is Mandatory its Type is specified
Steps to reproduce
------------------

```powershell
[Byte[]]$Value = [System.IO.File]::ReadAllBytes('C:\Some10MBfile')

#  ~1000ms for 10MB array
#  ~0.5ms for 1byte array
function MandType {
    Param (
        [parameter(Mandatory=$True)]
        [Byte[]]$Value
    )
}

#  ~0.3ms for 10MB array
function TypeOnly {
    Param (
        [parameter()]
        [Byte[]]$Value
    )
}

#  ~0.3 for 10MB array
function MandOnly {
    Param (
        [parameter(Mandatory=$True)]
        $Value
    )
}

(Measure-Command { MandType $Value}).TotalMilliseconds
(Measure-Command { MandType ([Byte[]]@(0))}).TotalMilliseconds
(Measure-Command { TypeOnly $Value}).TotalMilliseconds
(Measure-Command { MandOnly $Value}).TotalMilliseconds

```

Expected behavior
-----------------
All versions of the above functions should behave identically and complete immediately, regardless of of parameter size as the functions perform no functions.
```none
1.1
1.1
1.1
1.1
```

Actual behavior
---------------
When setting a parameter to Mandatory, specifying its Type, and passing a large array in, the function will pause for some amount of time before starting.
```none
1000
1.1
1.1
1.1
```

Environment data
----------------

```powershell
PS C:\> $PSVersionTable

PS C:\> $PSVersionTable

PSVersion                      5.1.16299.19
BuildVersion                   10.0.16299.19
```

daxian-dbw:
Fix #5417

- For `Mandatory` parameter, `ValidateNotNull` and `ValidateNotNullOrEmpty` attributes, skip null-element check if the collection's element type is value type.
- For `ValidateNotNull` and `ValidateNotNullOrEmpty`, they are changed to do null/empty check only on collections (as defined in `ParameterCollectionTypeInformation`) and hashtable. They currently enumerate elements of any object that implements `IEnumerable` and also enumerate elements if the argument is an enumerator. I think this behavior is not right for the following reasons:
    - The argument may be an object that talks to Azure to retrieve elements back. It doesn't seem right to have the `ValidateNotNull/ValidateNotNullOrEmpty` attributes wait on that.
    - On Windows, COM object may be enumerable, but the `GetEnumerator()` method is not supported in .NET Core on COM object. (See the COM example below)
    - When the argument is an enumerator, the enumerator may not be useful after the validate attribute walks through all its elements. (See the enumerator example below)

COM Example
```
function bar {
   param ([ValidateNotNull()] $object)
}
> $shell = New-Object -ComObject Shell.Application
> $folder = $shell.NameSpace("F:\tmp\dd\")
> bar -object ($folder.Items())
bar : Cannot validate argument on parameter 'object'. Could not load type 'System.Runtime.InteropServices.ComTypes.IEnumerable' from assembly 'System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e'.
At line:1 char:13
+ bar -object ($folder.Items())
+             ~~~~~~~~~~~~~~~~~
+ CategoryInfo          : InvalidData: (:) [bar], ParameterBindingValidationException
+ FullyQualifiedErrorId : ParameterArgumentValidationError,bar
```

Enumerator Example
```
function bar {
   param ([ValidateNotNull()] $object)
   "MoveNext:  $($object.MoveNext())"
}

> bar -object @(1,2,3).GetEnumerator()
MoveNext:  False
```

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -568,7 +568,7 @@ Fix steps:
         if ($ReleaseTagToUse) {
             $ReleaseVersion = $ReleaseTagToUse
         } else {
-            $ReleaseVersion = (Get-PSCommitId -WarningAction SilentlyContinue) -replace '^v'
+            $ReleaseVersion = (Get-PSCommitId) -replace '^v'
         }
 
         Start-NativeExecution { & "~/.rcedit/rcedit-x64.exe" "$($Options.Output)" --set-icon "$PSScriptRoot\assets\Powershell_black.ico" `
--- a/build.psm1
+++ b/build.psm1
@@ -568,7 +568,7 @@ Fix steps:
         if ($ReleaseTagToUse) {
             $ReleaseVersion = $ReleaseTagToUse
         } else {
-            $ReleaseVersion = (Get-PSCommitId) -replace '^v'
+            $ReleaseVersion = (Get-PSCommitId -WarningAction SilentlyContinue) -replace '^v'
         }
 
         Start-NativeExecution { & "~/.rcedit/rcedit-x64.exe" "$($Options.Output)" --set-icon "$PSScriptRoot\assets\Powershell_black.ico" `
--- a/src/System.Management.Automation/engine/Attributes.cs
+++ b/src/System.Management.Automation/engine/Attributes.cs
@@ -1994,32 +1994,33 @@ namespace System.Management.Automation
             {
                 bool isEmpty = true;
                 IEnumerator ienum = LanguagePrimitives.GetEnumerator(arguments);
-                while (ienum.MoveNext())
+                if (ienum.MoveNext()) { isEmpty = false; }
+
+                // If the element of the collection is of value type, then no need to check for null
+                // because a value-type value cannot be null.
+                if (!isEmpty && !isElementValueType)
                 {
-                    isEmpty = false;
-                    // If the element of the collection is of value type, then no need to check for null
-                    // because a value-type value cannot be null.
-                    if (isElementValueType) { break; }
-
-                    object element = ienum.Current;
-                    if (element == null || element == AutomationNull.Value)
-                    {
-                        throw new ValidationMetadataException(
-                            "ArgumentIsNull",
-                            null,
-                            Metadata.ValidateNotNullOrEmptyCollectionFailure);
-                    }
-
-                    if (element is string elementAsString)
-                    {
-                        if (String.IsNullOrEmpty(elementAsString))
+                    do {
+                        object element = ienum.Current;
+                        if (element == null || element == AutomationNull.Value)
                         {
                             throw new ValidationMetadataException(
-                                "ArgumentCollectionContainsEmpty",
+                                "ArgumentIsNull",
                                 null,
                                 Metadata.ValidateNotNullOrEmptyCollectionFailure);
                         }
-                    }
+
+                        if (element is string elementAsString)
+                        {
+                            if (String.IsNullOrEmpty(elementAsString))
+                            {
+                                throw new ValidationMetadataException(
+                                    "ArgumentCollectionContainsEmpty",
+                                    null,
+                                    Metadata.ValidateNotNullOrEmptyCollectionFailure);
+                            }
+                        }
+                    } while (ienum.MoveNext());
                 }
 
                 if (isEmpty)
--- a/src/System.Management.Automation/engine/Attributes.cs
+++ b/src/System.Management.Automation/engine/Attributes.cs
@@ -1867,7 +1867,7 @@ namespace System.Management.Automation
     #region NULL validation attributes
 
     /// <summary>
-    /// Base type of Null Validation attributes
+    /// Base type of Null Validation attributes.
     /// </summary>
     public abstract class NullValidationAttributeBase : ValidateArgumentsAttribute
     {
--- a/src/System.Management.Automation/engine/Attributes.cs
+++ b/src/System.Management.Automation/engine/Attributes.cs
@@ -1865,11 +1865,40 @@ namespace System.Management.Automation
     #endregion
 
     #region NULL validation attributes
+
+    /// <summary>
+    /// Base type of Null Validation attributes
+    /// </summary>
+    public abstract class NullValidationAttributeBase : ValidateArgumentsAttribute
+    {
+        /// <summary>
+        /// Check if the argument type is a collection.
+        /// </summary>
+        protected bool IsArgumentCollection(Type argumentType, out bool isElementValueType)
+        {
+            isElementValueType = false;
+            var information = new ParameterCollectionTypeInformation(argumentType);
+            switch (information.ParameterCollectionType)
+            {
+                // If 'arguments' is an array, or implement 'IList', or implement 'ICollection<>'
+                // then we continue to check each element of the collection.
+                case ParameterCollectionType.Array:
+                case ParameterCollectionType.IList:
+                case ParameterCollectionType.ICollectionGeneric:
+                    Type elementType = information.ElementType;
+                    isElementValueType = elementType != null && elementType.IsValueType;
+                    return true;
+                default:
+                    return false;
+            }
+        }
+    }
+
     /// <summary>
     /// Validates that the parameters's argument is not null
     /// </summary>
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
-    public sealed class ValidateNotNullAttribute : ValidateArgumentsAttribute
+    public sealed class ValidateNotNullAttribute : NullValidationAttributeBase
     {
         /// <summary>
         /// Verifies the argument is not null and if it is a collection, that each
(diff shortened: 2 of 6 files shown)


build.psm1
```diff
@@ -568,7 +568,7 @@ Fix steps:
         if ($ReleaseTagToUse) {
             $ReleaseVersion = $ReleaseTagToUse
         } else {
-            $ReleaseVersion = (Get-PSCommitId) -replace '^v'
+            $ReleaseVersion = (Get-PSCommitId -WarningAction SilentlyContinue) -replace '^v'
```
adityapatwardhan: This change does not seem related to this PR.

test/powershell/engine/Basic/ValidateAttributes.Tests.ps1
```diff
@@ -260,4 +260,124 @@ Describe 'Validate Attributes Tests' -Tags 'CI' {
...
+            $byteArray = [System.IO.File]::ReadAllBytes($filePath)
+            $byteList  = [System.Collections.Generic.List[byte]] $byteArray
+            $byteCollection = [System.Collections.ObjectModel.Collection[byte]] $byteArray
+            ## Use the running time of 'MandType -Value $byteArray' as the baseline time
+            $baseline = (Measure-Command { MandType -Value $byteArray }).Milliseconds
+            ## Running time should be less than 'expected'
+            $expected = $baseline + 20
+
+            if ($IsWindows) {
+                $null = New-Item -Path $TESTDRIVE/file1 -ItemType File
```
adityapatwardhan: -ItemType File is not necessary anymore. The default itemtype is file

test/powershell/engine/Basic/ValidateAttributes.Tests.ps1
```diff
@@ -260,4 +260,124 @@ Describe 'Validate Attributes Tests' -Tags 'CI' {
...
+                        $items -join ","
+                    }
+                }
+            }
+
+            function NotNullOrEmptyFunc {
+                param(
+                    [ValidateNotNullOrEmpty()]
+                    $Value,
+                    [string] $TestType
```
adityapatwardhan: Can you add a parameter like: `[System.Nullable``1[[System.Int32]]] $nullableInt`

adityapatwardhan (changes requested):

src/System.Management.Automation/engine/ParameterBinderBase.cs
```diff
@@ -787,12 +787,18 @@ internal CommandLineParameters CommandLineParameters
...
+            bool isElementValueType = elementType != null && elementType.IsValueType;
             // Note - we explicitly don't pass the context here because we don't want
             // the overhead of the calls that check for stopping.
             while (ParserOps.MoveNext(null, null, ienum))
             {
-                object element = ParserOps.Current(null, ienum);
                 isEmpty = false;
+                // If the element of the collection is of value type, then no need to check for null
+                // because a value-type value cannot be null.
+                if (isElementValueType) { break; }
```
iSazonov: Why if we calculate this before the cycle?

iSazonov (commented):
Leave a comment

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1975,12 +1990,18 @@ protected override void Validate(object arguments, EngineIntrinsics engineIntrin
                         Metadata.ValidateNotNullOrEmptyFailure);
                 }
             }
-            else if ((ienum = arguments as IEnumerable) != null)
+            else if (IsArgumentCollection(arguments.GetType(), out bool isElementValueType))
```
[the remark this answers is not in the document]
daxian-dbw: No, the check of element type doesn't involve inspecting the elements of the collection. You can see that the type of `arguments` is passed in instead of the `arguments` itself.
Please see the constructor of [`ParameterCollectionTypeInformation`](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/CompiledCommandParameter.cs#L613) for details.

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1975,12 +1990,18 @@ protected override void Validate(object arguments, EngineIntrinsics engineIntrin
...
-                foreach (object element in ienum)
+                bool isEmpty = true;
+                IEnumerator ienum = LanguagePrimitives.GetEnumerator(arguments);
+                while (ienum.MoveNext())
                 {
-                    validElements++;
+                    isEmpty = false;
+                    // If the element of the collection is of value type, then no need to check for null
+                    // because a value-type value cannot be null.
+                    if (isElementValueType) { break; }
```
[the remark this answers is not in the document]
daxian-dbw: We don't check the elements of `arguments` in `IsArgumentCollection`. The argument passed in is the type of `arguments`, not `arguments` itself.

src/System.Management.Automation/engine/ParameterBinderBase.cs
daxian-dbw: Same reason as above. The element type is resolved using the collection type.

build.psm1
daxian-dbw: Yes, I will revert it.

test/powershell/engine/Basic/ValidateAttributes.Tests.ps1
daxian-dbw: A `[System.Nullable[int]]` parameter is not useful in this case because the type is not a collection. Hence `ValidateNotNullOrEmpty` attribute won't validate it as a collection.

test/powershell/engine/Basic/ValidateAttributes.Tests.ps1
daxian-dbw: OK. I will remove it.

src/System.Management.Automation/engine/Attributes.cs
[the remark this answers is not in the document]
iSazonov: Yes, I see that no element check is in ParameterCollectionTypeInformation.

But the code is not clear. Maybe:
```c#
if (ienum.MoveNext()) isEmpty = false;

if (!isElementValueType)
{
    do
    {
        ...
        object element = ienum.Current; 
        ...
    } while(ienum.MoveNext()))
}
```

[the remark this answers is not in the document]
daxian-dbw: I think the logic is clear -- if the element type of the collection is a value type, then skip checking each element. The `isEmpty` check is needed in either case.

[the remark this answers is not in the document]
iSazonov: The logic is _right_ but the code is slightly tricky for me. I added your comments in my sample above. Also we do two actions that can be carried before the cycle - if we are thinking about performance it is important - I think Measure-Object show a delay starting with 10000 elements.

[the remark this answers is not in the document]
daxian-dbw: I get your point. I will update as you suggested. Thanks!

adityapatwardhan (approved):

daxian-dbw:
@iSazonov You comments have been addressed. Please take another look when you have time. Thanks!

test/powershell/engine/Basic/ValidateAttributes.Tests.ps1
```diff
@@ -260,4 +260,124 @@ Describe 'Validate Attributes Tests' -Tags 'CI' {
                 $ScriptBlock | Should Not Throw
         }
     }
+
+    Context "ValidateNotNull, ValidateNotNullOrEmpty and Not-Null-Or-Empty check for Mandatory parameter" {
+
+        BeforeAll {
+            function MandType {
```
iSazonov: Could you please use more nice name? MandatoryType?

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1865,11 +1865,40 @@ public ValidateUserDriveAttribute()
     #endregion
 
     #region NULL validation attributes
+
+    /// <summary>
+    /// Base type of Null Validation attributes
```
iSazonov: Could you please add final dot in _public_ comments?

iSazonov (approved):
LGTM with two minor comments.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and approves with this change

adityapatwardhan:
@daxian-dbw restarted AppVeyor build. Failed due to #5567

src/System.Management.Automation/engine/Attributes.cs
```diff
@@ -1891,20 +1920,23 @@ public sealed class ValidateNotNullAttribute : ValidateArgumentsAttribute
...
-            else if ((ienum = arguments as IEnumerable) != null)
+            else if (IsArgumentCollection(arguments.GetType(), out bool isElementValueType))
             {
-                foreach (object element in ienum)
+                // If the element of the collection is of value type, then no need to check for null
+                // because a value-type value cannot be null.
+                if (isElementValueType) { return; }
+
+                IEnumerator ienum = LanguagePrimitives.GetEnumerator(arguments);
+                while (ienum.MoveNext())
```
TravisEz13: Should we really check that every element is not null?  To me, the fact that you have at least one element should satisfy the condition?

TravisEz13 (approved):

Issue #5175, opened 2017-10-19
Parameters Should Support Mutual Exclusion Natively


markekraus:
# Problem

Parameter sets are great for grouping parameters into sets that work together. This can, to some degree, also be used to make parameters exclusive. The problem arises when you have mutually exclusive parameters that need to exist in all parameter sets. The current implementation would require creating additional parameter sets to accommodate the actual parameter sets and the mutually exclusive parameters.

For very simplistic demonstration, lets say you have a `New-Widget`:

```powershell
function New-Widget {
    [CmdletBinding()]
    param (
        [Parameter(ParameterSetName="Size")]
        [int]$Height,
        [Parameter(ParameterSetName="Size")]
        [int]$Width,
        [Parameter(ParameterSetName="Size")]
        [int]$Length,

        [Parameter(ParameterSetName="Color")]
        [int]$Red,
        [Parameter(ParameterSetName="Color")]
        [int]$Green,
        [Parameter(ParameterSetName="Color")]
        [int]$Blue
    )
}
``` 

Widgets can be defined either by color or size. Now, let say that all widgets, regardless of begin defined by size or color, can either have a flavor or a sound but not both. Currently, to accommodate this you would need to do the following:

```powershell
function New-Widget {
    [CmdletBinding()]
    param (
        [Parameter(ParameterSetName="SizeFlavor")]
        [Parameter(ParameterSetName="SizeSound")]
        [int]$Height,
        [Parameter(ParameterSetName="SizeFlavor")]
        [Parameter(ParameterSetName="SizeSound")]
        [int]$Width,
        [Parameter(ParameterSetName="SizeSound")]
        [Parameter(ParameterSetName="SizeFlavor")]
        [int]$Length,

        [Parameter(ParameterSetName="ColorFlavor")]
        [Parameter(ParameterSetName="ColorSound")]
        [int]$Red,
        [Parameter(ParameterSetName="ColorSound")]
        [Parameter(ParameterSetName="ColorFlavor")]
        [int]$Green,
        [Parameter(ParameterSetName="ColorSound")]
        [Parameter(ParameterSetName="ColorFlavor")]
        [int]$Blue,

        [Parameter(ParameterSetName="SizeFlavor")]
        [Parameter(ParameterSetName="ColorFlavor")]
        [string]$Flavor,

        [Parameter(ParameterSetName="SizeSound")]
        [Parameter(ParameterSetName="ColorSound")]
        [string]$Sound
    )
}
```

[...]

PR #5435, opened 2017-11-13
Added more helpful message for AmbiguousParameterSet exception

kvprasoon:
Added more helpful message for AmbiguousParameterSet exception which always throws exception "Parameter set cannot be resolved using the specified named parameters" which is not properly conveying the end user for further steps required.

### DIFF
--- a/src/System.Management.Automation/resources/ParameterBinderStrings.resx
+++ b/src/System.Management.Automation/resources/ParameterBinderStrings.resx
@@ -142,7 +142,7 @@
     <value>Cannot bind positional parameters because no names were given.</value>
   </data>
   <data name="AmbiguousParameterSet" xml:space="preserve">
-    <value>Parameter set cannot be resolved using the specified named parameters.</value>
+    <value>Parameter set cannot be resolved using the specified named parameters.One or few of the parameters mentioned cannot be used together.Do a Get-command {0} to see the available parameter sets.</value>
   </data>
   <data name="MissingMandatoryParameter" xml:space="preserve">
     <value>Cannot process command because of one or more missing mandatory parameters:{1}.</value>
--- a/src/System.Management.Automation/resources/ParameterBinderStrings.resx
+++ b/src/System.Management.Automation/resources/ParameterBinderStrings.resx
@@ -110,7 +110,7 @@
   </resheader>
   <resheader name="version">
     <value>2.0</value>
-   </resheader>
+  </resheader>
   <resheader name="reader">
     <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
@@ -120,7 +120,7 @@
   <data name="NamedParameterNotFound" xml:space="preserve">
     <value>A parameter cannot be found that matches parameter name '{1}'.</value>
   </data>
-   <data name="PositionalParameterNotFound" xml:space="preserve">
+  <data name="PositionalParameterNotFound" xml:space="preserve">
     <value>A positional parameter cannot be found that accepts argument '{1}'.</value>
   </data>
   <data name="MissingArgument" xml:space="preserve">
--- a/src/System.Management.Automation/resources/ParameterBinderStrings.resx
+++ b/src/System.Management.Automation/resources/ParameterBinderStrings.resx
@@ -142,7 +142,7 @@
     <value>Cannot bind positional parameters because no names were given.</value>
   </data>
   <data name="AmbiguousParameterSet" xml:space="preserve">
-    <value>Parameter set cannot be resolved using the specified named parameters. One or more of the parameters issued cannot be used together. Do a Get-Help on your script/function/cmdlet for the syntax help.</value>
+    <value>Parameter set cannot be resolved using the specified named parameters. One or more parameters issued cannot be used together or a parameter which will decide the parameter set is not provided. Use Get-Help with your script, function, or cmdlet for syntax help.</value>
   </data>
   <data name="MissingMandatoryParameter" xml:space="preserve">
     <value>Cannot process command because of one or more missing mandatory parameters:{1}.</value>
--- a/src/System.Management.Automation/resources/ParameterBinderStrings.resx
+++ b/src/System.Management.Automation/resources/ParameterBinderStrings.resx
@@ -142,7 +142,7 @@
     <value>Cannot bind positional parameters because no names were given.</value>
   </data>
   <data name="AmbiguousParameterSet" xml:space="preserve">
-    <value>Parameter set cannot be resolved using the specified named parameters. One or more of the parameters issued cannot be used together. Do a Get-Help on your script/function for the syntax help.</value>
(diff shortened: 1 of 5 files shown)


src/System.Management.Automation/resources/ParameterBinderStrings.resx
```diff
@@ -142,7 +142,7 @@
     <value>Cannot bind positional parameters because no names were given.</value>
   </data>
   <data name="AmbiguousParameterSet" xml:space="preserve">
-    <value>Parameter set cannot be resolved using the specified named parameters.</value>
+    <value>Parameter set cannot be resolved using the specified named parameters.One or few of the parameters mentioned cannot be used together.Do a Get-command {0} to see the available parameter sets.</value>
```
markekraus: * Add spaces between periods and the first word of a sentence
* `One or few` should be `One or more` but, this could also occur when an insufficient number of parameters have been supplied, not just when 2 parameters cannot be used together. 
* `parameters mentioned` should be `parameters issued`.
* `Do a Get-command {0}` do all instances of this error pass the current command name to this string for formatting? if not, this could be very confusing.
* `Get-Help` would be preferable to `Get-Command` as `Get-Command` does not show this information be default.
*  `parameter sets` is some what esoteric to function./cmdlet authors. To an admin user it might be better to say `For syntax help`.

markekraus (changes requested):
Thanks for contributing! Please make modifications to your pull request.

src/System.Management.Automation/resources/ParameterBinderStrings.resx
kvprasoon: Hi , Thanks for the review, I'll do the required modifications

kvprasoon:
I have updated it as per the comments

kvprasoon (commented):
Thanks for the review,I have updated it as per the comments

• Add spaces between periods and the first word of a sentence - Done
• One or few  should be  One or more  but, this could also occur when an insufficient number of parameters have been supplied, not just when 2 parameters cannot be used together. - Done
• parameters mentioned  should be  parameters issued . - Done
• Do a Get-command {0}  do all instances of this error pass the current command name to this string for formatting? if not, this could be very confusing. - Done, changed to a generic message without deending on the first element
• Get-Help  would be preferable to  Get-Command  as  Get-Command  does not show this information be default. - Done, changed to Get-Help
• parameter sets  is some what esoteric to function./cmdlet authors. To an admin user it might be better to say  For syntax help . - Done

kvprasoon (commented):
Thanks for the review,I have updated it as per the comments

• Add spaces between periods and the first word of a sentence - Done
• One or few  should be  One or more  but, this could also occur when an insufficient number of parameters have been supplied, not just when 2 parameters cannot be used together. - Done
• parameters mentioned  should be  parameters issued . - Done
• Do a Get-command {0}  do all instances of this error pass the current command name to this string for formatting? if not, this could be very confusing. - Done, changed to a generic message without deending on the first element
• Get-Help  would be preferable to  Get-Command  as  Get-Command  does not show this information be default. - Done, changed to Get-Help
• parameter sets  is some what esoteric to function./cmdlet authors. To an admin user it might be better to say  For syntax help . - Done

src/System.Management.Automation/resources/ParameterBinderStrings.resx
```diff
@@ -142,7 +142,7 @@
     <value>Cannot bind positional parameters because no names were given.</value>
   </data>
   <data name="AmbiguousParameterSet" xml:space="preserve">
-    <value>Parameter set cannot be resolved using the specified named parameters.</value>
+    <value>Parameter set cannot be resolved using the specified named parameters. One or more of the parameters issued cannot be used together. Do a Get-Help on your script/function/cmdlet for the syntax help.</value>
```
markekraus: * `One or more of the parameters issued cannot be used together.` This still doesn't address the case where too few parameters are supplied. For example: `foo` has 2 parameter sets `-A -B -C` and `-A -B -D` but the user only supplied `-A -B`. It's not that `-A` and `-B` can't be used together, but that `-C` or `-D` are needed to determine the parameter set.
* `Do a Get-Help on your script/function/cmdlet for the syntax help.`  I think this is would be somewhat confusing. I'm not sure how to fix it. The wording could at least be improved `Use Get-Help with your script, function, or cmdlet for syntax help.`

markekraus (changes requested):

kvprasoon (commented):
One or more of the parameters issued cannot be used together.  This still doesn't address the case where too few parameters are supplied. For example:  foo  has 2 parameter sets  -A -B -C  and  -A -B -D  but the user only supplied  -A -B . It's not that  -A  and  -B  can't be used together, but that  -C  or  -D  are needed to determine the parameter set. - Added the information about not using the decision making parameter
• Do a Get-Help on your script/function/cmdlet for the syntax help.  I think this is would be somewhat confusing. I'm not sure how to fix it. The wording could at least be improved  Use Get-Help with your script, function, or cmdlet for syntax help.  - changed to suggested message

kvprasoon (commented):
please review

kvprasoon (commented):
can anyone help me by reviewing this

iSazonov:
@kvprasoon I guess PowerShell team is fully obsessed with the release of the new version.
Also please rebase to get latest updates - I think this help to pass CIs.

/cc @SteveL-MSFT  Should we use direct tips in error messages or use HelpSystem capabilites?

src/System.Management.Automation/resources/ParameterBinderStrings.resx
```diff
@@ -142,7 +142,7 @@
     <value>Cannot bind positional parameters because no names were given.</value>
   </data>
   <data name="AmbiguousParameterSet" xml:space="preserve">
-    <value>Parameter set cannot be resolved using the specified named parameters.</value>
+    <value>Parameter set cannot be resolved using the specified named parameters. One or more parameters issued cannot be used together or a parameter which will decide the parameter set is not provided. Use Get-Help with your script, function, or cmdlet for syntax help.</value>
```
markekraus: > or a parameter which will decide the parameter set is not provided.

Perhaps this should be `or an insufficient number of parameters were provided.`  There could certainly be more than one missing.

SteveL-MSFT:
@iSazonov since you added `Type 'help' for help` as part of the startup banner, it seems unnecessary to point users to `get-help` in any error message.  PowerShell does have a suggestion system that we should try to take advantage of in the future to provide more concise help on how to resolve issues (like if they type a parameter name wrong, we can do some matching and suggest what they meant which is different from this PR).

For this particular issue, I think the ideal error message would tell the user which parameters are mutually exclusive and can't be used together or if another parameter is required to resolve the parameter set.  I added a note to https://github.com/PowerShell/PowerShell/issues/5175 which covers a new proposal on defining mutually exclusive parameters to have better error messages.

For this PR, I think it would be sufficient to update the error message to what @markekraus is proposing without the `get-help` mention on the end which is to generic.

iSazonov:
@SteveL-MSFT Thanks for clarify and confirm my thoughts.

markekraus:
#5435 has been opened instead.

PR #5451, opened 2017-11-14
Add a test for the powershell hang fix

daxian-dbw:
Add test for the fix in PR #5356 
Also add a minor change to build.psm1 to ignore the 'Sync-PSTags' warning.

### DIFF
--- a/test/powershell/engine/Api/BasicEngine.Tests.ps1
+++ b/test/powershell/engine/Api/BasicEngine.Tests.ps1
@@ -21,3 +21,29 @@ Describe 'Basic engine APIs' -Tags "CI" {
         }
     }
 }
+
+Describe "Clean up open Runspaces when exit powershell process" -Tags "Feature" {
+    It "PowerShell process should not freeze at exit" {
+        $command = @'
+-c $rs = [runspacefactory]::CreateRunspacePool(1,5)
+$rs.Open()
+$ps = [powershell]::Create()
+$ps.RunspacePool = $rs
+$null = $ps.AddScript(1).Invoke()
+write-host should_not_hang_at_exit
+exit
+'@
+        $process = Start-Process pwsh -ArgumentList $command -PassThru
+        for ($i = 0; -not $process.HasExited -and $i -lt 5; $i++) {
+            Start-Sleep -Seconds 1
+        }
+
+        $expect = "powershell process exits in 5 seconds"
+        if (-not $process.HasExited) {
+            Stop-Process -InputObject $process -Force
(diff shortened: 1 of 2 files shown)


test/powershell/engine/Api/BasicEngine.Tests.ps1
```diff
@@ -21,3 +21,29 @@ Describe 'Basic engine APIs' -Tags "CI" {
...
+-c $rs = [runspacefactory]::CreateRunspacePool(1,5)
+$rs.Open()
+$ps = [powershell]::Create()
+$ps.RunspacePool = $rs
+$null = $ps.AddScript(1).Invoke()
+write-host should_not_hang_at_exit
+exit
+'@
+        $process = Start-Process pwsh -ArgumentList $command -PassThru
+        for ($i = 0; -not $process.HasExited -and $i -lt 5; $i++) {
```
adityapatwardhan: Please use the helper at: https://github.com/PowerShell/PowerShell/blob/master/test/tools/Modules/HelpersCommon/HelpersCommon.psm1#L1

test/powershell/engine/Api/BasicEngine.Tests.ps1
```diff
@@ -21,3 +21,29 @@ Describe 'Basic engine APIs' -Tags "CI" {
...
+            Start-Sleep -Seconds 1
+        }
+
+        $expect = "powershell process exits in 5 seconds"
```
adityapatwardhan: Should the verification be like:
```Powershell
$process.HasExited | Should Be $true
Stop-Process $process -ErrorAction SilentlyContinue

adityapatwardhan (changes requested):

test/powershell/engine/Api/BasicEngine.Tests.ps1
daxian-dbw: OK, I will use `Wait-UntilTrue`

test/powershell/engine/Api/BasicEngine.Tests.ps1
daxian-dbw: The `Should be $true` check will not generate a useful error message when it fails, so I prefer to the current check. But I will add `-ErrorAction SilentlyContinue`

daxian-dbw:
@adityapatwardhan Your comments are addressed. Please take another look. Thanks!

adityapatwardhan (approved):

iSazonov (approved):
Leave a comment

PaulHigin (approved):
LGTM

Issue #5440, opened 2017-11-14
Can not build with VS Code


iSazonov:
I can not build with VS Code:

```
PowerShell output: c:\Users\sie\Documents\GitHub\iSazonov\PowerShell/debug
WARNING: Run Sync-PSTags to update tags
Unable to load file: "c:\Users\sie\Documents\GitHub\iSazonov\PowerShell/debug"
Execution of { & "~/.rcedit/rcedit-x64.exe" "$($Options.Output)" --set-icon "$PSScriptRoot\assets\Powershell_black.ico" `
            --set-file-version $ReleaseVersion --set-product-version $ReleaseVersion --set-version-string "ProductName" "PowerShell
Core 6" `
            --set-requested-execution-level "asInvoker" --set-version-string "LegalCopyright" "(C) Microsoft Corporation.  All Rights
Reserved." } failed with exit code 1
At C:\Users\sie\Documents\GitHub\iSazonov\PowerShell\build.psm1:1862 char:13
+             throw "Execution of {$sb} failed with exit code $LASTEXIT ...
+             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (Execution of { ...ith exit code 1:String) [], RuntimeException
    + FullyQualifiedErrorId : Execution of { & "~/.rcedit/rcedit-x64.exe" "$($Options.Output)" --set-icon "$PSScriptRoot\assets\Power
   shell_black.ico" `
                --set-file-version $ReleaseVersion --set-product-version $ReleaseVersion --set-version-string "ProductName" "PowerShe
   ll Core 6" `
                --set-requested-execution-level "asInvoker" --set-version-string "LegalCopyright" "(C) Microsoft Corporation.  All Ri
   ghts Reserved." } failed with exit code 1

The terminal process terminated with exit code: 1
```

SteveL-MSFT:
I have a fix for the build, but not able to get the build output to show in vscode using a 2.0.0 task

iSazonov:
I see the output in Terminal window not Output windows.

rkeithhill:
VSCode tasks using schema version 2.0.0, will output to the terminal window instead of the Output window.  That's the way it is supposed to work.

PR #5453, opened 2017-11-14
fix build in vscode

SteveL-MSFT:
There were several issues:

In vscode, depending on where you started it from, it may not get the git commit id so it ends up empty.  Add a default value to use.

From the console, the commit id was typically in a form that is not parse-able as a fileversion.  Fix is to chop off everything right of the dash.

It also seems that building from the console and building from vscode, the output path may or may not include `pwsh`, however, rcedit needs the actual executable in the path.  

Fix https://github.com/PowerShell/PowerShell/issues/5440

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,17 +1,74 @@
 {
-    "version": "0.1.0",
-    "command": "pwsh",
-    "isShellCommand": true,
-    "showOutput": "always",
-    "suppressTaskName": true,
-    "args": [ "-command" ],
+    "version": "2.0.0",
+
+    "windows": {
+        "options": {
+            "shell": {
+                "executable": "pwsh.exe",
+                "args": [
+                    "-NoProfile",
+                    "-ExecutionPolicy",
+                    "Bypass",
+                    "-Command"
+                ]
+            }
+        }
+    },
+    "linux": {
+        "options": {
+            "shell": {
+                "executable": "/usr/bin/pwsh",
+                "args": [
+                    "-NoProfile",
+                    "-Command"
+                ]
+            }
+        }
+    },
+    "osx": {
+        "options": {
+            "shell": {
+                "executable": "/usr/local/bin/pwsh",
+                "args": [
+                    "-NoProfile",
+                    "-Command"
+                ]
+            }
+        }
+    },
 
     "tasks": [
         {
-            "taskName": "build",
-            "args": [ "Import-Module ${workspaceRoot}/build.psm1; Start-PSBuild -Output ${workspaceRoot}/debug" ],
-            "isBuildCommand": true,
+            "label": "Bootstrap",
+            "type": "shell",
+            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSBootstrap",
+            "problemMatcher": []
+        },
+        {
+            "label": "Clean Build",
+            "type": "shell",
+            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSBuild -Clean",
+            "problemMatcher": []
+        },
+        {
+            "label": "Build",
+            "type": "shell",
+            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSBuild -Output (Join-Path ${workspaceFolder} debug)",
+            "group": {
+                "kind": "build",
+                "isDefault": true
+            },
             "problemMatcher": "$msCompile"
+        },
+        {
+            "label": "Test",
+            "type": "shell",
+            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSPester -Path ${workspaceFolder}/test",
+            "group": {
+                "kind": "test",
+                "isDefault": true
+            },
+            "problemMatcher": "$pester"
         }
     ]
 }
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -47,7 +47,7 @@
         {
             "label": "Clean Build",
             "type": "shell",
-            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSBuild -Clean",
+            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSBuild -Clean -Output (Join-Path ${workspaceFolder} debug)",
             "problemMatcher": "$msCompile"
         },
         {
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -48,7 +48,7 @@
             "label": "Clean Build",
             "type": "shell",
             "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSBuild -Clean",
-            "problemMatcher": []
+            "problemMatcher": "$msCompile"
         },
         {
             "label": "Build",
@@ -59,16 +59,6 @@
                 "isDefault": true
             },
             "problemMatcher": "$msCompile"
-        },
-        {
-            "label": "Test",
-            "type": "shell",
-            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSPester -Path ${workspaceFolder}/test",
-            "group": {
-                "kind": "test",
-                "isDefault": true
-            },
-            "problemMatcher": "$pester"
         }
     ]
 }
--- a/build.psm1
+++ b/build.psm1
@@ -569,10 +569,28 @@ Fix steps:
             $ReleaseVersion = $ReleaseTagToUse
         } else {
             $ReleaseVersion = (Get-PSCommitId) -replace '^v'
+            $fileVersion = $ReleaseVersion.Split("-")[0]
+        }
+        if (!$ReleaseVersion) {
+            $ReleaseVersion = "6.0.0"
+            $fileVersion = "6.0.0"
         }
 
-        Start-NativeExecution { & "~/.rcedit/rcedit-x64.exe" "$($Options.Output)" --set-icon "$PSScriptRoot\assets\Powershell_black.ico" `
-            --set-file-version $ReleaseVersion --set-product-version $ReleaseVersion --set-version-string "ProductName" "PowerShell Core 6" `
+        Write-Verbose "Release version: $ReleaseVersion" -Verbose
+
+        $pwshPath = $Options.Output
+        if ($Environment.IsWindows) {
+            if (!$pwshPath.EndsWith("pwsh.exe")) {
+                $pwshPath = Join-Path $Options.Output "pwsh.exe"
+            }
+        } else {
+            if (!$pwshPath.EndsWith("pwsh")) {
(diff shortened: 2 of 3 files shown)


adityapatwardhan (approved):
LGTM

build.psm1
```diff
@@ -569,10 +569,28 @@ Fix steps:
...
+        }
+        if (!$ReleaseVersion) {
+            $ReleaseVersion = "6.0.0"
+            $fileVersion = "6.0.0"
+        }
+
+        Write-Verbose "Release version: $ReleaseVersion" -Verbose
+
+        $pwshPath = $Options.Output
+        if ($Environment.IsWindows) {
```
daxian-dbw: The whole `RCEdit` work is in `if ($Environment.IsWindows) {` block (see line 561), so no need to split the code path here.

build.psm1
```diff
@@ -569,10 +569,28 @@ Fix steps:
             $ReleaseVersion = $ReleaseTagToUse
         } else {
             $ReleaseVersion = (Get-PSCommitId) -replace '^v'
+            $fileVersion = $ReleaseVersion.Split("-")[0]
+        }
+        if (!$ReleaseVersion) {
```
daxian-dbw: Can you please add a comment about "In vscode, depending on where you started it from, it may not get the git commit id so it ends up empty. Add a default value to use."?

build.psm1
```diff
@@ -569,10 +569,28 @@ Fix steps:
...
+            $ReleaseVersion = "6.0.0"
+            $fileVersion = "6.0.0"
+        }
+
+        Write-Verbose "Release version: $ReleaseVersion" -Verbose
```
daxian-dbw: Please use `log` like rest of the file. I think it's better to make the message more specific on the RCEdit operation, like "Embed the icon resource to pwsh.exe, use release version: $ReleaseVersion"

build.psm1
```diff
@@ -569,10 +569,28 @@ Fix steps:
...
+        } else {
+            if (!$pwshPath.EndsWith("pwsh")) {
+                $pwshPath = Join-Path $Options.Output "pwsh"
+            }
         }
 
-        Start-NativeExecution { & "~/.rcedit/rcedit-x64.exe" "$($Options.Output)" --set-icon "$PSScriptRoot\assets\Powershell_black.ico" `
-            --set-file-version $ReleaseVersion --set-product-version $ReleaseVersion --set-version-string "ProductName" "PowerShell Core 6" `
+        Start-NativeExecution { & "~/.rcedit/rcedit-x64.exe" $pwshPath --set-icon "$PSScriptRoot\assets\Powershell_black.ico" `
+            --set-file-version $fileVersion --set-product-version $ReleaseVersion --set-version-string "ProductName" "PowerShell Core 6" `
```
daxian-dbw: `$fileVersion` is not set when `$ReleaseTagToUse` exists.

daxian-dbw (changes requested):

rkeithhill:
You might want to try this tasks.json:
```json
{
    "version": "2.0.0",

    "windows": {
        "options": {
            "shell": {
                "executable": "C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe",
                "args": [
                    "-NoProfile",
                    "-ExecutionPolicy",
                    "Bypass",
                    "-Command"
                ]
            }
        }
    },
    "linux": {
        "options": {
            "shell": {
                "executable": "/usr/bin/pwsh",
                "args": [
                    "-NoProfile",
                    "-Command"
                ]
            }
        }
    },
    "osx": {
        "options": {
            "shell": {
                "executable": "/usr/local/bin/pwsh",
                "args": [
                    "-NoProfile",
                    "-Command"
                ]
            }
        }
    },

    "tasks": [
        {
            "label": "Bootstrap",
            "type": "shell",
            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSBootstrap",
            "problemMatcher": []
        },
        {
            "label": "Clean Build",
            "type": "shell",
            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSBuild -Clean",
            "problemMatcher": []
        },
        {
            "label": "Build",
            "type": "shell",
            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSBuild -Output ${workspaceFolder}/debug",
            "group": {
                "kind": "build",
                "isDefault": true
            },
            "problemMatcher": "$msCompile"
        },
        {
            "label": "Test",
            "type": "shell",
            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSPester -Path ${workspaceFolder}/test",
            "group": {
                "kind": "test",
                "isDefault": true
            },
            "problemMatcher": "$pester"
        }
    ]
}
```
There seems to be an issue with the `$msCompile` problem matcher at the moment but that's an issue for VSCode.

build.psm1
SteveL-MSFT: Will do

build.psm1
SteveL-MSFT: I'll just remove this, only needed it to figure out why it wasn't working

build.psm1
SteveL-MSFT: Yeah, forgot rcedit only applied to Windows, will fix

build.psm1
SteveL-MSFT: Will fix

SteveL-MSFT:
I'll add a slightly modified version of @rkeithhill 's tasks.json.  Thanks for that!

.vscode/tasks.json
```diff
@@ -1,17 +1,74 @@
...
+            "label": "Bootstrap",
+            "type": "shell",
+            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSBootstrap",
+            "problemMatcher": []
+        },
+        {
+            "label": "Clean Build",
+            "type": "shell",
+            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSBuild -Clean",
+            "problemMatcher": []
```
rkeithhill: Doh! I forgot to put the $msCompile problem matcher here.  I could be wrong about this but Start-PSBuild -Clean doesn't **just** clean.  It cleans first and then builds, right?

.vscode/tasks.json
```diff
@@ -1,17 +1,74 @@
...
+            "group": {
+                "kind": "build",
+                "isDefault": true
+            },
             "problemMatcher": "$msCompile"
+        },
+        {
+            "label": "Test",
+            "type": "shell",
+            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSPester -Path ${workspaceFolder}/test",
```
rkeithhill: You should consider adding the `-Tag CI` parameter here.  Running VSCode as admin is not something most folks do and if you run these tests not elevated, you get a lot of failures.  You could also add two tasks - one that does a full test run and another that tests just the CI tests.

SteveL-MSFT: Actually, I think I'll remove this task as intent is people run CI + Feature (which is default), but you are right that the tests expect user to be elevated and, in general, shouldn't be running vscode elevated.

SteveL-MSFT:
@rkeithhill to answer some of your questions:

1. the `debug` folder is used as it's included in `.gitignore`
2. since this tasks.json is only used to build pwsh, I think it's preferable to depend on pwsh rather than Windows PowerShell (the rest of the tooling expects it)
3. `-Clean` cleans then builds

.vscode/tasks.json
```diff
@@ -1,17 +1,64 @@
...
+            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSBuild -Output (Join-Path ${workspaceFolder} debug)",
+            "group": {
+                "kind": "build",
+                "isDefault": true
+            },
             "problemMatcher": "$msCompile"
         }
     ]
-}
+}
```
iSazonov: New line.

iSazonov (approved):
@SteveL-MSFT @rkeithhill Thanks for the great PR!

.vscode/tasks.json
```diff
@@ -1,17 +1,64 @@
...
+        {
+            "label": "Clean Build",
+            "type": "shell",
+            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSBuild -Clean",
+            "problemMatcher": "$msCompile"
+        },
+        {
+            "label": "Build",
+            "type": "shell",
+            "command": "Import-Module ${workspaceFolder}/build.psm1; Start-PSBuild -Output (Join-Path ${workspaceFolder} debug)",
```
rkeithhill: Nice!  Can you add the same parameter to the `Clean Build` task above?  Also, `Clean Build` reads a little awkward.  What if we called this task `Rebuild`?  On VS at least, that implies Clean, then Build.

rkeithhill: BTW I was just going to warn you about this issue when I saw you not only already knew about it but solved it.  👍  Also, I tagged onto the VSCode issue on the $mscompile problem matcher not working - https://github.com/Microsoft/vscode/issues/34660

rkeithhill:
I think I'm caught up to your version of tasks.json but when I execute the `Rebuild`  task (formerly `Clean Build`), I'm getting this error:
```
PowerShell output: c:\Users\Keith\GitHub\rkeithhill\PowerShell\debug
WARNING: Run Sync-PSTags to update tags
Unable to load file: "c:\Users\Keith\GitHub\rkeithhill\PowerShell\debug"
Execution of { & "~/.rcedit/rcedit-x64.exe" "$($Options.Output)" --set-icon "$PSScriptRoot\assets\Powershell_black.ico" `
            --set-file-version $ReleaseVersion --set-product-version $ReleaseVersion --set-version-string "ProductName" "PowerShell Core 6" `
            --set-requested-execution-level "asInvoker" --set-version-string "LegalCopyright" "(C) Microsoft Corporation.  All Rights Reserved." } failed with exit code 1
```
Are you getting this error?  FWIW I get this error from the command line so maybe my source is in a bad state.

.vscode/tasks.json
SteveL-MSFT: will add

.vscode/tasks.json
SteveL-MSFT: Will add to `Clean Build` which I think is named fine

rkeithhill: @SteveL-MSFT I got a response from the VSCode team on the $mscompile issue.  All we need to do is add the command line arg - `/property:GenerateFullPaths=true`.  I tried putting this around line 485:
```
    $Arguments += "/property:GenerateFullPaths=true"
```
And it works!  There may be other places where this arg needs to be provided.

SteveL-MSFT: Will add.  Thanks!

daxian-dbw (approved):

PR #5455, opened 2017-11-15
Add HighEntropyVA to building pwsh

TravisEz13:

### DIFF
--- a/PowerShell.Common.props
+++ b/PowerShell.Common.props
@@ -104,6 +104,7 @@
     <DelaySign>true</DelaySign>
     <AssemblyOriginatorKeyFile>../signing/visualstudiopublic.snk</AssemblyOriginatorKeyFile>
     <SignAssembly>true</SignAssembly>
+    <HighEntropyVA>true</HighEntropyVA>
   </PropertyGroup>
 
 </Project>


SteveL-MSFT (approved):

src/powershell-unix/powershell-unix.csproj
```diff
@@ -7,6 +7,7 @@
     <AssemblyName>pwsh</AssemblyName>
     <OutputType>Exe</OutputType>
     <RuntimeIdentifiers>linux-x64;osx.10.12-x64;</RuntimeIdentifiers>
+    <HighEntropyVA>true</HighEntropyVA>
```
iSazonov: Why we don't add this in all csproj files via `PowerShell.Common.props`?

iSazonov (approved):
LGTM.

One minor question. Sometimes msbuild is a surprise. How we can check that the option really works?

TravisEz13:
We will run binskim during the official build which will tell us.
https://github.com/Microsoft/binskim

daxian-dbw:
We don't build `pwsh.exe`, so is the executable also has this option set on?
According to https://msdn.microsoft.com/en-us/library/jj835761.aspx,  If not, powershell may not be able to benefit from this option.

iSazonov:
I see the option in CoreCLR https://github.com/dotnet/coreclr/search?utf8=%E2%9C%93&q=HIGHENTROPYVA&type=

daxian-dbw (approved):
LGTM. :shipit:

PR #5457, opened 2017-11-15
Add standard set of vscode workspace setting files

rkeithhill:
Add an extensions.json file that will prompt folks to install extensions that are recommended for this workspace (c++, c#, powershell, cmake (editing syntax) and markdown linter.

Add settings.json to start to configure C# syntax, PoweShell trim trailing whitespace and configure the markdown lint extension.

Update launch.json to provide ability to debug PowerShell scripts.

BTW I'm hoping as more people on the team use the VSCode extension , it will help get more attention to fixing some lingering issues in the PS extension.  Dogfooding FTW!!

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,11 @@
+{
+  // See http://go.microsoft.com/fwlink/?LinkId=827846
+  // for the documentation about the extensions.json format
+  "recommendations": [
+    "ms-vscode.cpptools",
+    "ms-vscode.csharp",
+    "ms-vscode.PowerShell",
+    "twxs.cmake",
+    "DavidAnson.vscode-markdownlint"
+  ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,16 @@
+// Place your settings in this file to overwrite default and user settings.
+{
+  "editor.tabSize": 4,
+  "editor.insertSpaces": true,
+  "markdownlint.config": {
+    "MD004": false,
+    "MD024": false,
+    "MD033": false,
+    "MD034": false,
+    "MD038": false,
+    "MD042": false
+  },
+  "[powershell]": {
+    "files.trimTrailingWhitespace": true
+  }
+}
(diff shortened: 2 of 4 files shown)


rkeithhill:
BTW I should have added, the markdown linter integration into VSCode is really, really nice.  Much better to see issues while you are editing them than to wait for the Pester test to find them.  :-)

.vscode/settings.json
```diff
@@ -0,0 +1,16 @@
+// Place your settings in this file to overwrite default and user settings.
+{
+  "editor.tabSize": 4,
+  "editor.insertSpaces": true,
+  "markdownlint.config": {
+    "MD004": false,
```
SteveL-MSFT: can you add comment why these are being excluded?

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):
LGTM

SteveL-MSFT (approved):

.vscode/extensions.json
```diff
@@ -0,0 +1,11 @@
+{
+  // See http://go.microsoft.com/fwlink/?LinkId=827846
+  // for the documentation about the extensions.json format
+  "recommendations": [
```
iSazonov: GitHub don't like Json comments. I see https://stackoverflow.com/questions/244777/can-comments-be-used-in-json

Should we remove the comments or convert in another form?

rkeithhill: Perhaps but it didn't stop the VSCode team from checking json with comments: https://github.com/Microsoft/vscode/blob/master/.vscode/extensions.json
https://github.com/Microsoft/vscode/blob/master/tslint.json

And that comment is the default comment added when you run the `Extensions: Configure Recommended Extensions` command in VSCode.

iSazonov:
Reopen thePR to restart CI Appveyor

.vscode/extensions.json
iSazonov: If VS Code understands it, I'll merge.

iSazonov:
@rkeithhill Many thanks for your contribution!

PR #5462, opened 2017-11-15
get-childitem -LiteralPath should accept 'Include' or 'Exclude' filter

### LINKED ISSUE #3730: Get-ChildItem -Recurse shouldn't quietly ignore -Include and -Exclude when -LiteralPath is used instead of -Path
Even without using wildcards, `Get-ChildItem -Recurse` allows matching items on every level of the target hierarchy, possibly filtered with `-Include` and `-Exclude`.

However, if you use `-LiteralPath` rather than the (implied) `-Path` parameter, `-Include` and `-Exclude` are _quietly ignored_:

_If_ there's a good reason for this behavior (none that I'm aware of), then at the very least a _warning_ about `-Include` / `-Exclude`  getting ignored should be issued, or a separate parameter set should prevent combining `-LiteralPath` with `-Include` / `-Exclude`.

Steps to reproduce
------------------

```powershell
Get-ChildItem              $PSHOME -Recurse -Exclude *   # implies -Path $PSHOME
Get-ChildItem -LiteralPath $PSHOME -Recurse -Exclude *
```

Expected behavior
-----------------

Both calls should yield _no_ output, because all items should be excluded by virtue of `-Exclude *`.

Actual behavior
---------------

The 2nd command, due to `-Exclude *` getting _ignored_, outputs the entire subtree of `$PSHOME`.

Environment data
----------------

```powershell
PowerShell Core v6.0.0-alpha (v6.0.0-alpha.18) on macOS 10.12.4
PowerShell Core v6.0.0-alpha (v6.0.0-alpha.18) on Microsoft Windows 10 Pro (64-bit; v10.0.14393)
```

chunqingchen:
This is to resolve #3730 

When -LiteralPath is specified, filter is never applied. The fix is to adding filter even if user set -LiteralPath as true

Before fix:
Get-childitem -LiteralPath $PSHOME -exclude * 
will return all the files under $pshome as filter is not working

After fix:
Include/Exclude filters work as expected.


src/System.Management.Automation/engine/SessionStateContainer.cs
```diff
@@ -1648,7 +1648,25 @@ internal Collection<PSObject> GetChildItems(string[] paths, bool recurse, uint d
...
+                {
+                    // Temeporary set literal path as false to apply filter
+                    context.SuppressWildcardExpansion = false;
+                    ProcessPathItems(providerInstance, path, recurse, depth, context, out unUsedChildrenNotMatchingFilterCriteria, ProcessMode.Enumerate);
+                }
+                finally
+                {
+                    context.SuppressWildcardExpansion = true;       
+                }
+                }
```
iSazonov: Please correct indentation.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1
```diff
@@ -76,10 +76,15 @@ Describe "Get-ChildItem" -Tags "CI" {
         }
 
         It "Should return items recursively when using 'Include' or 'Exclude' parameters" {
-            (Get-ChildItem -Path $TestDrive -Depth 1).Count | Should Be 6
+            (Get-ChildItem -Path $TestDrive -Depth 1).Count | Should Be 6
```
iSazonov: Please revert the change (remove last space).

src/System.Management.Automation/engine/SessionStateContainer.cs
```diff
@@ -1648,7 +1648,25 @@ internal Collection<PSObject> GetChildItems(string[] paths, bool recurse, uint d
 
                 ContainerCmdletProvider providerInstance = GetContainerProviderInstance(provider);
 
-                if (path != null && this.ItemExists(providerInstance, path, context))
+                if (
+                    (context.Include != null && context.Include.Count > 0) ||
+                    (context.Exclude != null && context.Exclude.Count > 0))
+                {
+                    // Do the recursion manually so that we can apply the
+                    // include and exclude filters
```
iSazonov: Should we check "context.Stopping" as in line 1587?

test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1
```diff
@@ -76,10 +76,15 @@ Describe "Get-ChildItem" -Tags "CI" {
...
         It "Should return items recursively when using 'Include' or 'Exclude' parameters" {
-            (Get-ChildItem -Path $TestDrive -Depth 1).Count | Should Be 6
+            (Get-ChildItem -Path $TestDrive -Depth 1).Count | Should Be 6 
             (Get-ChildItem -Path $TestDrive -Depth 1 -Include $item_G).Count | Should Be 1
             (Get-ChildItem -Path $TestDrive -Depth 1 -Exclude $item_a).Count | Should Be 5
         }
+        
+        It "get-childitem -LiteralPath should accept 'Include' or 'Exclude' filter" {
+            (Get-ChildItem -LiteralPath $PSHOME -Exclude *).Count | Should Be 0
+            (Get-ChildItem -LiteralPath $PSHOME -Include *.dll).Count | Should Be (Get-ChildItem $PSHOME -Include *.dll).Count
```
iSazonov: I'd prefer to use a directory tree created in `BeforeAll`. We can use previous test if replece Path with LiteralPath.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1
chunqingchen: resolved

test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1
chunqingchen: Good point, but I also propose we are using $pshome as it is complicated enough. Many customer feedback bugs are failing on this exact scenario. Besides, this just helps me find another issue, it works~

src/System.Management.Automation/engine/SessionStateContainer.cs
chunqingchen: resolved

src/System.Management.Automation/engine/SessionStateContainer.cs
chunqingchen: This is not necessary.  In the previous code, if a user try to cancel the search in between, the code will stop on next path search iteration. In this situation, since we only have one literal path, it won't work.

iSazonov: Thanks for clarify!

Closed.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-ChildItem.Tests.ps1
iSazonov: I expect a symmetric in tests. Previous test use `-Path`. If we replace `-Path` with '-LiteralPath' we get symmetric test. This makes it easy to understand these tests. The test with `$PSHOME` could be additional test. But what do it really test? What is difference? If it test another code path(s) it is better to be explicit and create a test directory tree for these cases - $PSHOME doesn't make it clear which code path we're really checking.

chunqingchen:
@iSazonov Is it good to go?

iSazonov:
@chunqingchen I have one comment above https://github.com/PowerShell/PowerShell/pull/5462#discussion_r151317412
>I expect a symmetric in tests. Previous test use -Path. If we replace -Path with '-LiteralPath' we get symmetric test. This makes it easy to understand these tests. The test with $PSHOME could be additional test. But what do it really test? What is difference? If it test another code path(s) it is better to be explicit and create a test directory tree for these cases - $PSHOME doesn't make it clear which code path we're really checking.

adityapatwardhan (approved):

PR #5465, opened 2017-11-15
Change synopsis of install-powershell.ps1 to reflect that it works on Linux and macOS as well

bergmeister:
According to @SteveL-MSFT's tweet it works not only on Windows, but also Linux and macOS: https://twitter.com/Steve_MSFT/status/930599397460623360

### DIFF
--- a/tools/install-powershell.ps1
+++ b/tools/install-powershell.ps1
@@ -1,6 +1,6 @@
 <#
 .Synopsis
-    Install PowerShell Core on Windows.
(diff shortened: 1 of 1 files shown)


daxian-dbw (approved):

TravisEz13 (approved):

adityapatwardhan (approved):

TravisEz13:
Marking as GA because it's very low risk.  @adityapatwardhan agreed offline

Issue #5434, opened 2017-11-13
Invoke-Command should have a -Timeout parameter


mklement0:
Related: 
* #5433
* #5185
* #15696

I don't know if it would be appropriate for all `Invoke-Command` use cases (parameter sets), but being able to synchronously invoke a command with a timeout would be handy; e.g.:

```powershell
Invoke-Command -Timeout 120 { ... }
```

Environment data
----------------

Written as of beta.9

Groostav:
Any ideas what the status of this issue is?

Also, whats the best workaround? I'm having trouble getting `Start-Job` with the various job collectors to give me the semantics I want.

woodli:
This feature is quite useful.
I do not think to postpone again and again is a good idea.

branepar:
What is status? Basic functionality which would be real useful to get as the default 3 minute default is quite long and dirty to use the asjob workaround

rhubarb-geek-nz:
Hi, I have put together a proof of concept for an Invoke-CommandWithCancellationToken on [GitHub](https://github.com/rhubarb-geek-nz/CancellationTokenEvent)

[CommandWithCancellationToken.cs](https://github.com/rhubarb-geek-nz/CancellationTokenEvent/blob/main/CancellationTokenEvent/CommandWithCancellationToken.cs)

The solution contains complete(?) set of unit tests covering the behaviour

[CommandWithCancellationTokenTests.cs](https://github.com/rhubarb-geek-nz/CancellationTokenEvent/blob/main/UnitTests/CommandWithCancellationTokenTests.cs)

The unit tests demonstrate using the CancellationTokenSource.CancelAfter() to set the timeout.

rhubarb-geek-nz:
https://github.com/PowerShell/PowerShell/issues/19685

kasini3000:
hi all:
"kasini3000" base sqlite. 
It achieves multi threaded and multi process job timeout stopping through the time field. There are also task failure retry times and priority.
So, powershell is the best engine, and Kasini3000 is the best car.
https://github.com/kasini3000/kasini3000

3052:
https://github.com/PowerShell/PowerShell/issues/24289

3052:
reposted. that bot is bad, and you should feel bad.

Issue #5433, opened 2017-11-13
Receive-Job should have a -Timeout parameter to complement -Wait


mklement0:
`Receive-Job -Wait` is a convenient way of receiving job output (almost) as it is being produced, but it waits indefinitely for the job to complete.

Having `Receive-Job -Wait -Timeout <secs>` would be handy to enforce a timeout.

(`Wait-Job` already has a `-Timeout` parameter, but it doesn't support receiving job output)

Environment data
----------------

Written as of beta.9

davidBar-On:
Is this issue still relevant or a workaround such as `$j="job7"; Wait-Job $j -Timeout 5 >$null; Receive-Job $j` is considered good enough?

I am asking partly because PR #5471 that resolves a related issue was submitted some years ago but was never merged. Therefore it seems that the [suggested workaround](https://github.com/PowerShell/PowerShell/pull/5471#pullrequestreview-77138717) in this PR is practically considered good enough.

doctordns:
The PowerShell Cmdlet Working Group has reviewed this issue and has decided to close it as:  Won't Fix. We have seen very little community input that this is a needed feature. And as pointed out in the issue thread, there is a good workaround should the need arise.

mklement0:
The alleged workaround isn't one (it relates to a different cmdlet, `Start-Process`).

Dismissing suggestions solely based on how much community input has been received rather than on their merits is unfortunate in general.

Specifically, by the standards of this repo, I don't think that an issue having garnered 4 up-votes (though I'm unsure of the timing) qualifies as "very little community input".

aarismendi:
I would like to have this as well. Would like to receive background job output in real time with an optional timeout.

somescout:
I agree, I just faced a situation when one of 400 jobs froze, I had to interrupt the script, and the result of all other jobs disappeared. Workarounds can be used of course, but there are already too many of them - I would like to just write working code.

Makumasa:
Everyone keeps talking about a workaround, but could someone enlighten me as to what the workaround actually is? As mentioned by mklement0, the workaround linked in an earlier comment doesn't apply here and something like `$j="job7"; Wait-Job $j -Timeout 5 >$null; Receive-Job $j` doesn't actually address the original issue since output isn't returned until after the job completes (or the timeout expires). As far as I can tell, we can either wait for a job with a timeout or wait for a job and poll it for output, but not both.

[...]

PR #5471, opened 2017-11-16
Start-Process: add parameter -ExitTimeout

### LINKED ISSUE #5185: Start-Process should support timeout
[System.Diagnostics.Process] has a .WaitForExit which has a timeout value.

Start-Process should have a parameter where we can pass that instead of needing to -Passthru and capture the process object...

stuajnht:
This is to resolve #5185 by adding a `Timeout` parameter to `Start-Process`.

If the timeout is reached and the process hasn't finished, the process is stopped.

To perform the stopping of processes, the `Stop-Process` cmdlet is invoked. This is done to prevent repeating any checks needed when killing a process.

Before the process is stopped, a "search" is attempted to find all descendant processes to hopefully avoid any orphaned processes remaining active. This uses WMI on Windows and the `ps` command on UNIX-like systems, as this provided the most reliable way for cross-platform support.

As mentioned in the contributing guidelines, this change may warrant a note in the changelog (I am unsure where to put it).

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
@@ -4,6 +4,7 @@ Copyright (c) Microsoft Corporation. All rights reserved.
 
 using System;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Specialized;
@@ -1780,6 +1781,28 @@ namespace Microsoft.PowerShell.Commands
         public SwitchParameter Wait { get; set; }
 
         /// <summary>
+        /// If specified, wait for this number of seconds
+        /// </summary>
+        [Parameter]
+        [Alias("TimeoutSec")]
+        [ValidateNotNullOrEmpty]
+        [ValidateRange(1, 32767)]
+        public int Timeout
+        {
+            get
+            {
+                return _timeout;
+            }
+            set
+            {
+                _timeout = value * 1000;
+                _timeoutSpecified = true;
+            }
+        }
+        private int _timeout = System.Threading.Timeout.Infinite;
+        private bool _timeoutSpecified;
+
+        /// <summary>
         ///  Default Environment
         /// </summary>
         [Parameter(ParameterSetName = "Default")]
@@ -2013,14 +2036,17 @@ namespace Microsoft.PowerShell.Commands
                 }
             }
 
-            if (Wait.IsPresent)
+            if (Wait.IsPresent || _timeoutSpecified)
             {
                 if (process != null)
                 {
                     if (!process.HasExited)
                     {
 #if UNIX
-                        process.WaitForExit();
+                        if (!process.WaitForExit(_timeout))
+                        {
+                            StopProcessOnTimeout(process);
+                        }
 #else
                         _waithandle = new ManualResetEvent(false);
 
@@ -2028,15 +2054,22 @@ namespace Microsoft.PowerShell.Commands
                         ProcessCollection jobObject = new ProcessCollection();
                         if (jobObject.AssignProcessToJobObject(process))
                         {
-                            // Wait for the job object to finish
-                            jobObject.WaitOne(_waithandle);
+                            // Wait for the job object to finish, or kill it if a timeout occurs
+                            jobObject.WaitOne(_waithandle, _timeout);
+                            if (!process.HasExited)
+                            {
+                                StopProcessOnTimeout(process);
+                            }
                         }
                         else if (!process.HasExited)
                         {
                             // WinBlue: 27537 Start-Process -Wait doesn't work in a remote session on Windows 7 or lower.
                             process.Exited += new EventHandler(myProcess_Exited);
                             process.EnableRaisingEvents = true;
-                            process.WaitForExit();
+                            if (!process.WaitForExit(_timeout))
+                            {
+                                StopProcessOnTimeout(process);
+                            }
                         }
 #endif
                     }
@@ -2121,6 +2154,123 @@ namespace Microsoft.PowerShell.Commands
                     processEnvironment.Add(entry.Key.ToString(), entry.Value.ToString());
                 }
             }
+        }
+
+        /// <summary>
+        /// Attempt to stop the process when the timeout has expired.
+        /// <see cref="StopProcessCommand" /> is used to stop the process
+        /// </summary>
+        /// <param name="process">
+        /// The process that should be stopped
+        /// </param>
+        private void StopProcessOnTimeout(Process process)
+        {
+            string message = StringUtil.Format(ProcessResources.StartProcessTimeoutExceeded, process.ProcessName);
+            ErrorRecord er = new ErrorRecord(new TimeoutException(message), "StartProcessTimeoutExceeded", ErrorCategory.OperationTimeout, process);
+
+            StopProcessCommand stop = new StopProcessCommand();
+            stop.Id = GetProcessTreeIds(process);
+            foreach (Process p in stop.Invoke<Process>()) { }
+
+            ThrowTerminatingError(er);
+        }
+
+        /// <summary>
+        /// Gets IDs of descendant processes, started by a process
+        /// On Windows, this reads output from WMI commands
+        /// On UNIX, this reads output from `ps axo pid,ppid`
+        /// </summary>
+        /// <param name="parentProcess">
+        /// The parent process to use to resolve the process tree IDs
+        /// </param>
+        /// <returns>
+        /// IDs of the parent process and all its descendants
+        /// </returns>
+        private int[] GetProcessTreeIds(Process parentProcess)
+        {
+            List<int> stopProcessIds = new List<int> {parentProcess.Id};
+            string processRelationships = "";
+#if UNIX
+            try
+            {
+                Process ps = new Process();
+                ps.StartInfo.FileName = "ps";
+                ps.StartInfo.Arguments = "axo pid,ppid";
+                ps.StartInfo.UseShellExecute = false;
+                ps.StartInfo.RedirectStandardOutput = true;
+                ps.Start();
+
+                processRelationships = ps.StandardOutput.ReadToEnd();
+
+                ps.WaitForExit();
+                ps.Close();
+            }
+            catch (Win32Exception)
+            {
+                WriteWarning(
+                    StringUtil.Format(ProcessResources.CouldNotResolveProcessTree, parentProcess.ProcessName)
+                    + " "
+                    + ProcessResources.DescendantProcessesPossiblyRunning
+                );
+                return stopProcessIds.ToArray();
+            }
+#else
+            string searchQuery = "Select ProcessID, ParentProcessID From Win32_Process";
+            try
+            {
+                using (CimSession cimSession = CimSession.Create(null))
+                {
+                    IEnumerable<CimInstance> processCollection =
+                        cimSession.QueryInstances("root/cimv2", "WQL", searchQuery);
+                    StringBuilder sb = new StringBuilder();
+                    foreach (CimInstance processInstance in processCollection)
+                    {
+                        sb.Append(processInstance.CimInstanceProperties["ProcessID"].Value.ToString());
+                        sb.Append(' ');
+                        sb.Append(processInstance.CimInstanceProperties["ParentProcessID"].Value.ToString());
+                        sb.Append(Environment.NewLine);
+                    }
+                    processRelationships = sb.ToString();
+                }
+            }
+            catch (CimException)
+            {
+                WriteWarning(
+                    StringUtil.Format(ProcessResources.CouldNotResolveProcessTree, parentProcess.ProcessName)
+                    + " "
+                    + ProcessResources.DescendantProcessesPossiblyRunning
+                );
+                return stopProcessIds.ToArray();
+            }
+#endif
+
+            // processList - key: process ID, value: parent process ID
+            Dictionary<int, int> processList = new Dictionary<int, int>();
+            int pid = 0;
+            int ppid = 0;
+            string relationshipPattern = @"\s?([0-9]+)\s+([0-9]+)";
+            foreach (Match psLine in Regex.Matches(processRelationships, relationshipPattern))
+            {
+                pid = int.Parse(psLine.Groups[1].Value);
+                ppid = int.Parse(psLine.Groups[2].Value);
+                processList.Add(pid, ppid);
+            }
+
+            int position = 0;
+            do
+            {
+                foreach (KeyValuePair<int, int> process in processList)
+                {
+                    if (process.Value == stopProcessIds[position])
+                    {
+                        stopProcessIds.Add(process.Key);
+                    }
+                }
+
+                position++;
+            } while (position <= (stopProcessIds.Count - 1));
+
+            return stopProcessIds.ToArray();
         }
 
         private Process Start(ProcessStartInfo startInfo)
@@ -2605,12 +2755,15 @@ namespace Microsoft.PowerShell.Commands
         /// <param name="waitHandleToUse">
         /// WaitHandle to use for waiting on the job object.
         /// </param>
-        internal void WaitOne(ManualResetEvent waitHandleToUse)
+        /// <param name="timeout">
+        /// Wait for this number of milliseconds before a time-out occurs
+        /// </param>
+        internal void WaitOne(ManualResetEvent waitHandleToUse, Int32 timeout = Timeout.Infinite)
         {
             TimerCallback jobObjectStatusCb = this.CheckJobStatus;
             using (Timer stateTimer = new Timer(jobObjectStatusCb, waitHandleToUse, 0, 1000))
             {
-                waitHandleToUse.WaitOne();
+                waitHandleToUse.WaitOne(timeout);
             }
         }
     }
--- a/src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
+++ b/src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
@@ -1781,26 +1781,12 @@ namespace Microsoft.PowerShell.Commands
         public SwitchParameter Wait { get; set; }
 
         /// <summary>
-        /// If specified, wait for this number of seconds
+        /// If specified, wait for this number of milliseconds for the process to exit.
         /// </summary>
         [Parameter]
-        [Alias("TimeoutSec")]
         [ValidateNotNullOrEmpty]
-        [ValidateRange(1, 32767)]
-        public int Timeout
-        {
-            get
-            {
-                return _timeout;
-            }
-            set
-            {
-                _timeout = value * 1000;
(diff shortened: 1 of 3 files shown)


src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
```diff
@@ -2123,6 +2156,123 @@ private void LoadEnvironmentVariable(ProcessStartInfo startinfo, IDictionary Env
...
+        /// </summary>
+        /// <param name="process">
+        /// The process that should be stopped
+        /// </param>
+        private void StopProcessOnTimeout(Process process)
+        {
+            string message = StringUtil.Format(ProcessResources.StartProcessTimeoutExceeded, process.ProcessName);
+            ErrorRecord er = new ErrorRecord(new TimeoutException(message), "StartProcessTimeoutExceeded", ErrorCategory.OperationTimeout, process);
+
+            StopProcessCommand stop = new StopProcessCommand();
```
SteveL-MSFT: Seems like Stop-Process should have a switch called `-IncludeChildProcesses` and you would use that here

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
```diff
@@ -1780,6 +1781,28 @@ public ProcessWindowStyle WindowStyle
         public SwitchParameter Wait { get; set; }
 
         /// <summary>
+        /// If specified, wait for this number of seconds
+        /// </summary>
+        [Parameter]
+        [Alias("TimeoutSec")]
+        [ValidateNotNullOrEmpty]
+        [ValidateRange(1, 32767)]
+        public int Timeout
```
SteveL-MSFT: I think that `Timeout` is a bit misleading as it's not really a timeout on starting the process, but on how long it's expected to run.  Perhaps calling this `-WaitForExitTimeout` would be more accurate?  cc @HemantMahawar @joeyaiello

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
```diff
@@ -2013,30 +2036,40 @@ protected override void BeginProcessing()
...
                         ProcessCollection jobObject = new ProcessCollection();
                         if (jobObject.AssignProcessToJobObject(process))
                         {
-                            // Wait for the job object to finish
-                            jobObject.WaitOne(_waithandle);
+                            // Wait for the job object to finish, or kill it if a timeout occurs
+                            jobObject.WaitOne(_waithandle, _timeout);
+                            if (!process.HasExited)
+                            {
+                                StopProcessOnTimeout(process);
```
SteveL-MSFT: I'm wondering if we actually want to stop the process on timeout by default.  I think it might be sufficient to return a non-terminating error so that `-PassThru` returns the process so the user can terminate it if so desired?  cc @HemantMahawar @joeyaiello

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
```diff
@@ -2123,6 +2156,123 @@ private void LoadEnvironmentVariable(ProcessStartInfo startinfo, IDictionary Env
...
+                Process ps = new Process();
+                ps.StartInfo.FileName = "ps";
+                ps.StartInfo.Arguments = "axo pid,ppid";
+                ps.StartInfo.UseShellExecute = false;
+                ps.StartInfo.RedirectStandardOutput = true;
+                ps.Start();
+
+                processRelationships = ps.StandardOutput.ReadToEnd();
+
+                ps.WaitForExit();
```
SteveL-MSFT: Is there a potential this could not exit?

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
```diff
@@ -2123,6 +2156,123 @@ private void LoadEnvironmentVariable(ProcessStartInfo startinfo, IDictionary Env
...
+                ps.Close();
```
SteveL-MSFT: Should we check the exit code?

test/powershell/Modules/Microsoft.PowerShell.Management/Start-Process.Tests.ps1
```diff
@@ -116,6 +116,52 @@ Describe "Start-Process" -Tags @("Feature") {
...
+    BeforeAll {
+        if ($IsWindows) {
+            $pingParam = "-n 30 localhost"
+        }
+        elseif ($IsLinux -Or $IsMacOS) {
+            $pingParam = "-c 30 localhost"
+        }
+
+        # Find where test/powershell is so we can find the testexe command relative to it
+        $powershellTestDir = $PSScriptRoot
```
SteveL-MSFT: You shouldn't have to search for testexe, I think the current pattern is to expect using `Start-PSPester` which builds testexe and puts it into the path

SteveL-MSFT (changes requested):
Thanks for the contribution!

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
bergmeister: From a user's perspective, I'd expect `PowerShell` to stop the process for me because this is how at least I would use this parameter.

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
```diff
@@ -1780,6 +1781,28 @@ public ProcessWindowStyle WindowStyle
         public SwitchParameter Wait { get; set; }
 
         /// <summary>
+        /// If specified, wait for this number of seconds
```
iSazonov: Please follow common convention and add final dot in public comments.

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
```diff
@@ -2013,30 +2036,40 @@ protected override void BeginProcessing()
                 }
             }
 
-            if (Wait.IsPresent)
+            if (Wait.IsPresent || _timeoutSpecified)
```
iSazonov: If we'll use autogenerated property we can check Timeout != Threading.Timeout.Infinite

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
```diff
@@ -2013,30 +2036,40 @@ protected override void BeginProcessing()
...
             {
                 if (process != null)
                 {
                     if (!process.HasExited)
                     {
 #if UNIX
```
iSazonov: We should remove two code paths and use `public bool WaitForExit(int milliseconds);` for all plaforms.

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
iSazonov: I wonder why we're actually making this change if the following works fine `Start-Process -PassThru | Wait-Process -Timeout`? It seems we duplicate a code.
I'd expect that we enhance Wait-Process cmdlet with -PassThru to get `Start-Process -PassThru | Wait-Process -Timeout 10 -PassThru | Stop-Process`.

iSazonov (commented):
I wonder why we're actually making this change if the following works fine `Start-Process -PassThru | Wait-Process -Timeout`? It seems we duplicate a code.
I'd expect that we enhance Wait-Process cmdlet with -PassThru to get `Start-Process -PassThru | Wait-Process -Timeout 10 -PassThru | Stop-Process`.

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
```diff
@@ -1780,6 +1781,28 @@ public ProcessWindowStyle WindowStyle
         public SwitchParameter Wait { get; set; }
 
         /// <summary>
+        /// If specified, wait for this number of seconds
+        /// </summary>
+        [Parameter]
+        [Alias("TimeoutSec")]
+        [ValidateNotNullOrEmpty]
+        [ValidateRange(1, 32767)]
```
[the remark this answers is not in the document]
stuajnht: @SteveL-MSFT No reason, just that the `Wait-Process` cmdlet has this limit so I used the same range.
@iSazonov I have changed it to `[ValidateRange(ValidateRangeKind.Positive)]`.

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
stuajnht: Agreed; `-ExitTimeout` makes more sense. Would changing `-Wait` to an int cause any problems with existing scripts?

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
stuajnht: @SteveL-MSFT @bergmeister I thought about both of these ways when implementing this, but settled on stopping the process on timeout, similar to [https://ss64.com/bash/timeout.html](https://ss64.com/bash/timeout.html). The reasoning is if I started a process and gave it a time limit, I would expect it to stop if the limit is reached, rather than just let me know it had gone over the limit. Is it worth adding a switch `-StopOnExitTimeout` (or similar)? Or, should the below suggestion be used?

@iSazonov I see your point, and didn't think of it being run this way. I suppose adding the `-ExitTimeout` parameter is about simplifying the amount of cmdlets needed. Would you prefer me to implement `-PassThru` for `Wait-Process` instead?

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
stuajnht: Ideally. As there currently isn't one, I implemented the `GetProcessTreeIds` function to do this role.

If this code is accepted, I had planned to open an issue / PR about somehow integrating this function into the `Stop-Process` cmdlet in the future (so as not to change too much code in this PR), instead of it being located in `Start-Process`.

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
stuajnht: Agreed. Shall add a few seconds wait before returning, just in case.

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
stuajnht: I had assumed this would either succeed or throw an error. Shall add a check just in case.

test/powershell/Modules/Microsoft.PowerShell.Management/Start-Process.Tests.ps1
stuajnht: Sorry, didn't realise it worked that way. Shall change to just use `textexe`.

stuajnht:
Hi all,

Thanks for your review feedback. I have made a number of changes based on it:
* Changed `Timeout` parameter to `ExitTimeout`
* Skipped searching for `testexe`, as `Start-PSPester` puts it into the path
* Set the upper limit for `ExitTimeout` to be a positive range, rather than 32767
* Used autogenerated property for `ExitTimeout`
* Set a timeout and checked the exit code for the `ps axo` process
* Used common coding convention for comments

Any questions that I have are in the relevant feedback comments.

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
stuajnht: I'm sorry, I don't quite understand what you are asking here. Are you saying to remove the `ProcessCollection` / `jobObject` and just have a `WaitForExit` for all platforms? e.g.

```
if (!process.WaitForExit(ExitTimeout))
{
    StopProcessOnTimeout(process);
}
```

From what I can see, the Windows specific code keeps track of any child processes that are created, which doesn't happen for UNIX. I am unsure of the reasoning behind this, as the earliest commit on GitHub for these lines does not contain any helpful information.

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
stuajnht: Have changed as requested

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
```diff
@@ -1780,6 +1781,28 @@ public ProcessWindowStyle WindowStyle
...
+            {
+                return _timeout;
+            }
+            set
+            {
+                _timeout = value * 1000;
+                _timeoutSpecified = true;
+            }
+        }
+        private int _timeout = System.Threading.Timeout.Infinite;
```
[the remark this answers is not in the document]
stuajnht: Thanks. I was not aware of doing it this way.

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
stuajnht: Added. The other public comments in `Start-Process` didn't have a dot, so I followed the same format as those.

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
SteveL-MSFT: We should probably leave `-Wait` as-is for compat reasons.  `-ExitTimeout` works for me.

SteveL-MSFT:
The concept is definitely a good one and something needed.  I'd like @PowerShell/powershell-committee to review the design to make sure it fits with PowerShell norms.

iSazonov:
As I said above I'd prefer to see `Start-Process -PassThru | Wait-Process -Timeout 10 -PassThru | Stop-Process`.
Another thought is that `-Wait` and `-ExitTimeout` is duplicates. We'd accept a slight breaking change `-Wait | -Wait <timeout>` where `-Wait` is infinite timeout. It is breaking only `-Wait:$true` that is rare case.
Another thought is - can there be other timeouts? If no we'd use `-Timeout` name.

mattpwhite:
Is this a race?  It looks like PIDs are being collected as integers (as opposed to handles that might prevent re-use), then later those PIDs are killed, but there's nothing to say that the PID hasn't been reused by the OS for a different process at that point, or is there?

mklement0:
@stuajnht: Consistency of parameter names across cmdlets is important, so I agree with @iSazonov that we should revert to `-Timeout`.

@iSazonov: Being able to specify a timeout value directly is a welcome convenience that we should add to other cmdlets as well - see #5434 and #5433 

PowerShell lacks the concept of an optional argument value altogether, and I suggest not opening that can of worms with something `-Wait | -Wait <timeout>`. (If anything, to avoid ambiguity, it would have to have the form `-Wait | -Wait:10`, analogous to how POSIX shells do it.)

To me, there's nothing wrong with a separate `-Timeout` parameter.

stuajnht:
@mklement0 @iSazonov I am happy to change the parameter back to `-Timeout` if no-one else has any objections about the parameter name needing to be clarified about what it does.

@mattpwhite After trying with a variety of programs, I believe that you are right. When I was coding this, killing the parent process would keep the child processes running. As was mentioned in [one of the comments](https://github.com/PowerShell/PowerShell/pull/5471#discussion_r151304540), ideally the `Stop-Process` cmdlet would have the ability to terminate child processes. My idea was to include something in this PR to prevent child processes being leftover when the timeout occurs, but I realise now it should be in its own PR instead.

@SteveL-MSFT Has anything been considered if the process should be stopped on the timeout being reached, or if it should just warn the user?

Thanks all, and I'll update the PR once I have more information.

mklement0:
@stuajnht:

> if the process should be stopped on the timeout being reached

To me, that calls for another switch: `-StopOnTimeout`; by default, _no_ action (other than the timeout error) should be taken.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and would like the parameter to be called `-MaximumDuration` and validate that when used with `-PassThru`, the dead process object should be emitted

mklement0:
May I ask what the rationale is for not sticking with the well-established `-Timeout` name for the parameter (used in several other cmdlets)?

Aside from this inconsistency, "duration" smacks of `[timespan]`.

mklement0:
Implying `-Wait` when a timeout is specified definitely makes sense, but, again ... uh ... calling a timeout a timeout makes sense too.

SteveL-MSFT:
@mklement0 the main push back on calling it `Timeout` is that it's not a timeout on the starting of a process, but the max lifetime allowed for the started process.  Duration sounded better than Lifetime.

mklement0:
@SteveL-MSFT: 

I appreciate your letting me know.

I don't think there's any point in distinguishing between _launch_ duration and _execution_ duration - they should be treated as a _single_ timeout, starting at the launch point in time.

If there's a _technical_ reason for not being able to enforce the timeout during the _launch_ phase a simple note in the documentation will do.

Even if you did want to make the distinction you're suggesting, `-MaximumDuration` is no less ambiguous than `-Timeout`: duration of _what_?

iSazonov:
I'd prefer to see unified parameter names. These parameters (Timeout) are easier to discover and remember than the unique parameters (MaximumDuration) in each cmdlet.

SteveL-MSFT:
@mklement0 @iSazonov appreciate all the feedback and you both make valid points, will bring it back up

SteveL-MSFT:
@PowerShell/powershell-committee revisited this and would be ok with `-ExecutionTimeout`.  In general, documentation should not be required to understand the intent or purpose of a parameter.  The name should be self documenting as much as possible without being overly verbose.  Just `-Timeout` is ambiguous of the purpose, but `-ExecutionTimeout` is concise and reasonable.

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
```diff
@@ -1779,6 +1780,14 @@ public ProcessWindowStyle WindowStyle
         [Parameter]
         public SwitchParameter Wait { get; set; }
 
+        /// <summary>
+        /// If specified, wait for this number of milliseconds for the process to exit.
+        /// </summary>
+        [Parameter]
+        [ValidateNotNullOrEmpty]
+        [ValidateRange(ValidateRangeKind.Positive)]
+        public int ExitTimeout { get; set; } = Timeout.Infinite;
```
TravisEz13: needs to be updated per committee review

TravisEz13 (changes requested):

iSazonov:
@stuajnht Could you please continue?

TravisEz13:
@iSazonov Please consider the bots comment the request to continue.  By commenting, you are defeating the bot.

iSazonov:
@TravisEz13 Thanks for clarify. I thought the bot only counted commits.
I think the PR is useful and we could grab and continue.

TravisEz13:
@iSazonov But a closed PR can be continued as well as long as the owner doesn't delete the branch.  The bot will label all of them as `review-abandoned` so we can query for them.  If you think the wording of the message of the bot needs to be updated, create a PR or just file an issue with a proposal and mention me.

iSazonov:
@TravisEz13 I open Issue #6446 with suggestion.

PR #5473, opened 2017-11-16
Fix the regex for '-ReleaseTag'

daxian-dbw:


TravisEz13 (approved):

adityapatwardhan (approved):

daxian-dbw:
This PR is only to change the regex for '-ReleaseTag', so that we can build for 'v6.0.0-rc'. There is no need to wait for the CI to finish.

Issue #5391, opened 2017-11-09
Remove PerformWSManPluginReportCompletion from pwrshplugin.dll


mirichmo:
[PerformWSManPluginReportCompletion](https://github.com/PowerShell/PowerShell/blob/b3940fe15babae000dc6ee4f31bebf3f6a843a3b/src/powershell-native/nativemsh/pwrshplugin/pwrshplugin.h#L1095) is dead code that is no longer used. It should have been removed during the purge of the workflow code and is safe to remove.

The [corresponding entry in pwrshplugin.def](https://github.com/PowerShell/PowerShell/blob/b3940fe15babae000dc6ee4f31bebf3f6a843a3b/src/powershell-native/nativemsh/pwrshplugin/pwrshplugin.def#L18) should also be removed as well.

### Version Info
-------------------------
6.0.0-beta.9

PR #5498, opened 2017-11-18
Remove PerformWSManPluginReportCompletion from pwrshplugin.dll

bergmeister:
This resolves issue #5391, which is tagged with `6.0.0-GA`, therefore this should help with getting pwsh out of the door in January.

It removes `PerformWSManPluginReportCompletion` as stated in the issue and its associated class member `g_pPluginContext `, which is now unused as well.

The best reviewer is probably @mirichmo , who raised the initial issue with very helpful descriptions or @dantraMSFT who is currently assigned to the issue.


dantraMSFT (approved):
LGTM

Issue #5012, opened 2017-10-04
Feature: Windows Installer to offer option of context menu to open PowerShell in explorer


bergmeister:
As a user, I often want to start PowerShell in the current directory in which I am browsing in explorer.exe.
Being able to have a context menu so that I can just right click and select `Open PowerShell here` or `Open PowerShell as Admininstrator` here would be very useful. Although a user can setup this manually via the Windows registry (see e.g. [here](https://www.howtogeek.com/165268/how-to-add-open-powershell-here-to-the-context-menu-in-windows/)), this is a tedious process that is often not worth the effort and therefore should rather be replaced by an option in the installer.
What do you think?

SteveL-MSFT:
I think if you can make it a checkbox in the installer, then it seems like a great feature.

xee5ch:
I have been meaning to learn about WiX for years now, and a quick examination shows this is a WiX-based installer.  Can I take a stab at this?

bergmeister:
Yes, it is a WiX based installer. I have not started on it yet, so feel free to give it a try and also ask questions. It is very easy to build via this PowerShell function [here](https://github.com/PowerShell/PowerShell/blob/master/build.psm1#L1802) and you do not even need to build the whole project, you could just point the `-ProductSourcePath` parameter to a bin folder of a dummy project as long as it contains a file named `powershell.exe`
What you will need to do is basically creating and setting registry entries, which means that this action needs to happen in the part where the installer runs with elevated rights, i.e. you will probably need to create a new dialogue with 2 checkboxes. Note that the example link that I have given at the top uses `HKEY_CLASSES_ROOT` but it might be better to use the `HKEY_CURRENT_USER`. In terms of registry entries, I found that 2 entries need to be added:
- One in the `Directory\Background` folder to enable the context menu when right clicking any where in explorer
- Another one in `Directory\shell` to enable the context menu when right clicking on a specific folder
The syntax between the 2 will slightly differently
You will find that there are already folders named `Powershell` of `Windows PowerShell` in the registry so you will need to come up with a folder name, so maybe `PowerShellCore6`? Since the installer does not allow side by side installation any more, a version number is not needed at the moment but it might be better since this might change in the future.,

SteveL-MSFT:
@xee5ch consider this yours!  I've assigned it to myself as GitHub won't allow me to assign it to you, but I don't want other people duplicating your work.

xee5ch:
[...]

PR #5499, opened 2017-11-18
WIP Make it easier to build an installable package locally for WiX development and document it

bergmeister:
Issue #5012 has highlighted that it is not easy to get started with development work on the Windows installer.

This PR documents for the first time how to build an installable package locally and simplifies the process by moving `New-MSIPackage` to the  `packaging` module where it better belongs to and more defaults/validation are added as well.

Since `New-MSIPackage` throws a very nice error with a download link to the required WiX toolset, I decided to not duplicate this in the documentation about having to install the WiX toolset.

I guess the best reviewer is @SteveL-MSFT who recently did a lot of installer work.


docs/building/windows-core.md
```diff
@@ -77,3 +77,13 @@ You can run our cross-platform Pester tests with `Start-PSPester`.
 ## Building in Visual Studio
 
 We currently have the issue [#3400](https://github.com/PowerShell/PowerShell/issues/3400) tracking this task.
+
+## Building an installer
```
iSazonov: We use `Start-PSPackage` function. See https://github.com/PowerShell/PowerShell/blob/cb048ea9f4f7ccc3612141ee76f7937b9ca1c823/docs/maintainers/releasing.md

build.psm1
```diff
@@ -1986,6 +1998,11 @@ function New-MSIPackage
     $wixCandleExePath = Join-Path $wixToolsetBinPath "Candle.exe"
     $wixLightExePath = Join-Path $wixToolsetBinPath "Light.exe"
 
+    if ($null -eq (Get-Module packaging))
+    {
+        # Import Get-PackageSemanticVersion function from packaging module
+        Import-Module "$PSScriptRoot\tools\packaging\packaging.psd1"
```
iSazonov: I was sure we were using auto-loading.

docs/building/windows-core.md
bergmeister: `Start-PSPackage` is not very developer friendly because it has ridiculous requirements such as that the checkout must be at the root of the filesystem and that PowereShell is compiled in a very specific way. The  purpose of this PR is to make it easier to build MSIs locally when changing WiX code.

build.psm1
bergmeister: See the example where it only imports build.psm1. We would need to have a module manifest for the build module then. Or maybe the `New-MSIPackage` should be moved into the packaging module?

build.psm1
```diff
@@ -1918,6 +1918,17 @@ function Get-PackageVersionAsMajorMinorBuildRevision
     $packageVersion
 }
 
+<#
+.Synopsis
+    Creates a Windows installer MSI file and assumes that the binaries are already built using 'Start-PSBuild'.
```
SteveL-MSFT: change the word `file` to `package`

docs/building/windows-core.md
SteveL-MSFT: `Start-PSPackage` also does some other checks, but that's more intended for creating a release package and I can understand the scenario for a developer testing their wix changes.  However, it's actually pretty easy to have `PowerShell` repo as your root using either `New-PSDrive` or `Subst`.

SteveL-MSFT: Instead of calling it `Building an installer`, I think it's more correct to say `Building an Installable Package` (where installer is msi)

SteveL-MSFT (changes requested):

docs/building/windows-core.md
iSazonov: I think we should support one entry point - it is easily to support (in docs too).
Also it will be best to have the same code (csproj?) to build on local and on GitHub.

build.psm1
TravisEz13: New-MSIPackage should be moved

docs/building/windows-core.md
```diff
@@ -77,3 +77,13 @@ You can run our cross-platform Pester tests with `Start-PSPester`.
...
 We currently have the issue [#3400](https://github.com/PowerShell/PowerShell/issues/3400) tracking this task.
+
+## Building an installable Package
+
+The installer is `WiX` based. You can create a Debug-x64 MSI as follows:
+
+````powershell
+Import-Module .\build.psm1
+Start-PSBuild
+New-MSIPackage -ProductSourcePath '.\src\powershell-win-core\bin\Debug\netcoreapp2.0\win7-x64\publish' -ProductTargetArchitecture x64 -ProductVersion '1.2.3'
```
TravisEz13: New-MSI package should not be exposed.

TravisEz13 (changes requested):
see comments

iSazonov:
@bergmeister Could you please correct the PR title and description?

docs/building/windows-core.md
TravisEz13: You could document how to use start-PsPackage for this purpose with the `-SkipReleaseChecks` parameter.

TravisEz13 (approved):

bergmeister:
@iSazonov OK. I've adapt the title and description now.

SteveL-MSFT:
Thinking further about this, I think I prefer not exposing `New-MSIPackage` and treat that as an internal function.  Have users use `Start-PSPackage` (so a single way to create msi packages rather than two ways).  However, we can update `Start-PSPackage` to automatically create a PSDrive to resolve the "root folder" requirement (to not expose "private" folder path information) and use that.

bergmeister:
@SteveL-MSFT If you do not want to expose `New-MSIPackage` any more, then an alternative would be to remove it from the manifest in `CmdletsToExport` and then only insiders would import the psm1 instead of the psd1 instead. WDYT?

SteveL-MSFT:
I think we should remove `New-MSIPackage` from exported cmdlets

tools/packaging/packaging.psm1
```diff
@@ -1173,7 +1173,7 @@ function Get-NugetSemanticVersion
     cd $RootPathOfPowerShellCheckout
     Import-Module .\build.psm1
     Start-PSBuild
-    Import-Module .\tools\packaging\packaging.psd1
+    Import-Module .\tools\packaging\packaging.psm1
```
SteveL-MSFT: this line is no longer needed since it's part of packaging.psm1 already

SteveL-MSFT (changes requested):

tools/packaging/packaging.psm1
bergmeister: I see what you mean but the idea was to provide a copy-pasteable example that works out of the box. Would it be OK to rather squash the initialization as `Import-Module .\build.psm1; Start-PSBuild; Import-Module .\tools\packaging\packaging.psm1`? Otherwise I'll of course remove the line if you insist.

tools/packaging/packaging.psm1
```diff
@@ -1163,3 +1163,148 @@ function Get-NugetSemanticVersion
...
+<#
+.Synopsis
+    Creates a Windows installer MSI package and assumes that the binaries are already built using 'Start-PSBuild'.
+    This only works on a Windows machine due to the usage of WiX.
+.EXAMPLE
+    # This example shows how to produce a Debug-x64 installer for WiX development purposes only.
+    cd $RootPathOfPowerShellCheckout
+    Import-Module .\build.psm1
+    Start-PSBuild
+    Import-Module .\tools\packaging\packaging.psm1
```
TravisEz13: Just point to import the folder as a module.

tools/packaging/packaging.psm1
SteveL-MSFT: For the comment based help, you would only see that if you already have the packaging module loaded, so it just seems weird.  I think if you want to provide an easy cut-paste, perhaps it's better to have it in https://github.com/PowerShell/PowerShell/blob/master/docs/building/windows-core.md

SteveL-MSFT (approved):

iSazonov:
@bergmeister Please resolve merge conflicts.

bergmeister:
Ok, thanks for letting me know. I will come back to it soon since it needs more time carefully extracting the diff because the conflict is quite horrible due to the move of a function that has received updates in the meantime.

iSazonov:
@bergmeister  Perhaps easy make new PR.

PR #5508, opened 2017-11-20
Fix docker tests

TravisEz13:
- remove `OpenSUSE` and `amazonlinux` as they are not community images and not published.


adityapatwardhan:
@TravisEz13  Can you add in the description why openSUSE and AmazonLinux were removed?

daxian-dbw (approved):
@adityapatwardhan AmazonLinux and openSUSE are removed because we don't publish docker images for these 2 distros for now.

daxian-dbw:
@TravisEz13 I think we should always target the mater branch, and cherrypick commits to 6.0.0 branch, right?

TravisEz13:
Marking for GA so we can test docker correctly for GA

PR #5510, opened 2017-11-20
Replace strlen with strnlen in native code

### LINKED ISSUE #4975: Remove use for strlen from native code

TravisEz13:
fixes #4975 
fix expected vs actual for easier troubleshooting

### DIFF
--- a/src/libpsl-native/test/test-createsymlink.cpp
+++ b/src/libpsl-native/test/test-createsymlink.cpp
@@ -92,7 +92,7 @@ TEST_F(CreateSymLinkTest, SymLinkToFile)
     std::string target = FollowSymLink(fileSymLink.c_str());
     char buffer[PATH_MAX];
     std::string expected = realpath(file, buffer);
-    EXPECT_EQ(target, expected);
+    EXPECT_EQ(expected, target);
 }
 
 TEST_F(CreateSymLinkTest, SymLinkToDirectory)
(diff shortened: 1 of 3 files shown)


src/libpsl-native/test/test-createsymlink.cpp
```diff
@@ -92,7 +92,7 @@ TEST_F(CreateSymLinkTest, SymLinkToFile)
     std::string target = FollowSymLink(fileSymLink.c_str());
     char buffer[PATH_MAX];
     std::string expected = realpath(file, buffer);
-    EXPECT_EQ(target, expected);
+    EXPECT_EQ(expected, target);
```
daxian-dbw: Can you please explain a bit about why you made the changes to `EXPECT_EQ`?

daxian-dbw:
Again, I think we agreed to submit PR targeting master branch, and then cherrypick commits to 6.0.0. Is that process changed?

src/libpsl-native/test/test-createsymlink.cpp
TravisEz13: The pattern for `EXPECT_EQ` is `EXPECT_EQ(<expected>,<actual>)` when you call it with the parameters reverse it makes the error very confusing.

daxian-dbw (approved):

daxian-dbw:
Note that the `libpsl` nuget package needs to be updated once this change is merged.

PR #5524, opened 2017-11-22
build: Enable building for win-arm and win-arm64

### LINKED ISSUE #5286: Pwrshplugin Config File Paths Should be Validated Before Use
The paths to CoreCLR and PSHOME read by pwrshplugin should be validated before use to protect against bad input.

A potential API to use is [PathIsDirectory](https://msdn.microsoft.com/en-us/library/windows/desktop/bb773621(v=vs.85).aspx).

It is part of api-ms-win-shlwapi-ie-l1-1-0.dll, but that API set is not present on Windows 7 (shlwapi.dll is though).

SteveL-MSFT:
Requires VS2017 latest preview to build for arm64.  Validated win-arm build on Win10 IOT with remoting, but current install script doesn't work correctly on Win10 IOT, need to update.

Also fixed issue with pwrshplugin.dll not validating path before using them.

Fix https://github.com/PowerShell/PowerShell/issues/5286


SteveL-MSFT:
Waiting on mmi.dll to be published for win-arm and win-arm64

SteveL-MSFT:
Confirmed that basic PSCore6 works on win-arm, I think we can merge this.

adityapatwardhan (approved):

SteveL-MSFT:
@adityapatwardhan hold off on merging this, looks like I need to build pwshplugin.dll for arm

build.psm1
```diff
@@ -2117,8 +2127,10 @@ function Start-CrossGen {
     $crossGenRuntime = if ($Environment.IsWindows) {
         if ($Runtime -match "-x86") {
             "win-x86"
-        } else {
+        } elseif ($Runtime -match "-x64") {
             "win-x64"
+        } elseif (!$env:PROCESSOR_ARCHITECTURE -match "arm") {
```
daxian-dbw: > !$env:PROCESSOR_ARCHITECTURE -match "arm"

`!$env:PROCESSOR_ARCHITECTURE` will be evaluated first in this case.

daxian-dbw:
Once the changes to building `pwshplugin.dll` is done, the `psrp.windows` package needs to be updated, and make sure the arm version plugin dll is signed before creating a new nuget package.

build.psm1
SteveL-MSFT: Will fix

adityapatwardhan:
@SteveL-MSFT please update here when `psrp.windows` package is updated.

daxian-dbw:
Note that PR #5617 would also require `psrp.windows` nuget package to be refreshed, so it's better to coordinate them to avoid duplicate work.

SteveL-MSFT:
Have some paths hardcoded to validate build.  Will update, but requires user to have VS2017 latest installed to build for win-arm.

SteveL-MSFT:
Verified pwrshplugin.dll works on both arm and arm64.  @adityapatwardhan can you re-review this PR as I've made substantial changes for the native compilation?

src/powershell-native/Install-PowerShellRemoting.ps1
```diff
@@ -44,38 +44,37 @@ function Register-WinRmPlugin
...
-    $fileName = "$pluginEndpointName.reg"
-
-    Set-Content -path .\$fileName "$header$regKey$keyValuePair`n"
-
-    Write-Verbose "Performing WinRM registration with: $fileName"
-    reg.exe import .\$fileName
 
-    # Clean up
-    Remove-Item .\$fileName
+    New-Item $regKey -Force
```
adityapatwardhan: `New-Item $regKey -Force  > $null `

adityapatwardhan (approved):
LGTM with minor comment.

adityapatwardhan:
@daxian-dbw Can you have a look too?

src/powershell-native/Install-PowerShellRemoting.ps1
SteveL-MSFT: Will change

daxian-dbw:
@adityapatwardhan Yes, will look tonight or first thing tomorrow morning.

build.psm1
```diff
@@ -193,51 +195,80 @@ function Start-BuildNativeWindowsBinaries {
...
+    }
+
     try {
         Push-Location "$PSScriptRoot\src\powershell-native"
 
         # setup cmakeGenerator
+        $cmakeGeneratorPlatform = ""
         if ($Arch -eq 'x86') {
-            $cmakeGenerator = 'Visual Studio 14 2015'
+            $cmakeGenerator = 'Visual Studio 15 2017'
```
daxian-dbw: Question: Is it going to be required to use VS2017 to build the plugin dll?

adityapatwardhan: @daxian-dbw The build worker image we use in AppVeyor is `Visual Studio 2015`. Do we need to update that?

adityapatwardhan: Actually we are using `Visual Studio 2017` through the AppVeyor.yml

daxian-dbw (approved):
LGTM. Left a question comment.

build.psm1
daxian-dbw: We don't build the native binaries in CI, so it shouldn't matter. Maybe we should switch to an image that doesn't have VS pre-installed.

build.psm1
```diff
@@ -172,8 +172,10 @@ function Start-BuildNativeWindowsBinaries {
         [ValidateSet('Debug', 'Release')]
         [string]$Configuration = 'Release',
 
-        [ValidateSet('x64', 'x86')]
-        [string]$Arch = 'x64'
+        [ValidateSet('x64', 'x86', 'x64_arm64', 'x64_arm')]
```
TravisEz13: `x64_arm` and `x64_arm64` seems weird to me.  Why are we combining two architectures?

TravisEz13 (approved):

build.psm1
SteveL-MSFT: I'll add a comment.  This is the form that the Visual Studio build environment expects for cross compilation.  The first is the host and the second the target.  VS supports x86_arm64, for example which I'm not supporting here.

build.psm1
SteveL-MSFT: @daxian-dbw VS2017 is required to build arm64

build.psm1
adityapatwardhan: @SteveL-MSFT Can you comment?

SteveL-MSFT:
@TravisEz13 @daxian-dbw @adityapatwardhan please review the change to the pwrshplugin config reader to address https://github.com/PowerShell/PowerShell/issues/5286

Decided to do that as part of this PR since I'm publishing a new drop to myget might as well do it once rather than twice.

SteveL-MSFT:
Published https://powershell.myget.org/feed/powershell-core/package/nuget/psrp.windows/6.0.0-rc

src/powershell-native/nativemsh/pwrshcommon/ConfigFileReader.cpp
```diff
@@ -44,30 +44,42 @@ namespace NativeMsh
...
+                    std::wstring psHomeDir = this->getValueFromLine(line, psHomeDirTag);
+                    HANDLE dirHandle = CreateFileW(psHomeDir.c_str(), GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, NULL);
+                    if (INVALID_HANDLE_VALUE != dirHandle)
                     {
-                        std::wstring::const_iterator slashIter = this->pathToPowerShellAssemblies.end();
-                        slashIter--;
-                        if (*slashIter != L'\\')
+                        CloseHandle(dirHandle);
+                        this->pathToPowerShellAssemblies = psHomeDir;
+                        if (this->pathToPowerShellAssemblies.size() > 0) // Found a match
```
daxian-dbw: It seems the `size() > 0` check should be moved to before the `CreateFileW` call.
Or maybe the `size() > 0` is not needed, because if `getValueFromLine` returns an empty string, I thinkg `CreateFileW` will fail.

src/powershell-native/nativemsh/pwrshcommon/ConfigFileReader.cpp
```diff
@@ -44,30 +44,42 @@ namespace NativeMsh
...
-                    if (this->coreClrDirectory.size() > 0) // Found a match
+                    std::wstring coreClrDir = this->getValueFromLine(line, coreClrDirTag);
+                    HANDLE dirHandle = CreateFileW(coreClrDir.c_str(), GENERIC_READ, FILE_SHARE_READ, NULL, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, NULL);
+                    if (INVALID_HANDLE_VALUE != dirHandle)
                     {
-                        std::wstring::const_iterator slashIter = this->coreClrDirectory.end();
-                        slashIter--;
-                        if (*slashIter != L'\\')
+                        CloseHandle(dirHandle);
+                        this->coreClrDirectory = this->getValueFromLine(line, coreClrDirTag);
```
daxian-dbw: Why calling `this->getValueFromLine(line, coreClrDirTag);` again instead of use `coreClrDir`?

src/powershell-native/nativemsh/pwrshcommon/ConfigFileReader.cpp
```diff
@@ -44,30 +44,42 @@ namespace NativeMsh
...
+                        if (this->coreClrDirectory.size() > 0) // Found a match
```
daxian-dbw: Same comment about the `size()` check here.

daxian-dbw (commented):
A few **non-blocking** comments.

src/powershell-native/nativemsh/pwrshcommon/ConfigFileReader.cpp
SteveL-MSFT: Mistake, should have been `coreClrDir`.  Will fix.

mirichmo (approved):

src/powershell-native/nativemsh/pwrshcommon/ConfigFileReader.cpp
SteveL-MSFT: Wrote a small program to test this and an empty string does return an invalid handle.  I can remove the size check.

src/powershell-native/nativemsh/pwrshcommon/ConfigFileReader.cpp
SteveL-MSFT: same as above

daxian-dbw:
@SteveL-MSFT Making those changes means we need to create a new `psrp.windows` package. But that work doesn't block merging this PR.

SteveL-MSFT:
@daxian-dbw yeah, I realize that.  After this is merged, I'll create one called RC2.

daxian-dbw (approved):

adityapatwardhan:
@daxian-dbw @TravisEz13 Any ideas why WIP might be blocking this?

TravisEz13:
it hasn't run, edit the title and it will run.

PR #5525, opened 2017-11-22
Fix Set-Service -Status Stopped runs failed on the service with dependencies 

### LINKED ISSUE #5517: Set-Service -Status Stopped should not prevent stopping of services without dependents only because they depend on others
`Set-Service -Status Stopped` not only prevents stopping services that other services _depend on_ (`.DependentServices`), but, curiously, also services that _themselves_ depend on others (`.ServicesDependedOn`.

For instance, the `wuauserv` service has no dependents, but itself depends on the `rpcss` service.

As an aside: `Set-Service -Status Stopped` lacks a `-Force` switch that would allow it to stop dependent services too, the way `Stop-Service` offers.

Steps to reproduce
------------------

```powershell
Set-Service -Status Stopped wuauserv
```

Expected behavior
-----------------

Given that the service has no dependents, it should be stopped.

Actual behavior
---------------

The following error occurs:

```none
Set-Service : Cannot stop service 'Windows Update (wuauserv)' because it is dependent on other services.
...
```

That the service depends on _other_ services shouldn't prevent its stopping.

Environment data
----------------

```powershell
PowerShell Core 6.0.0-rc on on Microsoft Windows 7 Enterprise  (64-bit; v6.1.7601)
Windows PowerShell v5.1.14409.1012 on Microsoft Windows 7 Enterprise  (64-bit; v6.1.7601)
```

zhenggu:
## PR Summary

Fixes #5517 

In the previous version, if one service has dependency services, it cannot be stopped by "Set-Service -Stop", this PR fixes this issue.

And also add the Parameter "-Force" to Set-Service, when it is enabled, the function will stop it's dependent services, and then stop the target's service itself. 

Without the parameter "-Force" it will raise a exception, if the service has running dependent services.

## PR Checklist

- **User-facing changes**
        - [x] Issue filed - Issue link: https://github.com/PowerShell/PowerShell-Docs/pull/2858

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell/5525)


src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -1784,24 +1793,17 @@ protected override void ProcessRecord()
...
                                 {
                                     WriteNonTerminatingError(service, null, "ServiceHasDependentServicesNoForce", ServiceResources.ServiceHasDependentServicesNoForce, ErrorCategory.InvalidOperation);
                                     return;
                                 }
 
-                                ServiceController[] servicedependedon = service.ServicesDependedOn;
-
-                                if ((servicedependedon != null) && (servicedependedon.Length > 0))
-                                {
-                                    WriteNonTerminatingError(service, null, "ServiceIsDependentOnNoForce", ServiceResources.ServiceIsDependentOnNoForce, ErrorCategory.InvalidOperation);
```
iSazonov: We should remove the resource string from RESX file  it is no longer used.

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
```diff
@@ -1784,24 +1793,17 @@ protected override void ProcessRecord()
                         {
                             if (!service.Status.Equals(ServiceControllerStatus.Stopped))
                             {
-                                //check for the dependent services as set-service dont have force parameter
+                                //check for the dependent services
```
iSazonov: Please correct the comment "// Check ...."

iSazonov (commented):
@mklement0 Could you please review the PR?

src/Microsoft.PowerShell.Commands.Management/commands/management/Service.cs
zhenze12345: Hi iSazonov, I am a new guy, can you please tell me I should create a new pull request or I can add new code into this pull request?

zhenze12345: The new code is committed.

zhenze12345:
Travis CI build failed for timeout, how to retry?

markekraus:
I'm curious if this is even the right direction. Doesn't `Stop-Service -Force` do this already? Shouldn't `Set-Service` _only_ support a a very basic stop? The `-Status` on `Set-Service` is redundant to `stop-service` which _should be_ the de facto way to stop services.

iSazonov:
We have a similar question in #5471

zhenggu:
I think it is necessary to remove the ServicesDependedOn check for it will block the normal function. and for parameter "Force", we can discuss it in other thread, if the parameter may be not acceptable.

iSazonov:
@SteveL-MSFT Could you please review the PR with #5471 on PowerShell committee?

mklement0:
Yes, `Set-Service -Status` duplicates `Start-Service` / `Resume-Service` / `Stop-Service` / `Suspend-Service` functionality, but:

* it is a one-stop solution with more of a _desired-state_ feel.
* it offers functionality that `Start-Service` doesn't have: if the target service's startup mode is currently `Disabled`, `Start-Service` invariably fails. By contrast, you can use `Set-Service -StartupType Manual -State Running` in such a case, for instance.

On the flip side, it currently doesn't support `-Force` in order to stop a service that have _dependents_ - which I think should be fixed too.

As for the issue at hand:

Preventing a service from stopping because _it_ depends on other services - as opposed  to having _other services_ depend on it (having _dependents_) - is an arbitrary and nonsensical restriction.

I presume it was introduced based on a misconception and should therefore be removed, which will `-Status Stopped` work as expected, as long as the target service doesn't have dependents.

@iSazonov: How is #5471 related to this?

iSazonov:
@mklement0 Thanks for useful comments!
> How is #5471 Start-Process: add parameter 'ExitTimeout' related to this?

Design review needed.

markekraus:
@mklement0 

> it is a one-stop solution with more of a desired-state feel.

Yea, but stopping a service is NOT setting a service. We have specific verbs for that kind of action. if there is a current bug in how it operates, that should be fixed, but adding `-force` for `set-` to only affect the `stop` is ridiculous. We have the `stop` verb, use it for advanced stop functionality. If you can get away with basic stop on `set`, so be it, that's what `stop` is for. If I could go back in time I would say the `set` verb should not have had this functionality to begin with. if it did, then the `stop` verb should not have been included.

On your second point about `start-service`, I will say that `set` is _also_ not `start` and including it originally is silly, IMO. The correction there is to add a `-force` to `start` if it is not already there that enables the service if it is not disabled and starts any services it depends on. 

I'm just trying to keep things simple here and it seems like this PR is going in the wrong direction from simple. There is nothing wrong with doing something like `set-service @params -PassThru | stop-service` or `set-service @params -PassThru | start-service`. That keeps to the _do one thing_ philosophy. increasing complexity in a command to do another thing when there is another command that does that one thing is an odd design choice. (setting aside any perceived convenience in doing so).

mklement0:
I don't why it was decided to provide duplicate functionality in `Set-Service`, but given that it was:

* fixing the bug is a must, as we agree.

* however, _not_ duplicating _all_ the functionality seems silly, if adding `-Force` is all that it takes, especially given that, as demonstrated, in other aspects it can do _more_ than one of its counterparts.  
(On a side note: status value `suspended` should probably be added as an alias for `paused`).

Adding a `-Force` switch to `Start-Service` is _not_ a solution, because there is no sensible default for the _startup type_ that invariably has to change in that case - unless you add `-StartupType` to `Start-Service` too, but that would then duplicate what `Set-Service` already does.

Yes, you can argue that the `-Status` parameter should never have been implemented, but, looking at the bigger picture, the `Set` verb has always had desired-state aspects in certain cmdlets, resulting in duplication as well; two examples:

* You can use `Set-Variable` without ever having to touch `New-Variable` - the variable is created on demand.

* You can use `Set-Content` to create a file on demand, without needing to use `New-Item`.

And, finally, another aspect in which `Set-Service -Status` provides extra (again, desired-state) functionality is that `-Status Running` either _starts_ or _resumes_ (continues) a service, as appropriate, whereas applying `Start-Service` to a suspended (paused) service results in an error.

markekraus:
> however, not duplicating all the functionality seems silly,

No, it doesn't at all. `Set-` is for setting, `Stop-`  is for stopping. if the ability to stop on `set-` goes beyond anything basic, you should use `stop-`. The ability to stop on `set-` is a convenience that arguably should not be there (and I'm not advocating for its removal).

Tell me, absent `-Status Stopped`, what purposed does `-Force` have? If I see `-Force` on `Set-Service` is expect it to allow me to force service settings. In this case, the only thing force is good for is stopping the service. We already  have `Stop-Service -Force` for that. For desired state, you are not likely going to be forcing the service to stop.  Again, this adds complexity and redundancy, and I'm arguing for simplicity.

> Adding a -Force switch to Start-Service is not a solution, because there is no sensible default for the startup type that invariably has to change in that case 

I disagree, setting it to manual, when you are "manually" forcing the service to start certainly serves as a sensible setting. But this is a discussion for another time, lets not continue it in this thread.

SteveL-MSFT:
@iSazonov committee is on taking time off this week for Thanksgiving (US holiday), will resume next week

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this.  Given that `-Status Stopped` is already there, it would make sense to support `-Force` for `Set-Service`

adityapatwardhan:
@zhenggu  Please add tests at : https://github.com/PowerShell/PowerShell/blob/master/test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1

markekraus:
@SteveL-MSFT I would like the comittee to reconsider the approval or `-Force` for this PS. re https://github.com/PowerShell/PowerShell/pull/6113#discussion_r167348028 and a few other discussions we have had here. The `-Force` in this PR not only applies to a single parameter, it also only applies to a single option for that parameter. I still believe including a `ForceStopped` or `StoppedForced` option for `-Status` is the better design choice and that we should avoid `-Force` abuse.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this.  The only applicable parameter to `-Force` is `-Status` (when value is `Stopped`) and there's no applicability to other parameters for this cmdlet.  The usage is consistent with other cmdlets like `Stop-Service`, so committee continues to recommend using `-Force` for this case.

adityapatwardhan:
@zhenggu Thanks for your contribution. Please add the PR submission template and update the appropriate fields. https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission

zhenggu:
I want to reopen this PR

iSazonov:
@zhenggu The PR is reopened. Thanks!

zhenggu:
Thanks @iSazonov

zhenggu:
any idea about CodeFactor issues? Seems no issue is caused by my code.

iSazonov:
@zhenggu You must fix style issue in new code and changed code but you shouldn't fix all CodeFactor issues.

zhenggu:
@iSazonov got it, I will take care of this

iSazonov:
I don't see style errors in your code.

zhenggu:
one question, There are many test cases uses pwsh.exe as the the binary file of the service in Set-Service.Tests.ps1, but actually we cannot run pwsh.exe as a service for it will not response any thing to the OS. As in this way, can I implement a new service by powershell to test the stop case, or I can just use wuauserv in #5517 as the service.

zhenggu:
I have added one test case to check -Status Stopped can run normally to stop wuauserv.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -329,4 +329,17 @@ Describe "Set/New/Remove-Service cmdlet tests" -Tags "Feature", "RequireAdminOnW
...
         { & $cmdlet @parameters } | Should -Throw -ErrorId $errorid
     }
+
+    It "Set-Service can run -Status Stopped to stop a service with dependencies" {
+        $servicename = "wuauserv"
+        $originalStatus = (Get-CimInstance -ClassName Win32_Service -Filter "Name='$($servicename)'").State
+        try {
+            Set-Service -Status Running $servicename
+            $script = { Set-Service -Status Stopped $servicename }
+            { & $script } | Should -Not -Throw
```
iSazonov: The test should be povitive - if we are trying to stop a service we should check that the service stoped.
Perhaps print spooler service ("Spooler") stops/starts fasters.

zhenggu: Thanks @iSazonov I will modify it.

zhenggu:
Hi @iSazonov, can you please help to review again.

zhenggu:
It will be difficult to add test cases -Force parameter, for I cannot find two services which has dependency relationship, and I can stop them both without any impact to the test windows server. Any suggestion?

iSazonov:
We could create simple custom services for the test.

zhenggu:
@iSazonov So can I put my custom service into test\powershell\Modules\Microsoft.PowerShell.Management\assets, is there any variable which points to the source code base directory in the test case?

iSazonov:
I guess it will be binary test module in `/test/tools/`. I can look WebListener as a sample.

zhenggu:
Have added one TestService written in C, but still don't know how to let the whole project call this CMakeFile.txt. And if it is called, will it be built in the same directory as pwsh? Please help to confirm them, and then I can continue. Thanks

iSazonov:
@adityapatwardhan @JamesWTruher Could you please help? Perhaps there is more simple way to create test windows services to test *-Service cmdlets.

zhenggu:
For I don't find a way to add a custom service, remove force flag and do the bug fix first.

iSazonov:
@zhenggu You could look https://www.c-sharpcorner.com/UploadFile/naresh.avari/develop-and-install-a-windows-service-in-C-Sharp/

zhenggu:
Thanks @iSazonov for your always support.

zhenggu:
@iSazonov there is one interesting thing that when I try your way, System.ServiceProcess.ServiceBase is only provided by .net framework, not .net core, according: http://dotnet.github.io/dotnet-web/api/System.ServiceProcess.html

TravisEz13 (commented):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/powershell/powershell/5525)* status: 2 of 3 files reviewed

zhenggu:
___
*[test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1, line 339 at r4](https://reviewable.io/reviews/powershell/powershell/5525#-LH_-He4I8g9HrQL5aD_-r4-339:-LHuP_F023Sv6VE-aUmT:bdsbdpm) ([raw file](https://github.com/powershell/powershell/blob/f1e071f695ced8272b3238fffb493c3c59f3ab16/test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1#L339)):*
<details><summary><i>Previously, TravisEz13 (Travis Plunk) wrote…</i></summary><blockquote>

I verified.  I believe this is the minimal set of services:
```
Status   Name               DisplayName
------   ----               -----------
Running  cexecsvc           Container Execution Agent
Running  CryptSvc           Cryptographic Services
Running  DcomLaunch         DCOM Server Process Launcher
Running  Dhcp               DHCP Client
Running  DiagTrack          Connected User Experiences and Tele...
Running  Dnscache           DNS Client
Running  EventLog           Windows Event Log
Stopped  KeyIso             CNG Key Isolation
Stopped  LanmanServer       Server
Running  LanmanWorkstation  Workstation
Stopped  lmhosts            TCP/IP NetBIOS Helper
Stopped  mpssvc             Windows Defender Firewall
Stopped  Netlogon           Netlogon
Stopped  NetSetupSvc        Network Setup Service
Running  nsi                Network Store Interface Service
Stopped  Power              Power
Running  ProfSvc            User Profile Service
Running  RpcEptMapper       RPC Endpoint Mapper
Running  RpcSs              Remote Procedure Call (RPC)
Running  SamSs              Security Accounts Manager
Running  Schedule           Task Scheduler
Stopped  seclogon           Secondary Logon
Running  SystemEventsBroker System Events Broker
Running  TimeBrokerSvc      Time Broker
Stopped  TrustedInstaller   TrustedInstaller
Running  UserManager        User Manager
Stopped  VaultSvc           Credential Manager
Stopped  WerSvc             Windows Error Reporting Service
Stopped  WinHttpAutoProx... WinHTTP Web Proxy Auto-Discovery Se...
Stopped  wisvc              Windows Insider Service
```
</blockquote></details>

have added a pre check of spooler for that.

zhenggu:
___
*[test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1, line 339 at r4](https://reviewable.io/reviews/powershell/powershell/5525#-LH_-He4I8g9HrQL5aD_-r4-339:-LHuQBF47l7Xxg8hBaVe:b-896fix) ([raw file](https://github.com/powershell/powershell/blob/f1e071f695ced8272b3238fffb493c3c59f3ab16/test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1#L339)):*
<details><summary><i>Previously, zhenggu wrote…</i></summary><blockquote>

have added a pre check of spooler for that.
</blockquote></details>

Done.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -332,15 +332,18 @@ Describe "Set/New/Remove-Service cmdlet tests" -Tags "Feature", "RequireAdminOnW
...
-        $originalStatus = (Get-Service $servicename).Status
-        try {
-            Set-Service -Status Running $servicename
-            $script = { Set-Service -Status Stopped $servicename }
-            { & $script } | Should -Not -Throw
-            (Get-Service $servicename).Status | Should -BeExactly "Stopped"
-        }
-        finally {
-            Set-Service -Status $originalStatus $servicename -ErrorAction SilentlyContinue
+        $service = Get-Service -Name $servicename -ErrorAction SilentlyContinue
```
TravisEz13: This would report the test as passed not skipped

zhenggu: I am sorry, but I don't know how to do it, could you give me any suggestion?

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -329,4 +329,21 @@ Describe "Set/New/Remove-Service cmdlet tests" -Tags "Feature", "RequireAdminOnW
         }
         { & $cmdlet @parameters } | Should -Throw -ErrorId $errorid
     }
+
+    It "Set-Service can run -Status Stopped to stop a service with dependencies" {
+        $servicename = "spooler"
+        $service = Get-Service -Name $servicename -ErrorAction SilentlyContinue
```
SteveL-MSFT: I think this code should look more like:

```powershell
$service = Get-Service -Name $servicename
$service | Should -Not -BeNullOrEmpty
$originalStatus = $service.Status
try {
  Set-Service -Status Running $servicename
  (Get-Service $servicename).Status | Should -BeExactly "Running"
  Set-Service -Status Stopped $servicename
  (Get-Service $servicename).Status | Should -BeExactly "Stopped"
}
finally {
  Set-Service -Status $originalStatus $servicename
}
```

In general, you should not use `-ErrorAction SilentlyContinue` in the tests unless a failure is acceptable and doesn't impact the test result.

SteveL-MSFT (changes requested):

iSazonov:
> there is one interesting thing that when I try your way, System.ServiceProcess.ServiceBase is only provided by .net framework, not .net core, 

The API is in WCP https://blogs.msdn.microsoft.com/dotnet/2017/11/16/announcing-the-windows-compatibility-pack-for-net-core/ . We use WCP in PowerShell Core and I think we can use it in test module too.

test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1
```diff
@@ -329,4 +329,47 @@ Describe "Set/New/Remove-Service cmdlet tests" -Tags "Feature", "RequireAdminOnW
         }
         { & $cmdlet @parameters } | Should -Throw -ErrorId $errorid
     }
+
+    It "Set-Service can run -Status Stopped to stop a service with dependencies, and can use -Force to stop a service with running dependent services" {
+        $testservicename1 = "testservice1"
```
SteveL-MSFT: The testservice setup should be in a BeforeAll{} block and the cleanup in a AfterAll{} block as I expect future tests may use the test service

SteveL-MSFT (changes requested):

zhenggu:
Seems the Travis CI build failed is not caused by my code.

SteveL-MSFT (approved):
LGTM.  Did not review the CodeFactor issues.

zhenggu:
any other thing can I do for this PR?

SteveL-MSFT:
@TravisEz13 can you continue your review?

TravisEz13:
___
*[test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1, line 358 at r12](https://reviewable.io/reviews/powershell/powershell/5525#-LJGZCwB0vC9VMkxlDvx:-LJGZCwB0vC9VMkxlDvy:byiu53u) ([raw file](https://github.com/powershell/powershell/blob/b31d813af3a8397d3c7e143eabf6d6e4894cc5e2/test/powershell/Modules/Microsoft.PowerShell.Management/Set-Service.Tests.ps1#L358)):*
> ```PowerShell
>         (Get-Service $testservicename2).Status | Should -BeExactly "Stopped"
> 
>         # stop a service with running dependent service
> ```

It seems that this should be a separate it statement
and `Set-Service -Status Running $testservicename2 | Should -Not -Throw` should be part of a `BeforeEach`

TravisEz13:
___
*[test/tools/TestService/Program.cs, line 1 at r12](https://reviewable.io/reviews/powershell/powershell/5525#-LJGZwHSEqQwrCzQkkY2:-LJGZwHSEqQwrCzQkkY3:bfu1bim) ([raw file](https://github.com/powershell/powershell/blob/b31d813af3a8397d3c7e143eabf6d6e4894cc5e2/test/tools/TestService/Program.cs#L1)):*
> ```Smalltalk
> using System.ServiceProcess;
> ```

missing copyright headers

TravisEz13:
___
*[test/tools/TestService/Service1.cs, line 1 at r12](https://reviewable.io/reviews/powershell/powershell/5525#-LJG_0o7CBWx3pl3xu5R:-LJG_0o7CBWx3pl3xu5S:bfu1bim) ([raw file](https://github.com/powershell/powershell/blob/b31d813af3a8397d3c7e143eabf6d6e4894cc5e2/test/tools/TestService/Service1.cs#L1)):*
> ```Smalltalk
> using System.ServiceProcess;
> ```

missing copyright headers

TravisEz13:
___
*[test/tools/TestService/Service1.Designer.cs, line 1 at r12](https://reviewable.io/reviews/powershell/powershell/5525#-LJG_5vm9NohwjlVOtYv:-LJG_5vm9NohwjlVOtYw:bfu1bim) ([raw file](https://github.com/powershell/powershell/blob/b31d813af3a8397d3c7e143eabf6d6e4894cc5e2/test/tools/TestService/Service1.Designer.cs#L1)):*
> ```Smalltalk
> namespace TestService
> ```

missing copyright headers

TravisEz13 (changes requested):
> Reviewed 1 of 1 files at r8, 6 of 7 files at r11, 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/powershell/powershell/5525)* status: :shipit: complete! all files reviewed

zhenggu:
@TravisEz13 Please help to review again. Thanks.

TravisEz13 (commented):
> Reviewed 4 of 4 files at r13.
> *[Reviewable](https://reviewable.io/reviews/powershell/powershell/5525)* status: :shipit: complete! all files reviewed

TravisEz13 (approved):

zhenggu:
Is this PR need another review?

anmenaga:
@adityapatwardhan This looks ready for merge.

adityapatwardhan:
@zhenggu Thanks for your contribution. Please add the PR submission template and update the appropriate fields. https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission

The template is here: https://github.com/PowerShell/PowerShell/blob/master/.github/PULL_REQUEST_TEMPLATE.md

zhenggu:
@adityapatwardhan for this PR adds a "-Force" parameter for Set-Service, but I am not sure should I modify https://github.com/PowerShell/PowerShell-Docs/blob/staging/reference/6/Microsoft.PowerShell.Management/Set-Service.md before this PR or it should be changed after it this PR is merged?

iSazonov:
@zhenggu Please open an issue or PR in PowerShell-Docs repo and add the link here.

zhenggu:
@iSazonov will add it

zhenggu:
@adityapatwardhan have added the document, and all the actions are ready, please help to review.

adityapatwardhan (approved):

adityapatwardhan:
@zhenggu  The tests are classified as `Feature` so I pushed an empty commit with `[Feature]` to run all tests.

adityapatwardhan:
@zhenggu Please have a look at the failures and for the next commit have `[Feature]` in your commit message to ensure all tests are executed. See [this](https://github.com/PowerShell/PowerShell/blob/56e50ce63279d11ffb0f391fce40091c53a9adbb/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr) for more details.

adityapatwardhan:
@zhenggu Thank you for your contribution!

zhenggu:
Thanks everyone's help for my first PR

iSazonov:
@zhenggu Thanks for your contribution and patience! I hope you will continue.

Issue #5506, opened 2017-11-20
Duplicates in ZIP distribution file.


thezim:
In [PowerShell-6.0.0-rc-win-x64.zip](https://github.com/PowerShell/PowerShell/releases/download/v6.0.0-rc/PowerShell-6.0.0-rc-win-x64.zip) there is a folder called `signed`. All files within in `signed` folder and subfolders exist at the parent level, their hashes are identical.

Is this a leftover from packaging or is there some other purpose for these files?

SteveL-MSFT:
Seems like leftover from packaging

PR #5527, opened 2017-11-22
remove `signed` folder in zip after copying signed binaries for packaging

SteveL-MSFT:
It appears that signing puts the signed binaries in a folder called [signed](https://github.com/PowerShell/PowerShell/blob/400747040300bb6a1a2dfcaec846a8fca092aadc/tools/releaseBuild/packagesigning.xml).  The current packaging copies the signed binaries over the unsigned binaries, but the signed folder is a child folder of the unsigned binaries so it gets added to the zip resulting in duplicate files that aren't used.  Fix is to remove the `signed` folder after copying the signed binaries.

Fix https://github.com/PowerShell/PowerShell/issues/5506

### DIFF
--- a/tools/packaging/packaging.psm1
+++ b/tools/packaging/packaging.psm1
@@ -448,6 +448,12 @@ function New-PSSignedBuildZip
         Copy-Item -Path $_ -Destination $destination -force
     }
 
+    # Remove `signed` folder in buildpath now that signed binaries are copied
+    if (Test-Path $BuildPath\signed)
+    {
+        Remove-Item -Recurse -Force -Path $BuildPath\signed
+    }
+
     $name = split-path -Path $BuildPath -Leaf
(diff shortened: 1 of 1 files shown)


tools/packaging/packaging.psm1
```diff
@@ -448,6 +448,12 @@ function New-PSSignedBuildZip
         Copy-Item -Path $_ -Destination $destination -force
     }
 
+    # Remove `signed` folder in buildpath now that signed binaries are copied
+    if (Test-Path $BuildPath\signed)
+    {
```
bergmeister: Maybe put the brace on the same line as this seems to be the most common style? I have opened PR #5529 to help with the style in the future.

SteveL-MSFT: I noticed both styles in this file, but it seems that for `if` statements, at least near the code I added, the braces were on a new line

TravisEz13 (approved):

PR #5529, opened 2017-11-22
Add PowerShell VSCode style settings

bergmeister:
Add the following VS Code settings to have a more consistent `PowerShell` style already at the PR stage. It is mainly based on the [K&R/OTBS](https://github.com/PoshCode/PowerShellPracticeAndStyle/issues/81#issuecomment-285835313) style, which basically means braces are on the same line and that else, catch, and other keywords are "cuddled" (e.g. `} else {`)
Personally I'd prefer rather the [Stroustrup](https://github.com/PoshCode/PowerShellPracticeAndStyle/issues/81#issuecomment-285835361) style where there is no "cuddling" but I have seen that the `K&R/OTBS` is used more often in the codebase.

````
  // Sets the codeformatting options to follow the given indent style in a way that is compatible with PowerShell syntax. For more information about the brace styles please refer to https://github.com/PoshCode/PowerShellPracticeAndStyle/issues/81.
  "powershell.codeFormatting.preset": "OTBS",
  
  // Adds a space between a keyword and its associated scriptblock expression.
  "powershell.codeFormatting.whitespaceBeforeOpenBrace": true,
  
  // Adds a space between a keyword (if, elseif, while, switch, etc) and its associated conditional expression.
  "powershell.codeFormatting.whitespaceBeforeOpenParen": true,
  
  // Adds spaces before and after an operator ('=', '+', '-', etc.).
  "powershell.codeFormatting.whitespaceAroundOperator": true,
  
  // Adds a space after a separator (',' and ';').
  "powershell.codeFormatting.whitespaceAfterSeparator": true
````

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -19,5 +19,19 @@
   "[powershell]": {
     "files.trimTrailingWhitespace": true
   },
-  "powershell.codeFormatting.preset": "Stroustrup"
+
+  // Sets the codeformatting options to follow the given indent style in a way that is compatible with PowerShell syntax. For more information about the brace styles please refer to https://github.com/PoshCode/PowerShellPracticeAndStyle/issues/81.
+  "powershell.codeFormatting.preset": "OTBS",
+  
+  // Adds a space between a keyword and its associated scriptblock expression.
+  "powershell.codeFormatting.whitespaceBeforeOpenBrace": true,
+  
+  // Adds a space between a keyword (if, elseif, while, switch, etc) and its associated conditional expression.
+  "powershell.codeFormatting.whitespaceBeforeOpenParen": true,
+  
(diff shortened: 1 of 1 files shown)


rkeithhill:
When I last touched that file I considered adding these settings but I notice that PowerShell script in this repo is not terribly consistent.  I like the idea of it being more consistent but it is also easier for folks to mix semantic changes with style changes.  I wonder if it would be better to just take the "styling" hit once and make all the script files conform to this style setup?

bergmeister:
One could apply the styling to the files since VSVode has the `Ctrl+K+F` shortcut for that. But this is certainly a decision and probably action that the PowerShell team has to make.

SteveL-MSFT:
I'm fine with the proposed setting changes.  We can apply the style universally after GA.

SteveL-MSFT (approved):

PR #5538, opened 2017-11-24
Stop running TypeGen every time

iSazonov:
Because of an error in the path, `Start-Build` cannot find the `powershell_xxx.inc` file and generates it every time. This causes the _entire_ project to be recompiled.


daxian-dbw (approved):

bergmeister:
Thanks @iSazonov , I did not spot that.

PR #5542, opened 2017-11-25
Increase version to 6.1 for PSSessionConfiguration tests

markekraus:
bump version to 6.1 for `Get-PSSessionConfiguration` and `Register-PSSessionConfiguration` tests to prevent false fails ([example](https://ci.appveyor.com/project/PowerShell/powershell/build/v6.1.0-preview.7017#L3173))

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Core/PSSessionConfiguration.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Core/PSSessionConfiguration.Tests.ps1
@@ -488,7 +488,7 @@ namespace PowershellTestConfigNamespace
 
                     $Result.Session.Name | Should be $TestSessionConfigName
                     $Result.Session.SessionType | Should be "Default"
-                    $Result.Session.PSVersion | Should be 6.0
+                    $Result.Session.PSVersion | Should be 6.1
                     $Result.Session.Enabled | Should be $true
                     $Result.Session.lang | Should be $Result.Culture
                     $Result.Session.pssessionthreadoptions | Should be $pssessionthreadoptions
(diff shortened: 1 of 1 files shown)


iSazonov:
Can we do these tests more version independent? Can we get the version from `PSVersionTable`?

markekraus:
I'm not familiar with these tests or the code they are testing, but, one fear I have about doing that kind of thing is if it is somehow referencing itself. if a corrupt version number is leaking into the PSSession, then the test will pass, even though the expected version is wrong. In that case, we should just remove the test as it wont provide any benefit. That is how we get tests that are passing even though the code is broken.

iSazonov:
I think the meaning of this test is that this property has a version the same as the current PowerShell Core version.

markekraus:
Is there a point to that test? What does it test? what is it verifying? why is it important?  Maybe it's better to remove this check. if it is just going to echo back the same version every time, is there any value in testing it? Is there any scenario where this would not have the same version? Is this version information significant in any way?

Tests are one area where I think it is OK have these kinds of literal tests. The provide more valuable information that way. If the same version just gets echoed back, testing against that sem version is no different than `$PSVersionTable.PSVersion | should be $PSVersionTable.PSVersion`

iSazonov:
`Register-PSSessionConfiguration` and `Set-PSSessionConfiguration` can have `-PSVersion`.  So currently we check that PSVersion is by default current PowerShell version. Also we should add tests that explicit `-PSVersion` works too.

markekraus:
@iSazonov so, does this check even matter then?

also, please open an issue for the missing tests.

iSazonov:
I open #5554 to fix CIs and add absent tests.

PR #5554, opened 2017-11-27
Fix PSVersion in PSSessionConfiguration tests

iSazonov:
Check that we can register and change (Set-) remote endpoint
configurations with given PSVersion.

PR #5556, opened 2017-11-27
Remove hard coded expected version string

dantraMSFT:
PSSessionConfiguration.Tests.ps1 contains tests that compare Session.PSVersion to a  hard coded version string. These tests break when the next release version doesn't match the string.

Update the test to rely on PSVersionTable.PSVersion instead of the hard code value.


markekraus:
This PR duplicates what is being done in #5554

PR #5568, opened 2017-11-28
Don't add trailing spaces to last column when using Format-Table

### LINKED ISSUE #5532: When copying output to the clipboard on macOS, output is padded to 80 columns wide with whitespace
Steps to reproduce
------------------
Install PowerShell on macOS
Open a new PowerShell instance
Do something that causes text to be output to the terminal
Copy and paste this text into somewhere else, e.g. a text editor or an email.
There is extraneous whitespace at the end of every line to pad it out to the width of my Terminal window, which is by default 80 columns.

Expected behavior
-----------------
It is expected that the text output in the Terminal will not have extra whitespace at the end of each and every line.

Actual behavior
---------------
Output is padded to the width of the Terminal window.
e.g. if you look at the output for $PSVersionTable below, copy and paste it into a text editor and you will notice that every line is 150 characters long, padded out with the space character. My Terminal window was 150 characters wide so that it could fit in the output for the OS line without truncating it.

Environment data
----------------
```
PS /Users/kai> $PSVersionTable                                                                                                                        

OS                             Darwin 16.7.0 Darwin Kernel Version 16.7.0: Thu Jun 15 17:36:27 PDT 2017; root:xnu-3789.70.16~2/RELEASE_X86_64        
```

SteveL-MSFT:
Separate fixes for the header and the data rows.  If it's the last column, don't add trailing padding.

Also fixed some Format-Table tests that were `Pending` but the issue was closed.

One side-effect is that using `out-file -width` you don't get the padding on the right most column which seems fine as the parameter is for limiting the width rather than ensuring the width.  But it is a breaking change.

Fixes https://github.com/PowerShell/PowerShell/issues/5532

### DIFF
--- a/src/System.Management.Automation/FormatAndOutput/common/TableWriter.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/TableWriter.cs
@@ -244,11 +244,18 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 return null;
 
             StringCollection[] scArray = new StringCollection[validColumnCount];
+            bool addPadding = true;
             for (int k = 0; k < scArray.Length; k++)
             {
+                // for the last column, don't pad it with trailing spaces
+                if (k == scArray.Length-1)
+                {
+                    addPadding = false;
+                }
+
                 // obtain a set of tokens for each field
                 scArray[k] = GenerateMultiLineRowField(values[validColumnArray[k]], validColumnArray[k],
-                                                                        alignment[validColumnArray[k]], ds);
+                    alignment[validColumnArray[k]], ds, addPadding);
 
                 // NOTE: the following padding operations assume that we
                 // pad with a blank (or any character that ALWAYS maps to a single screen cell
@@ -283,7 +290,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             }
 
             // add padding for the columns that are shorter
-            for (int col = 0; col < scArray.Length; col++)
+            for (int col = 0; col < scArray.Length-1; col++)
             {
                 int paddingBlanks = _si.columnInfo[validColumnArray[col]].width;
                 if (col > 0)
@@ -317,15 +324,18 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             return rows;
         }
 
-        private StringCollection GenerateMultiLineRowField(string val, int k, int alignment, DisplayCells dc)
+        private StringCollection GenerateMultiLineRowField(string val, int k, int alignment, DisplayCells dc, bool addPadding)
         {
             StringCollection sc = StringManipulationHelper.GenerateLines(dc, val,
                                         _si.columnInfo[k].width, _si.columnInfo[k].width);
-            // if length is shorter, do some padding
-            for (int col = 0; col < sc.Count; col++)
+            if (addPadding)
             {
-                if (dc.Length(sc[col]) < _si.columnInfo[k].width)
-                    sc[col] = GenerateRowField(sc[col], _si.columnInfo[k].width, alignment, dc);
+                // if length is shorter, do some padding
+                for (int col = 0; col < sc.Count; col++)
+                {
+                    if (dc.Length(sc[col]) < _si.columnInfo[k].width)
+                        sc[col] = GenerateRowField(sc[col], _si.columnInfo[k].width, alignment, dc, addPadding);
+                }
             }
             return sc;
         }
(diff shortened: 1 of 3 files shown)


iSazonov:
I think we should be more flexible - while we trim trailing whitespaces by default the `out-file -width` (and maybe `Format-Table -Width`) should be padding.

SteveL-MSFT:
I'll bring this up at the Committee meeting today.  I think it depends on what `-width` means.  Does it mean exact width or max width.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and `-width` should be maximum width only.  This is consistently with behavior today such as:

```powershell
"hello" | out-file -width 100 -path .\hello.txt
```
Also, formatting has not been guaranteed to be a specific format, so this is not a breaking change.

src/System.Management.Automation/FormatAndOutput/common/TableWriter.cs
```diff
@@ -248,11 +248,18 @@ private string[] GenerateTableRow(string[] values, int[] alignment, DisplayCells
...
 
             StringCollection[] scArray = new StringCollection[validColumnCount];
+            bool addPadding = true;
             for (int k = 0; k < scArray.Length; k++)
             {
+                // for the last column, don't pad it with trailing spaces
+                if (k == scArray.Length-1)
+                {
+                    addPadding = false;
+                }
```
iSazonov: See my comment below.

src/System.Management.Automation/FormatAndOutput/common/TableWriter.cs
```diff
@@ -339,8 +349,15 @@ private string GenerateRow(string[] values, int[] alignment, DisplayCells dc)
...
             StringBuilder sb = new StringBuilder();
 
+            bool addPadding = true;
             for (int k = 0; k < _si.columnInfo.Length; k++)
             {
+                // don't pad the last column
+                if (k == _si.columnInfo.Length -1)
+                {
+                    addPadding = false;
+                }
```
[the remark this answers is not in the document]
SteveL-MSFT: I think for outputting table which is mostly an interactive experience, I don't think performance has a big implication here.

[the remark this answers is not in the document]
iSazonov: We have CPU spikes mainly because of the reoutput of symbols.

Closed.

adityapatwardhan:
@daxian-dbw @JamesWTruher @anmenaga Can you have a look?

adityapatwardhan:
@daxian-dbw @JamesWTruher @anmenaga  ping ...

adityapatwardhan:
@daxian-dbw @JamesWTruher @anmenaga ping ...

test/powershell/Modules/Microsoft.PowerShell.Utility/Format-Table.Tests.ps1
```diff
@@ -138,42 +138,32 @@ Describe "Format-Table DRT Unit Tests" -Tags "CI" {
...
-				$result = $p|Format-Table -Property "foo","bar"|Out-String
-				$result | Should BeNullOrEmpty
+				$result = $p | Format-Table -Property "foo","bar" -Force | Out-String
+				$result.Replace(" ","").Replace([Environment]::NewLine,"") | Should BeExactly "foobar------"
 		}
 
-		#pending on issue#900
-		It "Format-Table with multiple line string for End-To-End should work" -pending{
+		It "Format-Table with multiple line string for End-To-End should work" {
 				$p = "Line1\nLine2"
```
JamesWTruher: shouldn't this be 
```
$p = "Line1`nLine2"
```
This looks like it was migrated incorrectly from c#

JamesWTruher (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/Format-Table.Tests.ps1
SteveL-MSFT: Yes, the escaped newline is wrong, but I'm actually not sure what this test is supposed to be validating.  The multiline doesn't have an effect on format-table.  Is that the test?

adityapatwardhan:
@JamesWTruher  Can you have another look?

test/powershell/Modules/Microsoft.PowerShell.Utility/Format-Table.Tests.ps1
JamesWTruher: I believe that is the test, but I'm having trouble reconciling the title of the test with the code. If we want to check multiple line strings in format-table, I see 2 tests which are interesting:
```
[pscustomobject]@{ foo = "line1`nline2"; bar = 2 }| format-table -prop foo,bar -wrap | out-string -str | where-object {$_}
foo   bar
---   ---
l1      1
l2
```
and
```
[pscustomobject]@{ foo = "line1`nline2"; bar = 2 }| format-table -prop foo,bar | out-string -str | where-object {$_}
foo   bar
---   ---
l1...   1
```
but I didn't go back and look at all the tests in this file to see if those cases are covered.
certainly it's not a blocking issue, and the test has some utility now as is

JamesWTruher (approved):

PR #5569, opened 2017-11-29
References to RunspaceConfiguration still present

Bhaal22:
RunspaceConfiguration still mentionned in RunspacePoolInternal ctor API documentation
Removed 2 obsolete test files

Those 2 files should even not compile.

 #4942

### DIFF
--- a/src/System.Management.Automation/engine/hostifaces/RunspacePoolInternal.cs
+++ b/src/System.Management.Automation/engine/hostifaces/RunspacePoolInternal.cs
@@ -63,7 +63,6 @@ namespace System.Management.Automation.Runspaces.Internal
         /// The explicit PSHost implementation.
         /// </param>
         /// <exception cref="ArgumentNullException">
-        /// RunspaceConfiguration is null.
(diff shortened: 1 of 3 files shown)


SteveL-MSFT (approved):
I don't think we currently compile and run those csharp based tests

Bhaal22:
Indeed. I do not think you compile them ... otherwise they should fail for a couple of weeks.
Initially I removed them ... then I just dediced to remove reference to RunspaceConfiguration and update the  AutomationEngine construction with the right parameters.

SteveL-MSFT:
@Bhaal22 I think the changes to the test are correct, separate from this PR, someone should understand if that test code is still relevant if so, get it running.

Bhaal22:
@SteveL-MSFT do you mean you would like to have 2 PRs?

SteveL-MSFT:
@Bhaal22 no, I think it's fine to have these changes as part of a single PR.  I'm suggesting that investigating the utility of that test code is separate from this PR.

Issue #5567, opened 2017-11-28
WMF link test intermittently fails


TravisEz13:
1. The test does not give the url which fails to verify the failure
1. I've seen similar failures twice.
https://travis-ci.org/PowerShell/PowerShell/jobs/308584873#L4450

```
TEST FAILURES
Description: Windows Installer/
Name:        Pre-Requisistes link for 'WMF 4.0' is reachable
message:
Error 404 - Not Found/**/Warning: This site requires the use of scripts, which your browser does not currently allow.See how to enable scriptsTry Microsoft EdgeA fast and secure browser that's designed for Windows 10No thanksGet startedSkip to main contentMicrosoftDownload CenterOfficeWindowsSurfaceXboxDealsSupportMoreSoftwareWindows appsOneDriveOutlookSkypeOneNotePCs &amp; Devices  PCs &amp; tabletsAccessoriesVR &amp; mixed realityMicrosoft HoloLensEntertainmentXbox gamesPC gamesWindows digital gamesMovies &amp; TVBooksBusinessMicrosoft AzureMicrosoft Dynamics 365Microsoft 365Cloud platformEnterpriseData platformDeveloper &amp; IT  Microsoft.NETVisual StudioWindows Dev CenterDocsOtherMicrosoft StoreFree downloads &amp; securityEducationStore locationsGift cardsHoliday gift guideView allNo results0Sign inDownload CenterWindowsOfficeWeb browsersDeveloper toolsXboxWindows Phone
```

daxian-dbw:
I saw the URL for WMF 5.1 failed a few times. Th tests here started to fail very recently but the URLs work fine when I open them in a web browser. The cause of test failure seems to be `"This site requires the use of scripts, which your browser does not currently allow"`.

PR #5570, opened 2017-11-29
Output url being validated for pre-req test so when test fails you can more easily troubleshoot

SteveL-MSFT:
Addresses one of the concerns in https://github.com/PowerShell/PowerShell/issues/5567


TravisEz13 (approved):

iSazonov:
Make sense to cache WMF on CIs?

iSazonov (approved):

markekraus:
@iSazonov I think the point of the test is to make sure the prerequisite links are still valid so we don't have stale links in the documentation. Caching WMF would provide no benefit there.

Are these links available in the windows installer somehow?

PR #5592, opened 2017-12-01
Fix Get-EnvironmentInformation IsCoreCLR logic

### LINKED ISSUE #5581: Get-EnvironmentInformation does not properly check for CoreCLR
The following code incorrectly reports that I am using CoreCLR on Windows PowerShell 5.1

https://github.com/PowerShell/PowerShell/blob/7dce4116c7182c99db33c5ad202743abe7cc2d88/build.psm1#L100-L117

Steps to reproduce
------------------

```powershell
Import-Module .\build.psm1
Get-EnvironmentInformation
```

Expected behavior
-----------------

```none
IsAdmin           : False
IsLinux           : False
IsCoreCLR         : False
IsWindows         : True
IsMacOS           : False
nugetPackagesRoot : C:\Users\MarkLocal\.nuget\packages
```

Actual behavior
---------------

```none
IsAdmin           : False
IsLinux           : False
IsCoreCLR         : True
IsWindows         : True
IsMacOS           : False
nugetPackagesRoot : C:\Users\MarkLocal\.nuget\packages
```

Environment data
----------------

```none
PSVersion                      5.1.15063.726
BuildVersion                   10.0.15063.726
```

markekraus:
closes #5581 

.NET 4.7.1 adds the APIs `Get-EnvironmentInformation` was using to determine if it is on CoreCLR or not. This is not `$IsCoreCLR`, but a part of the environment gathering used in the build and packaging.

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -97,19 +97,12 @@ function Get-PSCommitId
 function Get-EnvironmentInformation
 {
     $environment = @{}
-    # Use the .NET Core APIs to determine the current platform.
-    # If a runtime exception is thrown, we are on Windows PowerShell, not PowerShell Core,
-    # because System.Runtime.InteropServices.RuntimeInformation
-    # and System.Runtime.InteropServices.OSPlatform do not exist in .NET versions older than 4.7.1.
-    try {
-        $Runtime = [System.Runtime.InteropServices.RuntimeInformation]
-        $OSPlatform = [System.Runtime.InteropServices.OSPlatform]
-
-        $environment += @{'IsCoreCLR' = 'Core' -eq $PSVersionTable.PSEdition}
-        $environment += @{'IsLinux' = $Runtime::IsOSPlatform($OSPlatform::Linux)}
-        $environment += @{'IsMacOS' = $Runtime::IsOSPlatform($OSPlatform::OSX)}
-        $environment += @{'IsWindows' = $Runtime::IsOSPlatform($OSPlatform::Windows)}
-    } catch {
+    if ($PSVersionTable.ContainsKey("PSEdition") -and "Core" -eq $PSVersionTable.PSEdition) {
+        $environment += @{'IsCoreCLR' = $true}
+        $environment += @{'IsLinux' = $IsLinux}
+        $environment += @{'IsMacOS' = $IsMacOS}
+        $environment += @{'IsWindows' = $IsWindows}
+    } else {
         $environment += @{'IsCoreCLR' = $false}
         $environment += @{'IsLinux' = $false}
         $environment += @{'IsMacOS' = $false}
--- a/build.psm1
+++ b/build.psm1
@@ -97,6 +97,7 @@ function Get-PSCommitId
 function Get-EnvironmentInformation
 {
     $environment = @{}
+    # PowerShell Core will likely not be built on pre-1709 nanoserver
     if ($PSVersionTable.ContainsKey("PSEdition") -and "Core" -eq $PSVersionTable.PSEdition) {
         $environment += @{'IsCoreCLR' = $true}
         $environment += @{'IsLinux' = $IsLinux}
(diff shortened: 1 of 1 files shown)


build.psm1
```diff
@@ -100,12 +100,12 @@ function Get-EnvironmentInformation
     # Use the .NET Core APIs to determine the current platform.
     # If a runtime exception is thrown, we are on Windows PowerShell, not PowerShell Core,
```
iSazonov: The line confuse me. I think it is better to re-write all the comment.

daxian-dbw:
How about changing the check to the following?
```powershell
if ($PSVersionTable.ContainsKey("PSEdition") -and $PSVersionTable.PSEdition -eq "Core") {
    $environment += @{'IsCoreCLR' = $true}
    $environment += @{'IsLinux' = $IsLinux}
    $environment += @{'IsMacOS' = $IsMacOS}
    $environment += @{'IsWindows' = $IsWindows}
} else {
    $environment += @{'IsCoreCLR' = $false}
    $environment += @{'IsLinux' = $false}
    $environment += @{'IsMacOS' = $false}
    $environment += @{'IsWindows' = $true}
}
```

iSazonov:
We can remove `$PSVersionTable.ContainsKey("PSEdition")`.

markekraus:
@daxian-dbw I thought the entire point of this was to not rely on `pwsh` for this information as much as possible. If I break PowerShell so that it reports `$IsWindows` and `$IsLinux` all kinds of trouble will happen. The environment from this function is for building and testing. It should be set apart. ideally we would using some .NET API to figure out it this is CoreCLR or not and _not_ rely on `$PSVersionTable.PSEdition`. Any ideas?

build.psm1
markekraus: I don't understand what's confusing about it. It's accurate.

daxian-dbw:
@markekraus If you break powershell, how would you trust the type resolution in powershell still works? So `[System.Runtime.InteropServices.RuntimeInformation]` could throw a unrelated exception and make the script believe it's in a Windows PowerShell environment 😄
I think `$PSVersionTable.PSEdition` and variables like `$IsWindows`, `$IsLinux` and `$IsMacOS` are pretty static and would be quite unlikely to be changed. Also, I think the PowerShell session you use to build powershell usually is a comparatively stable version.

daxian-dbw:
@iSazonov We can remove `$PSVersionTable.ContainsKey("PSEdition")` in build.psm1, but for a script that has `Set-Strict -version latest` at the top, this check will be needed in case it's running on PSv5 or earlier versions.

daxian-dbw:
I think maybe `RuntimeInformation` was used initially because of the NanoServer PowerShell, which has `$PSVersionTable.PSEdition == Core` but doesn't have `$IsCoreCLR`, `$IsWindows`, `$IsLinux` and `$IsMacOS`. But I don't think anyone will build powershell in a NanoServer. Also, starting from RS3, the NanoServer image doesn't have the NanoServer PowerShell in it.

build.psm1
markekraus: comment completely removed as it is no longer relevant.

markekraus:
¯\\\_(ツ\)_/¯
updated with the suggestions.

build.psm1
```diff
@@ -97,19 +97,12 @@ function Get-PSCommitId
...
     $environment = @{}
-    # Use the .NET Core APIs to determine the current platform.
-    # If a runtime exception is thrown, we are on Windows PowerShell, not PowerShell Core,
-    # because System.Runtime.InteropServices.RuntimeInformation
-    # and System.Runtime.InteropServices.OSPlatform do not exist in Windows PowerShell.
-    try {
-        $Runtime = [System.Runtime.InteropServices.RuntimeInformation]
-        $OSPlatform = [System.Runtime.InteropServices.OSPlatform]
-
+    if ($PSVersionTable.ContainsKey("PSEdition") -and "Core" -eq $PSVersionTable.PSEdition) {
```
TravisEz13: maybe a comment about we don't expect anyone to build on Pre-1709 nanoserver.  Post-1709, nanoserver user powershell core, but you can probably just say nanoserver for now.

TravisEz13 (approved):

iSazonov (approved):

build.psm1
markekraus: added.

daxian-dbw (approved):

Issue #5594, opened 2017-12-01
Test-FileCatalog does not work on PSCore.


BerheAbrha:
Steps to reproduce
------------------
1) Create new file catalog for one of the modules in your system (using New-FileCatlog) 
2) Run Test-FileCatlog on the catalog created in step 1.
```powershell
New-FileCatalog -CatalogFilePath "C:\Program Files\WindowsPowerShell\Modules\Pester\3.4.0\pester.cat"  -Path "C:\Program Files\WindowsPowerShell\Modules\Pester\3.4.0"

Test-FileCatalog "C:\Program Files\WindowsPowerShell\Modules\Pester\3.4.0\pester.cat"

```

Expected behavior
-----------------

```none
PS C:\DSC-tests\DSC\Tests\PullServerClient\BVT> Test-FileCatalog "C:\Program Files\WindowsPowerShell\Modules\Pester\3.4.0\pester.cat" -Detailed


Status        : Valid
HashAlgorithm : SHA1
CatalogItems  : {[Functions\Coverage.ps1, F4022ABD45FE0FB2C3527A539F022E2AFDF16851], [Functions\Assertions\BeGreaterThan.ps1, 
                EEE62A89040021D1F3B2B86D8B6C4884AF7C8925], [bin\Pester.bat, E20AF79E7EC1B2DBA3DF7600C46E9CA613DF1A3C], [Functions\GlobalMock-A.Tests.ps1, 
                DF7F6F009A46AE9386C01A2350741C264FB4163D]...}
PathItems     : {[Build.bat, B4FEDB2F7522EB6EEA68EEA73A095EBA8E441926], [build.psake.ps1, 70DA7E6DF0664C0E791C32C31E14924B779A6938], [CHANGELOG.md, 
                86F6D42CC7389E38952ED741BFACE73A2C755119], [chocolateyInstall.ps1, 013A7B468B830C37FA51518E56933ACE54889CCC]...}
Signature     : System.Management.Automation.Signature


```

Actual behavior
---------------

```none
PS C:\InGuestPolicyx64> Test-FileCatalog "C:\Program Files\WindowsPowerShell\Modules\Pester\3.4.0\pester.cat"

<NO OUTPUT>
```

Environment data
----------------

```powershell

PS C:\InGuestPolicyx64> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.6-dirty
[...]

PR #5596, opened 2017-12-01
Fix FileCatalog cmdlets to work if -Path is not specified

SteveL-MSFT:
Remove unnecessary check for Paths.count > 0 as there is code later to use the current working directory since -Path is not a mandatory parameter.
Updated ShouldProcess to output the internal action on adding paths rather than the user action (which is the cmdlet name).

Updated tests to not specify -Path

Fix https://github.com/PowerShell/PowerShell/issues/5594

### DIFF
--- a/src/Microsoft.PowerShell.Security/security/CatalogCommands.cs
+++ b/src/Microsoft.PowerShell.Security/security/CatalogCommands.cs
@@ -95,7 +95,7 @@ namespace Microsoft.PowerShell.Commands
                 {
                     foreach (PathInfo tempPath in SessionState.Path.GetResolvedPSPathFromPSPath(p))
                     {
-                        if (ShouldProcess(tempPath.ProviderPath))
+                        if (ShouldProcess("Including path " + tempPath.ProviderPath, "", ""))
                         {
                             paths.Add(tempPath.ProviderPath);
                         }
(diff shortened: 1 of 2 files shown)


iSazonov (approved):
LGTM.

markekraus (approved):
LGTM

TravisEz13:
Reviewing for RC2 triage:  Should process should not hardcode message string

src/Microsoft.PowerShell.Security/security/CatalogCommands.cs
```diff
@@ -95,25 +95,24 @@ protected override void ProcessRecord()
...
-                {
-                    catalogFilePath = SessionState.Path.GetUnresolvedProviderPathFromPSPath(catalogFilePath);
-                }
+            // resolve catalog destination Path
+            if (!SessionState.Path.IsPSAbsolute(catalogFilePath, out drive) && !System.IO.Path.IsPathRooted(catalogFilePath))
+            {
+                catalogFilePath = SessionState.Path.GetUnresolvedProviderPathFromPSPath(catalogFilePath);
+            }
 
+            if (ShouldProcess(catalogFilePath))
```
iSazonov: Sorry for being late. Above we made `"Including path " + tempPath.ProviderPath` but here we show only path `catalogFilePath` - different style message.

SteveL-MSFT: `ShouldProcess()` here automatically shows the cmdlet name as the action, so it says something like: What If: Executing Test-FileCatalog against "catalogFilePath"

iSazonov: My question is about the difference - if an user set both `Path` and `CatalogFilePath` parameters he can get some requests for _files_ from `Path` and then different style request for catalogFilePath - he can not understand that last request is for just catalog _file_.

SteveL-MSFT: If both is specified, you see this:

```powershell
PS C:\Users\steve\test> New-FileCatalog -Path C:\Users\steve\test,c:\,c:\Users\steve -CatalogFilePath c:\users\steve\tes
t\test.cat -WhatIf
What if: Including path C:\Users\steve\test
What if: Including path C:\
What if: Including path C:\Users\steve
What if: Performing the operation "New-FileCatalog" on target "c:\users\steve\test\test.cat".
```

iSazonov: Looks good. Thanks!

PR #5617, opened 2017-12-04
SAL annotation updates & fix warnings

dantraMSFT:
Various SAL annotations were either incorrect or not backed up by code. This PR address these issues.

NOTE: By default, Start-BuildNativeWindowsBinaries does not enable code analysis and issues detected by SAL annotations are not reported.  To identify these issues, run Start-BuildNativeWindowsBinaries to generate the solution and vcxproj files.  Launch a Visual Studio developer prompt,  cd to src\powershell-native and run msbuild manually. 

```
msbuild PowerShellNative.sln /p:RunCodeAnalysis=true /p:Configuration=RelWithDebInfo /p:Platform=x64
```

The following changes address all code analysis warnings:
* GetRegistryInfo in NativeMsh had incorrect out annotations, remove __opt
* Fix handling of various out parameters - check for non-null and initialize
* Update and Align  SAL annotations for GetFormattedErrorMessage overloads
* Allow PluginException to accept NULL message.

### DIFF
--- a/src/powershell-native/nativemsh/pwrshcommon/pwrshcommon.cpp
+++ b/src/powershell-native/nativemsh/pwrshcommon/pwrshcommon.cpp
@@ -1238,18 +1238,19 @@ namespace NativeMsh
     // Note: During successful calls the following values must be freed by the caller:
     //      pwszMonadVersion
     //      pwszRuntimeVersion
-    //      pwzsRegKeyValue
+    //      pwszRegKeyValue
     //
     // The caller must take care to check to see if they must be freed during error scenarios
     // because the function may fail after allocating one or more strings.
     //
+    _Success_(return == 0)
     unsigned int PwrshCommon::GetRegistryInfo(
-        __deref_out_opt PWSTR * pwszMonadVersion,
+        __out PWSTR * pwszMonadVersion,
         __inout_ecount(1) int * lpMonadMajorVersion,
         int monadMinorVersion,
-        __deref_out_opt PWSTR * pwszRuntimeVersion,
+        __out PWSTR * pwszRuntimeVersion,
         LPCWSTR lpszRegKeyNameToRead,
-        __deref_out_opt PWSTR * pwzsRegKeyValue)
+        __out PWSTR * pwszRegKeyValue)
     {
         HKEY hEngineKey = NULL;
         bool bEngineKeyOpened = true;
@@ -1257,12 +1258,27 @@ namespace NativeMsh
         wchar_t * wszMshEngineRegKeyPath = NULL;
         LPWSTR wszFullMonadVersion = NULL;
 
+        if (NULL != pwszRegKeyValue)
+        {
+            *pwszRegKeyValue = NULL;
+        }
+
+        if (NULL != pwszRuntimeVersion)
+        {
+            *pwszRuntimeVersion = NULL;
+        }
+
+        if (NULL != pwszMonadVersion)
+        {
+            pwszMonadVersion = NULL;
+        }
+
         do
         {
             if (NULL == pwszMonadVersion ||
                 NULL == lpMonadMajorVersion ||
                 NULL == pwszRuntimeVersion ||
-                NULL == pwzsRegKeyValue)
+                NULL == pwszRegKeyValue)
             {
                 exitCode = EXIT_CODE_READ_REGISTRY_FAILURE;
                 break;
@@ -1315,7 +1331,7 @@ namespace NativeMsh
             {
                 LPCWSTR wszRequestedRegValueName = lpszRegKeyNameToRead;
                 if (!this->RegQueryREG_SZValue(hEngineKey, wszRequestedRegValueName,
-                    wszMshEngineRegKeyPath, pwzsRegKeyValue))
+                    wszMshEngineRegKeyPath, pwszRegKeyValue))
                 {
                     exitCode = EXIT_CODE_READ_REGISTRY_FAILURE;
                     break;
@@ -1361,12 +1377,13 @@ namespace NativeMsh
         return exitCode;
     }
 
+    _Success_(return == 0)
     unsigned int PwrshCommon::GetRegistryInfo(
-        __deref_out_opt PWSTR * pwszMonadVersion,
+        __out PWSTR * pwszMonadVersion,
         __inout_ecount(1) int * lpMonadMajorVersion,
         int monadMinorVersion,
-        __deref_out_opt PWSTR * pwszRuntimeVersion,
-        __deref_out_opt PWSTR * pwszConsoleHostAssemblyName)
+        __out PWSTR * pwszRuntimeVersion,
+        __out PWSTR * pwszConsoleHostAssemblyName)
     {
         return GetRegistryInfo(pwszMonadVersion,
             lpMonadMajorVersion,
--- a/src/powershell-native/nativemsh/pwrshcommon/pwrshcommon.cpp
+++ b/src/powershell-native/nativemsh/pwrshcommon/pwrshcommon.cpp
@@ -1270,7 +1270,7 @@ namespace NativeMsh
 
         if (NULL != pwszMonadVersion)
         {
-            pwszMonadVersion = NULL;
+            *pwszMonadVersion = NULL;
         }
 
         do
--- a/src/powershell-native/nativemsh/pwrshplugin/entrypoints.cpp
+++ b/src/powershell-native/nativemsh/pwrshplugin/entrypoints.cpp
@@ -31,15 +31,17 @@ LPCWSTR g_MAIN_BINARY_NAME = L"pwrshplugin.dll";
 // the caller should free pwszErrorMessage using LocalFree().
 // returns: If the function succeeds the return value is the number of CHARs stored int the output
 // buffer, excluding the terminating null character. If the function fails the return value is zero.
-#pragma prefast(push)
-#pragma prefast (disable: 28196)
-DWORD GetFormattedErrorMessage(__deref_out PWSTR * pwszErrorMessage, DWORD dwMessageId, va_list* arguments)
-#pragma prefast(pop)
+_Success_(return >= 0)
+DWORD GetFormattedErrorMessage(__out PWSTR * pwszErrorMessage, DWORD dwMessageId, va_list* arguments)
 {
     DWORD dwLength = 0;
 
     do
     {
+        if (NULL == pwszErrorMessage)
+        {
+            break;
+        }
         *pwszErrorMessage = NULL;
 
         if (NULL == g_hResourceInstance)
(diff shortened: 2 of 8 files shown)


daxian-dbw:
Note that the `psrp.windows` nuget package needs to be regenerated. Make sure that the binaries are signed before using them to create the nuget package.

adityapatwardhan (approved):

TravisEz13:
@dantraMSFT  Can you please investigate test failures?

dantraMSFT:
@TravisEz13: The failures in these two runs are unrelated to the change; they are timing issues.  I've created https://github.com/PowerShell/PowerShell/issues/5627 to track the issue.

TravisEz13:
I've restarted windows and mac CI.

dantraMSFT:
@daxian-dbw: I don't expect to be able to get the nuget package signed and published before I leave for vacation (EOD today).

daxian-dbw:
@SteveL-MSFT @TravisEz13 Can you please reivew this PR to see if it's needed for RC2 or GA?

src/powershell-native/nativemsh/pwrshcommon/pwrshcommon.cpp
```diff
@@ -1238,31 +1238,47 @@ namespace NativeMsh
...
+        }
+
+        if (NULL != pwszRuntimeVersion)
+        {
+            *pwszRuntimeVersion = NULL;
+        }
+
+        if (NULL != pwszMonadVersion)
+        {
+            pwszMonadVersion = NULL;
```
daxian-dbw: The two check above are assigning `NULL` to `*<pointer>` while this one is not. Is it intentional?

daxian-dbw (dismissed):

src/powershell-native/nativemsh/pwrshplugin/entrypoints.cpp
```diff
@@ -299,6 +301,7 @@ unsigned int GetCLRVersionForPSVersion(int iPSMajorVersion,
 
 DWORD ReportOperationComplete(WSMAN_PLUGIN_REQUEST *requestDetails, DWORD errorCode)
 {
+    WCHAR szMessage[256] = L"\0";
```
mirichmo: Why did you make this code change? It looks unnecessary and potentially error-prone.

It is incorrect to provide a fixed-length buffer for the error string. The string is read from pwrshplugin.dll and may be of any length. GetFormattedErrorMessage correctly handles the arbitrary length, but this change will potentially truncate it.

src/powershell-native/nativemsh/pwrshplugin/entrypoints.cpp
```diff
@@ -307,11 +310,14 @@ DWORD ReportOperationComplete(WSMAN_PLUGIN_REQUEST *requestDetails, DWORD errorC
...
-    GetFormattedErrorMessage(&pwszErrorMessage, errorCode);
+    if (GetFormattedErrorMessage(&pwszErrorMessage, errorCode) == 0)
+    {
+        swprintf_s(szMessage,_countof(szMessage), L"ReportOperationComplete: %d", errorCode);
+    }
 
     result = WSManPluginOperationComplete(requestDetails, 0, errorCode, pwszErrorMessage);
 
-    if (NULL != pwszErrorMessage)
+    if (NULL != pwszErrorMessage && pwszErrorMessage != szMessage)
```
mirichmo: This change will conditionally leak memory. If pwszErrorMessage has a value, it was allocated by GetFormattedErrorMessage and must be freed (via delete[]).

src/powershell-native/nativemsh/pwrshplugin/pwrshclrhost.cpp
```diff
@@ -270,7 +270,10 @@ unsigned int PowerShellClrWorker::LoadWorkerCallbackPtrs(
     {
         PWSTR msg = NULL;
         exitCode = g_MANAGED_METHOD_RESOLUTION_FAILED;
-        GetFormattedErrorMessage(&msg, exitCode);
+        /* NOTE: dont' use a string literal for a fallback; PlugInException expects msg to allocated
```
mirichmo: This comment seems unnecessary. The behavior is documented in the class definition.

src/powershell-native/nativemsh/pwrshplugin/pwrshplugin.h
```diff
@@ -490,6 +490,11 @@ class PwrshPlugInMediator
         {
             iMajorVersion = iVPSMajorVersion;
 
+            if (NULL != wszMgdPlugInFileName)
```
mirichmo: This already occurs on line 671. It seems unnecessary to do it twice. Did the scanner require it?

src/powershell-native/nativemsh/pwrshplugin/pwrshplugin.h
```diff
@@ -768,7 +780,9 @@ class PwrshPlugInMediator
             else
             {
                 PWSTR msg = NULL;
-                GetFormattedErrorMessage(&msg, g_OPTION_SET_NOT_COMPLY, g_BUILD_VERSION);
+                (void) GetFormattedErrorMessage(&msg, g_OPTION_SET_NOT_COMPLY, g_BUILD_VERSION);
+                /* NOTE: Passing possible NULL msg. PlugInExpecption requires allocated
+                   or NULL. Literal strings are not supported. */
```
mirichmo: Comment seems unnecessary here too

mirichmo (changes requested):

src/powershell-native/nativemsh/pwrshplugin/entrypoints.cpp
dantraMSFT: Fixed

src/powershell-native/nativemsh/pwrshplugin/pwrshplugin.h
dantraMSFT: This check is added because wszMgdPlugInFileName was not being set in all paths; specifically error paths.

src/powershell-native/nativemsh/pwrshplugin/entrypoints.cpp
dantraMSFT: There is an assignment missing, which I'll add in the next commit; szMessage is used when GetFormmatedMessage fails.

src/powershell-native/nativemsh/pwrshplugin/pwrshclrhost.cpp
dantraMSFT: I added this comment because the constructor annotations were wrong and, since the return value of GetFormmatedErrorMessage was being ignored, I added the comment to document the behavior as intentional.

src/powershell-native/nativemsh/pwrshcommon/pwrshcommon.cpp
dantraMSFT: Fixed

src/powershell-native/nativemsh/pwrshplugin/entrypoints.cpp
mirichmo: This has not been resolved yet.

src/powershell-native/nativemsh/pwrshplugin/entrypoints.cpp
mirichmo: The assignment does not address my comment. This change is fundamentally flawed because it will potentially truncate the string. I do not see a benefit over the original code.

mirichmo (changes requested):

dantraMSFT:
@mirichmo: If you take a look at the code again, you'll see that szMessage is to populate a placeholder error when when GetFormattedErrorMessage fails. It does not truncate an arbitrary error string.

As far as a possible leak, there isn't one.

When GetFormattedErrorMessage returns zero; pwszErrorMessage is null and, szMessage is used instead; thus, the delete call only frees memory when pwszErrorMessage is non-null and not szMessage.

Are you seeing something that I have missed?

src/powershell-native/nativemsh/pwrshplugin/entrypoints.cpp
SteveL-MSFT: After talking to Dan, `szMessage` contains a default error message in the case where the actual error message can't be allocated.  So `szMessage` should be called something else like `szFallbackErrorMessage` making this more clear.  Second, the 256 bytes is arbitrary.  Since we know the max length based on line 315 below, we should only allocate that much.

src/powershell-native/nativemsh/pwrshplugin/entrypoints.cpp
```diff
@@ -307,11 +310,15 @@ DWORD ReportOperationComplete(WSMAN_PLUGIN_REQUEST *requestDetails, DWORD errorC
 
     DWORD result = EXIT_CODE_SUCCESS;
     PWSTR pwszErrorMessage = NULL;
-    GetFormattedErrorMessage(&pwszErrorMessage, errorCode);
+    if (GetFormattedErrorMessage(&pwszErrorMessage, errorCode) == 0)
+    {
+        swprintf_s(szMessage,_countof(szMessage), L"ReportOperationComplete: %d", errorCode);
+        pwszErrorMessage = szMessage;
```
SteveL-MSFT: Perhaps for readability of the code, it would be better to change this to call WSManPluginOperationComplete with szMessage and else use pwszErrorMessage?

This would also remove the change on line 321 below.

SteveL-MSFT (changes requested):

src/powershell-native/nativemsh/pwrshplugin/entrypoints.cpp
dantraMSFT: Fixed.

src/powershell-native/nativemsh/pwrshplugin/entrypoints.cpp
dantraMSFT: Fixed.
Note that I left the stack buffer at 64.

dantraMSFT:
@mirichmo, @SteveL-MSFT  
I've updated ReportOperationComplete to clarify the stack buffer usage.
After another review, I found that GetFormattedErrorMessage was still not reliably returning zero for all error paths.  I've rewritten it to be explicit in each case and avoid the nested If statements by leveraging the existing do/while(false)  construct and moving the localfree call for wszSystemErrorMessage to the end of the function.

SteveL-MSFT (approved):

mirichmo (approved):

TravisEz13:
@dantraMSFT Can you open an issue to publish a new package before I merge this so we don't lose the work?

dantraMSFT:
@TravisEz13 See https://github.com/PowerShell/PowerShell/issues/5802

PR #5619, opened 2017-12-04
Run tests for Windows Installer only on Windows

JamesWTruher:
We don't need to run these tests on all platforms, especially as the endpoint has intermittent failures.


adityapatwardhan (approved):

markekraus (approved):
LGTM

mirichmo (approved):

adityapatwardhan:
@JamesWTruher Please run with `[Feature]` so the modified test is executed.

adityapatwardhan:
`[Feature]` tag does not run `scenario` tests. Merging this.

Issue #5458, opened 2017-11-15
Unbundle Pester from PSCore pkg


SteveL-MSFT:
We originally decided to ship Pester inbox in Windows to help bootstrap the ecosystem with a common test framework.  I believe we’ve achieved that goal.  We should stop shipping Pester with PSCore6 and rely on users installing it from PSGallery.  Long term we should move other modules we ship/own to be available on PSGallery and compatible with a range of PSCore6 versions.  This should happen after we update our tests to work on latest Pester

rkeithhill:
I'd agree with this except for `PackageManagement` and `PowerShellGet`.  On various PowerShell module open source projects (like posh-git), it is very convenient to tell folks that all they need to do is run `Install-Module posh-git -Scope CurrentUser` to install the module.  That simplicity is watered down a lot if you have to put a bunch of verbiage around bootstrapping the package management system first.

SteveL-MSFT:
PackageManagement and PSGet are needed to install from PSGallery so they will, of course, be bundled :)

markekraus:
There was a suggestion somewhere that we also split the web cmdlets and JSON cmdltes out. I somewhat agree with that one on the basis that the web cmdlets are browser-ish and may need a more rapid iteration cycle to keep up with the ever changing threat dynamics of the interwebs and increasing feature support in CoreFX for non-windows platforms. But I wouldn't want to have PowerShell Core completely void of a download mechanism, though. So would a very simple `Invoke-WebDownload` cmdlet that only has a very very basic feature set with only the ability to save to file (no content processing) with separation of some (or all) of the utility cmdltes make sense?

iSazonov:
What will be the service model for PowerShell Core? 
If something like VS Code - we'll publish new stable release every month - the need for multiple packets disappears.
VS Code has a manager for extentions I guess it's overkill for PowerShell..

For example, the [Exchange Server team](https://blogs.technet.microsoft.com/exchange/2013/02/08/servicing-exchange-2013/) releases new version every 3 months (quarter) and the build expires 6 months later. Also security patches can be released for previous non-expired builds.

dragonwolf83:
Unbundling from PowerShell Core doesn't have to mean not including as a post-install script. After install, a separate command could be run to install key packages straight from PowerShell Gallery. There is precedence for this with `Update-Help`. For those requiring offline installations, an alternate method may need to be figured out.

[...]

PR #5369, opened 2017-11-07
Change $OutputEncoding to be utf8 without BOM rather than ASCII

JamesWTruher:
address issue https://github.com/PowerShell/PowerShell/issues/4681

PR #5620, opened 2017-12-04
Add the breaking changes doc

maertendMSFT:
## PR Checklist

### DIFF
--- a/docs/BREAKINGCHANGES.md
+++ b/docs/BREAKINGCHANGES.md
@@ -30,25 +30,25 @@ Today, this breaks the `ActiveDirectory` and `DnsClient` modules in Windows and 
 Due to the complexity of supporting two sets of WMI-based modules,
 we removed the WMI v1 cmdlets from PowerShell Core:
 
-* Get-WmiObject
-* Invoke-WmiMethod
-* Register-WmiEvent
-* Set-WmiInstance
+* `Get-WmiObject`
+* `Invoke-WmiMethod`
+* `Register-WmiEvent`
+* `Set-WmiInstance`
 
 Instead, we recommend that you the use the CIM (aka WMI v2) cmdlets which provide the same functionality with new functionality and a redesigned syntax:
 
-* Get-CimAssociatedInstance
-* Get-CimClass
-* Get-CimInstance
-* Get-CimSession
-* Invoke-CimMethod
-* New-CimInstance
-* New-CimSession
-* New-CimSessionOption
-* Register-CimIndicationEvent
-* Remove-CimInstance
-* Remove-CimSession
-* Set-CimInstance
+* `Get-CimAssociatedInstance`
+* `Get-CimClass`
+* `Get-CimInstance`
+* `Get-CimSession`
+* `Invoke-CimMethod`
+* `New-CimInstance`
+* `New-CimSession`
+* `New-CimSessionOption`
+* `Register-CimIndicationEvent`
+* `Remove-CimInstance`
+* `Remove-CimSession`
+* `Set-CimInstance`
 
 ### Microsoft.PowerShell.LocalAccounts
 
@@ -101,7 +101,7 @@ When an API returns just `null`, Invoke-RestMethod was serializing this as the s
 
 ### Remove `-ComputerName` from `\*-Computer` cmdlets [#5277](https://github.com/PowerShell/PowerShell/issues/5277)
 
-Due to issues with RPC remoting in CoreFx (particularly on non-Windows platforms) and ensuring a consistent remoting experience in PowerShell,
+Due to issues with RPC remoting in CoreFX (particularly on non-Windows platforms) and ensuring a consistent remoting experience in PowerShell,
 the `-ComputerName` parameter was removed from the `\*-Computer` cmdlets.
 Use `Invoke-Command` instead as the way to execute cmdlets remotely.
 
@@ -199,7 +199,7 @@ Due to unsupported APIs, the `LocalAccounts` module and the `Counter` cmdlets in
 
 ### Executing powershell script with bool parameter does not work [#4036](https://github.com/PowerShell/PowerShell/issues/4036)
 
-Previously, using powershell.exe (now pwsh.exe) to execute a PowerShell script using `-File` provided no way to pass $true/$false as parameter values.
+Previously, using powershell.exe (now `pwsh.exe`) to execute a PowerShell script using `-File` provided no way to pass $true/$false as parameter values.
 Support for $true/$false as parsed values to parameters was added.
 Switch values are also supported as currently documented syntax doesn't work.
 
@@ -243,7 +243,11 @@ Going forward all of the output will be displayed with a true representation of 
 
 ### PowerShell as a default shell doesn't work with script command [#3319](https://github.com/PowerShell/PowerShell/issues/3319)
 
-On Unix, it is a convention for shells to accept `-i` for an interactive shell and many tools expect this behavior (`script` for example, and when setting PowerShell as the default shell) and calls the shell with the `-i` switch. This change is breaking in that `-i` previously could be used as short hand to match `-inputformat`, which now needs to be `-in`.
+On Unix, it is a convention for shells to accept `-i` for an interactive shell and many tools expect this behavior
+(`script` for example, and when setting PowerShell as the default shell)
+and calls the shell with the `-i` switch.
+This change is breaking in that `-i` previously could be used as short hand to match `-inputformat`,
+which now needs to be `-in`.
 
 ### Typo fix in Get-ComputerInfo property name [#3167](https://github.com/PowerShell/PowerShell/issues/3167)
 
--- a/docs/BREAKINGCHANGES.md
+++ b/docs/BREAKINGCHANGES.md
@@ -1,0 +1,324 @@
+### Skip null-element check for collections with a value-type element type [5432](https://github.com/PowerShell/PowerShell/issues/5432)
+TODO
+
+
+### Remove AddTypeCommandBase class [5407](https://github.com/PowerShell/PowerShell/issues/5407)
+TODO
+
+
+### Add error on Legacy Credential over non-HTTPS for Web Cmdlets [5402](https://github.com/PowerShell/PowerShell/issues/5402)
+TODO
+
+
+### Change $OutputEncoding to use iso-8859-1 encoding rather than ASCII [5361](https://github.com/PowerShell/PowerShell/issues/5361)
+TODO
+
+
+### Fix Single Value JSON null in Invoke-RestMethod [5338](https://github.com/PowerShell/PowerShell/issues/5338)
+TODO
+
+
+### Invoke-RestMethod doesn't return useful info when no data is returned. [5320](https://github.com/PowerShell/PowerShell/issues/5320)
+TODO
+
+
+### Remove DCOM support from *-Computer cmdlets (replacement for 5152) [5277](https://github.com/PowerShell/PowerShell/issues/5277)
+TODO
+
+
+### Remove AllScope from most default aliases [5268](https://github.com/PowerShell/PowerShell/issues/5268)
+TODO
+
+
+### Fix performance issues in Add-Type [5243](https://github.com/PowerShell/PowerShell/issues/5243)
+TODO
+
+
+### ConvertFrom-Json silently throws away duplicated keys with identical casing [5199](https://github.com/PowerShell/PowerShell/issues/5199)
+TODO
+
+
+### fix get-item -literalpath a*b if `a*b` doesn't actually exist to return error [5197](https://github.com/PowerShell/PowerShell/issues/5197)
+TODO
+
+
+### don't insert linebreaks to output (except for tables) [5193](https://github.com/PowerShell/PowerShell/issues/5193)
+TODO
+
+
+### Don't wrap stderr as ErrorRecord [5190](https://github.com/PowerShell/PowerShell/issues/5190)
+TODO
+
+
+### Passing a ScriptBlock (in syntax only) to native commands produces surprising results [5187](https://github.com/PowerShell/PowerShell/issues/5187)
+TODO
+
+
+### Make -NoTypeInformation Default on Export-Csv and ConvertTo-Csv [5164](https://github.com/PowerShell/PowerShell/issues/5164)
+TODO
+
+
+### Remove DCOM support from *-Computer cmdlets [5152](https://github.com/PowerShell/PowerShell/issues/5152)
+TODO
+
+
+### Import-Csv Should apply PSTypeNames uppon import when Type Information is present in the CSV [5134](https://github.com/PowerShell/PowerShell/issues/5134)
+TODO
+
+
+### -NoTypeInformation should be default on Export-Csv [5131](https://github.com/PowerShell/PowerShell/issues/5131)
+TODO
+
+
+### allow pscustomobject to have property with empty string name [5129](https://github.com/PowerShell/PowerShell/issues/5129)
+TODO
+
+
+### StdErr/ErrorRecord should not be modified on output [5127](https://github.com/PowerShell/PowerShell/issues/5127)
+TODO
+
+
+### -Verbose should not override $ErrorActionPreference [5113](https://github.com/PowerShell/PowerShell/issues/5113)
+TODO
+
+
+### Web Cmdlets should warn when  legacy -Credential is sent over unencrypted connections [5112](https://github.com/PowerShell/PowerShell/issues/5112)
+TODO
+
+
+### Use InvariantCulture for implicit argument conversion [5106](https://github.com/PowerShell/PowerShell/issues/5106)
+TODO
+
+
+### Rename powershell.exe to pwsh.exe [5101](https://github.com/PowerShell/PowerShell/issues/5101)
+TODO
+
+
+### Remove '-ComputerName' from 'Get/Set/Remove-Service' [5094](https://github.com/PowerShell/PowerShell/issues/5094)
+TODO
+
+
+### Unify cmdlets with parameter 'Encoding' to be of type System.Text.Encoding [5080](https://github.com/PowerShell/PowerShell/issues/5080)
+TODO
+
+
+### Add better error message for empty and null -UFormat arg [5055](https://github.com/PowerShell/PowerShell/issues/5055)
+TODO
+
+
+### Clean up console code [4995](https://github.com/PowerShell/PowerShell/issues/4995)
+TODO
+
+
+### correct passing of parameter so args show up as args instead of input [4963](https://github.com/PowerShell/PowerShell/issues/4963)
+TODO
+
+
+### remove -computer parameter from get-process [4960](https://github.com/PowerShell/PowerShell/issues/4960)
+TODO
+
+
+### removed RunspaceConfiguration support [4942](https://github.com/PowerShell/PowerShell/issues/4942)
+TODO
+
+
+### Add Tests for *-computer cmdlets [4926](https://github.com/PowerShell/PowerShell/issues/4926)
+TODO
+
+
+### CommandInvocationIntrinsics.InvokeScript bind arguments to $input instead of $args [4923](https://github.com/PowerShell/PowerShell/issues/4923)
+TODO
+
+
+### remove unsupported -showwindow switch from get-help [4903](https://github.com/PowerShell/PowerShell/issues/4903)
+TODO
+
+
+### allow * to be used in registry path for remove-item [4866](https://github.com/PowerShell/PowerShell/issues/4866)
+TODO
+
+
+### fix set-service failing test [4802](https://github.com/PowerShell/PowerShell/issues/4802)
+TODO
+
+
+### Enhance the -split operator with negative maximum token counts to split from the end [4765](https://github.com/PowerShell/PowerShell/issues/4765)
+TODO
+
+
+### Rename $IsOSX to $IsMacOS [4757](https://github.com/PowerShell/PowerShell/issues/4757)
+TODO
+
+
+### Rename $IsOSX to $IsmacOS [4700](https://github.com/PowerShell/PowerShell/issues/4700)
+TODO
+
+
+### New-TemporaryFile should throw terminating error [4634](https://github.com/PowerShell/PowerShell/issues/4634)
+TODO
+
+
+### S.M.A.PowerShell.HadErrors and $? return false positives when errors are suppressed [4613](https://github.com/PowerShell/PowerShell/issues/4613)
+TODO
+
+
+### Make error message consistent when invalid script is passed to -File, better error when passed ambiguous arg [4573](https://github.com/PowerShell/PowerShell/issues/4573)
+TODO
+
+
+### Fix Get-Date -UFormat '%V' week number output [4508](https://github.com/PowerShell/PowerShell/issues/4508)
+TODO
+
+
+### Improper usage of $input as a function parameter is silently ignored [4391](https://github.com/PowerShell/PowerShell/issues/4391)
+TODO
+
+
+### Due to the use of unsupported APIs, we must remove the Counter CmdLets in the Diagnostics Module until a better solution is found. [4303](https://github.com/PowerShell/PowerShell/issues/4303)
+TODO
+
+
+###  Due to the use of unsupported APIs, we must remove the LocalAccounts module until a better solution is found. [4302](https://github.com/PowerShell/PowerShell/issues/4302)
+TODO
+
+
+### Remove UTC and SQM telemetry code [4190](https://github.com/PowerShell/PowerShell/issues/4190)
+TODO
+
+
+### Unify file encoding when a cmdlet creates a file [4119](https://github.com/PowerShell/PowerShell/issues/4119)
+TODO
+
+
+### Executing powershell script with bool parameter doesnt work [4036](https://github.com/PowerShell/PowerShell/issues/4036)
+TODO
+
+
+### Remove ClrVersion property from $PSVersionTable [4027](https://github.com/PowerShell/PowerShell/issues/4027)
+TODO
+
+
+### Command line arguments with a dollar sign [4024](https://github.com/PowerShell/PowerShell/issues/4024)
+TODO
+
+
+### Change positional parameter for powershell.exe from -Command to -File [4019](https://github.com/PowerShell/PowerShell/issues/4019)
+TODO
+
+
+### Implement Unicode escape parsing [3958](https://github.com/PowerShell/PowerShell/issues/3958)
+TODO
+
+
+### Change New-ModuleManifest encoding to UTF8NoBOM on non-Windows platforms [3940](https://github.com/PowerShell/PowerShell/issues/3940)
+TODO
+
+
+### Fix 'Get-Content -Delimiter' to not include the delimiter in the returned lines [3808](https://github.com/PowerShell/PowerShell/issues/3808)
+TODO
+
+
+### Prevent Get-ChildItem from recursing into symlinks (#1875). [3780](https://github.com/PowerShell/PowerShell/issues/3780)
+TODO
+
+
+### COM objects are not enumerating properly [3775](https://github.com/PowerShell/PowerShell/issues/3775)
+TODO
+
+
+### System.IO.DirectoryInfo and System.IO.FileInfo instances output by Get-Item / Get-ChildItem bind to -Path rather than -LiteralPath and sometimes by filename only [3772](https://github.com/PowerShell/PowerShell/issues/3772)
+TODO
+
+
+### Why do handled exceptions show in ErrorVariable? [3768](https://github.com/PowerShell/PowerShell/issues/3768)
+TODO
+
+
+### Align PowerShell's CLI / startup behavior with POSIX-like shells such as Bash - command-line arguments [3743](https://github.com/PowerShell/PowerShell/issues/3743)
+TODO
+
+
+### Get-Content -Delimiter unexpectedly keeps the delimiter in the lines returned [3706](https://github.com/PowerShell/PowerShell/issues/3706)
+TODO
+
+
+### Should we detect/Deny using $_ as a user defined variable? [3695](https://github.com/PowerShell/PowerShell/issues/3695)
+TODO
+
+
+### Added -i swtich to powershell for interactive [3558](https://github.com/PowerShell/PowerShell/issues/3558)
+TODO
+
+
+### PowerShell use different logic, when converting passed value to parameter type, for compiled and script cmdlets [3348](https://github.com/PowerShell/PowerShell/issues/3348)
+TODO
+
+
+### Implement Format-Hex in C# [3320](https://github.com/PowerShell/PowerShell/issues/3320)
+TODO
+
+
+### PowerShell as a default shell doesn't work with script command [3319](https://github.com/PowerShell/PowerShell/issues/3319)
+TODO
+
+
+### Completions for environment variables differ between platforms [3227](https://github.com/PowerShell/PowerShell/issues/3227)
+TODO
+
+
+### Get-ComputerInfo - fix typo in property name [3167](https://github.com/PowerShell/PowerShell/issues/3167)
+TODO
+
+
+### -OutVariable doesn't unwrap single-object output and creates [System.Collections.ArrayList] values rather than [System.Object[]] [3154](https://github.com/PowerShell/PowerShell/issues/3154)
+TODO
+
+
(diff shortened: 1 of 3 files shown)


TravisEz13:
You have spelling issues to fix:  https://travis-ci.org/PowerShell/PowerShell/jobs/311493778#L3284

TravisEz13:
Also markdown issues,  If you install markdownlint in vs code it will detect these: 
https://ci.appveyor.com/project/PowerShell/powershell/build/v6.1.0-preview.7128#L84

maertendMSFT:
@TravisEz13, thanks for point these out.  I am aware of the issues, the doc is not ready to merge.  This is a working PR

iSazonov:
I see here a problem - not all PRs with breaking changes was labeled.
Currently I am working on Add-Type cmdlet and discover that it was changed a lot, contains bugs and other issues. The important thing is, if you compare the version, it's a lot different than Windows PowerShell cmdlet, but it's not reflected here. 
I believe we have many more cases like this.
In perfect case, we would have to check all the cmdlets and all the parameters.

TravisEz13:
I updated the description with the current PR template.  Please fill it out when the PR is ready to be reviewed.

joeyaiello:
Need to add removal of Pester (#5458) and cmdlet/module availability per https://github.com/PowerShell/PowerShell/blob/2be13a623eae2583604ab51449eefc10a8fd9c12/test/powershell/engine/Basic/DefaultCommands.Tests.ps1

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
 
 
-### Add error on Legacy Credential over non-HTTPS for Web Cmdlets [5402](https://github.com/PowerShell/PowerShell/issues/5402)
-TODO
-
-
 ### Change $OutputEncoding to use iso-8859-1 encoding rather than ASCII [5361](https://github.com/PowerShell/PowerShell/issues/5361)
-TODO
-
+This is marked as a breaking change because the output of the pipeline is not altered as it was previously, anyone relying on ascii encoding (especially for extended ascii sequences) will no longer have that behavior.
```
SteveL-MSFT: This description isn't correct.  Also, I don't think we need to mention "breaking change" in the description since the doc itself is about breaking changes.  The PR linked to is not valid as we closed that one and decided to use UTF-8 instead of ISO-8859-1.  Correct PR is https://github.com/PowerShell/PowerShell/pull/5369.  I suggest using:

```none
The default encoding of output was previously ASCII (7-bit) which in some cases resulted in incorrect
alteration of the output.  Change is to default to UTF-8 NoBOM which preserves unicode output with an
encoding supported by most tools and operating systems.
```

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
-TODO
+TODO: This is closed in favor of a new PR which simply changes the output encoding to utf8nobom, the new PR is not referenced.  Need to find and see if it is still a breaking change.
 
 
 ### Invoke-RestMethod doesn't return useful info when no data is returned. [5320](https://github.com/PowerShell/PowerShell/issues/5320)
-TODO
+When an API returns just `null`, Invoke-RestMethod was serializing this as the string `"null"` instead of `$null`.  This change fixes the logic in `Invoke-RestMethod` to properly serialize a valid single value JSON `null` literal as `$null`.
 
 
 ### Remove DCOM support from *-Computer cmdlets (replacement for 5152) [5277](https://github.com/PowerShell/PowerShell/issues/5277)
```
SteveL-MSFT: You need to escape the asterisk otherwise markdown things you're emphasizing everything:

```markdown
\*-Computer
```

Applies everywhere below.

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
-TODO
+DCOM is not supported in corefx, therefore the *-Computer cmdlets were removed.
```
SteveL-MSFT: The \*-Computer cmdlets are still there on Windows (since they rely on WMI), but the `-ComputerName` parameter was removed since it relied on RPC (DotNet remoting).

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
 
 
 ### Remove AllScope from most default aliases [5268](https://github.com/PowerShell/PowerShell/issues/5268)
-TODO
-
-
-### Fix performance issues in Add-Type [5243](https://github.com/PowerShell/PowerShell/issues/5243)
-TODO
+To speed up spope creation, AllScope was removed from most default aliases.  AllScope was left for a few frequently used aliases where the lookup was faster.
```
SteveL-MSFT: Should we have different sections for impactful `Breaking Changes` and technical `Breaking Changes`?  This one falls under the latter where it shouldn't affect anyone.

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
 
 
 ### ConvertFrom-Json silently throws away duplicated keys with identical casing [5199](https://github.com/PowerShell/PowerShell/issues/5199)
-TODO
+TODO: Still open
 
 
 ### fix get-item -literalpath a*b if `a*b` doesn't actually exist to return error [5197](https://github.com/PowerShell/PowerShell/issues/5197)
-TODO
+Previously, `-literalpath` given a wildcard will treat it the same as -path and if the wildcard found no files, it would silently exit. Correct behavior should be that -literalpath is literal so if the file doesn't exist, it should error. Fix is to also see if wildcard is being suppressed.
```
SteveL-MSFT: Last sentence should be: Change is to treat wildcards used with `-Literal` as literal.

Also, use semantic linefeeds.

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
 
 
 ### don't insert linebreaks to output (except for tables) [5193](https://github.com/PowerShell/PowerShell/issues/5193)
-TODO
+TODO: Not sure what there is besides the title
```
SteveL-MSFT: The gist of this is that previously, output was aligned to the width of the console and linebreaks were added at the end width of the console.  This means that if you resize the terminal, the output doesn't get reformatted as expected.  Tables need linebreaks to keep columns aligned.

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
 
 
 ### Don't wrap stderr as ErrorRecord [5190](https://github.com/PowerShell/PowerShell/issues/5190)
-TODO
+Keep stderr as string and don't wrap as ErrorRecord.  This means that errors from running pwsh.exe within pwsh are now just stderr strings even though they visually look like ErrorRecords so tests had to be changed.
```
SteveL-MSFT: This was closed as the change I made didn't do what I thought it was going to do.  Remove this.

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
-### Make -NoTypeInformation Default on Export-Csv and ConvertTo-Csv [5164](https://github.com/PowerShell/PowerShell/issues/5164)
-TODO
-
-
-### Remove DCOM support from *-Computer cmdlets [5152](https://github.com/PowerShell/PowerShell/issues/5152)
-TODO
+TODO: still open
 
 
 ### Import-Csv Should apply PSTypeNames uppon import when Type Information is present in the CSV [5134](https://github.com/PowerShell/PowerShell/issues/5134)
```
SteveL-MSFT: typoL `upon`

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
-TODO
+Adds PSTypeName preservation on `Import-Csv` and `ConvertFrom-Csv`.
```
SteveL-MSFT: More context here may make this more useful.  I suggest:

```none
Previously objects exported using `Export-CSV` with TypeInformation imported with `ConvertFrom-Csv`
was not retaining the type information.  This change adds the type information to PSTypeNames member
if available in the csv file.
```

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
 
 
 ### -NoTypeInformation should be default on Export-Csv [5131](https://github.com/PowerShell/PowerShell/issues/5131)
-TODO
+Sets `-NoTypreInformation` as the default for `Export-Csv` and `ConvertTo-Csv`, hides the `-NoTypeInformation` parameter switch, adds `-IncludeTypeInformation` switch to `Export-Csv` and `ConvertTo-Csv` to enable legacy behavior, and provides a terminating error when both `-NoTypeInformation` and `-IncludeTypeInformation` are supplied.
```
SteveL-MSFT: I think the important part of this is:

```none
Previously, the cmdlet would output a comment as the first line containing the type name of the object.  
The change is to suppress this by default as it's not understood by most tools.  
Use `-IncludeTypeInformation` to retain the previous behavior.
```

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
 
 
 ### StdErr/ErrorRecord should not be modified on output [5127](https://github.com/PowerShell/PowerShell/issues/5127)
-TODO
+TODO: Closed, not merged
 
 
 ### -Verbose should not override $ErrorActionPreference [5113](https://github.com/PowerShell/PowerShell/issues/5113)
-TODO
+`$ErrorActionPreference` is longer overwritten by the `-Verbose` or `-Debug` parameters.
```
SteveL-MSFT:
```none
Previously, if -Verbose or -Debug was specified, it overrode the behavior of $ErrorActionPreference.
With this change, -Verbose and -Debug no longer affect the behavior of $ErrorActionPreference.
```

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
 
 
 ### Web Cmdlets should warn when  legacy -Credential is sent over unencrypted connections [5112](https://github.com/PowerShell/PowerShell/issues/5112)
```
SteveL-MSFT: Extra space between `when` and `legacy`.  However, we should probably just remove the word `legacy`

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
-TODO
+An error was added when a user tries to use `-Credential` (legacy usage without `-Authentication`) or `-UseDefaultCredentials` over a non-HTTPS URI. Users can bypass the error with `-AllowUnencryptedAuthentication`.
```
SteveL-MSFT:
```none
When using HTTP, content including passwords are sent as clear-text.
This change is to not allow this by default and return an error if credentials are being passsed
in an insecure manner.  Users can bypass this by using the `-AllowUnencryptedAuthethentication` switch.
```

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
 
 
 ### Use InvariantCulture for implicit argument conversion [5106](https://github.com/PowerShell/PowerShell/issues/5106)
-TODO
+TODO: Closed, not merged
 
 
 ### Rename powershell.exe to pwsh.exe [5101](https://github.com/PowerShell/PowerShell/issues/5101)
-TODO
+TODO: (Not sure how to summarize this change further than the title)
```
SteveL-MSFT:
```none
For PowerShell Core 6, the console host was renamed from powershell.exe to pwsh.exe.
The primary motivation is to ensure side-by-side use of Windows PowerShell and PowerShell Core 6
starts the expected version of PowerShell as scripts are not 100% compatible across the two.
```

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
 
 
-### Remove '-ComputerName' from 'Get/Set/Remove-Service' [5094](https://github.com/PowerShell/PowerShell/issues/5094)
-TODO
+### Remove '-ComputerName' from *-Service cmdlets [5090](https://github.com/PowerShell/PowerShell/issues/5094)
+Although `-ComputerName` worked correctly, it was removed from `*-Service` cmdlets so that users consistently rely on PSRP and only have that port open.
```
SteveL-MSFT: Should have similar language to the one about \*-Computer cmdlets above

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
 
 
 ### Unify cmdlets with parameter 'Encoding' to be of type System.Text.Encoding [5080](https://github.com/PowerShell/PowerShell/issues/5080)
-TODO
+The `-Encoding` Byte has been removed from the filesystem provider cmdlets. A new parameter -Byte is now the signal that a byte stream is required as input or output will be a stream of bytes.
```
SteveL-MSFT:
```none
Previously, an enum value of `Byte` was used with `-Encoding` to indicate a byte stream.
However, `Byte` is not an actual encoding.  The change is remove that enum value and replace
it with a `-Byte` switch to indicate input or output should be a byte stream.
```

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
-
-
-### remove -computer parameter from get-process [4960](https://github.com/PowerShell/PowerShell/issues/4960)
-TODO
+The following features were removed as they are not supported in PowerShell Core, and there are no plans to add support as they exist for legacy reasons for Windows PowerShell: psconsolefile switch and code, importsystemmodulues switch and code, and font changing code.
 
 
 ### removed RunspaceConfiguration support [4942](https://github.com/PowerShell/PowerShell/issues/4942)
-TODO
+In PSCore6 only InitialSessionState is supported.  Since the RunspaceConfiguration api was made internal, the code related to RunspaceConfiguration has been removed.  This means that some public apis have changed and it was decided that it was best to have a compile time error.
```
SteveL-MSFT:
```none
When creating a PowerShell runspace programmatically using the api, previously you could use the legacy
RunspaceConfiguration or the newer InitialSessionState.  This change removed support for RunspaceConfiguration and only supports InitialSessionState.
```

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
 
 
 ### remove unsupported -showwindow switch from get-help [4903](https://github.com/PowerShell/PowerShell/issues/4903)
-TODO
+`-showwindow` relies on WPF, which is not supported on CoreCLR.
 
 
 ### allow * to be used in registry path for remove-item [4866](https://github.com/PowerShell/PowerShell/issues/4866)
-TODO
+Using `-literalpath` with an asterisk and the filesystem provider will now return an error rather than returning quietly.
```
SteveL-MSFT: Add language similar to the -LiteralPath wildcard issue above or consider combining them.

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
 
 
 ### fix set-service failing test [4802](https://github.com/PowerShell/PowerShell/issues/4802)
-TODO
-
+`New-Service` should throw on System `-StartupType`
+TODO : Looks like this a change in test cases, not sure what the documentation process should be
```
SteveL-MSFT: The change isn't only in test code.  Previously if you use `New-Service -StartupType foo`, I believe it quietly ignored `foo` and created the service with some default startup type.  The change is to explicitly throw an error for invalid startup type.

docs/BREAKINGCHANGES.md
```diff
@@ -6,151 +8,125 @@ TODO
...
 
 ### Enhance the -split operator with negative maximum token counts to split from the end [4765](https://github.com/PowerShell/PowerShell/issues/4765)
 TODO
 
 
-### Rename $IsOSX to $IsMacOS [4757](https://github.com/PowerShell/PowerShell/issues/4757)
-TODO
+### Rename $IsOSX to $IsacOS [4700](https://github.com/PowerShell/PowerShell/issues/4700)
```
SteveL-MSFT: typo `$IsMacOS`

docs/BREAKINGCHANGES.md
```diff
@@ -163,7 +139,7 @@ TODO
 
 
 ### Make error message consistent when invalid script is passed to -File, better error when passed ambiguous arg [4573](https://github.com/PowerShell/PowerShell/issues/4573)
-TODO
+TODO (I believe the breaking change here is for "Enable -WindowStyle to work", changing the exit code.  Need to confirm)
```
SteveL-MSFT: No need to mention enabling -WindowStyle to work, the breaking change is modifying exit codes of pwsh.exe to align with Unix conventions.

docs/BREAKINGCHANGES.md
```diff
@@ -175,11 +151,11 @@ TODO
 
 
 ### Due to the use of unsupported APIs, we must remove the Counter CmdLets in the Diagnostics Module until a better solution is found. [4303](https://github.com/PowerShell/PowerShell/issues/4303)
-TODO
+TODO (Not sure how to document more than the title)
```
SteveL-MSFT: Should be a separate section on cmdlet coverage compared to Windows PowerShell

docs/BREAKINGCHANGES.md
```diff
@@ -175,11 +151,11 @@ TODO
...
 
 
 ###  Due to the use of unsupported APIs, we must remove the LocalAccounts module until a better solution is found. [4302](https://github.com/PowerShell/PowerShell/issues/4302)
-TODO
+TODO (Not sure how to document more than the title)
```
SteveL-MSFT: Should be a separate section on cmdlet coverage compared to Windows PowerShell

docs/BREAKINGCHANGES.md
```diff
@@ -195,16 +171,16 @@ TODO
...
 
 
 ### Command line arguments with a dollar sign [4024](https://github.com/PowerShell/PowerShell/issues/4024)
 TODO
 
 
 ### Change positional parameter for powershell.exe from -Command to -File [4019](https://github.com/PowerShell/PowerShell/issues/4019)
-TODO
-
+This fixes the usage of `#!` in PowerShell scripts that are being executed from non-PowerShell shells on non-Windows platforms.
```
SteveL-MSFT: Enable shebang use of PowerShell on non-Windows platforms.  This means on Unix based systems,
you can make a script executable which would invoke PowerShell automatically rather than explicitly
invoking pwsh.

SteveL-MSFT (changes requested):

docs/BREAKINGCHANGES.md
rkeithhill: I think the parameter is `-AsByteStream`, right?

SteveL-MSFT:
@maertendMSFT looks like you made a big update while I was adding comments so they appear "out dated", but you should look at them to see if they've been addressed since GitHub doesn't know

docs/BREAKINGCHANGES.md
maertendMSFT: Good catch

docs/BREAKINGCHANGES.md
```diff
@@ -107,13 +115,20 @@ TODO: Closed, not merged
 
 
 
+### Unify cmdlets with parameter `Encoding` to be of type `System.Text.Encoding` [#5080](https://github.com/PowerShell/PowerShell/issues/5080)
 
+<<<<<<< HEAD
```
iSazonov: Typo.

docs/BREAKINGCHANGES.md
```diff
@@ -107,13 +115,20 @@ TODO: Closed, not merged
...
 
 ### Unify cmdlets with parameter `-Encoding` to be of type System.Text.Encoding [#5080](https://github.com/PowerShell/PowerShell/issues/5080)
 Previously, an enum value of `Byte` was used with `-Encoding` to indicate a byte stream.
 However, `Byte` is not an actual encoding.  The change is remove that enum value and replace
 it with a `-AsByteStream` switch to indicate input or output should be a byte stream.
+=======
+The `-Encoding` value `Byte` has been removed from the filesystem provider cmdlets.
+A new parameter `-Byte` is now used to specify that a byte stream is required as input or that output will be a stream of bytes.
+
+>>>>>>> reformatting to BREAKINGCHANGES
```
iSazonov: Typo.

docs/BREAKINGCHANGES.md
```diff
@@ -107,13 +115,20 @@ TODO: Closed, not merged
...
+### Unify cmdlets with parameter `Encoding` to be of type `System.Text.Encoding` [#5080](https://github.com/PowerShell/PowerShell/issues/5080)
 
+<<<<<<< HEAD
 
 
 ### Unify cmdlets with parameter `-Encoding` to be of type System.Text.Encoding [#5080](https://github.com/PowerShell/PowerShell/issues/5080)
 Previously, an enum value of `Byte` was used with `-Encoding` to indicate a byte stream.
 However, `Byte` is not an actual encoding.  The change is remove that enum value and replace
 it with a `-AsByteStream` switch to indicate input or output should be a byte stream.
+=======
```
iSazonov: Typo.

docs/BREAKINGCHANGES.md
```diff
@@ -0,0 +1,217 @@
...
+### Enhance the -split operator with negative maximum token counts to split from the end [#4765](https://github.com/PowerShell/PowerShell/issues/4765)
+After the enhancement, only a `<Max-substrings>` value of `0` will be accepted as the explicit signal that all tokens should be returned, however many are found in the input string(s).
+Negative `<Max-substrings>` values will work analogously to the already supported positive values, except that: they return the specified number of strings from the end of the string(s) and all individually extracted tokens are returned in input order, and whatever unsplit part remains, if any, is returned as the first token.
+
+
+### Rename $IsOSX to $IsMacOS [#4700](https://github.com/PowerShell/PowerShell/issues/4700)
+The naming in PowerShell should be consistent with our naming andd conform to Apple's use of macOS instead of OSX, however, for readability and consistently we are staying with Pascal casing.
+
+
+### Prepare for BOM-less UTF-8 default character encoding with respect to $OutputEncoding and console code page  [#4681](https://github.com/PowerShell/PowerShell/issues/4681)
```
SteveL-MSFT: This is dupe of above as this is the issue and the other is the PR

docs/BREAKINGCHANGES.md
```diff
@@ -0,0 +1,217 @@
...
+### Make error message consistent when invalid script is passed to -File, better error when passed ambiguous arg [#4573](https://github.com/PowerShell/PowerShell/issues/4573)
+Change the exit codes of pwsh.exe to align with Unix conventions
+
+
+### Removal of `LocalAccount` and cmdlets from  `Diagnostics` modules. [#4302](https://github.com/PowerShell/PowerShell/issues/4302) [#4303](https://github.com/PowerShell/PowerShell/issues/4303)
+Due to unsupported APIs, the `LocalAccounts` module and the `Counter` cmdlets in the `Diagnostics` module were removed until a better solution is found. 
+
+
+### Remove UTC and SQM telemetry code [#4190](https://github.com/PowerShell/PowerShell/issues/4190)
+TODO: Should this still be in the doc if the change is not visible to the user and is about telemetry collection?
```
SteveL-MSFT: This shouldn't be in the list, it's not customer facing

docs/BREAKINGCHANGES.md
```diff
@@ -0,0 +1,217 @@
...
+
+
+### Executing powershell script with bool parameter doesnt work [#4036](https://github.com/PowerShell/PowerShell/issues/4036)
+Previously, using powershell.exe (now pwsh.exe) to execute a PowerShell script using `-File` provided no way to pass $true/$false as parameter values.  Support for $true/$false as parsed values to parameters was added. Switch values are also supported as currently documented syntax doesn't work.
```
SteveL-MSFT: Semantic line feeds.

docs/BREAKINGCHANGES.md
```diff
@@ -0,0 +1,217 @@
...
+
+
+### Remove ClrVersion property from $PSVersionTable [#4027](https://github.com/PowerShell/PowerShell/issues/4027)
+The ClrVersion property of `$PSVersionTable` is not useful with CoreCLR,  end users should not be using that value to determine compatibility.
```
SteveL-MSFT: Extra space between ", end"

docs/BREAKINGCHANGES.md
```diff
@@ -0,0 +1,217 @@
...
+
+
+### Change positional parameter for powershell.exe from -Command to -File [#4019](https://github.com/PowerShell/PowerShell/issues/4019)
+Enable shebang use of PowerShell on non-Windows platforms. This means on Unix based systems,
+you can make a script executable which would invoke PowerShell automatically rather than explicitly invoking pwsh.
+This also means that you can now do things like `powershell foo.ps1` or `powershell fooScript` without specifying `-File`.
+However, this change now requires that you explicitly specify `-c` or `-Command` when trying to do things like `powershell.exe Get-Command`.
+
+### Implement Unicode escape parsing [#3958](https://github.com/PowerShell/PowerShell/issues/3958)
+TODO
```
SteveL-MSFT: I think you can just say:

\`u#### or \`u{####} is converted to the corresponding Unicode character.  To output a literal \`u, escape the backtick: ``u.

docs/BREAKINGCHANGES.md
```diff
@@ -0,0 +1,217 @@
...
+ The `-raw` parameter is now No-op. Going forward all of the output will be displayed with a true representation of numbers that includes all of the bytes for its type (what the Raw parameter was formally doing).
+
+
+### PowerShell as a default shell doesn't work with script command [#3319](https://github.com/PowerShell/PowerShell/issues/3319)
+On Unix, it is a convention for shells to accept `-i` for an interactive shell and many tools expect this behavior (`script` for example, and when setting powershell as the default shell) and calls the shell with the `-i` switch. This change is breaking in that -i previously could be used as short hand to match `-inputformat` which now will need to be `-in`.
+
+
+### Completions for environment variables differ between platforms [#3227](https://github.com/PowerShell/PowerShell/issues/3227)
+The casing of `$PSModulePath` has been corrected to be consistent with Windows PowerShell.
+TODO: not sure if this actually a breaking change.  If the change is to make something consistent with Windows, there would be no difference between 6.0 and 5.1
```
SteveL-MSFT: I think you can remove this, this was only a breaking change between releases of PSCore6

docs/BREAKINGCHANGES.md
```diff
@@ -0,0 +1,217 @@
...
+
+
+### Typo fix in Get-ComputerInfo property name [#3167](https://github.com/PowerShell/PowerShell/issues/3167)
+`BiosSerialNumber` was misspelled as "BiosSeralNumber".
+
+
+### Add Get-StringHash and Get-FileHash cmdlets [#3024](https://github.com/PowerShell/PowerShell/issues/3024)
+TODO: Labeled with breaking change so the difference between windows and core can be described
```
SteveL-MSFT: The breaking change is that some hash algorithms are not supported by corefx and so no longer available: MACTripleDES, RIPEMD160

docs/BREAKINGCHANGES.md
```diff
@@ -0,0 +1,217 @@
...
+
+
+### Add validation on get-* cmdlets where passing $null returns all objects instead of error [#2672](https://github.com/PowerShell/PowerShell/issues/2672)
+Passing `$null` to any of the following will now throw an error: `Get-Credential -UserName`, `Get-Event -SourceIdentifier`, `Get-EventSubscriber -SourceIdentifier`, `Get-Help -Name`, `Get-PSBreakpoint -Script`, `Get-PSProvider -PSProvider`, `Get-PSSessionConfiguration -Name`, `Get-PSSnapin -Name`, `Get-Runspace -Name`, `Get-RunspaceDebug -RunspaceName`, `Get-Service -Name`, `Get-TraceSource -Name`, `Get-Variable -Name`, `Get-WmiObject -Class`, and `Get-WmiObject -Property`.
+
+
+### Add support W3C Extended Log File Format in Import-Csv [#2482](https://github.com/PowerShell/PowerShell/issues/2482)
+Previously, the `Iport-Csv` cmdlet cannot be used to directly import the log files in W3C extended log format and additional action would be required.
```
SteveL-MSFT: typo `Import-CSV`

docs/BREAKINGCHANGES.md
```diff
@@ -0,0 +1,217 @@
...
+With this change, W3C extended log format is supported.
+
+
+### Parameter binding problem with ValueFromRemainingArguments in PS functions [#2035](https://github.com/PowerShell/PowerShell/issues/2035)
+TODO
```
SteveL-MSFT: ValueFromRemainingArguments now returns the values as an array instead of a single value which itself is an array

SteveL-MSFT (changes requested):

docs/BREAKINGCHANGES.md
alexandair: @SteveL-MSFT Casing issue. Import-Csv.

docs/BREAKINGCHANGES.md
```diff
@@ -0,0 +1,217 @@
...
+
+### Don't wrap `stderr` as `ErrorRecord` [#5190](https://github.com/PowerShell/PowerShell/issues/5190)
+
+Keep stderr as string and don't wrap as an `ErrorRecord`.
+This means that errors from running `pwsh.exe` within PowerShell Core are now just `stderr` strings even though they visually look like ErrorRecords
+This change my require you to change how your tests handle native commmand errors.
+
+### `-Verbose` and `-Debug` no longer overrides `$ErrorActionPreference` [#5113](https://github.com/PowerShell/PowerShell/issues/5113)
+
+Previously, if `-Verbose` or `-Debug` were specified, it overrode the behavior of `$ErrorActionPreference`.
```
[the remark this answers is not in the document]
maertendMSFT: I think this could go either way, since there are 2 parameters specified I used there plural `were`.  But `was` would be appropriate because the parameters can be used individually.  In this case, I think `were` is correct since we are listing multiple parameters.

docs/BREAKINGCHANGES.md
maertendMSFT: Removed

docs/BREAKINGCHANGES.md
maertendMSFT: Agreed, removed

docs/BREAKINGCHANGES.md
SteveL-MSFT: @alexandair thanks for the correction of the correction :)

docs/BREAKINGCHANGES.md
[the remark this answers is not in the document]
SteveL-MSFT: Ok

joeyaiello:
@TravisEz13 these spelling errors in Travis CI don't make any sense, you mind helping out?

markekraus:
The change in Web Cmdlets from `WebRequest` to `HttpClient` introduced several breaking changes As well as some breaking changes around the lack of IE interop. I'm not sure if they should be documented here or not, But I feel the need to call them out.

* Basic HTML Parsing only (`ParsedHtml` and `Forms` properties no longer exist and `Invoke-WebRequest` now always returns a `BasicHtmlWebResponseObject`)
* `BasicHtmlWebResponseObject.Headers` values are now `String[]` instead of `String`
* `BasicHtmlWebResponseObject.BaseResponse` is now `System.Net.Http.HttpResponseMessage` (has error handling implications)
* Strict RFC header parsing is now defualt for `-Headers` and `-UserAgent` (bypass with `-SkipHeaderValidation`)
* `file://` and `ftp://` URI's are no longer supported.
* `System.Net.ServicePointManager` settings are not honored
* No Certificate based authentication on macOS

joeyaiello:
@markekraus thanks for the callout! This is definitely going to be a living document. We're doing the launch in just a few mins...would you mind submitting a PR with that stuff into here? I can't afford to break spelling right now.... :\

TravisEz13:
One spelling issue, https://travis-ci.org/PowerShell/PowerShell/jobs/327363249#L3337, please fix.

markekraus:
@joeyaiello Might be a few hours, but yes. I will put in a PR for that.. erm.. How do you want it and where?

joeyaiello:
Oh, definitely no rush. It'd be awesome if you just added another H3 summarizing "changes to web cmdlets" or something like. And into this doc in master, please.

You're the best!

TravisEz13 (approved):

SteveL-MSFT (approved):

Issue #5566, opened 2017-11-28
NullReferenceException in PrioritySendDataCollection.Clear()


dantraMSFT:
We have two reports of a NullReferenceException in this method on macOS but it's not clear if the field(s) being accessed are null of if elements in the array are null.  
The code is guarded by a Dbg.Assert but it's clear that a non-debug check is called for.
see Dbg.Assert(null != _dataToBeSent, "Serialized streams are not initialized");

See issue https://github.com/PowerShell/PowerShell/issues/4029 for repro steps.

Note that this is not repro'ing on machines available in-house.

PR #5622, opened 2017-12-04
remoting: Handle DLLImport failure of libpsrpclient

dantraMSFT:
Issue https://github.com/PowerShell/PowerShell/issues/4029 exposed two problems when failing to load libpsrpclient. 
* WSManAPIDataCommon.ctor was not identifying the binary that failed DLLImport; hindering diagnosability 
* PrioritySendDataCollection.Clear() would throw a NullReferenceException when called from the finalizer after an error path.

This change addresses both of the above:

* Guard against null arrays in PrioritySendDataCollection.Clear() - called from finalizer in error paths.

* Diagnosability: Log the DllNotFoundException in WSManAPIDataCommon ctor and also include it as the internal exception when throwing PSRemotingTransportException.

### DIFF
--- a/src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
+++ b/src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
@@ -46,11 +46,17 @@ namespace System.Management.Automation.Remoting
         // actual data store(s) to store priority based data and its
         // corresponding sync objects to provide thread safety.
         private SerializedDataStream[] _dataToBeSent;
+        // array of sync objects, one for each element in _dataToBeSent
+        private object[] _dataSyncObjects;
+
+        // synchronize access to _dataSyncObjects and _dataToBeSent as a set.
+        // while the elements in _dataSyncObjects are used to synchronize
+        // access to a specific element in _dataToBeSent, _syncObject is used
+        // to syncronize access to the set.
+        private object _syncObject = new Object();
+
         // fragmentor used to serialize & fragment objects added to this collection.
         private Fragmentor _fragmentor;
-        private object[] _syncObjects;
-        // synchronize access to _syncObjects and _dataToBeSent as a set.
-        private object _syncObject = new Object();
 
         // callbacks used if no data is available at any time.
         // these callbacks are used to notify when data becomes available under
@@ -97,11 +103,11 @@ namespace System.Management.Automation.Remoting
                 // create serialized streams using fragment size.
                 string[] names = Enum.GetNames(typeof(DataPriorityType));
                 _dataToBeSent = new SerializedDataStream[names.Length];
-                _syncObjects = new object[names.Length];
+                _dataSyncObjects = new object[names.Length];
                 for (int i = 0; i < names.Length; i++)
                 {
                     _dataToBeSent[i] = new SerializedDataStream(_fragmentor.FragmentSize);
-                    _syncObjects[i] = new object();
+                    _dataSyncObjects[i] = new object();
                 }
             }
         }
@@ -128,7 +134,7 @@ namespace System.Management.Automation.Remoting
             // make sure the only one object is fragmented and added to the collection
             // at any give time. This way the order of fragment is maintained
             // in the SendDataCollection(s).
-            lock (_syncObjects[(int)priority])
+            lock (_dataSyncObjects[(int)priority])
             {
                 _fragmentor.Fragment<T>(data, _dataToBeSent[(int)priority]);
             }
@@ -156,28 +162,28 @@ namespace System.Management.Automation.Remoting
         /// </summary>
         internal void Clear()
         {
-            /* Lock _syncObjects and _dataToBeSent as a set to ensure atomic cleanup. */
+            /* Lock _dataSyncObjects and _dataToBeSent as a set to ensure atomic cleanup. */
             lock (_syncObject)
             {
                 /*
-                    NOTE: Error paths during initialization can cause _syncObjects to be null
+                    NOTE: Error paths during initialization can cause _dataSyncObjects to be null
                     causing an unhandled exception in finalize and a process crash.
                 */
-                if (null != _syncObjects)
+                if (null != _dataSyncObjects)
                 {
                     Dbg.Assert(null != _dataToBeSent, "Serialized streams are not initialized");
-                    lock (_syncObjects[(int)DataPriorityType.PromptResponse])
+                    lock (_dataSyncObjects[(int)DataPriorityType.PromptResponse])
                     {
                         _dataToBeSent[(int)DataPriorityType.PromptResponse].Dispose();
                     }
 
-                    lock (_syncObjects[(int)DataPriorityType.Default])
+                    lock (_dataSyncObjects[(int)DataPriorityType.Default])
                     {
                         _dataToBeSent[(int)DataPriorityType.Default].Dispose();
                     }
 
                     _dataToBeSent = null;
-                    _syncObjects = null;
+                    _dataSyncObjects = null;
                 }
             }
         }
--- a/src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
+++ b/src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
@@ -170,7 +170,6 @@ namespace System.Management.Automation.Remoting
                     {
                         _dataToBeSent[(int)DataPriorityType.PromptResponse].Dispose();
                     }
-                    _dataToBeSent = null;
 
                     lock (_syncObjects[(int)DataPriorityType.Default])
                     {
--- a/src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
+++ b/src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
@@ -49,6 +49,8 @@ namespace System.Management.Automation.Remoting
         // fragmentor used to serialize & fragment objects added to this collection.
         private Fragmentor _fragmentor;
         private object[] _syncObjects;
+        // synchronize access to _syncObjects and _dataToBeSent as a set.
+        private object _syncObject = new Object();
 
         // callbacks used if no data is available at any time.
         // these callbacks are used to notify when data becomes available under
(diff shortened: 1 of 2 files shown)


dantraMSFT:
@mirichmo and @PaulHigin will either of you be able to look at this?

src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
```diff
@@ -154,15 +156,29 @@ internal void Add<T>(RemoteDataObject<T> data)
         /// </summary>
         internal void Clear()
         {
-            Dbg.Assert(null != _dataToBeSent, "Serialized streams are not initialized");
-            lock (_syncObjects[(int)DataPriorityType.PromptResponse])
+            /* Lock _syncObjects and _dataToBeSent as a set to ensure atomic cleanup. */
+            lock (_syncObject)
```
anmenaga: Is it possible to lock on '_syncObjects' array itself without creating additional field "_syncObject" ?

src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
```diff
@@ -49,6 +49,8 @@ internal class PrioritySendDataCollection
         // fragmentor used to serialize & fragment objects added to this collection.
         private Fragmentor _fragmentor;
         private object[] _syncObjects;
+        // synchronize access to _syncObjects and _dataToBeSent as a set.
+        private object _syncObject = new Object();
```
anmenaga: This field is different from array "_syncObjects" only by last letter - it is confusing.
It would be good to either add comment for "_syncObjects" array about what it is; or rename "_syncObject"; or both.

anmenaga (commented):
Leave a comment

src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
dantraMSFT: @anmenaga: Actually, after reviewing this logic again, _syncObject is insufficient since it only guards Clear(). Additionally, _syncObjects also appears to be fragile; it guards against individual usage of a given element but assumes the elements are valid; which is the actual problem.

My thinking is _syncObjects should be removed and _syncObject used to serialize access to the set instead of individual members.

Thoughts?

TravisEz13:
also related to issue #5566

src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
anmenaga: @dantraMSFT I'm all for simplification, but there should be a reason why they decided to go more complex 'individual-sync-object' route in the first place... Maybe something will deadlock with a single syncObject, and parallel "Add"s won't be able to execute for different DataPriorityTypes (not sure how critical this really is).

TravisEz13:
LGTM, but I would like @PaulHigin  to have a look.

TravisEz13:
@anmenaga Do you want to sign off, or did you just comment?

anmenaga:
@TravisEz13 
The conversation continues about refactoring that Dan proposed; but I think that refactoring (if needs to be done at all) can be done as separate PR.

anmenaga (approved):
LGTM

dantraMSFT:
@TravisEz13 This should be ready to merge.

TravisEz13:
I've been asked to wait for a review by @PaulHigin  or @daxian-dbw

src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
```diff
@@ -154,15 +162,29 @@ internal void Add<T>(RemoteDataObject<T> data)
...
             {
-                _dataToBeSent[(int)DataPriorityType.PromptResponse].Dispose();
-            }
+                /*
+                    NOTE: Error paths during initialization can cause _dataSyncObjects to be null
+                    causing an unhandled exception in finalize and a process crash.
+                */
+                if (null != _dataSyncObjects)
+                {
+                    Dbg.Assert(null != _dataToBeSent, "Serialized streams are not initialized");
```
daxian-dbw: This assert should be removed because apparently `_dataToBeSent` can be null in that error path.

src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
```diff
@@ -154,15 +162,29 @@ internal void Add<T>(RemoteDataObject<T> data)
         /// </summary>
         internal void Clear()
         {
-            Dbg.Assert(null != _dataToBeSent, "Serialized streams are not initialized");
-            lock (_syncObjects[(int)DataPriorityType.PromptResponse])
+            /* Lock _dataSyncObjects and _dataToBeSent as a set to ensure atomic cleanup. */
+            lock (_syncObject)
```
daxian-dbw: I think we can avoid the sync object `_syncObject` by the following code:
```CSharp
if (null != _dataSyncObjects)
{
    int promptResponseIndex = (int)DataPriorityType.PromptResponse;
    int defaultIndex = (int)DataPriorityType.Default;

    if (_dataToBeSent[promptResponseIndex] != null)
    {
        lock (_dataSyncObjects[promptResponseIndex])
        {
            if (_dataToBeSent[promptResponseIndex] != null)
            {
                _dataToBeSent[promptResponseIndex].Dispose();
                _dataToBeSent[promptResponseIndex] = null;
            }
        }
    }

    if (_dataToBeSent[defaultIndex] != null)
    {
        lock (_dataSyncObjects[defaultIndex])
        {
            if (_dataToBeSent[defaultIndex] != null)
            {
                _dataToBeSent[defaultIndex].Dispose();
                _dataToBeSent[defaultIndex] = null;
            }
        }
    }
}
```

daxian-dbw (dismissed):
Left two comments.

src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
daxian-dbw: > Actually, after reviewing this logic again, _syncObject is insufficient since it only guards Clear().

`_dataSyncObjects` and `_dataToBeSent` are used in multiple methods, but it appears that there is an assumption that those methods won't be called until the `Fragmentor` property is set (where both `_dataSyncObjects` and `_dataToBeSent` would be initialized). With this assumption, I think it's fine to directly use `_dataSyncObjects` and `_dataToBeSent` in methods other than `Clear()`. But we probably should add a comment about that assumption.

For `Clear()`, it's possible to be called before `Fragmentor` being set (in error path), so it's necessary to check `_dataSyncObjects` for null beforehand.

src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
daxian-dbw: Never mind this comment. It's certainly right to assume `_dataToBeSent != null` when `null != _dataSyncObjects` 😄

src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
dantraMSFT: You're right; I'll make the change.

src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
dantraMSFT: Will do.

src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
dantraMSFT: fixed

src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
dantraMSFT: fixed

dantraMSFT:
@daxian-dbw do you have any additional feedback?

dantraMSFT:
@PaulHigin, @daxian-dbw, please take a look at this; it's marked as GA. Thanks.

daxian-dbw (approved):

src/System.Management.Automation/engine/remoting/fanin/PriorityCollection.cs
```diff
@@ -154,15 +156,33 @@ internal void Add<T>(RemoteDataObject<T> data)
...
             {
-                _dataToBeSent[(int)DataPriorityType.PromptResponse].Dispose();
-            }
+                int promptResponseIndex = (int)DataPriorityType.PromptResponse;
+                int defaultIndex = (int)DataPriorityType.Default;
 
-            lock (_syncObjects[(int)DataPriorityType.Default])
-            {
-                _dataToBeSent[(int)DataPriorityType.Default].Dispose();
+                lock (_dataSyncObjects[promptResponseIndex])
```
PaulHigin: Sorry for the late review.  I am not familiar with the error path but if _dataToBeSent array can contain null entries then it seems like so can _dataSyncObjects (_dataSyncObjects[promptResponseIndex] == null ?).

It seems like these arrays should be created in the PrioritySendDataCollection constructor...

dantraMSFT: @PaulHigin: _dataSyncObjects is populated when it is created.  The entries are not null if the array is not null.
As far as when it is created, it appears the intent was to create both arrays at the same point.  I saw no strong reason to change it.

PaulHigin: But the _dataSyncObjects array is filled within a loop:
```
for (int i = 0; i < names.Length; i++)
    {
        _dataToBeSent[i] = new SerializedDataStream(_fragmentor.FragmentSize);
        _dataSyncObjects[i] = new object();
    }
```
If an exception is thrown within the loop then it is possible an array element is null.  Is this not the error path we are worried about?

dantraMSFT: SerializedDataStream's ctor does not currently throw.

Note that it should (it relies on Dbg.Assert) and I opened an issue for this: https://github.com/PowerShell/PowerShell/issues/5800

PR #5633, opened 2017-12-06
PowerShellGet: Change AllUsers scope install location to SHARED_MODULES path

bmanikm:
## PR Checklist

- [x] [Update the changelog](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---changelog)
- [NA] [Change is not breaking](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes)
- [x] [Make sure you've added a new test, if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
    - [NA] [Add `[feature]` if the change is significant or affectes feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

## PR Summary
- Changed the install location of AllUsers scope on PWSH to SHARED_MODULES location. PowerShell/PowerShellGet#175.
- Refreshing the PowerShellGet.Tests.ps1 file from PowerShell/PowerShellGet repository.
- Removed $IsCoreCLR variable usage.
- Updated the logic for retrieving MyDocuments path on Windows.

### DIFF
--- a/test/powershell/Modules/PowerShellGet/PowerShellGet.Tests.ps1
+++ b/test/powershell/Modules/PowerShellGet/PowerShellGet.Tests.ps1
@@ -13,7 +13,7 @@ $Initialized = $false
 
 function IsInbox { $PSHOME.EndsWith('\WindowsPowerShell\v1.0', [System.StringComparison]::OrdinalIgnoreCase) }
 function IsWindows { $PSVariable = Get-Variable -Name IsWindows -ErrorAction Ignore; return (-not $PSVariable -or $PSVariable.Value) }
-function IsCoreCLR { $PSVariable = Get-Variable -Name IsCoreCLR -ErrorAction Ignore; return ($PSVariable -and $PSVariable.Value) }
+function IsCoreCLR { $PSVersionTable.ContainsKey('PSEdition') -and $PSVersionTable.PSEdition -eq 'Core' }
 
 #endregion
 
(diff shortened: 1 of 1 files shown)


daxian-dbw (approved):
LGTM. Please add `[feature]` tag to the commit message to trigger a full test run.

test/powershell/Modules/PowerShellGet/PowerShellGet.Tests.ps1
```diff
@@ -48,13 +52,22 @@ if(IsInbox)
...
+            Microsoft.PowerShell.Management\Join-Path -Path $script:MyDocumentsFolderPath -ChildPath 'PowerShell'
+        }
+        else
+        {
+            Microsoft.PowerShell.Management\Join-Path -Path $HOME -ChildPath "Documents\PowerShell"
+        }
+    }
+    else
+    {
+        $script:MyDocumentsPSPath = Split-Path -Path ([System.Management.Automation.Platform]::SelectProductNameForDirectory('USER_MODULES')) -Parent
```
JamesWTruher: we don't really have a way to handle this if you don't have the privilege to write in the directory on non-windows platforms, we can look at extending the test framework for next release.

I recommend a test hook approach, which allows you to provide an alternative location. From a test perspective, the only thing you want to be sure of is when you select a location the files get placed there. It doesn't really matter what that location is, it could be anywhere - A testhook which makes [system.management.automation.platform]::SelectProductNameForDirectory('SHARED_MODULES') point to `$TESTDRIVE` should be sufficient as long as you have a separate test which returns what the default location of `SHARED_MODULES` should be.
This provides far more flexibility in the long run
If you can't do that, I would mark this specific test as `pending` for non-windows

test/powershell/Modules/PowerShellGet/PowerShellGet.Tests.ps1
```diff
@@ -154,7 +167,7 @@ Describe "PowerShellGet - Module tests (Admin)" -tags @('Feature', 'RequireAdmin
         $module = Get-Module $ContosoServer -ListAvailable
         $module.Name | Should be $ContosoServer
         $module.ModuleBase | Should Be $installedModuleInfo.InstalledLocation
-    }
+    } -Skip:$IsLinux
```
daxian-dbw: Please use `-Pending` instead of `-Skip`. We track all pending tests because they are supposed to work but now at the time. `-Skip` means they are not applicable to the platform.

daxian-dbw: And also, please add `-Pending` before the open curly bracket, like
```
It "Should install a module correctly to the required location with default AllUsers scope" -Pending:$IsLinux {
```

test/powershell/Modules/PowerShellGet/PowerShellGet.Tests.ps1
```diff
@@ -219,7 +232,7 @@ Describe "PowerShellGet - Script tests (Admin)" -tags @('Feature', 'RequireAdmin
         $installedScriptInfo | Should Not Be $null
         $installedScriptInfo.Name | Should Be $FabrikamServerScript
         $installedScriptInfo.InstalledLocation.StartsWith($script:ProgramFilesScriptsPath, [System.StringComparison]::OrdinalIgnoreCase) | Should Be $true
-    }
+    } -Skip:$IsLinux
```
daxian-dbw: Same comment here.

daxian-dbw:
In fact, with `sudo: required` in `.travis.yml`, you can use `sudo` without providing a password on Travis CI Linux to get the root privilege. So we should be able to support `RequireAdminOnLinux`. #5645 is opened to track that.

As for macOS, it turns out the Shared-Module path is not protected, and that's why it passed on macOS.

daxian-dbw (approved):

daxian-dbw:
@adityapatwardhan Can you take a look at the failure in AppVeyor to see what might be wrong?

daxian-dbw:
@bmanikm You didn't address my comment at https://github.com/PowerShell/PowerShell/pull/5633#discussion_r155403480
I have addressed it and also add comments about why we mark them pending on Linux.

Issue #5610, opened 2017-12-02
Merge, Move and Clean Web Cmdlets


markekraus:
This issue is to track work on merging, moving , and cleaning the Web Cmdlets code

* [x] Merge partial classes into single class files
* [ ] Clean up and organize the cmdlets code
  * [ ] Remove `partial`'s
  * [ ] Sort `usings`'s, Properties, Methods, etc
  * [ ] Fix comments
  * [ ] Fix Formatting
  * [ ] Convert to automatic properties where possible
  * [ ] Remove `using Microsoft.Win32;`
* [ ] Move all web cmdlet code from sub folders to `src\Microsoft.PowerShell.Commands.Utility\commands\utility\WebCmdlet\`  top level folder
* [ ] Move independent enums and classes out of WebRequestPSCmdlet.cs to separate files

markekraus:
@SteveL-MSFT @TravisEz13  The next phase of this cleanup is to sort and order the web cmdlet code. I looked at the [coding guidlines](https://github.com/PowerShell/PowerShell/blob/master/docs/dev-process/coding-guidelines.md), but I don't see any guidance for the order of fields, properties, constructors, methods, etc other than "fields at the top unless they are backing properties". The web cmdlets are a mess with code of varying types and access all over the place.

I was looking at this http://stylecop.soyuz5.com/Ordering%20Rules.html and wondering if that would be a good reference or if this project had its own guidance that hasn't been published yet.

SteveL-MSFT:
@markekraus I would be fine starting with stylecop.  Since the codebase is a decade old and produced by various engineers over time, it's inherently inconsistent and something that we should only address over time.

iSazonov:
See [comment](https://github.com/PowerShell/PowerShell/pull/5972#discussion_r162855127) 
(null should be on first place).

PR #5662, opened 2017-12-09
Sort and Merge Classes in WebCmdlets\Common

markekraus:
## PR Summary

Reference #5610

* Merges partial classes in the WebCmdlets\Common folder
* Sorts the `using` statements
* Sorts class elements according to StyleCop guidance https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1201.md
* Sorts within the same elements by access type according to StyleCop guidance https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1202.md
* Sorts elements of the same type and access alphabetically.
* Order of cmdlet public properties and parameters was untouched to preserve their order in PowerShell
* Proper inheritance was added to `InvokeRestMethodCommand`.
* Some new regions were created where it makes sense
* Formatting and comments of code blocks was purposefully maintained. These will be handled in a separate cleanup action.
* No logic was modified

## PR Checklist

- NA User facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - NA Issue filed - Issue link:
- NA [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/ContentHelper.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/ContentHelper.Common.cs
@@ -3,12 +3,12 @@ Copyright (c) Microsoft Corporation. All rights reserved.
 --********************************************************************/
 
 using System;
-using System.Management.Automation;
-using System.Text;
-using Microsoft.Win32;
 using System.Linq;
+using System.Management.Automation;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Text;
+using Microsoft.Win32;
 
 namespace Microsoft.PowerShell.Commands
 {
@@ -26,22 +26,22 @@ namespace Microsoft.PowerShell.Commands
 
         #region Internal Methods
 
-        internal static bool IsText(string contentType)
+        internal static string GetContentType(HttpResponseMessage response)
         {
-            contentType = GetContentTypeSignature(contentType);
-            return CheckIsText(contentType);
+            // ContentType may not exist in response header.  Return null if not.
+            return response.Content.Headers.ContentType?.MediaType;
         }
 
-        internal static bool IsXml(string contentType)
+        internal static Encoding GetDefaultEncoding()
         {
-            contentType = GetContentTypeSignature(contentType);
-            return CheckIsXml(contentType);
+            return GetEncodingOrDefault((string)null);
         }
 
-        internal static bool IsJson(string contentType)
+        internal static Encoding GetEncoding(HttpResponseMessage response)
         {
-            contentType = GetContentTypeSignature(contentType);
-            return CheckIsJson(contentType);
+            // ContentType may not exist in response header.
+            string charSet = response.Content.Headers.ContentType?.CharSet;
+            return GetEncodingOrDefault(charSet);
         }
 
         internal static Encoding GetEncodingOrDefault(string characterSet)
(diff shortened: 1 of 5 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/ContentHelper.Common.cs
```diff
@@ -26,22 +26,22 @@ internal static partial class ContentHelper
...
+        internal static Encoding GetEncoding(HttpResponseMessage response)
         {
-            contentType = GetContentTypeSignature(contentType);
-            return CheckIsJson(contentType);
+            // ContentType may not exist in response header.
+            string charSet = response.Content.Headers.ContentType?.CharSet;
+            return GetEncodingOrDefault(charSet);
         }
 
         internal static Encoding GetEncodingOrDefault(string characterSet)
```
iSazonov: @markekraus We could consider to change the default to Utf8 in later PR.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -3,28 +3,49 @@
...
-using System.Net.Http;
-using System.Net.Http.Headers;
+using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Xml;
-using System.Collections.Generic;
-using System.Text.RegularExpressions;
-using System.Linq;
+using Microsoft.Win32;
```
iSazonov: Do we really need this?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebResponseObject.Common.cs
```diff
@@ -3,11 +3,11 @@
 --********************************************************************/
 
 using System;
+using System.Collections.Generic;
```
iSazonov: The file is not sorted but I agree to merge "as is" - if we'll have follow formatting PR.

iSazonov (approved):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebResponseObject.Common.cs
markekraus: @iSazonov I'm confused. This file is sorted. What is out of order?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
markekraus: It looks like we don't.  Should I remove it in this PR or in another?

iSazonov: Better in next - the PR is very large :-)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebResponseObject.Common.cs
iSazonov: Line 67.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
markekraus: I have added it to the task list in #5610

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebResponseObject.Common.cs
markekraus: ah. that one can probably be fixed with automatic property `{ get; private set }`, which is on the checklist already.

iSazonov: @markekraus Pow("Please", 10) make the follow PRs as small as possible - split work by file/type formattings. :-)

markekraus: @iSazonov if you have ideas on how the work should be split, please update #5610. I still don't have a good handle on what is too big or too small and, honestly, sometimes the feedback on that is inconsistent.

iSazonov: Maybe interesting https://smartbear.com/learn/code-review/best-practices-for-peer-code-review/

TravisEz13:
1. If you have enabled styleCop, can you enable https://www.codefactor.io on your branch so I can see that your changes improve the style?
1. Such a large amount of code was changed, I have to take your word that no logic was changed.  Can we break this up into smaller PRs to get this done?   The following are ideas on how to break it up:
    - Perhaps merge the partial classes without sorting ( should be easy to review )
    - Sort using statements (should be easy to review)
    - Sort various element attempting to keep under a line limit (I would complete at least one file.)  The link [@iSazonov](https://smartbear.com/learn/code-review/best-practices-for-peer-code-review/) suggested says 400.
    - Sometimes between coworkers, we use commits for breaking up reviews

markekraus:
@TravisEz13 I actually haven't enabled stylecop. I tried but couldn't get it to work. I believe it would apply to the entire utility assembly which would make it very hard to track down just the web cmdlet code issues. This was all somewhat beyond me. So I resorted to sorting the code manually and using the stylecop rules as guidance. We can't use them all as, for example, our contributor guidelines say to group the private backing fields with public properties. that will make it hard to see what private field  and private-before-public sorting issues are actually issues.

I can break it up, yes. I tried breaking up a large change into the multiple commits in a single PR before, but @iSazonov didn't like that. I don't care how the work is broken up, but I could use some guidance, clarity, and consistency. I can be told one thing in a PR/issue, follow that to the letter in a new PR, and then have reviewers telling me to redo it a different way. This pattern has repeated several times and results in a lot of wasted time for all involved.

So, should I break this out into separate PRs? Or should I do multiple commits in a single PR or should I group a few commits in a few PRs? Should I move the files first? start with the partial class merges?

TravisEz13:
@markekraus Can you point me to the previous PR with the objection from @iSazonov ?

markekraus:
https://github.com/PowerShell/PowerShell/pull/5518#issuecomment-346425873

TravisEz13:
I think we should try to take an incremental approach.  One of the reasons to enable codefactor.io on your branch (maybe we should enable it after 6.0.0 is released) is to show that the PR was an overall improvement (did not introduce more issues than in removed,) not to say that you have to fix every issue.   

I reviewed this link (https://smartbear.com/learn/code-review/best-practices-for-peer-code-review/) further and I agree with the line limit and the reason behind it.  I spend an hour reviewing this PR and completed a few hundred lines.

Is there a reason that we cannot do a simple merge of the partial classes without any sorting (perhaps you could sort the using statements?)  This seems like a good starting point.

markekraus:
Right now the pieces of the partial classes are in the same file. What constitutes a simple merge in this case? do I just move that code so it's in the same class brackets at the end of the class? 

One reason why I was merging and sorting at the same time is that the classes are so out of alignment it's hard to even chose a logical place to move the partial code into. I can move the code directly in the end of the larger class, but then that code is all unsorted and that's sort of how it is now, only its in its own partial class block. If i try to move it with similar element types, I might have 2 to 3 places to chose from. If i try to move it to where it _should_ be it will be out of place because the code around it is not where it should be.

TravisEz13:
Another approach would be to focus on a limited number of classes to keep under the line limit.  I see one file that is over the line limit by itself.  I'd say try this approach with the smaller files and we can proceed from there.  @daxian-dbw @adityapatwardhan We should update the guidelines, but there should not be a requirement that a certain change is completed in a single PR unless splitting the change would have a negative effect on the project.

markekraus:
Ok. I'm going to close this PR so I can keep the branch for reference. I turned on codefactor.io, but I'm not real happy with the full read/write access it requires. I will work on smaller chunks per PR.

iSazonov:
@TravisEz13 Good news (although preview only) https://developer.github.com/v3/enterprise-admin/pre_receive_hooks/
Another way is GitHub Webhooks https://developer.github.com/v3/orgs/hooks/

PR #5702, opened 2017-12-15
Fix ValidateSet with generator in a module

### LINKED ISSUE #5661: Getting error "Unable to find type" when using IValidateSetValuesGenerator in a PowerShell module
Steps to reproduce
------------------
1) Create a module file (*.psm1) and add the following code block.
2) Attempt to call Test-ValidateSet -Item 'Hello' from PowerShell

```powershell
class ValidateSetTest : System.Management.Automation.IValidateSetValuesGenerator
{
    [string[]] GetValidValues()
    {
        return 'Hello', 'World'
    }
}

function Test-ValidateSet
{
    [CmdletBinding()]
    param (
        [Parameter(Mandatory = $true)]
        [ValidateSet([ValidateSetTest])]
        [string[]]
        $Item
    )
    $Item
}
```
Expected behavior
-----------------

```none
Hello
```

Actual behavior
---------------

```none
Unable to find type [ValidateSetTest].
At C:\PSTest\PSTest.psm1:1364 char:9
+         [ValidateSet([ValidateSetTest])]
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : InvalidOperation: ([ValidateSet([ValidateSetTest])]:AttributeAst) [], RuntimeException
+ FullyQualifiedErrorId : TypeNotFound
```

Environment data
----------------

```powershell
```

This happens because the defined class is only known to the module. In order for this feature to work currently "Using Module \<ModuleName>" must be called from PowerShell to load the class definitions or a second option is to define a *.ps1 file that contains your class definitions, then in the *.psd1 file, add that script path to the ScriptsToProcess attribute list.

The advantages of modifying the manifest file is that it does not require anything from the end user to get the module to work, however since the class definition lives in an external file, that also means it cannot reference of the internal workings of the module. On the flip side, "using module" allows for the class to remain within the module, but also mean a script must be written or end user action must be taken to bring the class definition into the current session. Additionally this will load any internal class definitions inside the module into their session as well.

Without doing something around these lines, this new feature #3744 cannot be used in modules and only works if you dot-source a script file. As this feature was added to get around having to create dynamic parameters to simply add dynamic ValidateSet values, I feel like modules were the intended use case for this feature as they typically are responsible for exposing multiple functions to the end user whereas scripts as generally just called as a script.

Also note that when using the workaround solutions mentioned above, it also exposes those class definitions to the end users session, something that they really should not care about.

lzybkr:
## PR Summary

The symbol resolver was properly resolving the type to the class
definition, but that resolution was ignored at runtime because we used
the string instead of the ITypeName - the ITypeName kept the reference
to the class definition.

Fix #5661

## PR Checklist

- [NA] User facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission) `NA`
    - [NA] Issue filed - Issue link: `NA`
    - [ ] [Add `[feature]` if the change is significant or affectes feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

### DIFF
--- a/src/System.Management.Automation/engine/parser/Compiler.cs
+++ b/src/System.Management.Automation/engine/parser/Compiler.cs
@@ -1309,14 +1309,16 @@ namespace System.Management.Automation.Language
             // 'ValidateSet([CustomGeneratorType], IgnoreCase=$false)' is supported in scripts.
             if (ast.PositionalArguments.Count == 1 && ast.PositionalArguments[0] is TypeExpressionAst generatorTypeAst)
             {
-                if (TypeResolver.TryResolveType(generatorTypeAst.TypeName.FullName, out Type generatorType))
+                var generatorType = TypeResolver.ResolveITypeName(generatorTypeAst.TypeName, out Exception exception);
+                if (generatorType != null)
                 {
                     result = new ValidateSetAttribute(generatorType);
                 }
                 else
                 {
-                    throw InterpreterError.NewInterpreterException(ast, typeof(RuntimeException), ast.Extent,
-                        "TypeNotFound", ParserStrings.TypeNotFound, generatorTypeAst.TypeName.FullName, typeof(System.Management.Automation.IValidateSetValuesGenerator).FullName);
+                    throw InterpreterError.NewInterpreterExceptionWithInnerException(
+                        ast, typeof(RuntimeException), ast.Extent, "TypeNotFound", ParserStrings.TypeNotFound, exception,
+                        generatorTypeAst.TypeName.FullName, typeof(System.Management.Automation.IValidateSetValuesGenerator).FullName);
                 }
             }
             else


src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -1309,14 +1309,16 @@ private static Attribute NewValidateSetAttribute(AttributeAst ast)
             // 'ValidateSet([CustomGeneratorType], IgnoreCase=$false)' is supported in scripts.
             if (ast.PositionalArguments.Count == 1 && ast.PositionalArguments[0] is TypeExpressionAst generatorTypeAst)
             {
-                if (TypeResolver.TryResolveType(generatorTypeAst.TypeName.FullName, out Type generatorType))
+                var generatorType = TypeResolver.ResolveITypeName(generatorTypeAst.TypeName, out Exception exception);
```
iSazonov: `ResolveITypeName` again check `iTypeName as TypeName` - can we use `internal static Type ResolveTypeName(TypeName typeName, out Exception exception)`?

lzybkr: No, you might have something else like a generic or array type instead of a simple type.

iSazonov: Thanks for clarify!
Should we test these cases too (generic or array)?

lzybkr: The ast must be fully general which explains why you must use ResolveITypeName here.

In this specific case, it is extremely unlikely anyone would ever use a generic (which should work) or an array (which should be an error), so testing has marginal usefulness.

vors (approved):

Issue #5576, opened 2017-11-29
Start-Process doesn't pass arguments with embedded whitespace and double quotes correctly


mklement0:
<sup>Updated later to include the problem with embedded double quotes.</sup>

Steps to reproduce
------------------

Embedded whitespace:

```powershell
'"Hi!"' > './t 1.ps1'; Start-Process -Wait -NoNewWindow pwsh -ArgumentList '-noprofile', '-file', './t 1.ps1'
```

Embedded double quotes:

```powershell
Start-Process -Wait -NoNewWindow pwsh -ArgumentList '-noprofile', '-command', '"Hi!"'
```

Expected behavior
-----------------

In both cases:

```none
Hi!
```

That is, script file `./t 1.ps1` should execute, and double-quoted string literal `"Hi!"` should print.

Actual behavior
---------------

Embedded whitespace:

Invocation fails, because the `./t 1.ps1` is passed as _two_ arguments:

```none
The argument './t' is not recognized as the name of a script file. Check the spelling of the name, or if a path was included, verify that the path is
correct and try again.
```

The only way to make this currently work is to _embed_ (potentially escaped) _double quotes_: `'"./t 1.ps1"'` or ``"`"./t 1.ps1`""``; e.g.:

```powershell
Start-Process -Wait -NoNewWindow pwsh -ArgumentList '-noprofile', '-file', "`"./t 1.ps1`""
```

_Update_: Overall, the **best workaround** is to pass a _single_ string containing _all_ arguments to `-ArgumentList` and use embedded quoting and escaping as necessary:

```powershell
Start-Process -Wait -NoNewWindow pwsh -ArgumentList '-noprofile -file "./t 1.ps1"'
```

Embedded double quotes:

The embedded double quotes are unexpectedly _removed_.

The only way to make this currently work is to `\`-escape the embedded double quotes: 

```powershell
Start-Process -Wait -NoNewWindow pwsh -ArgumentList '-noprofile', '-command', '\"Hi!\"'
```

_Update_: Again, the best workaround is to use a _single_ string:

```powershell
Start-Process -Wait -NoNewWindow pwsh -ArgumentList '-noprofile -command \"Hi!\"'
```

Environment data
----------------

```powershell
PowerShell Core v6.0.0-rc (v6.0.0-rc) on Microsoft Windows 7 Enterprise  (64-bit; v6.1.7601)
[...]

PR #5703, opened 2017-12-15
Fix Start-Process whereby arguments in ArgumentList contain whitespace.

bergmeister:
## PR Summary
Fix Start-Process whereby arguments in ArgumentList contain whitespace by quoting arguments if they contain whitespace to resolve issue #5576 
Please refer to the issue discussion that highlights it might be better to just wait for corefx to implement a new API. Maybe it is better to wait for that instead of applying this improvement (both would be a  breaking change anyway)

## PR Checklist

- [ ] NA User facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - [x] Issue filed - Issue link: https://github.com/PowerShell/PowerShell/issues/5576
- [ ] [Change is not breaking](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes) Breaking change for scripts that use quotes already as a workaround of this issue. An example is the provided workaround in the issue.
    - [ x] [Add `[feature]` if the change is significant or affectes feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr) new test is in feature test suite


test/powershell/Modules/Microsoft.PowerShell.Utility/XMLCommand.Tests.ps1
```diff
@@ -102,9 +102,9 @@ Describe "XmlCommand DRT basic functionality Tests" -Tags "CI" {
 		$deserialized_one.two.three.num | Should BeNullOrEmpty
 	}
 
-	It "Import-Clixml should work with XML serialization from pwsh.exe" {
+	It "Import-Clixml should work with XML serialization from pwsh binary" {
 		# need to create separate process so that current powershell doesn't interpret clixml output
-		Start-Process -FilePath $pshome\pwsh -RedirectStandardOutput $testfile -Args "-noprofile -nologo -outputformat xml -command get-command import-clixml" -Wait
+		Start-Process -FilePath $pshome\pwsh -RedirectStandardOutput $testfile -ArgumentList '-noprofile', '-nologo', '-outputformat', 'xml', '-command', 'get-command import-clixml' -Wait -NoNewWindow
```
bergmeister: This initially failing test was a good example that it is a breaking change.

src/Microsoft.PowerShell.Commands.Management/commands/management/Process.cs
```diff
@@ -1857,10 +1857,22 @@ protected override void BeginProcessing()
             if (ArgumentList != null)
             {
                 StringBuilder sb = new StringBuilder();
-                foreach (string str in ArgumentList)
+                foreach (string argument in ArgumentList)
                 {
-                    sb.Append(str);
-                    sb.Append(' ');
+                    if (argument != null && argument.Contains(" "))
```
bergmeister: I am wondering if this condition needs to be tweaked to also add the clause `&& !argument.StartsWith("\"")` because the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.arguments?view=netcore-2.0#System_Diagnostics_ProcessStartInfo_Arguments) say that 2 double quotes are `interpreted as if the quotes weren't there` and `Triple-escape quotation marks to include the character in the final argument received`. A user does not know that low level detail/quirks of `ProcessStartInfo.Arguments` and it would reduce the number of breaking changes.

test/powershell/Modules/Microsoft.PowerShell.Management/Start-Process.Tests.ps1
```diff
@@ -48,6 +48,17 @@ Describe "Start-Process" -Tags @("Feature") {
 	    $process = Start-Process ping -ArgumentList $pingParam -Wait -PassThru -RedirectStandardOutput "$TESTDRIVE/output"
     }
 
+    # Ensures that issue 5576 remains fixed: https://github.com/PowerShell/PowerShell/issues/5576
+    It "Should handle whitespace in arguments of ArgumentList correctly" {
+        try {
+            '"Hello World!"' > '.\testFile with WhiteSpace In Name.ps1'
```
SteveL-MSFT: You should use TESTDRIVE which will auto-cleanup

SteveL-MSFT (changes requested):

bergmeister:
@SteveL-MSFT Did you also read the discussion in the original issue 5576? It seems that it might be possible that .Net Core gets a new overload. What are your thoughts on this?

SteveL-MSFT:
@bergmeister forgot about that discussion.  I agree that we should wait for the change in corefx.

PR #5732, opened 2017-12-21
Fix range operator

### LINKED ISSUE #5519: Range operator (aka DotDot operator) with char operands in pipeline throws error
Steps to reproduce
------------------

```'
A'..'B'
'A'..'B' | ForEach-Object { $_ }
```

Expected behavior
-----------------

```
A
B
```

Actual behavior
---------------
Cannot convert value "A" to type "System.Int32". Error: "Input string was not in a correct format."
At line:1 char:1
+ 'A'..'B' | ForEach-Object { $_ }
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ FullyQualifiedErrorId : InvalidCastFromStringToInteger```

Environment data
----------------

```Name                           Value
WSManStackVersion              3.0```

### LINKED ISSUE #5731: Range operator with char returns unexpected value
Steps to reproduce
------------------

```powershell
[char]'a'..[char]'e'
```

Expected behavior
-----------------

```powershell
# Char array ([char]'a', [char]'b', [char]'c', [char]'d', [char]'e')
a
b
c
d
e
```

Actual behavior
---------------

```powershell
# Int array (97, 98, 99, 100, 101)
97
98
99
100
101
```

Environment data
----------------

```powershell
> $PSVersionTable

```

iSazonov:
## PR Summary
Fix range operator enumerator. Close #5519
Fix [char] cast in range operator. Close #5731

Breaking-change:  "0".."9" returns [char] previously in PSCore, now it returns [int]. After the change, the behavior is the same as in Windows PowerShell.

## PR Checklist

    - [NA] [Add `[feature]` if the change is significant or affectes feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

### DIFF
--- a/src/System.Management.Automation/engine/lang/parserutils.cs
+++ b/src/System.Management.Automation/engine/lang/parserutils.cs
@@ -1478,7 +1478,13 @@ namespace System.Management.Automation
         }
 
         private int _current;
-        public object Current
+
+        Object IEnumerator.Current
+        {
+            get { return Current; }
+        }
+
+        public virtual int Current
         {
             get { return _current; }
         }
@@ -1521,6 +1527,81 @@ namespace System.Management.Automation
             return true;
         }
     }
+
+    /// <summary>
+    /// The simple enumerator class is used for the range operator '..'
+    /// in expressions like 'A'..'B' | ForEach-Object { $_ }
+    /// </summary>
+    internal class CharRangeEnumerator : IEnumerator
+    {
+        private int _lowerBound;
+
+        internal int LowerBound
+        {
+            get { return _lowerBound; }
+        }
+
+        private int _upperBound;
+
+        internal int UpperBound
+        {
+            get { return _upperBound; }
+        }
+
+        private int _current;
+
+        Object IEnumerator.Current
+        {
+            get { return Current; }
+        }
+
+        public string Current
+        {
+            get { return new String((char)_current, 1); }
+        }
+
+        internal int CurrentValue
+        {
+            get { return _current; }
+        }
+
+        private int _increment = 1;
+
+        private bool _firstElement = true;
+
+        public CharRangeEnumerator(char lowerBound, char upperBound)
+        {
+            _lowerBound = lowerBound;
+            _current = _lowerBound;
+            _upperBound = upperBound;
+            if (lowerBound > upperBound)
+                _increment = -1;
+        }
+
+        public void Reset()
+        {
+            _current = _lowerBound;
+            _firstElement = true;
+        }
+
+        public bool MoveNext()
+        {
+            if (_firstElement)
+            {
+                _firstElement = false;
+                return true;
+            }
+
+            if (_current == _upperBound)
+            {
+                return false;
+            }
+
+            _current += _increment;
+            return true;
+        }
+    }
+
     #endregion RangeEnumerator
 
     #region InterpreterError
--- a/src/System.Management.Automation/engine/lang/parserutils.cs
+++ b/src/System.Management.Automation/engine/lang/parserutils.cs
@@ -863,7 +863,6 @@ namespace System.Management.Automation
         // and so on.
         //
         // In special case:
-        //     [char]'0'..[char]'9'
         //     "0".."9"
         // return objects of [int] type.
         private static object AsChar(object obj)
--- a/src/System.Management.Automation/engine/lang/parserutils.cs
+++ b/src/System.Management.Automation/engine/lang/parserutils.cs
@@ -786,6 +786,35 @@ namespace System.Management.Automation
         }
 
         /// <summary>
+        /// The implementation of the PowerShell range operator.
+        /// </summary>
+        /// <param name="context">The execution context in which to evaluate the expression</param>
+        /// <param name="errorPosition">The position to use for error reporting.</param>
+        /// <param name="lval">The object on which to replace the values</param>
+        /// <param name="rval">The replacement description.</param>
+        /// <returns>The result of the operator</returns>
+        internal static object RangeOperator(ExecutionContext context, IScriptExtent errorPosition, object lval, object rval)
+        {
+            var lbase = PSObject.Base(lval);
+            var rbase = PSObject.Base(rval);
+
+            try
+            {
+                var l = Convert.ToChar(lbase);
+                var r = Convert.ToChar(rbase);
+
+                return CharOps.Range(l, r);
+            }
+            catch
+            {
+                var l = Convert.ToInt32(lbase);
+                var r = Convert.ToInt32(rbase);
+
+                return IntOps.Range(l, r);
+            }
+        }
+
+        /// <summary>
         /// The implementation of the PowerShell -replace operator....
         /// </summary>
         /// <param name="context">The execution context in which to evaluate the expression</param>
--- a/src/System.Management.Automation/engine/lang/parserutils.cs
+++ b/src/System.Management.Automation/engine/lang/parserutils.cs
@@ -788,12 +788,10 @@ namespace System.Management.Automation
         /// <summary>
         /// The implementation of the PowerShell range operator.
         /// </summary>
-        /// <param name="context">The execution context in which to evaluate the expression.</param>
-        /// <param name="errorPosition">The position to use for error reporting.</param>
         /// <param name="lval">The object on which to start.</param>
         /// <param name="rval">The object on which to stop.</param>
         /// <returns>The array of objects.</returns>
-        internal static object RangeOperator(ExecutionContext context, IScriptExtent errorPosition, object lval, object rval)
+        internal static object RangeOperator(object lval, object rval)
         {
             var lbase = PSObject.Base(lval);
             var rbase = PSObject.Base(rval);
--- a/src/System.Management.Automation/engine/lang/parserutils.cs
+++ b/src/System.Management.Automation/engine/lang/parserutils.cs
@@ -788,30 +788,89 @@ namespace System.Management.Automation
         /// <summary>
         /// The implementation of the PowerShell range operator.
         /// </summary>
-        /// <param name="context">The execution context in which to evaluate the expression</param>
+        /// <param name="context">The execution context in which to evaluate the expression.</param>
         /// <param name="errorPosition">The position to use for error reporting.</param>
-        /// <param name="lval">The object on which to replace the values</param>
-        /// <param name="rval">The replacement description.</param>
-        /// <returns>The result of the operator</returns>
+        /// <param name="lval">The object on which to start.</param>
+        /// <param name="rval">The object on which to stop.</param>
+        /// <returns>The array of objects.</returns>
         internal static object RangeOperator(ExecutionContext context, IScriptExtent errorPosition, object lval, object rval)
         {
             var lbase = PSObject.Base(lval);
             var rbase = PSObject.Base(rval);
 
-            try
-            {
-                var l = Convert.ToChar(lbase);
-                var r = Convert.ToChar(rbase);
+            var lType = lbase.GetType();
+            var rType = rbase.GetType();
 
-                return CharOps.Range(l, r);
-            }
-            catch
+            // Return objects of [char] type for explicit [char] cast.
+            // Common case:
+            //     [char]'A'..[char]'Z'
+            // Special case:
+            //     [char]'0'..[char]'9'
+            if (lType == typeof(char) && rType == typeof(char))
             {
-                var l = Convert.ToInt32(lbase);
-                var r = Convert.ToInt32(rbase);
-
-                return IntOps.Range(l, r);
+                return CharOps.Range((char)lbase, (char)rbase);
             }
+
+            // For arguments of string type and length == 1
+            // in the common case:
+            //     "A".."Z" - return objects of [char] type.
+            //
+            // For arguments of string type and length == 1
+            // in the special case:
+            //     "0".."9" - return objects of [int] type
+            if (lType == typeof(string) && rType == typeof(string))
+            {
+                var lstr = (string)lbase;
+                var rstr = (string)rbase;
+                if (lstr.Length == 1 && lstr.Length == 1 && !Char.IsDigit(lstr, 0) && !Char.IsDigit(rstr, 0))
+                {
+                    return CharOps.Range(lstr[0], rstr[0]);
+                }
+            }
+
+            // As a last resort, the range operator tries to return objects of [int] type.
+            //    1..10
+            //    "1".."10"
+            //    [int]"1"..[int]"10"
+            var l = Convert.ToInt32(lbase);
+            var r = Convert.ToInt32(rbase);
+
+            return IntOps.Range(l, r);
+        }
+
+        /// <summary>
+        /// The implementation of an enumerator for the PowerShell range operator.
+        /// </summary>
+        /// <param name="lval">The object on which to start.</param>
+        /// <param name="rval">The object on which to stop.</param>
+        /// <returns>The enumerator.</returns>
+        internal static IEnumerator GetRangeEnumerator(object lval, object rval)
+        {
+            var lbase = PSObject.Base(lval);
+            var rbase = PSObject.Base(rval);
+
+            var lType = lbase.GetType();
+            var rType = rbase.GetType();
+
+            if (lType == typeof(char) && rType == typeof(char))
+            {
+                return new CharRangeEnumerator((char)lval, (char)rval);
+            }
+
+            if (lType == typeof(string) && rType == typeof(string))
+            {
+                var lstr = (string)lbase;
+                var rstr = (string)rbase;
+                if (lstr.Length == 1 && lstr.Length == 1 && !Char.IsDigit(lstr, 0) && !Char.IsDigit(rstr, 0))
+                {
+                    return new CharRangeEnumerator(lstr[0], rstr[0]);
+                }
+            }
+
+            var l = Convert.ToInt32(lbase);
+            var r = Convert.ToInt32(rbase);
+
+            return new RangeEnumerator(l, r);
         }
 
         /// <summary>
--- a/src/System.Management.Automation/engine/lang/parserutils.cs
+++ b/src/System.Management.Automation/engine/lang/parserutils.cs
@@ -798,34 +798,12 @@ namespace System.Management.Automation
             var lbase = PSObject.Base(lval);
             var rbase = PSObject.Base(rval);
 
-            var lType = lbase.GetType();
-            var rType = rbase.GetType();
-
-            // Return objects of [char] type for explicit [char] cast.
-            // Common case:
-            //     [char]'A'..[char]'Z'
-            // Special case:
-            //     [char]'0'..[char]'9'
-            if (lType == typeof(char) && rType == typeof(char))
+            // If both arguments is [char] type or [string] type with length==1
+            // return objects of [char] type.
+            // In special case "0".."9" return objects of [int] type.
+            if (AsChar(lbase) is char lc && AsChar(rbase) is char rc)
             {
-                return CharOps.Range((char)lbase, (char)rbase);
-            }
-
-            // For arguments of string type and length == 1
-            // in the common case:
-            //     "A".."Z" - return objects of [char] type.
-            //
-            // For arguments of string type and length == 1
-            // in the special case:
-            //     "0".."9" - return objects of [int] type
-            if (lType == typeof(string) && rType == typeof(string))
-            {
-                var lstr = (string)lbase;
-                var rstr = (string)rbase;
-                if (lstr.Length == 1 && lstr.Length == 1 && !Char.IsDigit(lstr, 0) && !Char.IsDigit(rstr, 0))
-                {
-                    return CharOps.Range(lstr[0], rstr[0]);
-                }
+                return CharOps.Range(lc, rc);
             }
 
             // As a last resort, the range operator tries to return objects of [int] type.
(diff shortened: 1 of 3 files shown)


src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -1521,6 +1527,81 @@ public bool MoveNext()
...
+        private int _upperBound;
+
+        internal int UpperBound
+        {
+            get { return _upperBound; }
+        }
+
+        private int _current;
+
+        Object IEnumerator.Current
```
markekraus: Public properties should go before Internal properties https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1202.md

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -1521,6 +1527,81 @@ public bool MoveNext()
...
+        {
+            get { return Current; }
+        }
+
+        public string Current
```
markekraus: Public properties should go before Internal properties https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1202.md

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -1521,6 +1527,81 @@ public bool MoveNext()
...
+        {
+            get { return new String((char)_current, 1); }
+        }
+
+        internal int CurrentValue
+        {
+            get { return _current; }
+        }
+
+        private int _increment = 1;
```
markekraus: Fields not backing public properties should be defined before properties . https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1202.md

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -1478,7 +1478,13 @@ internal int UpperBound
         }
 
         private int _current;
-        public object Current
+
+        Object IEnumerator.Current
```
markekraus: This should swap places with `private int _current;` so that the field can be closed to the property it backs.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -1521,6 +1527,81 @@ public bool MoveNext()
             return true;
         }
     }
+
+    /// <summary>
+    /// The simple enumerator class is used for the range operator '..'
+    /// in expressions like 'A'..'B' | ForEach-Object { $_ }
+    /// </summary>
+    internal class CharRangeEnumerator : IEnumerator
```
markekraus: I'm curious if we should start following the one-type-per-file rule https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1402.md

markekraus (approved):
Style comments only. Not blocking.

src/System.Management.Automation/engine/lang/parserutils.cs
lzybkr: It is often useful to have related types in the same file, so I'm strongly opposed to following this rule. I'm strongly in favor of organizing code in ways that make sense to the author and reviewers. We shouldn't blindly follow the guidelines of some tool.

src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -4295,9 +4298,18 @@ private Expression GetRangeEnumerator(ExpressionAst condExpr)
                     Expression lhs = Compile(binaryExpr.Left);
                     Expression rhs = Compile(binaryExpr.Right);
 
-                    result = Expression.New(CachedReflectionInfo.RangeEnumerator_ctor,
-                                            lhs.Convert(typeof(int)),
-                                            rhs.Convert(typeof(int)));
+                    if (lhs.Type == typeof(string) || lhs.Type == typeof(Char))
```
lzybkr: This doesn't support variables, e.g.
```powershell
$a = 'a'
$z = 'z'
$a..$z
```

src/System.Management.Automation/engine/lang/parserutils.cs
markekraus: I agree that it make sense to group related code and that following the rules blindly is silly. But, some of the files in this project are so massive that there is 0 benefit gained by the code being grouped, IMO. If I need to have 3 tabs open on the same file to work on it effectively, the code is probably better split, IMO. Doesn't have to be strictly followed, but, maybe give some serious consideration?

src/System.Management.Automation/engine/parser/Compiler.cs
iSazonov: Added support variables.

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: @markekraus I reformatted the class.

iSazonov:
MacOS CI temporary failed.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -1521,6 +1556,81 @@ public bool MoveNext()
...
+    {
+        private int _increment = 1;
+
+        private bool _firstElement = true;
+
+        public CharRangeEnumerator(char lowerBound, char upperBound)
```
markekraus: This should probably go with the other methods in the class.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -1521,6 +1556,81 @@ public bool MoveNext()
             return true;
         }
     }
+
+    /// <summary>
+    /// The simple enumerator class is used for the range operator '..'
+    /// in expressions like 'A'..'B' | ForEach-Object { $_ }
+    /// </summary>
+    internal class CharRangeEnumerator : IEnumerator
```
markekraus: Should this implement `IEnumerable<String>` instead?

markekraus:
@iSazonov I restarted the macOS job for you.

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: I think we cast all to `object` type before send to pipe. So it seems there is no sense to specify here the exact type.

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: I use your link https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1201.md
 - constructors is after fields.

markekraus: Ah I missed that this was the constructor... My bad

vors:
On this topic, it would be nice, if we can utilize new C# Slice syntax someday for perf reasons.

src/System.Management.Automation/engine/lang/parserutils.cs
lzybkr: Are you suggesting the author of said code saw 0 benefit of a huge file, because you might be saying that to said author. (I'm the primary author of 4 of the 10 largest non-generated files in the repo.)

To pick on the largest file - Ast.cs - there are at least 72 classes in that file. When I know I'm looking for something Ast related, but not sure what, if I end up with 72 open tabs that I need to close, I'm less productive than if I open 1 file.

And if you needed 3 tabs open to effectively work on some code, you needed 3 tabs to see some code, 1 file or 3 files, right? I do realize 

There is always room for improvement, but it should be thoughtful and not rely on some tool - tools are good at enforcing things that are easy to check, less good at providing good judgement.

src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -4295,12 +4299,24 @@ private Expression GetRangeEnumerator(ExpressionAst condExpr)
                     Expression lhs = Compile(binaryExpr.Left);
                     Expression rhs = Compile(binaryExpr.Right);
 
-                    result = Expression.New(CachedReflectionInfo.RangeEnumerator_ctor,
-                                            lhs.Convert(typeof(int)),
-                                            rhs.Convert(typeof(int)));
+                    if ((lhs.Type == typeof(string) || lhs.Type == typeof(char))
```
lzybkr: I think this will break a weird but supported scenario:
```powershell
"0".."10"
```
I know it's dumb code, but I've seen it often enough that I wouldn't break this case.

src/System.Management.Automation/engine/lang/parserutils.cs
markekraus: >Are you suggesting the author of said code saw 0 benefit of a huge file,

Not at all. I'm suggesting *I* see 0 benefit.

> because you might be saying that to said author.

I know 🙂 .

> And if you needed 3 tabs open to effectively work on some code, you needed 3 tabs to see some code, 1 file or 3 files, right?

Right, I'm saying I see no benefit to these being in the same file, where I do see benefit in it being in separate files. I think it's a difference of preference, I prefer smaller files grouped in folders over monolith files. It's easier to jump between code in different files using the IDE, IMO. When the code is in the same file, the default action is to jump to the code in the current tab. When it is in another file, a new tab opens with that file and my current place in the previous file is preserved. *shrugs

I'm also trying to approach this from a maintainable open source project perspective. If this rule doen't makes sense for this project, then it doesn't make sense and we don't use it. I just want to have something in place that can help maintain and encourage an enforceable style for the life of the project, preferably through CI. It can help shift the burden from reviewers to contributors. 

If I'm wasting my efforts there, then I will gladly redirect my attention elsewhere.

src/System.Management.Automation/engine/parser/Compiler.cs
iSazonov: 😄 I wonder why do we have CIs if we have Jason Shirk?

While I don't know how fix this here. And I see the same problem in line 4893. I guess we can not resolve the conflict in `RangeOperator` method: "0".."9", $var1.."9", "0"..$var2 - covert string to char or int? What if user want get char array? 😕 

Another question. In line 4893 I tried to make this code faster for constants. It seems that with this addition it can becomes too complicated and duplicated. I think to leave only Expression.Call() and put all magic code in one method. Right?
I see all `Expression.Call()` has a ToDo comment ("replace this with faster code") - could you please explain the idea? Maybe we can make more right thing now?

lzybkr: Assume any string with Length > 1 or if the only character is 0-9, assume int, otherwise char - that should be sufficient.

The `TODO` was written when I first wrote that code and just getting things working. I don't remember what I had in mind, if anything.

It's your call on how you want to write the code - you can avoid some duplication with a helper method if you prefer to keep the optimization.

iSazonov:
@lzybkr @SteveL-MSFT Could you please review?

SteveL-MSFT (changes requested):
Looks good.  However, I think for the letter range tests, it would be better to go A..C to ensure B gets generated rather than just the two ends for A..B

SteveL-MSFT (approved):

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -785,6 +785,94 @@ internal static object JoinOperator(ExecutionContext context, IScriptExtent erro
...
+
+            var lType = lbase.GetType();
+            var rType = rbase.GetType();
+
+            // Return objects of [char] type for explicit [char] cast.
+            // Common case:
+            //     [char]'A'..[char]'Z'
+            // Special case:
+            //     [char]'0'..[char]'9'
+            if (lType == typeof(char) && rType == typeof(char))
```
lzybkr: You can avoid the calls to GetType() and casts by using C#7 pattern matching like this:
```C#
if (lbase as char lc && rbase as char rc)
{
    return CharOps.Range(lc, rc);
}
```

The same would apply below for strings.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -785,6 +785,94 @@ internal static object JoinOperator(ExecutionContext context, IScriptExtent erro
...
+            }
+
+            // For arguments of string type and length == 1
+            // in the common case:
+            //     "A".."Z" - return objects of [char] type.
+            //
+            // For arguments of string type and length == 1
+            // in the special case:
+            //     "0".."9" - return objects of [int] type
+            if (lType == typeof(string) && rType == typeof(string))
```
lzybkr: It might nice to support:

```
[char]"a" .. "z"
```

It would be less useful to have `"a"..[char]"z"`, but it's not complicated to support both.

You could introduce a local helper function:
```c#
object AsChar(object o)
{
    if (o is char) return o;
    if (o as string s && s.Length == 1 && !Char.IsDigit(s, 0)) return o[0];
    return null;
}

if (AsChar(lbase) as char lc && AsChar(rbase) as char rc)
{
    return CharOps.Range(lc, rc);
}
```

This covers all combos of char/char,string/string, and mixed. I think - browser compiled so may not work :)

src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -4874,14 +4871,11 @@ public object VisitBinaryExpression(BinaryExpressionAst binaryExpressionAst)
...
                 case TokenKind.DotDot:
-                    if(lhs.Type == typeof(string)){
-                        return Expression.Call(CachedReflectionInfo.CharOps_Range,
-                                               lhs.Convert(typeof(char)),
-                                               rhs.Convert(typeof(char)));
-                    }
-                    return Expression.Call(CachedReflectionInfo.IntOps_Range,
-                                           lhs.Convert(typeof(int)),
-                                           rhs.Convert(typeof(int)));
+                    // TODO: replace this with faster code
```
lzybkr: The idea behind "faster" is to avoid extra runtime type checks.

The general pattern using `Expression.Dynamic` is to generate code like (for a binary operator like this):

```
if (lhs.GetType() == int && rhs.GetType() == int))
    return IntOps.Range((int)lhs, (int)rhs);
Find some other binding
```

If there is no binding that works, the binder generates a new site, perhaps to throw an error, or add conversions, or whatever. But the key thing here is there are no extra checks in the ideal scenario, if the lhs and rhs don't change types the next time you execute the same line in script, you have exactly the code you need.

The assumption is types don't change much, but you need to generate code that can handle that.

At any rate, it's probably not that important here, but I thought I'd expand on why the code can be made faster.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -785,6 +785,94 @@ internal static object JoinOperator(ExecutionContext context, IScriptExtent erro
...
+        /// <returns>The enumerator.</returns>
+        internal static IEnumerator GetRangeEnumerator(object lval, object rval)
+        {
+            var lbase = PSObject.Base(lval);
+            var rbase = PSObject.Base(rval);
+
+            var lType = lbase.GetType();
+            var rType = rbase.GetType();
+
+            if (lType == typeof(char) && rType == typeof(char))
```
lzybkr: My suggestions for C#7 pattern matching apply here as well.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -1521,6 +1615,81 @@ public bool MoveNext()
...
+        }
+
+        Object IEnumerator.Current
+        {
+            get { return Current; }
+        }
+
+        internal int LowerBound
+        {
+            get { return _lowerBound; }
```
lzybkr: Can you use an auto-property instead?
```
internal int LowerBound
{
    get; private set;
}
```

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -1521,6 +1615,81 @@ public bool MoveNext()
             return true;
         }
     }
+
+    /// <summary>
+    /// The simple enumerator class is used for the range operator '..'
+    /// in expressions like 'A'..'B' | ForEach-Object { $_ }
+    /// </summary>
+    internal class CharRangeEnumerator : IEnumerator
```
lzybkr: The code is basically the same as RangeEnumerator, I wonder if there is a clever way we could avoid the duplication.

One thought - you could maybe derive from RangeEnumerator and just override Current.

It's probably not worth spending time on this though - C# doesn't handle this scenario nicely like C++ template would.

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: Thanks for great comment!

Done.

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: Now all properties is auto-property.

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: I started with this but it looked too confusing. In addition, I thought that if we wanted to enhance this operator in the future, it would be easier if each type would have its own enumerator. Also possible we can make optimization and reduce string allocations.

src/System.Management.Automation/engine/parser/Compiler.cs
iSazonov: Thanks for the deep explaination!

So if we cast all to `object` here we generate the code only one time? And now we delegate type checks to `AsChar()`. If we could move these checks from the function to this level (line 4875), then we could generate more specific sites and be faster, but I don't know if it is possible and reasonable.

lzybkr: Definitely possible and should be faster, but it might only be worthwhile as an exercise in learning how `Expression.Dynamic` and call sites work.

src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -292,6 +286,10 @@ internal static class CachedReflectionInfo
             typeof(ParserOps).GetMethod(nameof(ParserOps.LikeOperator), staticFlags);
         internal static readonly MethodInfo ParserOps_MatchOperator =
             typeof(ParserOps).GetMethod(nameof(ParserOps.MatchOperator), staticFlags);
+        internal static readonly MethodInfo ParserOps_RangeOperator =
+            typeof(ParserOps).GetMethod(nameof(ParserOps.RangeOperator), staticFlags);
+        internal static readonly MethodInfo ParserOps_RangeEnumerator =
```
lzybkr: There is a convention here - the field name is class_member - so this should be ParserOps_GetRangeEnumerator.

src/System.Management.Automation/engine/lang/parserutils.cs
lzybkr: Sorry, I don't see that change.

iSazonov: My bad - lost commit.

Fixed.

src/System.Management.Automation/engine/parser/Compiler.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -1478,7 +1567,13 @@ internal int UpperBound
         }
 
         private int _current;
-        public object Current
+
+        Object IEnumerator.Current
```
lzybkr: Can you explain why this method was added? It seems like we only need the explicit method impl or the other, but not both.

iSazonov: I saw it in CoreFX implementations ( [sample](https://github.com/dotnet/corefxlab/blob/b5a998d2758b8ee968dad324f104c9310057eda3/src/System.IO.Pipelines.Text.Primitives/SplitEnumerator.cs#L30) ). I don't remember exactly - it seems I'm checking that it doesn't work at all without this method if we use enheritance and call by IEnumerator interface.

iSazonov: @lzybkr What is your conclusion?

lzybkr: The example you point to is needed to implement the strongly typed interface `IEnumerable<T>`. If you wanted to follow that pattern, that's fine, but you would replace the `CurrentValue` property with the `Current` property.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -1521,6 +1616,75 @@ public bool MoveNext()
...
+            LowerBound = lowerBound;
+            CurrentValue = LowerBound;
+            UpperBound = upperBound;
+            if (lowerBound > upperBound)
+                _increment = -1;
+        }
+
+        public string Current
+        {
+            get { return new String((char)CurrentValue, 1); }
```
lzybkr: Why convert to string? I was thinking the output should be char.

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: Yes, it is _Char_ enumerator. 

Fixed.

iSazonov:
@lzybkr Do we ready to merge?

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -785,6 +785,95 @@ internal static object JoinOperator(ExecutionContext context, IScriptExtent erro
...
+        //     "A".."Z"
+        //     [char]"A"..[string]"Z"
+        //     "A"..[char]"Z"
+        //     [string]"A".."Z"
+        // and so on.
+        //
+        // In special case:
+        //     [char]'0'..[char]'9'
+        //     "0".."9"
+        // return objects of [int] type.
```
lzybkr: This comment only applies to the later case, not the former, right?

I don't recall the discussion around this, but I have a feeling we'll regret going this way despite it being a breaking change.

iSazonov: Yes, I removed ` [char]'0'..[char]'9'`

iSazonov:
@daxian-dbw Could you please fast review before I merge?

daxian-dbw:
@iSazonov I will quickly go through the changes later this morning.

daxian-dbw (dismissed):

src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -4286,21 +4281,23 @@ public object VisitForEachStatement(ForEachStatementAst forEachStatementAst)
...
-            Expression result = null;
             if (condExpr != null)
             {
                 var binaryExpr = condExpr as BinaryExpressionAst;
                 if (binaryExpr != null && binaryExpr.Operator == TokenKind.DotDot)
                 {
                     Expression lhs = Compile(binaryExpr.Left);
                     Expression rhs = Compile(binaryExpr.Right);
+                    var l = lhs.Cast(typeof(object));
+                    var r = rhs.Cast(typeof(object));
```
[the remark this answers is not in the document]
iSazonov: Fixed.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -785,6 +785,94 @@ internal static object JoinOperator(ExecutionContext context, IScriptExtent erro
...
 
+        /// <summary>
+        /// The implementation of the PowerShell range operator.
+        /// </summary>
+        /// <param name="context">The execution context in which to evaluate the expression.</param>
+        /// <param name="errorPosition">The position to use for error reporting.</param>
+        /// <param name="lval">The object on which to start.</param>
+        /// <param name="rval">The object on which to stop.</param>
+        /// <returns>The array of objects.</returns>
+        internal static object RangeOperator(ExecutionContext context, IScriptExtent errorPosition, object lval, object rval)
```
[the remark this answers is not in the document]
iSazonov: All other help methods use the pattern.
I can remove if you request.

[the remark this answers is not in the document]
daxian-dbw: > All other help methods use the pattern.

If the arguments are not used in those help methods, they should also be removed. But you don't need to do that in this PR. Just fix the method added in this PR.

src/System.Management.Automation/engine/parser/Compiler.cs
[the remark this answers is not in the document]
daxian-dbw: I saw that discussion. The idea should be captured in the comment.

src/System.Management.Automation/engine/lang/parserutils.cs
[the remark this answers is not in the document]
iSazonov: Fixed.

src/System.Management.Automation/engine/parser/Compiler.cs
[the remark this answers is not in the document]
iSazonov: The comment added.

daxian-dbw:
@iSazonov According to https://github.com/PowerShell/PowerShell/pull/5732#discussion_r158558312, it seems to me you have fixed the breaking change. Is there still a breaking change introduced by this PR?

iSazonov:
@daxian-dbw I modify the PR description: "Add support for special case "0".."9" - return [int] objects."

@lzybkr Have you any thoughts or we can merge?

daxian-dbw (approved):

daxian-dbw:
@iSazonov I updated PR description a bit to make the breaking change more clear.
When squash merging, please make sure add `[breaking-change]` in the commit description, and summarize the breaking change.

iSazonov:
@lzybkr @daxian-dbw Thanks for help and great comments!

daxian-dbw:
@iSazonov The `[breaking-change]` tag should be added to the body of the commit message instead of the title of it. Sorry that I didn't point out that specifically.

When squash-merging a breaking change PR in future, please do like the following:
image

PR #5759, opened 2017-12-29
Pull PSReadLine from PSGallery

### LINKED ISSUE #996: Export PSReadLine to changes to public project
This is just so I remember to do this for @lzybkr.

lzybkr:
## PR Summary
Instead of building PSReadLine from this repo, pull it from the gallery.

This pulls v2.0 of PSReadLine which does have documented breaking changes from v1.2, but the risk is small - the features that have changed are typically only used in a profile and aren't used all that often anyway.

Fix #996

Hardcodes version of modules pulled from PSGallery

## PR Checklist

- [x] **NA** User facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - [x] **NA** Issue filed - Issue link:
- [x] **NA** [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
    - [x] **NA** [Add `[feature]` if the change is significant or affectes feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)


lzybkr:
I tested these changes locally and had no problems building (I didn't run tests though.)

The CI failures show 2 problems:

1. If restoring a modules fails, tests probably should not run.
1. We rely on `PowerShellGet` to build. This is unnecessary, we could use `nuget` directly.

I'd like some feedback on `2` - should we update `PowerShellGet` as part of the build, or just use `nuget`.

build.psm1
```diff
@@ -2514,6 +2514,7 @@ function Restore-PSModule
                         Name=$_
                         Path = $Destination
                         Repository =$RepositoryName
+                        AllowPreRelease = $true
```
TravisEz13: This is causing failures

lzybkr: Yes, I know, see my request for feedback [here](https://github.com/PowerShell/PowerShell/pull/5759#issuecomment-354506893).

TravisEz13:
It is very difficult to use NuGet directly across platforms.  Please update `PowerShellGet`

SteveL-MSFT:
I think we can take this early for 6.1.0, but we should stick with the released version of PSReadline for 6.0.0-GA

build.psm1
TravisEz13: Also, We should only do this if needed.

lzybkr:
@TravisEz13 - I tried updating PowerShellGet, but in the end it was easier to use nuget (indirectly, via a csproj) and copy out of the nuget cache.

It looks like everything passes now, so I think this can be merged.

Note that PSReadLine will no longer be cross-gen'd after this PR, so if you merge, I'd open an issue to track adding a call to crossgen for PSReadLine - I'm not sure when I'll get a chance to do that.

build.psm1
```diff
@@ -2452,102 +2458,57 @@ function Restore-GitModule
...
-            log "Unregistering PSRepository with name: $RepositoryName"
-            PowerShellGet\UnRegister-PSRepository -Name $RepositoryName
-        }
+        # Nuget seems to always use lowercase in the cache
+        $src = "$nugetCache/$($name.ToLower())/$srcVer"
+        $dest = "$Destination/$name/$destVer"
+
+        Remove-Item -Force -ErrorAction Ignore -Recurse "$Destination/$name"
+        New-Item -Path $dest -ItemType Directory -Force -ErrorAction Stop > $null
+        $dontCopy = '*.nupkg', '*.nupkg.sha512', '*.nuspec', 'System.Runtime.InteropServices.RuntimeInformation.dll'
```
TravisEz13: This is replicating powershellget behavior which may change over time.  I'd rather not do this if we don't have to.

build.psm1
```diff
@@ -646,17 +648,21 @@ function Restore-PSModuleToBuild
...
 
     $modulesDir = Join-Path -Path $publishPath -ChildPath "Modules"
 
-    # Restore modules from powershellgallery feed
-    Restore-PSModule -Destination $modulesDir -Name @(
-        # PowerShellGet depends on PackageManagement module, so PackageManagement module will be installed with the PowerShellGet module.
+    # These modules are restored to the global nuget package cache at the
+    # same time we restore nuget packages to build PowerShell, we just
+    # copy them from the cache to the final layout location now.
+    $modules = @(
```
daxian-dbw: `$modules` seems not used anywhere.

TravisEz13:
What was the issue with PowerShellGet?  Can we investigate and fix that issue?

src/Microsoft.PowerShell.SDK/Microsoft.PowerShell.SDK.csproj
```diff
@@ -12,6 +12,7 @@
     <ProjectReference Include="..\Microsoft.PowerShell.ConsoleHost\Microsoft.PowerShell.ConsoleHost.csproj" />
     <ProjectReference Include="..\Microsoft.PowerShell.Security\Microsoft.PowerShell.Security.csproj" />
     <ProjectReference Include="..\System.Management.Automation\System.Management.Automation.csproj" />
+    <ProjectReference Include="..\Modules\PSGalleryModules.csproj" />
```
daxian-dbw: When building nuget packages for individual assemblies, I think this reference will cause a nuget package to be generated for `PSGalleryModules.csproj` and the generated `Microsoft.PowerShell.SDK` package will have a reference to the `PSGalleryModules` nuget package. We should avoid creating the unneeded `PSGalleryModules` nuget package.

build.psm1
lzybkr: Yeah, I get that, but this was easier to make work and has less impact on the user's environment or build machine.

build.psm1
lzybkr: Right, I'll remove it.

lzybkr:
PowerShellGet would update successfully, but I would get errors from PackageManagement, presumably because it is a binary module and the newly installed version wasn't being picked up.

I considered "fixing" that by running the update in a child process, but:
* Running in a child process can be harder to debug, harder to pass parameters (no live objects e.g.)
* I just think it's cleaner to use nuget for the package restore, e.g. the packages are cached, so you don't pull the package every time you build.

build.psm1
lzybkr: Also notice how much less code there is doing it this way - I think that's valuable too.

src/Microsoft.PowerShell.SDK/Microsoft.PowerShell.SDK.csproj
lzybkr: Indeed, I'll remove this and do the restore on PSGalleryModules.csproj explicitly.

lzybkr:
I updated to address the feedback.

I don't plan on changing back to using PowerShellGet - I think this approach could be turned on by default - we can rely on the nuget cache instead of fetching from the network on every build.

Also regarding this approach versus using PowerShellGet - in my opinion:
* Either way, we need some logic to remove something, PowerShellGet artifacts or nuget artifacts
* The build fails quickly this way if the package is missing - the previous way the build succeeds and you don't discover a missing module until running tests, assuming there are tests.
* With a csproj, the dependencies are easier to express, maybe not in the SDK csproj, but the dependency is more discoverable than being hidden in build.psm1.

daxian-dbw (approved):

adityapatwardhan:
Reverting this change as Console tests are failing on non-windows. The failure is: `An error has occurred that was not properly handled. Additional information is shown below. The PowerShell process will exit.`

PR #5760, opened 2017-12-29
Added functionality to retry in Invoke-RestMethod and Invoke-WebRequest.

### LINKED ISSUE #5582: WebCmdlets should support retry
suggested design
`Invoke-WebRequest ... -RetryCount <int> -RetryIntervalSec <int>`

adityapatwardhan:
Fixes #5582

## PR Summary
The change added two parameters, `RetryCount` and `RetryIntervalSec` to enable retry functionality.
When retrying a verbose message is sent out to inform the user.

## PR Checklist

- [X] User facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - [X] Issue filed - Issue link: https://github.com/PowerShell/PowerShell-Docs/issues/1991
    - [X] [Add `[feature]` if the change is significant or affectes feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
@@ -237,7 +237,7 @@ namespace Microsoft.PowerShell.Commands
         public virtual int RetryCount { get; set; } = 0;
 
         /// <summary>
-        /// gets or sets the TimeOut property
+        /// gets or sets the RetryIntervalSec property
         /// </summary>
         [Parameter]
         [ValidateRange(1, Int32.MaxValue)]
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
@@ -21,6 +21,7 @@ using System.Xml;
 using System.Collections.Generic;
 using System.Text.RegularExpressions;
 using System.Linq;
+using System.Threading.Tasks;
 
 namespace Microsoft.PowerShell.Commands
 {
@@ -227,6 +228,20 @@ namespace Microsoft.PowerShell.Commands
             set { _maximumRedirection = value; }
         }
         private int _maximumRedirection = -1;
+
+        /// <summary>
+        /// gets or sets the RetryCount property
+        /// </summary>
+        [Parameter]
+        [ValidateRange(0, Int32.MaxValue)]
+        public virtual int RetryCount { get; set; } = 0;
+
+        /// <summary>
+        /// gets or sets the TimeOut property
+        /// </summary>
+        [Parameter]
+        [ValidateRange(1, Int32.MaxValue)]
+        public virtual int RetryIntervalSec { get; set; } = 10;
 
         #endregion
 
@@ -1389,121 +1404,133 @@ namespace Microsoft.PowerShell.Commands
                             WriteVerbose(linkVerboseMsg);
                         }
 
-                        using (HttpRequestMessage request = GetRequest(uri))
+                        int retriesLeft = RetryCount;
+                        HttpResponseMessage response = null;
+
+                        bool isSuccessful = true;
+
+                        do
                         {
-                            FillRequestStream(request);
-                            try
+                            using (HttpRequestMessage request = GetRequest(uri, stripAuthorization: false))
                             {
-                                long requestContentLength = 0;
-                                if (request.Content != null)
-                                    requestContentLength = request.Content.Headers.ContentLength.Value;
-
-                                string reqVerboseMsg = String.Format(CultureInfo.CurrentCulture,
-                                    WebCmdletStrings.WebMethodInvocationVerboseMsg,
-                                    request.Method,
-                                    request.RequestUri,
-                                    requestContentLength);
-                                WriteVerbose(reqVerboseMsg);
-
-                                HttpResponseMessage response = GetResponse(client, request, keepAuthorization);
-
-                                string contentType = ContentHelper.GetContentType(response);
-                                string respVerboseMsg = string.Format(CultureInfo.CurrentCulture,
-                                    WebCmdletStrings.WebResponseVerboseMsg,
-                                    response.Content.Headers.ContentLength,
-                                    contentType);
-                                WriteVerbose(respVerboseMsg);
-
-                                bool _isSuccess = response.IsSuccessStatusCode;
-
-                                // Check if the Resume range was not satisfiable because the file already completed downloading.
-                                // This happens when the local file is the same size as the remote file.
-                                if (Resume.IsPresent &&
-                                    response.StatusCode == HttpStatusCode.RequestedRangeNotSatisfiable &&
-                                    response.Content.Headers.ContentRange.HasLength &&
-                                    response.Content.Headers.ContentRange.Length == _resumeFileSize)
+                                FillRequestStream(request);
+                                try
                                 {
-                                    _isSuccess = true;
-                                    WriteVerbose(String.Format(CultureInfo.CurrentCulture, WebCmdletStrings.OutFileWritingSkipped, OutFile));
-                                    // Disable writing to the OutFile.
-                                    OutFile = null;
-                                }
+                                    long requestContentLength = 0;
+                                    if (request.Content != null)
+                                        requestContentLength = request.Content.Headers.ContentLength.Value;
 
-                                if (!_isSuccess)
-                                {
-                                    string message = String.Format(CultureInfo.CurrentCulture, WebCmdletStrings.ResponseStatusCodeFailure,
-                                        (int)response.StatusCode, response.ReasonPhrase);
-                                    HttpResponseException httpEx = new HttpResponseException(message, response);
-                                    ErrorRecord er = new ErrorRecord(httpEx, "WebCmdletWebResponseException", ErrorCategory.InvalidOperation, request);
-                                    string detailMsg = string.Empty;
-                                    StreamReader reader = null;
-                                    try
+                                    string reqVerboseMsg = String.Format(CultureInfo.CurrentCulture,
+                                        WebCmdletStrings.WebMethodInvocationVerboseMsg,
+                                        request.Method,
+                                        request.RequestUri,
+                                        requestContentLength);
+                                    WriteVerbose(reqVerboseMsg);
+
+                                    response = GetResponse(client, request, stripAuthorization);
+
+                                    bool causeFailureForTesting = (System.Management.Automation.Internal.InternalTestHooks.FailWebRequestCount > 0) && retriesLeft > 0;
+
+                                    isSuccessful = response.IsSuccessStatusCode && !causeFailureForTesting;
+
+                                    if (!isSuccessful && retriesLeft > 0)
                                     {
-                                        reader = new StreamReader(StreamHelper.GetResponseStream(response));
-                                        // remove HTML tags making it easier to read
-                                        detailMsg = System.Text.RegularExpressions.Regex.Replace(reader.ReadToEnd(), "<[^>]*>", string.Empty);
+                                        retriesLeft--;
+
+                                        string retryMessage = string.Format(CultureInfo.CurrentCulture,
+                                            WebCmdletStrings.RetryVerboseMsg,
+                                            RetryIntervalSec,
+                                            response.StatusCode
+                                            );
+
+                                        WriteVerbose(retryMessage);
+                                        Task.Delay(RetryIntervalSec * 1000).GetAwaiter().GetResult();
+                                        continue;
                                     }
-                                    catch (Exception)
+
+                                    string contentType = ContentHelper.GetContentType(response);
+                                    string respVerboseMsg = string.Format(CultureInfo.CurrentCulture,
+                                        WebCmdletStrings.WebResponseVerboseMsg,
+                                        response.Content.Headers.ContentLength,
+                                        contentType);
+                                    WriteVerbose(respVerboseMsg);
+
+                                    if (!response.IsSuccessStatusCode)
                                     {
-                                        // catch all
-                                    }
-                                    finally
-                                    {
-                                        if (reader != null)
+                                        string message = String.Format(CultureInfo.CurrentCulture, WebCmdletStrings.ResponseStatusCodeFailure,
+                                            (int)response.StatusCode, response.ReasonPhrase);
+                                        HttpResponseException httpEx = new HttpResponseException(message, response);
+                                        ErrorRecord er = new ErrorRecord(httpEx, "WebCmdletWebResponseException", ErrorCategory.InvalidOperation, request);
+                                        string detailMsg = "";
+                                        StreamReader reader = null;
+                                        try
                                         {
-                                            reader.Dispose();
+                                            reader = new StreamReader(StreamHelper.GetResponseStream(response));
+                                            // remove HTML tags making it easier to read
+                                            detailMsg = System.Text.RegularExpressions.Regex.Replace(reader.ReadToEnd(), "<[^>]*>", "");
+                                        }
+                                        catch (Exception)
+                                        {
+                                            // catch all
+                                        }
+                                        finally
+                                        {
+                                            if (reader != null)
+                                            {
+                                                reader.Dispose();
+                                            }
+                                        }
+                                        if (!String.IsNullOrEmpty(detailMsg))
+                                        {
+                                            er.ErrorDetails = new ErrorDetails(detailMsg);
+                                        }
+                                        ThrowTerminatingError(er);
+                                    }
+
+                                    if (_parseRelLink || _followRelLink)
+                                    {
+                                        ParseLinkHeader(response, uri);
+                                    }
+                                    ProcessResponse(response);
+                                    UpdateSession(response);
+
+                                    // If we hit our maximum redirection count, generate an error.
+                                    // Errors with redirection counts of greater than 0 are handled automatically by .NET, but are
+                                    // impossible to detect programmatically when we hit this limit. By handling this ourselves
+                                    // (and still writing out the result), users can debug actual HTTP redirect problems.
+                                    if (WebSession.MaximumRedirection == 0) // Indicate "HttpClientHandler.AllowAutoRedirect == false"
+                                    {
+                                        if (response.StatusCode == HttpStatusCode.Found ||
+                                            response.StatusCode == HttpStatusCode.Moved ||
+                                            response.StatusCode == HttpStatusCode.MovedPermanently)
+                                        {
+                                            ErrorRecord er = new ErrorRecord(new InvalidOperationException(), "MaximumRedirectExceeded", ErrorCategory.InvalidOperation, request);
+                                            er.ErrorDetails = new ErrorDetails(WebCmdletStrings.MaximumRedirectionCountExceeded);
+                                            WriteError(er);
                                         }
                                     }
-                                    if (!String.IsNullOrEmpty(detailMsg))
+                                }
+                                catch (HttpRequestException ex)
+                                {
+                                    ErrorRecord er = new ErrorRecord(ex, "WebCmdletWebResponseException", ErrorCategory.InvalidOperation, request);
+                                    if (ex.InnerException != null)
                                     {
-                                        er.ErrorDetails = new ErrorDetails(detailMsg);
+                                        er.ErrorDetails = new ErrorDetails(ex.InnerException.Message);
                                     }
                                     ThrowTerminatingError(er);
                                 }
 
-                                if (_parseRelLink || _followRelLink)
+                                if (_followRelLink)
                                 {
-                                    ParseLinkHeader(response, uri);
-                                }
-                                ProcessResponse(response);
-                                UpdateSession(response);
-
-                                // If we hit our maximum redirection count, generate an error.
-                                // Errors with redirection counts of greater than 0 are handled automatically by .NET, but are
-                                // impossible to detect programmatically when we hit this limit. By handling this ourselves
-                                // (and still writing out the result), users can debug actual HTTP redirect problems.
-                                if (WebSession.MaximumRedirection == 0) // Indicate "HttpClientHandler.AllowAutoRedirect == false"
-                                {
-                                    if (response.StatusCode == HttpStatusCode.Found ||
-                                        response.StatusCode == HttpStatusCode.Moved ||
-                                        response.StatusCode == HttpStatusCode.MovedPermanently)
+                                    if (!_relationLink.ContainsKey("next"))
                                     {
-                                        ErrorRecord er = new ErrorRecord(new InvalidOperationException(), "MaximumRedirectExceeded", ErrorCategory.InvalidOperation, request);
-                                        er.ErrorDetails = new ErrorDetails(WebCmdletStrings.MaximumRedirectionCountExceeded);
-                                        WriteError(er);
+                                        return;
                                     }
+                                    uri = new Uri(_relationLink["next"]);
+                                    followedRelLink++;
                                 }
                             }
-                            catch (HttpRequestException ex)
-                            {
-                                ErrorRecord er = new ErrorRecord(ex, "WebCmdletWebResponseException", ErrorCategory.InvalidOperation, request);
-                                if (ex.InnerException != null)
-                                {
-                                    er.ErrorDetails = new ErrorDetails(ex.InnerException.Message);
-                                }
-                                ThrowTerminatingError(er);
-                            }
-
-                            if (_followRelLink)
-                            {
-                                if (!_relationLink.ContainsKey("next"))
-                                {
-                                    return;
-                                }
-                                uri = new Uri(_relationLink["next"]);
-                                followedRelLink++;
-                            }
-                        }
+                        } while (!isSuccessful);
                     }
                     while (_followRelLink && (followedRelLink < _maximumFollowRelLink));
                 }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
@@ -230,14 +230,14 @@ namespace Microsoft.PowerShell.Commands
         private int _maximumRedirection = -1;
 
         /// <summary>
-        /// Gets or Sets the RetryCount property, which determines the number of retries of a failed web request.
+        /// Gets or sets the RetryCount property, which determines the number of retries of a failed web request.
         /// </summary>
         [Parameter]
         [ValidateRange(0, Int32.MaxValue)]
         public virtual int RetryCount { get; set; } = 0;
 
         /// <summary>
-        /// Gets or Sets the RetryIntervalSec property, which determines the number seconds between retries.
+        /// Gets or sets the RetryIntervalSec property, which determines the number seconds between retries.
         /// </summary>
         [Parameter]
         [ValidateRange(1, Int32.MaxValue)]
@@ -646,7 +646,7 @@ namespace Microsoft.PowerShell.Commands
                 }
             }
 
-            if(RetryCount > 0)
+            if (RetryCount > 0)
             {
                 WebSession.RetryCount = RetryCount;
 
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
@@ -230,18 +230,18 @@ namespace Microsoft.PowerShell.Commands
         private int _maximumRedirection = -1;
 
         /// <summary>
-        /// Gets or sets the RetryCount property, which determines the number of retries of a failed web request.
+        /// Gets or sets the MaximumRetryCount property, which determines the number of retries of a failed web request.
         /// </summary>
         [Parameter]
         [ValidateRange(0, Int32.MaxValue)]
-        public virtual int RetryCount { get; set; } = 0;
+        public virtual int MaximumRetryCount { get; set; } = 0;
 
         /// <summary>
         /// Gets or sets the RetryIntervalSec property, which determines the number seconds between retries.
         /// </summary>
         [Parameter]
         [ValidateRange(1, Int32.MaxValue)]
-        public virtual int RetryIntervalSec { get; set; } = 60;
+        public virtual int RetryIntervalSec { get; set; } = 5;
 
         #endregion
 
@@ -646,9 +646,9 @@ namespace Microsoft.PowerShell.Commands
                 }
             }
 
-            if (RetryCount > 0)
+            if (MaximumRetryCount > 0)
             {
-                WebSession.RetryCount = RetryCount;
+                WebSession.MaximumRetryCount = MaximumRetryCount;
 
                 // only set retry interval if retry count is set.
                 WebSession.RetryIntervalInSeconds = RetryIntervalSec;
@@ -1294,7 +1294,7 @@ namespace Microsoft.PowerShell.Commands
         {
             int intCode = (int)code;
 
-            if (((intCode == 304) || (intCode >= 400 && intCode <= 599)) && WebSession.RetryCount > 0)
+            if (((intCode == 304) || (intCode >= 400 && intCode <= 599)) && WebSession.MaximumRetryCount > 0)
             {
                 return true;
             }
@@ -1308,7 +1308,7 @@ namespace Microsoft.PowerShell.Commands
             if (request == null) { throw new ArgumentNullException("request"); }
 
             // Add 1 to account for the first request.
-            int totalRequests = WebSession.RetryCount + 1;
+            int totalRequests = WebSession.MaximumRetryCount + 1;
             HttpRequestMessage req = request;
             HttpResponseMessage response = null;
 
@@ -1387,7 +1387,8 @@ namespace Microsoft.PowerShell.Commands
 
                 _resumeSuccess = response.StatusCode == HttpStatusCode.PartialContent;
 
-                if (ShouldRetry(response.StatusCode))
+                // When MaximumRetryCount is not specified, the totalRequests == 1.
+                if (totalRequests > 1 && ShouldRetry(response.StatusCode))
                 {
                     string retryMessage = string.Format(
                         CultureInfo.CurrentCulture,
@@ -1396,9 +1397,15 @@ namespace Microsoft.PowerShell.Commands
                         response.StatusCode);
 
                     WriteVerbose(retryMessage);
-                    Task.Delay(WebSession.RetryIntervalInSeconds * 1000).GetAwaiter().GetResult();
+
+                    _cancelToken = new CancellationTokenSource();
+                    Task.Delay(WebSession.RetryIntervalInSeconds * 1000, _cancelToken.Token).GetAwaiter().GetResult();
+                    _cancelToken.Cancel();
+                    _cancelToken = null;
+
                     req.Dispose();
-                    req = GetRequest(Uri);
+                    req = GetRequest(currentUri);
+                    FillRequestStream(req);
                 }
 
                 totalRequests--;
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
@@ -230,18 +230,18 @@ namespace Microsoft.PowerShell.Commands
         private int _maximumRedirection = -1;
 
         /// <summary>
-        /// gets or sets the RetryCount property
+        /// Gets or Sets the RetryCount property, which determines the number of retries of a failed web request.
         /// </summary>
         [Parameter]
         [ValidateRange(0, Int32.MaxValue)]
         public virtual int RetryCount { get; set; } = 0;
 
         /// <summary>
-        /// gets or sets the RetryIntervalSec property
+        /// Gets or Sets the RetryIntervalSec property, which determines the number seconds between retries.
         /// </summary>
         [Parameter]
         [ValidateRange(1, Int32.MaxValue)]
-        public virtual int RetryIntervalSec { get; set; } = 10;
+        public virtual int RetryIntervalSec { get; set; } = 60;
 
         #endregion
 
@@ -645,6 +645,14 @@ namespace Microsoft.PowerShell.Commands
                     WebSession.Headers[key] = Headers[key].ToString();
                 }
             }
+
+            if(RetryCount > 0)
+            {
+                WebSession.RetryCount = RetryCount;
+
+                // only set retry interval if retry count is set.
+                WebSession.RetryIntervalInSeconds = RetryIntervalSec;
+            }
         }
 
         #endregion Virtual Methods
@@ -677,7 +685,6 @@ namespace Microsoft.PowerShell.Commands
         #endregion Helper Properties
 
         #region Helper Methods
-
         private Uri PrepareUri(Uri uri)
         {
             uri = CheckProtocol(uri);
@@ -1300,7 +1307,7 @@ namespace Microsoft.PowerShell.Commands
             if (client == null) { throw new ArgumentNullException("client"); }
             if (request == null) { throw new ArgumentNullException("request"); }
 
-            int retriesLeft = RetryCount;
+            int retriesLeft = WebSession.RetryCount;
             HttpRequestMessage req = request;
             HttpResponseMessage response = null;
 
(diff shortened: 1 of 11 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -229,6 +230,20 @@ public virtual int MaximumRedirection
         }
         private int _maximumRedirection = -1;
 
+        /// <summary>
+        /// gets or sets the RetryCount property
```
markekraus: Please add a more meaningful summary. Capitalize the first work and end with a period.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -229,6 +230,20 @@ public virtual int MaximumRedirection
...
+        /// </summary>
+        [Parameter]
+        [ValidateRange(0, Int32.MaxValue)]
+        public virtual int RetryCount { get; set; } = 0;
+
+        /// <summary>
+        /// gets or sets the RetryIntervalSec property
```
markekraus: Please add a more meaningful summary. Capitalize the first work and end with a period.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1265,106 +1280,133 @@ protected override void ProcessRecord()
                             WriteVerbose(linkVerboseMsg);
                         }
 
-                        using (HttpRequestMessage request = GetRequest(uri, stripAuthorization:false))
+                        int retriesLeft = RetryCount;
+                        HttpResponseMessage response = null;
+
+                        bool isSuccessful = true;
+
+                        do
```
markekraus: Can you investigate collapsing this to a single do/while instead of nested?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1265,106 +1280,133 @@ protected override void ProcessRecord()
...
+
+                                        string retryMessage = string.Format(CultureInfo.CurrentCulture,
+                                            WebCmdletStrings.RetryVerboseMsg,
+                                            RetryIntervalSec,
+                                            response.StatusCode
+                                            );
+
+                                        WriteVerbose(retryMessage);
+                                        Task.Delay(RetryIntervalSec * 1000).GetAwaiter().GetResult();
+                                        continue;
```
markekraus: We should probably dispose the current HttpResponseMessage before retrying.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1265,106 +1280,133 @@ protected override void ProcessRecord()
...
+
+                                    if (!isSuccessful && retriesLeft > 0)
+                                    {
+                                        retriesLeft--;
+
+                                        string retryMessage = string.Format(CultureInfo.CurrentCulture,
+                                            WebCmdletStrings.RetryVerboseMsg,
+                                            RetryIntervalSec,
+                                            response.StatusCode
+                                            );
```
markekraus: This should move to the end of the previous line.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1265,106 +1280,133 @@ protected override void ProcessRecord()
                             WriteVerbose(linkVerboseMsg);
                         }
 
-                        using (HttpRequestMessage request = GetRequest(uri, stripAuthorization:false))
+                        int retriesLeft = RetryCount;
```
markekraus: I believe it would be better to have `RetryCount` and `RetryIntervalSec` set on `WebSession` in `PrepareSession()` and then have this area of code retrieve the values from `WebSession`. It would mean extending `WebRequestSession` to include new Properties, but I believe that is the correct direction for these settings.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1265,106 +1280,133 @@ protected override void ProcessRecord()
...
+                                        requestContentLength);
+                                    WriteVerbose(reqVerboseMsg);
+                                    
+                                    response = GetResponse(client, request, stripAuthorization);
+                                    
+                                    bool causeFailureForTesting = (System.Management.Automation.Internal.InternalTestHooks.FailWebRequestCount > 0) && retriesLeft > 0;
+
+                                    isSuccessful = response.IsSuccessStatusCode && !causeFailureForTesting;                                
+
+                                    if (!isSuccessful && retriesLeft > 0)
```
markekraus: This current logic leads to an interestingly weird combination. If you set `-MaximumRedirection 0 -RetryCount 5 -RetryIntervalSec 1` and the response code is `301`, this will still retry the same link 5 times because `HttpResponseMessage.IsSuccessStatusCode` is only true for response codes `200`-`299` inclusive... not sure how to rectify that.

markekraus: also... since the logic for reducing the current redirection count is handled in `GetResponse()`, we may need to reset `WebSession.MaximumRedirection` to the `MaximumRedirection` property value before continuing.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1265,106 +1280,133 @@ protected override void ProcessRecord()
...
+                                        retriesLeft--;
+
+                                        string retryMessage = string.Format(CultureInfo.CurrentCulture,
+                                            WebCmdletStrings.RetryVerboseMsg,
+                                            RetryIntervalSec,
+                                            response.StatusCode
+                                            );
+
+                                        WriteVerbose(retryMessage);
+                                        Task.Delay(RetryIntervalSec * 1000).GetAwaiter().GetResult();
```
markekraus: See my comment about setting and retrieving `RetryIntervalSec` and `RetryCount` from `WebSession`.

markekraus (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
SteveL-MSFT: Best that I can find, the HTTP spec doesn't say enough about rate limiting to make it something we can depend on reliably.  However, I think we can say that retry only occurs on a 429 or 403 and document it as such unless there's data showing that web services are using other error codes.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -825,6 +825,29 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
...
+
+            $uri = Get-WebListenerUrl -Test 'Get'
+            $command = "Invoke-WebRequest -Uri '$uri' -Verbose -RetryCount 2 -RetryIntervalSec 1 4>$outputFilePath"
+            $result = ExecuteWebCommand -command $command
+
+            ## The webrequest call succeeds after 2 retries.
+            $result.output.StatusCode | Should Be "200"
+
+            # Expect 2 verbose messages as we are retrying twice.
+            (Select-String -Path $outputFilePath -Pattern 'Retrying after interval').Count | Should Be 2
```
SteveL-MSFT: This won't work once we start localizing the error messages

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
markekraus: @SteveL-MSFT I can see other usecases for this beyond just rate limiting. I could see using this to test 404 errors until a new page comes online after a publish/push and it $x attempts fail alert that the push/publish failed. 

Perhaps it we limit it to 400-499 inclusive it would be best. In my experience with APIs I have seen more than 403 and 429 used for busy endpoints (including 404... Reddit... ugh) or rate limiting, but it's always a 400 level error

SteveL-MSFT: @markekraus that's a fair point.  I think 4xx would be fine.

markekraus: After further consideration, this should work for the following status codes: 304. 400-599 inclusive

304 could be used for monitoring the change of a resource and is not a redirect status even though it in the 300 range. Retries on 5XX errors makes perfect sense to overcome temporary service glitches and cloudflare shenanigans.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: I don't really like this test.  I'd rather have a real failure emulated by `WebListener`. Add a new endpoint which take a session ID (new GUID), a status code, and a fail count. Until the initially provided fail count is reached when using that session ID it produces a response with the provided status code After fail count is reached it returns a normal 200 response. The response body should include the request count for the given session ID. This way we can test how it reacts with various responses and differing fail counts provided by actual endpoint errors instead of relying on fake logic to enter the code path and without relying on verbose messages.

iSazonov:
@adityapatwardhan Could you please continue?

iSazonov:
@adityapatwardhan Please resolve conflicts.

adityapatwardhan:
I plan to resume work on this next week.

ffeldhaus:
Did you consider implementing Exponential Backof or different retry strategies? [AWS SDKs for instance use an Exponential Backof](https://docs.aws.amazon.com/general/latest/gr/api-retries.html) and it is good practice for many other web services.

iSazonov:
@ffeldhaus Please open new Issue for your great suggestion - we shouldn't lost this.

ffeldhaus:
I thought about this feature some more and I don't think it should be included in the current form without further discussion to prevent retries to cause issues or unwanted behavior and I don't think a Pull Request is the right place to discuss the design of this feature. As issue #5582 was already created for this feature, we should first discuss the design there and then continue with the Pull Request.

I will add my concerns and design ideas to issue #5582.

markekraus:
@iSazonov Retry and Resume are separate features. Resume continues downloading a partially downloaded file. Retry tries the URI again if a non-susccess status code was received. This PR is for the Retry feature, #6447 is for the Resume feature.

iSazonov:
I wouldn't want to lose this.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
adityapatwardhan: Fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
adityapatwardhan: Fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
adityapatwardhan: Moved logic to GetResponse()

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
adityapatwardhan: fixed

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
adityapatwardhan: Fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
adityapatwardhan: Fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
adityapatwardhan: Fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -229,6 +230,20 @@ public virtual int MaximumRedirection
...
+        [Parameter]
+        [ValidateRange(0, Int32.MaxValue)]
+        public virtual int RetryCount { get; set; } = 0;
+
+        /// <summary>
+        /// gets or sets the RetryIntervalSec property
+        /// </summary>
+        [Parameter]
+        [ValidateRange(1, Int32.MaxValue)]
+        public virtual int RetryIntervalSec { get; set; } = 10;
```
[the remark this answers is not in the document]
adityapatwardhan: Fixed.

adityapatwardhan:
@iSazonov @markekraus @SteveL-MSFT  I have addressed all the feedback. Please have another look.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1268,80 +1290,117 @@ static bool IsRedirectToGet(HttpStatusCode code)
             );
         }
 
+        bool ShouldRetry(HttpStatusCode code)
+        {
+            int intCode = (int) code;
```
iSazonov: Please remove extra space after `(int)`. (See CodeFactor issues.)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -630,6 +645,14 @@ internal virtual void PrepareSession()
                     WebSession.Headers[key] = Headers[key].ToString();
                 }
             }
+
+            if(RetryCount > 0)
```
iSazonov: Please add space after `if`. (See CodeFactor issues.))

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1268,80 +1290,117 @@ static bool IsRedirectToGet(HttpStatusCode code)
...
+                        WriteVerbose(reqVerboseMsg);
+
+                        return GetResponse(client, requestWithoutRange, keepAuthorization);
+                    }
+                }
+
+                _resumeSuccess = response.StatusCode == HttpStatusCode.PartialContent;
 
-                using (HttpRequestMessage requestWithoutRange = GetRequest(currentUri))
+                if(ShouldRetry(response.StatusCode))
```
iSazonov: Please add space after `if`. (See CodeFactor issues.)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -228,6 +229,20 @@ public virtual int MaximumRedirection
         }
         private int _maximumRedirection = -1;
 
+        /// <summary>
+        /// Gets or Sets the RetryCount property, which determines the number of retries of a failed web request.
```
iSazonov: CodeFactor wants  "Gets or sets".

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -228,6 +229,20 @@ public virtual int MaximumRedirection
...
+        /// </summary>
+        [Parameter]
+        [ValidateRange(0, Int32.MaxValue)]
+        public virtual int RetryCount { get; set; } = 0;
+
+        /// <summary>
+        /// Gets or Sets the RetryIntervalSec property, which determines the number seconds between retries.
```
iSazonov: CodeFactor wants  "Gets or sets".

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/WebRequestSession.cs
```diff
@@ -64,6 +64,16 @@ public class WebRequestSession
         /// </summary>
         public int MaximumRedirection { get; set; }
 
+        /// <summary>
+        /// Gets or Sets the count of retries for request failures
+        /// </summary>
+        public int RetryCount { get; set; }
+
+        /// <summary>
+        /// Gets or Sets the interval in seconds between retries.
```
iSazonov: CodeFactor wants  "Gets or sets".

test/tools/WebListener/Startup.cs
```diff
@@ -79,6 +79,10 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
                     template: "Delete",
                     defaults: new {controller = "Get", action = "Index"},
                     constraints: new RouteValueDictionary(new { httpMethod = new HttpMethodRouteConstraint("DELETE") }));
+                routes.MapRoute(
+                    name: "retry",
+                    template: "Retry/{sessionId?}/{failureCode?}/{failureCount?}",
+                    defaults: new {controller = "Retry", action = "Retry"});
```
iSazonov: CodeFactor wants spaces with braces.
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1012.md

test/tools/WebListener/Controllers/RetryController.cs
```diff
@@ -0,0 +1,64 @@
...
+                    return Json(resp);
+                }
+            }
+            else
+            {
+                //initialize the failureResponsesSent as 1.
+                var newRetryInfoItem = Tuple.Create(failureCode, failureCount, 1);
+                retryInfo.Add(sessionId, newRetryInfoItem);
+                Response.StatusCode = failureCode;
+                Hashtable error = new Hashtable {{"error", $"Error: HTTP - {failureCode} occurred."}};
```
iSazonov: CodeFactor wants spaces with braces.

test/tools/WebListener/Controllers/RetryController.cs
```diff
@@ -0,0 +1,64 @@
...
+                else
+                {
+                    retryInfo.Remove(sessionId);
+                    var resp = new Hashtable {{"failureResponsesSent", retry.Item3}};
+                    return Json(resp);
+                }
+            }
+            else
+            {
+                //initialize the failureResponsesSent as 1.
```
iSazonov: Please add space after `//`

test/tools/WebListener/Controllers/RetryController.cs
```diff
@@ -0,0 +1,64 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Net;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Http.Extensions;
```
iSazonov: Wrong sorting - please swap lines 11 and 12.

test/tools/WebListener/Controllers/RetryController.cs
```diff
@@ -0,0 +1,64 @@
...
+namespace mvc.Controllers
+{
+    public class RetryController : Controller
+    {
+        // Dictionary for sessionId as key and failureCode, failureCount and failureResponsesSent as the value.
+        private static Dictionary<string, Tuple<int, int, int>> retryInfo;
+
+        public JsonResult Retry(string sessionId, int failureCode, int failureCount)
+        {
+            if(retryInfo == null)
```
iSazonov: Please add space after `if`.

test/tools/WebListener/Controllers/RetryController.cs
```diff
@@ -0,0 +1,64 @@
...
+            {
+                retryInfo = new Dictionary<string, Tuple<int, int, int>>();
+            }
+
+            if(retryInfo.TryGetValue(sessionId, out Tuple<int, int, int> retry))
```
iSazonov: Please add space after `if`.

test/tools/WebListener/Controllers/RetryController.cs
```diff
@@ -0,0 +1,64 @@
...
+            {
+                // if failureResponsesSent is less than failureCount
+                if(retry.Item3 < retry.Item2)
```
iSazonov: Please add space after `if`.

test/tools/WebListener/Controllers/RetryController.cs
```diff
@@ -0,0 +1,64 @@
...
+                var newRetryInfoItem = Tuple.Create(failureCode, failureCount, 1);
+                retryInfo.Add(sessionId, newRetryInfoItem);
+                Response.StatusCode = failureCode;
+                Hashtable error = new Hashtable {{"error", $"Error: HTTP - {failureCode} occurred."}};
+                return Json(error);
+            }
+        }
+
+        public IActionResult Error()
+
```
iSazonov: Please remove extra line.

test/tools/WebListener/Controllers/RetryController.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Linq;
+using System.Net;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.Http.Extensions;
+using Microsoft.Extensions.Primitives;
+using mvc.Models;
+
+namespace mvc.Controllers
```
iSazonov: CodeFactor - Namespace names begin with an upper-case letter
https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1300.md

iSazonov (commented):
My comments come from CodeFactor - I believe we should fix the issues if we have changed the code.

test/tools/WebListener/Controllers/RetryController.cs
adityapatwardhan: This will cause unnecessary changes across the WebListener project. I prefer not fixing this.

adityapatwardhan:
@iSazonov code review comments addressed.

adityapatwardhan:
@markekraus Added documentation and fixed more CodeFactor issues.

adityapatwardhan:
@SteveL-MSFT  Can you have another look?

iSazonov:
Restarted Travis CI Linux - seems temporary failed.

iSazonov (approved):

adityapatwardhan:
@SteveL-MSFT  @markekraus  Please update your reviews.

markekraus:
@adityapatwardhan I may not have time until this weekend to give this the attention it deserves. I do want to review it for edge cases before it's pulled.

ffeldhaus:
Is it really ok to do the retry for Status Code 304 and 400 to 599? There are several Status Codes were a retry just does not make sense and will just increase the load/error counters on the server. 400, 409, 410, 429, 431, 505 would be some were I don't think a retry should be done. 413 would require special retry handling.

markekraus:
@ffeldhaus I have plenty of use cases for retries just about every status code. For example, the status code is server side bug, a fix is being released to prod, and I want the script to retry until x amount of time passes or the fix is live.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/WebRequestSession.cs
```diff
@@ -64,6 +64,16 @@ public class WebRequestSession
         /// </summary>
         public int MaximumRedirection { get; set; }
 
+        /// <summary>
+        /// Gets or sets the count of retries for request failures.
+        /// </summary>
+        public int RetryCount { get; set; }
```
SteveL-MSFT: To be consistent, I wonder if this should be `MaximumRetry`

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/WebRequestSession.cs
adityapatwardhan: I like this suggestion, I will make this change.

markekraus: `MaximumRetry` sounds weird. Maybe `MaximumRetryCount`?

Looking at existing commands, `New-PSSessionOption` uses `-MaxConnectionRetryCount`

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1268,80 +1290,121 @@ static bool IsRedirectToGet(HttpStatusCode code)
...
+                    string retryMessage = string.Format(
+                        CultureInfo.CurrentCulture,
+                        WebCmdletStrings.RetryVerboseMsg,
+                        RetryIntervalSec,
+                        response.StatusCode);
+
+                    WriteVerbose(retryMessage);
+                    Task.Delay(WebSession.RetryIntervalInSeconds * 1000).GetAwaiter().GetResult();
+                    req.Dispose();
+                    req = GetRequest(Uri);
```
markekraus: `Uri` should be `currentUri`

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1268,80 +1290,121 @@ static bool IsRedirectToGet(HttpStatusCode code)
...
+                        WriteVerbose(reqVerboseMsg);
 
-                using (HttpRequestMessage requestWithoutRange = GetRequest(currentUri))
+                        return GetResponse(client, requestWithoutRange, keepAuthorization);
+                    }
+                }
+
+                _resumeSuccess = response.StatusCode == HttpStatusCode.PartialContent;
+
+                if (ShouldRetry(response.StatusCode))
```
markekraus: The logic here results in the verbose message printed an extra time when the final result is still an error:

```powershell
Invoke-RestMethod -RetryCount 1 -RetryIntervalSec 1 https://httpbin.org/status/404 -Verbose -SkipCertificateCheck
```

```none
VERBOSE: GET https://httpbin.org/status/404 with 0-byte payload
VERBOSE: Retrying after interval of 1 seconds. Status code for previous attempt: NotFound
VERBOSE: Retrying after interval of 1 seconds. Status code for previous attempt: NotFound
VERBOSE: received 0-byte response of content type text/html
irm : Response status code does not indicate success: 404 (NOT FOUND).
At line:1 char:1
+ irm -RetryCount 1 -RetryIntervalSec 1 https://httpbin.org/status/404  ...
+ CategoryInfo          : InvalidOperation: (Method: GET, Re...rShell/6.1.0
}:HttpRequestMessage) [Invoke-RestMethod], HttpResponseException
+ FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeRestMethodCommand
```

Maybe `if (totalRequests<0 && ShouldRetry(response.StatusCode))`?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1268,80 +1290,121 @@ static bool IsRedirectToGet(HttpStatusCode code)
...
-
-                    return GetResponse(client, requestWithoutRange, keepAuthorization);
+                    string retryMessage = string.Format(
+                        CultureInfo.CurrentCulture,
+                        WebCmdletStrings.RetryVerboseMsg,
+                        RetryIntervalSec,
+                        response.StatusCode);
+
+                    WriteVerbose(retryMessage);
+                    Task.Delay(WebSession.RetryIntervalInSeconds * 1000).GetAwaiter().GetResult();
```
markekraus: is there some better way to do this that will be responsive to ctrl+alt+del? Currently, if you use the default 60 seconds, got an error endpoint without verbose it looks like powershell is hung because ctrl+alt+delete doesn't end the command.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
markekraus: `FillRequestStream(req)` needs to called to ensure the body is present.

There should also be a test to ensure body data survives the retries.

```powershell
Invoke-WebRequest -Uri $Uri -Body 'a' -Method 'POST' -RetryCount 1  -RetryIntervalSec 1
```

with fiddler the body on the first request is correct. the body on the second request is empty.

markekraus (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
adityapatwardhan: Good catch. Thanks.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
adityapatwardhan: Added a cancellation token for the Delay. Ctrl+C will cancel the delay and the webrequest.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -228,6 +229,20 @@ public virtual int MaximumRedirection
...
+        [Parameter]
+        [ValidateRange(0, Int32.MaxValue)]
+        public virtual int RetryCount { get; set; } = 0;
+
+        /// <summary>
+        /// Gets or sets the RetryIntervalSec property, which determines the number seconds between retries.
+        /// </summary>
+        [Parameter]
+        [ValidateRange(1, Int32.MaxValue)]
+        public virtual int RetryIntervalSec { get; set; } = 60;
```
markekraus: 60 seconds as the default just feels too long. So many times testing this feature and forgetting the `-RetryIntervalSec` I felt like the console froze or the thread locked. between 10 and 15 seconds I can't decide which is better. 10 feels more natural, but 15 is probably better. anything over 15 really starts to feel like the command has frozen.

adityapatwardhan: According to Azure docs [here](https://docs.microsoft.com/en-us/aspnet/aspnet/overview/developing-apps-with-windows-azure/building-real-world-cloud-apps-with-windows-azure/transient-fault-handling#use-smart-retryback-off-logic-to-mitigate-the-effect-of-transient-failures) the sample shows 5 seconds. @markekraus  @SteveL-MSFT what do you think?

markekraus: I would agree with 5 seconds for transient errors, but we are covering more than transient errors (for example, 404). But... maybe it's ok. 5 seconds is still only 12 requests a minute which isn't that spammy, IMO.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
adityapatwardhan: Fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/WebRequestSession.cs
adityapatwardhan: Fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
adityapatwardhan: Changing it to 5 seconds, if we get feedback about it to increase the defaults, we can reconsider.

SteveL-MSFT (approved):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1268,80 +1290,128 @@ static bool IsRedirectToGet(HttpStatusCode code)
...
+                    WriteVerbose(retryMessage);
+
+                    _cancelToken = new CancellationTokenSource();
+                    Task.Delay(WebSession.RetryIntervalInSeconds * 1000, _cancelToken.Token).GetAwaiter().GetResult();
+                    _cancelToken.Cancel();
+                    _cancelToken = null;
+
+                    req.Dispose();
+                    req = GetRequest(currentUri);
+                    FillRequestStream(req);
```
markekraus: @adityapatwardhan Can you either add a test for retry POST data or open an issue to have one added?

adityapatwardhan: Added a test for `POST`

adityapatwardhan:
@markekraus Added a test for POST, please have another look. Thanks!

markekraus:
@adityapatwardhan That test doesn't provide a body payload nor does it test the response. The WebListener endpoint would need to echo back the body payload and then the pester test would need to check the response from WebListener to ensure the echoed payload matches the sent payload.

adityapatwardhan:
@markekraus Updated test and controller with echo for sessionId.

markekraus:
@adityapatwardhan Two more things and I think this will be done. I just realized there are only tests for `Invoke-Webrequest`. Please add the same tests for `Invoke-RestMethod`. Also, instead of wrapping the tests in `#region Retry tests`, please use a `Context` block.

adityapatwardhan:
@markekraus Since the code path is same, I believe adding all the tests will be duplication. I will add one test for Invoke-RestMethod to cover the scenario.

markekraus:
@adityapatwardhan The convention for the web cmdlets is to test all features on both, even common features. There is always the possibility that the common features clash with the implementation features.

adityapatwardhan:
@markekraus Please review again. I have added tests for `Invoke-RestMethod`. I moved them to a new describe block so they can be grouped together. The describes for other Invoke-RestMethod and Invoke-WebRequest tests were separate. Thanks for the feedback, very much appreciated.

markekraus:
@adityapatwardhan there is a `Describe "Invoke-WebRequest tests" -Tags "Feature"` block and a `Describe "Invoke-RestMethod tests" -Tags "Feature"` and in those are various context blocks 

for example, 

https://github.com/PowerShell/PowerShell/blob/e8d2a6b27827dd366f638468f96487ddf2de1db4/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1#L827

and 

https://github.com/PowerShell/PowerShell/blob/e8d2a6b27827dd366f638468f96487ddf2de1db4/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1#L2119

adityapatwardhan:
@markekraus I decided to go with a separate `Describe` block so that we can use `TestCase` for `It` and re-use most of the test code. If we have tests for `Invoke-RestMethod` and `Invoke-WebRequest` in separate `Describe` blocks then there will be test code duplication. I prefer a separate `Describe` but since there is precedence I am not completely against adding them to existing `Describe` blocks.

Let me know your thoughts?

markekraus:
@adityapatwardhan For all new features they have been separate context blocks in the top level describe blocks for each cmdlet. I'd prefer to stick to this pattern for now and look at changing things up in one of the open issues I have on refactoring the code and tests.

adityapatwardhan:
@markekraus Ok, agreed. I will send an update in a few minutes.

adityapatwardhan:
@markekraus Made the changes to the tests as discussed above.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1723,6 +1723,49 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
             $response.Headers.'Content-Range'[0] | Should -BeExactly "bytes */$referenceFileSize"
         }
     }
+
+    Context "Invoke-WebRequest retry tests" {
+
+        It "<Command> can retry - <Name>" -TestCases @(
```
markekraus: `<Command>` -> `Invoke-WebRequest` since it's no longer in the testcases hash.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1723,6 +1723,49 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
...
+        It "Invoke-WebRequest should fail when failureCount is greater than MaximumRetryCount" {
+
+            $uri = Get-WebListenerUrl -Test 'Retry' -Query @{ sessionid = (New-Guid).Guid; failureCode = 400; failureCount = 4 }
+            $command = "Invoke-WebRequest -Uri '$uri' -MaximumRetryCount 1 -RetryIntervalSec 1"
+            $result = ExecuteWebCommand -command $command
+            $jsonError = $result.error | ConvertFrom-Json
+            $jsonError.error | Should -BeExactly 'Error: HTTP - 400 occurred.'
+        }
+
+        It "<Command> can retry with POST" {
```
markekraus: `<Command>` -> `Invoke-WebRequest` since it's no longer in the testcases hash.

markekraus (changes requested):

adityapatwardhan:
@markekraus  Good catch! Fixed.

test/tools/WebListener/README.md
```diff
@@ -682,3 +682,20 @@ X-WebListener-Has-Range: false
...
+
+```powershell
+$response = Invoke-WebRequest -Uri 'http://127.0.0.1:8083/Retry?failureCode=599&failureCount=2&sessionid=100&' -MaximumRetryCount 2 -RetryIntervalSec 1
+```
+
+Response Body:
+
+```json
+{
+  "failureResponsesSent":2
```
markekraus: Can you update this to include the `sessionId` that is now also returned.

markekraus (changes requested):
One final thing. Sorry :(

adityapatwardhan:
@markekraus  Fixed documentation.

markekraus (approved):
@adityapatwardhan Thanks for you patience and diligence.  LGTM

iSazonov:
Reopen to restart CIs.

adityapatwardhan:
@iSazonov  Thanks for restarting CIs. All tests passed, ready to merge?

iSazonov:
@adityapatwardhan Thanks for great contribution!

Issue #5643, opened 2017-12-06
PowerShell should support creating an List similar to how it supports arrays


TravisEz13:
Powershell supports creating arrays with `$array = 'a', 1, '3'` .  Then you can add an element to the array with `$array += 4`, but this creates a new array which is not performant.

Powershell should have a syntax which allows creating lists.
Assuming the operator is `@[...]`,  you could create a list with `$list = @['a', 1, '3']` and then you could add an element to the existing list with `$list += 4` without PowerShell having to create a new list.
Note: this new operator might function more like `@(...)`

This design assumes that changing `,` would be a breaking change.  I'm open to discussing changing `,` as well.

I filed this based on an offline discussion about this comment on a PR:  https://github.com/PowerShell/PowerShell/pull/5625#discussion_r155106230

rkeithhill:
I like the idea but to truly impact performance you'd need to be operating on **large** lists.  For a convenient way to create large lists, I would expect something like this to work `@[Get-ChildItem C:\Windows -r -file *.dll -ea 0]`.  While the list literal form is nice, I don't see folks creating lists large enough to gain much of a perf benefit over using an array.  Well, unless the list literal is created inside a busy (large n) loop.

lzybkr:
Two points:

* `@()` says - make sure the thing inside is an array. It's not necessary if you use `,` because the comma operator always creates an array.
* I've often wondered if the comma operator could create a list instead of an array. I have a feeling most scripts would never notice a difference because of how freely things are converted to an object array.

TravisEz13:
@lzybkr I updated my description based on @lzybkr 's comments

daxian-dbw:
I like the idea of having a list literal in powershell. I think it could have a syntax like `@[1, 2, 3]` to directly create a list with elements 1, 2 and 3 without first create an array literal from `1,2,3` and then make it a list using `@[]`.

cchiu1979:
@lzybkr is ti something like this?
1,2,3
(1,2,3).length
( , (1,2,3) ).length
( @(1,2,3) ).length

mklement0:
> I've often wondered if the comma operator could create a list instead of an array.

If that's not considered too much of a breaking change, it would certainly be the best solution.

Otherwise:

@rkeithhill: 

[...]

PR #5761, opened 2017-12-29
[WIP] Support ListLiteralExpression '[]' in PowerShell

daxian-dbw:
## PR Summary
An attempt to address #5643 

This design adds the `ListLiteralExpression '[]'` to PowerShell, which is similar to the `ArrayLiteralExpression`. They are different in mainly two ways:
- `ListLiteralExpression` always returns a `List<object>`
- When there is only one element expression, `ListLiteralExpression` unwraps the value of the expression while `ArrayLiteralExpression` directly uses the value as the first element of the returned array.

Please review with #5762 as a comparison, where list is supported as a `ListExpression '@[]'`, similar to `ArrayExpression '@()'`.

### Example
```
# an empty list
> $l = []
> $l.GetType()

True     True     List`1                                   System.Object
```
```
# one element list, the element is a type
> $l = [[string]]
> $l.Count
1
> $l[0]

```
```
# multi-element list
> $l = [$pid,2,$home,4]
> $l.Count
4
```
```
# unwrap the only expression element in a list literal expression, like '@()'
> $l = [(dir)]
> $l.GetType()

True     True     List`1                                   System.Object

> $l.Count
30

# as a comparison, see the behavior of ArrayLiteralExpression
> $a = ,(dir)
> $a.Count
1
```

## PR Checklist

### DIFF
--- a/src/System.Management.Automation/engine/interpreter/InstructionFactory.cs
+++ b/src/System.Management.Automation/engine/interpreter/InstructionFactory.cs
@@ -61,6 +61,7 @@ namespace System.Management.Automation.Interpreter
         protected internal abstract Instruction DefaultValue();
         protected internal abstract Instruction NewArray();
         protected internal abstract Instruction NewArrayInit(int elementCount);
+        protected internal abstract Instruction NewListInit(int elementCount);
     }
 
     internal sealed class InstructionFactory<T> : InstructionFactory
@@ -75,6 +76,7 @@ namespace System.Management.Automation.Interpreter
         private Instruction _newArray;
         private Instruction _typeAs;
         private Instruction[] _newArrayInit;
+        private Instruction[] _newListInit;
         // This number is somewhat arbitrary - trying to avoid some gc without keeping
         // objects (instructions) around that aren't used that often.
         private const int MaxArrayInitElementCountCache = 32;
@@ -123,6 +125,20 @@ namespace System.Management.Automation.Interpreter
                 return _newArrayInit[elementCount] ?? (_newArrayInit[elementCount] = new NewArrayInitInstruction<T>(elementCount));
             }
             return new NewArrayInitInstruction<T>(elementCount);
+        }
+
+        protected internal override Instruction NewListInit(int elementCount)
+        {
+            if (elementCount < MaxArrayInitElementCountCache)
+            {
+                if (_newListInit == null)
+                {
+                    _newListInit = new Instruction[MaxArrayInitElementCountCache];
+                }
+
+                return _newListInit[elementCount] ?? (_newListInit[elementCount] = new NewListInitInstruction<T>(elementCount));
+            }
+            return new NewListInitInstruction<T>(elementCount);
         }
     }
 }
--- a/src/System.Management.Automation/engine/parser/AstVisitor.cs
+++ b/src/System.Management.Automation/engine/parser/AstVisitor.cs
@@ -175,8 +175,15 @@ namespace System.Management.Automation.Language
         object VisitDynamicKeywordStatement(DynamicKeywordStatementAst dynamicKeywordAst);
     }
 
+    /// <summary/>
+    public interface ICustomAstVisitor3 : ICustomAstVisitor2
+    {
+        /// <summary/>
+        object VisitListLiteral(ListLiteralAst listLiteralAst);
+    }
+
 #if DEBUG
-    class CheckAllParentsSet : AstVisitor2
+    class CheckAllParentsSet : AstVisitor3
     {
         internal CheckAllParentsSet(Ast root)
         {
@@ -239,6 +246,7 @@ namespace System.Management.Automation.Language
         public override AstVisitAction VisitInvokeMemberExpression(InvokeMemberExpressionAst ast) { return CheckParent(ast); }
         public override AstVisitAction VisitArrayExpression(ArrayExpressionAst ast) { return CheckParent(ast); }
         public override AstVisitAction VisitArrayLiteral(ArrayLiteralAst ast) { return CheckParent(ast); }
+        public override AstVisitAction VisitListLiteral(ListLiteralAst ast) { return CheckParent(ast); }
         public override AstVisitAction VisitHashtable(HashtableAst ast) { return CheckParent(ast); }
         public override AstVisitAction VisitScriptBlockExpression(ScriptBlockExpressionAst ast) { return CheckParent(ast); }
         public override AstVisitAction VisitParenExpression(ParenExpressionAst ast) { return CheckParent(ast); }
@@ -258,7 +266,7 @@ namespace System.Management.Automation.Language
     /// <summary>
     /// Check if <see cref="TypeConstraintAst"/> contains <see cref="TypeBuilder "/> type
     /// </summary>
-    class CheckTypeBuilder : AstVisitor2
+    class CheckTypeBuilder : AstVisitor3
     {
         public override AstVisitAction VisitTypeConstraint(TypeConstraintAst ast)
         {
@@ -275,7 +283,7 @@ namespace System.Management.Automation.Language
     /// <summary>
     /// Searches an AST, using the evaluation function provided by either of the constructors
     /// </summary>
-    internal class AstSearcher : AstVisitor2
+    internal class AstSearcher : AstVisitor3
     {
         #region External interface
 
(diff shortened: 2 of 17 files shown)


src/System.Management.Automation/engine/interpreter/InstructionFactory.cs
```diff
@@ -124,5 +126,19 @@ protected internal override Instruction NewArrayInit(int elementCount)
             }
             return new NewArrayInitInstruction<T>(elementCount);
         }
+
+        protected internal override Instruction NewListInit(int elementCount)
```
lzybkr: I don't think we should add this to the interpreter - this way it stays as close as possible to original and/or the one used for `dynamic` when now JIT is available.

It is possible to add new instructions without touching the interpreter it that's how you want to do it - see `PowerShellLoopExpression` for an example.

src/System.Management.Automation/engine/parser/AstVisitor.cs
```diff
@@ -175,8 +175,15 @@ public interface ICustomAstVisitor2 : ICustomAstVisitor
         object VisitDynamicKeywordStatement(DynamicKeywordStatementAst dynamicKeywordAst);
     }
 
+    /// <summary/>
+    public interface ICustomAstVisitor3 : ICustomAstVisitor2
```
lzybkr: I think numbered interfaces made sense when the release cycle spanned years, but I'm worried we'll end up with `ICustomAstVistor9` before too long.

lzybkr: One (maybe not awesome) idea is to avoid the problem for this feature is to add a `bool` flag to `ArrayLiteralAst` - because they should be structurally the same.

The downside to this approach is that some analysis tools that aren't aware of list literals will analyze the ast as an array, e.g. they may incorrectly give warnings about performance.

markekraus:
why `[]` and not `@[]`? seeing just `[]` is confusing compared to types. also `[()]` is weird... I'd expect something similar to `@()`:

```powershell
$DateList = @[
    Get-Date
    Start-Sleep 4
    Get-Date
]
```

src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -5537,6 +5543,46 @@ public object VisitArrayLiteral(ArrayLiteralAst arrayLiteralAst)
...
+            if (listLiteralAst.Elements.Count == 0)
+            {
+                return ExpressionCache.NewEmptyObjectList;
+            }
+
+            if (listLiteralAst.Elements.Count == 1)
+            {
+                Expression values = Compile(listLiteralAst.Elements[0]);
+
+                if (values.Type == typeof(List<object>)) { return values; }
```
lzybkr: I don't understand this.  Does this mean the following would be true?

```
$x = [1]
$y = [[1]]
$x.Count -eq $y.Count -and $x[0] -eq $y[0]
```

lzybkr: Reading more of the implementation, I'm confused now. Is this supposed to emulate `ArrayExpressionAst` or `ArrayLiteralAst` - it seems like the former, but it's named like the later.

daxian-dbw:
@markekraus My first design attempt is to support `@[]`, behaving like `@()`, see PR #5762. However, unlike character `')'` that always starts a new token, the character `']'` doesn't force to start a new token and therefore can be included in `GenericToken`, meaning command names, arguments, function names and etc. It makes impossible to determine where `'@[]'` ends and I cannot find an elegant solution without breaking changes.

We need more discussions on the design, and that is why I submitted 2 PRs for the list support: #5762 to add `@[]`, and this one to add `[]`.

src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -5987,6 +5987,68 @@ private ExpressionAst ErrorRecoveryParameterInExpression(ParameterToken paramTok
...
+
+            var lBracket = NextLBracket();
+            if (lBracket == null) { return null; }
+
+            var listValues = new List<ExpressionAst>();
+            Token commaToken = null;
+
+            do {
+                if (commaToken != null) { SkipToken(); }
+                V3SkipNewlines();
```
lzybkr: It's probably time to get rid of `V3SkipNewlines` - I was overly cautious introducing that method and it hasn't been used.

src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -5987,6 +5987,68 @@ private ExpressionAst ErrorRecoveryParameterInExpression(ParameterToken paramTok
...
+            }
+
+            IScriptExtent extent;
+            if (rBracket.Kind == TokenKind.RBracket)
+            {
+                extent = ExtentOf(lBracket, rBracket.Extent);
+            }
+            else if (listValues.Count > 0)
+            {
+                extent = ExtentOf(lBracket, listValues[listValues.Count-1].Extent);
```
lzybkr: This excludes a trailing comma. The simple approach to tracking the end is to just have a variable `endExtent` and keep updating it with each token that you consume or ast you add to `listValues`.

lzybkr (changes requested):
Since this is `wip`, I just did a quick skim.

src/System.Management.Automation/engine/interpreter/InstructionFactory.cs
daxian-dbw: The `NewListInitInstruction` is a minor optimization which can be removed.
But I need to implement `LightCompiler.CompileListInitExpression` because it's currently not implemented. Are you suggesting to not use `ListInitExpression` in our compiler generated code?

src/System.Management.Automation/engine/parser/AstVisitor.cs
daxian-dbw: I thought about reusing the `ArrayLiteralAst`, but didn't go that direction for two reasons:
1. An explicit `ListLiteralAst` is more discoverable.
2. `ArrayLiteralAst` can appear on the left side of an assignment. I didn't plan to make `[]` also assignable. (but I don't object the idea to make it able to do the same).

I would like to see more discussion on this before updating the code.

src/System.Management.Automation/engine/parser/Compiler.cs
daxian-dbw: This is a change I made that is different from `ArrayLiteral` -- when there is only one expression element in the `ListLiteral`, unwrap the value of the expression element like `@()` does.
This is because I think it may be better for `[(dir)]` to return a list containing all files than it to return a list containing an object array.
It's just one of my thoughts, not necessarily a right one 😄

src/System.Management.Automation/engine/parser/Parser.cs
daxian-dbw: Will replace `V3SkipNewlines` with `SkipNewlines`

src/System.Management.Automation/engine/parser/Compiler.cs
lzybkr: You shouldn't call it a literal if you can't guarantee the number of objects in the result.

src/System.Management.Automation/engine/parser/Parser.cs
daxian-dbw: Update to use a `endExtent` variable.

src/System.Management.Automation/engine/interpreter/InstructionFactory.cs
lzybkr: I see the method is now implemented: http://source.dot.net/#System.Linq.Expressions/System/Linq/Expressions/Interpreter/LightCompiler.cs,a988c37e1da7a799

I wonder if there is a supported way of invoking the interpreter now? We'd have to make sure they have any of our fixes first, but it'd be nice to not ship two copies of basically the same thing.

src/System.Management.Automation/engine/parser/Compiler.cs
daxian-dbw: Good point. I will remove that part of logic.

src/System.Management.Automation/engine/interpreter/InstructionFactory.cs
daxian-dbw: Great, I can use their implementation then.

daxian-dbw:
Per the discussions in #5643, it's commonly agreed that we want `ListExpressionAst` instead of a `ListLiteralAst`. Therefore, I will close this PR.

PR #5762, opened 2017-12-29
[WIP] Support ListExpression '@[]' in PowerShell

daxian-dbw:
## PR Summary
Another attempt to address #5643 

This design is to add the `ListExpression '@[]'` to PowerShell, which is similar to `ArrayExpression '@()'` but always returns a `List<object>`.

Please review with #5761 as a comparison, where list is supported as a `ListLiteralExpression`, similar to `ArrayLiteralExpression`.

### A blocking issue
`@[]` has a `SubExpression` like `'@()'` and `'$()'`. However, unlike the closing parenthesis character `')'`, the closing bracket character `']'` doesn't always force to start a new token, and it can be included in a generic token, meaning that `']'` can appear in a command name, argument, or function name. This makes it impossible for `@[dir]` to determine the ending of the list expression because `dir]` will be treated as a single generic token.

This PR adds the property `InListSubExpression` to `Tokenizer`, and makes `']'` a force-to-start-new-token character when `_tokenizer.InListSubExpression` is set. This approach solves the most common UX problem but is by no way perfect, for example, comparing to `@(funcHas[]inName)` or `@(dir has[]inpath)`, `@[funcHas[]inName]` and `@[dir has[]inpath]` won't work because the first `']'` will force the command name to end.

**I couldn't find a less hacky solution without breaking changes (make `'['` and `']'` always not included by `GenericToken`). So I started to think what if we support list in PowerShell like the `ArrayLiteralExpression` instead of `@()`. In that case, a list can only contain `Expression` elements and hence command name, arguments, and function names won't be a problem for the ending bracket. #5761 was also submitted to support list as a `ListLiteralExpression`.**

### Example
```
# an empty list
> $l = @[]
> $l.GetType()

True     True     List`1                                   System.Object
```
```
# one element list, the element is a type
> $l = @[[string]]
> $l.Count
1
> $l[0]

```
```
# multi-element list
> $l = @[$pid,2,$home,4]
> $l.Count
4
> $l.GetType()

True     True     List`1                                   System.Object
```
```
# arbitrary statements in @[], like @()
> $l = @[dir]
> $l.GetType()

True     True     List`1                                   System.Object

> $l.Count
30
```

## PR Checklist

    - [ ] [Add `[feature]` if the change is significant or affectes feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

### DIFF
--- a/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
+++ b/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
@@ -6702,7 +6702,7 @@ namespace System.Management.Automation
     /// use is for intellisense where you don't want to run arbitrary code, but you do want to know the values
     /// of various expressions so you can get the members.
     /// </summary>
-    internal class SafeExprEvaluator : ICustomAstVisitor2
+    internal class SafeExprEvaluator : ICustomAstVisitor3
     {
         internal static bool TrySafeEval(ExpressionAst ast, ExecutionContext executionContext, out object value)
         {
@@ -6850,6 +6850,11 @@ namespace System.Management.Automation
         public object VisitIndexExpression(IndexExpressionAst indexExpressionAst)
         {
             return (bool)indexExpressionAst.Target.Accept(this) && (bool)indexExpressionAst.Index.Accept(this);
+        }
+
+        public object VisitListExpression(ListExpressionAst listExpressionAst)
+        {
+            return listExpressionAst.SubExpression.Accept(this);
         }
 
         public object VisitArrayExpression(ArrayExpressionAst arrayExpressionAst)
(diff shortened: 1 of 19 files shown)


lzybkr:
You could consider 2 character tokens.

For example, F# uses this syntax for an array literal:

```f#
[| 1; 2 |]
```

There are other possibilities that probably aren't breaking changes, e.g. `[< 1, 2 >]`.

The key here is to use a second character that can't be in a command name.

daxian-dbw:
@lzybkr that's a good idea! I didn't think in that way. It would be great if we can have `@[]` because it resembles the existing `@()` and `$()`, but I'm fine to have new token pairs for List support.
I will start prototype with `[<>]`, and wait for more discussion at the same time.

daxian-dbw:
The proposal now is to use a type accelerator `[list]` for `List<object>`, so that users can easily create a list using `$list = [list]@(...)`. This way, we can avoid the issue we have here, but still provide an easy way to create a list. Given that, I will close this PR. Please see #5643 for more discussions.

PR #5791, opened 2018-01-04
Make PowerShell able to enable logging of script block execution on Unix platforms

### LINKED ISSUE #5695: Logging of script block execution cannot be enabled on Unix
PowerShell checks whether to log script block activities in [`ShouldLogScriptBlockActivity`](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs#L1478), which query Registry on windows to get the configuration via [`GetGroupPolicySetting`](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/Utils.cs#L527). However, `GetGroupPolicySetting` does nothing on Unix, and thus script block activity logging cannot be enabled on Unix.

A potential fix is to make PowerShell checks against an entry in `pwsh.runtimeconfig.json`.

daxian-dbw:
## PR Summary
Fix #5695

Enable ScriptBlockLogging on Linux/macOS by moving the GroupPolicy configurations to the configuration file `PowerShellProperties.json`. Including the following tasks:
1. Enable the logging via `MshLog.cs` on Unix platforms. (**the 1st commit**)
2. Refactor `PropertyAccessor.cs`. (**the 2nd commit**)
    - Remove the unneeded base type
    - Rename 'ConfigPropertyAccessor' to `PowerShellConfig`
    - Move `PowerShellConfig` to the namespace `System.Management.Automation.Configuration`
3. Rename the enum `PropertyScope` to `ConfigScope`. (**the 3rd commit**)
4. Rename file `PropertyAccessor.cs` to `PSConfiguration.cs`. (**the 4th commit**)
5. Refactor GroupPolicy setting related code to use the configuration file. (**the 5th commit**)
    - Followings are the policy settings that are moved to the configuration file.
       - ScriptExecution settings (used in `SecuritySupport.cs`)
       - ScriptBlockLogging settings (used in `CompiledScriptBlock.cs`)
       - ModuleLogging settings (used in `ModuleCmdletBase.cs`)
       - ProtectedEventLogging settings (used in `CompiledScriptBlock.cs`)
       - Transcription settings (used in `MshHostUserInterface.cs`)
       - UpdatableHelp setting (used in `UpdatableHelpSystem.cs`)
       - ConsoleSessionConfiguration settings (used in `CommandLineParameterParser.cs`)
    - In `PSConfiguration.cs`, updated `ReadValueFromFile<T>` to make it flexible to allow a custom `JsonObject` parsing implementation delegate passed in.
    - In `PSConfiguration.cs`, added `GetPowerShellPolicies(ConfigScope scope)` and make `DefaultSourcePath` setting part of the new policy settings because it is part of the group policy on windows powershell.
    - In `Utils.cs`, replace the method `GetGroupPolicySetting` with `T GetPolicySetting<T>(ConfigScope[] preferenceOrder) where T : PolicyBase`. The old method was for reading from Registry, while the new one is for reading from configuration file.
    - Changes in other files are the corresponding changes due to `GetGroupPolicySetting` being replaced.
6. Add xUnit tests to test reading policy settings from the configuration files in different scenarios. (**the 6th commit**)
    - The xUnit tests are running in parallel by default. Tests for reading policy settings manipulate the configuration file and thus are in conflict with other xUnit tests. So the existing xUnit tests are moved to the namespace `PSTests.Parallel` and the new tests are moved to `PSTests.Sequential` and the corresponding build scripts are updated to run sequential xUnit tests first and then run other xUnit tests in parallel.
7. A minor change to fix an issue in xUnit test and another minor change to `SysLogProvider` to make ScriptBlock logging not ignored on Unix platforms. (**the 7th commit**)
8. Re-enable GPO support on Windows. On Windows, we first query GPO from registry, if the required policy is not defined, then we query policies from the configuration file. (**the 9th commit**)
9. Rename `PowerShellProperties.json` to `powershell.config.json`. (**the 10th commit**)

### An example of the configuration file `powershell.config.json`

```
{
  "Microsoft.PowerShell:ExecutionPolicy": "RemoteSigned",
  "PowerShellPolicies": {
    "ScriptExecution": {
      "ExecutionPolicy": "RemoteSigned",
      "EnableScripts": true
    },
    "ScriptBlockLogging": {
      "EnableScriptBlockInvocationLogging": true,
      "EnableScriptBlockLogging": true
    },
    "ModuleLogging": {
      "EnableModuleLogging": false,
      "ModuleNames": [
        "PSReadline",
        "PowerShellGet"
      ]
    },
    "ProtectedEventLogging": {
      "EnableProtectedEventLogging": false,
      "EncryptionCertificate": [
        "Joe"
      ]
    },
    "Transcription": {
      "EnableTranscripting": true,
      "EnableInvocationHeader": true,
      "OutputDirectory": "F:\\tmp\\new"
    },
    "UpdatableHelp": {
      "DefaultSourcePath": "f:\\temp"
    },
    "ConsoleSessionConfiguration": {
      "EnableConsoleSessionConfiguration": false,
      "ConsoleSessionConfigurationName": "name"
    }
  },
  "LogLevel": "verbose"
}
```

### Limitation that needs to be documented

On Unix platform, the default logging level filter is `information`, while the logging of script block execution writes out `verbose` level logging. So the configuration entry `"LogLevel": "verbose"` is needed along with the settings in `"ScriptBlockLogging"`. This needs to be documented.

### Documentation needed

We need documentation to clarify how logging works in PowerShell. Not sure if we already have some docs about it. Issue https://github.com/PowerShell/PowerShell-Docs/issues/2017 is opened.

## PR Checklist


daxian-dbw:
There are a lot of changes in this PR. When reviewing this PR, I recommend you review each commit because I intentionally group logically related changes into separate commits.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -459,29 +460,17 @@ internal void Parse(string[] args)
...
-        private static string s_groupPolicyBase = @"Software\Policies\Microsoft\Windows\PowerShell";
-        private static string s_consoleSessionConfigurationKey = "ConsoleSessionConfiguration";
-        private static string s_enableConsoleSessionConfiguration = "EnableConsoleSessionConfiguration";
-        private static string s_consoleSessionConfigurationName = "ConsoleSessionConfigurationName";
         private static string GetConfigurationNameFromGroupPolicy()
         {
             // Current user policy takes precedence.
-            var groupPolicySettings = Utils.GetGroupPolicySetting(s_groupPolicyBase, s_consoleSessionConfigurationKey, Utils.RegCurrentUserThenLocalMachine);
-            if (groupPolicySettings != null)
+            var consoleSessionSetting = Utils.GetPolicySetting<ConsoleSessionConfiguration>(Utils.CurrentUserThenSystemWideConfig);
```
SteveL-MSFT: With this change, the Windows PowerShell GPO settings don't apply to PSCore6, correct?  This seems correct to me, but is a `breaking change`.  We should label this PR as such.  cc @joeyaiello

daxian-dbw: Yes, GPO settings in Registry won't apply to PowerShell Core. I have added the `breaking change` label. Thanks for catching that.

SteveL-MSFT (changes requested):
Still reviewing

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
TravisEz13: This does not seem correct to me. @Indhukrishna

dantraMSFT (changes requested):

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -4749,28 +4750,19 @@ internal static ModuleLoggingGroupPolicyStatus GetModuleLoggingInformation(out I
...
             ModuleLoggingGroupPolicyStatus status = ModuleLoggingGroupPolicyStatus.Undefined;
-            Dictionary<string, object> groupPolicySettings = Utils.GetGroupPolicySetting("ModuleLogging", Utils.RegLocalMachineThenCurrentUser);
 
-            if (groupPolicySettings != null)
+            var moduleLogging = Utils.GetPolicySetting<ModuleLogging>(Utils.SystemWideThenCurrentUserConfig);
+            if (moduleLogging != null)
             {
-                object enableModuleLoggingValue = null;
-                if (groupPolicySettings.TryGetValue("EnableModuleLogging", out enableModuleLoggingValue))
+                if (moduleLogging.EnableModuleLogging == false)
```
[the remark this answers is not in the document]
daxian-dbw: I didn't change the original code, will update it to set `status` in both cases.

src/System.Management.Automation/engine/hostifaces/MshHostUserInterface.cs
```diff
@@ -921,7 +921,7 @@ internal static TranscriptionOption GetSystemTranscriptOption(TranscriptionOptio
                 {
                     if (systemTranscript == null)
```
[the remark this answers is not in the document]
daxian-dbw: I didn't change the original code. I guess you are right, this method could return a `TranscriptionOption` object or null depending on whether `StopAllTranscribing` set `systemTranscript` to `null` before `return systemTranscript;` in this method. But I don't know the transcribing code very well to know if that could be a problem. Can you open an issue tracking this comment?

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -677,5 +493,128 @@ private void RemoveValueFromFile<T>(string fileName, string key)
...
+
+    internal abstract class PolicyBase { }
+
+    /// <summary>
+    /// Setting about ScriptExecution
+    /// </summary>
+    internal sealed class ScriptExecution : PolicyBase
+    {
+        public string ExecutionPolicy { get; set; }
+        public bool? EnableScripts { get; set; }
```
SteveL-MSFT: Not sure I understand the value of having this nullable.  EnableScripts == null is same as EnableScripts == true, correct?  Same question for the nullable properties below where they should have a default value and being null is the same as the default value.

SteveL-MSFT (changes requested):

src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
```diff
@@ -1234,13 +1235,14 @@ internal bool Compile(bool optimized)
...
-            if (force || ShouldLogScriptBlockActivity("EnableScriptBlockLogging"))
+            ScriptBlockLogging logSetting = GetScriptBlockLoggingSetting();
+            if (force || logSetting?.EnableScriptBlockLogging == true)
             {
                 if (!scriptBlock.HasLogged || InternalTestHooks.ForceScriptBlockLogging)
                 {
                     // If script block logging is explicitly disabled, or it's from a trusted
                     // file or internal, skip logging.
-                    if (ScriptBlockLoggingExplicitlyDisabled() ||
+                    if (logSetting?.EnableScriptBlockLogging == false ||
```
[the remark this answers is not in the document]
daxian-dbw: There is no behavior change, the check `logSetting?.EnableScriptBlockLogging == false` is to make sure that the `EnableScriptBlockLogging` is explicitly set to `false`. If you look at the original implementation of `ScriptBlockLoggingExplicitlyDisabled()`, it returns `true` only if the registry key `EnableScriptBlockLogging` exists and its value is explicitly 0.

I agree to have docs to clarify how logging works in PowerShell. Not sure if we already have some docs about it.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: `EnableScripts == null` means this setting is not defined in the config file, so it's up to powershell to define the default behavior. `EnableScripts == true` means it's defined in the config file and the value is `true`. Likewise, `EnableScripts == false` means it's defined in the config file and the value is `false`.

It's the same in registry -- a key is not defined, which means the application needs to use the pre-defined behavior; a key is defined, and the value is 1 or 0.

src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
```diff
@@ -1234,13 +1235,14 @@ internal bool Compile(bool optimized)
...
                         scriptBlock.ScriptBlockData.IsProductCode)
```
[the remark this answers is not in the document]
daxian-dbw: Theoretically, this can be re-written as
```csharp
if ((force && logSetting?.EnableScriptBlockLogging != false) || logSetting?.EnableScriptBlockLogging == true)
{
    if ((!scriptBlock.HasLogged || InternalTestHooks.ForceScriptBlockLogging) && !scriptBlock.ScriptBlockData.IsProductCode)
    {
        ....
    }
}
```
Or even more aggressively, these two `if` blocks can be merged into one. But this doesn't look very readable, so I think I'm fine with the original code.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: An example is the use of `EnableScriptBlockLogging` in `LogScriptBlockCreation(ScriptBlock scriptBlock, bool force)`
- when it's explicitly set to 'true', we always do script block logging; 
- when it's not defined, we do script block logging when `force == true`
- when it's explicitly set to 'false', we always disable script block logging, even if `force == true`.

src/System.Management.Automation/engine/hostifaces/LocalPipeline.cs
```diff
@@ -269,7 +269,7 @@ internal static int ReadRegistryInt(string policyValueName, int defaultValue)
             RegistryKey key;
             try
             {
-                key = Registry.LocalMachine.OpenSubKey(Utils.GetRegistryConfigurationPrefix());
+                key = Registry.LocalMachine.OpenSubKey("SOFTWARE\\Microsoft\\PowerShell\\1\\ShellIds");
```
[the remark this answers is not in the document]
daxian-dbw: This code is in a `#if !CORECLR ... #endif` block. I leave the registry key path here only for reference, so that when we clean up this `!CORECLR` block, we know what the key is.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
daxian-dbw: I will start work on enabling GPO as well on windows. The preference order of the settings should be config file --> GPO, meaning if a policy setting is defined in the config file, then use it; if it's not defined in the config file, query registry. This way, you can have control whether you want the GPO settings to take effect.

iSazonov: Common rule is that Computer GPO is high priority, then User GPO, then computer wide config. then user config. It is critical rule and security related.

src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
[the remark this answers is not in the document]
iSazonov: @daxian-dbw Could you please add this in PR description for future docs?

src/System.Management.Automation/engine/PSConfiguration.cs
SteveL-MSFT: Ok, the force case makes sense that there is a difference with null.  Thanks.

src/System.Management.Automation/engine/PropertyAccessor.cs
```diff
@@ -56,142 +28,41 @@ internal enum PropertyScope
...
-            psHomeConfigDirectory = Path.GetDirectoryName(assembly.Location);
+            psHomeConfigDirectory = Utils.DefaultPowerShellAppBase;
 
-            //
             // Sets the per-user configuration directory
             // Note: This directory may or may not exist depending upon the
             // execution scenario. Writes will attempt to create the directory
             // if it does not already exist.
-            //
             appDataConfigDirectory = Utils.GetUserConfigurationDirectory();
```
iSazonov: We can install PowerShell Core side-by-side with other PowerShell Core-s. So we should have configs for "default" system wide installation and for "local" for every user installation.
Also it seems we should have GPO option - "Force GPO for Side-By-Side PowerShell installations" - by default we should apply GPO only for system wide PowerShell installation.

src/System.Management.Automation/help/UpdatableHelpSystem.cs
```diff
@@ -1682,7 +1683,7 @@ internal static void SetDisablePromptToUpdateHelp()
         {
             try
             {
-                ConfigPropertyAccessor.Instance.SetDisablePromptToUpdateHelp(true);
+                PowerShellConfig.Instance.SetDisablePromptToUpdateHelp(true);
```
iSazonov: We have plans to make Help System optional. So it seems Config API should be public in future.

src/System.Management.Automation/engine/CommandDiscovery.cs
```diff
@@ -1643,59 +1643,6 @@ private int GetCmdletRemovalIndex(List<CmdletInfo> cacheEntry, string PSSnapin)
 
         internal ExecutionContext Context { get; }
 
-        /// <summary>
-        /// Reads the path for the appropriate shellID from the registry.
```
iSazonov: Maybe leave this? How Windows user or Domain Admins can system-wide switch from Windows PowerShell to PowerShell Core?

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -4,29 +4,25 @@
...
 
 namespace System.Management.Automation.Configuration
 {
+    internal enum ConfigScope
+    {
+        // SystemWide configuration applies to all users.
+        SystemWide = 0,
+
+        // CurrentUser configuration applies to the current user.
+        CurrentUser = 1
```
iSazonov: Previously I have put a comment that we should consider Side-By-Side local config too. It seems we should add new value here.

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -362,46 +331,35 @@ internal PSKeyword GetLogKeywords()
             return result;
         }
 #endif // UNIX
-
-        private T ReadValueFromFile<T>(string fileName, string key)
+ 
+        private T ReadValueFromFile<T>(string fileName, string key, T defaultValue = default(T),
```
iSazonov: Do we read option by option? I'd expect that we read whole config file, parse and apply.

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -535,5 +493,128 @@ private void RemoveValueFromFile<T>(string fileName, string key)
...
+        public UpdatableHelp UpdatableHelp { get; set; }
+        public ConsoleSessionConfiguration ConsoleSessionConfiguration { get; set; }
+    }
+
+    internal abstract class PolicyBase { }
+
+    /// <summary>
+    /// Setting about ScriptExecution
+    /// </summary>
+    internal sealed class ScriptExecution : PolicyBase
```
iSazonov: Maybe `ScriptExecutionPolicy`?

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -362,46 +331,35 @@ internal PSKeyword GetLogKeywords()
...
+                using (var streamReader = new StreamReader(fileName))
+                using (var jsonReader = new JsonTextReader(streamReader))
                 {
-                    // Safely determines whether there is content to read from the file
-                    bool isReadSuccess = jsonReader.Read();
-                    if (isReadSuccess)
+                    var settings = new JsonSerializerSettings() { TypeNameHandling = TypeNameHandling.None, MaxDepth = 10 };
+                    var serializer = JsonSerializer.Create(settings);
+
+                    var configData = serializer.Deserialize<JObject>(jsonReader);
```
PaulHigin: Has a security review been done on this code?  Is it possible for the deserializer to create arbitrary types or run arbitrary code?  If so is it possible to apply a schema to mitigate?  This could be a possible security hole when we (re) implement PowerShell white listing protection.  Can you add a comment about this possible vulnerability and create an issue to track?

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -535,5 +493,128 @@ private void RemoveValueFromFile<T>(string fileName, string key)
...
+    internal sealed class ScriptExecution : PolicyBase
+    {
+        public string ExecutionPolicy { get; set; }
+        public bool? EnableScripts { get; set; }
+    }
+
+    /// <summary>
+    /// Setting about ScriptBlockLogging
+    /// </summary>
+    internal sealed class ScriptBlockLogging : PolicyBase
```
iSazonov: The same - maybe `ScriptBlockLoggingPolicy`?

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -535,5 +493,128 @@ private void RemoveValueFromFile<T>(string fileName, string key)
...
+    {
+        public bool? EnableScriptBlockInvocationLogging { get; set; }
+        public bool? EnableScriptBlockLogging { get; set; }
+    }
+
+    /// <summary>
+    /// Setting about ModuleLogging
+    /// </summary>
+    internal sealed class ModuleLogging : PolicyBase
```
iSazonov: The same - maybe `ModuleLoggingPolicy`?

PaulHigin (dismissed):

src/System.Management.Automation/engine/PSConfiguration.cs
iSazonov: I see check null for instance the classes - maybe static is better if we always process config?

test/csharp/test_PSConfiguration.cs
```diff
@@ -33,6 +33,12 @@ public PowerShellPolicyFixture()
...
             currentUserConfigDirectory = Utils.GetUserConfigurationDirectory();
 
+            if (!Directory.Exists(currentUserConfigDirectory))
+            {
+                // Create the CurrentUser config directory if it doesn't exist
+                Directory.CreateDirectory(currentUserConfigDirectory);
+            }
+
             systemWideConfigFile = Path.Combine(systemWideConfigDirectory, "PowerShellProperties.json");
             currentUserConfigFile = Path.Combine(currentUserConfigDirectory, "PowerShellProperties.json");
```
iSazonov: I believe `Properties` should be replaced with `Config`.
`PowerShellConfig.json` or `PowerShell.Config.json` or `PowerShell.config.json`

src/System.Management.Automation/engine/PropertyAccessor.cs
```diff
@@ -56,142 +28,41 @@ internal enum PropertyScope
...
-    {
         private string psHomeConfigDirectory;
         private string appDataConfigDirectory;
         private const string configFileName = "PowerShellProperties.json";
 
         /// <summary>
-        /// Lock used to enable multiple concurrent readers and singular write locks within a
-        /// single process.
-        /// TODO: This solution only works for IO from a single process. A more robust solution is needed to enable ReaderWriterLockSlim behavior between processes.
+        /// Lock used to enable multiple concurrent readers and singular write locks within a single process.
```
iSazonov: We have `lock (s_cachedPolicies)` in Utils.cs - maybe that's enough?

src/System.Management.Automation/engine/PSConfiguration.cs
iSazonov: > is it possible to apply a schema to mitigate

We have a PR for Test-Json - it can help here.

src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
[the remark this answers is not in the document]
dantraMSFT: If we don't, the new about_Logging.md would be a reasonable place.
As far as the behavior, the implementation currently has three behaviors - 
1: EnableScriptBlockLogging == true: always log scriptblocks, 
2: EnableScriptBlockLogging == false: don't log scripblocks
3: EnableScriptBlockLogging not defined: log suspicious scriptblocks

Reading through this again, the implementation should be simplified: 
The force parameter is explicitly tied to HasSuspiciousContent in the caller so you might as well do the work in LogScriptBlockCreation.
In the current implementation, #2 and #3 are effectively identical since we pass force = true for suspicious script blocks. In short, false and default have the same behavior.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: The `SystemWide` option is using the config file under $PSHome, it is the system-wide config in terms of a specific PSCore deployment, but in side-by-side scenario where multiple PSCore deployments are involved, each PSCore deployment will have its own system-wide config file.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: Yes, we currently read what is being requested, so that we don't need to parse unnecessary portion of the config file.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: This policy controls script execution. In `Local Group Policy Editor`, its name is "Turn on Script Execution".

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: We are using the same name as the keys in GPO, which I think will make it easy to correlate them, and also make it easy to read GPO settings on Windows. You will see that in my new commits.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: Ditto.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: @PaulHigin the use of `JsonSerailizer` here is the same as in `ConverFrom/ConverTo-Json` cmdlets. I will add a comment and create an issue.

src/System.Management.Automation/engine/PropertyAccessor.cs
daxian-dbw: There are other settings except for the policy related ones and they may query the config file simultaneously, so this lock is needed.

test/csharp/test_PSConfiguration.cs
daxian-dbw: I don't like this file name either. I will see if I can rename it.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: I had a closer look at the usage, and think it should be safe. The usage here can only convert values to the limited types: string, bool, string[], bool?, and sub types of `PolicyBase`.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
daxian-dbw: For the policy related settings, we are having a debate on which one should take preference. I will talk to @TravisEz13 later today about it.
For non-policy related settings, we shouldn't use registry settings anymore but only refer to the config file.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: > I see check null for instance the classes - maybe static is better if we always process config?

I'm not sure what this comment is referring to. Can you be more specific?

src/System.Management.Automation/engine/CommandDiscovery.cs
daxian-dbw: I don't understand your question. This method is removed because it's not used anywhere.

src/System.Management.Automation/help/UpdatableHelpSystem.cs
daxian-dbw: The config APIs needs more improvements for sure, one thing is is the singleton pattern really needed -- can the methods just be static methods?

src/System.Management.Automation/engine/PSConfiguration.cs
PaulHigin: @daxian-dbw  If that is the case then we should be Ok.  Thanks.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
daxian-dbw: The conclusion is: on windows, we query group policies from registry first, and if the requested policy is not defined in GP, then we query from the configuration file.
New commit has been pushed to enable this.

test/csharp/test_PSConfiguration.cs
daxian-dbw: Pushed new commit to rename the file. The file name is now `powershell.config.json`.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -497,125 +497,207 @@ internal static bool IsValidPSEditionValue(string editionValue)
...
-        // We use a static to avoid creating "extra garbage."
-        private static Dictionary<string, object> s_emptyDictionary = new Dictionary<string, object>(StringComparer.OrdinalIgnoreCase);
+        private readonly static ConcurrentDictionary<ConfigScope, PowerShellPolicies> s_cachedPoliciesFromConfigFile =
+            new ConcurrentDictionary<ConfigScope, PowerShellPolicies>();
 
-        internal static Dictionary<string, object> GetGroupPolicySetting(string groupPolicyBase, string settingName, RegistryKey[] preferenceOrder)
+        /// <summary>
+        /// Get a specific kind of policy setting from the configuration file.
+        /// </summary>
+        internal static T GetPolicySettingFromConfigFile<T>(ConfigScope[] preferenceOrder) where T : PolicyBase, new()
```
dantraMSFT: Shouldn't this be private?

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -497,125 +497,207 @@ internal static bool IsValidPSEditionValue(string editionValue)
...
-                // No group policy settings, then return null
-                if (settings.Count == 0)
+                return (T) tInstance;
+            }
+        }
+
+        /// <summary>
+        /// Get a specific kind of policy setting from the group policy registry key.
+        /// </summary>
+        internal static T GetPolicySettingFromGPO<T>(ConfigScope[] preferenceOrder) where T : PolicyBase, new()
```
dantraMSFT: This should be private, right?

dantraMSFT (changes requested):

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -29,7 +29,7 @@ internal sealed class PowerShellConfig
 
         private string psHomeConfigDirectory;
         private string appDataConfigDirectory;
-        private const string configFileName = "PowerShellProperties.json";
+        private const string configFileName = "powershell.config.json";
```
dantraMSFT: about_Logging.md will need to be updated. 
See https://github.com/PowerShell/PowerShell-Docs/blob/staging/reference/6/Microsoft.PowerShell.Core/About/about_Logging.md

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: I think I misunderstood your comment :) You mean it may be more readable if we add 'Policy' as a suffix to those type names. I would like to keep those names as close to GP registry key as possible, especially since we are still using GP in PSCore.

src/System.Management.Automation/engine/PSConfiguration.cs
dantraMSFT: Regarding your comment about static; this class should be static. There's nothing in this class that requires an instance and the fields are all read-only in usage.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: Opened https://github.com/PowerShell/PowerShell-Docs/issues/2017 to track the documentation update.

dantraMSFT: Note that about_logging.md needs to be updated regardless since it refers to 'PowerShellProperties.json'.

src/System.Management.Automation/engine/runtime/CompiledScriptBlock.cs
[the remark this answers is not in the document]
daxian-dbw: Yes, a documentation section is added to PR description.

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -197,33 +186,13 @@ internal void SetDisablePromptToUpdateHelp(bool prompt)
...
 
         /// <summary>
-        /// Existing Key = HKCU and HKLM\Software\Policies\Microsoft\Windows\PowerShell\UpdatableHelp
-        /// Proposed value = blank.This should be supported though
-        ///
-        /// Schema:
-        /// {
-        ///     "DefaultSourcePath" : "path to local updatable help location"
-        /// }
+        /// Corresponding settings of the original Group Policies
```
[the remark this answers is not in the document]
daxian-dbw: Since we now re-enabled the GPO support on Windows, I will leave this comment unchanged.

src/System.Management.Automation/engine/Utils.cs
daxian-dbw: Yes, it should. Will update.

src/System.Management.Automation/engine/Utils.cs
daxian-dbw: Yes, it should. Will update.

src/System.Management.Automation/engine/PropertyAccessor.cs
SteveL-MSFT: Based on internal discussion today, the current plan is to have PSCore6 GPO settings separate from Windows PowerShell.  For now, we don't plan to have side-by-side GPO settings per version of PSCore6 until there is customer need for it.  New admx will be published later for the PSCore6 settings (which will be the same as Windows PowerShell, but reside in a different location in registry).

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -538,8 +556,148 @@ internal static bool IsValidPSEditionValue(string editionValue)
...
+        {
+            {nameof(ScriptExecution), @"Software\Policies\Microsoft\PowerShellCore"},
+            {nameof(ScriptBlockLogging), @"Software\Policies\Microsoft\PowerShellCore\ScriptBlockLogging"},
+            {nameof(ModuleLogging), @"Software\Policies\Microsoft\PowerShellCore\ModuleLogging"},
+            {nameof(ProtectedEventLogging), @"Software\Policies\Microsoft\Windows\EventLog\ProtectedEventLogging"},
+            {nameof(Transcription), @"Software\Policies\Microsoft\PowerShellCore\Transcription"},
+            {nameof(UpdatableHelp), @"Software\Policies\Microsoft\PowerShellCore\UpdatableHelp"},
+            {nameof(ConsoleSessionConfiguration), @"Software\Policies\Microsoft\PowerShellCore\ConsoleSessionConfiguration"}
+        };
+        private readonly static ConcurrentDictionary<Tuple<ConfigScope, string>, PolicyBase> s_cachedPoliciesFromRegistry =
```
TravisEz13: Really, we should register for GP change notifications and clear the cache.   Not a regression, so, can you add a comment or file an issue:  https://msdn.microsoft.com/en-us/library/windows/desktop/aa374404(v=vs.85).aspx

src/System.Management.Automation/engine/PSConfiguration.cs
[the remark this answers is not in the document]
SteveL-MSFT: Correct

src/System.Management.Automation/engine/hostifaces/LocalPipeline.cs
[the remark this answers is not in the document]
SteveL-MSFT: ok

src/System.Management.Automation/engine/Utils.cs
daxian-dbw: Issue https://github.com/PowerShell/PowerShell/issues/5804 is opened to track this.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: @dantraMSFT Agreed. I left a similar comment at https://github.com/PowerShell/PowerShell/pull/5791#discussion_r159966649. We need to revisit the configuration APIs when working on the experimental feature work.

src/System.Management.Automation/engine/PropertyAccessor.cs
iSazonov: I think we could find a way to use only one global lock - if we want to have consistency, it doesn't matter how and what settings change.

src/System.Management.Automation/engine/PSConfiguration.cs
iSazonov: Do PowerShell silently eat wrong config options without the schema check? It is bad UX.

src/System.Management.Automation/engine/PSConfiguration.cs
iSazonov: Yes, I meant that it may be more readable if we add 'Policy' as a suffix to those type names. It is not blocking comment.

src/System.Management.Automation/engine/PSConfiguration.cs
iSazonov: I don't understand why we do that. We have some config files and registry - all so small that we'd read and parse them all in one step without performance issue.

src/System.Management.Automation/engine/CommandDiscovery.cs
iSazonov: My thoughts was about side-by-side scenario but it seems not help.
Closed.

src/System.Management.Automation/help/UpdatableHelpSystem.cs
iSazonov: Yes. we can revisit the API later. I think we could create a common API to re-use the same config files and registry for customer modules.

src/System.Management.Automation/engine/PropertyAccessor.cs
iSazonov: GPO is Enterprise feature. GPO should allow Domain Admins get full control on PowerShell. I full trust MSFT experts but I'd prefer discuss GPO in a Issue. Also we forget about AppLocker.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
iSazonov: >For non-policy related settings, we shouldn't use registry settings anymore but only refer to the config file.

I don't agree - GPO is great Enterprise feature and Domain Admins should have full and flexible control on PowerShell. We could discuss this in a Issue (as I mention below too).

src/System.Management.Automation/engine/PSConfiguration.cs
iSazonov: I don't like the nullable too. I see no reason to emulate registry keys. If now we use static classes it is more readable and maintainable to set defaults *explicitly*. `Force` works with this too.

src/System.Management.Automation/engine/PSConfiguration.cs
iSazonov: I do not understand completely how it should work on Unix but well imagine how it should work on Windows. What about `/etc/PowerShell`? It seems it is system-wide and should works like Windows GPOs.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: I don't think it's mimicking registry keys, but a requirement to represent three states: enabled, disabled and not defined.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: To me, it's an implementation detail to either read all settings at a time or allow the code to only read what it interests in. I don't favor one over another. The configuration APIs were already doing the latter, and I just follow that design.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: PowerShell currently doesn't do schema check on the config file, which needs to be addressed. I opened #5809 with 3 tasks. Feel free to bring up more to discuss.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
daxian-dbw: I mean "non-policy related settings", like the settings in "SOFTWARE\Microsoft\PowerShell" today.

src/System.Management.Automation/help/UpdatableHelpSystem.cs
daxian-dbw: I would rather to not re-use the same config file for modules. Synchronizing the updating of the config file among processes is expensive.

src/System.Management.Automation/engine/PropertyAccessor.cs
daxian-dbw: The querying for GP setting is a hot code path, especially for the script block logging, every script block execution will query for it. The lock `s_cachedPolicies` is to make sure that when multiple threads are querying for GP settings at the same time, only one thread will actually hit and read the configuration file.

src/System.Management.Automation/engine/PropertyAccessor.cs
SteveL-MSFT: @iSazonov since PSCore6 is not inbox in Windows and doesn't replace Windows PowerShell, we shouldn't use the same settings.  PSCore6 in this case is no different from any 3rd party app.

daxian-dbw:
@dantraMSFT @SteveL-MSFT @iSazonov @TravisEz13 I think all comments have been addressed (code updated, or question answered, or issue opened). Can you please take another look to see if the PR is ready to merge?

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -677,5 +493,129 @@ private void RemoveValueFromFile<T>(string fileName, string key)
             }
         }
     }
-} // Namespace System.Management.Automation
 
+    #region GroupPolicy Configs
+
+    /// <summary>
+    /// The GroupPolicy related settings used in PowerShell are as follows in Registry:
+    ///  - Software\Policies\Microsoft\Windows\PowerShell -- { EnableScripts (0 or 1); ExecutionPolicy (string) }
```
SteveL-MSFT: Needs to be updated as we're putting it under `...\Microsoft\PowerShellCore` now

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -497,125 +497,210 @@ internal static bool IsValidPSEditionValue(string editionValue)
...
+                    switch (typeof(T).Name)
+                    {
+                        case nameof(ScriptExecution):             result = policies.ScriptExecution; break;
+                        case nameof(ScriptBlockLogging):          result = policies.ScriptBlockLogging; break;
+                        case nameof(ModuleLogging):               result = policies.ModuleLogging; break;
+                        case nameof(ProtectedEventLogging):       result = policies.ProtectedEventLogging; break;
+                        case nameof(Transcription):               result = policies.Transcription; break;
+                        case nameof(UpdatableHelp):               result = policies.UpdatableHelp; break;
+                        case nameof(ConsoleSessionConfiguration): result = policies.ConsoleSessionConfiguration; break;
+                        default: Diagnostics.Assert(false, "Unreachable code"); break;
```
SteveL-MSFT: Perhaps have the same assert message as line 656 below?

SteveL-MSFT (changes requested):

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: Will do.

src/System.Management.Automation/engine/Utils.cs
daxian-dbw: Done.

SteveL-MSFT (approved):

dantraMSFT (approved):

TravisEz13 (approved):
We should add GPO tests at some point.

daxian-dbw:
Thanks all for the reviews!

SteveL-MSFT:
@TravisEz13 can you open an issue to track the GPO tests?

PR #5792, opened 2018-01-04
Windows installer: Allow `Launch PowerShell` checkbox to be toggled using the space bar (the Enter key already defaults to the `Finish` button)

bergmeister:
## PR Summary
Move definition of `Launch PowerShell` checkbox to the top to allow toggling it using the spacebar (the Enter key already defaults to the `Finish` button).
This is quite common amongst best practice WiX installers to make it possible to navigate through the dialogs without having to touch the mouse. This behvaiour is the same as the licence dialogue that allows toggling the checkbox using the space bar as well,
This does not affect the visual appearance.
Also reuse product name variable

## PR Checklist

- [x]  User facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission) This should go into the MSI document
- [ ] NA [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
    - [ ]  NA [Add `[feature]` if the change is significant or affectes feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

### DIFF
--- a/assets/Product.wxs
+++ b/assets/Product.wxs
@@ -35,6 +35,9 @@
     <CustomAction Id="LaunchApplication" BinaryKey="WixCA" DllEntry="WixShellExec" Impersonate="yes" />
     <UI>
       <Dialog Id="MyExitDialog" Width="370" Height="270" Title="!(loc.ExitDialog_Title)">
+        <Control Id="LaunchCheckBox" Type="CheckBox" X="10" Y="243" Width="170" Height="17" Property="LAUNCHAPPONEXIT" Hidden="yes" CheckBoxValue="1" Text="Launch $(env.ProductName)">
+          <Condition Action="show">NOT Installed</Condition>
+        </Control>
         <Control Id="Finish" Type="PushButton" X="236" Y="243" Width="56" Height="17" Default="yes" Cancel="yes" Text="!(loc.WixUIFinish)" />
         <Control Id="Cancel" Type="PushButton" X="304" Y="243" Width="56" Height="17" Disabled="yes" Text="!(loc.WixUICancel)" />
         <Control Id="Bitmap" Type="Bitmap" X="0" Y="0" Width="370" Height="234" TabSkip="no" Text="!(loc.ExitDialogBitmap)" />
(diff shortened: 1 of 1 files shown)


iSazonov:
It works. But I caught another problem: I install current build PowerShell-6.1.0-preview.7428-win-x64.msi over previous PowerShell-6.1.0-preview.7420-win-x64.msi and lost the local Explorer menu 😕

SteveL-MSFT (approved):

bergmeister:
@iSazonov : I cannot repro your problem. I installed `6.1.0-preview.7420-win-x64` first and then `6.1.0-preview.7436-win-x64` over it and the context menu is still there. Are you sure you did not forget to check the checkbox when installing it the 2nd time? Maybe it is missleading that the checkbox says `Add context menu` and the word `Add` should maybe be replaced with something like `Configure`. I might be able to change the behaviour but I would like to get clarification first what the expected behaviour should be when not checking the checkbox the 2nd time. I think what happens behind the scenes is that the upgrade uninstalls some things first (leading to the removal of registry entries) and then does the upgrade installation.
One could argue for and against either case, e.g.:
- Person A wants to upgrade PowerShell but this time decided she does not want the context menu any more, so he/she does not check the checkbox to get rid of it. In this case the current behaviour would be nice because otherwise the person has to uninstall and re-install just to get rid of the context menu.
- Person B wants to only upgrade PowerShell but keep the current configuration without having to remember the configuration. In this case the current behaviour is not good.

The only way to fulfill both cases would be to detect that it is an upgrade install and set the defaults to the current configuration. This might be quite difficult and error prone to implement.

SteveL-MSFT:
@bergmeister I would defer worrying about those cases until there is sufficient feedback we need to support it (whichever case)

iSazonov (approved):

iSazonov:
@bergmeister Never mind, I used non-rebased branch.

Issue #5597, opened 2017-12-01
Can't install 32-bit/64-bit PowerShell together when using MSI installer


stknohg:
Steps to reproduce
------------------

I installed 64-bit PowerShell 6.0-rc using MSI installer first.
Next I installed 32-bit PowerShell 6.0-rc using MSI installer.

Expected behavior
-----------------

Both 64-bit and 32-bit PowerShell is installed.

Actual behavior
---------------

64-bit PowerShell was removed.

Environment data
----------------

* 64-bit Windows 10 Pro(1703)

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.0.0-rc.2
PSEdition                      Core
GitCommitId                    v6.0.0-rc.2
OS                             Microsoft Windows 10.0.15063
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

bergmeister:
I don't have a dev machine this weekend but I think the `-arch x64` argument [here](https://github.com/PowerShell/PowerShell/blob/master/build.psm1#L2167) is the culprit since it apparently overrides the `Platform` originally define by `ProductTargetArchitecture`. Therefore WiX wrongly thinks that the 32 bit installer is 64 bit. @SteveL-MSFT Does this need to be fixed for the RC?

TravisEz13:
Yes, Thanks for reporting this bug.

bergmeister:
@TravisEz13 @SteveL-MSFT  I think we can close the issue now with the PR being merged into master and the release branch.

markekraus:
@bergmeister in the future, you can have issues auto-close on merge with special key words: https://help.github.com/articles/closing-issues-using-keywords/

bergmeister:
Thanks, I did not know that. But it also sounds dangerous because it should be a conscious decision and review of maintainers. What if the PR says 'partially fixes '? GitHub would close the issue then. VSTS has a checkbox to close related task that can be ticked upon completion of a PR.

markekraus:
@bergmeister It is something the PR assignee is responsible for checking before merging. Collaborators/maintainers/owners can update the PR details if needed before merging or they can re-open the issue. Issue watchers also get notified on close and can always request it be re-opened. I'd say the risk is low.

iSazonov:
[...]

PR #5774, opened 2018-01-02
Add option to add explorer context menu in Windows installer

bergmeister:
## PR Summary
A new dialogue is added in the Windows installer to offer the option of adding explorer.exe context menus to open PowerShell in the current location as a normal shell or as administrator. The context menu entries are achiveven via registry entries and is available when right-clicking on:

- the background of explorer
- a folder in explorer
- a drive in explorer
- on the Desktop

Adding the new custom dialog required copy-pasting the WiX source code of [WixUI_InstallDir.wxs](https://github.com/wixtoolset/wix3/blob/7c8457fa3802858add8981b92996e19574e762da/src/ext/UIExtension/wixlib/WixUI_InstallDir.wxs) as a base as described in the official WiX documentation [here](http://wixtoolset.org/documentation/manual/v3/wixui/wixui_customizations.html).
The registry entries are quite standard but the approach for the context menu of opening PowerShell as an Administrator must be discussed:
The Windows registry has the limitation that it only offers a `runas/command` structure but not a `runas/proogram/command` structure, which means there can only be one main context entry to open PowerShell as an Admin. Since a user might have alerady configured this context menu, I decided that it is best to not stomp over it but rather open PowerShell normally but then elevate it. This of course has the disadvantage that it takes a bit longer and a window is visible for a brief moment. The only alternative that would allow to launch PowerShell elevated the first time is to have the 2 context menu in a submenu, see [here](https://blogs.msdn.microsoft.com/andrew_richards/2017/03/01/enhancing-the-open-command-prompt-here-shift-right-click-context-menu-experience/)
The registry keys contain the PowerShell version to cater for side by side installations of future versions.
Some stuff had to be created in a WiX fragment, which means that this XML section could be factored out into its own file (with the disadvantage of having to linkt this file then into the compilation process of candle,, light, etc.)
The best person to review is probably @SteveL-MSFT 

This is how it currently looks like: 
Installer: 
image
Explorer context menu:
image

## PR Checklist

Note: Please mark anything not applicable to this PR `NA`.

- [x] [PR has a meaningful title](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - [x] Use the present tense and imperative mood when describing your changes
- [x] [Summarized changes](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
[...]

PR #5812, opened 2018-01-07
Fix x86 installer bug: it gets installed as an x64 component and has the same UpgradeCode as the x64 installer leading to an uninstallation of an x64 installation (and vice versa)

bergmeister:
## PR Summary

Fix bug #5597: x86/x64 installer are uninstalling each other when installing either of them:

-Make x86 installer to be installed as a x86 component (-arch argument to candle.exe, which sets the `sys.BUILDARCH` variable)
-Make the UpgradeCode unique per platform

Replace `var.ProductTargetArchitecture` variable with sys.BUILDARCH use to have only 1 variable for the architecture
Additionally, the architecture was appended to the package name to be able to distinguish the installations.

Windows installer PRs usually got reviewed by @SteveL-MSFT in the past.

## PR Checklist

- [x] User facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission) : Not sure? Package name now includes architecture...
    - [x] Issue filed - Issue link: https://github.com/PowerShell/PowerShell/issues/5597
- [ ] NA [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
    - [] NA [Add `[feature]` if the change is significant or affectes feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

### DIFF
--- a/assets/Product.wxs
+++ b/assets/Product.wxs
@@ -20,7 +20,7 @@
   <!-- Explorer context submenu entries. The ampersand denotes the keyboard shortcut. -->
   <?define ExplorerContextMenuDialogText = "Open &here"?>
   <?define ExplorerContextMenuElevatedDialogText = "Open here as &Administrator"?>
-
+  <!-- The ProductCode is Product Id: http://wixtoolset.org/documentation/manual/v3/xsd/wix/product.html -->
   <Product Id="$(var.ProductGuid)" Name="$(var.ProductSemanticVersionWithName)-$(sys.BUILDARCH)" Language="1033" Version="$(var.ProductVersion)" Manufacturer="Microsoft Corporation" UpgradeCode="$(var.UpgradeCode)">
     <!-- Properties About The Package -->
     <Package Id="*" Keywords="Installer" Platform="$(sys.BUILDARCH)" InstallerVersion="200" Compressed="yes" InstallScope="perMachine" Description="PowerShell package" Comments="PowerShell for every system" />
--- a/assets/Product.wxs
+++ b/assets/Product.wxs
@@ -6,21 +6,24 @@
   <?define InfoURL="https://github.com/PowerShell/PowerShell" ?>
   <?define ProductName = "$(env.ProductName)" ?>
   <?define ProductGuid = "$(env.ProductGuid)" ?>
+  <!-- UpgradeCode GUID MUST REMAIN SAME THROUGHOUT ALL VERSIONS, otherwise, updates won't occur. -->
+  <?if $(sys.BUILDARCH)=x64?>
+    <?define UpgradeCode = "31ab5147-9a97-4452-8443-d9709f0516e1" ?>
+  <?else?>
+    <?define UpgradeCode = "1d00683b-0f84-4db8-a64f-2f98ad42fe06" ?>
+  <?endif?>
   <?define ProductVersion = "$(env.ProductVersion)" ?>
   <?define ProductSemanticVersion = "$(env.ProductSemanticVersion)" ?>
   <?define ProductVersionWithName = "$(var.ProductName)_$(var.ProductVersion)"?>
   <?define ProductSemanticVersionWithName = "$(var.ProductName)-$(env.ProductSemanticVersion)"?>
-  <?define ProductTargetArchitecture = "$(env.ProductTargetArchitecture)"?>
   <?define ProductProgFilesDir = "$(env.ProductProgFilesDir)" ?>
   <!-- Explorer context submenu entries. The ampersand denotes the keyboard shortcut. -->
   <?define ExplorerContextMenuDialogText = "Open &here"?>
   <?define ExplorerContextMenuElevatedDialogText = "Open here as &Administrator"?>
-  <!-- Generate Your Own GUID for both ID and UpgradeCode attributes. -->
-  <!-- Note:  UpgradeCode GUID MUST REMAIN SAME THROUGHOUT ALL VERSIONS -->
-  <!-- Otherwise, updates won't occur -->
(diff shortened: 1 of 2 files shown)


bergmeister:
@SteveL-MSFT @joeyaiello I am not sure if this bug was ever triaged but it seems quite severe to me and would cause problems in the future when trying to upgrade install the x86 version in 6.1 if this fix does not go into 6.0. On the other hand, I can understand that this is last minute (maybe building only the x86 installer using this fix would mitigate the risk?)...

iSazonov:
Should Explorer menu be fixed too?

bergmeister:
@iSazonov : The explorer context menu goes into 6.1 (I already asked this question [here](https://github.com/PowerShell/PowerShell/pull/5774#issuecomment-355353375)).  @SteveL-MSFT said [here](https://github.com/PowerShell/PowerShell/pull/5792#issuecomment-355709921) that it is not clear whether the current behaviour is right or wrong and therefore wait for feedback.

iSazonov:
@bergmeister My request was that in the previous PR we added the menu and in this PR we add a second PowerShell version - what version will be called from this menu? I suppose that if we want this menu, then in this PR you have to add the menu for both versions.

bergmeister:
@iSazonov I see, you have a fair point that I did not consider the case of installing both x64 and x86 for the context menu.
But this bug fix itself has nothing to do with the context menu since it should have always worked to install both versions, hence the question to steve if this fix needed for the release branch as well. At the moment the PR targets fixing the installer problem so that it is also easily possible to merge it into the release branch if needed/required. I shall open a 2nd WIP PR for the architecture specific context menu to separate the issues.

SteveL-MSFT:
I think the number of users using both x86 and x64 are very minimal.  Giving that where we are with 6.0.0, I think we should document this (Known Issues @joeyaiello) and fix it for 6.1.0.

SteveL-MSFT (approved):

TravisEz13:
I think we need to take this for GA to ensure that we can create servicing packages correctly without creating a complicated servicing wrapper.

SteveL-MSFT:
I talked to @TravisEz13 offline, agree that servicing would be an issue so we should take this for GA.

TravisEz13:
filed:  https://github.com/PowerShell/PowerShell/blob/master/build.psm1#L2071

TravisEz13 (approved):

bergmeister:
OK. I will prepare now a separate PR that targets the release branch to check for merge problems beforehand.

assets/Product.wxs
```diff
@@ -6,21 +6,24 @@
...
   <?define ProductSemanticVersionWithName = "$(var.ProductName)-$(env.ProductSemanticVersion)"?>
-  <?define ProductTargetArchitecture = "$(env.ProductTargetArchitecture)"?>
   <?define ProductProgFilesDir = "$(env.ProductProgFilesDir)" ?>
   <!-- Explorer context submenu entries. The ampersand denotes the keyboard shortcut. -->
   <?define ExplorerContextMenuDialogText = "Open &here"?>
   <?define ExplorerContextMenuElevatedDialogText = "Open here as &Administrator"?>
-  <!-- Generate Your Own GUID for both ID and UpgradeCode attributes. -->
-  <!-- Note:  UpgradeCode GUID MUST REMAIN SAME THROUGHOUT ALL VERSIONS -->
-  <!-- Otherwise, updates won't occur -->
-  <Product Id="$(var.ProductGuid)" Name="$(var.ProductSemanticVersionWithName)" Language="1033" Version="$(var.ProductVersion)" Manufacturer="Microsoft Corporation" UpgradeCode="{f7ba3e58-0be8-443b-ac91-f99dd1e7bd3b}">
```
TravisEz13: Can you add an comment that `<Product Id` is the product code?
https://stackoverflow.com/questions/10330534/in-wix-where-is-the-productcode-specified
http://wixtoolset.org/documentation/manual/v3/xsd/wix/product.html

TravisEz13:
@bergmeister I will take care of any merge conflicts with the release branch.  Just get it working on this branch.  I've finished my review.  It looks good otherwise.

assets/Product.wxs
bergmeister: @TravisEz13 : I created a separate PR for 6.0.0 [here](https://github.com/PowerShell/PowerShell/pull/5822). The difference is that it is based on the 6.0.0. branch where I cherry picked the fixing commit (an auto-merge of whitespace was needed) so that local testing becomes easier. I have addressed your comment about adding the product code related comment.
This PR shall now as WIP and for 6.1.0.
For the record: the mention context menu issue applies only to 6.1 and is not related to this PR

bergmeister:
OK. Whatever you prefer then.

TravisEz13:
Can you get the comment added in the next few hours?

bergmeister:
I shall do it now, in 1 min.

TravisEz13 (approved):

TravisEz13:
@bergmeister Thanks

bergmeister:
No worries, my pleasure. I am glad I found it by coincidence (I guess the bug must have been in there since day one). I think the next time we have to ensure that issues get triaged better because if the impact was understood better from the reported issue then it would not have been so much last minute.

bergmeister:
I did some analysis and it showed that the bug about the upgrade code being constant and the `-arch` parameter being hardcoded to `x64` was already present in the first commit [here](https://github.com/PowerShell/PowerShell/commit/8f1a5194c563a3a74e7b92181c87b9e33803c4b4) that added the WiX file and the MSI build script...

Issue #5725, opened 2017-12-20
Future of Add-Type 


iSazonov:
Current version of Add-Type cmdlet is _partially_ ported on Roslyn from Dom. It is large breaking change. It is true for most of the parameters. Some parameters was related to Dom and was removed. Some parameters don't implemented. _In fact, we already have a new cmdlet._
Before I push new enhancements I want discuss a few points. We could remove old things.

 -  [public enum `Language`](https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs#L33)
Should we keep JScript for backward compatibility?
Should we add FSharp for future?
Should we keep CSharpVersion1, CSharpVersion2 and so on? If yes should we add VisualBasicVersion9 and so on? Could we leave only CSharp and VisualBasic in the enum? Why anybody can want to use CSharpVersion1 in PowerShell?! I found nothing on GitHub - I think we should old elements.
Currently I have to add language family type with CSharp and VisualBasic internally. So I'd prefer simplify `public enum Language` and don't use the new type and helper functions to translate these types.
- Roslyn can accept parsing and compile options. It is _different types_ for every language (currently CSharp and VusualBasic). Some options (not all) can be useful for Add-Type (language version, preprocessor variables). 
How we can better expose the options? Generic or by language? The question related with previous one - with options we can specify the language version. We could Roslyn generic classes but they are not convenient to work in PowerShell. I have no conclusion.
- Should we support creation new assembly without loading? Currently we always load created assembly, even with `-Passthrou`. I think we could add `-CompileOnly` switch.

/cc @SteveL-MSFT @daxian-dbw

SteveL-MSFT:
`Add-Type` was a big change to dotnetcore so I'm personally ok with not maintaining strict compatibility for legacy features that people don't use (with sufficient data).

Ideally, we should use dynamic parameters per language.

I think adding a `-CompileOnly` switch would be useful.

HerbM:
Please do add FSharp.

Current versions of FSharp.Compiler.Codedom seem to be broken relative to incorrect version of FSharp.Core -- and no amount of installing packages for various versions had helped me -- or asking on FSharp Slack.

Whose problem would this be?  Where should it be reported?
(I think it is stuck half-way between PowerShell & FSharp with no one to care.

iSazonov:
[...]

PR #5829, opened 2018-01-09
Cleanup Add-Type

iSazonov:
## PR Summary
Related #5725.

It's part of the big PR I'm trying to break into smaller pieces.
Subsequent PRs will focus on enhancements.

- Remove CSharp language versions and JScript from Language enum. Later we can use Roslyn parse option if we need set a language version. JScript is not in Roslyn. It is a breaking change (Approved by @SteveL-MSFT in #5725).
- Correct comments.
- Remove IgnoreWarnings parameter.  It is not implemented and no plans do that. We could use the standard parameter `WarningAction` if we'll need.

## PR Checklist

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
@@ -602,7 +602,7 @@ namespace Microsoft.PowerShell.Commands
                     }
                     break;
                 default:
-                        Diagnostics.Assert(false, "GetUsingSet: Unsupported language family.");
+                    Diagnostics.Assert(false, "GetUsingSet: Unsupported language family.");
                     break;
             }
 
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
@@ -602,9 +602,7 @@ namespace Microsoft.PowerShell.Commands
                     }
                     break;
                 default:
-                    {
                         Diagnostics.Assert(false, "GetUsingSet: Unsupported language family.");
-                    }
                     break;
             }
 
@@ -1025,15 +1023,7 @@ namespace Microsoft.PowerShell.Commands
             CSharpParseOptions parseOptions;
             if (Language == Language.CSharp)
             {
-                switch (Language)
-                {
-                    case Language.CSharp:
-                        parseOptions = new CSharpParseOptions();
-                        break;
-                    default:
-                        parseOptions = null;
-                        break;
-                }
+                parseOptions = new CSharpParseOptions();
             }
             else
             {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
@@ -483,6 +483,12 @@ namespace Microsoft.PowerShell.Commands
         [Parameter()]
         public SwitchParameter PassThru { get; set; }
 
+        /// <summary>
+        /// Flag to ignore warnings during compilation.
+        /// </summary>
+        [Parameter()]
+        public SwitchParameter IgnoreWarnings { get; set; }
+
         #endregion Parameters
 
         #region GererateSource
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
@@ -33,73 +33,33 @@ namespace Microsoft.PowerShell.Commands
     public enum Language
     {
         /// <summary>
-        /// The C# programming language: latest version.
+        /// The C# programming language.
         /// </summary>
         CSharp,
 
         /// <summary>
-        /// The C# programming language v7
+        /// The Visual Basic programming language.
         /// </summary>
-        CSharpVersion7,
-
-        /// <summary>
-        /// The C# programming language v6
-        /// </summary>
-        CSharpVersion6,
-
-        /// <summary>
-        /// The C# programming language v5
-        /// </summary>
-        CSharpVersion5,
-
-        /// <summary>
-        /// The C# programming language v4
-        /// </summary>
-        CSharpVersion4,
-
-        /// <summary>
-        /// The C# programming language v3 (for Linq, etc)
-        /// </summary>
-        CSharpVersion3,
-
-        /// <summary>
-        /// The C# programming language v2
-        /// </summary>
-        CSharpVersion2,
-
-        /// <summary>
-        /// The C# programming language v1
-        /// </summary>
-        CSharpVersion1,
-
-        /// <summary>
-        /// The Visual Basic programming language
-        /// </summary>
-        VisualBasic,
-
-        /// <summary>
-        /// The Managed JScript programming language
-        /// </summary>
-        JScript,
+        VisualBasic
     }
 
     /// <summary>
-    /// Types supported for the OutputAssembly parameter
+    /// Types supported for the OutputAssembly parameter.
     /// </summary>
     public enum OutputAssemblyType
     {
         /// <summary>
-        /// A Dynamically linked library (DLL)
+        /// A Dynamically linked library (DLL).
         /// </summary>
         Library,
 
         /// <summary>
-        /// An executable application that targets the console subsystem
+        /// An executable application that targets the console subsystem.
         /// </summary>
         ConsoleApplication,
 
         /// <summary>
-        /// An executable application that targets the graphical subsystem
+        /// An executable application that targets the graphical subsystem.
         /// </summary>
         WindowsApplication
     }
@@ -148,6 +108,8 @@ namespace Microsoft.PowerShell.Commands
     [OutputType(typeof(Type))]
     public sealed class AddTypeCommand : PSCmdlet
     {
+        #region Parameters
+
         /// <summary>
         /// The source code of this type.
         /// </summary>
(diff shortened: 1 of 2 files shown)


iSazonov:
@SteveL-MSFT We need an approve from PowerShell Committee.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
```diff
@@ -553,149 +509,110 @@ internal string GenerateTypeSource(string typeNamespace, string name, string sou
...
+                case Language.VisualBasic:
+                    foreach (string namespaceValue in UsingNamespace)
+                    {
+                        usingNamespaceSet.Append("Imports " + namespaceValue + "\n");
+                    }
+                    break;
+                default:
+                    {
+                        Diagnostics.Assert(false, "GetUsingSet: Unsupported language family.");
+                    }
```
daxian-dbw: No need to put the assert in a code block `{}`
One more minor thing -- the assert is put in `default` case here while not in the above similar switch code, maybe you can keep them consistent.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
```diff
@@ -1106,31 +1023,10 @@ private void WriteTypes(Assembly assembly)
         private void CompileSourceToAssembly(string source)
         {
             CSharpParseOptions parseOptions;
-            if (IsCSharp(Language))
+            if (Language == Language.CSharp)
             {
                 switch (Language)
```
daxian-dbw: No need to switch on `Language` anymore, since you already checked `Language == Language.CSharp`

daxian-dbw (changes requested):
Left two comments

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
iSazonov: The PR is formal cleanup to reduce the size of subsequent PRs to simplify a code review (my develop branch has over 1200 changed lines and I still haven't done all the work) - latter we'll again edit the code. So I don't think it makes sense to make today such minor changes.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
daxian-dbw: The same. If you are going to push more commits to the same PR, then ignore this comment for now. Otherwise, this comment needs to be resolved for this PR to be merged.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
daxian-dbw: If you are going to push more commits to the same PR, then just ignore the comments now.
If you want this PR to be merged before doing the further work, then I think you need to resolve those comments.

daxian-dbw: > One more minor thing --

The minor part is not a blocking comment, you can leave it as is.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this:
- ok with removing jscript since it's not supported by Roslyn
- ask is to investigate if passing the different versions of csharp are supported today with current implementation; prefer to not break existing users who may be passing the csharp version even though we agree that most users are likely not caring about specific version of csharp

So pending outcome of second bullet, we can decide if it's appropriate to remove the csharp versions from the enum.  Leaving as `Review - Committee` pending investigation.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
iSazonov: About `Diagnostics.Assert` - code block removed.
About minor comment - here we should return value, above we use formal `return null;`

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
iSazonov: `switch` removed.

I want keep the PR small. No plans add new code.

iSazonov:
1. GitHub search results (without repetitions and `Add-Type clones`):

| Version | Script | C#
|- | - | -
|`CSharpVersion3` | 118 | 0
|`CSharpVersion2` | 1 | 0
|`CSharpVersion1` | 1 | 0

`CSharpVersion4`, `CSharpVersion5`, `CSharpVersion6`, `CSharpVersion7`, `CSharpVersion7` was added in 
PowerShell Core 6.0. Nobody still use it.

Note that this cmdlet is rarely used in compilation mode, and with these exact language settings it is much less.

2. I don't know scenarios where users should use lower C# version with Roslyn. Why would anyone want to block new versions or features in his _own_ code? 
 I rather expect that users will request to add support to latest Roslyn and C# versions. Currently we have to change the cmdlet to add new version explicitly but if we'll use native Roslyn types we get this automatically. (With .Net Core hosting model even more automatically).

3. @SteveL-MSFT suggested using dynamic parameters because Roslyn exposes different types for different languages. This allow us to expose exact language version strings in a parameter in any cases. I don't investigate but I guess we could to try emulate current `-Language` behavior if we could give up current `Language` enum type.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs
daxian-dbw: Thanks for the update. I removed the extra spaces before `Diagnostics.Assert`

daxian-dbw (approved):

daxian-dbw:
More information for the committee:
### In Windows PowerShell 5.1
the enum `Microsoft.PowerShell.Commands.Language` has the following members: `CSharp`, `CSharpVersion3`, `CSharpVersion2`, `VisualBasic`, `JScript`.
Among them, `CSharpVersion2` requires Microsoft .NET Framework 2 to be installed and `CSharpVersion3` requires Microsoft .NET Framework 3.5 to be installed. Therefore, on PSv5.1 these 2 options don't work unless the user manually installs .NET 2 and .NET 3.5. See examples below:
```
PS:2> Add-Type -TypeDefinition $source -Language CSharpVersion2
Add-Type : Cannot add type due to the following exception: Compiler executable file csc.exe cannot be found.. Verify
that Microsoft .NET Framework 2  is installed. On x64-based versions of Windows, you must also install the WOW64
feature.
At line:1 char:1
+ Add-Type -TypeDefinition $source -Language CSharpVersion2
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Add-Type], InvalidOperationException
    + FullyQualifiedErrorId : System.InvalidOperationException,Microsoft.PowerShell.Commands.AddTypeCommand

PS:3> Add-Type -TypeDefinition $source -Language CSharpVersion3
Add-Type : Cannot add type due to the following exception: Compiler executable file csc.exe cannot be found.. Verify
that Microsoft .NET Framework 3.5  is installed. On x64-based versions of Windows, you must also install the WOW64
feature.
At line:1 char:1
+ Add-Type -TypeDefinition $source -Language CSharpVersion3
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Add-Type], InvalidOperationException
    + FullyQualifiedErrorId : System.InvalidOperationException,Microsoft.PowerShell.Commands.AddTypeCommand
```

### In PowerShell Core
The following members were added to the enum `Language` when refactoring the code with Roslyn: `CSharpVersion1`, `CSharpVersion4`, `CSharpVersion5`, `CSharpVersion6`, `CSharpVersion7`.
Based on the search results from GitHub, none of them are used in any PowerShell scripts except in our tests.

### Summary

Given that `CSharpVersion2` and `CSharpVersion3` by default don't work in PSv5.1 (maybe PSv5 too), maybe it's not a too bad breaking change to remove them from `Language`. As for the other `CSharpVersionN` members, none of them are used anywhere.

iSazonov:
@SteveL-MSFT I'm really looking forward 😄

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and would be fine with removing the specific versions of csharp from the enum; we do not believe there is utility to having the different versions of the csharp language specified and for users that have specified a version they will get an error indicating to just use `csharp`.  

We will be reviewing the `Breaking Change Contract` and updating as necessary to reflect how we came to this decision.

daxian-dbw:
The `-IgnoreWarnings` parameter shouldn't be removed yet. It's implemented in Windows PowerShell by setting [`CompilerParameters.TreatWarningsAsErrors = !((bool)ignoreWarnings)`](https://github.com/PowerShell/PowerShell/blob/v6.0.0-beta.1/src/Microsoft.PowerShell.Commands.Utility/commands/utility/AddType.cs#L1723).

The parameter was not implemented when porting to Roslyn in PowerShell Core, but that doesn't mean it should be removed from PowerShell Core. I think it makes sense to support it in PowerShell Core.

I pushed a commit to add this parameter back. Work can be done in future to actually support it in PowerShell Core.

iSazonov:
@daxian-dbw I agree with `-IgnoreWarnings`. I removed it because the name does not reflect the meaning - treat warnings as errors. I plan to add new parameter with name like `-WarnAsErrors` - we could add `IgnoreWarnings` as alias. 
Also I plan add support for preprocessor constants. The rest does not look so useful - if you have other considerations please let me know.

daxian-dbw:
> I plan to add new parameter with name like -WarnAsErrors - we could add IgnoreWarnings as alias.

That idea would work too :) The bottom line is that we need to continue to support this parameter in PowerShell Core. @iSazonov, thanks for initiating the rework of `Add-Type`!

PR #5839, opened 2018-01-10
Set-Location should use path with wildcard characters if it exists instead of globbing

### LINKED ISSUE #5752: Powershell does not open in current directory if its name contains a left square bracket (wildcard)
Steps to reproduce
------------------

```powershell
C:\>mkdir "[T]est"

C:\>cd "[T]est"

C:\[T]est>pwsh.exe
PowerShell v6.0.0-rc.2

https://aka.ms/pscore6-docs
Type 'help' to get help.

PS C:\Programs\PowerShell\6.0.0-rc.2>
```
1. Make a directory called "[T]est".
2. Run cmd (or powershell) and cd to the directory.
3. Run powershell.

Expected behavior
-----------------

```none
Powershell always starts in the folder the parent process runs it in, regardless of the name of the folder.
```

Actual behavior
---------------

```none
3 observed behaviors:
- Powershell starts in its home directory if the folder has left and right brackets. (ex: "Te[st]")
- Powershell starts in C:\ if the folder only has left brackets. (ex: "Te[st")
- Powershell starts in the current directory if it only has right brackets. (ex: "Te]st")
```

Environment data
----------------

```powershell
> $PSVersionTable
```

Additional Information
----------------

Hope this isn't a known issue, I've experienced it myself both when starting powershell from cmd and from Visual Studio Code ("Open in Terminal"). I've seen it discussed a little bit [here](http://mxii.eu.org/2017/03/19/powershell-explorer-square-brackets-and-unicode-paths/). It might be related to Issue #4726. Hope we can see this fixed.

SteveL-MSFT:
## PR Summary
When InitialSessionState initializes it tries to SetLocation to current working directory,
but if the directory name contains PowerShell wildcard characters, it fails and reverts
to $PSHOME.  
The change affects Set-Location in that if the path exists (even if containing wildcard characters), just use it.

Fix https://github.com/PowerShell/PowerShell/issues/5752

## PR Checklist

- [na] User facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)

### DIFF
--- a/src/System.Management.Automation/namespaces/LocationGlobber.cs
+++ b/src/System.Management.Automation/namespaces/LocationGlobber.cs
@@ -502,7 +502,8 @@ namespace System.Management.Automation
 
             Collection<string> stringResult = new Collection<string>();
 
-            if (!context.SuppressWildcardExpansion)
+            // if the directory contains wildcard characters, but exists, just return it
+            if (!context.SuppressWildcardExpansion && !Utils.NativeDirectoryExists(userPath))
             {
                 // See if the provider will expand the wildcard
                 if (CmdletProviderManagementIntrinsics.CheckProviderCapabilities(
--- a/src/System.Management.Automation/namespaces/LocationGlobber.cs
+++ b/src/System.Management.Automation/namespaces/LocationGlobber.cs
@@ -502,8 +502,21 @@ namespace System.Management.Automation
 
             Collection<string> stringResult = new Collection<string>();
 
-            // if the directory contains wildcard characters, but exists, just return it
-            if (!context.SuppressWildcardExpansion && !Utils.NativeDirectoryExists(userPath))
+            // if the directory exists, just return it
+            try
+            {
+                if (Utils.NativeDirectoryExists(userPath))
+                {
+                    result.Add(new PathInfo(drive, provider, userPath, _sessionState));
+                    return result;
+                }
+            }
+            catch
+            {
+                // in cases of Access Denied or other errors, fallback to previous behavior and let provider handle it
+            }
+
+            if (!context.SuppressWildcardExpansion)
             {
                 // See if the provider will expand the wildcard
                 if (CmdletProviderManagementIntrinsics.CheckProviderCapabilities(
--- a/test/powershell/Host/Base-Directory.Tests.ps1
+++ b/test/powershell/Host/Base-Directory.Tests.ps1
@@ -118,6 +118,7 @@ Describe "Configuration file locations" -tags "CI","Slow" {
 
 Describe "Working directory on startup" -Tag "CI" {
     BeforeAll {
+        $powershell = Join-Path -Path $PSHOME -ChildPath "pwsh"
         $testPath = New-Item -ItemType Directory -Path "$TestDrive\test[dir]"
         $currentDirectory = Get-Location
     }
@@ -128,6 +129,12 @@ Describe "Working directory on startup" -Tag "CI" {
 
     It "Can start in directory where name contains wildcard characters" {
         Set-Location -LiteralPath $testPath.FullName
-        & $powershell -noprofile -c get-location | Should BeExactly $testPath.FullName
+        if ($IsMacOS) {
+            # on macOS, /tmp is a symlink to /private so the real path is under /private/tmp
+            $expectedPath = "/private" + $testPath.FullName
+        } else {
+            $expectedPath = $testPath.FullName
+        }
+        & $powershell -noprofile -c { $PWD.Path } | Should BeExactly $expectedPath
     }
 }
--- a/test/powershell/Host/Base-Directory.Tests.ps1
+++ b/test/powershell/Host/Base-Directory.Tests.ps1
@@ -115,3 +115,19 @@ Describe "Configuration file locations" -tags "CI","Slow" {
         }
     }
 }
+
+Describe "Working directory on startup" -Tag "CI" {
+    BeforeAll {
+        $testPath = New-Item -ItemType Directory -Path "$TestDrive\test[dir]"
+        $currentDirectory = Get-Location
+    }
+
+    AfterAll {
+        Set-Location $currentDirectory
+    }
+
+    It "Can start in directory where name contains wildcard characters" {
+        Set-Location -LiteralPath $testPath.FullName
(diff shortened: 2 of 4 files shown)


test/powershell/Host/Base-Directory.Tests.ps1
```diff
@@ -115,3 +115,19 @@ Describe "Configuration file locations" -tags "CI","Slow" {
...
+        $currentDirectory = Get-Location
+    }
+
+    AfterAll {
+        Set-Location $currentDirectory
+    }
+
+    It "Can start in directory where name contains wildcard characters" {
+        Set-Location -LiteralPath $testPath.FullName
+        & $powershell -noprofile -c get-location | Should BeExactly $testPath.FullName
```
iSazonov: `$powershell` don't  defined in the Describe block.

SteveL-MSFT: Will fix.  Broke when I made this test it's own Describe.

iSazonov (approved):

lzybkr (commented):
The code change is surprising given the PR title. Based on the code, I would have thought that Set-Location was broken. Given that it's not, I'm guessing similar code exists elsewhere and either:
a) that code is not necessary
b) that code is subtly different than this fix

So I'd ask for a little more digging. You might also look for other ways this code is reachable - that might suggest the need for more tests.

SteveL-MSFT:
@lzybkr Changed title a bit to clarify where the issue is.  This internal `SetLocation` method doesn't differentiate between `path` and `literalpath` (it's always `path`).  The *-Location cmdlets do end up using this same path but sets SuppressWildcardExpansion to true when `-LiteralPath` is used.

lzybkr:
With your change - I think `Set-Location -Path` has a change in behavior. Consider:

```
mkdir aa
mkdir ba
mkdir [ab]a
Set-Location [ab]a
```

Before your change, this reports an error, after it works without error.

I think this is desirable - it almost seems like `Set-Location` never should have had the `-LiteralPath` parameter in the first place, but I've only thought about this for like a second.

SteveL-MSFT:
@lzybkr you are correct regarding the behavior change in `set-location -path` and agree it's probably the right thing.  I'll add `breaking change` although I don't think anyone will be impacted.

iSazonov:
If `[ab]a` is allowed path I'd said that it is not a breaking change, but bug fix.
Also I'd add new explicit test for `Set-Location -Path [ab]a`.

SteveL-MSFT:
After adding test for `Set-Location`, it didn't actually do what I thought it would.  Apparently the check I had wasn't working as the path was relative and thus didn't "exist".  Put the change in the LocationGlobber so now `Set-Location` works as discussed.  Appreciate the feedback.

src/System.Management.Automation/namespaces/LocationGlobber.cs
```diff
@@ -502,7 +502,8 @@ out providerInstance
 
             Collection<string> stringResult = new Collection<string>();
 
-            if (!context.SuppressWildcardExpansion)
+            // if the directory contains wildcard characters, but exists, just return it
+            if (!context.SuppressWildcardExpansion && !Utils.NativeDirectoryExists(userPath))
```
iSazonov: Maybe the comment and logic should be "if the directory exists, just return it"?

SteveL-MSFT: I'm leveraging falling into the `else` clause which is the same as if wildcard expansion is suppressed.  Alternatively, I could rearrange the code so that the suppression/existing happens first which may make it a bit easier to read:

```csharp
if (context.SuppressWildcardExpansion || Utils.NativeDirectoryExists(userPath))
...
```

SteveL-MSFT: Need to change this code to handle exceptions during check

iSazonov (approved):
LGTM

iSazonov:
@lzybkr Could you please continue the review?

lzybkr (approved):
I sign off.

You should squash the commits and write a good commit message with an explanation of the breaking change so that appears in the git history.

iSazonov:
@SteveL-MSFT Could you please edit PR header and description to address Jason request?

SteveL-MSFT:
Done

KillyMXI:
I'm a bit late to the party, but I hope it's OK now, while 6.1.0 still in preview.

Something with implemented fix still bothers me.
When started playing with v6.1.0-preview.2, I realized that `Get-ChildItem` is also affected by this change.
And everything else using wildcard expansion now has added conditional branch with try-catch (what kind of performance overhead it might have, btw?)

```
PS C:\ps_test> Get-ChildItem


    Directory: C:\ps_test


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----       24.05.2018      2:03                [ab]a
d-----       24.05.2018      2:03                aa
d-----       24.05.2018      2:03                ba


PS C:\ps_test> Get-ChildItem [ab]a


    Directory: C:\ps_test


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----       24.05.2018      2:03                [ab]a


PS C:\ps_test> Get-ChildItem [abc]a


    Directory: C:\ps_test


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----       24.05.2018      2:03                aa
d-----       24.05.2018      2:03                ba
```

While this might even make it work as expected more times than before, it has an increased [level of surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). It might make issues caused by the use of `-Path` instead of `-LiteralPath` to show up less frequently but will make them harder to trace when they finally show up.
Using wildcards by default and literal path as an option was a major design mistake, imo. This patch makes a step back, but in a way that makes it less predictable. Is it OK to have more "magic"?

I also wonder why it was made in this way instead of just calling `Set-Location` with `-LiteralPath` when initializing a session? Working folder is by definition a literal path. It would've kept behavior changes to minimum.

SteveL-MSFT:
@KillyMXI you bring up some good points, can you open a new issue to discuss this?

KillyMXI:
Done #6927

Issue #4306, opened 2017-07-20
Implement Counter CmdLets using Supported APIs


TravisEz13:
Implement Counter CmdLets using Supported APIs
The previous implementation is here:
https://github.com/PowerShell/PowerShell/tree/master/src/Microsoft.PowerShell.Commands.Diagnostics
Which was removed due to #4272

iSazonov:
Could you please clarify what is "unsupported API" in previous implementation?

TravisEz13:
I'll restate @iSazonov question which I removed.

> How do I know what APIs are supported?

You can assume that APIs documented in MSDN are supported as long as you follow the requirements in MSDN.

SteveL-MSFT:
System.Diagnostics.PerformanceCounter is part of WindowsCompatibilityPack for dotnetcore so this work is being done as part of WindowsPowerShellCompatibilityPack

iSazonov:
@SteveL-MSFT The problem is that there is no public API - MSFT can distribute the cmdlets only in  _binary package._

SteveL-MSFT:
@iSazonov need to figure out how to handle non-public APIs as I think only components shipped with Windows can use those APIs so even a binary package on PSGallery wouldn't be acceptable

iSazonov:
It seems the API is already ported to .Net Core and present in WCP https://github.com/dotnet/corefx/issues/3906

TravisEz13:
Specifically https://www.nuget.org/packages/System.Diagnostics.PerformanceCounter

iSazonov:
I fast review the CoreFX code - it seems it implement only a counter reader but not file (blg) reader/writer. This is enough for `Get-Counter` but not for `Import/Export-Counter`.
We should make a request in CoreFX if we want `System.Diagnostics.Eventing.Reader` and etc.

bill-long:
Is this being considered for Powershell 7? I noticed Get-Counter still doesn't work in the preview.

TravisEz13:
@bill-long I think our current thought is to get the Windows Inbox module to work.  It currently works but has a naming conflict with with a PowerShell 7 module (also meaning both modules cannot be loaded at the same time).  We considering a solution that resolves the naming conflict.

PR #5845, opened 2018-01-10
WIP - Fix performance and memory allocation issues in ReadPdhMultiString.

jcarino:
## PR Summary

String concats in ReadPdhMultiString have been replaced with a StringBuilder to significantly reduce heap allocation and GC.

ReadPdhMultiString reads characters one by one from a pointer location into a string, then splits that string into a string collection (on the null character '\0'). However, it does this through a looped string concat, allocating a new string on the heap for each character read in.  This has significant GC implications for large strings.  The string concats have been replaced with a string builder, and the strings are split and added to the StringCollection on the fly.

## PR Checklist

- [NA ] User facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - [ NA] Issue filed - Issue link:

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Diagnostics/PdhHelper.cs
+++ b/src/Microsoft.PowerShell.Commands.Diagnostics/PdhHelper.cs
@@ -3,6 +3,7 @@
 //
 
 using System;
+using System.Text;
 using System.Diagnostics;
 using System.Globalization;
 using System.Runtime.InteropServices;
@@ -485,7 +486,7 @@ namespace Microsoft.Powershell.Commands.GetCounter.PdhNative
         {
             Debug.Assert(strSize >= 2);
             int offset = 0;
-            string allSubstringsWithNulls = "";
+            StringBuilder sb = new StringBuilder(strSize);
             while (offset <= ((strSize * sizeof(char)) - 4))
             {
                 Int32 next4 = Marshal.ReadInt32(strNative, offset);
@@ -494,14 +495,19 @@ namespace Microsoft.Powershell.Commands.GetCounter.PdhNative
                     break;
                 }
 
-                allSubstringsWithNulls += (char)next4;
+                char nextChar = (char)next4;
+                if (nextChar == '\0')
+                {
+                    strColl.Add(sb.ToString());
+                    sb.Clear();
+                }
+                else
+                {
+                    sb.Append(nextChar);
+                }
 
                 offset += 2;
             }
-
-            allSubstringsWithNulls = allSubstringsWithNulls.TrimEnd('\0');
-
-            strColl.AddRange(allSubstringsWithNulls.Split('\0'));
         }
 
 


iSazonov:
@jcarino We can not merge the contribution until you sign CLA.

iSazonov:
Ah, `PdhHelper.cs` isn't used currently and excluded from compiling.

jcarino:
CLA has now been signed.
I know this module is excluded due to issue #4303 but this is an issue in production, and if/when related modules are added back this might be useful.

src/Microsoft.PowerShell.Commands.Diagnostics/PdhHelper.cs
```diff
@@ -494,14 +495,19 @@ private void ReadPdhMultiString(ref IntPtr strNative, Int32 strSize, ref StringC
                     break;
                 }
 
-                allSubstringsWithNulls += (char)next4;
+                char nextChar = (char)next4;
+                if (nextChar == '\0')
+                {
+                    strColl.Add(sb.ToString());
```
dantraMSFT: Adding to strColl should occur outside the while loop. The updated logic causes the contents of sb to be dropped on the floor by the break statement when next4 == 0.

Essentially, replace the original `allSubstringsWithNulls += (char)next4` statement with `sb.Append(nextChar) `and move the `strColl.Add` statement to the original location after the while loop.

dantraMSFT (changes requested):

iSazonov:
@dantraMSFT Please clarify - this file is excluded from compilation - is it acceptable to change it without tests? And will we use it later?

src/Microsoft.PowerShell.Commands.Diagnostics/PdhHelper.cs
jcarino: You're right, but adding to strColl only outside the while loop has some memory implications (has to alloc at least one extra string before Split). Can we just add another `strColl.Add(sb.ToString());` outside the while to catch whatever is left in sb?

daxian-dbw:
> this file is excluded from compilation 

@iSazonov Thanks for bringing it up :) You are right that we don't have a way to officially validate the changes. @TravisEz13, any thoughts?

For background: the perf counter cmdlets are disabled due to unsupported APIs. But the files are left in place so that
1. They can be brought back easily in future once the issue is resolved.
2. Users can enable building those cmdlets themselves and use the cmdlets privately.

iSazonov:
We should not make changes that are not tested.

I think we can postpone this change until the API is ported in the future. I've seen that CoreFS team still doesn't have a consensus how to do it.

TravisEz13:
Not only is it not tested.  The file is not being compiled.  I don't think we can take any changes to this file except possibly adding a comment to this effect and pointing to the issue to add the CmdLets back using supported APIs.  https://github.com/PowerShell/PowerShell/issues/4306

iSazonov:
I think we should close the PR until we compile this code again.

TravisEz13:
@jcarino The consensus is that we cannot take changes to this code at this time.

PR #5894, opened 2018-01-13
Create PowerShell Standard 6.0 reference assembly

JamesWTruher:
Tests are included to ensure that the assembly can be used to create a very simple cmdlet. I'll use this a the template to create the PowerShell Standard 5.1 assembly.

There is no actual implementation here, just the scaffolding of the classes, but this can be used to create assemblies (like cmdlet assemblies) which would then run against the actual code assembly. The expectation is that this source will be used to create the nupkg for distribution.

This also includes the source which was used for the PowerShell Standard 3.0 nuget package. The PowerShell Standard 3.0 assembly was initially created by jason, I'm including it in this PR to ensure that we keep track of it, since the compiled assembly is already available here: https://www.nuget.org/packages/PowerShellStandard.Library/3.0.0-preview-01

### DIFF
--- a/docs/SDK/PowerShellStandard/3.0/Reference/System.Management.Automation-lib.cs
+++ b/docs/SDK/PowerShellStandard/3.0/Reference/System.Management.Automation-lib.cs
@@ -1,0 +1,8621 @@
+/*
+
+First cut at removing apis that we might want to deprecate or not support
+
+CIM - cim support
+COM_APARTMENT_STATE - Missing from .Net, could come back
+COMPONENT_MODEL - Missing from .Net, could come back (not very important though)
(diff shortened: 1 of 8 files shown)


TravisEz13:
- normally reference assemblies are included in a `ref` folder under `src`.  see:  https://github.com/dotnet/corefx/tree/master/src/System.Collections/ref  Is there a reason you put this in docs?
- How are we going to build this?  
- if it's going to be in the docs folder, an `md` file describing the project and how to use it would be appropriate.

JamesWTruher:
The decision was to create a repo for the PowerShellStandard rather than add it to this repo, so I'm closing this

PR #5914, opened 2018-01-15
Updated install md bin path (not packaged documentation)

jbilinski:
Used explicit _6.0.0_ rather than _\<majorrelease\>_ since rest of documentation targets 6.0.0

## PR Checklist

- `NA` User facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - `NA` Issue filed - Issue link: GitHub documentation. not packaged documentation 
- `NA` [Change is not breaking](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes)
- `NA` [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
    - `NA` [Add `[feature]` if the change is significant or affects feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

### DIFF
--- a/docs/installation/windows.md
+++ b/docs/installation/windows.md
@@ -10,7 +10,7 @@ Once downloaded, double-click the installer and follow the prompts.
 There is a shortcut placed in the Start Menu upon installation.
 
 * By default the package is installed to `$env:ProgramFiles\PowerShell\`
-* You can launch PowerShell via the Start Menu or `$env:ProgramFiles\PowerShell\6.0.0\pwsh.exe`
+* You can launch PowerShell via the Start Menu or `pwsh` from a new command prompt.
 
 ### Prerequisites
 
--- a/docs/installation/windows.md
+++ b/docs/installation/windows.md
@@ -10,7 +10,7 @@ Once downloaded, double-click the installer and follow the prompts.
 There is a shortcut placed in the Start Menu upon installation.
 
 * By default the package is installed to `$env:ProgramFiles\PowerShell\`
-* You can launch PowerShell via the Start Menu or `$env:ProgramFiles\PowerShell\pwsh.exe`
+* You can launch PowerShell via the Start Menu or `$env:ProgramFiles\PowerShell\6.0.0\pwsh.exe`
 
 ### Prerequisites
 


docs/installation/windows.md
```diff
@@ -10,7 +10,7 @@ Once downloaded, double-click the installer and follow the prompts.
 There is a shortcut placed in the Start Menu upon installation.
 
 * By default the package is installed to `$env:ProgramFiles\PowerShell\`
-* You can launch PowerShell via the Start Menu or `$env:ProgramFiles\PowerShell\pwsh.exe`
+* You can launch PowerShell via the Start Menu or `$env:ProgramFiles\PowerShell\6.0.0\pwsh.exe`
```
bergmeister: Since pwsh is now on the [PATH](https://github.com/PowerShell/PowerShell/blob/master/assets/Product.wxs#L127), we should simplify this to something like `just type pwsh` because this works in any shell (cmd, PowerShell or just typing it into the address bar of explorer)

bergmeister (changes requested):

docs/installation/windows.md
jbilinski: agreed. unless installing from core cmd prompt.  Path in cmd is not updated until new cmd is invoked.

bergmeister: Well said, but I think this is standard behaviour and there are tricks to reload the PATH if one really wanted to. If someone installs it via cmd then I would say then they are clever enough to know that. I prefer simplicity in this case but the decision will be up to maintainers.

iSazonov:
@SteveL-MSFT Could you please review the doc change?

docs/installation/windows.md
SteveL-MSFT: I agree with just using `pwsh` and PATH not being updated is a corner case and pretty standard

iSazonov:
@jbilinski Could you please address comments?

docs/installation/windows.md
SteveL-MSFT: I would be ok if you added: "...Start Menu or from a new console window `pwsh`.

SteveL-MSFT (changes requested):

iSazonov:
@jbilinski Could you please continue?

bergmeister (approved):

iSazonov:
@SteveL-MSFT Could you please take another look?

SteveL-MSFT (approved):
LGTM

PR #5920, opened 2018-01-16
Support loading a custom powershell.config.json file via the command-line.

### LINKED ISSUE #5847: Provide command-line option for loading a custom powershell.config.json
This enhancement is primarily driven by testing use cases including the following:

1: On Linux, syslog imposes a measurable overhead impacting test runtimes. (@JamesWTruher: Please provide data)

2: For testing syslog and os_log there is a need to override the source name and output log to provide reliable verification.

One solution to both issues is to support launching an instance of PowerShell with a custom powershell.config.json file as a command-line option.  For test runs, the json could restrict log output to error and critical messages.  For syslog testing, the log file can be changed to ensure only the expected output is captured.  For syslog and os_log, the source name can be changed to enable filtering and verifying the expected log messages.

dantraMSFT:
## PR Summary

Fixes https://github.com/PowerShell/PowerShell/issues/5847

Support loading a custom powershell.config.json file via the command-line for use in testing.

This change supports replacing the the default powershell.config.json file loaded from the PSHome directory with a custom version.  

The primary use-cases for this command-line option are as follows:

1: Allow the CI system to disable settings that impact test run times; such as disabling syslog usage on Linux and MacOS

2: Support testing of syslog and os_log without interfering with normal PowerShell operations during test runs via launching an instance with custom log settings.

## PR Checklist

- [X] User facing [Documentation needed] - TBD
(https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
@@ -190,7 +190,7 @@ namespace Microsoft.PowerShell
             "file",
             "executionpolicy",
             "command",
-            "settingsFile",
+            "settingsfile",
             "help"
         };
 
@@ -718,14 +718,25 @@ namespace Microsoft.PowerShell
                             CommandLineParameterParserStrings.MissingSettingsFileArgument);
                         break;
                     }
-                    string configFile = args[i];
-                    if (!System.IO.File.Exists(configFile))
+                    string configFile = null;
+                    try
                     {
-                        WriteCommandLineError(
-                            CommandLineParameterParserStrings.SettingsFileNotExists);
+                        configFile = NormalizeFilePath(args[i]);
+                    }
+                    catch (Exception ex)
+                    {
+                        string error = string.Format(CultureInfo.CurrentCulture, CommandLineParameterParserStrings.InvalidSettingsFileArgument, args[i], ex.Message);
+                        WriteCommandLineError(error);
                         break;
                     }
-                    PowerShellConfig.Instance.SystemConfigFilePath = configFile;
+
+                    if (!System.IO.File.Exists(configFile))
+                    {
+                        string error = string.Format(CultureInfo.CurrentCulture, CommandLineParameterParserStrings.SettingsFileNotExists, configFile);
+                        WriteCommandLineError(error);
+                        break;
+                    }
+                    PowerShellConfig.Instance.SetSystemConfigFilePath(configFile);
                 }
 #if STAMODE
                 // explicit setting of the ApartmentState Not supported on NanoServer
@@ -867,6 +878,15 @@ namespace Microsoft.PowerShell
             executionPolicy = args[i];
         }
 
+        private static string NormalizeFilePath(string path)
+        {
+            // Normalize slashes
+            path = path.Replace(StringLiterals.AlternatePathSeparator,
+                                    StringLiterals.DefaultPathSeparator);
+
+            return Path.GetFullPath(path);
+        }
+
         private bool ParseFile(string[] args, ref int i, bool noexitSeen)
         {
             // Process file execution. We don't need to worry about checking -command
@@ -924,10 +944,7 @@ namespace Microsoft.PowerShell
                 string exceptionMessage = null;
                 try
                 {
-                    // Normalize slashes
-                    _file = args[i].Replace(StringLiterals.AlternatePathSeparator,
-                                            StringLiterals.DefaultPathSeparator);
-                    _file = Path.GetFullPath(_file);
+                    _file = NormalizeFilePath (args[i]);
                 }
                 catch (Exception e)
                 {
@@ -1004,11 +1021,11 @@ namespace Microsoft.PowerShell
                                 string argName = arg.Substring(0, offset);
                                 if (TryGetBoolValue(argValue, out bool boolValue))
                                 {
-                                        _collectedArgs.Add(new CommandParameter(argName, boolValue));
+                                    _collectedArgs.Add(new CommandParameter(argName, boolValue));
                                 }
                                 else
                                 {
-                                        _collectedArgs.Add(new CommandParameter(argName, argValue));
+                                    _collectedArgs.Add(new CommandParameter(argName, argValue));
                                 }
                             }
                         }
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
@@ -709,7 +709,7 @@ namespace Microsoft.PowerShell
                     }
                 }
 
-                else if (MatchSwitch(switchKey, "settingsFile", "settings") )
+                else if (MatchSwitch(switchKey, "settingsfile", "settings") )
                 {
                     ++i;
                     if (i >= args.Length)
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
@@ -882,7 +882,7 @@ namespace Microsoft.PowerShell
         {
             // Normalize slashes
             path = path.Replace(StringLiterals.AlternatePathSeparator,
-                                    StringLiterals.DefaultPathSeparator);
+                                StringLiterals.DefaultPathSeparator);
 
             return Path.GetFullPath(path);
         }
@@ -944,7 +944,7 @@ namespace Microsoft.PowerShell
                 string exceptionMessage = null;
                 try
                 {
-                    _file = NormalizeFilePath (args[i]);
+                    _file = NormalizeFilePath(args[i]);
                 }
                 catch (Exception e)
                 {
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
@@ -190,6 +190,7 @@ namespace Microsoft.PowerShell
             "file",
             "executionpolicy",
             "command",
+            "settingsFile",
             "help"
         };
 
@@ -706,6 +707,25 @@ namespace Microsoft.PowerShell
                     {
                         break;
                     }
+                }
+
+                else if (MatchSwitch(switchKey, "settingsFile", "settings") )
+                {
+                    ++i;
+                    if (i >= args.Length)
+                    {
+                        WriteCommandLineError(
+                            CommandLineParameterParserStrings.MissingSettingsFileArgument);
+                        break;
+                    }
+                    string configFile = args[i];
+                    if (!System.IO.File.Exists(configFile))
+                    {
+                        WriteCommandLineError(
+                            CommandLineParameterParserStrings.SettingsFileNotExists);
+                        break;
+                    }
+                    PowerShellConfig.Instance.SystemConfigFilePath = configFile;
                 }
 #if STAMODE
                 // explicit setting of the ApartmentState Not supported on NanoServer
--- a/src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
+++ b/src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
@@ -244,7 +244,10 @@ All parameters are case-insensitive.</value>
     Overrides the system-wide powershell.config.json settings file for the session.
     By default, system-wide settings are read from the powershell.config.json
     in the $PSHOME directory.
+
     Note that per-user settings take precedence over system-wide settings.
+    Also note that these settings are not used by the endpoint specified
+    by the -ConfigurationName argument.
 
     Example: pwsh -SettingsFile c:\myproject\powershell.config.json
 
--- a/src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
+++ b/src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
@@ -239,6 +239,15 @@ All parameters are case-insensitive.</value>
 
 -WindowStyle | -w
     Sets the window style to Normal, Minimized, Maximized or Hidden.
+
+-SettingsFile | -settings
+    Overrides the system-wide powershell.config.json settings file for the session.
+    By default, system-wide settings are read from the powershell.config.json
+    in the $PSHOME directory.
+    Note that per-user settings take precedence over system-wide settings.
+
+    Example: pwsh -SettingsFile c:\myproject\powershell.config.json
+
     </value>
   </data>
 </root>
--- a/src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
+++ b/src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
@@ -245,8 +245,7 @@ All parameters are case-insensitive.</value>
     By default, system-wide settings are read from the powershell.config.json
     in the $PSHOME directory.
 
-    Note that per-user settings take precedence over system-wide settings.
-    Also note that these settings are not used by the endpoint specified
+    Note that these settings are not used by the endpoint specified
     by the -ConfigurationName argument.
 
     Example: pwsh -SettingsFile c:\myproject\powershell.config.json
--- a/src/System.Management.Automation/engine/PSConfiguration.cs
+++ b/src/System.Management.Automation/engine/PSConfiguration.cs
@@ -62,15 +62,7 @@ namespace System.Management.Automation.Configuration
 
         private string GetConfigFilePath(ConfigScope scope)
         {
-            fileLock.EnterReadLock();
-            try
-            {
-                return (scope == ConfigScope.CurrentUser) ? perUserConfigFile : systemWideConfigFile;
-            }
-            finally
-            {
-                fileLock.ExitReadLock();
-            }
+            return (scope == ConfigScope.CurrentUser) ? perUserConfigFile : systemWideConfigFile;
         }
 
         /// <summary>
@@ -87,17 +79,9 @@ namespace System.Management.Automation.Configuration
             {
                 throw new FileNotFoundException(value);
             }
-            fileLock.EnterWriteLock();
-            try
-            {
-                FileInfo info = new FileInfo(value);
-                systemWideConfigFile = info.FullName;
-                systemWideConfigDirectory = info.Directory.FullName;
-            }
-            finally
-            {
-                fileLock.ExitWriteLock();
-            }
+            FileInfo info = new FileInfo(value);
+            systemWideConfigFile = info.FullName;
+            systemWideConfigDirectory = info.Directory.FullName;
         }
 
         /// <summary>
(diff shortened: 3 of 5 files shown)


src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -190,6 +190,7 @@ internal class CommandLineParameterParser
             "file",
             "executionpolicy",
             "command",
+            "settingsFile",
```
daxian-dbw: Maybe change the name to 'settingfile'? as none of the existing flag names here use an uppercase letter.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -707,6 +708,25 @@ private void ParseHelper(string[] args)
...
+                {
+                    ++i;
+                    if (i >= args.Length)
+                    {
+                        WriteCommandLineError(
+                            CommandLineParameterParserStrings.MissingSettingsFileArgument);
+                        break;
+                    }
+                    string configFile = args[i];
+                    if (!System.IO.File.Exists(configFile))
```
daxian-dbw: The validation of the file path should be similar to how `-File` does: normalize path separator and then get full path
See: https://github.com/dantraMSFT/PowerShell/blob/f587fd06d6046859ca1a9fe3470d5f9615dc5e01/src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs#L928

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
```diff
@@ -707,6 +708,25 @@ private void ParseHelper(string[] args)
...
+                    {
+                        WriteCommandLineError(
+                            CommandLineParameterParserStrings.SettingsFileNotExists);
```
daxian-dbw: The error message should also be similar to what's used for `-File`.
```
PS:3> powershell -File C:\Users\nonexist.ps1
The argument 'C:\Users\nonexist.ps1' to the -File parameter does not exist. Provide the path to an existing '.ps1' file as an argument to the -File parameter.
```

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
daxian-dbw: And, shall we check the extension of the file path to be `json`, like `-File` does?
```
PS:1> powershell -File F:\tmp\decode.json
Processing -File 'F:\tmp\decode.json' failed because the file does not have a '.ps1' extension. Specify a valid Windows PowerShell script file name, and then try again.
```

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
```diff
@@ -239,6 +239,15 @@ All parameters are case-insensitive.</value>
...
 -WindowStyle | -w
     Sets the window style to Normal, Minimized, Maximized or Hidden.
+
+-SettingsFile | -settings
+    Overrides the system-wide powershell.config.json settings file for the session.
+    By default, system-wide settings are read from the powershell.config.json
+    in the $PSHOME directory.
+    Note that per-user settings take precedence over system-wide settings.
+
+    Example: pwsh -SettingsFile c:\myproject\powershell.config.json
```
daxian-dbw: The scenario I think of is `pwsh -settingsfile blah.json -configurationname another-ps-instance-config-name`. In this case, shall the user expect the `blah.json` would be applied to the powershell instance that the `another-ps-instance-config-name` points to?

Maybe you can document the behavior in that case in the help content.

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -28,6 +28,10 @@ internal sealed class PowerShellConfig
         internal static PowerShellConfig Instance => s_instance;
 
         private string psHomeConfigDirectory;
```
daxian-dbw: I suggest we remove `psHomeConfigDirectory` and `appDataConfigDirectory`. Instead, add `psHomeConfigFile` and `appDataConfigFile` and initialize them in `private PowerShellConfig()` (e.g. `Path.Combine(Utils.DefaultPowerShellAppBase, configFileName)`), so we don't need to call `Path.Combine` every time try retrieving a setting value.

In that case, we don't need `psCustomHomeConfigFilePath`, but just set `psHomeConfigFile` in the `SystemConfigFilePath` property. In this way, the method `GetConfigFilePath` will be more simplified.

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -50,6 +54,65 @@ private PowerShellConfig()
...
+                try
+                {
+                    psCustomHomeConfigFilePath = value;
+                }
+                finally
+                {
+                    fileLock.ExitWriteLock();
+                }
+            }
+        }
```
daxian-dbw: I don't think the lock is necessary. This property should only be used during command parsing when starting the host. No Ruspace is created at that point and thus there won't be multiple threads access this property.

IMHO, this property should be setter only, and maybe we should make it only able to be set once.

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -133,12 +174,10 @@ internal void SetExecutionPolicy(ConfigScope scope, string shellId, string execu
                 // CreateDirectory will succeed if the directory already exists
                 // so there is no reason to check Directory.Exists().
                 Directory.CreateDirectory(appDataConfigDirectory);
```
daxian-dbw: If we decided to get rid of `appDataConfigDirectory`, then it needs to be changed to `Path.GetDirectoryName(appDataConfigFile)` here.

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -362,6 +407,21 @@ internal PSKeyword GetLogKeywords()
...
+        /// Update a value in the configuration file.
+        /// </summary>
+        /// <typeparam name="T">The type of the value</typeparam>
+        /// <param name="scope">The ConfigScope of the configuration file to update.</param>
+        /// <param name="key">The string key of the value.</param>
+        /// <param name="value">The value to set.</param>
+        /// <param name="addValue">Whether the key-value pair should be added to or removed from the file</param>
+        private void UpdateValueInFile<T>(ConfigScope scope, string key, T value, bool addValue)
+        {
+
```
daxian-dbw: Extra new line

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -331,7 +360,23 @@ internal PSKeyword GetLogKeywords()
...
+        /// <param name="scope">The ConfigScope of the configuration file to update.</param>
+        /// <param name="key">The string key of the value.</param>
+        /// <param name="defaultValue">The default value to return if the key is not present.</param>
+        /// <param name="readImpl"></param>
+        private T ReadValueFromFile<T>(ConfigScope scope, string key, T defaultValue = default(T),
+                                       Func<JToken, JsonSerializer, T, T> readImpl = null)
+        {
+
+            string fileName = GetConfigFilePath(scope);
+            return ReadValueFromFile<T>(fileName, key, defaultValue, readImpl);
```
daxian-dbw: How about just changing the old `ReadValueFromFile<T>(string fileName ...` to `ReadValueFromFile<T>(ConfigScope scope ...`? Same with the other file operation help methods below. It feels cleaner that way.

daxian-dbw (changes requested):

daxian-dbw:
@SteveL-MSFT Maybe the committee needs to review the new flag name added to `pwsh`? It's now `-SettingsFile`.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
dantraMSFT: @daxian-dbw: I assume you are referring to the camel casing, right or are you also suggesting the singular form.  I agree with removing the camel casing but I don't think using the singular makes grammatical sense for this option.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
dantraMSFT: I'm not aware of any requirement for json files to have the extension .json.  Is this really something we want to enforce?

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
dantraMSFT: I'll document it.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
daxian-dbw: I was referring to the camel casing.

src/System.Management.Automation/engine/PSConfiguration.cs
dantraMSFT: Agreed. In the process, I'll also change the field names. the current names are rather obfuscated considering their usage.

src/System.Management.Automation/engine/PSConfiguration.cs
dantraMSFT: I agree with just having a setter - will change it to a method.
The locks are used around writes to the configuration files. For correctness, they should be around the access to the file paths as well.

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
dantraMSFT: Will do

src/System.Management.Automation/engine/PSConfiguration.cs
dantraMSFT: Noted.

src/System.Management.Automation/engine/PSConfiguration.cs
dantraMSFT: @daxian-dbw Are you suggesting merging versus wrapping?

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
dantraMSFT: fixed

src/Microsoft.PowerShell.ConsoleHost/host/msh/CommandLineParameterParser.cs
dantraMSFT: Validation fixed

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
dantraMSFT: Fixed

src/System.Management.Automation/engine/PSConfiguration.cs
dantraMSFT: Fixed

src/System.Management.Automation/engine/PSConfiguration.cs
dantraMSFT: Removed getter and changed to Set method

src/System.Management.Automation/engine/PSConfiguration.cs
dantraMSFT: Fixed

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -108,37 +138,24 @@ internal string GetExecutionPolicy(ConfigScope scope, string shellId)
...
             // Defaults to system wide.
             if (ConfigScope.CurrentUser == scope)
             {
                 // Exceptions are not caught so that they will propagate to the
                 // host for display to the user.
                 // CreateDirectory will succeed if the directory already exists
                 // so there is no reason to check Directory.Exists().
-                Directory.CreateDirectory(appDataConfigDirectory);
-                scopeDirectory = appDataConfigDirectory;
+                Directory.CreateDirectory(perUserConfigFile.Directory.FullName);
```
daxian-dbw: I don't think it's necessary to use `FileInfo` type for `systemWideConfigFile` and `perUserConfigFile` to just use the `Directory` property here. Seems too heavy to me. I suggest we just add a private field `perUserConfigDirectory` to hold the directory path, and make `systemWideConfigFile` and `perUserConfigFile` strings.

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -40,14 +46,49 @@ internal sealed class PowerShellConfig
...
             // Note: This directory may or may not exist depending upon the
             // execution scenario. Writes will attempt to create the directory
             // if it does not already exist.
-            appDataConfigDirectory = Utils.GetUserConfigurationDirectory();
+            perUserConfigFile = new FileInfo(Path.Combine(Utils.GetUserConfigurationDirectory(), configFileName));
+        }
+
+        private string GetConfigFilePath(ConfigScope scope)
+        {
+            fileLock.EnterReadLock();
```
daxian-dbw: I don't think the lock is needed here. `SetSystemConfigFilePath` should only be used at the parameter parsing phase of `pwsh`, and around that time no accessing would be made to the configuration file.

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -40,14 +46,49 @@ internal sealed class PowerShellConfig
...
+                return (scope == ConfigScope.CurrentUser) ? perUserConfigFile.FullName : systemWideConfigFile.FullName;
+            }
+            finally
+            {
+                fileLock.ExitReadLock();
+            }
+        }
+
+        /// <summary>
+        /// Sets the system wide configuration file path
```
daxian-dbw: Can you please explicitly point out that this method should only be used when parsing the flag `-SettingsPath` for `pwsh` in `CommandLineParameterParser.cs`?

src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -40,14 +46,49 @@ internal sealed class PowerShellConfig
...
+        /// </summary>
+        /// <param name="value">A fully qualified path to the system wide configuration file.</param>
+        /// <exception cref="FileNotFoundException"><paramref name="value"/> is a null reference or the associated file does not exist.</exception>
+        internal void SetSystemConfigFilePath(string value)
+        {
+            if (!string.IsNullOrEmpty(value) && !File.Exists(value))
+            {
+                throw new FileNotFoundException(value);
+            }
+            fileLock.EnterWriteLock();
```
daxian-dbw: Same comment here.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/PSConfiguration.cs
dantraMSFT: Either we rewrite the logic to enforce that it can only be called once or we make it thread safe with the reset of the usage.  I'm inclined to keep it as it is.

src/System.Management.Automation/engine/PSConfiguration.cs
dantraMSFT: Fixed

src/System.Management.Automation/engine/PSConfiguration.cs
dantraMSFT: Won't fix

src/System.Management.Automation/engine/PSConfiguration.cs
dantraMSFT: Won't fix

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
```diff
@@ -239,6 +239,18 @@ All parameters are case-insensitive.</value>
 
 -WindowStyle | -w
     Sets the window style to Normal, Minimized, Maximized or Hidden.
+
+-SettingsFile | -settings
+    Overrides the system-wide powershell.config.json settings file for the session.
+    By default, system-wide settings are read from the powershell.config.json
+    in the $PSHOME directory.
+
+    Note that per-user settings take precedence over system-wide settings.
```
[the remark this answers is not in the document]
dantraMSFT: Fixed: I removed the statement.

src/System.Management.Automation/engine/PSConfiguration.cs
daxian-dbw: > 5.5 Atomicity of variable references
Reads and writes of the following data types are atomic: bool, char, byte, sbyte, short, ushort, uint, int, float, and reference types. In addition, reads and writes of enum types with an underlying type in the previous list are also atomic. Reads and writes of other types, including long, ulong, double, and decimal, as well as user-defined types, are not guaranteed to be atomic. Aside from the library functions designed for that purpose, there is no guarantee of atomic read-modify-write, such as in the case of increment or decrement.

Read and write a reference type is guaranteed to be atomic in the C# language spec. In this case, we are updating a string in `SetSystemConfigFilePath` which is a reference type, so the read and write to the field `systemWideConfigFile` is atomic, so the lock around the write and read operation doesn't really buy us anything.

dantraMSFT: @daxian-dbw: I fully understand the language's atomic guarantees but that is not why I added the lock.
However, I will remove it

daxian-dbw (approved):
Changes look good to me.
Now we are waiting for the committee to review the new flag name for pwsh.

daxian-dbw:
AppVeyor build failed due to the web listener startup. Add the `[Feature]` tag to the last commit and restart the test run.

daxian-dbw:
> [FETURE] Merge branch 'dantra/issue5312' of https://github.com/dantr

@dantraMSFT There is a typo in the feature tag in the last commit, and thus the full test run was not triggered. Can you please update the tag of the last commit?

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and is ok with `-settingsfile`

daxian-dbw:
@dantraMSFT The last commit doesn't have `[Feature]` tag in its message and thus it didn't trigger a full test run. If you don't mind, I will clean up your merge commits and set [feature] tag to the last commit.

daxian-dbw:
I have cleaned up the commits and the last commit has `[feature]` tag now. Will wait on the CI runs.

TravisEz13:
Can someone file an issue?  The latest message was detected to be `Return better error for 'pwsh -WindowStyle' on unsupported platforms` due to merge.

This can be solved by using `TRAVIS_PULL_REQUEST_SHA`. https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables to get the commit message for a PR instead of the latest.

dantraMSFT:
@daxian-dbw created https://github.com/PowerShell/PowerShell/issues/6022

daxian-dbw:
@TravisEz13 I submitted #6024 to use retrieve the commit message with `TRAVIS_PULL_REQUEST_SHA`.

Issue #4268, opened 2017-07-17
"call depth overflow" exception in Powershell scripts on Mac


mpeneva:
Some of of my PowerShell scripts that I run on Mac failed with the following exception:
"The script failed due to call depth overflow."
After some investigation, I came to the conclusion that the allow call depth in PowerShell v6.0 alpha,
is smaller than the defined in PowerShell v5.0 (which I find is 100).
Could you change or increase in PowerShell for Unix that fixed call depth to be at least 1000?

iSazonov:
Can you reproduce with latest version - Beta.4?
If so please add repo steps and more description.

mpeneva:
Hi and thanks for your response. I've updated the PS to version 4.0 beta, however it occurred a new issue, so we started looking at it with one of my colleagues, so once we're ready with that, I would try the above described issue with the beta version of PS.

iSazonov:
Could you please add simple repo steps?

mpeneva:
Sure, I would , but currently there is a 'local' problem, that's not connected at all with the PS version.

it-praktyk:
I've observed the same issue what you can easily reproduce.

Please clone the Pester branch [Dev4-1_PSCore_compatibility](https://github.com/pester/Pester/tree/Dev4-1_PSCore_compatibility), check out to the commit [89344732785b862c4c0fcc72050290df6fce9e5f](https://github.com/pester/Pester/tree/89344732785b862c4c0fcc72050290df6fce9e5f) and run tests from the file [Functions/Assertions/Be.Tests.ps1](https://github.com/pester/Pester/blob/89344732785b862c4c0fcc72050290df6fce9e5f/Functions/Assertions/Be.Tests.ps1).

The test result

```
<OUTPUT_PARTIALL_OMMITED>

    [-] throws exception when self-imposed recursion limit is reached 592ms
      Expected: the expression to throw an exception with message {recursion depth limit}, an exception was raised, message was {The script failed due to call depth overflow.}
          from <FOLDER_PATH>/Pester/Functions/Assertions/Be.Tests.ps1:99 char:21
          +             { $a1 | Should be $a2 } | Should throw 'recursion depth l ...
          +                     ~~~~~~~~~~~~~
      99:             { $a1 | Should be $a2 } | Should throw 'recursion depth limit'
      at Invoke-LegacyAssertion, <FOLDER_PATH>/Pester/Functions/Assertions/Should.ps1: line 190
[...]

Issue #5809, opened 2018-01-06
Improve the configuration APIs in PSConfiguration.cs


daxian-dbw:
The tasks that I can think of are:
1. maybe replace the singleton with static methods
2. schema check
3. synchronize updating the configuration file from multiple processes.
   - we currently have a read/write lock to protect updating/reading the config file. The lock doesn't help synchronize writing access from multiple processes.
4. holding the config file in one process (e.g. `write` access or `none` share) will cause a new pwsh instance to crash at startup because it cannot access the config file and we don't deal with the `IOException`.
   - we could return default value when the config file is not accessible, but need to consider to not allow bypassing security settings because of that.
      - one option: have two sets of default values for security settings. One set is used when the config file doesn't exist or the security setting is not defined in the file; the other set is used when the file exists but not accessible, to enforce all security features that are configurable in the config file.

#### Repro step for 4:
First, run the following code to hold the config file of the current pwsh
```
$path = "path-to-powershell.config.json"
$mode = "Open"
$access = "Read"
$share = "None"

$file = [System.IO.File]::Open($path, $mode, $access, $share)
```
Then, start a new instance of the pwsh, and it will crash.

iSazonov:
We could make public API to enable developers to work consistently with config files.

iSazonov:
It seems currently we read config option step-by-step by demand. I think re-read all config file(s) in one step and cache it is more optimal.

iSazonov:
Suggestions:
- use Windows PowerShell GPO by default to not allow users ignore security and Enterprise policies for PowerShell
- add new GPO branch for PowerShell Core to allow Enterprises configure Windows PowerShell and PowerShell Core separately.
- use GPO policies (Software\Policies\Microsoft\Windows\PowerShell) and GPO config options (Software\Microsoft\Windows\PowerShell)
- we want #5804 RegisterGPNotification we should do the same for config files too.

iSazonov:
Configuration priority:

Windows | Unix
-| -
GPO Computer Policy for PowerShell Core | File
GPO Computer Policy for Windows PowerShell | -
File Computer-Wide Version | File Computer-Wide Version
GPO User Policy | ?
File Local Version | File Local Version
File User-Wide (AppData) | File User-Wide
GPO Computer Config | ?
GPO User Config | ?

TravisEz13:
`GPO Computer Policy for Windows PowerShell` is not considered when in PS Core.

daxian-dbw:
[...]

PR #5927, opened 2018-01-17
Set local pipeline thread stack size

iSazonov:
## PR Summary
Related #4268.

1. Set local pipeline thread stack size to 10MB by default.
2. Read the `PipelineMaxStackSizeMB` configuration value from config.

Thread stack OS system default size is:
- on Windows - 1MB
- on MacOs - 0.5MB
- on Linux - 8MB

We should use the same stack size for all platforms to get predictable behavior.
This value should match the Windows PowerShell value for backward compatibility (I expect that .Net Core and Windows PowerShell consume the stack as well as .Net Framework and Windows PowerShell).
Perhaps default 10MB may be too large if we use many local pipelines for paralleling. We can reconsider this later.

## PR Checklist


src/System.Management.Automation/engine/PSConfiguration.cs
```diff
@@ -548,6 +548,7 @@ internal sealed class PowerShellPolicies
         public ProtectedEventLogging ProtectedEventLogging { get; set; }
         public Transcription Transcription { get; set; }
         public UpdatableHelp UpdatableHelp { get; set; }
+        public UpdatableHelp PipelineMaxStackSizeMB { get; set; }
```
daxian-dbw: The `PipelineMaxStackSizeMB` setting is not a policy in windows powershell, and thus it shouldn't be added to `PowerShellPolicies`. It should be a setting like `psmodulepath`, or `executionpolicy` or `ConsolePrompting`.

THe `PipelineMaxStackSizeMB` was stored at `SOFTWARE\Microsoft\PowerShell\1\ShellIds` in windows powershell, not at `SOFTWARE\Policies\...`

src/System.Management.Automation/engine/hostifaces/LocalPipeline.cs
```diff
@@ -247,54 +248,35 @@ private void SetupInvokeThread(Thread invokeThread, bool changeName)
...
-            {
-                int i = ReadRegistryInt("PipelineMaxStackSizeMB", 10);
-                if (i < 10)
-                    i = 10; // minimum 10MB
-                else if (i > 100)
-                    i = 100; // maximum 100MB
-                return i * 1000000;
-            }
-        }
+        internal static int StackSize => GetPipelineMaxStackSizeOption(10);
```
daxian-dbw: If we don't cache the stack size, then this static property seems not necessary as it only adds one more redirection to call `GetPipelineMaxStackSizeOption`.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/PSConfiguration.cs
iSazonov: I believe [previously](https://github.com/PowerShell/PowerShell/issues/5809#issuecomment-356059863) we agreed that PowerShell Core is a separate product and it should have its policies and not use policies and configs of Windows PowerShell.
Why don't we have consistency between the settings in the file and the registry if we the consistency by default on Unix? I wouldn't be surprised if MSFT decides to port Registry API to make it easier to port applications. (I would take this into account).
Also if the settings are in different places, it is difficult to find it and makes it very difficult to configure.
Another thought is that we must put (later?) config options and  GPO policies in different registry keys.

I think we should develop this my [comment](https://github.com/PowerShell/PowerShell/issues/5809#issuecomment-355843863) about priority - maybe open new tracking Issue?
Also we could open new tracking Issue (?) to enumerate config options and to solve where put different options - in policy key or/and in config key.

Today I suggest:
1. Have `PipelineMaxStackSizeMB` (and most other options) in registry and file.
2. Put `PipelineMaxStackSizeMB` in "Software\Policies\Microsoft\PowerShellCore". Later we can move/duplicate it in "Software\Microsoft\PowerShellCore" - today we haven't code to process the registry branches in priority.

daxian-dbw: I didn't mean `PipelineMaxStackSizeMB` should be put in Registry on Windows. I think this setting should be in the config file only on both Windows and Unix.

As far as I understand, only the GPO related settings will live in Registry and `powershell.config.json` on Windows for PowerShell Core (registry setting takes precedence). Non-GPO settings that used to be stored in Registry in Windows PowerShell now should live in the config file only in PowerShell Core, such as `executionpolicy`, `ConsolePrompting` and many others. You can take a look at `PSConfiguration.cs` to find more of them.

The type `PowerShellPolicies` only represents the GPO settings. `PipelineMaxStackSizeMB` is not a policy setting but a normal powershell setting, so it shouldn't be put in `PowerShellPolicies`.

iSazonov: @daxian-dbw I wanted to create Issue but the text was too big and I made a document #5947

iSazonov:
@daxian-dbw What is "policy" in the context? 

`Software\Policies\Microsoft\PowerShellCore` - I believe it is "policy"?

`Software\Microsoft\PowerShell\1\ShellIds` or 
`Software\PowerShellCore`  - I believe it is "regular settings"? Do you say that we shouldn't use Registry for this settings and keep its only in a config file?

daxian-dbw:
> What is "policy" in the context?

It's short for "Group Police Settings". Those are the settings that are originally under "Software\Policies". Each of the setting has a corresponding GPO in "Local Group Policy Editor" in Control Panel. The new PowerShell Core ones will also be available in the editor. We will publish a `.admx` file for them.

The settings that are not originally under "Software\Policies" are regular powershell settings. As far as I understand, we don't use Registry for those settings but only refer to the config file.

iSazonov:
Thanks for clarify!
>As far as I understand, we don't use Registry for those settings but only refer to the config file.

Is this PowerShell MSFT team solution? And why? 

Registry is common place for Windows applications to store settings:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms724871(v=vs.85).aspx
https://msdn.microsoft.com/en-us/library/ms973902.aspx

This is critical for enterprises. By using GPO/GPP, they can centrally manage (distribute polices and settings too) applications on workstations and server farms.
For example, if they receive feedback from employees that their scripts fail because of a stack overflow, the enterprise administrator can easily propagate the parameter value as a default that does not overlap local settings at the application or user level.

SteveL-MSFT:
For PSCore6, we want to have configuration not in the registry.  This makes it easier to deploy via zip, for example, as well as support side-by-side more simply.  I do agree that for an enterprise, we should support GPO.  In which case, there is no issue with having both configuration and policy.

iSazonov:
>This makes it easier to deploy via zip, for example, as well as support side-by-side more simply.

The ability to have settings in the registry does not conflict with zip and side-by-side. (We could even enhance our API to support applications by type and name)

@SteveL-MSFT What your conclusion about #5947? - I'd move it in RFC repo but it's frozen. I would have moved RFC repo here with label like `RFC-*` - so we get more feedback.

iSazonov:
I created https://github.com/PowerShell/PowerShell-RFC/pull/111.

Should we wait the RFC?

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and don't believe that the stack size should even be a configurable setting until there is a clear need to make it settable.  Recommendation is to not even have it settable despite what Windows PowerShell did.

iSazonov:
@SteveL-MSFT Stack size default is different on different platforms - should we set the same default _internally_ to get predictable behavior on all platforms? I think it make sense.

SteveL-MSFT:
@iSazonov yes, @PowerShell/powershell-committee agrees to changing the default on all systems.  We thought the only point that needed discussion from the committee was configuration.

daxian-dbw:
A new PR was submitted based on the committee's review: #6224 
I will close this PR.

Issue #5140, opened 2017-10-17
Web Cmdlet Error Syntax Re-style


markekraus:
Currently the errors in the Web Cmdlets are a bit tough to read. specifically the parameters are just unquoted strings. Theys hould be made more readable by using the format `'-Parameter'`.

iSazonov:
See [comment](https://github.com/PowerShell/PowerShell/pull/5972#discussion_r162857125)
>I see we use the file pattern but the messages can be shorter - "Conflicting parameters are specified: Body and Form."

PR #5972, opened 2018-01-21
Add Simplified multipart/form-data Support to Web Cmdlets Through -Form Parameter

### LINKED ISSUE #2112: Simplify usage of Multipart Uploads with WebRequestPSCmdlet

markekraus:
## PR Summary

* Closes #2112 🎉 
* Adds `-Form` Parameter to `Invoke-WebRequest` and `Invoke-RestMethod`
* `-Form` Accepts any `IDictionary`.
* Keys are used as `multipart/form-data` field names (`PSObject` unwrapped and `ConvertTo<String>(Object)`)
* `FileInfo` values are added as `StreamContent` with `application/octet-stream` content type and the `FileInfo.Name` as the file name.
* Strings are treated as `StringContent`
* Singe values are converted to string with `ConvertTo<String>(Object)` and treated as `StringContent`
* Top level collections are enumerated and converted as above. Nested collections are treated as a single value and converted accordingly.
* `-Form` is mutually exclusive with `-Body` and `-InFile`
* Per committee decision, `-Form` makes no assumptions about the HTTP method used. It can theoretically be used with any method. User will need to manually supply `-Method POST` to post the form.
* `-ContentType` and content related headers supplied to `-Headers` will be ignored/cleared as `MultipartFormDataContent` requires control of these headers.

## PR Checklist

- [X] User facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - [X] Issue filed - Issue link: PowerShell/PowerShell-Docs#2077


src/TypeCatalogGen/powershell_win7-x64.inc
```diff
@@ -0,0 +1,165 @@
+C:\Users\MarkLocal\AppData\Local\Microsoft\dotnet\sdk\NuGetFallbackFolder\microsoft.applicationinsights\2.4.0\lib\netstandard1.3\Microsoft.ApplicationInsights.dll;
```
iSazonov: Please remove the change - the file is auto generated.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1075,8 +1095,21 @@ internal virtual void FillRequestStream(HttpRequestMessage request)
                 }
             }
 
+            if (Form != null)
```
iSazonov: Could you please use the same pattern in whole file (null on first place)? I see only 2-3 places where we use old pattern - maybe fix it too.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1396,6 +1429,109 @@ protected override void StopProcessing()
...
 
+        
+        /// <summary>
+        /// Adds content to a <see cref="MultipartFormDataContent" />. Object type detection is used to determine if the value is String, File, or Collection.
+        /// </summary>
+        /// <param name="fieldName">The Field Name to use.</param>
+        /// <param name="fieldValue">The Field Value to use.</param>
+        /// <param name="formData">The <see cref="MultipartFormDataContent" />> to update.</param>
+        /// <param name="enumerate">If true, collection types in <paramref name="fieldValue" /> will be enumerated. If false, collections will be treated as single value.</param>
+        internal void AddMultipartContent (object fieldName, object fieldValue, MultipartFormDataContent formData, bool enumerate)
```
iSazonov: Extra space after method name - we never use the pattern. Below too.

Shouldn't the method be private? Helper methods below too.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1396,6 +1429,109 @@ protected override void StopProcessing()
...
+        {
+            if (null == formData)
+            {
+                formData = new MultipartFormDataContent();
```
iSazonov: Is `formData` a return (result) data? Shouldn't it be `out` parameter? I am confused that we use the constructor inside the method and outside.

src/Microsoft.PowerShell.Commands.Utility/resources/WebCmdletStrings.resx
```diff
@@ -138,6 +138,12 @@
   <data name="BodyConflict" xml:space="preserve">
     <value>The cmdlet cannot run because the following conflicting parameters are specified: Body and InFile. Specify either Body or Infile, then retry.  </value>
   </data>
+  <data name="BodyFormConflict" xml:space="preserve">
+    <value>The cmdlet cannot run because the following conflicting parameters are specified: Body and Form. Specify either Body or Form, then retry.  </value>
```
iSazonov: I see we use the file pattern but the messages can be shorter - "Conflicting parameters are specified: Body and Form."

Maybe fix this in the PR or new one?

iSazonov (commented):
Great!

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
markekraus: I'm not sure. I was following suit with the existing helper functions. I will try setting them private and see if anything breaks.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
markekraus: I guess this should throw? The thing is, I didn't want it to throw and out requires assignment. I don't care if someone supplies null to formData but I didn't want the code to fail if they did. maybe I should just return on or an assert?

src/Microsoft.PowerShell.Commands.Utility/resources/WebCmdletStrings.resx
markekraus: There is an existing issue #5140 can you comment there?

src/TypeCatalogGen/powershell_win7-x64.inc
markekraus: can we get this file added to the ignore? It is super annoying to need to watch out for this file in my commits all the time.

markekraus:
I just realized my tests were missing from the commit. They will be included with the PR feedback changes.

src/TypeCatalogGen/powershell_win7-x64.inc
iSazonov: https://github.com/PowerShell/PowerShell/blob/master/src/TypeCatalogGen/.gitignore

src/Microsoft.PowerShell.Commands.Utility/resources/WebCmdletStrings.resx
iSazonov: Done.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
iSazonov: We used `internal` because we had many files.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
iSazonov: This is only a helper function - I think it is enough to use `Assert()`.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
markekraus: fixed. Maybe comment on #5610 for the null ordering in the rest of the file (and files...)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
markekraus: fixed

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
markekraus: fixed. we are throwing `ArgumentNullException` in other methods, I changed it to the same.

src/TypeCatalogGen/powershell_win7-x64.inc
markekraus: fixed

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
iSazonov: Done.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1098,6 +1098,18 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
     }
 
     Context "Multipart/form-data Tests" {
+        BeforeAll {
+            $file1Name = "testfile1.txt"
+            $file1Path = Join-Path $testdrive $file1Name
+            $file1Contents = "Test123"
+            $file1Contents | Set-Content $file1Path
```
iSazonov: I think we should add ` -Force` to `Set-Content` to exclude random confirmation.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1130,6 +1144,108 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
...
+
+        It "Verifies Invoke-WebRequest -Form supports a collection of string values" {
+            $form = @{TestStrings = "TestValue", "TestValue2"}
+            $uri = Get-WebListenerUrl -Test 'Multipart'
+            $response = Invoke-WebRequest -Uri $uri -Form $form -Method 'POST'
+            $result = $response.Content | ConvertFrom-Json
+
+            $result.Headers.'Content-Type' | Should Match 'multipart/form-data'
+            $result.Items.TestStrings.Count | Should Be 2
+            $result.Items.TestStrings[0] | Should Be 'TestValue'
```
iSazonov: I think `BeExactly` is better for strings in the tests.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1130,6 +1144,108 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
...
+            $response = Invoke-WebRequest -Uri $uri -Form $form -Method 'POST'
+            $result = $response.Content | ConvertFrom-Json
+
+            $result.Headers.'Content-Type' | Should Match 'multipart/form-data'
+            $result.Files.Count | Should Be 1
+
+            $result.Files[0].Name | Should Be "TestFile"
+            $result.Files[0].FileName | Should Be $file1Name
+            $result.Files[0].ContentType | Should Be 'application/octet-stream'
+            $result.Files[0].Content | Should Match $file1Contents
```
iSazonov: If we want get the same content I think we should use `BeExactly`.

iSazonov (commented):
@markekraus Please clarify: I see we use ANSI chars in the tests - make sense test other charsets/encodings?

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: We have to use match here because kestrel doesn't act the same on all platforms.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: I will switch these for everything except the file contents which need to remain as match.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: Please clarify, if I send binary file (zip) I get it corrupted? Can we then trust these tests?

markekraus: It's not on the sending PS side. It has something to do with the result JSON serialization. Sometimes the newline is trimmed and sometimes not. It's a rendering issue. We can improve the trust of these tests with file hashes, return an array of bytes, or a base64.

iSazonov: As far as I understand ' application/octet-stream ' is a binary stream - then we have a serious bug in Kestrel 😕

markekraus: You misunderstand. This is when the stream is converted to test and again to JSON. Kestrels gets it just fine, but when the string is converted to JSON the behavior is a bit unstable between platforms. like I said, we should use one of the three methods to improve trust in the test. its not trivial work though and it will affect other tests. if we want to improve the trust, we should have a separate issue to include one of the three options.

iSazonov: > when the string is converted to JSON the behavior is a bit unstable between platforms.
For full my understanding  - where is the wrong code?

markekraus: It is somewhere between Kestrel and ASP.NET JSON result. no clue.. beyond my skill to track down. I did some basic testing where I had the server save the uploaded file to disk and manually ran hashes on the source and upload file to ensure they match. no problem there.. so something happens between the filestream being converted to text and then to a JSON result and then sent as a HttpResponse. I can see it in tcpdump that it's coming out of the server wrong.. so it's not a rendering issue on the PowerShell side either.  

That stack is too deep for my skill level to look into. I can't reliably reproduce it either. It's why the other existing multipart tests are doing match on the the rendered file content. But I'm not too concerned.. rendering binary to text is always a bit wonky... a better test from the start would be to have provide SHA256 hash for the byte stream. that should match the sha256 hash of the FileStream on the source side. 

I don't want to do it in this PR because the change is not trivial and will affect the other tests. I'd rather keep with the `match` here and address the trust improvements in another PR. (the stream is forward only, so if I want to provide text contents as well as hash I have to introduce a buffer, or I have to switch to just the hash.. either way has an impact on the existing tests.)

Someday when I have time to become an expert at debugging .NET Core... I might be able to actually find the root cause of such things. 😆

iSazonov: Many thanks! I think we should fix the issue in ??? - your description is great for openning Issue (in CoreFX repo if Kestrel use CoreFX System.Json.
After that we can use safely `BeExactly` in this tests. I believe we can't complicate our tests if we get the fix.
Also we already have meta Issue for tracking web cmdlets - we should add this in the Issue.

markekraus:
Revered the content-type tests to `Match`. We can't know the multipart/form-data boundary before the request is sent so we cannot do `BeExactly` assertion.

markekraus:
Remaining test failures are form #5973 being fixed in #5980.

iSazonov:
> We can't know the multipart/form-data boundary before the request is sent so we cannot do BeExactly assertion.

I want see more information about it and I think we should add appropriate comment in tests.

iSazonov:
@markekraus Thanks! I expected only `multipart/form-data`. The comment will be useful in tests.

markekraus:
@iSazonov comments added.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
SteveL-MSFT: Perhaps we should add a comment here

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -1590,6 +1623,108 @@ internal void ParseLinkHeader(HttpResponseMessage response, System.Uri requestUr
...
+        /// <summary>
+        /// Adds content to a <see cref="MultipartFormDataContent" />. Object type detection is used to determine if the value is String, File, or Collection.
+        /// </summary>
+        /// <param name="fieldName">The Field Name to use.</param>
+        /// <param name="fieldValue">The Field Value to use.</param>
+        /// <param name="formData">The <see cref="MultipartFormDataContent" />> to update.</param>
+        /// <param name="enumerate">If true, collection types in <paramref name="fieldValue" /> will be enumerated. If false, collections will be treated as single value.</param>
+        private void AddMultipartContent(object fieldName, object fieldValue, MultipartFormDataContent formData, bool enumerate)
+        {
+            if (null == formData)
```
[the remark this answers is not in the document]
iSazonov: The pattern is common in the file.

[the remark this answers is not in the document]
rkeithhill: FYI, this type of null check (null on the LHS) is a C++ idiom that does not apply to C# because if/while/do conditions in C# must evaluate to a Boolean.  Therefore a mistaken assignment e.g. `if (formData = null)` would fail to compile because the type of formData can't be a Boolean (not if you're assigning null to it).  In fact, the only time you need to put the literal on the LHS is when the variable type is a Boolean e.g. `if (false == aBooleanVariable)`.  

My team's internal coding guidelines suggest not using `==` or `!=` in the Boolean variable case but use `if (!aBooleanVariable)` or `if (aBooleanVariable)` instead.  Consequently, we never put the literal value on the LHS in C# code.  And IMO this approach reads easier e.g. `if (formData == null)`.

zhouyoufu:
~~

~~

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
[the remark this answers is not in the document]
markekraus: @rkeithhill  The problem is more that this file uses inconsistent forms and I forget which one is being used in which context. In some parts of the file null is left in others null is right. As long as there isn't some programmatic consideration (I suspected there wasn't), I don't have an opinion either way. `null == foo` and `foo == null` read exactly the same to me with out one being easier than the other. I have a habit of "literals go left" because it is a common pattern in many languages for various reasons.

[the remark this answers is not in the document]
markekraus: @SteveL-MSFT I'm not entire certain of the future of these new methods. I get the feeling they may end up moving to a helper class as static methods to be consumed in more places. throw seems more portable, IMO.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
```diff
@@ -306,6 +306,14 @@ public virtual string CustomMethod
         [Parameter(ValueFromPipeline = true)]
         public virtual object Body { get; set; }
 
+        /// <summary>
+        /// Dictionary for use with RFC-7578 multipart\form-data submissions.
```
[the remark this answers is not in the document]
markekraus: fixed

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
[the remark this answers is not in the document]
iSazonov: I like `(formData == null)` too.
But here we should use the file pattern.

iSazonov (approved):

markekraus:
@SteveL-MSFT GH won't let me reply to the thread, but I added the comments about file contents.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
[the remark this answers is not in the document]
SteveL-MSFT: @markekraus since the pattern already exists in this file, I'm fine keeping it for now.  Consider this closed for me.

SteveL-MSFT (approved):

PR #6029, opened 2018-01-25
Add ScriptBlock substitution overload to -replace operator

### LINKED ISSUE #6164: Enhancement: support a script block as the replacement operand of the -replace operator to enable dynamic substitutions
**Wishful thinking**:

```powershell
# Script block as replacement operand allows dynamic substitutions.
# TBD: how to refer to the match at hand in the script block.
> 'Increment these numbers: 1 2 3' -replace '\d+', { [int] $Args.Value + 1 }   
Increment these numbers: 2 3 4
```

Sometimes string replacements need to be _dynamic_, i.e., the replacement string must be _derived from the match at hand_.

Consider the following (contrived example):

You want to increment all numbers in an input string; e.g.:  

`'Increment these numbers: 1 2 3'` -> `'Increment these numbers: 2 3 4'`

Currently, `-replace` doesn't support this type of replacement, because the replacement operand can only _statically_ refer to the match at hand (`$&` for the match in full, `$1` for the 1st capture group's match (if any), ...).

Thus, to achieve this kind of substitution the .NET Framework must currently be used:

```powershell
> [regex]::Replace('Increment these numbers: 1 2 3', '\d+', { param($match) [int] $match.Value + 1 })
Increment these numbers: 2 3 4
```

Needless to say, this is neither PowerShell-like nor simple.

---

Supporting the syntax as proposed at the top would greatly simplify such operations.

### Syntax details

The simplest approach is to use `[regex]::Replace()` as-is behind the scenes, and let  the script block reference the match at hand - which is passed as a `[System.Text.RegularExpressions.Match]` instance - via `$Args` (or, optionally, an explicitly defined parameter via `param()`).

The alternative is to stick with the `$&`, `$1`, ... syntax, but that would require behind-the-scenes manipulation of the script block.

### Backward-compatibility considerations

Currently, attempting to use a script block as the replacement operand simply _stringifies_ that block (i.e., uses its _string contents_ as the replacement string).

Since this behavior isn't useful, it's hard to imagine anyone relying on it, so this change probably falls into [Bucket 3: Unlikely Grey Area](https://github.com/PowerShell/PowerShell/blob/master/docs/dev-process/breaking-change-contract.md#bucket-3-unlikely-grey-area)

Environment data
----------------

Written as of  PowerShell Core v6.0.1

IISResetMe:
## PR Summary

Fix #6164.

This PR adds rudimentary lambda support to the `-replace` operator:

    PS C:\> "ID 0000123" -replace "\b0+", {return " " * $args[0].Value.Length}
    ID     123

----

The current `-replace` operator implementation has the following syntax:

    <value> -replace <pattern>[, <substitution>]

This (functionally) maps directly onto:

    Regex.Replace(<value>, <pattern>, <substitution>)

`Regex.Replace()` already has an overload that supports a [`MatchEvaluator` delegate](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.matchevaluator?view=netframework-4.7.1), which in turn allows us to pass a scriptblock instead of a string value as the substitution pattern.

Since only wiring is required, we can accomplish this with:

    if `<substitution>` argument is a ScriptBlock
        convert `<substitution>` to a `MatchEvaluator` instance
        pass MatchEvaluator to Regex.Replace() instead of "`<substitution>`"

**Update:**

To support arbitrary delegates  (like existing `MatchEvaluator`s and `PSMethod`s), 

    if `<substitution>` argument is not a string
        attempt to convert `<substitution>` to a `MatchEvaluator` instance
        pass MatchEvaluator to Regex.Replace() instead of "`<substitution>`"

### DIFF
--- a/src/System.Management.Automation/engine/lang/parserutils.cs
+++ b/src/System.Management.Automation/engine/lang/parserutils.cs
@@ -895,14 +895,13 @@ namespace System.Management.Automation
                         "BadReplaceArgument", ParserStrings.BadReplaceArgument, ignoreCase ? "-ireplace" : "-replace", rList.Count);
                 }
 
-                if(rList.Count > 1)
-                {
-                    substitute = rList[1];
-                }
-
                 if (rList.Count > 0)
                 {
                     pattern = rList[0];
+                    if (rList.Count > 1)
+                    {
+                        substitute = rList[1];
+                    }
                 }
             }
             else
@@ -963,14 +962,29 @@ namespace System.Management.Automation
         /// <returns>The result of the regex.Replace operation</returns>
         private static object ReplaceOperatorImpl(ExecutionContext context, string input, Regex regex, object substitute)
         {
-            MatchEvaluator matchEvaluator = null;
-            if (!(substitute is string) && LanguagePrimitives.TryConvertTo(substitute, out matchEvaluator))
+            switch (substitute)
             {
-                return regex.Replace(input, matchEvaluator);
-            }
+                case ScriptBlock sb:
+                    MatchEvaluator me = match => {
+                        var result = sb.DoInvokeReturnAsIs(
+                            useLocalScope: false, /* Use current scope to be consistent with 'ForEach/Where-Object {}' and 'collection.ForEach{}/Where{}' */
+                            errorHandlingBehavior: ScriptBlock.ErrorHandlingBehavior.WriteToCurrentErrorPipe,
+                            dollarUnder: match,
+                            input: AutomationNull.Value,
+                            scriptThis: AutomationNull.Value,
+                            args: Utils.EmptyArray<object>());
 
-            string replacement = PSObject.ToStringParser(context, substitute);
-            return regex.Replace(input, replacement);
+                        return PSObject.ToStringParser(context, result);;
+                    };
+                    return regex.Replace(input, me);
+
+                case object val when LanguagePrimitives.TryConvertTo(val, out MatchEvaluator matchEvaluator):
+                    return regex.Replace(input, matchEvaluator);
+
+                default:
+                    string replacement = PSObject.ToStringParser(context, substitute);
+                    return regex.Replace(input, replacement);
+            }
         }
 
         /// <summary>
--- a/src/System.Management.Automation/engine/lang/parserutils.cs
+++ b/src/System.Management.Automation/engine/lang/parserutils.cs
@@ -881,8 +881,8 @@ namespace System.Management.Automation
         /// <returns>The result of the operator</returns>
         internal static object ReplaceOperator(ExecutionContext context, IScriptExtent errorPosition, object lval, object rval, bool ignoreCase)
         {
-            string replacement = "";
             object pattern = "";
+            object substitute = "";
 
             rval = PSObject.Base(rval);
             IList rList = rval as IList;
@@ -895,13 +895,14 @@ namespace System.Management.Automation
                         "BadReplaceArgument", ParserStrings.BadReplaceArgument, ignoreCase ? "-ireplace" : "-replace", rList.Count);
                 }
 
+                if(rList.Count > 1)
+                {
+                    substitute = rList[1];
+                }
+
                 if (rList.Count > 0)
                 {
                     pattern = rList[0];
-                    if (rList.Count > 1)
-                    {
-                        replacement = PSObject.ToStringParser(context, rList[1]);
-                    }
                 }
             }
             else
@@ -935,8 +936,7 @@ namespace System.Management.Automation
             {
                 string lvalString = lval?.ToString() ?? String.Empty;
 
-                // Find a single match in the string.
-                return rr.Replace(lvalString, replacement);
+                return ReplaceOperatorImpl(context, lvalString, rr, substitute);
             }
             else
             {
@@ -944,12 +944,33 @@ namespace System.Management.Automation
                 while (ParserOps.MoveNext(context, errorPosition, list))
                 {
                     string lvalString = PSObject.ToStringParser(context, ParserOps.Current(errorPosition, list));
-
-                    resultList.Add(rr.Replace(lvalString, replacement));
+                    resultList.Add(ReplaceOperatorImpl(context, lvalString, rr, substitute));
                 }
 
                 return resultList.ToArray();
             }
+        }
+
+        /// <summary>
+        /// ReplaceOperator implementation.
+        /// Abstracts away conversion of the optional substitute parameter to either a string or a MatchEvaluator delegate
+        /// and finally returns the result of the final Regex.Replace operation.
+        /// </summary>
+        /// <param name="context">The execution context in which to evaluate the expression</param>
+        /// <param name="input">The input string</param>
+        /// <param name="regex">A Regex instance.</param>
+        /// <param name="substitute">The substitute value</param>
+        /// <returns>The result of the regex.Replace operation</returns>
+        private static object ReplaceOperatorImpl(ExecutionContext context, string input, Regex regex, object substitute)
+        {
+            MatchEvaluator matchEvaluator = null;
+            if (!(substitute is string) && LanguagePrimitives.TryConvertTo(substitute, out matchEvaluator))
+            {
+                return regex.Replace(input, matchEvaluator);
+            }
+
+            string replacement = PSObject.ToStringParser(context, substitute);
+            return regex.Replace(input, replacement);
         }
 
         /// <summary>
@@ -1906,4 +1927,3 @@ namespace System.Management.Automation
     }
     #endregion ScriptTrace
 }
-
--- a/test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
+++ b/test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
@@ -1,0 +1,64 @@
+Describe "Replace Operator" -Tags CI {
+    Context "Replace operator" {
+        It "Replace operator can replace string values using regular expressions" {
+            $res = "Get-Process" -replace "Get", "Stop"
+            $res | Should BeExactly "Stop-Process"
+
+            $res = "image.gif" -replace "\.gif$",".jpg"
+            $res | Should BeExactly "image.jpg"
+        }
+
+        It "Replace operator can be case-insensitive and case-sensitive" {
+            $res = "book" -replace "B","C"
+            $res | Should BeExactly "Cook"
+
+            $res = "book" -ireplace "B","C"
+            $res | Should BeExactly "Cook"
+
+            $res = "book" -creplace "B","C"
+            $res | Should BeExactly "book"
+        }
+
+        It "Replace operator can take 2 arguments, a mandatory pattern, and an optional substitution" {
+            $res = "PowerPoint" -replace "Point","Shell"
+            $res | Should BeExactly "PowerShell"
+
+            $res = "PowerPoint" -replace "Point"
+            $res | Should BeExactly "Power"
+        }
+    }
+
+    Context "Replace operator substitutions" {
+        It "Replace operator supports numbered substitution groups using ```$n" {
+            $res = "domain.example" -replace ".*\.(\w+)$","Tld of '`$0' is - '`$1'"
+            $res | Should BeExactly "Tld of 'domain.example' is - 'example'"
+        }
+
+        It "Replace operator supports named substitution groups using ```${name}" {
+            $res = "domain.example" -replace ".*\.(?<tld>\w+)$","`${tld}"
+            $res | Should BeExactly "example"
+        }
+
+        It "Replace operator can take a ScriptBlock in place of a substitution string" {
+            $res = "ID ABC123" -replace "\b[A-C]+", {return "X" * $args[0].Value.Length}
+            $res | Should BeExactly "ID XXX123"
+        }
+
+        It "Replace operator can take a MatchEvaluator in place of a substitution string" {
+            $matchEvaluator = {return "X" * $args[0].Value.Length} -as [System.Text.RegularExpressions.MatchEvaluator]
+            $res = "ID ABC123" -replace "\b[A-C]+", $matchEvaluator
+            $res | Should BeExactly "ID XXX123"
+        }
+
+        It "Replace operator can take a static PSMethod in place of a substitution string" {
+            class R {
+                static [string] Replace([System.Text.RegularExpressions.Match]$Match) {
+                    return "X" * $Match.Value.Length
+                }
+            }
+            $substitutionMethod = [R]::Replace
+            $res = "ID 0000123" -replace "\b0+", $substitutionMethod
+            $res | Should BeExactly "ID XXXX123"
+        }
+    }
+}
--- a/test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
+++ b/test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
@@ -1,3 +1,6 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
 Describe "Replace Operator" -Tags CI {
     Context "Replace operator" {
         It "Replace operator can replace string values using regular expressions" {
@@ -40,7 +43,7 @@ Describe "Replace Operator" -Tags CI {
         }
 
         It "Replace operator can take a ScriptBlock in place of a substitution string" {
-            $res = "ID ABC123" -replace "\b[A-C]+", {return "X" * $args[0].Value.Length}
+            $res = "ID ABC123" -replace "\b[A-C]+", {return "X" * $_[0].Value.Length}
             $res | Should BeExactly "ID XXX123"
         }
 


src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -815,6 +816,10 @@ internal static object ReplaceOperator(ExecutionContext context, IScriptExtent e
                     pattern = rList[0];
                     if (rList.Count > 1)
                     {
+                        if(rList[1] is ScriptBlock)
```
powercode: Shouldn't this also support a `MatchEvaluator` delegate? Do that we can use a method from a PowerShell or .NET class.
That is, in addition to the scriptblock.

IISResetMe: Good call! 
Wondering whether it should accept/evaluate `System.Delegate`, `System.Func<string>` and `System.Func<Match, string>` as well...

IISResetMe: Nevermind, cast will fail.

IISResetMe: @powercode Passing a `PSMethodInfo` object would make sense as a use case, but I'm not sure how to safely construct a scriptblock or delegate that dispatches it.

Any ideas?

powercode:
I'd think it would be better to check if you have something that can be converted to a MatchEvaluator.
PSMethods can now be converted to matching delegates.

This works:
```
class R { static [string] RR([Text.RegularExpressions.Match] $m) {return "Foo"}}
[System.Text.RegularExpressions.MatchEvaluator] $m = [R]::RR
[Regex]::replace("bar", 'a', [R]::RR)
```
It would be nice if 
```
"bar" -replace "a", [R]::RR
```
worked too.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -816,7 +816,11 @@ internal static object ReplaceOperator(ExecutionContext context, IScriptExtent e
                     pattern = rList[0];
                     if (rList.Count > 1)
                     {
-                        if(rList[1] is ScriptBlock)
+                        if (rList[1] is MatchEvaluator)
```
powercode: Maybe have a fast path for the string replacement (likely the most common case), and then, if it isn't a string check for
scriptblock, MatchEvaluator, and PSMethod?
If it is a PSMethod, try to convert it to a MatchEvaluator.

test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
```diff
@@ -0,0 +1,56 @@
...
+            $res = "ID 0000123" -replace "\b0+", {return " " * $args[0].Value.Length}
+            $res | Should Be "ID     123"
+        }
+
+        It "Replace operator can take a MatchEvaluator in place of a substitution string" {
+            $matchEvaluator = {return " " * $args[0].Value.Length} -as [System.Text.RegularExpressions.MatchEvaluator]
+            $res = "ID 0000123" -replace "\b0+", $matchEvaluator
+            $res | Should Be "ID     123"
+        }
+    }
```
powercode: I would like a test like 
```
class R { static [string] ReplaceWithShell([Text.RegularExpressions.Match] $m) {return "Shell"}}
"PowerPoint" -replace "Point", [R]::ReplaceWithShell | Should be "PowerShell"
```

src/System.Management.Automation/engine/lang/parserutils.cs
IISResetMe: How about a single branch like:

    if(!(rList[!]] is string))
        Attempt conversion to MatchEvaluator

LaurentDardenne:
What happens if the delegate code raises an exception ?
Is the _Match_ delegate parameter named _**$args**_ under Powershell ?

IISResetMe:
@LaurentDardenne `Replace()` will throw a `RuntimeException` if the delegate throws during substitution

If you pass a scriptblock with no explicit params, the `Match` parameter argument populates the `$args` automatic variable. User could also explicitly name it if desired, no functional difference

LaurentDardenne:
Thank you.

test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
```diff
@@ -0,0 +1,67 @@
+Describe "Split Operator" -Tags CI {
```
iSazonov: Typo.

powercode (approved):

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -815,6 +818,10 @@ internal static object ReplaceOperator(ExecutionContext context, IScriptExtent e
                     pattern = rList[0];
                     if (rList.Count > 1)
                     {
+                        if (!(rList[1] is string))
+                        {
+                            useMatchEvaluator = LanguagePrimitives.TryConvertTo(rList[1], out matchEvaluator);
+                        }
                         replacement = PSObject.ToStringParser(context, rList[1]);
```
powercode: Should this line be run if we use a matchEvaluator?

IISResetMe: [`TryConvertEmpty()` will do a simple cast and return early](https://github.com/PowerShell/PowerShell/blob/a9c6292903d86a7bf3b3bfb59110d03d1234c525/src/System.Management.Automation/engine/LanguagePrimitives.cs#L1681) if the input object and out param are the same type, no need to complicate the `ReplaceOperator()` implementation for this (edge-)case

test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
IISResetMe: Thanks @iSazonov, fixed in commit aa54fe56a9efb2c0c49b551599ec193d67c47e36

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -851,6 +858,12 @@ internal static object ReplaceOperator(ExecutionContext context, IScriptExtent e
                 string lvalString = lval?.ToString() ?? String.Empty;
 
                 // Find a single match in the string.
+                if (useMatchEvaluator)
+                {
+                    // If a MatchEvaluator has been assigned, we
+                    // use the corresponding Regex.Replace() overload
+                    return rr.Replace(lvalString, matchEvaluator);
+                }
                 return rr.Replace(lvalString, replacement);
```
iSazonov: The code (L860-L867, L876-L883) should be in a help method.

test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
```diff
@@ -0,0 +1,67 @@
+Describe "Replace Operator" -Tags CI {
+    Context "Replace operator" {
+        It "Replace operator can replace string values using regular expressions" {
+            $res = "Get-Process" -replace "Get", "Stop"
+            $res | Should Be "Stop-Process"
+
+            $res = "image.gif" -replace "\.gif$",".jpg"
```
iSazonov: Why we need the test? Looks as dup line 4. Do it test another code path?

test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
```diff
@@ -0,0 +1,67 @@
...
+            $res = "book" -ireplace "B","C"
+            $res | Should BeExactly "Cook"
+
+            $res = "book" -creplace "B","C"
+            $res | Should BeExactly "book"
+        }
+
+        It "Replace operator can take 2 arguments, a mandatory pattern, and an optional substitution" {
+            $res = "PowerPoint" -replace "Point","Shell"
+            $res | Should Be "PowerShell"
```
iSazonov: Please use `BeExactly` for strings. Below too.

test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    Context "Replace operator substitutions" {
+        It "Replace operator supports numbered substitution groups using ```$n" {
+            $res = "domain.example" -replace ".*\.(\w+)$","`$0"
+            $res | Should Be "domain.example"
+
+            $res = "domain.example" -replace ".*\.(\w+)$","`$1"
+            $res | Should Be "example"
```
iSazonov: We could use one Should with `$res = "domain.example" -replace ".*\.(\w+)$","$0 - $1" `

test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        It "Replace operator supports named substitution groups using ```${name}" {
+            $res = "domain.example" -replace ".*\.(?<tld>\w+)$","`${tld}"
+            $res | Should Be "example"
+        }
+
+        It "Replace operator can take a ScriptBlock in place of a substitution string" {
+            $res = "ID 0000123" -replace "\b0+", {return " " * $args[0].Value.Length}
```
iSazonov: Spaces are not visible - it looks bad. Could you please use another char? Below too.

iSazonov (dismissed):

test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
IISResetMe: First test is the first and simplest example from the `about_Comparison_Operators` help file, this one is the first example that shows it's actually a regex operator and not just static string replacement.

test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
IISResetMe: Done

test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
IISResetMe: Done

test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
IISResetMe: Updated all three cases

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -850,23 +851,42 @@ internal static object ReplaceOperator(ExecutionContext context, IScriptExtent e
...
-                    resultList.Add(rr.Replace(lvalString, replacement));
+                    resultList.Add(ReplaceOperatorImpl(context, errorPosition, lvalString, rr, substitute));
                 }
 
                 return resultList.ToArray();
             }
         }
 
+        /// <summary>
+        /// ReplaceOperator implementation...
```
iSazonov: Really we don't need public comment for private method but if we use it please put one final dot in the comment and comments below. Also it is useful to mention that the method is needed to detect a matchEvaluator or use string regex (please reword :) ).

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -850,23 +851,42 @@ internal static object ReplaceOperator(ExecutionContext context, IScriptExtent e
...
+        /// </summary>
+        /// <param name="context">The execution context in which to evaluate the expression</param>
+        /// <param name="errorPosition">The position to use for error reporting.</param>
```
iSazonov: We don't use the parameter. Please remove.

iSazonov:
@IISResetMe Thanks for your contribution. Now we are waiting PowerShell Committee conclusion in days.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and agree with taking this change, would prefer `$_` implemented, but can accept `$args[0]` since it's automatically there

IISResetMe:
@SteveL-MSFT Cool! I agree that `$_` would be nicer, but since the we're passing the scriptblock directly to the method call as a delegate I can't really inject anything into the resulting expression . 

Only way I can think of is by parsing the user-supplied scriptblock, inspecting the `param` block (optionally adding `param([Match]$_)`) and then re-creating it. It feels a bit ugly.

If anyone has suggestions or comments on this approach (or better ideas) I'm all ears.

lzybkr:
@IISResetMe - you could create a closure as the delegate instead, something like this:

```C#
MatchEvaluator GetEvaluator(ScriptBlock sb)
{
    return m => sb.DoInvokeReturnAsIs(
        useLocalScope: true, // or false, not sure
        errorHandlingBehavior: ErrorHandlingBehavior.WriteToCurrentErrorPipe,
        dollarUnder: m,
        input: AutomationNull.Value,
        scriptThis: AutomationNull.Value,
        args: Utils.EmptyArray<object>());
}
```

IISResetMe:
@lzybkr Interesting, I thought `DoInvokeReturnAsIs()` would execute the scriptblock immediately?

IISResetMe:
As expected, the following works as intended:

        private static object ReplaceOperatorImpl(ExecutionContext context, string input, Regex regex, object substitute)
        {
            MatchEvaluator matchEvaluator = null;
            if (!(substitute is string) )
            {
                if(substitute is ScriptBlock) 
                {
                    ScriptBlock substituteScriptBlock = substitute as ScriptBlock;
                    ScriptBlockAst substituteAst = substituteScriptBlock.Ast as ScriptBlockAst;
                    if(substituteAst.ParamBlock == null && substituteAst.Find(ast => ast is VariableExpressionAst && (ast as VariableExpressionAst).VariablePath.UserPath == "_", false) != null)
                    {
                        substitute = ScriptBlock.Create(@"param([System.Text.RegularExpressions.Match]$_)" + substituteScriptBlock.ToString());
                    }
                }   
                if(LanguagePrimitives.TryConvertTo(substitute, out matchEvaluator)){
                    return regex.Replace(input, matchEvaluator);
                }
            }

            string replacement = PSObject.ToStringParser(context, substitute);
            return regex.Replace(input, replacement);
        }

... but it doesn't look or feel very clean :-\

IISResetMe:
@lzybkr Had a closer look at your suggestion, makes sense, but when I try to compile it I get:

    .\PowerShell\src\System.Management.Automation\engine\lang\parserutils.cs(909,25): 
      error CS1662: Cannot convert lambda expression to intended delegate type because some of the return types in the block are not implicitly convertible to the delegate return type

lzybkr:
Oh, right, I guess call `LanguagePrimitives.ConvertTo<string>` on the return value.

iSazonov:
Added fix #6164 in description.

iSazonov:
@IISResetMe What is the PR status?

daxian-dbw:
@IISResetMe This PR is a good enhancement to have. But it's been a while since this PR was left unattended, so I updated the changes to address the `$_` pending issue. Hope you don't mind.

My changes were:
1. Squash the existing commits to a single one.
2. Refactor `ReplaceOperatorImpl` with `switch` (pattern matching).

@iSazonov @lzybkr @powercode Can you please take another look?

test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
iSazonov: I'd exclude `substitutionMethod` too.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -963,14 +962,29 @@ internal static object ReplaceOperator(ExecutionContext context, IScriptExtent e
...
-            MatchEvaluator matchEvaluator = null;
-            if (!(substitute is string) && LanguagePrimitives.TryConvertTo(substitute, out matchEvaluator))
+            switch (substitute)
             {
-                return regex.Replace(input, matchEvaluator);
-            }
+                case ScriptBlock sb:
+                    MatchEvaluator me = match => {
+                        var result = sb.DoInvokeReturnAsIs(
+                            useLocalScope: false, /* Use local scope to be consistent with 'ForEach/Where-Object {}' and 'collection.ForEach{}/Where{}' */
```
iSazonov: I don't  understand the comment here - we use `false`.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -963,14 +962,29 @@ internal static object ReplaceOperator(ExecutionContext context, IScriptExtent e
...
+                    };
+                    return regex.Replace(input, me);
 
-            string replacement = PSObject.ToStringParser(context, substitute);
-            return regex.Replace(input, replacement);
+                case object val when LanguagePrimitives.TryConvertTo(val, out MatchEvaluator matchEvaluator):
+                    return regex.Replace(input, matchEvaluator);
+
+                default:
+                    string replacement = PSObject.ToStringParser(context, substitute);
```
iSazonov: I don't see tests for `PSObject.ToStringParser` - why we use this?

test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
```diff
@@ -40,7 +43,7 @@ Describe "Replace Operator" -Tags CI {
         }
 
         It "Replace operator can take a ScriptBlock in place of a substitution string" {
-            $res = "ID ABC123" -replace "\b[A-C]+", {return "X" * $args[0].Value.Length}
```
iSazonov: I'd keep the `$args` test too.

powercode:
@daxian-dbw LGTM!

IISResetMe:
Thanks @daxian-dbw! Couldn't get `$_` to work inside the MatchEvaluator and then life got in the way :)

Looks good!

src/System.Management.Automation/engine/lang/parserutils.cs
daxian-dbw: Good catch. It's a typo, should be "use current scope ..."

src/System.Management.Automation/engine/lang/parserutils.cs
daxian-dbw: It's used in the original code, to convert an object to a string:
```c#
if (rList.Count > 1)
{
    replacement = PSObject.ToStringParser(context, rList[1]);
}
```

test/powershell/Language/Operators/ReplaceOperator.Tests.ps1
daxian-dbw: `$args` won't work. When used in the form of `something -replace xxx, { ... }`, the "Match" object is passed to `$_` of the script block, just like `ForEach/Where-Object {...}` and `collection.ForEach{...}/Where{...}`.

daxian-dbw:
@iSazonov Your comments have been addressed. Can you please take another look?

lzybkr (approved):

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: What PowerShell magic it give us?

daxian-dbw: It converts an object to a string in a powershell way. You can see the code yourself for more information.

daxian-dbw: It's an internal method at https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/MshObject.cs#L1206

daxian-dbw:
@iSazonov Given that your pending comment is not blocking (questions about `PSObject.ToStringParser`), I will merge this PR. Feel free to continue the discussion in the comment thread.

src/System.Management.Automation/engine/lang/parserutils.cs
iSazonov: @daxian-dbw Thanks! I thought that using the method add some magic in the PR but no - it is only to get string.

PR #6043, opened 2018-01-26
Add script to set the shortcut working directory to the user's home folder.

### LINKED ISSUE #5705: Installer should start PS Core with user's home dir as the working dir
Steps to reproduce
------------------

```powershell
Install PS Core rc2 and at the select check the box to start PowerShell.
Find PowerShell Core in the task bar, right-click it and select "Pin to taskbar"
```

Expected behavior
-----------------
PowerShell Core as started by the installer, starts in my home dir.
PowerShell Core as started by the taskbar icon, starts in my home dir.

Actual behavior
---------------
PowerShell Core in both cases starts in `C:\Program Files\PowerShell\6.0.0-rc.2`.

I wind up having to edit the taskbar shortcut's properties to change the "Start in:" location to `%USERPROFILE%` - which works but I shouldn't have to do this.  I suspect other folks will use the PS Core started by the installer as the easy means to get PS Core pinned to their taskbar.

If the installer could set the working dir to the current user's home dir before launching PowerShell, that would probably fix both scenarios.

Environment data
----------------

```powershell
> $PSVersionTable
```

gpduck:
## PR Summary

Update the shortcut created by the installer to use %HOMEDRIVE%%HOMEPATH% for the working directory.

Fixes #5705

## PR Checklist

### DIFF
--- a/assets/Product.wxs
+++ b/assets/Product.wxs
@@ -43,6 +43,18 @@
     <!-- Default value of Checkbox of starting PowerShell after installation -->
     <Property Id="WixShellExecTarget" Value="[$(var.ProductVersionWithName)]pwsh.exe"/>
     <CustomAction Id="LaunchApplication" BinaryKey="WixCA" DllEntry="WixShellExec" Impersonate="yes" />
+    <CustomAction Id="SaveShortcutPath" Property="ShortcutPath" Value="[$ApplicationProgramsMenuShortcut]$(var.ProductSemanticVersionWithNameAndOptionalArchitecture).lnk" />
+    <CustomAction Id="FixShortcutWorkingDirectory" Script="vbscript" HideTarget="no" Impersonate="no">
+      shortcutPath = Session.Property("ShortcutPath")
+      Set fso = CreateObject("Scripting.FileSystemObject")
+      if (fso.FileExists(shortcutPath)) Then
+        Set wshShell = CreateObject("WSCript.Shell")
+        Set shortcut = wshShell.CreateShortcut(shortcutPath)
+
+        shortcut.workingdirectory = "%HOMEDRIVE%%HOMEPATH%"
+        shortcut.save
+      End If
+    </CustomAction>
     <UI>
       <Dialog Id="MyExitDialog" Width="370" Height="270" Title="!(loc.ExitDialog_Title)">
         <Control Id="LaunchCheckBox" Type="CheckBox" X="10" Y="243" Width="170" Height="17" Property="LAUNCHAPPONEXIT" Hidden="yes" CheckBoxValue="1" Text="Launch $(env.ProductName)">
@@ -185,6 +197,12 @@
         </Directory>
       </Directory>
     </Directory>
+    <InstallExecuteSequence>
+      <Custom Action="SaveShortcutPath" After="InstallFinalize" />
+      <Custom Action="FixShortcutWorkingDirectory" After="SaveShortcutPath">
+        NOT Installed OR NOT REMOVE OR UPGRADINGPRODUCTCODE
+      </Custom>
+    </InstallExecuteSequence>
   </Product>
 
 <!-- Explorer Context Menu Dialog -->


assets/Product.wxs
```diff
@@ -43,6 +43,18 @@
...
     <CustomAction Id="LaunchApplication" BinaryKey="WixCA" DllEntry="WixShellExec" Impersonate="yes" />
+    <CustomAction Id="SaveShortcutPath" Property="ShortcutPath" Value="[$ApplicationProgramsMenuShortcut]$(var.ProductSemanticVersionWithNameAndOptionalArchitecture).lnk" />
+    <CustomAction Id="FixShortcutWorkingDirectory" Script="vbscript" HideTarget="no" Impersonate="no">
+      shortcutPath = Session.Property("ShortcutPath")
+      Set fso = CreateObject("Scripting.FileSystemObject")
+      if (fso.FileExists(shortcutPath)) Then
+        Set wshShell = CreateObject("WSCript.Shell")
+        Set shortcut = wshShell.CreateShortcut(shortcutPath)
+
+        shortcut.workingdirectory = "%HOMEDRIVE%%HOMEPATH%"
```
iSazonov: Can we use [WorkingDirectory property in Shortcut](http://wixtoolset.org/documentation/manual/v3/howtos/files_and_registry/create_start_menu_shortcut.html)
in the [line](https://github.com/gpduck/PowerShell/blob/8119712fccb71a2c0d5c9b49d49aa212b067ada2/assets/Product.wxs#L192)

gpduck: That WIX property translates to [this](https://msdn.microsoft.com/library/aa371847.aspx#wkdir) Windows Installer property, which resolves the environment variable at install time.  If we put `%HOMEDRIVE%%HOMEPATH%` into the WIX configuration it creates the shortcut hardcoded to the value for the user who is doing the install.

As far as I can tell there is no way to override this Windows Installer behavior.

iSazonov: Maybe this works https://community.flexerasoftware.com/showthread.php?155257-Using-environment-variable-in-a-shortcut

gpduck: I tried it as a WIX property set to `%HOMEDRIVE%%HOMEPATH%` and Windows Installer still resolves it at install time to a hardcoded path.  I also tried various methods of escaping and they all got passed through as literal text that resulted in an invalid working directory (ex `[\%]HOMEDRIVE[\%]`).

The one other method that did show promise was:

1. Set an environment variable (ex TempWkDir) to `%HOMEDRIVE%%HOMEPATH%` (this was set on the target machine outside of the install process)
1. Create a WIX property with value of the environment variable
1. Point the shortcut WorkingDirectory attribute to the WIX property

This resulted in `%TempWkDir%` being resolved to `%HOMEDRIVE%%HOMEPATH%` at install time and the desired shortcut being created.  The problem is I could not find a way to temporarily set that environment variable on the install process's environment.  Using WIX's environment settings doesn't update the install process's environment, so the value was not resolved properly during the install.

Also setting up several layers of environment variables and WIX properties that get substituted for each other seems more complicated for future maintenance than just a simple script that explicitly updates the shortcut.

iSazonov: The link above suggest `[HOMEDRIVE][HOMEPATH]`.

gpduck: Those get passed through without being modified, I don't think they're valid properties in MSI: https://msdn.microsoft.com/en-us/library/windows/desktop/aa370905(v=vs.85).aspx

iSazonov: Thanks! I see we have to use a script.

I think an user profile root folder is not appropriate work directory - user should create nothing in the directory. What about the user `Documents` folder?

rkeithhill: Huh?  Then why doesn't PowerShell define `$Home` to be the Documents dir?  I put files in the root of my home dir all the time.

gpduck: I'm open to other folders but I'll leave these notes as well:

1. %HOMEDRIVE%%HOMEPATH% matches the behavior of both cmd.exe and powershell.exe which will give a consistent user experience
1. Users can change the value of HomePath (not that I've ever bothered to do this)
1. Some cloud sync services create folders in the profile directory (OneNote, Google Drive) that may be the preferred location for some people, this puts them at the root of their profile so they can choose which to go in to

TravisEz13:
Wouldn't this script only set the correct path for the user that installed PowerShell?

gpduck:
It sets the value to the literal string "%HOMEDRIVE%%HOMEPATH%" which is then resolved at launch time to the current user's path.

TravisEz13 (approved):

assets/Product.wxs
iSazonov: >I put files in the root of my home dir all the time.

😄 In my enterprise, we warn users of this practice, and even manage to block it as virus suspicious.

I'd reword this arguments:
1. User can have many workspaces and we can not predict better location.
2. User shouldn't change HomePath but they can change the shortcut or even better to create new one.
3. %HOMEDRIVE%%HOMEPATH% looks better for consistency with Unix.

iSazonov:
@gpduck Can you confirm that the shortcut correctly removed in uninstall time?

gpduck:
@iSazonov Yes, I just verified with the artifact from AppVeyor that the shortcut is properly removed on uninstall.

iSazonov (approved):
@gpduck Many thanks for your contribution!

PR #6013, opened 2018-01-24
Add WebListener Verbosity and CI VM sleep Workaround

markekraus:
## PR Summary

Closes  #6001

* Add verbosity to the WebListener when it fails to assist in troubleshooting
* Switch the WebListener initialization timeout to count cycles instead of using fixed dates to work around possible VM CI sleep/ issues.

## An example message when WebListener fails to start

https://ci.appveyor.com/project/PowerShell/powershell/build/v6.1.0-preview.7742#L2508

```none
Describing Invoke-WebRequest tests
 [-] Error occurred in Describe block 16.43s
   WebListener did not start before the timeout was reached.
   Errors:
   NotSpecified: (:String) [], RemoteException
   Unhandled Exception: System.ApplicationException: Bad things happened.
      at mvc.Program.Main(String[] args) in C:\projects\powershell\test\tools\WebListener\Program.cs:line 23
   
   Output:
   appDLL: WebListener.dll
   serverPfxPath: C:\projects\powershell\test\tools\Modules\WebListener\ServerCert.pfx
   serverPfxPassword: password
   HttpPort: 8083
   Https: 8084
   Tls11Port: 8085
   TlsPort: 8086
   
   Verbose:
   
   At C:\projects\powershell\test\tools\Modules\WebListener\WebListener.psm1:107 char:13
```

## PR Checklist

Note: Please mark anything not applicable to this PR `NA`.

- [x] [PR has a meaningful title](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - [x] Use the present tense and imperative mood when describing your changes
- [x] [Summarized changes](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [NA] User facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - [NA] Issue filed - Issue link:
- [x] [Change is not breaking](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes)
- [NA] [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
[...]

PR #6055, opened 2018-01-28
Remove -TimeOutSec from non timeout related Web Cmdlets Tests

markekraus:
## PR Summary

Timeouts were originally used in many tests because they relied on unreliable external sites. WebListener may be a bit slow inside CI, but it is reliable in that it is always present. This PR removes the `-TimeOutSec` option from tests where `-TimeOutSec` is not required. the remaining instances either it or would be unaffected (such as testing immediate parameter validation errors). 

## PR Checklist

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -76,9 +76,9 @@ function ExecuteRequestWithHeaders {
     try {
         $headers = @{ Connection = 'close'}
         if ($cmdletName -eq "Invoke-WebRequest") {
-            $result.Output = Invoke-WebRequest -Uri $uri -TimeoutSec 5 -Headers $headers
+            $result.Output = Invoke-WebRequest -Uri $uri -Headers $headers
         } else {
-            $result.Output = Invoke-RestMethod -Uri $uri -TimeoutSec 5 -Headers $headers
+            $result.Output = Invoke-RestMethod -Uri $uri -Headers $headers
         }
     } catch {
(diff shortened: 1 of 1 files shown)


iSazonov:
The timeouts look useful.
The timeouts in web tests helped us to identify the problem #6069 with long running tests.
If we remove the timeouts we can catch silent test hang.
Maybe it doesn't matter while we use WebListener with local Kestrel but once we can be forced to use another tool for some reason.
I'd rather leave the timeouts as is.

/cc @adityapatwardhan @JamesWTruher Could you please make conclusion?

markekraus:
The problem is that -TimeOutSec option affects the internal logic of the cmdlets and can pollute the results. Sure they are helpful in this one-off instance.. but.. in the case where someone messes with the logic we could suffer a regression.

If we want to timeout the cmdlets, we need to do it outside the cmdltes and not with the internal logic of the cmdlets.

daxian-dbw:
I think both of you have good points. Ideally, the timeout mechanism should be supported by the test framework (Pester here).

dantraMSFT:
The only suggestion I have is to query the history of these tests since the weblistener was introduced and determine if there are cases where the timeout was hit; otherwise, LGTM.

dantraMSFT (approved):

SteveL-MSFT:
Created https://github.com/pester/Pester/issues/983 to have them support `-Timeout`

adityapatwardhan (approved):
LGTM. I also prefer timeouts to be used only when testing the timeout functionality.

iSazonov:
@dantraMSFT This PR, #6069 and #6013 is just the result of the fact that these tests have recently started to get timeouts.

iSazonov:
I agree that it is better to have timeout support in Pester.

Issue #1267, opened 2016-07-01
More function should support pipelining


lzybkr:
The current definition of more does not pipeline, it collects all output in $input and then pipes to more after the upstream command finishes.

joeyaiello:
@lzybkr per our conversation with the @PowerShell/powershell-maintainers, would this actually be something that goes to `0 - Backlog`? And if so, what's the intended purpose of `2 - Ready`?

iSazonov:
@lzybkr I attached a script file to address the Issue. Could you please look when you'll have free time? If it is right direction I'll push a PR.
[Next-More.txt](https://github.com/PowerShell/PowerShell/files/1643154/Next-More.txt)

PR #6059, opened 2018-01-29
Remove 'more' function and move the $env:PAGER capability into the help function

### LINKED ISSUE #6497: Help cmdlet fails in PS 6.1.0-preview.1 when StrictMode is enabled
Steps to reproduce
------------------

```powershell
Set-StrictMode -Version Latest
help
```
or

```powershell
Set-StrictMode -Version Latest
man
```

Expected behavior
-----------------

```powershell
TOPIC                                                    
    PowerShell Help System                               
                                                         
SHORT DESCRIPTION                                        
    Displays help about PowerShell cmdlets and concepts. 
...
```

Actual behavior
---------------

```powershell
PSMessageDetails      :
Exception             : System.Management.Automation.RuntimeException: The variable '$moreArgs' cannot be retrieved because it has not been set.
                           at System.Management.Automation.VariableOps.GetVariableValue(VariablePath variablePath, ExecutionContext executionContext, VariableExpressionAst varAst)
                           at System.Management.Automation.Interpreter.FuncCallInstruction`4.Run(InterpretedFrame frame)
                           at System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame frame)
TargetObject          : moreArgs
CategoryInfo          : InvalidOperation: (moreArgs:String) [], RuntimeException
FullyQualifiedErrorId : VariableIsUndefined
ErrorDetails          :
InvocationInfo        : System.Management.Automation.InvocationInfo
ScriptStackTrace      : at <ScriptBlock>, <No file>: line 24
                        at <ScriptBlock>, <No file>: line 69
                        at <ScriptBlock>, <No file>: line 1
PipelineIterationInfo : {}
```

Environment data
----------------

```powershell
PSVersion                      6.1.0-preview.1
GitCommitId                    v6.1.0-preview.1
```

iSazonov:
## PR Summary

Address #1267.
Fix #6497.

Based on PowerShell Committee conclusion (see below) and PR Native pipe  #2450 we remove `more` function. Now users is free to use preffered pager directly or create a custom alias.
To keep current `help` function behavior we move a pager capability (with $env:PAGER support) to the function.

~~Before the change we had to wait 10 seconds before we could see the script output:~~
~~& { 1..100; sleep -seconds 10; 101..200 } | more~~
~~With this change we will see the output without delay.~~

~~I have no idea how to test this.~~

## PR Checklist


iSazonov:
Please review. I think I did everything I could. Some Help tests failed - looks as breaking change. Also see `Get-Process | more`.

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -4754,28 +4754,68 @@ internal static SessionStateAliasEntry[] BuiltInAliases
...
+        $moreCommand = (Get-Command -CommandType Application more | Select-Object -First 1).Definition
+    } else {
+        $moreCommand = (Get-Command -CommandType Application less | Select-Object -First 1).Definition
+    }
+
+    $process = [System.Diagnostics.Process]::new()
+
+    $process.StartInfo.FileName = 'C:\Windows\system32\more.com'
+    $process.StartInfo.UseShellExecute = $false
+    $process.StartInfo.RedirectStandardInput = $true
```
daxian-dbw: The design is apparently not cross-plat. Please revisit the design and make sure it works on all supported platforms.

BrucePay (commented):
Do we still need the more function? I thought @vors fixed the pipeline so that objects would be streamed into executables. Certainly
    & { 1..100; sleep -seconds 10; 101..200 } | more
is slow, but 
    & { 1..100; sleep -seconds 10; 101..200 } | more.com
and
    & { 1..100; sleep -seconds 10; 101..200 } | less.exe
show output immediately.

iSazonov:
@BrucePay Good catch! I think it is best way. Seems we can use an alias for more.com/less.
```powershell
Set-Alias -Name tst -Value "more.com"`
& { 1..100; sleep -seconds 10; 101..200 } | tst
```
show output immediately.
If it's acceptable I'll close the PR and make a new.

daxian-dbw:
Removing `more` function would be a breaking change in that the `more` function takes multiple file path strings separated by comma `,`, while `more.com` and `less` don't.

```powershell
PS:11> more.com F:\test.txt, F:\new.txt
Cannot access file F:\test.txt
```
Besides, the `more` function uses a pager if `env:PAGER` is set. Not sure if this is really practically used by anyone though.

rkeithhill:
I use Pager on Windows in order to use less.  An alias for Pager would be OK  as long as I  can override it and set it to less.

iSazonov:
We could create the alias at PowerShell startup and keep current platform specific logic - assign `more.com` on Windows,  `less` on Unix or use `env:PAGER`.

daxian-dbw:
Update the PR title to add `[WIP]`.

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -4754,28 +4754,68 @@ internal static SessionStateAliasEntry[] BuiltInAliases
...
+        $moreCommand = (Get-Command -CommandType Application $env:PAGER | Select-Object -First 1).Definition
+    } elseif ($IsWindows) {
+        $moreCommand = (Get-Command -CommandType Application more | Select-Object -First 1).Definition
+    } else {
+        $moreCommand = (Get-Command -CommandType Application less | Select-Object -First 1).Definition
+    }
+
+    $process = [System.Diagnostics.Process]::new()
+
+    $process.StartInfo.FileName = 'C:\Windows\system32\more.com'
```
rkeithhill: Why use a path that is only valid on Windows and a binary `more.com` that is only valid on Windows?  Also, why does the pipeline case ignore the PAGER utility specified in the $env:PAGER?  I always use `less`, even on Windows.  I **need** to be able specify `less` on Windows.  If I was forced to use `more.com` I'd go batsh!t in short order.  ;-)

iSazonov:
@daxian-dbw @SteveL-MSFT 
Could you please discuss the breaking change (implement alias for `more` with default `more.com` on Windows and `less` on Unix and given `env: PAGER`) on PowerShell Committee?

SteveL-MSFT:
@iSazonov can you be more specific on what you think needs to be discussed by @PowerShell/powershell-committee?  Is it whether we should remove the `more` function altogether?

iSazonov:
@SteveL-MSFT Yes, we could try to remove `more` function - but it is a breaking change as @daxian-dbw said above (` more.com F:\test.txt, F:\new.txt` ).

iSazonov:
Should we implement an alias for `more` with the $env:PAGER capability?
```powershell
New-Alias -Name more -Value more.com # for Windows
New-Alias -Name more -Value less           # for Unix
```

SteveL-MSFT:
@iSazonov the discussion in the committee is that `more` should just be `more.com` on Windows which seems to be working and not have issues any more based on the comments in the function (we tested against a nanoserver docker container).  It seemed wrong that users would have to use `more` to get `$env:PAGER` as they should just invoke that specific pager directly.  For tools that require automatic paging (like help), it makes sense to respect $env:PAGER (we couldn't think of any other case outside of `help` function which itself is a separate discussion if we should have that instead of just alias `help` directly to `get-help` and support paging...)

iSazonov:
@daxian-dbw @SteveL-MSFT The PR is ready for review.

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -4262,7 +4266,16 @@ .FORWARDHELPCATEGORY Cmdlet
     }
     else
     {
-        $help | more
+        # Respect PAGER, use more on Windows, and use less on Linux
+        if (Test-Path env:PAGER) {
+            $moreCommand = (Get-Command -CommandType Application $env:PAGER | Select-Object -First 1).Definition
```
rkeithhill: This will lose the ability to specify arguments to the pager utility - `$pager,$moreArgs = $env:PAGER -split '\s+' `.   That doesn't bother me personally but presumably somebody added these for a reason (an issue I suspect).

SteveL-MSFT: Yes, that was explicitly to fix https://github.com/PowerShell/PowerShell/issues/6142

Should just use the same code as from https://github.com/PowerShell/PowerShell/pull/6144/files

iSazonov: Thanks for the reminder! 

Fixed.

iSazonov: Is it possible to add tests to exclude such regression?

SteveL-MSFT (approved):

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -4262,7 +4266,17 @@ .FORWARDHELPCATEGORY Cmdlet
...
+        if (Test-Path env:PAGER) {
+            $pager,$moreArgs = $env:PAGER -split '\s+'
+            $moreCommand = (Get-Command -CommandType Application $pager | Select-Object -First 1).Definition
+        } elseif ($IsWindows) {
+            $moreCommand = (Get-Command -CommandType Application more | Select-Object -First 1).Definition
+        } else {
+            $moreCommand = (Get-Command -CommandType Application less | Select-Object -First 1).Definition
+        }
+
+        $help | & $moreCommand $moreArgs $moreArgs
```
rkeithhill: IIRC someone (@BrucePay) had initialized `$moreArgs` to an empty string above the preceding if/else statements to prevent a strict mode failure ($moreArgs not being defined when $env:PAGER is not defined).

SteveL-MSFT: https://github.com/PowerShell/PowerShell/pull/6781

iSazonov: What about
```c#
        $pager,$moreArgs = $env:PAGER -split '\s+'
        if (Test-Path env:PAGER) {
            $moreCommand = (Get-Command -CommandType Application $pager | Select-Object -First 1).Definition
        } elseif ($IsWindows) {
            $moreCommand = (Get-Command -CommandType Application more | Select-Object -First 1).Definition
        } else {
            $moreCommand = (Get-Command -CommandType Application less | Select-Object -First 1).Definition
        }

        $help | & $moreCommand $moreArgs $moreArgs
```

rkeithhill: You have `$moreArgs` listed twice after the invocation of `$moreCommand`.  That should only be one.

iSazonov: I updated my post.

rkeithhill: Also, watch out for `$pager != $null`.  That should be `-ne`.  That said, this line:
```
$pager,$moreArgs = $env:PAGER -split '\s+'
```
Will assign an empty string, **not** $Null, to $pager when `$env:PAGER` is not defined.  So I think the test should be simply: `if ($pager)`.   And ideally, if `get-command` can't find the app referred to by $env:PAGER, it should revert back to the default for the OS.  Otherwise, you left with a $null in $moreCommand and `help | & $null` will fail.

iSazonov: Why are we doing `Get-Command -CommandType Application `? We get the same error for `badapp`, `& badapp` and `Get-Command -CommandType Application badapp`.

```c#
        $pager,$moreArgs = $env:PAGER -split '\s+'
        if ($pager) {
            $moreCommand = $pager
        } elseif ($IsWindows) {
            $moreCommand = "more.exe"
        } else {
            $moreCommand = "less"
        }

        $help | & $moreCommand $moreArgs

```

daxian-dbw: > You have $moreArgs listed twice after the invocation of $moreCommand. There should only be one.

Not only in your post, that's also the case in your code change.

iSazonov:
Add a test for Strict Mode from #6781.

test/powershell/engine/Help/HelpSystem.Tests.ps1
```diff
@@ -48,6 +48,20 @@ Describe "Validate that <pshome>/<culture>/default.help.txt is present" -Tags @(
...
+
+    It "Help doesn't fail when strict mode is on" {
+
+        $help = & {
+            # run in nested scope to keep strict mode from affecting other tests
+            Set-StrictMode -Version Latest
+            Help
+        }
+        # the help function renders the help content as text so just verify that there is content
+        $help | Should -Not -BeNullOrEmpty $true
```
daxian-dbw: `$true` is not needed. See https://github.com/PowerShell/PowerShell/pull/6781#discussion_r185292977

iSazonov: Fixed.

daxian-dbw (approved):

rkeithhill (approved):
LGTM

iSazonov:
@rkeithhill Thank you that you caught the lost commits! My mistake was that I did not rebase the old PR.

PR #6063, opened 2018-01-29
Add Environment Variable override of telemetry

lucyllewy:
## PR Summary

Inspired by dotnet cli, and allows for Telemetry optout on packaing systems such as AppImage and Snapcraft where the filesystem is immutable.

## PR Checklist

    - [x] Issue filed - Issue link: https://github.com/PowerShell/PowerShell-Docs/issues/2088

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/Telemetry.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/Telemetry.cs
@@ -73,14 +73,14 @@ namespace Microsoft.PowerShell
             try
             {
                 // if the semaphore file exists, try to send telemetry
-                var Enabled = Utils.NativeFileExists(TelemetrySemaphoreFilePath) && !GetEnvironmentVariableAsBool(TelemetryOptoutEnvVar, false);
+                var enabled = Utils.NativeFileExists(TelemetrySemaphoreFilePath) && !GetEnvironmentVariableAsBool(TelemetryOptoutEnvVar, false);
 
-                if ( ! Enabled )
+                if (!enabled)
                 {
                     return;
                 }
 
-                if ( _telemetryClient == null )
+                if (_telemetryClient == null)
                 {
                     _telemetryClient = new TelemetryClient();
                 }
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/Telemetry.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/Telemetry.cs
@@ -21,7 +21,7 @@ namespace Microsoft.PowerShell
         // The name of the file by when present in $PSHOME will enable telemetry.
         // If this file is not present, no telemetry will be sent.
         private const string TelemetrySemaphoreFilename = "DELETE_ME_TO_DISABLE_CONSOLEHOST_TELEMETRY";
-        private const string TelemetryOptout = "POWERSHELL_TELEMETRY_OPTOUT";
+        private const string TelemetryOptoutEnvVar = "POWERSHELL_TELEMETRY_OPTOUT";
 
         // The path to the semaphore file which enables telemetry
         private static string TelemetrySemaphoreFilePath = Path.Combine(
@@ -73,7 +73,7 @@ namespace Microsoft.PowerShell
             try
             {
                 // if the semaphore file exists, try to send telemetry
-                var Enabled = Utils.NativeFileExists(TelemetrySemaphoreFilePath) && !GetEnvironmentVariableAsBool(TelemetryOptout, false);
+                var Enabled = Utils.NativeFileExists(TelemetrySemaphoreFilePath) && !GetEnvironmentVariableAsBool(TelemetryOptoutEnvVar, false);
 
                 if ( ! Enabled )
                 {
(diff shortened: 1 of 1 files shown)


src/Microsoft.PowerShell.ConsoleHost/host/msh/Telemetry.cs
```diff
@@ -21,6 +21,7 @@ internal static class ApplicationInsightsTelemetry
         // The name of the file by when present in $PSHOME will enable telemetry.
         // If this file is not present, no telemetry will be sent.
         private const string TelemetrySemaphoreFilename = "DELETE_ME_TO_DISABLE_CONSOLEHOST_TELEMETRY";
+        private const string TelemetryOptout = "POWERSHELL_TELEMETRY_OPTOUT";
```
iSazonov: Need rename the file in root repo folder.

TravisEz13: This is used for the environment variable not the file

SteveL-MSFT: Perhaps it should be called TelemetryOptoutEnvVar

diddledan: See a54f7dc where I have renamed the constant as you proposed.

TravisEz13:
@diddledan updated the PR

src/Microsoft.PowerShell.ConsoleHost/host/msh/Telemetry.cs
```diff
@@ -50,14 +73,18 @@ private static void SendTelemetry(string eventName, Dictionary<string,string>pay
             try
             {
                 // if the semaphore file exists, try to send telemetry
-                if (Utils.NativeFileExists(TelemetrySemaphoreFilePath))
+                var Enabled = Utils.NativeFileExists(TelemetrySemaphoreFilePath) && !GetEnvironmentVariableAsBool(TelemetryOptoutEnvVar, false);
```
SteveL-MSFT: local vars should be camelCased

src/Microsoft.PowerShell.ConsoleHost/host/msh/Telemetry.cs
```diff
@@ -50,14 +73,18 @@ private static void SendTelemetry(string eventName, Dictionary<string,string>pay
...
+
+                if ( ! Enabled )
```
SteveL-MSFT: for consistency (like line 76 above), remove the whitespace between `!` and `enabled`

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.ConsoleHost/host/msh/Telemetry.cs
diddledan: oops, sorry about that one :-) I think I might have accidentally copied the style from somewhere else. I'll add another commit shortly to correct it.

src/Microsoft.PowerShell.ConsoleHost/host/msh/Telemetry.cs
diddledan: haha, that is the WordPress coding style.. Polyglots: knowledge of many things, master of none. I'll fold that into the same commit as the camelCased `Enabled` variable.

diddledan:
I've updated the PR again

src/Microsoft.PowerShell.ConsoleHost/host/msh/Telemetry.cs
```diff
@@ -42,6 +43,28 @@ static ApplicationInsightsTelemetry()
             TelemetryConfiguration.Active.TelemetryChannel.DeveloperMode = _developerMode;
         }
 
+        internal static bool GetEnvironmentVariableAsBool(string name, bool defaultValue) {
```
SteveL-MSFT: is there a reason for `internal` instead of `private`?  if you think this would be generally useful, perhaps this method should be `psutils.cs` instead of this file.  I would be ok with just making this private for now and in the future if this method is used elsewhere we can move it.

SteveL-MSFT (changes requested):
Thanks for the contribution.  Sorry about the stream of feedback, been juggling multiple things today.

src/Microsoft.PowerShell.ConsoleHost/host/msh/Telemetry.cs
diddledan: To be honest, I think the `internal` came from dotnet-cli which is where I got the inspiration for this function. I have no rationale of choosing one over the other so I am happy with whichever you prefer.

SteveL-MSFT: I would suggest leaving it here and changing to `private`.

diddledan:
I've updated the PR.

SteveL-MSFT (approved):

TravisEz13:
This is user-facing.  filed a doc issue and updated the template

TravisEz13:
filed https://github.com/PowerShell/PowerShell/issues/6078 to evaluated if we should keep the file opt-out.

Issue #6068, opened 2018-01-30
Switch Tests to Use 127.0.0.1 Instead of Localhost


markekraus:
due to dotnet/corefx#24104 using `localhost` as the hostname in test URL results in significant slowdown of `HttpClient`. We should switch the tests to use `127.0.0.1` for the hostname instead to decrease testing time of web cmdlets.

We should also look at other tests that are using localhost and consider switching the ones we can to 127.0.0.1.

markekraus:
Resolved by #6069

PR #6069, opened 2018-01-30
Change Web Cmdlet Tests to Use 127.0.0.1 not Localhost

markekraus:
## PR Summary

* Change Web Cmdlets Tests to Use 127.0.0.1 instead of localhost due to dotnet/corefx#24104 
* This provides a decent speed boost to the WebCmdlet tests.

## PR Checklist

### DIFF
--- a/test/tools/Modules/WebListener/WebListener.psm1
+++ b/test/tools/Modules/WebListener/WebListener.psm1
@@ -158,6 +158,7 @@ function Get-WebListenerUrl {
             return $null
         }
         $Uri = [System.UriBuilder]::new()
+        # Use 127.0.0.1 and not localhost due to https://github.com/dotnet/corefx/issues/24104
         $Uri.Host = '127.0.0.1'
         $Uri.Port = $runningListener.HttpPort
         $Uri.Scheme = 'Http'
--- a/test/tools/Modules/WebListener/WebListener.psm1
+++ b/test/tools/Modules/WebListener/WebListener.psm1
@@ -158,7 +158,7 @@ function Get-WebListenerUrl {
             return $null
         }
         $Uri = [System.UriBuilder]::new()
-        $Uri.Host = 'localhost'
+        $Uri.Host = '127.0.0.1'
         $Uri.Port = $runningListener.HttpPort
         $Uri.Scheme = 'Http'
 
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -496,15 +496,15 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
     It "Validate Invoke-WebRequest error with -Proxy and -NoProxy option" {
         $uri = Get-WebListenerUrl -Test 'Delay' -TestValue '10'
-        $command = "Invoke-WebRequest -Uri '$uri' -Proxy 'http://localhost:8080' -NoProxy -TimeoutSec 2"
+        $command = "Invoke-WebRequest -Uri '$uri' -Proxy 'http://127.0.0.1:8080' -NoProxy -TimeoutSec 2"
 
         $result = ExecuteWebCommand -command $command
         $result.Error.FullyQualifiedErrorId | Should Be "AmbiguousParameterSet,Microsoft.PowerShell.Commands.InvokeWebRequestCommand"
     }
 
     $testCase = @(
-        @{ proxy_address = "http://localhost:9"; name = 'http_proxy'; protocol = 'http' }
-        @{ proxy_address = "http://localhost:9"; name = 'https_proxy'; protocol = 'https' }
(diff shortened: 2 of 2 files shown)


SteveL-MSFT:
@markekraus reading the dotnet thread, it seems that this is a side effect of DNS and not specifically httpclient.  If so, should we globally change all tests using `localhost` to `127.0.0.1`?

markekraus:
Well it depends on how you look at it.. DNS and the IPv6 vs IPv4 certainly plays a key role... but the fact that ,NET Core is orders of magnitude slower than .NET Framework at dealing with the exact same name resolution dilemma.. I'd say that is a .NET Core problem.... pretty much al HTTP APIs are wrapping the same low level API as HttpClient and some of the others (such as WebRequest) are wrapping HttpClient... anyway semantics aside...

Are there other tests outside of the Web Cmdlets that are accessing something on localhost? If they are and they are using any of the HTTP APIs then they could be converted to 127.0.0.1 for some performance gain. I wasn't aware of any such tests.

It's really important for the Web Client tests because we are breaking 5 seconds for redirect tests and the sheer volume of calls made to localhost really add up the test time.

SteveL-MSFT:
@markekraus I simply did a search for `localhost` in the test folder and had many hits.  The WSMan configuration ones should stay localhost as that's what the config path is called.  Specially for HTTP, I see some proxy tests that explicitly use `localhost`.

markekraus:
@SteveL-MSFT  The proxy tests in the Web Cmdlets are probably unaffected. I believe they are all immediate errors. I'll give the a second look, but last I checked they never get to the point where an HTTP call is made. I'll look at the other tests as well.

test/tools/Modules/WebListener/WebListener.psm1
```diff
@@ -158,7 +158,7 @@ function Get-WebListenerUrl {
             return $null
         }
         $Uri = [System.UriBuilder]::new()
-        $Uri.Host = 'localhost'
+        $Uri.Host = '127.0.0.1'
```
iSazonov: Could you please add comment and link on CoreFX Issue?

iSazonov:
Other PRs is experiencing problems with these web tests. I suggest fast merge the PR and continue explorations in next PR.

markekraus:
@iSazonov #6055 will also stop the false positives. the -TimeOutSec needs to be removed from tests not related to -TimeOutSec anyway because of the potential for a regression to be missed due to feature clash. That PR is ready.

test/tools/Modules/WebListener/WebListener.psm1
markekraus: done.

test/tools/Modules/WebListener/WebListener.psm1
```diff
@@ -158,7 +158,8 @@ function Get-WebListenerUrl {
             return $null
         }
         $Uri = [System.UriBuilder]::new()
-        $Uri.Host = 'localhost'
+        # Use 127.0.0.1 and not localhost due to https://github.com/dotnet/corefx/issues/24104
```
iSazonov: I'd add "TODO: remove after fix https://github.com/dotnet/corefx/issues/24104"

iSazonov (commented):
LGTM with one minor comment

test/tools/Modules/WebListener/WebListener.psm1
markekraus: why would we ever change it back? we are binding WebListener to 127.0.0.1, not to 'localhost'. it's more correct to keep it at 127.0.0.1.

iSazonov: Good question! 😄 It seems IPv4 never die.

iSazonov (approved):

markekraus:
I changed #6068 to include looking at localhost usage in all tests, not just Web Cmdlets. We can merge this PR and cover the rest which are not as numerous nor show stopping as the ones fixed by this PR.

test/tools/Modules/WebListener/WebListener.psm1
TravisEz13: At least we will have to make significant changes to the tests if we get rid of IPv4.  I was thinking that perhaps we should define a function to get the `value` for localhost.  Then, we only have one place to change, but this isn't a blocking comment.

TravisEz13 (approved):

test/tools/Modules/WebListener/WebListener.psm1
markekraus: It could be done. for the most part.. the web cmdlets already do that since they get the url's from `Get-WebListenerUrl` the web listener module would just need to call whatever function we share with the rest of the tests. 

The rest of the tests in the project may not even benefit from this change. I did a quick look this morning and they were mostly `\\localhost\` (I'm not sure SMB is affected, the CoreFX issue seems to indicate it's specific to HTTP, some of those tests may need localhost regardless) or `WSMan:\localhost\` (which wouldn't change).

iSazonov: In fact it's not worth.

iSazonov:
@markekraus Thanks for the temporary fix!

We definitely need performance/benchmark tests.

PR #6109, opened 2018-02-06
Make UTF-8 Default Encoding for application/json

### LINKED ISSUE #5530: WebCmdlets should read the encoding of Content-Type application/json per RFC
Steps to reproduce
------------------

```powershell
Invoke-RestMethod 'http://api.forismatic.com/api/1.0/?method=getQuote&format=json&lang=ru' -verbose
```

Expected behavior
-----------------

It should **detect** the utf-8 encoding, and produce the same output as this:

```posh
$resp = Invoke-WebRequest 'http://api.forismatic.com/api/1.0/?method=getQuote&format=json&lang=ru'
$char = $resp.RawContentStream.ToArray()
$str = [Text.Encoding]::UTF8.GetString($char)
ConvertFrom-Json $str
```

i.e. something like this:

```none
VERBOSE: GET http://api.forismatic.com/api/1.0/?method=getQuote&format=json with 0-byte payload
VERBOSE: received 536-byte response of content type application/json
VERBOSE: Content encoding: utf-8

quoteText   : Именно внутренний диалог прижимает к земле людей в повседневной жизни. Мир для нас такой-то и такой-то или этакий и этакий лишь потому, что мы сами себе говорим о нем, что он такой-то и такой-то или этакий и этакий.
quoteAuthor : Карлос Кастанеда
senderName  :
senderLink  :
quoteLink   : http://forismatic.com/ru/6309006412/
```

Actual behavior
---------------

It falls back to iso-8859-1 encoding and produces gobbledygook with a lot of Ð's in it.  Also, it utterly fails to produce a number for the `#-byte response` string in the verbose output.

```none
VERBOSE: GET http://api.forismatic.com/api/1.0/?method=getQuote&format=json with 0-byte payload
VERBOSE: received -byte response of content type application/json
VERBOSE: Content encoding: iso-8859-1

quoteText   : ÐÐ¾ Ð²ÑÑÐºÐ¾Ð¼Ñ Ð¿ÑÐ¸Ð±ÐµÐ¶Ð¸ÑÑ Ð¾Ð±ÑÐ°ÑÐ°ÑÑÑÑ Ð»ÑÐ´Ð¸, Ð¼ÑÑÐ¸Ð¼ÑÐµ ÑÑÑÐ°ÑÐ¾Ð¼: Ðº Ð³Ð¾ÑÐ°Ð¼
              Ð¸ Ðº Ð»ÐµÑÐ°Ð¼, Ðº Ð´ÐµÑÐµÐ²ÑÑÐ¼ Ð² ÑÐ¾ÑÐµ, Ðº Ð³ÑÐ¾Ð±Ð½Ð¸ÑÐ°Ð¼.
quoteAuthor : ÐÑÐ´Ð´Ð° ÐÐ°ÑÑÐ°Ð¼Ð°
senderName  :
senderLink  :
quoteLink   : http://forismatic.com/ru/804c7d14d9/
```

Discussion
----------

When calling an HTTP endpoint that returns a header: `Content-Type: application/json` the WebCmdlets are **incorrectly** defaulting to `iso-8859-1` rather than a proper unicode encoding, and are *disregarding* the application/json RFC's simple specification for how to determine the content encoding.

1. The JSON standard [ECMA-404 (PDF)](http://ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf) clearly states that JSON must be unicode
2. The [application/json RFC (in section 3)](https://tools.ietf.org/html/rfc4627#section-3) clearly indicates how the encoding should be determined from the first 4 bytes of the content.

**NOTE:** Please don't work around this by just defaulting to utf-8.  I'm sure that 90% of the time, you could probably get away with that, but it's not actually correct, and the RFC implementation is _trivial_.

**ALSO NOTE:** The WebCmdlets *do* respect the [`; charset=utf-8`](https://github.com/PowerShell/PowerShell/blob/f5f3fab1ea8a51dd2e5d34e0e580ab39176eb696/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/StreamHelper.cs#L442) attribute if it's present on the content-type header -- which makes sense, but isn't technically standards compliant for an `application/*` content-type, as far as I can tell.

To get started: [`ProcessResponse`](https://github.com/PowerShell/PowerShell/blob/f5f3fab1ea8a51dd2e5d34e0e580ab39176eb696/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/CoreCLR/InvokeRestMethodCommand.CoreClr.cs#L54) and [`TryGetEncoding`](https://github.com/PowerShell/PowerShell/blob/f5f3fab1ea8a51dd2e5d34e0e580ab39176eb696/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/StreamHelper.cs#L411) 

See also #5528 which was a specific instance of this problem. @lipkau was incorrectly convinced by early responders that the problem was in the webserver, but it's actually in PowerShell's cmdlets. If you invoke the rest API against the Atlassian wiki, you can see the problem happening in the Verbose stream:

```posh
$r = IRM $url -Credential $mycred -Authentication basic -Verbose
VERBOSE: GET https://powershell.atlassian.net/wiki/rest/api/content/13009245?expand=space,version with 0-byte payload
VERBOSE: received -byte response of content type application/json
VERBOSE: Content encoding: iso-8859-1
```

The content is **actually** correctly utf-8 encoded (as you could tell from the positions of the nulls in the first 4 bytes), and iso-8859-1 is _**never** a valid encoding for application/json_, period.

```plain
PS C:\Program Files\PowerShell\6.0.0-rc> $PSVersionTable

```

markekraus:
## PR Summary

When a charset is not supplied for a JSON response, the default encoding should be UTF-8 per RFC 8259. This PR changes the default charset to UTF-8 for JSON responses when a charset is not defined.

I'm not sure if this is a breaking change or not. If someone was running into encoding issues, they would already be converting `BasicHtmlWebResponseObject.RawContentStream.` If they were unaware, then this would just correct characters in their strings.

Closes #5530

## PR Checklist

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
@@ -993,6 +993,22 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
             $response.Output.Encoding.EncodingName | Should Be $expectedEncoding.EncodingName
             $response.Output | Should BeOfType 'Microsoft.PowerShell.Commands.BasicHtmlWebResponseObject'
         }
+
+        It "Verifies Invoke-WebRequest defaults to UTF8 on application/json when no charset is present" {
+            # when contenttype is set, WebListener suppresses charset unless it is included in the query
+            $query = @{
+                contenttype = 'application/json'
+                body        = '{"Test": "Test"}'
+            }
+            $uri = Get-WebListenerUrl -Test 'Response' -Query $query
+            $expectedEncoding = [System.Text.Encoding]::UTF8
+            $response = ExecuteWebRequest -Uri $uri -UseBasicParsing
+
+            $response.Error | Should BeNullOrEmpty
+            $response.Output.Encoding.EncodingName | Should Be $expectedEncoding.EncodingName
+            $response.Output | Should BeOfType 'Microsoft.PowerShell.Commands.BasicHtmlWebResponseObject'
+            $response.Output.Content | Should BeExactly $query.body
+        }
     }
 
     #endregion charset encoding tests
@@ -2302,6 +2318,21 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
 
             $response.Error | Should BeNullOrEmpty
             $response.Encoding.EncodingName | Should Be $expectedEncoding.EncodingName
+        }
+
+        It "Verifies Invoke-RestMethod defaults to UTF8 on application/json when no charset is present" {
+            # when contenttype is set, WebListener suppresses charset unless it is included in the query
+            $query = @{
+                contenttype = 'application/json'
+                body        = '{"Test": "Test"}'
+            }
+            $uri = Get-WebListenerUrl -Test 'Response' -Query $query
+            $expectedEncoding = [System.Text.Encoding]::UTF8
+            $response = ExecuteRestMethod -Uri $uri -UseBasicParsing
+
+            $response.Error | Should BeNullOrEmpty
+            $response.Encoding.EncodingName | Should Be $expectedEncoding.EncodingName
+            $response.output.Test | Should BeExactly 'Test'
         }
     }
 
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/BasicHtmlWebResponseObject.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/BasicHtmlWebResponseObject.Common.cs
@@ -166,6 +166,12 @@ namespace Microsoft.PowerShell.Commands
                 Encoding encoding = null;
                 // fill the Content buffer
                 string characterSet = WebResponseHelper.GetCharacterSet(BaseResponse);
+
+                if (String.IsNullOrEmpty(characterSet) && ContentHelper.IsJson(contentType))
+                {
+                    characterSet = Encoding.UTF8.HeaderName;
+                }
+
                 this.Content = StreamHelper.DecodeStream(RawContentStream, characterSet, out encoding);
                 this.Encoding = encoding;
             }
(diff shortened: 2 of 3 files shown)


JamesWTruher (approved):

dantraMSFT (approved):

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -993,6 +993,22 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
             $response.Output.Encoding.EncodingName | Should Be $expectedEncoding.EncodingName
             $response.Output | Should BeOfType 'Microsoft.PowerShell.Commands.BasicHtmlWebResponseObject'
         }
+
+        It "Verifies Invoke-WebRequest defaults to UTF8 on application/json when no charset is present" {
+            # when contenttype is set, WebListener suppresses charset unless it is included in the query
+            $query = @{
+                contenttype = 'application/json'
+                body        = '{"Test": "Test"}'
```
iSazonov: Should we add multibyte characters to ensure Utf8 encoding works?

iSazonov:
I believe we should add the default in documentation.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
markekraus: No need. We are not testing the ability of .NET to properly decode UTF-8. We are testing the ability of the web cmdlets to chose the correct default encoding for application/json. This is an acceptable tests because we already have tests for the `iso-8859-1` defaults for other mime type and we have UTF-8 tests which do ensure we can work with UTF-8 characters. Here we only need to make sure the UTF-8 encoding was selected.

iSazonov:
Last question - is this a breaking change or not?

markekraus:
@iSazonov I need someone more experienced to make that call. Normally, the changes I deal with are more black-and-white on being breaking changes. This one is a bit of grey. This certainly changes the public behavior of the cmdlets. However, as I said the the description anyone who ran into this behavior is likely using a work around that would be unaffected by the change. The only user story I can come up with where this is a breaking change is where someone was purposely using the broken encoding... that's not a compelling story to me.

markekraus:
@iSazonov should with have the PowerShell committee review it then?

iSazonov:
Let's ask @SteveL-MSFT

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this an ok with having UTF-8 as default and marked as breaking change

iSazonov:
@markekraus Thanks for the change!

iSazonov:
@daxian-dbw Could you please rename the commit with [breaking change]? I haven't permission to correct the lost.

PR #6113, opened 2018-02-06
Add Password parameter to Get-PfxCertificate cmdlet

### LINKED ISSUE #3970: Get-PFXCertificate does not support -Password like Import-PFXCertificate
Get-PFXCertificate doesn't have a -Password param like Import-PFXCertificate. So when I try to import a password protected pfx, it _prompts_ for a password. It would be better if we could provide a password to it so we could use it in non-interactive code. 

It looks like [here](https://github.com/PowerShell/PowerShell/blob/7a55bf98b2370ab4817ab2533cb67673053ee446/src/Microsoft.PowerShell.Security/security/CertificateCommands.cs#L153) it is doing the prompt

I am hoping that we can get functionality like [Import-PFXCertificate -Password](https://technet.microsoft.com/en-us/itpro/powershell/windows/pkiclient/import-pfxcertificate#-password) on [Get-PFXCertificate](https://msdn.microsoft.com/en-us/powershell/reference/5.1/microsoft.powershell.security/get-pfxcertificate)

As a work around I have been able to do 
```powershell
$certFile    = '\\server\share\pwd-protected.pfx'
$certPass = Read-Host -AsSecureString -Prompt 'Enter the password for certificate: '

$PFXCert =  Get-PfxData -FilePath $certFile -Password $certPass
$PFXCert.EndEntityCertificates.ThumbPrint
```

Steps to reproduce
------------------

```powershell
$certFile    = '\\server\share\pwd-protected.pfx'
$certPass = Read-Host -AsSecureString -Prompt 'Enter the password for certificate: '

#Will prompt for password
$certThumbPrint = (Get-PfxCertificate -FilePath $certFile ).ThumbPrint

#Works fine, but imports the cert, which one might not want to do. 
$PFXCert = Import-PfxCertificate -FilePath $certFile -Password $certPass -CertStoreLocation Cert:\LocalMachine\My 

```

Expected behavior
-----------------

```powershell
$certFile    = '\\server\share\pwd-protected.pfx'
$certPass = Read-Host -AsSecureString -Prompt 'Enter the password for certificate: '

$certThumbPrint = (Get-PfxCertificate -FilePath $certFile  -Password $certPass ).ThumbPrint


```

Actual behavior
---------------

```powershell
$certFile    = '\\server\share\pwd-protected.pfx'
$certPass = Read-Host -AsSecureString -Prompt 'Enter the password for certificate: '

$certThumbPrint = (Get-PfxCertificate -FilePath $certFile  -Password $certPass ).ThumbPrint

Get-PfxCertificate : A parameter cannot be found that matches parameter name 'Password'.
At line:1 char:60
+ ... umbPrint = (Get-PfxCertificate -FilePath $certFile  -Password $certPa ...
+                                                         ~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Get-PfxCertificate], ParameterBindingException
    + FullyQualifiedErrorId : NamedParameterNotFound,Microsoft.PowerShell.Commands.GetPfxCertificateCommand
```

Environment data
----------------

```powershell
> $PSVersionTable
PSVersion                      5.1.14409.1005
BuildVersion                   10.0.14409.1005
```

maybe-hello-world:
## PR Summary

Add Password parameter to Get-PfxCertificate cmdlet to allow automatization instead of prompting for password every time.

Fix #3970

Possibly breaking change: 
Calling cmdlet without -Password parameter assumes passing empty password instead of prompting for pass as before. In order to avoid breaking change cmdlet can ask for password if Password is missed but it will be more complicated to use empty password string in automated scripts.

Also about tests:
How it's better to create password-protected PFX certificate to test not-null password? Store as raw bytes in certificateCommon.psm1?

## PR Checklist

    - [x] Issue filed - Issue link: PowerShell/PowerShell-Docs#2150

### DIFF
--- a/src/Microsoft.PowerShell.Security/security/CertificateCommands.cs
+++ b/src/Microsoft.PowerShell.Security/security/CertificateCommands.cs
@@ -74,7 +74,7 @@ namespace Microsoft.PowerShell.Commands
         /// Do not prompt for password if not given.
         /// </summary>
         [Parameter(Mandatory = false)]
-        public SwitchParameter NoPasswordPrompt { get; set; }
+        public SwitchParameter NoPromptForPassword { get; set; }
 
         //
         // list of files that were not found
@@ -142,7 +142,7 @@ namespace Microsoft.PowerShell.Commands
                     }
                     else
                     {
-                        if (Password == null && !NoPasswordPrompt.IsPresent) 
+                        if (Password == null && !NoPromptForPassword.IsPresent) 
                         {
                             try 
                             {
--- a/src/Microsoft.PowerShell.Security/security/CertificateCommands.cs
+++ b/src/Microsoft.PowerShell.Security/security/CertificateCommands.cs
@@ -144,18 +144,18 @@ namespace Microsoft.PowerShell.Commands
                     {
                         if (Password == null && !Force.IsPresent) 
                         {
-							try 
-							{
-								cert = GetCertFromPfxFile(resolvedProviderPath, null);
-								WriteObject(cert);
-								continue;
-							} 
-							catch (CryptographicException) 
-							{
+                            try 
+                            {
+                                cert = GetCertFromPfxFile(resolvedProviderPath, null);
+                                WriteObject(cert);
+                                continue;
+                            } 
+                            catch (CryptographicException) 
+                            {
                                 Password = SecurityUtils.PromptForSecureString(
                                     Host.UI,
                                     CertificateCommands.GetPfxCertPasswordPrompt);
-							}                            
+                            }                            
                         }
                         
                         try
--- a/src/Microsoft.PowerShell.Security/security/CertificateCommands.cs
+++ b/src/Microsoft.PowerShell.Security/security/CertificateCommands.cs
@@ -65,7 +65,7 @@ namespace Microsoft.PowerShell.Commands
         private bool _isLiteralPath = false;
 
         /// <summary>
-        /// Gets or sets the password for unlocking the certificate
+        /// Gets or sets the password for unlocking the certificate.
         /// </summary>
         [Parameter(Mandatory = false)]
         public SecureString Password { get; set; }
@@ -190,7 +190,6 @@ namespace Microsoft.PowerShell.Commands
 
         private static X509Certificate2 GetCertFromPfxFile(string path, SecureString password)
         {
-            // Passing null password and DefaultKeySet here is equivalent to calling "new X509Certificate2(path)"
             var cert = new X509Certificate2(path, password, X509KeyStorageFlags.DefaultKeySet);
             return cert;
         }
--- a/src/Microsoft.PowerShell.Security/security/CertificateCommands.cs
+++ b/src/Microsoft.PowerShell.Security/security/CertificateCommands.cs
@@ -64,6 +64,24 @@ namespace Microsoft.PowerShell.Commands
         }
         private bool _isLiteralPath = false;
 
+        /// <summary>
+        /// Gets or sets the password for unlocking the certificate
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        public SecureString Password
+        {
+            get
+            {
+                return _password;
+            }
+
+            set
+            {
+                _password = value;
+            }
+        }
+        private SecureString _password = null;
+
         //
         // list of files that were not found
         //
@@ -132,45 +150,17 @@ namespace Microsoft.PowerShell.Commands
                     {
                         try
                         {
-                            cert = GetCertFromPfxFile(resolvedProviderPath);
+                            cert = GetCertFromPfxFile(resolvedProviderPath, _password);
                         }
-                        catch (CryptographicException)
+                        catch (CryptographicException e)
                         {
-                            //
-                            // CryptographicException is thrown when any error
-                            // occurs inside the crypto class library. It has a
-                            // protected member HResult that indicates the exact
-                            // error but it is not available outside the class.
-                            // Thus we have to assume that the above exception
-                            // was thrown because the pfx file is password
-                            // protected.
-                            //
-                            SecureString password = null;
-
-                            string prompt = null;
-                            prompt = CertificateCommands.GetPfxCertPasswordPrompt;
-
-                            password = SecurityUtils.PromptForSecureString(Host.UI, prompt);
-                            try
-                            {
-                                cert = GetCertFromPfxFile(resolvedProviderPath,
-                                                          password);
-                            }
-                            catch (CryptographicException e)
-                            {
-                                //
-                                // since we cannot really figure out the
-                                // meaning of a given CryptographicException
-                                // we have to use NotSpecified category here
-                                //
-                                ErrorRecord er =
-                                    new ErrorRecord(e,
-                                                    "GetPfxCertificateUnknownCryptoError",
-                                                    ErrorCategory.NotSpecified,
-                                                    null);
-                                WriteError(er);
-                                continue;
-                            }
+                            ErrorRecord er =
+                                new ErrorRecord(e,
+                                                "GetPfxCertificateUnknownCryptoError",
+                                                ErrorCategory.NotSpecified,
+                                                null);
+                            WriteError(er);
+                            continue;
                         }
 
                         WriteObject(cert);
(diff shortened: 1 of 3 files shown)


src/Microsoft.PowerShell.Security/security/CertificateCommands.cs
```diff
@@ -64,6 +64,24 @@ public string[] LiteralPath
...
+            {
+                return _password;
+            }
+
+            set
+            {
+                _password = value;
+            }
+        }
+        private SecureString _password = null;
```
markekraus: Please use Auto-Implemented Property instead.

markekraus (changes requested):

src/Microsoft.PowerShell.Security/security/CertificateCommands.cs
```diff
@@ -64,6 +64,12 @@ public string[] LiteralPath
         }
         private bool _isLiteralPath = false;
 
+        /// <summary>
+        /// Gets or sets the password for unlocking the certificate
```
iSazonov: Please add final dot in the public comment.

src/Microsoft.PowerShell.Security/security/CertificateCommands.cs
```diff
@@ -210,14 +188,9 @@ protected override void ProcessRecord()
...
 
-        private static X509Certificate2 GetCertFromPfxFile(string path)
-        {
-            X509Certificate2 cert = new X509Certificate2(path);
-            return cert;
-        }
-
         private static X509Certificate2 GetCertFromPfxFile(string path, SecureString password)
         {
+            // Passing null password and DefaultKeySet here is equivalent to calling "new X509Certificate2(path)"
```
iSazonov: Please remove obvious comment.

src/Microsoft.PowerShell.Security/security/CertificateCommands.cs
```diff
@@ -132,45 +138,17 @@ protected override void ProcessRecord()
...
-                            // Thus we have to assume that the above exception
-                            // was thrown because the pfx file is password
-                            // protected.
-                            //
-                            SecureString password = null;
-
-                            string prompt = null;
-                            prompt = CertificateCommands.GetPfxCertPasswordPrompt;
-
-                            password = SecurityUtils.PromptForSecureString(Host.UI, prompt);
```
TravisEz13: We should not change the behavior to prompt when a password is not supplied.

TravisEz13 (changes requested):
see comment

TravisEz13:
This will need a user-facing documentation change, see the PR Template.

TravisEz13:
We need tests for this functionality.  At least:

1. A test verifying a password works
2. A test verifying that a bad supplied password fails and does not prompt.

And it would be nice if we could have a test that verified the host ui prompt scenario

SteveL-MSFT (changes requested):
I don't like the breaking change.  I think you can make `Password` nullable and if it's not set use the existing prompting code.

iSazonov:
We could ignoring UI prompt with `-Force` for best UX in scripts.

maybe-hello-world:
But isn't it an unnatural behaviour of cmdlet? PFX certificates (unlike pem) are designed to be password-protected by default, isn't it better to treat them in such way?

SteveL-MSFT:
@maybe-hello-world is your comment replying to @iSazonov?  I would agree that we should only support either `-Password` supplied or prompt for password.

maybe-hello-world:
@SteveL-MSFT No, it's more about situation when we get no password from user, try to open pfx cert and in case of exception ask user for password. IMHO in either/or situation it's better to use password parameter and treat it as null if it's missed because:

- It doesn't break existing automated scripts (where pfx isn't protected), scripts that works with protected pfx require interaction anyway
- Prompt can possibly make it more complex to use cmdlet with GUI that relies on Powershell 
- It's more intuitive to look for parameter then wait for prompt

But of course it's just IMHO =)

markekraus:
@maybe-hello-world 
> It's more intuitive to look for parameter then wait for prompt

Except that it has always prompted up to now. Interactive console scripting may rely on that prompt. Replacing the lack of password with an error breaks that flow. In fact. I have CLI that does just that: menu selections for a few PFXs, then something like this

```powershell
Write-Host "Unlock PFX"
$cert = $null
$attemptsLeft = 3
Do {
    $cert = Get-PfxCertificate -FilePath $pfxChoice -erroraction 'SilentlyContinue'
    $attemptsLeft--
    if (!$cert -and $attemptsLeft -gt 0) { 
        Write-Host "Wrong password. Try again (attempts left: $attemptsLeft)"
    }
} While ($null -eq $cert -and $attemptsLeft -gt 0)
````

I'm not sure how wide spread that kind of usage is, but I'm guessing if I use it, at least one other person may be. 🙂

**Comment by SteveL-MSFT** [comment]
@maybe-hello-world I do appreciate this contribution and if the existing behavior was not already there, I would perhaps agree with you, but it seems unnecessary to introduce a breaking change.

**Comment by maybe-hello-world** [comment]
@SteveL-MSFT Well, you're right. I've changed behaviour to non-breaking so could you review it again?

File: test/powershell/Modules/Microsoft.PowerShell.Security/certificateCommon.psm1 [hunk 5]
```diff
@@ -66,6 +66,46 @@ OksttXT1kXf+aez9EzDlsgQU4ck78h0WTy01zHLwSKNWK4wFFQM=
     return $certLocation
 }
 
+Function New-ProtectedCertificate
```
markekraus: We do have 2 password protected PFX in the project already

* https://github.com/PowerShell/PowerShell/blob/master/test/tools/Modules/WebListener/ClientCert.pfx
* https://github.com/PowerShell/PowerShell/blob/master/test/tools/Modules/WebListener/ServerCert.pfx

Maybe they could just be copied to the test drive or accessed directly?

markekraus (approved):
LGTM. My comment is non-blocking

src/Microsoft.PowerShell.Security/security/CertificateCommands.cs
```diff
@@ -64,6 +64,18 @@ public string[] LiteralPath
...
+        /// Gets or sets the password for unlocking the certificate.
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        public SecureString Password { get; set; }
+
+        /// <summary>
+        /// Do not prompt for password if not given.
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        public SwitchParameter Force { get; set; }
```
SteveL-MSFT: I'm not personally a fan of the overloaded `-Force` being the `make it happen` switch and prefer something more descriptive and specific like: `-NoPasswordPrompt`.  Let me bring it up at the next @PowerShell/powershell-committee meeting.

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Security/certificateCommon.psm1
maybe-hello-world: Well, but is it good to rely on WebListener's location and certs? If WebListener change in future you'll have to find what it'll break in rather unusual places like this.

markekraus: While that seems like a logical point.. that change would break the tests here and we would know about it. it's not like moving the certs would leave this tests here passing. Yes, we would have to update the tests at that time.  But WebListener is a part of the test suite so we might as well reuse what we can form it.

There is an open issue for consolidating all certs for all tests anyway as they exists in far to many places. My comment is an attempt to keep the complexity of certs in our tests at the current level without increasing it further.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and agree for the need of a switch to suppress prompting for password and recommends the name of the switch to be "-NoPromptForPassword"

iSazonov:
@maybe-hello-world Could you please continue?

iSazonov:
@SteveL-MSFT @TravisEz13 Please continue the review.

test/powershell/Modules/Microsoft.PowerShell.Security/certificateCommon.psm1
SteveL-MSFT: I agree that we should try to have the test certs used commonly rather than having different ones.  However, I would be fine having an issue open to refactor this as a separate PR.

SteveL-MSFT (approved):
My concern with the switch name has been addressed

TravisEz13 (approved):

PR #6144, opened 2018-02-10
Enable $env:PAGER to work correctly if arguments are used

### LINKED ISSUE #6142: Invalid handling of the PAGER environment variable
Hi there,

I'm new to PowerShell and wanted to try it out but it quickly failed when I tried to use `Get-Help`.
I think this is related to #798 but I actually don't have any clue on how to fix so please bear with me (well, it looks like `| pwsh -c "$env:PAGER"` could work but that seems too awful to be the proper solution) :)

Steps to reproduce
------------------

```powershell
> $env:PAGER
less -r
```

Expected behavior
-----------------

I would have expected the behavior could conform to http://pubs.opengroup.org/onlinepubs/9699919799/utilities/man.html:
```
ENVIRONMENT VARIABLES
  PAGER
    Determine an output filtering command for writing the output to a terminal.
    Any string acceptable as a command_string operand to the sh -c command shall be valid.
    When standard output is a terminal device, the reference page output shall be piped through the command.
    If the PAGER variable is null or not set, the command shall be either more or another paginator utility documented in the system documentation.
```

Actual behavior
---------------

```powershell
> help help
Get-Command : The term 'less -r' is not recognized as the name of a cmdlet, function, script file, or operable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:12 char:21
+ ... oreCommand = (Get-Command -CommandType Application $env:PAGER | Selec ...
+                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : ObjectNotFound: (less -r:String) [Get-Command], CommandNotFoundException
+ FullyQualifiedErrorId : CommandNotFoundException,Microsoft.PowerShell.Commands.GetCommandCommand

The expression after '&' in a pipeline element produced an object that was not valid. It must result in a command name, a script block, or a CommandInfo object.
At line:23 char:21
+ } else { $input | & $moreCommand }
+                     ~~~~~~~~~~~~
+ FullyQualifiedErrorId : BadExpression
```

Environment data
----------------

```powershell
> $PSVersionTable

OS                             Linux 4.14.0-3-amd64 #1 SMP Debian 4.14.12-2 (2018-01-06)
```

Regards.

SteveL-MSFT:
## PR Summary

If `$env:PAGER` is a command with args, it doesn't work correctly as the current code tries to find a command matching `$env:PAGER` which fails due to the args.  Fix is to split `$env:PAGER` into the command and the args, find the first occurrence of the command and pass back the args.

No tests as this is interactive only.  Manually validated with `less -w` which highlights the first lien of a new page.

Fix https://github.com/PowerShell/PowerShell/issues/6142

## PR Checklist

### DIFF
--- a/src/System.Management.Automation/engine/InitialSessionState.cs
+++ b/src/System.Management.Automation/engine/InitialSessionState.cs
@@ -4764,7 +4764,7 @@ $OutputEncoding = if ([System.Management.Automation.Platform]::IsNanoServer -or 
 
 # Respect PAGER, use more on Windows, and use less on Linux
 if (Test-Path env:PAGER) {
-    $pager,$moreArgs = $env:PAGER -split ' ',2
+    $pager,$moreArgs = $env:PAGER -split '\s+'
     $moreCommand = (Get-Command -CommandType Application $pager | Select-Object -First 1).Definition
 } elseif ($IsWindows) {
     $moreCommand = (Get-Command -CommandType Application more | Select-Object -First 1).Definition
--- a/src/System.Management.Automation/engine/InitialSessionState.cs
+++ b/src/System.Management.Automation/engine/InitialSessionState.cs
@@ -4764,7 +4764,8 @@ $OutputEncoding = if ([System.Management.Automation.Platform]::IsNanoServer -or 
 
 # Respect PAGER, use more on Windows, and use less on Linux
 if (Test-Path env:PAGER) {
-    $moreCommand = (Get-Command -CommandType Application $env:PAGER | Select-Object -First 1).Definition
+    $pager,$moreArgs = $env:PAGER -split ' ',2
+    $moreCommand = (Get-Command -CommandType Application $pager | Select-Object -First 1).Definition
 } elseif ($IsWindows) {
     $moreCommand = (Get-Command -CommandType Application more | Select-Object -First 1).Definition
 } else {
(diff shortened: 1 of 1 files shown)


src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -4764,7 +4764,8 @@ internal static SessionStateAliasEntry[] BuiltInAliases
 
 # Respect PAGER, use more on Windows, and use less on Linux
 if (Test-Path env:PAGER) {
-    $moreCommand = (Get-Command -CommandType Application $env:PAGER | Select-Object -First 1).Definition
+    $pager,$moreArgs = $env:PAGER -split ' ',2
```
lzybkr: If there are multiple arguments, this fix would put quotes around the single value in `$moreArgs`.

For a quick fix, this should probably be `... -split '\s+'`.

By removing the limit, `$moreArgs` will be either a string or array, and passing an array to a native command will put spaces between the args. And you should support multiple spaces between args.

Note that this code is still wrong for quoted arguments. If that mattered, you would use the PowerShell parser to parse the command line and extract arguments from the ast.

SteveL-MSFT: @lzybkr for now, I'm going with the quick fix which should address the majority of usage.  We can revisit support for quoted args if it impacts someone.

lzybkr (approved):

iSazonov:
@anmenaga @daxian-dbw Have you any thoughts or change requests?

anmenaga (approved):

PR #6182, opened 2018-02-18
Throw terminating error in New-TemporaryFile and make it not rely on the presence of the 'TEMP' environment variable

### LINKED ISSUE #4634: New-TemporaryFile should throw terminating error
As part of PR 4612, it was discussed [here](https://github.com/PowerShell/PowerShell/pull/4612#discussion_r134333394) that New-TemporaryFile should throw a terminating error (instead of a non-terminating one). Everyone was clear on the expected behaviour but the issue is that this change would be a breaking change in behaviour.

bergmeister:
## PR Summary

- Fixes #4634 by throwing a terminating error as agreed
- Makes the command not rely on the presence of the `TEMP` environment variable to get path to temp directory and use the .Net method `Path.GetTempPath()` instead
- Catch exception more specific as given by the documentation
- Improve existing test

## PR Checklist

- [ ] User facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [ ] [Change is not breaking](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes) Calls to `New-TemporaryFile` that are in a pipeline will abort (by design)
- [ ] [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting) I have tested it locally but we would be rather testing .Net and test setup would be expensive because one needs to first create 65535  temp files

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
@@ -10,7 +10,7 @@ namespace Microsoft.PowerShell.Commands
     /// <summary>
     /// The implementation of the "New-TemporaryFile" cmdlet
     /// </summary>
-    [Cmdlet(VerbsCommon.New, "TemporaryFile", SupportsShouldProcess = true, HelpUri = "https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/new-temporaryfile")]
+    [Cmdlet(VerbsCommon.New, "TemporaryFile", SupportsShouldProcess = true, HelpUri = "https://docs.microsoft.com/powershell/module/microsoft.powershell.utility/new-temporaryfile")]
     [OutputType(typeof(System.IO.FileInfo))]
     public class NewTemporaryFileCommand : Cmdlet
     {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
@@ -10,7 +10,7 @@ namespace Microsoft.PowerShell.Commands
     /// <summary>
     /// The implementation of the "New-TemporaryFile" cmdlet
     /// </summary>
-    [Cmdlet(VerbsCommon.New, "TemporaryFile", SupportsShouldProcess = true, HelpUri = "https://docs.microsoft.com/powershell/module/microsoft.powershell.utility/new-temporaryfile")]
+    [Cmdlet(VerbsCommon.New, "TemporaryFile", SupportsShouldProcess = true, HelpUri = "https://go.microsoft.com/fwlink/?LinkId=526726")]
     [OutputType(typeof(System.IO.FileInfo))]
     public class NewTemporaryFileCommand : Cmdlet
     {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
@@ -10,7 +10,7 @@ namespace Microsoft.PowerShell.Commands
     /// <summary>
     /// The implementation of the "New-TemporaryFile" cmdlet
     /// </summary>
-    [Cmdlet(VerbsCommon.New, "TemporaryFile", SupportsShouldProcess = true, HelpUri = "https://go.microsoft.com/fwlink/?LinkId=526726")]
+    [Cmdlet(VerbsCommon.New, "TemporaryFile", SupportsShouldProcess = true, HelpUri = "https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/new-temporaryfile")]
     [OutputType(typeof(System.IO.FileInfo))]
     public class NewTemporaryFileCommand : Cmdlet
     {
@@ -20,18 +20,18 @@ namespace Microsoft.PowerShell.Commands
         protected override void EndProcessing()
         {
             string filePath = null;
-            string tempPath = System.Environment.GetEnvironmentVariable("TEMP");
+            string tempPath = Path.GetTempPath();
             if (ShouldProcess(tempPath))
             {
                 try
                 {
                     filePath = Path.GetTempFileName();
                 }
-                catch (Exception e)
+                catch (IOException ioException)
                 {
-                    WriteError(
+                    ThrowTerminatingError(
                         new ErrorRecord(
-                            e,
+                            ioException,
                             "NewTemporaryFileWriteError",
                             ErrorCategory.WriteError,
                             tempPath));
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
@@ -17,14 +17,13 @@
 Describe "New-TemporaryFile" -Tags "CI" {
 
     It "creates a new temporary file" {
-            $tempFile = New-TemporaryFile
+        $tempFile = New-TemporaryFile
 
-            $tempFile | Should Exist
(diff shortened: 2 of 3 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.PowerShell.Commands
     /// <summary>
     /// The implementation of the "New-TemporaryFile" cmdlet
     /// </summary>
-    [Cmdlet(VerbsCommon.New, "TemporaryFile", SupportsShouldProcess = true, HelpUri = "https://go.microsoft.com/fwlink/?LinkId=526726")]
+    [Cmdlet(VerbsCommon.New, "TemporaryFile", SupportsShouldProcess = true, HelpUri = "https://docs.microsoft.com/powershell/module/microsoft.powershell.utility/new-temporaryfile")]
```
iSazonov: Please revert the change - we should use official short link.

test/powershell/engine/Help/assets/HelpURI/V3Cmdlets.csv
```diff
@@ -63,7 +63,7 @@ Enter-PSHostProcess,https://go.microsoft.com/fwlink/?LinkId=403736
 Exit-PSHostProcess,https://go.microsoft.com/fwlink/?LinkId=403737
 Clear-RecycleBin,https://go.microsoft.com/fwlink/?LinkId=524082
 ConvertFrom-String,https://go.microsoft.com/fwlink/?LinkId=507579
-New-TemporaryFile,https://go.microsoft.com/fwlink/?LinkId=526726
+New-TemporaryFile,https://docs.microsoft.com/powershell/module/microsoft.powershell.utility/new-temporaryfile
```
iSazonov: The same - please revert.

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
```diff
@@ -14,17 +14,16 @@
...
-        Test-Path $tempFile | Should be $true
-        $tempFile | Should BeOfType System.IO.FileInfo
-
-        if(Test-Path $tempFile)
-        {
-            Remove-Item $tempFile -ErrorAction SilentlyContinue -Force
+            $tempFile | Should Exist
+            $tempFile | Should BeOfType System.IO.FileInfo
+        } finally {
+            [System.IO.File]::Delete($tempFile)
```
iSazonov: Previous `Test-Path` and `Remove-Item` was better.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
bergmeister: Well, the official short link is broken.,,

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
bergmeister: Why? The .net method is only one line because it internally performs the check if the file exists and does not throw in this case.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
iSazonov: Feel free to open new Issue if the link is broken.

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
iSazonov: The method can throw https://msdn.microsoft.com/en-us/library/system.io.file.delete(v=vs.110).aspx?f=255&mspperror=-2147217396#Anchor_1

bergmeister:
@iSazonov I addressed your comments and opened an issue for the help link [here](https://github.com/PowerShell/PowerShell/issues/6186).

iSazonov:
@bergmeister Please update PR summary and title.

bergmeister:
@iSazonov Ok. Done.

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
```diff
@@ -14,17 +14,18 @@
...
-        $tempFile = New-TemporaryFile
-
-        Test-Path $tempFile | Should be $true
-        $tempFile | Should BeOfType System.IO.FileInfo
-
-        if(Test-Path $tempFile)
-        {
-            Remove-Item $tempFile -ErrorAction SilentlyContinue -Force
+        try {
+            $tempFile = New-TemporaryFile
```
iSazonov: I think we should remove the `try` block - if the line throw we skip the test silently.
```Powershell
{ $script:tempFile = New-TemporaryFile } | Should -Not -Throw
$script:tempFile| Should Exist
...
```

bergmeister: @iSazonov try-finally would still throw and not make the test skip silently due to the missing catch block but I also agree with you that in this special case, we do not need to use the try-finally pattern, so I will remove it

bergmeister: Also, the `Should -Not -Throw` syntax is already deprecated in Pester 4.2: https://github.com/pester/Pester/releases/tag/4.2.0

src/Microsoft.PowerShell.Commands.Utility/commands/utility/NewTemporaryFileCommand.cs
SteveL-MSFT: Agree w/ @iSazonov, we should keep the temp link and have separate issue to fix that link.  cc @joeyaiello

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
```diff
@@ -14,16 +14,15 @@
...
 #>
 
-Describe "NewTemporaryFile" -Tags "CI" {
+Describe "New-TemporaryFile" -Tags "CI" {
 
     It "creates a new temporary file" {
         $tempFile = New-TemporaryFile
 
-        Test-Path $tempFile | Should be $true
+        $tempFile | Should Exist
```
SteveL-MSFT: From a black box testing point of view, it seems that we should check that the $tempFile path starts with the path returned from GetTempPath()?

SteveL-MSFT (approved):
LGTM, one non-blocking comment

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
bergmeister: You could... Would you be ok with this one:
```` powershell
$tempFile | Should BeLikeExactly "$([System.IO.Path]::GetTempPath())*"
````

test/powershell/Modules/Microsoft.PowerShell.Utility/NewTemporaryFile.Tests.ps1
iSazonov: I see 4.2 introduce right behavior for `Should -Not -Throw `

iSazonov (approved):

daxian-dbw:
@iSazonov The `[breaking-change]` tag should be added to the body of the commit message instead of the title of it. Sorry that I didn't point out that specifically.

When squash-merging a breaking change PR in future, please do like the following:
image

iSazonov:
@daxian-dbw Thanks! Clear! Please re-check last commits which I merged.

PR #6240, opened 2018-02-25
WIP: Adding base class for custom argument completers

powercode:
## PR Summary

This codifies the  most common patterns encountered when writing argument completers in  an easy-to-use base class.

* Complete if a text is starting with the word to complete
* Complete if a text contains the word to complete
* Complete if any of the text, listItem or tooltip starts with the word to complete
* Complete if any of the text, listItem or tooltip contains the word to complete

It also quotes text with space in the completionText.

Finally it provodes a slightly less intimidating interface to override by having the AST as a member, 
hence not visible in the 

The interface for the consumer is 

```csharp
public abstract class ArgumentCompleterBase : IArgumentCompleter 
{
    protected string WordToComplete { get; private set; }
    protected CommandAst CommandAst { get; private set; }
    protected abstract void AddCompletionsFor(string commandName, string parameterName, IDictionary fakeBoundParameters);
    protected void CompleteWith(CompletionResult completionResult);
    protected void CompleteWith(string text, string listItemText = null, 
            string toolTip = null,  CompletionResultType resultType = CompletionResultType.ParameterValue);
    protected void CompleteWithIfTextStartsWithWordToComplete(string text, string listItemText = null,
            string toolTip = null, CompletionResultType resultType = CompletionResultType.ParameterValue);
     protected void CompleteWithIfAnyStartsWithWordToComplete(string text, string listItemText = null,
            string toolTip = null, CompletionResultType resultType = CompletionResultType.ParameterValue);
    protected void CompleteWithIfAnyContainsWordToComplete(string text, string listItemText = null,
            string toolTip = null, CompletionResultType resultType = CompletionResultType.ParameterValue);
    protected void CompleteWithIfTextContainsWordToComplete(string text, string listItemText = null,
            string toolTip = null, CompletionResultType resultType = CompletionResultType.ParameterValue);
    protected bool StartWithWordToComplete(string text);
    protected bool StartWithWordToComplete(string text, StringComparison stringComparison);
    protected bool ContainsWordToComplete(string text);
    protected bool ContainsWordToComplete(string text, StringComparison stringComparison);
}
```
In PowerShell, a consumer of the base class can do the following:

```powershell
class CustomArgumentCompleter : System.Management.Automation.ArgumentCompleterBase {
    # override
    [void] AddCompletionsFor([string] $commandName, [string] $parameterName, [Collections.IDictionary] $fakeBoundParameters) {
        switch ("${commandName}:$parameterName") {
            # handle completion based on both command and parameter name
            "Verb-Noun:ParamName" { $this.CompleteVerbNounParam($fakeBoundParameters); break }
            #
            #  Add one switch clases for each parameter with generic completion
            #

            default {
                # handle generic parameter completion
                switch ($parameterName) {
                    "ParameterName" { $this.CompleteParameterName($fakeBoundParameters); break}
                    #
                    #  Add one switch clases for each parameter with generic completion
                    #
                }
            }
        }
    }

    [void] CompleteVerbNounParam([Collections.IDictionary] $fakeBoundParameter) {
        # complete with "completiontext" if wordToComplete is anywhere in "completiontext"
        #$this.CompleteWithIfTextContainsWordToComplete("completiontext")
    }

    [void] CompleteParameterName([Collections.IDictionary] $fakeBoundParameter) {
        # complete some ParameterNames that starts with wordToComplete
        #$this.CompleteWithIfStartsWithWordToComplete("Some ParameterName")
    }
}
```

This is both easy describe, as it follows a clear pattern, and 

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/CommandCompletion/ExtensibleCompletion.cs
+++ b/src/System.Management.Automation/engine/CommandCompletion/ExtensibleCompletion.cs
@@ -207,4 +207,168 @@ namespace System.Management.Automation
             }
         }
     }
+
+
+    /// <summary>
+    ///     Base class for writing custom Argument Completers
+    /// </summary>
+    /// <para>Derived classes should override <see cref="AddCompletionsFor(string,string,IDictionary)"/> </para>
+    public abstract class ArgumentCompleterBase : IArgumentCompleter
+    {
+        private List<CompletionResult> _results;
+
+        /// <summary>
+        ///     The word to complete
+        /// </summary>
+        protected string WordToComplete { get; private set; }
+
+        /// <summary>
+        ///     The Command Ast for the command to complete
+        /// </summary>
+        protected CommandAst CommandAst { get; private set; }
+
+        private List<CompletionResult> Results => _results ?? (_results = new List<CompletionResult>());
+
+        IEnumerable<CompletionResult> IArgumentCompleter.CompleteArgument(string commandName, string parameterName,
+            string wordToComplete, CommandAst commandAst, IDictionary fakeBoundParameters)
+        {
+            WordToComplete = wordToComplete;
+            CommandAst = commandAst;
+            AddCompletionsFor(commandName, parameterName, fakeBoundParameters);
+            return _results;
+        }
+
+        /// <summary>
+        ///     Override in child class to add completions by calling <see ref="CompleteWith"/>
+        /// </summary>
+        /// <param name="commandName">the command to complete parameters for</param>
+        /// <param name="parameterName">the parameter to complete</param>
+        /// <param name="fakeBoundParameters">previously specified command parameters</param>
+        protected abstract void AddCompletionsFor(string commandName, string parameterName, IDictionary fakeBoundParameters);
+
+        /// <summary>
+        ///     Adds a completion result to the result set
+        /// </summary>
+        /// <param name="completionResult">the completion result to add</param>
+        protected void CompleteWith(CompletionResult completionResult)
+            => Results.Add(item: completionResult);
+
+        /// <summary>
+        ///     Adds a completion result to the result set with the specified parameters
+        /// </summary>
+        /// <param name="text">the text to be used as the auto completion result</param>
+        /// <param name="listItemText">the text to be displayed in a list</param>
+        /// <param name="toolTip">the text for the tooltip with details to be displayed about the object</param>
+        /// <param name="resultType">the type of completion result</param>
+        protected void CompleteWith(string text, string listItemText = null, string toolTip = null,
+            CompletionResultType resultType = CompletionResultType.ParameterValue)
+        {
+            var quotedText = QuoteIfSpace(text: text);
+            var completionResult = new CompletionResult(completionText: quotedText, listItemText: listItemText ?? text,
+                resultType: resultType, toolTip: toolTip ?? text);
+            Results.Add(item: completionResult);
+        }
+
+        /// <summary>
+        ///     Adds a completion result to the result set if the text starts with <see cref="WordToComplete" />
+        /// </summary>
+        /// <para>The comparison is case insensitive</para>
+        /// <param name="text">the text to be used as the auto completion result</param>
+        /// <param name="listItemText">the text to be displayed in a list</param>
+        /// <param name="toolTip">the text for the tooltip with details to be displayed about the object</param>
+        /// <param name="resultType">the type of completion result</param>
+        protected void CompleteWithIfTextStartsWithWordToComplete(string text, string listItemText = null,
+            string toolTip = null, CompletionResultType resultType = CompletionResultType.ParameterValue)
+        {
+            if (StartWithWordToComplete(text: text))
+                CompleteWith(text: text, listItemText: listItemText ?? text, toolTip: toolTip ?? text,
+                    resultType: resultType);
+        }
+
+        /// <summary>
+        ///     Adds a completion result to the result set if the any string argument starts with <see cref="WordToComplete" />
+        /// </summary>
+        /// <para>The comparison is case insensitive</para>
+        /// <param name="text">the text to be used as the auto completion result</param>
+        /// <param name="listItemText">the text to be displayed in a list</param>
+        /// <param name="toolTip">the text for the tooltip with details to be displayed about the object</param>
+        /// <param name="resultType">the type of completion result</param>
+        protected void CompleteWithIfAnyStartsWithWordToComplete(string text, string listItemText = null,
+            string toolTip = null, CompletionResultType resultType = CompletionResultType.ParameterValue)
+        {
+            if (StartWithWordToComplete(text: text) || StartWithWordToComplete(listItemText) || StartWithWordToComplete(toolTip))
+                CompleteWith(text: text, listItemText: listItemText ?? text, toolTip: toolTip ?? text,
+                    resultType: resultType);
+        }
+
+
+        /// <summary>
+        ///     Adds a completion result to the result set if the any string argument starts with <see cref="WordToComplete" />
+        /// </summary>
+        /// <para>The comparison is case insensitive</para>
+        /// <param name="text">the text to be used as the auto completion result</param>
+        /// <param name="listItemText">the text to be displayed in a list</param>
+        /// <param name="toolTip">the text for the tooltip with details to be displayed about the object</param>
+        /// <param name="resultType">the type of completion result</param>
+        protected void CompleteWithIfAnyContainsWordToComplete(string text, string listItemText = null,
+            string toolTip = null, CompletionResultType resultType = CompletionResultType.ParameterValue)
+        {
+            if (ContainsWordToComplete(text) || ContainsWordToComplete(listItemText) || ContainsWordToComplete(toolTip))
+                CompleteWith(text: text, listItemText: listItemText ?? text, toolTip: toolTip ?? text,
+                    resultType: resultType);
+        }
+
+        /// <summary>
+        ///     Adds a completion result to the result set if the text contains <see cref="WordToComplete" />
+        /// </summary>
+        /// <para>The comparison is case insensitive</para>
+        /// <param name="text">the text to be used as the auto completion result</param>
+        /// <param name="listItemText">the text to be displayed in a list</param>
+        /// <param name="toolTip">the text for the tooltip with details to be displayed about the object</param>
+        /// <param name="resultType">the type of completion result</param>
+        protected void CompleteWithIfTextContainsWordToComplete(string text, string listItemText = null,
+            string toolTip = null, CompletionResultType resultType = CompletionResultType.ParameterValue)
+        {
+            if (ContainsWordToComplete(text))
+                CompleteWith(text: text, listItemText: listItemText ?? text, toolTip: toolTip ?? text,
+                    resultType: resultType);
+        }
+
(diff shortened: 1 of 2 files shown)


iSazonov:
@powercode Please correct PR description - it seems you lost something after "This is both easy describe, as it follows a clear pattern, and".

src/System.Management.Automation/engine/CommandCompletion/ExtensibleCompletion.cs
```diff
@@ -207,4 +207,168 @@ public IEnumerable<CompletionResult> CompleteArgument(string commandName, string
             }
         }
     }
+
+
+    /// <summary>
+    ///     Base class for writing custom Argument Completers
```
iSazonov: Usually we use one space after `///`. And please add final dot in public comments.

iSazonov:
@SteveL-MSFT We need PowerShell Committee conclusion about public API in the PR.

joeyaiello:
Summary of  @PowerShell/powershell-committee discussion on this: we all *love* this in concept, and think that it generally looks sound, but we also know that none of us have real expertise on argument completers. This is relatively small, so we don't feel like kicking it out to the RFC process, but we'd all feel a lot more comfortable if it had some bake time for feedback. 

Specifically, it would be great if folks have written a lot of argument completers to provide their feedback. Off the top of my head, @lzybkr knows a ton here (not least of which because of PSReadline), and I know both @dahlbyk and @LeoColomb work on posh-git. Probably @rkeithhill has some experience in this space as well? It'd be great to hear from you all to make sure  we get the details right here.

powercode:
Something that could be discussed is if we should provide helpers for writing completers for native commands too. This PR only targets parameters on cmdlets/functions, but there are common patterns there too.

For native commands, you quite often you have to find the previous item and understand if it is a flag, a parameter or a subcommand.

src/System.Management.Automation/engine/CommandCompletion/ExtensibleCompletion.cs
```diff
@@ -207,4 +207,168 @@ public IEnumerable<CompletionResult> CompleteArgument(string commandName, string
...
+
+        /// <summary>
+        ///     Adds a completion result to the result set with the specified parameters
+        /// </summary>
+        /// <param name="text">the text to be used as the auto completion result</param>
+        /// <param name="listItemText">the text to be displayed in a list</param>
+        /// <param name="toolTip">the text for the tooltip with details to be displayed about the object</param>
+        /// <param name="resultType">the type of completion result</param>
+        protected void CompleteWith(string text, string listItemText = null, string toolTip = null,
+            CompletionResultType resultType = CompletionResultType.ParameterValue)
```
dahlbyk: One of my biggest worries about porting posh-git to be a PowerShell completer is that Git's command line conventions don't necessarily align with PowerShell's. For example, most commands should parse anything after `--` as a file path. I'm not familiar enough to know, but it may be that `ParameterValue` isn't the most reasonable default `CompletionResultType` for my use case. (Does anyone have a link where I can read up on the various types? The [docs](https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.completionresulttype) don't tell me much about how to think about native command arguments.)

With that in mind, it might be useful to make the default configurable, e.g.

```csharp

        protected virtual DefaultCompletionResultType => CompletionResultType.ParameterValue;

        protected void CompleteWith(string text, string listItemText = null, string toolTip = null,
            CompletionResultType? resultType = null)
        {
            var quotedText = QuoteIfSpace(text: text);
            var completionResult = new CompletionResult(completionText: quotedText, listItemText: listItemText ?? text,
                resultType: resultType ?? DefaultCompletionResultType, toolTip: toolTip ?? text);
            Results.Add(item: completionResult);
        }
```

dahlbyk (commented):
I regret that posh-git argument completion is still 100% regex, so I don't have experience with the PowerShell `IArgumentCompleter`. The same is true for Chocolatey's [tab expansion](https://github.com/ferventcoder/choco/blob/878b8bb16a2323a52e86112f2e292ea3127e5f8e/src/chocolatey.resources/helpers/ChocolateyTabExpansion.ps1), though @ferventcoder may still have thoughts.

That being said, the biggest gap from a posh-git standpoint is that a match nearly always adds multiple completion results, e.g. `git branch m<tab>` adds `master` and `my-feature`. For these cases, supporting `CompleteWith(IEnumerable<CompletionResult> completionResult)` (a pass-through to `Results.AddRange()`) and/or `CompleteWith(IEnumerable<string> textResults, CompletionResultType resultType = CompletionResultType.ParameterValue)` could make for simpler code.

src/System.Management.Automation/engine/CommandCompletion/ExtensibleCompletion.cs
ferventcoder: I like the regex. Very fast parsing. It does have a tendency to take some time to get all of the context back in my head when working with it though - "write once, read never" ;)

dahlbyk:
> Something that could be discussed is if we should provide helpers for writing completers for native commands too. This PR only targets parameters on cmdlets/functions, but there are common patterns there too.
> 
> For native commands, you quite often you have to find the previous item and understand if it is a flag, a parameter or a subcommand.
> 
> It is considerably more difficult to write good completers for complex native executables than it is for powershell parameters, so some helpers would be very valuable there too.

I missed this since I first saw the issue last night, but I agree. I mentioned it earlier, but `--` handling is just one of a handful of complicating factors for posh-git:

- Nearly all (but not all - shout out to `git filter-branch`) commands treat anything after ` -- ` as a file path
- Git aliases need to be parsed to determined which command we're actually doing expansion for
- Expanding a remote name (e.g. `origin`) might be just the origin name, or it might be part of a branch reference (e.g. `origin/master`) depending on context
- When completing file paths, we sometimes give preference to paths relevant to the command (e.g. `git add <tab>` expands paths that have been modified)
- I'm not sure how PowerShell completers handle command arguments aliases (`--patch` / `-p`)

src/System.Management.Automation/engine/CommandCompletion/ExtensibleCompletion.cs
```diff
@@ -207,4 +207,168 @@ public IEnumerable<CompletionResult> CompleteArgument(string commandName, string
...
+        /// <param name="resultType">the type of completion result</param>
+        protected void CompleteWithIfTextContainsWordToComplete(string text, string listItemText = null,
+            string toolTip = null, CompletionResultType resultType = CompletionResultType.ParameterValue)
+        {
+            if (ContainsWordToComplete(text))
+                CompleteWith(text: text, listItemText: listItemText ?? text, toolTip: toolTip ?? text,
+                    resultType: resultType);
+        }
+
+        private static string QuoteIfSpace(string text)
```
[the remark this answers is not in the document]
powercode: Good point. Actually found such a function existing in CompletionCompleters.cs

src/System.Management.Automation/engine/CommandCompletion/ExtensibleCompletion.cs
powercode: @dahlbyk You cannot use this base class for the git completer. It is not intended as such, but it would be interesting to take a stab at that to.

The native commands has a different signature for their completers, and I'm thinking about adding a base class which implements 

```csharp
enum NativeCompletionTokenKind {
    None,
    Command,
    Dash,
    DashDash,
    Text
}

class NativeArgumentCompleterBase  {
    protected string CommandName {get;}
    protected string WordToComplete {get;}
    protected string PreviousToken {get;}
    protected NativeCompletionTokenKind PreviousTokenKind {get;}
    protected NativeCompletionTokenKind TokenKind {get;}
    protected CommandAst Ast {get;}
    protected int CursorPosition {get;}
    
    /// TBD
    /// Helpers for 
  
    /// this is used to register the completer with the engine
    static void RegisterCompleter(NativeArgumentCompleterBase completer);
    /// this is the registered public entry point
    public IList<CompletionResult> CompleteInput(string wordToComplete, 
                               CommandAst ast, int cursorPosition) {
       /// Set useful properties from the AST and cursorPosition
      
       Complete();
    }

    protected abstract void Complete();
  
}


class GitCompleter : NativeArgumentCompleterBase , IModuleAssemblyInitializer {
    /// here we have access to an easier abstraction that the CommandAst
    public Complete() {
         ///
    }    

    /// Register the completer.    
    void IModuleAssemblyInitializer.OnImport() =>RegisterCompleter(typeof(GitCompleter );
}
```

From PowerShell, 


One thing that Would be nice in such a case is some built in support for testing.

I regularly write completers where I have a mocked version of a native executable runner, and if the command line matches one of my named commands, I just return the a result from a file stream or a resource.

That way it is easy to experiment with different parsing/completions without even having the command under test installed.

powercode:
To be fair, I think we can see a couple of different use cases with completion. Git is in one extreme, with a very complicated command line interface. Others provide more regularity. A lot of the logic to parse git commands will be custom, but I am more on the chase for helpers that would make that task easier.

lzybkr:
It's worth point folks to my stab at making native completers easier to write [here](https://github.com/lzybkr/TabExpansionPlusPlus/blob/master/WindowsExe.ArgumentCompleters.ps1) - the `net.exe` example is similar to what I could see for `git`.

dahlbyk:
> You cannot use this base class for the git completer. It is not intended as such

In that case, ignore my Git/native-specific feedback in the context of this particular PR. 😀

@powercode glad to participate in a discussion about how we can make native parsers easier to write. There's a bit of overlap (e.g. collecting `Results`, the `StartWithWordToComplete`/`ContainsWordToComplete` helpers), but perhaps not enough to try to share any code.

> One thing that Would be nice in such a case is some built in support for testing.

We have some infrastructure to mock `git` results for a given command, but it would be great to be able to test tab expansion in isolation.

src/System.Management.Automation/engine/CommandCompletion/ExtensibleCompletion.cs
```diff
@@ -207,4 +207,168 @@ public IEnumerable<CompletionResult> CompleteArgument(string commandName, string
...
+            => text.Contains(" ") ? $@"""{text}""" : text;
+
+        /// <summary>
+        ///     Predicate to test if a string starts with <see cref="WordToComplete" />
+        /// </summary>
+        /// <param name="text"></param>
+        /// <returns>true if the text contains <see cref="WordToComplete" />, otherwise false</returns>
+        protected bool StartWithWordToComplete(string text)
+            => text.StartsWith(value: WordToComplete, comparisonType: StringComparison.CurrentCultureIgnoreCase);
```
dahlbyk: Following on from @lzybkr's [link](https://github.com/lzybkr/TabExpansionPlusPlus/blob/deea6b96fcfb290f98da44104336f27d0be68fed/WindowsExe.ArgumentCompleters.ps1#L16), is `OrginalIgnoreCase` maybe a more correct default? (Says the American that never uses alphabets where they're not equivalent.)

SteveL-MSFT:
@powercode I think native command completion is definitely something we want to support and making it easier for the community to author completers for popular native commands on Windows/Linux would be awesome

powercode:
I've started prototyping.

powercode:
@lzybkr I have looked at your completion tree before, and I really like the idea! I would like to have them being lazy, so we don't allocate when we don't have to.

Issue #6228, opened 2018-02-23
Tests in "Implicit.Remoting.Tests.ps1" fails after #5568 was merged


daxian-dbw:
Tests in "Implicit.Remoting.Tests.ps1" fails after #5568 was merged.

Here is the failure: https://ci.appveyor.com/project/PowerShell/powershell-f975h/build/v6.1.0-preview.587
```
  Describing Import-PSSession with FormatAndTypes
VERBOSE: Using Global AppVeyor Credential
    Context Importing format file works
      [-] Error occurred in Context block 640ms
        Expected strings to be the same, but they were different.
        Expected length: 48
        Actual length:   46
        Strings differ at index 37.
        Expected: '\r
        Width Height\r
        ----- ------\r
          123    456\r
        \r
        \r
        '
        But was:  '\r
        Width Height\r
        ----- ------\r
          123  456\r
        \r
        \r
        '
        ------------------------------------------------------^
        642:                 $originalLocalFormatting | Should Be $originalRemoteFormatting
        at Invoke-LegacyAssertion, C:\projects\powershell-f975h\src\powershell-win-core\bin\Release\netcoreapp2.0\win7-x64\publish\Modules\Pester\4.3.1\Functions\Assertions\Should.ps1: line 188
        at <ScriptBlock>, C:\projects\powershell-f975h\test\powershell\Modules\Microsoft.PowerShell.Utility\Implicit.Remoting.Tests.ps1: line 642
        at Invoke-Blocks, C:\projects\powershell-f975h\src\powershell-win-core\bin\Release\netcoreapp2.0\win7-x64\publish\Modules\Pester\4.3.1\Functions\SetupTeardown.ps1: line 140
        at Invoke-TestGroupSetupBlocks, C:\projects\powershell-f975h\src\powershell-win-core\bin\Release\netcoreapp2.0\win7-x64\publish\Modules\Pester\4.3.1\Functions\SetupTeardown.ps1: line 125
        at DescribeImpl, C:\projects\powershell-f975h\src\powershell-win-core\bin\Release\netcoreapp2.0\win7-x64\publish\Modules\Pester\4.3.1\Functions\Describe.ps1: line 157
```

PR #6241, opened 2018-02-26
Fix `Start-PSPester` support for running tests w/ `sudo`

### LINKED ISSUE #6227: Two tests in 'Set-Date.Tests.ps1' consistently fails after #6145 was merged
Two tests in 'Set-Date.Tests.ps1' consistently fails after #6145 was merged

Here is the failure:
```
  Describing Set-Date for admin
    [-] Set-Date should be able to set the date in an elevated context 166ms
      Expected: the expression not to throw an exception. Message was {Operation not permitted}
          from /home/travis/build/PowerShell/PowerShell/test/powershell/Modules/Microsoft.PowerShell.Utility/Set-Date.Tests.ps1:5 char:22
          +         { Get-Date | Set-Date } | Should Not Throw
          +                      ~~~~~~~~
      5:         { Get-Date | Set-Date } | Should Not Throw
      at Invoke-LegacyAssertion, /home/travis/build/PowerShell/PowerShell/src/powershell-unix/bin/Linux/netcoreapp2.0/linux-x64/publish/Modules/Pester/4.1.1/Functions/Assertions/Should.ps1: line 190
      at <ScriptBlock>, /home/travis/build/PowerShell/PowerShell/test/powershell/Modules/Microsoft.PowerShell.Utility/Set-Date.Tests.ps1: line 5
    [-] Set-Date should be able to set the date with -Date parameter 111ms
      Win32Exception: Operation not permitted
      at <ScriptBlock>, /home/travis/build/PowerShell/PowerShell/test/powershell/Modules/Microsoft.PowerShell.Utility/Set-Date.Tests.ps1: line 11
```
Here is the failure history:
image

SteveL-MSFT:
## PR Summary

if `Start-PSPester -sudo` is used, make sure to only run tests tagged with `RequireSudoOnUnix` (note that `CI`,`Feature`,etc.. no longer apply)
if `Start-PSPester` is used, make sure to exclude `RequireSudoOnUnix` tests on Unix

Fix https://github.com/PowerShell/PowerShell/issues/6227

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -898,7 +898,7 @@ function Get-PesterTag {
                     }
                     $values = $vAst.FindAll({$args[0] -is "System.Management.Automation.Language.StringConstantExpressionAst"},$true).Value
                     $values | ForEach-Object {
-                        if (@('REQUIREADMINONWINDOWS', 'REQUIRESUDOONUNIX', 'SLOW') -contains $_) {
+                        if (@('REQUIREADMINONWINDOWS', 'REQUIRESUDOONUNIX', 'SLOW', 'REQUIRENOSUDOONUNIX') -contains $_) {
                             # These are valid tags also, but they are not the priority tags
                         }
                         elseif (@('CI', 'FEATURE', 'SCENARIO') -contains $_) {
@@ -1023,6 +1023,20 @@ Restore the module to '$Pester' by running:
         if (-not $PSBoundParameters.ContainsKey('ExcludeTag'))
         {
             $ExcludeTag += 'RequireAdminOnWindows'
+        }
+    }
+    elseif (-not $Environment.IsWindows -and (-not $Sudo.IsPresent))
+    {
+        if (-not $PSBoundParameters.ContainsKey('ExcludeTag'))
+        {
+            $ExcludeTag += 'RequireSudoOnUnix'
+        }
+    }
+    elseif (-not $Environment.IsWindows -and $Sudo.IsPresent)
+    {
+        if (-not $PSBoundParameters.ContainsKey('ExcludeTag'))
+        {
+            $ExcludeTag += 'RequireNoSudoOnUnix'
         }
     }
 
--- a/build.psm1
+++ b/build.psm1
@@ -898,7 +898,7 @@ function Get-PesterTag {
                     }
                     $values = $vAst.FindAll({$args[0] -is "System.Management.Automation.Language.StringConstantExpressionAst"},$true).Value
                     $values | ForEach-Object {
-                        if (@('REQUIREADMINONWINDOWS', 'REQUIRESUDOONUNIX', 'SLOW', 'REQUIRENOSUDOONUNIX') -contains $_) {
+                        if (@('REQUIREADMINONWINDOWS', 'REQUIRESUDOONUNIX', 'SLOW') -contains $_) {
                             # These are valid tags also, but they are not the priority tags
                         }
                         elseif (@('CI', 'FEATURE', 'SCENARIO') -contains $_) {
@@ -1034,9 +1034,9 @@ Restore the module to '$Pester' by running:
     }
     elseif (-not $Environment.IsWindows -and $Sudo.IsPresent)
     {
-        if (-not $PSBoundParameters.ContainsKey('ExcludeTag'))
+        if (-not $PSBoundParameters.ContainsKey('Tag'))
         {
-            $ExcludeTag += 'RequireNoSudoOnUnix'
+            $Tag = 'RequireSudoOnUnix'
         }
     }
 
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/Set-Date.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/Set-Date.Tests.ps1
@@ -1,6 +1,5 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
-
 Describe "Set-Date for admin" -Tag @('CI', 'RequireAdminOnWindows', 'RequireSudoOnUnix') {
     It "Set-Date should be able to set the date in an elevated context" -Pending {
         { Get-Date | Set-Date } | Should Not Throw
@@ -13,7 +12,7 @@ Describe "Set-Date for admin" -Tag @('CI', 'RequireAdminOnWindows', 'RequireSudo
     }
 }
 
-Describe "Set-Date" -Tag @('CI') {
+Describe "Set-Date" -Tag 'CI' {
     It "Set-Date should produce an error in a non-elevated context" {
         { Get-Date | Set-Date } | ShouldBeErrorId "System.ComponentModel.Win32Exception,Microsoft.PowerShell.Commands.SetDateCommand"
     }
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/Set-Date.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/Set-Date.Tests.ps1
@@ -1,5 +1,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
+
 Describe "Set-Date for admin" -Tag @('CI', 'RequireAdminOnWindows', 'RequireSudoOnUnix') {
     It "Set-Date should be able to set the date in an elevated context" -Pending {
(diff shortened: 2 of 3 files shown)


build.psm1
```diff
@@ -898,7 +898,7 @@ function Get-PesterTag {
                     }
                     $values = $vAst.FindAll({$args[0] -is "System.Management.Automation.Language.StringConstantExpressionAst"},$true).Value
                     $values | ForEach-Object {
-                        if (@('REQUIREADMINONWINDOWS', 'REQUIRESUDOONUNIX', 'SLOW') -contains $_) {
+                        if (@('REQUIREADMINONWINDOWS', 'REQUIRESUDOONUNIX', 'SLOW', 'REQUIRENOSUDOONUNIX') -contains $_) {
```
iSazonov: I wonder why we add 'REQUIRENOSUDOONUNIX' if we haven't 'REQUIRENOADMINONWINSOWS'?

SteveL-MSFT:
AppVeyor failure due to #6228

build.psm1
SteveL-MSFT: On Windows, tests are always run elevated (remoting tests, update-help, and some others require elevation).  Tests that explicitly require non-admin (there's a few since I wrote some to check for Access Denied) uses runas to start a non-admin user.

On Linux, tests are usually run not under sudo.  There's just one test (currently) that explicitly tests the case where you need to be root, but running as not root you should get an error.  Only a few tests that require sudo.  We could change this design to always require running the tests under sudo, but that feels worse than requiring tests to run elevated on Windows.

It may be worth discussing in a separate issue so we can unblock the nightly runs with this PR.

iSazonov: I see we run tests on Windows with `-Unelevated` by default and elevated if 'REQUIREADMINONWINDOWS' present - I'd expect the same for Unix - run test unelevated by default (that we have already) and elevated if 'REQUIRESUDOONUNIX' present - I don't see why we need another new tag.
It is non-blocking comment.

SteveL-MSFT: I was thinking more of the individual engineer and not the nightly.  I think we could support `-Unelevated` on UNIX by creating a non-root user account and running the tests under that account.  Created https://github.com/PowerShell/PowerShell/issues/6246 to track.

iSazonov: We'll need update docs https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/WritingPesterTests.md#admin-privileges-in-tests

tools/travis.ps1
```diff
@@ -236,7 +235,7 @@ elseif($Stage -eq 'Build')
 
     # Running tests, which require sudo.
     $pesterParam['Tag'] = @('RequireSudoOnUnix')
-    $pesterParam['ExcludeTag'] = @()
+    $pesterParam['ExcludeTag'] = @('RequireNoSudoOnUnix')
```
daxian-dbw: I don't think the `REQUIRENOSUDOONUNIX` tag is needed.
As you have fixed here, the problem is at line 217. The `ExcludeTag` was set to empty incorrectly when running without `SUDO`.

When running with `SUDO`, the `Tag` and `ExcludeTag` were set correctly -- only `Describe` blocks with the `RequireSudoOnUnix` tag will be included.

daxian-dbw (changes requested):

tools/travis.ps1
SteveL-MSFT: Ok, will remove.

test/powershell/Modules/Microsoft.PowerShell.Utility/Set-Date.Tests.ps1
```diff
@@ -12,7 +13,7 @@ Describe "Set-Date for admin" -Tag @('CI', 'RequireAdminOnWindows', 'RequireSudo
     }
 }
 
-Describe "Set-Date" -Tag 'CI' {
+Describe "Set-Date" -Tag @('CI', 'RequireNoSudoOnUnix') {
```
daxian-dbw: This tag is not needed for this test. Only tests with `RequireSudoOnUnix` tag will be included in a SUDO test run.
I think this new tag should be removed from `build.psm1` too.

daxian-dbw (changes requested):

build.psm1
TravisEz13: RequiresNoSudoOnUnix is not needed.  Please don't needlessly added a tag.

test/powershell/Modules/Microsoft.PowerShell.Utility/Set-Date.Tests.ps1
TravisEz13: Yes, please remove this tag

TravisEz13 (changes requested):

tools/travis.ps1
```diff
@@ -214,7 +214,6 @@ elseif($Stage -eq 'Build')
 
     if ($isFullBuild) {
         $pesterParam['Tag'] = @('CI','Feature','Scenario')
-        $pesterParam['ExcludeTag'] = @()
```
TravisEz13: I'll submit a complete fix to fix this issue.

test/powershell/Modules/Microsoft.PowerShell.Utility/Set-Date.Tests.ps1
SteveL-MSFT: Without this tag, this test is run when `sudo` is used and this test fails because it's expecting an exception, but no exception occurs.

TravisEz13: We should only run `REQUIRESUDOONUNIX` tests with -sudo.  I have a fix for travis.ps1 for this issue.

TravisEz13: cc @JamesWTruher

SteveL-MSFT: @TravisEz13 that would make sense, so then we should exclude all tests that don't have `RequiresSudoOnUnix` when `-Sudo` is used, correct?

build.psm1
JamesWTruher: we really don't want to add a tag for this, the default behavior should be to not use sudo

test/powershell/Modules/Microsoft.PowerShell.Utility/Set-Date.Tests.ps1
TravisEz13: when running sudo, only include the tests with the `RequiresSudoOnUnix` Tag.

SteveL-MSFT: Ok, I think I got it figured out, but it means that `RequireSudoOnUnix` tag means `CI`,`Feature`,`Slow`,etc... no longer apply.

dantraMSFT:
At the risk of splitting hairs; why isn't the tag simply RequireSudo or Sudo?  Isn't the 'OnUnix' redundant?
FWIW: My first thought, does it work on MacOS?

SteveL-MSFT:
@dantraMSFT yes, it works on macOS.  I think the `OnUnix` is just to be consistent with `RequireAdminOnWindows`.

TravisEz13:
@dantraMSFT Some people on the original PR though it needed clarification.  I agreed to be consistent.

tools/travis.ps1
TravisEz13: Please rebase

daxian-dbw:
The `set-date` test failures have been fixed in #6249 
Shall we close this PR?

TravisEz13:
this gets closer to something that @anmenaga asked for.  making the travis and start-pspester behavior the same.

daxian-dbw:
I guess we want the changes in `build.psm1`.
@SteveL-MSFT can you please revert the changes in other files?

SteveL-MSFT:
@daxian-dbw @TravisEz13 in general, the way individuals run tests and the way our automated systems run it should be as similar as possible otherwise tests fail in one environment and not the other making it more difficult to diagnose

SteveL-MSFT:
@daxian-dbw will do

daxian-dbw (approved):

TravisEz13 (approved):

iSazonov (approved):

Issue #6245, opened 2018-02-26
Clean up test code to use `-Parameter` syntax for Pester tests


SteveL-MSFT:
Pester documentation for 4.x recommends using `Should -Be` instead of `Should Be` syntax.  We should update all our tests and documentation to adhere to this setting a good example for the community.
- [x] engine #6298                                                                                                               
- [x] Host #6250
- [x] Installer #6250 
- [x] Language #6304
- [x] Modules/CimCmdlets #6306
- [x] Microsoft.PowerShell.Core #6349
- [x] Microsoft.PowerShell.Diagnostics #6351
- [x] Microsoft.Powershell.Host #6250
- [x] Microsoft.PowerShell.LocalAccounts #6499
- [x] Microsoft.PowerShell.Management #6294
- [x] Microsoft.PowerShell.Security #6256
- [x] Microsoft.PowerShell.Utility #6366
- [x] Microsoft.PowerShell.Utility Web #6257                                                                                    
- [x] Microsoft.WSMan.Management #6366
- [x] PackageManagement #6488 
- [x] Modules/PowerShellGet #6488 
- [x] Modules/PSDesiredStateConfiguration #6622 
- [x] Modules/PSReadLine #6488
- [x] Provider #6490
- [x] SDK #6490

iSazonov:
Maybe Meta-Issue? I'd rather break the work apart otherwise we'll get a huge size PR which is hard to review.

markekraus:
hehe I planned to do this with the web cmdlet tests, yes.

SteveL-MSFT:
@iSazonov breaking it up makes sense!

bergmeister:
I haven't tried it yet but [here](https://gist.github.com/indented-automation/aeb14825e39dd8849beee44f681fbab3) is a helper script to upgrade to the new syntax.

KevinMarquette:
I have a PR out for these items on the list:
PowerShellGet, PackageManagement, PSReadLine #6488
SDK and Provider tests #6490

KevinMarquette:
I have a PR out for the LocalAccounts module #6499

xtqqczze:
#6499 was not merged

xtqqczze:
@iSazonov Can you reopen?

iSazonov:
@xtqqczze The Microsoft.PowerShell.LocalAccounts module was removed from PowerShell Core.

xtqqczze:
[...]

PR #6250, opened 2018-02-26
Upgrade Pester tests in test\powershell\Host folder to dashed v4 syntax

bergmeister:
## PR Summary

Fixes a subset of #6245
Use new dashed Pester v4 operators for tests in `test\powershell\Host`

## PR Checklist

- [NA] [Make sure all `.h`, `.cpp`, `.cs`, `.ps1` and `.psm1` files have the correct copyright header](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/test/powershell/Host/ConsoleHost.Tests.ps1
+++ b/test/powershell/Host/ConsoleHost.Tests.ps1
@@ -128,7 +128,7 @@ Describe "ConsoleHost unit tests" -tags "Feature" {
             {
                 $a="blahblah"
                 $a | & $powershell -noprofile -out text -com { $input }
-            } | Should Not Throw
+            } | Should -Not Throw
         }
 
         It "Verify Parsing Error Input Format Single Shell should throw exception" {
@@ -154,7 +154,7 @@ Describe "ConsoleHost unit tests" -tags "Feature" {
         }
         foreach ($x in "--help", "-help", "-h", "-?", "--he", "-hel", "--HELP", "-hEl") {
             It "Accepts '$x' as a parameter for help" {
-                & $powershell -noprofile $x | Where-Object { $_ -match "pwsh[.exe] -Help | -? | /?" } | Should Not BeNullOrEmpty
+                & $powershell -noprofile $x | Where-Object { $_ -match "pwsh[.exe] -Help | -? | /?" } | Should -Not BeNullOrEmpty
             }
         }
 
--- a/test/powershell/Host/ConsoleHost.Tests.ps1
+++ b/test/powershell/Host/ConsoleHost.Tests.ps1
@@ -154,7 +154,7 @@ Describe "ConsoleHost unit tests" -tags "Feature" {
         }
         foreach ($x in "--help", "-help", "-h", "-?", "--he", "-hel", "--HELP", "-hEl") {
             It "Accepts '$x' as a parameter for help" {
-                & $powershell -noprofile $x | Where-Object { $_ -match "pwsh[.exe] -Help | -? | /?" } | Should -Not BeNullOrEmpty
+                & $powershell -noprofile $x | Where-Object { $_ -match "pwsh[.exe] -Help | -? | /?" } | Should -Not -BeNullOrEmpty
             }
         }
 
--- a/test/powershell/Host/ConsoleHost.Tests.ps1
+++ b/test/powershell/Host/ConsoleHost.Tests.ps1
@@ -125,7 +125,9 @@ Describe "ConsoleHost unit tests" -tags "Feature" {
         }
 
         It "Verify Validate Output Format As Text Explicitly Child Single Shell does not throw" {
-            "blahblah" | & $powershell -noprofile -out text -com { $input }
+            {
+                "blahblah" | & $powershell -noprofile -out text -com { $input }
+            } | Should -Not -Throw
         }
 
         It "Verify Parsing Error Input Format Single Shell should throw exception" {
--- a/test/powershell/Host/ConsoleHost.Tests.ps1
+++ b/test/powershell/Host/ConsoleHost.Tests.ps1
@@ -124,11 +124,8 @@ Describe "ConsoleHost unit tests" -tags "Feature" {
             }
         }
 
-        It "Verify Validate Output Format As Text Explicitly Child Single Shell should works" {
-            {
-                $a="blahblah"
-                $a | & $powershell -noprofile -out text -com { $input }
-            } | Should -Not Throw
+        It "Verify Validate Output Format As Text Explicitly Child Single Shell doess not throw" {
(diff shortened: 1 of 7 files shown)


iSazonov:
@bergmeister Please clarify about "-Not -Throw is deprecated"

/cc @it-praktyk @nohwnd

test/powershell/Host/PSVersionTable.Tests.ps1
```diff
@@ -25,51 +25,51 @@ Describe "PSVersionTable" -Tags "CI" {
...
-       $PSVersionTable.ContainsKey("PSVersion")                 | Should Be True
-       $PSVersionTable.ContainsKey("PSEdition")                 | Should Be True
-       $PSVersionTable.ContainsKey("WSManStackVersion")         | Should Be True
-       $PSVersionTable.ContainsKey("SerializationVersion")      | Should Be True
-       $PSVersionTable.ContainsKey("PSCompatibleVersions")      | Should Be True
-       $PSVersionTable.ContainsKey("PSRemotingProtocolVersion") | Should Be True
-       $PSVersionTable.ContainsKey("GitCommitId")               | Should Be True
-       $PSVersionTable.ContainsKey("Platform")                  | Should Be True
-       $PSVersionTable.ContainsKey("OS")                        | Should Be True
+       $PSVersionTable.ContainsKey("PSVersion")                 | Should -Be True
```
iSazonov: We can use `-BeTrue`.

test/powershell/Host/ConsoleHost.Tests.ps1
```diff
@@ -115,20 +115,17 @@ Describe "ConsoleHost unit tests" -tags "Feature" {
...
             }
         }
 
-        It "Verify Validate Output Format As Text Explicitly Child Single Shell should works" {
-            {
-                $a="blahblah"
-                $a | & $powershell -noprofile -out text -com { $input }
-            } | Should Not Throw
+        It "Verify Validate Output Format As Text Explicitly Child Single Shell doess not throw" {
+            "blahblah" | & $powershell -noprofile -out text -com { $input }
```
iSazonov: Typo in title.

iSazonov (changes requested):

bergmeister:
@iSazonov  See [here](https://github.com/pester/Pester/releases/tag/4.2.0) (bottom section about breaking changes) about why `Should -Not -Throw` should not be used any more. I agree with your other suggestions.

iSazonov:
If you means:
>Should -Not -Throw will fail if any exception is thrown. The filters applied on Should -Not -Throw do not apply, because when an exception is thrown, it automatically means that the test should not pass.

I don't see where said that it is "deprecated". My understanding is that with the change we get expected behavior "Should -Not -Throw _will fail_ if any exception is thrown."

bergmeister:
I would like to hear first what the Pester guys suggest. I read it first the sentence `Should -Not -Throw will fail if any exception is thrown.` as that the operator itself will fail but you might be right that they changed it to make it work. But in general I do not understand why such an assertion is needed at all. Does a Pester test not fail automatically if an terminating or non-terminating error occurs?

iSazonov:
Pester `Should` like C# [Fact] in Unit tests - we should _explicitly_ see that we expect. Also if we remove `Should` we (can) get noisy output and we have to suppress it by redirection to null.

nohwnd:
`Should -Not -Throw` is not deprecated. Only the behavior changed. Before, `Should -Not -Throw`was a negation of `Should -Throw`, which allowed it to catch an exception, and still pass the test. So I corrected the behavior to fail if any exception is thrown. And now I think I should make the assertion fail, if `-Not` and any other filter is used.

The change of behavior no longer allows for writing tests that say that "code should not throw a particular exception, but any other exception is fine". This is a sacrifice I am willing to make, as this kind of test does not make much sense, and is easy to replicate via `Should -Throw`. 

Personally I would remove `Should -Not -Throw` from the assertion library completely, if it was easily possible. Because as with any other .NET test framework, any exception thrown within a test fails the test. So `Should -Not -Throw` is implied on every line of code that we write, and there is no need to explicitly call it out. In fact it only help us to write incomplete tests, because it looks like we have an assertion, while we should be testing for a result or an effect instead.

@iSazonov I am not sure, what you mean by your last remark about `Should` and `Fact`. Should is an assertion, but Fact is an attribute that mark a test in xUnit, did you mean Assert?. Anyways removing Should -Not -Throw, shouldn't be a problem.

iSazonov:
@nohwnd Thanks for confirmation.
I means that "Should" brings useful information - we _explicitly_ see the expected result. If this is not present, the `It` block looks like useless - we can not know that the expected result of code execution should be an exception - the first thought is that we skip anything.

nohwnd:
@iSazonov 
> we explicitly see the expected result.

I cannot agree with that. That code does not throw is not a good result of a test. A test is supposed to validate that something useful happened. A function can be useful by either returning the correct result, or producing the correct side effect. (eg. adding 1+1 and making sure you got 2, or deleting a file and making sure it's deleted).  

A function that does not return any value, and does not have any side effects is not useful. (but feel free to prove me wrong :) )

Coming back to our useful functions we can see that pre-requisites for successfully returning the correct value, or producing the correct side-effect is that the code does not throw. So a test that only validates that you did not throw, is just a subset of a more useful test that would test the resulting value, or the resulting side-effect.

( If this does not make a sense, you can view it from TDD point of view. A test is a mean that allows you to force your code to have the desired behavior. By writing a test that validates that your function returns 2 for input 1 and 1 you are forcing the function to implement some behavior, because by default it does not return anything. BUT by expecting the code to not throw, you are not forcing anything new, because empty functions do not throw by default. )

iSazonov:
This logic is perfectly correct in the general case. We would have to do all our tests that way. But we have _edge_ cases where we are forced to use this "Should -Not -Throw". For example, in this PR we say that modified psversiontable can be formatted, but we do not check the exact result - such a test would be unduly complicated, so we just check that there is no exception - then we _trust_ formatting system tests that exist separately. There are other examples when there is output to the console, user interaction, changing system settings (Set-Date). So "Should -Not -Throw" helps us to simplify tests if it is justified or impossible otherwise.

bergmeister:
So what is the outcome? Do I need to revert the removal of `Should -Not -Throw`?

iSazonov:
@bergmeister Yes, please revert - we'll use this until we make a decision to replace such edge cases with full positive tests.

iSazonov (approved):

daxian-dbw (approved):
LGTM

bergmeister:
@daxian-dbw Please tick the corresponding checkbox in the referenced issue 6245

PR #6310, opened 2018-03-04
Added Service Point Manager call to force Tls12.

DarqueWarrior:
## PR Summary
Calls [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12 during bootstrap to download PowerShell core to avoid following error.

```
PS C:\Users\dlbm3\source\repos\PowerShell> Start-PSBootstrap
Installing PowerShell build dependencies
dotnet is already installed.  Skipping installation.
pwsh.exe not found. Install latest PowerShell Core release and add it to Path
VERBOSE: Destination: C:\Users\dlbm3\AppData\Local\Microsoft\powershell
VERBOSE: About to download package from 'https://github.com/PowerShell/PowerShell/releases/download/v6.0.1/PowerShell-6.0.1-win-x64.zip'
Invoke-WebRequest : The request was aborted: Could not create SSL/TLS secure channel.
At C:\Users\dlbm3\source\repos\PowerShell\tools\install-powershell.ps1:139 char:9
+         Invoke-WebRequest -Uri $downloadURL -OutFile $packagePath
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:HttpWebRequest) [Invoke-WebRequest], WebException
    + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeWebRequestCommand
```

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/tools/install-powershell.ps1
+++ b/tools/install-powershell.ps1
@@ -137,7 +137,7 @@ try {
 
         # Setting Tls to 12 to prevent the Invoke-WebRequest : The request was
         # aborted: Could not create SSL/TLS secure channel. error.
-        [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+        [Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
 
         $packagePath = Join-Path -Path $tempDir -ChildPath $packageName
         Invoke-WebRequest -Uri $downloadURL -OutFile $packagePath
--- a/tools/install-powershell.ps1
+++ b/tools/install-powershell.ps1
@@ -135,6 +135,10 @@ try {
         $downloadURL = "https://github.com/PowerShell/PowerShell/releases/download/v${release}/${packageName}"
         Write-Verbose "About to download package from '$downloadURL'" -Verbose
 
+        # Setting Tls to 12 to prevent the Invoke-WebRequest : The request was
+        # aborted: Could not create SSL/TLS secure channel. error.
+        [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+
         $packagePath = Join-Path -Path $tempDir -ChildPath $packageName
         Invoke-WebRequest -Uri $downloadURL -OutFile $packagePath
         $contentPath = Join-Path -Path $tempDir -ChildPath "new"
--- a/tools/install-powershell.ps1
+++ b/tools/install-powershell.ps1
@@ -91,6 +91,11 @@ $architecture = if (-not $IsWinEnv) {
 $tempDir = Join-Path ([System.IO.Path]::GetTempPath()) ([System.IO.Path]::GetRandomFileName())
 New-Item -ItemType Directory -Path $tempDir -Force > $null
 try {
+    # Setting Tls to 12 to prevent the Invoke-WebRequest : The request was
+    # aborted: Could not create SSL/TLS secure channel. error.
+    $originalValue = [Net.ServicePointManager]::SecurityProtocol
+    [Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
+
     if ($Daily) {
         if (-not (Get-Module -Name PackageManagement -ListAvailable)) {
             throw "PackageManagement module is required to install daily PowerShell Core."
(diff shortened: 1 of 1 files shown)


tools/install-powershell.ps1
```diff
@@ -135,6 +135,10 @@ try {
         $downloadURL = "https://github.com/PowerShell/PowerShell/releases/download/v${release}/${packageName}"
         Write-Verbose "About to download package from '$downloadURL'" -Verbose
 
+        # Setting Tls to 12 to prevent the Invoke-WebRequest : The request was
+        # aborted: Could not create SSL/TLS secure channel. error.
+        [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
```
iSazonov: I believe we should use the same from #6236 
[Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12

markekraus: Yes.. and for `Install-` the original setting needs to be reverted. after the calls.

TravisEz13 (changes requested):
please address comments for @iSazonov

DarqueWarrior:
I think I have addressed all the review comments.

tools/install-powershell.ps1
```diff
@@ -137,12 +137,16 @@ try {
 
         # Setting Tls to 12 to prevent the Invoke-WebRequest : The request was
         # aborted: Could not create SSL/TLS secure channel. error.
-        [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+        $originalValue = [Net.ServicePointManager]::SecurityProtocol
+        [Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
```
TravisEz13: This should be tls1.0, tls1.1 and tls1.2.  Then no need to revert

TravisEz13 (changes requested):

tools/install-powershell.ps1
```diff
@@ -135,10 +135,18 @@ try {
...
+        # aborted: Could not create SSL/TLS secure channel. error.
+        $originalValue = [Net.ServicePointManager]::SecurityProtocol
+        [Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
+
         $packagePath = Join-Path -Path $tempDir -ChildPath $packageName
         Invoke-WebRequest -Uri $downloadURL -OutFile $packagePath
         $contentPath = Join-Path -Path $tempDir -ChildPath "new"
 
+        # Restore original value
+        [Net.ServicePointManager]::SecurityProtocol = $originalValue
```
markekraus: The `Invoke-WebRequest` call should be wrapped in a Try/Finally. Pseudo code:

```powershell
$originalValue = [Net.ServicePointManager]::SecurityProtocol
[Net.ServicePointManager]::SecurityProtocol = ...
try {
    Invoke-WebRequest ...
}
finally {
    [Net.ServicePointManager]::SecurityProtocol = $originalValue
}
```

markekraus (changes requested):
Please use a try/finally

tools/install-powershell.ps1
markekraus: I'm torn. Enabling it in the boot strap function makes sense because we are setting up the PowerShell environment for builds and tests. Leaving the change in the install code means we are modifying the user's PowerShell environment. That's not exactly what I would expect (even though it is a non-breaking environment change that arguably increases their security....).

TravisEz13: It is what the defaults should be... you could argue that 1.0 should not be included.  I included it to be more compatible.  If you are connecting to anything older, there are serious issues.

daxian-dbw:
Please note that there is another `Invoke-WebRequset` call at line 181:
```
Invoke-WebRequest -OutFile "~/.rcedit/rcedit-x64.exe" -Uri $rceditUrl
```
This call downloads `rcedit-x64.exe` from github.

TravisEz13:
@daxian-dbw comments still need to be addressed.

tools/install-powershell.ps1
```diff
@@ -135,8 +135,21 @@ try {
...
 
+        # Setting Tls to 12 to prevent the Invoke-WebRequest : The request was
+        # aborted: Could not create SSL/TLS secure channel. error.
+        $originalValue = [Net.ServicePointManager]::SecurityProtocol
+        [Net.ServicePointManager]::SecurityProtocol = [Net.ServicePointManager]::SecurityProtocol -bor [Net.SecurityProtocolType]::Tls12
+
         $packagePath = Join-Path -Path $tempDir -ChildPath $packageName
-        Invoke-WebRequest -Uri $downloadURL -OutFile $packagePath
+
+        try {
```
daxian-dbw: Given that there is another `Invoke-WebRequest` call below, I suggest that we move the configuration of "SecurityProtocol" before line 93 (the outmost `try` block), and then revert to original protocol value in the `finally` block at the very end.

daxian-dbw (changes requested):

tools/install-powershell.ps1
markekraus: I agree that's what the default settings _should_ be.... unfortunately, they are not the defaults. It is possible that changing these settings without reverting (and in a way that is somewhat hidden from the user) could potentially break their environment. There are some sites that di respond differently on TLS 1.0, 1.1, and 1.2. That's why I think the context matters as to whether the settings changes are left or reverted. `Start-PSBootStrap` is a "configure my environment to be able to build, debug, and test PowerShell" function where as `install-powershell.ps1` is a script for adding pwsh to the environment. In the former I expect the environment to change, in the later, the only environment change I expect is to have pwsh installed. *shrug

DarqueWarrior:
Moved the code to the outer try and finally block as requested.

markekraus (approved):

daxian-dbw (approved):

Issue #4605, opened 2017-08-17
Can we get a release that runs on Alpine?


JasonMorgan:
Could we get an apk package for alpine linux?  The main reason we don't use powershell as part of our ci/cd pipeline is that all our tasks run in alpine containers.

SteveL-MSFT:
Are you able to use the AppImage?

JasonMorgan:
I tried and couldn't get it to work.  I don't see it in their list of supported os's.

markekraus:
I took a stab at running the Appimage in Alpine. It looks like you could do it only if you have a full glibc chroot. The primary issue seems to be the musl libc that alpine uses and Appimage is made for glibc environments. It certainly increases the size of Alpine if you have to tack on a full glibc chroot to run the Appimage.

JasonMorgan:
@SteveL-MSFT any thoughts on this? Adding an apk package would make it so I could start using it in most of my tools containers and I'm sure I'm not the only one.

SteveL-MSFT:
@JasonMorgan Alpine isn't one of the official supported OS for corefx, so for us this isn't currently a priority.  If a community PR makes it work, we'd be happy to take that change and start producing packages.

SteveL-MSFT:
Looks like Alpine is sort of supported by CoreCLR, they have a build, but it doesn't appear to be tested regularly.

markekraus:
dotnet/coreclr#917 It looks like they are still working the kinks out.

markekraus:
dotnet/dotnet-docker#22 looks like they added an alpine docker for .NET Core. https://github.com/dotnet/dotnet-docker-samples/tree/master/dotnetapp-prod-alpine-preview so it should be theoretically possible to get a pwsh docker from that now.

kasper3:
How can we make this theory a fact? Something like:

```
docker run ...something?
apk install git
git clone git@github.com:PowerShell/PowerShell
./PowerShell/tools/install-powershell.sh
```

or are there any dependencies involved? Please publish an additional docker layer for PS sitting on top of that previewed docker from .NET Core.

JasonMorgan:
@kasper3 I can confirm the install script doesn't handle alpine yet.  I'm looking at building from source myself now.  I'll let you know if I get it.

JasonMorgan:
[...]

PR #6367, opened 2018-03-12
Add Alpine Linux support

ghost:
## PR Summary

`sys/sysctl` header is not available in musl-libc based systems and system calls are available without any header inclusion. I have ported the cmake test from CoreFX repo (https://github.com/dotnet/corefx/blob/431475b8/src/Native/Unix/configure.cmake#L560).

Progress towards: #4605

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/libpsl-native/src/CMakeLists.txt
+++ b/src/libpsl-native/src/CMakeLists.txt
@@ -1,3 +1,5 @@
+include(CheckIncludeFiles)
+
 add_library(psl-native SHARED
   getstat.cpp
   getpwuid.cpp
@@ -23,5 +25,13 @@ add_library(psl-native SHARED
   followsymlink.cpp
   createprocess.cpp
   nativesyslog.cpp)
+
+check_include_files(
+    "sys/types.h;sys/sysctl.h"
+    HAVE_SYS_SYSCTL_H)
+
+configure_file(
+    ${CMAKE_CURRENT_SOURCE_DIR}/pal_config.h.in
+    ${CMAKE_CURRENT_BINARY_DIR}/pal_config.h)
 
 target_include_directories(psl-native PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
--- a/src/libpsl-native/src/getppid.cpp
+++ b/src/libpsl-native/src/getppid.cpp
@@ -1,11 +1,15 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#include "pal_config.h"
 #include "getppid.h"
 
 #include <sys/user.h>
 #include <sys/param.h>
+
+#if HAVE_SYS_SYSCTL_H
 #include <sys/sysctl.h>
+#endif
 
 //! @brief GetPPid returns the parent process id for a process
 //!
--- a/src/libpsl-native/src/getuserfrompid.cpp
+++ b/src/libpsl-native/src/getuserfrompid.cpp
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 #include "pal.h"
+#include "pal_config.h"
 #include "getfileowner.h"
 #include "getpwuid.h"
 #include "getuserfrompid.h"
@@ -9,7 +10,10 @@
 #include <string>
 #include <sstream>
 #include <errno.h>
+
+#if HAVE_SYS_SYSCTL_H
 #include <sys/sysctl.h>
+#endif
 
 char* GetUserFromPid(pid_t pid)
 {
--- a/src/libpsl-native/src/pal_config.h.in
+++ b/src/libpsl-native/src/pal_config.h.in
@@ -1,0 +1,3 @@
+#pragma once
+
+#cmakedefine01 HAVE_SYS_SYSCTL_H
--- a/tools/releaseBuild/Images/microsoft_powershell_alpine3/Dockerfile
+++ b/tools/releaseBuild/Images/microsoft_powershell_alpine3/Dockerfile
@@ -1,0 +1,10 @@
+# Docker image file that describes an Centos7 image with PowerShell installed from Microsoft YUM Repo
+
+FROM microsoft/dotnet:2.1-sdk-alpine
+LABEL maintainer="PowerShell Team <powershellteam@hotmail.com>" 
+
+# Install dependencies and clean up
+RUN apk update \
+    && apk add cmake clang build-base git bash
+
+COPY build-and-run-pwsh.sh /
--- a/tools/releaseBuild/Images/microsoft_powershell_alpine3/build-and-run-pwsh.sh
+++ b/tools/releaseBuild/Images/microsoft_powershell_alpine3/build-and-run-pwsh.sh
@@ -1,20 +1,27 @@
 #!/bin/sh
 
-apk update
-apk add cmake clang build-base git bash
-git clone --recursive https://github.com/kasper3/powershell
-cd powershell/src/libpsl-native
+repoRoot=$1
+destination=$2
+releaseTag=$3
+
+# Build libpsl-native
+cd $repoRoot/src/libpsl-native
 
 cmake -DCMAKE_BUILD_TYPE=Debug .
 make -j
 
+# Restore packages
 cd ../..
 dotnet restore
+
+# Add telemetry file
 touch DELETE_ME_TO_DISABLE_CONSOLEHOST_TELEMETRY
 
+# run ResGen
 cd src/ResGen
 dotnet run
 
+# Create typeCatalog
 cd ..
 targetFile="Microsoft.PowerShell.SDK/obj/Microsoft.PowerShell.SDK.csproj.TypeCatalog.targets"
 cat > $targetFile <<-"EOF"
@@ -28,12 +35,21 @@ cat > $targetFile <<-"EOF"
   </Target>
 </Project>
 EOF
+
 dotnet msbuild Microsoft.PowerShell.SDK/Microsoft.PowerShell.SDK.csproj /t:_GetDependencies "/property:DesignTimeBuild=true;_DependencyFile=$(pwd)/TypeCatalogGen/powershell.inc" /nologo
 
 cd TypeCatalogGen
 dotnet run ../System.Management.Automation/CoreCLR/CorePsTypeCatalog.cs powershell.inc
 
+# build PowerShell
 cd ../powershell-unix
 dotnet publish --configuration Linux --runtime linux-x64
-mv libpsl-native.so bin/Linux/netcoreapp2.0
-dotnet run -c Linux
+
+# add libpsl-native to build
+mv libpsl-native.so bin/Linux/netcoreapp2.0/linux-x64
+
+# tar build for output
+cd bin/Linux/netcoreapp2.0/linux-x64
+
+# TODO make format of file name the same as other packages
+tar -czvf $destination/powershell-alpine.3.tar.gz .
--- a/tools/releaseBuild/Images/microsoft_powershell_alpine3/build-and-run-pwsh.sh
+++ b/tools/releaseBuild/Images/microsoft_powershell_alpine3/build-and-run-pwsh.sh
@@ -1,0 +1,39 @@
+#!/bin/sh
+
+apk update
+apk add cmake clang build-base git bash
+git clone --recursive https://github.com/kasper3/powershell
+cd powershell/src/libpsl-native
+
+cmake -DCMAKE_BUILD_TYPE=Debug .
+make -j
+
+cd ../..
+dotnet restore
+touch DELETE_ME_TO_DISABLE_CONSOLEHOST_TELEMETRY
+
+cd src/ResGen
+dotnet run
+
+cd ..
+targetFile="Microsoft.PowerShell.SDK/obj/Microsoft.PowerShell.SDK.csproj.TypeCatalog.targets"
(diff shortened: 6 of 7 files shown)


kasper3:
cc @SteveL-MSFT 

This is how i built it:

```sh
docker run -it microsoft/dotnet:2.1-sdk-alpine

# in container
apk update
apk add cmake clang build-base git bash
git clone --recursive https://github.com/kasper3/powershell
cd powershell/src/libpsl-native

cmake -DCMAKE_BUILD_TYPE=Debug .
make -j
```

then, for managed,

```sh
cd ../..
dotnet restore
dotnet publish
```

`dotnet publish` throws error:

```
  Restore completed in 8.86 ms for /powershell/src/powershell-win-core/powershell-win-core.csproj.
DotNetCode/Eventing/EventDescriptor.cs(41,61): error CS0103: The name 'DotNetEventingStrings' does not exist in the current context [/powershell/src/Microsoft.PowerShell.CoreCLR.Eventing/Microsoft.PowerShell.CoreCLR.Eventing.csproj]
DotNetCode/Eventing/EventDescriptor.cs(46,103): error CS0103: The name 'DotNetEventingStrings' does not exist in the current context [/powershell/src/Microsoft.PowerShell.CoreCLR.Eventing/Microsoft.PowerShell.CoreCLR.Eventing.csproj]
DotNetCode/Eventing/EventDescriptor.cs(58,63): error CS0103: The name 'DotNetEventingStrings' does not exist in the current context [/powershell/src/Microsoft.PowerShell.CoreCLR.Eventing/Microsoft.PowerShell.CoreCLR.Eventing.csproj]
DotNetCode/Eventing/EventDescriptor.cs(63,105): error CS0103: The name 'DotNetEventingStrings' does not exist in the current context [/powershell/src/Microsoft.PowerShell.CoreCLR.Eventing/Microsoft.PowerShell.CoreCLR.Eventing.csproj]
DotNetCode/Eventing/EventProviderTraceListener.cs(45,49): error CS0103: The name 'DotNetEventingStrings' does not exist in the current context [/powershell/src/Microsoft.PowerShell.CoreCLR.Eventing/Microsoft.PowerShell.CoreCLR.Eventing.csproj]
DotNetCode/Eventing/EventProviderTraceListener.cs(76,45): error CS0103: The name 'DotNetEventingStrings' does not exist in the current context [/powershell/src/Microsoft.PowerShell.CoreCLR.Eventing/Microsoft.PowerShell.CoreCLR.Eventing.csproj]
DotNetCode/Eventing/EventProvider.cs(627,75): error CS0103: The name 'DotNetEventingStrings' does not exist in the current context [/powershell/src/Microsoft.PowerShell.CoreCLR.Eventing/Microsoft.PowerShell.CoreCLR.Eventing.csproj]
DotNetCode/Eventing/EventProvider.cs(664,83): error CS0103: The name 'DotNetEventingStrings' does not exist in the current context [/powershell/src/Microsoft.PowerShell.CoreCLR.Eventing/Microsoft.PowerShell.CoreCLR.Eventing.csproj]
```

kasper3:
Following this guide https://github.com/PowerShell/PowerShell/blob/76526c6f1df08c0a55fbae79a7148a4a58805626/docs/building/internals.md, i made a breakthrough. Now stuck with another error (updated previous comment)

kasper3:
Fixed, had to add `touch DELETE_ME_TO_DISABLE_CONSOLEHOST_TELEMETRY` at repo root. Strange requirement 😝

SteveL-MSFT:
Got it working:

```powershell
GitCommitId                    v6.0.1-252-g982dbfe552b41ec3cd9b64f649849e8b942643d6
OS                             Linux 4.9.60-linuxkit-aufs #1 SMP Mon Nov 6 16:00:12 UTC 2017
```

kasper3:
Hey Steve, after `dotnet publish --runtime linux-x64`, what exact command you ran to get into the shell, i am still trying to figure that out 😮

SteveL-MSFT:
@kasper3 I had to use `dotnet publish --configuration Release --runtime linux-x64`.  Also made some changes to some .csproj files, but need to determine if those were actually necessary.  To run use `dotnet run`

SteveL-MSFT:
@kasper3 here's the updated instructions from your list above:

```bash
# remove your last two lines and replace with
cd ../powershell-unix
dotnet publish --configuration Linux --runtime linux-x64
mv libpsl-native.so bin/Linux/netcoreapp2.0
dotnet run -c Linux
```

Perhaps for now, we can have this as a bash script until we have a build available then we can update `build.psm1` to work correctly on Alpine (it makes assumptions about being on Ubuntu)

kasper3:
Thanks, it worked! I will add the build script with instructions in a separate commit today.

SteveL-MSFT:
@kasper3 if you can update the PR description above to use the standard PR template, we can take this one just to get libpsl-native building on Alpine

kasper3:
@SteveL-MSFT, done.

Please note that we are not bound to use that particular docker provided by Microsoft/dotnet, we can use a latest alpine one (or install Alpine 3.7 in VM or even bare-metal) and setup fully functional dotnet sdk like this described here: https://github.com/dotnet/cli/issues/8767. There is one additional thing I had to do (create a nuget.config, before creating the first app, since it's depending on preview2 packages, which are not yet available in nuget). Here is the final setup:

```sh
docker run -it alpine:latest
```

then in container

```sh
apk update
apk add curl libstdc++ libunwind libintl icu-libs openssl

curl -O https://dotnetcli.blob.core.windows.net/dotnet/Sdk/master/dotnet-sdk-latest-alpine.3.6-x64.tar.gz
mkdir -p /usr/bin/dotnet
tar -xzvf dotnet-sdk-latest-alpine.3.6-x64.tar.gz  -C /usr/bin/dotnet
export PATH=$PATH:/usr/bin/dotnet
rm dotnet-sdk-latest-alpine.3.6-x64.tar.gz

targetFile="/root/.nuget/NuGet/NuGet.Config"
cat > $targetFile <<-"EOF"
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <packageSources>
    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
 <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
  </packageSources>
</configuration>
EOF
```

Then we can use `dotnet new console` etc. and continue installing `git` and stuff to compiler powershell.

Should we have two scripts `obtain-dotnet-alpine.sh` and `build-pwsh-alpine.sh`? Preferably, those scripts could use Bourne Shell shebang `#!/bin/sh` instead of `bash`, because:

1. there is no `bash`ism in these simple commands
2. as an example, all Unix shell scripts shipped by npm and node.js are `/bin/sh`. .NET Core heavily used bash but now they are slowly converting to sh as well.
3. in latest Ubuntu, default shell is dash, in Alpine default shell is ash, in NetBSD, default shell is csh etc... and all of those require extra installation of `bash`. The only common shell between all kinds of unices is POSIX complaint Bourne Shell.. it gets bit harder to convert existing bash script to bourne shell but we can try not to introduce more `bash`ism than required and stay impartial. We can also try to convert existing .sh files to `bin/sh` at slow pace. ✌️

krytarowski:
`#!/bin/sh` would be nice.

A correction - upstream for tcsh is NetBSD, but the default shell is Almquist Shell (ash / sh). Debian's dash is a fork of NetBSD's ash.

TravisEz13:
dotnet said they fixed `install-dotnet.sh`, but I don't know how long it takes to go into production

kasper3:
@TravisEz13, can't `docker run -it microsoft/dotnet:2.1-sdk-alpine` be used? Or does PowerShell repo can't use docker and requires it to be available in a VM or bare-metal?

TravisEz13:
@kasper3 I'd have to see a working script.
I would recommend you add any custom script to build this in `tools/releaseBuild/images`   The idea is that the build is described in `tools/releaseBuild/build.json`.  The readme in `tools/releaseBuild` describes how to call a build described in `build.json`
If you even just add a docker file that does the build in that folder, I'm happy to refactor it into the build.json

Hopefully, we can get something working in `build.psm1` eventually too.

kasper3:
@TravisEz13, I have assembled the pieces together and added a naive build script for Alpine. I couldn't quite figure out the Docker build image integration. This script clones the repository and build and run pwsh. Here is one way to run it:

```sh
# in windows 10 powershell
git clone https://github.com/kasper3/powershell
cd powershell/tools/releaseBuild/images/microsoft_powershell_alpine3
docker run -it -v $pwd\:/host microsoft/dotnet:2.1-sdk-alpine

# in alpine-dotnet container
/host/build_and_run_pwsh.sh

# after sometime you will enter powershell
```

TravisEz13:
@kasper3 I think I can work with this.  I have some deadlines.  It may take me some time.  If I don't reply my next month, can you ping me?

kasper3:
@TravisEz13, sure thanks!

TravisEz13:
@kasper3 The build command from the root of the repo is `./tools/releaseBuild/vstsbuild.ps1 -Name alpine.3 -ReleaseTag v6.1.0-alpine3`

There are still two things to do:
 - [ ] the package should have the same format as the other packages (should be `powershell-6.0.2-alpine.3-x64.tar.gz`)
 - [ ] the build should be triggered with the release tag ([build.psm1 example](https://github.com/PowerShell/PowerShell/blob/master/build.psm1#L524)

kasper3:
This is a great progress! TIL, thanks.

I was thinking that maybe we could try using a mini rootfs: https://alpinelinux.org/downloads/ in hyperv or vmwhere (it's 1.9 mbs of gz), and test if it works in non-Docker as well. I heard yesterday from musl-libc core team member (@rofl) something in between the lines of "don't just trust that it works on Alpine just because it works on Alpine docker, they have things like `libbacktrace` and `sys/queue` used in docker to make it work with underlying kernel, test on real thing too!", hence my suggestion-cum-curiosity. :)

TravisEz13:
@kasper3 that sound like a great idea to investigate.  I would get this working as is first.  I think if we get those things done since this is an unofficial package, we can merge the PR.

Once, we have released a package, we should produce a docker file.

TravisEz13:
@kasper3   The regex is here: https://github.com/PowerShell/PowerShell/blob/master/tools/packaging/packaging.psm1#L16
but either will work... mine was just an example.  `v6.1.0-preview.1` will be the next tag then after a series of non-stable releases we will eventually get to `v6.1.0`

kasper3:
Thanks. I have made some changes locally and created the tarball with the matching name. The final tarball builds but..

We need to build other native components as well, `pwsh` itself, and maybe https://github.com/PowerShell/psl-omi-provider, https://github.com/Microsoft/omi? We cannot rely on prebuild Linux binaries, the tarball being produced is not distributable.

`libhostfxr.so` and other `.so` files except for the one we built *on Alpine* `libpsl-native.so` are corrupted when we try to consume pwsh from the tarball. `pwsh` itself is corrupted as it was prebuilt against glibc.

The main difference in Alpine Linux and other distros is that it uses different libc, i.e. musl-libc. So we need to compile it from ground up. .NET Team is even considering changing the inheritance of Alpine Linux RID from `unixx64->linuxx64->alpinex64` to `unixx54->alpinex64` or `unixx64->linuxmuslx64->alpninex64` etc. so the fallback RID is not a linuxX64 also known as `GLIBC based Linux` (regular one).

Should we block this PR until then? Is everything that is required to build `pwsh` native executable in the open? It seems like pwshexe project is Windows only in this repo and Linux binaries are coming from somewhere I couldn't figure out.

kasper3:
All the native pieces need to be recompiled on Alpine then package in tarball.

SteveL-MSFT:
@kasper3 my thinking is that this bash script is temporary to produce a working pwsh we can use on Alpine until we can update build.psm1 and packaging.psm1 to produce official Alpine builds regularly, so I wouldn't worry about the OMI aspect for now.  Regarding glibc vs musl, I would defer updating build.psm1/packaging.psm1 until dotnetcore resolves that.

rofl0r:
@kasper3 said:
> yesterday from musl-libc core team member (@rofl0r) something in between the lines of "don't trust that it works on Alpine just because it works on Alpine docker, they have things like libbacktrace and sys/queue used in docker to make it work with underlying kernel, test on real thing too!"

correction, what i actually said was conceptually: 
> "don't trust that it works on **musl libc** just because it works on Alpine docker, they have things like libbacktrace and sys/queue <strike>used in docker to make it work with underlying kernel</strike> **to go the path of least resistance instead of identifying offenders and reporting the issue upstream**, test on real thing too!"

TravisEz13:
@kasper3 my thinking is along the same line as @SteveL-MSFT 's this PR as is should get us there.  I think it's okay to add a new distribution as partially working.  Getting it to the point we can build it in the main build system will open it to more people to contribute, but this PR is an important step... I have the PR blocked with the WIP in the title, but I still think we should fix the two issue and proceed to fix the others.

kasper3:
@TravisEz13, pushed my changes and opened an issue upstream: https://github.com/Microsoft/omi/issues/497. Will try to patch those separately if i get time. :)

@rofl0r, thanks for the correction! :)

TravisEz13 (approved):

SteveL-MSFT (approved):
Great contribution!  Thanks!

PR #6392, opened 2018-03-14
Support StopProcessing in ConvertTo-Json

### LINKED ISSUE #6388: ConvertTo-Json cmdlet hangs with -Depth 10 for object created using Get-Content
Steps to reproduce
------------------

```powershell
echo "testline1" > test.txt
echo "testline2" >> test.txt

$textContent = Get-Content .\test.txt
$json=ConvertTo-Json -InputObject $textContent -Depth 10 
```

Expected behavior
-----------------

```none
ConvertTo-Json should not hang and it should display proper error or exception

```

Actual behavior
---------------

```none
ConvertTo-Json cmdlet hangs
```

Environment data
----------------

```powershell
> $PSVersionTable
PSVersion                      5.1.16299.248                                                                                
BuildVersion                   10.0.16299.248                                                                               

```

SteveL-MSFT:
## PR Summary

If ConvertTo-Json is serializing many objects or many deeply nested objects, it can appear to hang and Ctrl+C doesn't stop the cmdlet.
Fix is to check if `Stopping` is set in locations where potentially expensive processing is happening.

Fix https://github.com/PowerShell/PowerShell/issues/6388

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
@@ -116,7 +116,6 @@ namespace Microsoft.PowerShell.Commands
                 {
                     return;
                 }
-                CheckStopping();
                 JsonSerializerSettings jsonSettings = new JsonSerializerSettings { TypeNameHandling = TypeNameHandling.None, MaxDepth = 1024 };
                 if (EnumsAsStrings)
                 {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
@@ -107,7 +107,15 @@ namespace Microsoft.PowerShell.Commands
                 object objectToProcess = (_inputObjects.Count > 1) ? (_inputObjects.ToArray() as object) : (_inputObjects[0]);
                 // Pre-process the object so that it serializes the same, except that properties whose
                 // values cannot be evaluated are treated as having the value null.
-                object preprocessedObject = ProcessValue(objectToProcess, 0);
+                object preprocessedObject = null;
+                try
+                {
+                    preprocessedObject = ProcessValue(objectToProcess, 0);
+                }
+                catch (StoppingException)
+                {
+                    return;
+                }
                 CheckStopping();
                 JsonSerializerSettings jsonSettings = new JsonSerializerSettings { TypeNameHandling = TypeNameHandling.None, MaxDepth = 1024 };
                 if (EnumsAsStrings)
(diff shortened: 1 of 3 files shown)


dantraMSFT (changes requested):
The logic rely on throwing StopUpstreamCommandsException.
Perhaps a private getter around _stopping that throws when it is set.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -401,6 +429,11 @@ private ErrorRecord NewError()
         /// <returns>An object suitable for serializing to JSON</returns>
         private object ProcessValue(object obj, int depth)
         {
+            if (_stopping)
```
markekraus: hmm perhaps instead of repeating this, add a `IsStopping()` method. That way if, for whatever reason, we need to implement this in more places or change/expand the stop actions we can do it in one place.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -177,11 +175,7 @@ private string ConvertToPrettyJsonString(string json)
         /// <returns></returns>
         private int ConvertList(string json, int index, StringBuilder result, string padString, int numberOfSpaces)
         {
-            if (_stopping)
-            {
-                throw new StopUpstreamCommandsException(this);
-            }
-
+            CheckStopping();
```
dantraMSFT: I think this should be either inside in for loop or mirrored there.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -251,11 +245,7 @@ private int ConvertList(string json, int index, StringBuilder result, string pad
         /// <returns></returns>
         private int ConvertQuotedString(string json, int index, StringBuilder result)
         {
-            if (_stopping)
-            {
-                throw new StopUpstreamCommandsException(this);
-            }
-
+            CheckStopping();
```
dantraMSFT: Like above, consider moving this inside the for loop.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -296,11 +286,7 @@ private int ConvertQuotedString(string json, int index, StringBuilder result)
         /// <returns></returns>
         private int ConvertDictionary(string json, int index, StringBuilder result, string padString, int numberOfSpaces)
         {
-            if (_stopping)
-            {
-                throw new StopUpstreamCommandsException(this);
-            }
-
+            CheckStopping();
```
dantraMSFT: Again, inside the for loop.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -619,11 +601,7 @@ private void AppendPsProperties(PSObject psobj, IDictionary receiver, int depth,
         /// <returns></returns>
         private object ProcessDictionary(IDictionary dict, int depth)
         {
-            if (_stopping)
-            {
-                throw new StopUpstreamCommandsException(this);
-            }
-
+            CheckStopping();
```
dantraMSFT: Consider inside the for loop.

dantraMSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
SteveL-MSFT: The only potential expensive calls in that loop are ConvertDictionary() and ConvertList() which already handles stopping now

dantraMSFT (approved):

BrucePay (dismissed):
The StopProcessing() overload should be removed and the existing Stopping property on the cmdlet should be used instead of creating a new _stopping property.

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Json.Tests.ps1
```diff
@@ -16,4 +16,19 @@ Describe 'ConvertTo-Json' -tags "CI" {
...
+        $tmpFile = Join-Path $TestDrive "test.txt"
+        Set-Content -Path $tmpFile -Value "hello"
+        $ps = [PowerShell]::Create()
+        $null = $ps.AddCommand("Get-Content")
+        $null = $ps.AddParameter("Path", $tmpFile)
+        $null = $ps.AddCommand("ConvertTo-Json")
+        $null = $ps.AddParameter("Depth", 10)
+        $null = $ps.BeginInvoke()
+        Start-Sleep -Milliseconds 100
+        $null = $ps.Stop()
```
markekraus: I tested this with 6.0.1 and a less extreme example and it passed. In a regression this could potentially block the pester thread for a long time and give a false pass. Perhaps we should do something closer to this:

```powershell
$null = $ps.BeginStop($null, $null)
Start-Sleep -Milliseconds 100
$ps.InvocationStateInfo.State | Should -Be "Stopped"
$ps.Dispose()
```

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Json.Tests.ps1
```diff
@@ -16,4 +16,19 @@ Describe 'ConvertTo-Json' -tags "CI" {
...
+        $ps.InvocationStateInfo.State | should be "Stopped"
```
markekraus: `Should -Be`

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Json.Tests.ps1
```diff
@@ -16,4 +16,19 @@ Describe 'ConvertTo-Json' -tags "CI" {
         $jsonFormat | Should Match '"TestValue2": 78910'
         $jsonFormat | Should Match '"TestValue3": 99999'
     }
+
+	It "StopProcessing should succeed" {
+        $tmpFile = Join-Path $TestDrive "test.txt"
+        Set-Content -Path $tmpFile -Value "hello"
+        $ps = [PowerShell]::Create()
+        $null = $ps.AddCommand("Get-Content")
```
markekraus: I've a minor concern that if we end up changing the behavior of `Get-Content`, This test will be a false pass. Perhaps it would be better to generate a deep object on the fly that we know will always be deep?

```powershell
$ps = [PowerShell]::Create()
$null = $ps.AddScript({
    $obj = [PSCustomObject]@{P1 = ''; P2 = ''; P3 = ''; P4 = ''; P5 = ''; P6 = ''}
    $obj.P1 = $obj.P2 = $obj.P3 = $obj.P4 = $obj.P5 = $obj.P6 = $obj
    1..100 | Foreach-Object { $obj } | ConvertTo-Json -Depth 10
})
$handler = $ps.BeginInvoke()
```

markekraus (changes requested):
Perhaps this is a better test:

```powershell
Set-Content -Path $tmpFile -Value "hello"
$ps = [PowerShell]::Create()
$null = $ps.AddScript({
    $obj = [PSCustomObject]@{P1 = ''; P2 = ''; P3 = ''; P4 = ''; P5 = ''; P6 = ''}
    $obj.P1 = $obj.P2 = $obj.P3 = $obj.P4 = $obj.P5 = $obj.P6 = $obj
    1..100 | Foreach-Object { $obj } | ConvertTo-Json -Depth 10
})
$handler = $ps.BeginInvoke()
Start-Sleep -Milliseconds 100
$null = $ps.BeginStop($null, $null)
Start-Sleep -Milliseconds 100
$ps.InvocationStateInfo.State | Should -Be "Stopped"
$ps.Dispose()
```

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -108,20 +108,33 @@ protected override void EndProcessing()
...
+        private bool _stopping = false;
+
+        /// <summary>
+        /// Allow Ctrl+C to stop processing.
+        /// </summary>
+        protected override void StopProcessing()
+        {
+            _stopping = true;
+        }
+
```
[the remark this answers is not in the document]
SteveL-MSFT: Will fix.  Wasn't familiar with the Stopping property.

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Json.Tests.ps1
SteveL-MSFT: Will use your test above

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Json.Tests.ps1
```diff
@@ -16,4 +16,21 @@ Describe 'ConvertTo-Json' -tags "CI" {
...
+        $ps = [PowerShell]::Create()
+        $null = $ps.AddScript({
+            $obj = [PSCustomObject]@{P1 = ''; P2 = ''; P3 = ''; P4 = ''; P5 = ''; P6 = ''}
+            $obj.P1 = $obj.P2 = $obj.P3 = $obj.P4 = $obj.P5 = $obj.P6 = $obj
+            1..100 | Foreach-Object { $obj } | ConvertTo-Json -Depth 10
+            # the conversion is expected to take some time, this throw is in case it doesn't
+            throw "Should not have thrown exception"
+        })
+        $null = $ps.BeginInvoke()
+        Start-Sleep -Milliseconds 100
```
iSazonov: Is the tests time sensitive? Will it work stably on slow CI system?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -683,5 +690,16 @@ private object ProcessCustomObject<T>(object o, int depth)
             }
             return result;
         }
+
+        /// <summary>
+        /// Check if Stopping is set and throw special exception to stop
```
iSazonov: Please add final dot in public comment.

SteveL-MSFT: Will add

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Json.Tests.ps1
SteveL-MSFT: The sleep is to ensure the script has reached `ConvertTo-Json`.  I think 100ms is suffiicient even on CI.  I can add a comment to the code.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -683,5 +690,16 @@ private object ProcessCustomObject<T>(object o, int depth)
...
         }
+
+        /// <summary>
+        /// Check if Stopping is set and throw special exception to stop.
+        /// </summary>
+        private void CheckStopping()
+        {
+            if (Stopping)
+            {
+                throw new StopUpstreamCommandsException(this);
```
daxian-dbw: I don't think we should use this exception. It's supposed to be thrown from a `ProcessRecord` method, to stop the upstream `Process` blocks from running. For the `ConvertTo-Json` cmdlet, we do the real work at `EndProcessing`, and by that time all upstream commands have finished running.

I suggest you use a custom exception instead, can catch the exception in `EndProcessing` when calling `ProcessValue`, like this:
```c#
object preprocessedObject = null;
try
{
    preprocessedObject = ProcessValue(objectToProcess, 0);
}
catch (MyCustomException ex)
{
    return;
}
```

daxian-dbw (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -174,6 +175,7 @@ private int ConvertList(string json, int index, StringBuilder result, string pad
 
             for (int i = index; i < json.Length; i++)
             {
+                CheckStopping();
```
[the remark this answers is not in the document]
SteveL-MSFT: That probably was necessary before moving to Newtonsoft.json.  Will remove.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -590,6 +595,7 @@ private object ProcessDictionary(IDictionary dict, int depth)
 
             foreach (DictionaryEntry entry in dict)
             {
+                CheckStopping();
```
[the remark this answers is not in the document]
SteveL-MSFT: Will remove

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -620,6 +626,7 @@ private object ProcessEnumerable(IEnumerable enumerable, int depth)
 
             foreach (object o in enumerable)
             {
+                CheckStopping();
```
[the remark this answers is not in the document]
SteveL-MSFT: Will remove

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
SteveL-MSFT: Makes sense, will update.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -107,7 +107,16 @@ protected override void EndProcessing()
...
+                object preprocessedObject = null;
+                try
+                {
+                    preprocessedObject = ProcessValue(objectToProcess, 0);
+                }
+                catch (StoppingException)
+                {
+                    return;
+                }
+                CheckStopping();
```
PaulHigin: I am a little confused by this.  The StoppingException is caught and ignored in one case but allowed to throw if ProcessValue() completes via CheckStopping().  What is the user experience?  Is the user intended to see the exception or does the cmdlet just stop?

PaulHigin (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
daxian-dbw: I don't think we need to check for stopping again here. You may argue that the user can have a final chance to stop the command before the `JsonConvert.SerializeObject` has the control, though :)

If you really want to check for stop another time here, then `if (Stopping) { return }` should be sufficient.

SteveL-MSFT: Yes, will change.

daxian-dbw (approved):
LGTM

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -683,5 +426,21 @@ private object ProcessCustomObject<T>(object o, int depth)
             }
             return result;
         }
+
+        /// <summary>
+        /// Check if Stopping is set and throw special exception to stop.
+        /// </summary>
+        private void CheckStopping()
```
iSazonov: Seems we should remove the the help method - used one time.

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Json.Tests.ps1
```diff
@@ -16,4 +16,23 @@ Describe 'ConvertTo-Json' -tags "CI" {
...
+        $null = $ps.AddScript({
+            $obj = [PSCustomObject]@{P1 = ''; P2 = ''; P3 = ''; P4 = ''; P5 = ''; P6 = ''}
+            $obj.P1 = $obj.P2 = $obj.P3 = $obj.P4 = $obj.P5 = $obj.P6 = $obj
+            1..100 | Foreach-Object { $obj } | ConvertTo-Json -Depth 10
+            # the conversion is expected to take some time, this throw is in case it doesn't
+            throw "Should not have thrown exception"
+        })
+        $null = $ps.BeginInvoke()
+        # wait to ensure invocation has reached ConvertTo-Json
+        Start-Sleep -Milliseconds 100
```
iSazonov: I'd prefer deterministic way to to ensure invocation has reached ConvertTo-Json

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
SteveL-MSFT: Yes, with the removal of the unused code, this helper is no longer needed.  Will remove.

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Json.Tests.ps1
SteveL-MSFT: Let me work on that

daxian-dbw (approved):
LGTM

markekraus (approved):
LGTM

iSazonov (approved):

PaulHigin (approved):
LGTM

PR #6394, opened 2018-03-14
Fix 'PropertyOnlyAdapter' to allow calling base methods

### LINKED ISSUE #6386: Calling base class method in a System.Xml.XmlNode-derived custom class fails; invokes own method instead, resulting in infinite recursion
The documented syntax for invoking a base-class method from a PS custom class is  
`([<base-class>] $this).<method>()`, but that doesn't seem to work, as discovered in [this SO question](https://stackoverflow.com/q/49259408/45375).

Instead, the derived class' method itself is invoked, resulting in infinite recursion.

**Update**: The problem appears to be specific to `[System.Xml.XmlNode]`-derived classes; see [@daxian-dbw's comment below](https://github.com/PowerShell/PowerShell/issues/6386#issuecomment-372869624).

A reflection-based workaround can be found in [this answer](https://stackoverflow.com/a/49265123/45375).

Steps to reproduce
------------------

```powershell
class child : System.Xml.XmlDocument {
  [void] LoadXml([string]$content) {
    Write-Host 'Inside overridden LoadXml.'
    # Try to call the base type's .LoadXml() method.
    ([System.Xml.XmlDocument] $this).LoadXml($content)
  }
}

[child]::new().LoadXml('<foo>bar</foo>')
```

Expected behavior
-----------------

```none
Inside overridden LoadXml.
```

The message should print once, and the XML string should be loaded into the instance via the call to the base-class method.

Actual behavior
---------------

```none
Inside overridden LoadXml.
Inside overridden LoadXml.
...
```

The method keeps calling itself and eventually runs into a call-depth overflow.

Environment data
----------------

```powershell
PowerShell Core v6.0.1 on macOS 10.13.3
PowerShell Core v6.0.1 on Ubuntu 16.04.3 LTS
PowerShell Core v6.0.1 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
Windows PowerShell v5.1.15063.674 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
```

daxian-dbw:
## PR Summary

Fix #6386
For a PropertyOnlyAdapter, the property may come from various sources, but methods, including parameterized properties, still come from DotNetAdapter. So, the binder should be able to optimize on method calls for objects that map to a custom PropertyOnlyAdapter.

Therefore, the original `SiteBinderCanOptimize` property is changed to a method `CanSiteBinderOptimize(MemberTypes typeToOperateOn)`. For `PropertyOnlyAdapter`, the method returns `false` for property operations and `true` for method operations.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/CoreAdapter.cs
+++ b/src/System.Management.Automation/engine/CoreAdapter.cs
@@ -38,19 +38,6 @@ namespace System.Management.Automation
     internal abstract class Adapter
     {
         /// <summary>
-        /// The member type that a callsite binder operates on.
-        /// </summary>
-        /// <remarks>
-        /// An adapter can decide whether the binder can optimize the operations
-        /// on an object member based on the type of the member.
-        /// </remarks>
-        internal enum SiteBinderOperatesOn
-        {
-            Property,
-            Method,
-        }
-
-        /// <summary>
         /// tracer for this and derivate classes
         /// </summary>
         [TraceSource("ETS", "Extended Type System")]
@@ -59,7 +46,7 @@ namespace System.Management.Automation
 
         #region member
 
-        internal virtual bool CanSiteBinderOptimize(SiteBinderOperatesOn opType) { return false; }
+        internal virtual bool CanSiteBinderOptimize(MemberTypes typeToOperateOn) { return false; }
 
         protected static IEnumerable<string> GetDotNetTypeNameHierarchy(Type type)
         {
@@ -3532,7 +3519,7 @@ namespace System.Management.Automation
 
         #region member
 
-        internal override bool CanSiteBinderOptimize(SiteBinderOperatesOn opType) { return true; }
+        internal override bool CanSiteBinderOptimize(MemberTypes typeToOperateOn) { return true; }
 
         private static ConcurrentDictionary<Type, ConsolidatedString> s_typeToTypeNameDictionary =
             new ConcurrentDictionary<Type, ConsolidatedString>();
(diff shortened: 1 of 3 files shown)


src/System.Management.Automation/engine/CoreAdapter.cs
```diff
@@ -37,6 +37,19 @@ namespace System.Management.Automation
...
     internal abstract class Adapter
     {
+        /// <summary>
+        /// The member type that a callsite binder operates on.
+        /// </summary>
+        /// <remarks>
+        /// An adapter can decide whether the binder can optimize the operations
+        /// on an object member based on the type of the member.
+        /// </remarks>
+        internal enum SiteBinderOperatesOn
```
lzybkr: You might consider using the existing flags enum  `System.Management.Automation.PSMemberTypes` instead.

lzybkr (approved):

src/System.Management.Automation/engine/CoreAdapter.cs
daxian-dbw: I think `System.Reflection.MemberTypes` is a better option :) I updated the code to use that enum type.

iSazonov:
With last commit it seems the PR summary should be corrected.

test/powershell/engine/ETS/Adapter.Tests.ps1
```diff
@@ -315,3 +315,49 @@ Describe "DataRow and DataRowView Adapter tests" -tags "CI" {
...
+'@
+        try {
+            $null = [BaseMethodCallTest.OnXmlDocument.Foo]
+        } catch {
+            Add-Type -TypeDefinition $code
+        }
+
+        $foo = [BaseMethodCallTest.OnXmlDocument.Foo]::new()
+        $foo.LoadXml('<test>bar</test>')
+        $foo.MyName | Should -Be '<test>bar</test>'
```
iSazonov: Usually we use `-BeExactly` for strings.

test/powershell/engine/ETS/Adapter.Tests.ps1
```diff
@@ -315,3 +315,49 @@ Describe "DataRow and DataRowView Adapter tests" -tags "CI" {
...
+        $foo.ChildNodes.Count | Should -Be 0
+
+        ([System.Xml.XmlDocument]$foo).LoadXml('<test>bar</test>')
+        $foo.test | Should -Be 'bar'
```
iSazonov: Usually we use `-BeExactly` for strings.

test/powershell/engine/ETS/Adapter.Tests.ps1
```diff
@@ -315,3 +315,49 @@ Describe "DataRow and DataRowView Adapter tests" -tags "CI" {
...
+                $this.MyName = $content
+                # Try to call the base type's .LoadXml() method.
+                ([System.Xml.XmlDocument] $this).LoadXml($content)
+            }
+        }
+
+        $child = [XmlDocChild]::new()
+        $child.LoadXml('<test>bar</test>')
+        $child.MyName | Should -Be '<test>bar</test>'
+        $child.test | Should -Be 'bar'
```
iSazonov: Usually we use `-BeExactly` for strings.

Issue #3773, opened 2017-05-12
Common parameters -OutVariable, -ErrorVariable, -WarningVariable, and -InformationVariable always return a [System.Collections.ArrayList] instance, even with a single object


mklement0:
Note: The problem was first discovered with `-OutVariable`  - see  #3154.  
The problem applies to all `-*Variable` common parameters, however, with the exception of `-PipelineVariable` (which inherently captures a _single_ item at time).

To recap the problem:

* Even if only a _single_ value is captured by these parameters, an _array list_ rather than a scalar is returned (the single-element collection isn't unwrapped - a deviation from PowerShell's normal collection handling).

* Also - another deviation from usual PowerShell behavior - it is type `[System.Collections.ArrayList]` that is always used, in contrast with the `[System.Object[]]` instances you get with _regular_ assignments.

Steps to reproduce
------------------

```powershell
# Adv. function that outputs a single item of each type.
function foo {
  [CmdletBinding()]
  param()
  
  Write-Warning 'warn'
  Write-Information 'info'
  Write-Error 'err'
  'out'
}

foo -OutVariable ov -ErrorVariable ev -WarningVariable wv -InformationVariable iv  *>$null

$ov, $ev, $wv, $iv | % GetType | % Name
```

Expected behavior
-----------------

```
String
ErrorRecord
WarningRecord
InformationRecord
```

Actual behavior
---------------

```
ArrayList
ArrayList
ArrayList
ArrayList
```

Environment data
----------------

```powershell
PowerShell Core v6.0.0-beta (v6.0.0-beta.1)
```

Jaykul:
While I agree that it's weird that this doesn't behave like pipeline output (and I insist that the *only* thing that PowerShell does this for is pipeline output) ...

It has been like this since the beginning.

Changing it at this point doesn't help much, and risks breaking all existing code that relies on these common parameters ...

mklement0:
@Jaykul: If the consensus is that this change is too risky, so be it, and _documenting_ the issue is the best we can do.
Personally, I think it's a wart that warrants removal - it can trip up new users (as it did me).

> and I insist that the only thing that PowerShell does this for is pipeline output

That is not much of a restriction, however, given the pervasive use of pipelines, even in places where it is not obvious:`$(...)`, `@(...)`, `&`, `.`,  script-block arguments.

To give a few examples, all of which output `$True`:

```powershell
$(, 1) -is [int]
(. { , 1 }) -is [int]
[...]

PR #6409, opened 2018-03-16
Stop stream variables (e.g. OutVariable) becoming ArrayLists

### LINKED ISSUE #3154: -OutVariable doesn't unwrap single-object output and creates [System.Collections.ArrayList] values rather than [System.Object[]]
<sup>Note: Also affects other common parameters: see #3773</sup>

Steps to reproduce
------------------

```powershell
$null = Get-Date -OutVariable ov; $ov.GetType().FullName
```

Expected behavior
-----------------

Type name `System.DateTime`

Actual behavior
---------------

Type name `System.Collections.ArrayList`

That is, the single-object output stored in the variable targeted with `-OutVariable` received an _array list_ rather than a scalar (the single-element collection wasn't unwrapped).

Also, it is type `[System.Collections.ArrayList]` that is always used, in contrast with the `[System.Object[]]` instances you get with _regular_ assignments.

See also: http://stackoverflow.com/a/40666568/45375

Environment data
----------------

```powershell
PowerShell v6.0.0-alpha (v6.0.0-alpha.15) on Darwin Kernel Version 16.4.0: Thu Dec 22 22:53:21 PST 2016; root:xnu-3789.41.3~3/RELEASE_X86_64
PowerShell v5.1.14393.693 on Microsoft Windows 10 Pro (64-bit; v10.0.14393)
```

rjmholt:
## PR Summary
Fixes #3154 and #3773.

Currently, when a value is passed out of a command using `-OutVariable`, we use an `ArrayList` to collect the value to give back to the user, but we leak that detail so that the variable comes back as an `ArrayList` and not `object[]` or as itself.

This PR fixes this, so that implementation differences between
```powershell
$null = Write-Output -OutVariable x $value; $x
```
and
```powershell
Write-Output $value
```
are transparent.

To summarize:

| Input | Non-OutVar Result Type | Old Result Type | New Result Type |
| :-: | :-: | :-: | :-: |
| `'Hello'` | System.String | System.Collections.ArrayList | System.String |
| `@(1, 2)` | object[] | System.Collections.ArrayList | object[] |
| `[System.Collections.ArrayList]::new(@(1,2))` | object[] | System.Collections.ArrayList | object[] |
| `@(,[System.Collections.ArrayList]::new(@(1,2))` | System.Collections.ArrayList | System.Collections.ArrayList | System.Collections.ArrayList |
| `$null` | ⊥ | System.Collections.ArrayList | ⊥ |

This is technically a **breaking change**. The PowerShell commitee reviewed it in #3154. A suggested classification was as a class-3 unlikely grey area change, where the new behaviour is more likely to be expected.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/MshCommandRuntime.cs
+++ b/src/System.Management.Automation/engine/MshCommandRuntime.cs
@@ -930,8 +930,6 @@ namespace System.Management.Automation
                 {
                     this.OutputPipe.AddVariableList(VariableStreamKind.Output, _outVarList);
                 }
-
-                _state.PSVariable.Set(this.OutVariable, _outVarList);
             }
             else
             {
@@ -2498,34 +2496,21 @@ namespace System.Management.Automation
 
             if (variableName.StartsWith("+", StringComparison.Ordinal))
             {
+                varList = new ArrayList();
+
                 variableName = variableName.Substring(1);
                 object oldValue = PSObject.Base(_state.PSVariable.GetValue(variableName));
-                varList = oldValue as IList;
-                if (null == varList)
-                {
-                    varList = new ArrayList();
 
-                    if (null != oldValue && AutomationNull.Value != oldValue)
+                if (oldValue is object[] oldArray)
+                {
+                    foreach (object streamObject in oldArray)
                     {
-                        IEnumerable enumerable = LanguagePrimitives.GetEnumerable(oldValue);
-                        if (null != enumerable)
-                        {
-                            foreach (object o in enumerable)
-                            {
-                                varList.Add(o);
-                            }
-                        }
-                        else
-                        {
-                            varList.Add(oldValue);
-                        }
+                        varList.Add(streamObject);
                     }
                 }
-                else if (varList.IsFixedSize)
+                else if (oldValue != null && oldValue != AutomationNull.Value)
                 {
-                    ArrayList varListNew = new ArrayList();
-                    varListNew.AddRange(varList);
-                    varList = varListNew;
+                    varList.Add(oldValue);
                 }
             }
             else
@@ -2537,7 +2522,6 @@ namespace System.Management.Automation
             {
                 this.OutputPipe.AddVariableList(streamKind, varList);
             }
-            _state.PSVariable.Set(variableName, varList);
         } // SetupVariable
 
         /// <summary>
@@ -3677,21 +3661,25 @@ namespace System.Management.Automation
 
             if (_outVarList != null)
             {
+                SetNonArrayListVariable(OutVariable, _outVarList);
                 this.OutputPipe.RemoveVariableList(VariableStreamKind.Output, _outVarList);
             }
 
             if (_errorVarList != null)
             {
+                SetNonArrayListVariable(ErrorVariable, _errorVarList);
                 this.OutputPipe.RemoveVariableList(VariableStreamKind.Error, _errorVarList);
             }
 
             if (_warningVarList != null)
             {
+                SetNonArrayListVariable(WarningVariable, _warningVarList);
                 this.OutputPipe.RemoveVariableList(VariableStreamKind.Warning, _warningVarList);
             }
 
             if (_informationVarList != null)
             {
+                SetNonArrayListVariable(InformationVariable, _informationVarList);
                 this.OutputPipe.RemoveVariableList(VariableStreamKind.Information, _informationVarList);
             }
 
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/engine/MshCommandRuntime.cs
```diff
@@ -2498,34 +2496,21 @@ internal void SetupVariable(VariableStreamKind streamKind, string variableName,
...
                 object oldValue = PSObject.Base(_state.PSVariable.GetValue(variableName));
-                varList = oldValue as IList;
-                if (null == varList)
-                {
-                    varList = new ArrayList();
 
-                    if (null != oldValue && AutomationNull.Value != oldValue)
+                if (oldValue is object[] oldArray)
+                {
+                    foreach (object v in oldArray)
```
iSazonov: Could you please use more informative variable name then `v`?

src/System.Management.Automation/engine/MshCommandRuntime.cs
```diff
@@ -3701,5 +3689,45 @@ internal void RemoveVariableListsInPipe()
...
+        /// <param name="variableName">the name of the variable to set</param>
+        /// <param name="varList">the variable value stored in the ArrayList</param>
+        private void SetNonArrayListVariable(string variableName, IList varList)
+        {
+            if (variableName == null)
+            {
+                return;
+            }
+
+            // No values -> return null
```
iSazonov: Seems `return` is not correct term in comments here and below.

src/System.Management.Automation/engine/MshCommandRuntime.cs
```diff
@@ -3701,5 +3689,45 @@ internal void RemoveVariableListsInPipe()
...
+            if (varList.Count == 0)
+            {
+                _state.PSVariable.Set(variableName, null);
```
iSazonov: Do we really need this assign? It seems is default.

test/powershell/Language/Scripting/OutErrorVariable.Tests.ps1
```diff
@@ -75,6 +75,48 @@ Describe "Tests OutVariable only" -Tags "CI" {
...
+                                    },
+                                @{ Name = 'Pass $null through OutVariable';
+                                    Value = $null;
+                                    Expected = $null
+                                    },
+                                @{ Name = 'Pass AutomationNull.Value through OutVariable';
+                                    Value = [System.Management.Automation.Internal.AutomationNull]::Value;
+                                    Expected = $null
+                                    }
+                                )
```
iSazonov: Indentation. I believe it should be in 4th position.

test/powershell/Language/Scripting/OutErrorVariable.Tests.ps1
```diff
@@ -75,6 +75,48 @@ Describe "Tests OutVariable only" -Tags "CI" {
         $script:a | Should Be 'foo'
         $b | Should Be @("bar", "foo")
     }
+
+    $outVarTypeTestData = @(
```
iSazonov: It is used once - please move directly in `-TestCases`.

test/powershell/Language/Scripting/OutErrorVariable.Tests.ps1
```diff
@@ -75,6 +75,48 @@ Describe "Tests OutVariable only" -Tags "CI" {
...
+                                @{ Name = 'Pass single object through OutVariable';
```
iSazonov: Indentation. I believe it should be 8th position.

test/powershell/Language/Scripting/OutErrorVariable.Tests.ps1
rjmholt: Happy to do this. Is there an example of the code style you can link where `$TestCases` is inlined in the call?

src/System.Management.Automation/engine/MshCommandRuntime.cs
rjmholt: If you take a case like
```powershell
$x = Write-Output -OutVariable ov @()
```
Then `$x` is `$null` but without the assignment, `$ov` is an empty `object[]`. If our intent is for them to be the same, then I think we need the assignment to `null` here.

src/System.Management.Automation/engine/MshCommandRuntime.cs
rjmholt: Good point!

rjmholt:
Quick question - the `[feature]` checkbox - is that supposed to just be in the commit message and not in the title of the PR?

Jaykul:
This would be a pretty major breaking change...

markekraus:
Umm... I disagree that this is a `Bucket 3` breaking change and request it be classified as `Bucket 1`. Especially if you are going to return a string as a `String` instead of an ArrayList.

I have been preaching for years as well as using the `-OutVariable`  specifically because it always returns an ArrayList. Since it is always a predictable type you can always use the same methods and actions on it.

If Strings are suddenly returned... this becomes problematic:

```powershell
$null = Get-MaybeOneOrMoreStrings -OutVariable Results
$FirstString = $Results[0]
```

This is a simplified example of something I make great use of in many places.

markekraus:
To clarify, I would be ok with these returning any collection type (changed in a major version bump). In order of preference `List<Object>`, `ArrayList`, `Object[]`. But I am wholeheartedly against this returning single values and not a collection.

markekraus:
This PR breaks the following usage:

```powershell
$Results = [System.Collections.ArrayList]::new()
$Results.Add('before')
Write-Output 'a' -OutVariable '+Results'
$Results.Add('After')

$Results.Count
$Results[0]
$Results[1]
$Results[2]
```

```none
Exception calling "Add" with "1" argument(s): "Collection was of a fixed size."
At line:1 char:1
+ $Results.Add('After')
+ ~~~~~~~~~~~~~~~~~~~~~
+ FullyQualifiedErrorId : NotSupportedException
```

markekraus (changes requested):
Please have PowerShell Committee review this again. I believe this is a Bucket 1 breaking change and should go through RFC.

rjmholt:
Apologies -- after reading through #3154, the responses there seemed to encourage the change.

Since I've already created a PR, I've gone ahead an created an [RFC](https://github.com/PowerShell/PowerShell-RFC/pull/120), so feel free to comment there.

I may also submit a PR to change the docs to emphasise and strengthen the language on RFCs for breaking changes. After going through the PR checklist, I didn't catch the reference to possibly submitting an RFC.

I will mothball this PR for now.

markekraus:
@rjmholt No worries! if I had seen those other issues I would have raised a stink about it needing an RFC before the PR! 😆

iSazonov:
>Quick question - the [feature] checkbox - is that supposed to just be in the commit message and not in the title of the PR?

In last commit header.

Issue #6413, opened 2018-03-16
Tests (especially Get-FileHash) depend on machine-specific details


rjmholt:
There are a number of tests in the PowerShell test suite that depend on a file asset in a way that depends on the specific bytes of the file. These tests fail on platforms that somehow differ from the expected encodings. For example, my Git configuration converts LF line endings to CRLF locally on Windows when checking out (and converts back on check-in), which is the default Git configuration on Windows, but because of this and Windows introducing a BOM, my file hashes come out differently.

I've made a small list of the tests I see failing on my machine because of this:
- [x] Get-FileHash tests depend on file encoding -- CRLF/BOM breaks it
- [ ] A Get-Process test depends on having a process with the name "notepad"
- [ ] Write-Error ErrorRecord tests seem to be affected (get 11 and not the expected 4)
- [ ] Get-PSSessionConfiguration assumes there is only a single PSSessionConfiguration on the machine
- [ ] NativeStreams testing using the `$longtext` like in the Write-Error tests have a similar outcome

There might be a couple more tests like this too (I will update if I find any).

For the string/encoding-dependent tests, my suggestion is to use byte streams directly instead.

Steps to reproduce
------------------

```powershell
cd $powershellBuildDir
Import-Module ./build.psm1
cd ./test/powershell/Modules/Microsoft.PowerShell.Utility
Start-PSPester Get-FileHash.Tests.ps1
```

Expected behavior
-----------------

Get-FileHash tests pass.

Actual behavior
---------------

Hashes all come out differently

![Get-FileHash tests all failing on Win10](https://imgur.com/MqcloqO.jpg)

Environment data
----------------

```powershell
> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.0.1
PSEdition                      Core
GitCommitId                    v6.0.1-261-g71cb49a99d00ae4257e63f6564fd122325b64a48
OS                             Microsoft Windows 10.0.16299 
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0


```

rjmholt:
This may be the cause of #5960 and #6369.

Issue #6369, opened 2018-03-12
Get-FileHash.Tests are failing on server skus.


dantraMSFT:
Steps to reproduce
------------------

```powershell
start-pspester -path test\powershell\Modules\Microsoft.PowerShell.Utility\Get-FileHash.Tests.ps1
```

Expected behavior
-----------------

All tests pass

Actual behavior
---------------

All tests in the 'Default result tests' and 'Algorithm tests' Context blocks fail  comparing the expected to the actual hash values.

Environment data
----------------

```powershell
> $PSVersionTable
PSVersion                      6.0.1
PSEdition                      Core
GitCommitId                    v6.0.1-276-g8ced4aa8635bf44aeaf40ba9efaf3ea37705e4c4
OS                             Microsoft Windows 10.0.17619
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

rjmholt:
@dantraMSFT Does #6430 fix this?

Issue #5960, opened 2018-01-19
Get-FileHash tests consistently failing on Server 2016


dantraMSFT:
Steps to reproduce
------------------

Build current sources and run Start-PSPester, no parameters

Expected behavior
-----------------

All Tests pass.

Actual behavior
---------------
Each test Get-FileHash.Tests.ps1 that is comparing against an expected hash is failing with a mismatched hash value.

Note that this mismatch is not a case-sensitivity failure; while the string lengths match for each algorithm, the the hash values are completely different.

Environment data
----------------
Running on server datacenter 2016

$PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.0.0
PSEdition                      Core
GitCommitId                    v6.0.0-128-gf587fd06d6046859ca1a9fe3470d5f9615dc5e01
OS                             Microsoft Windows 10.0.17080
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

rjmholt:
@dantraMSFT Does #6430 fix this?

PR #3395, opened 2017-03-22
WIP - Implement Get-Hash cmdlet

iSazonov:
It is first experimental implementation Get-Hash cmdlet for https://github.com/PowerShell/PowerShell-RFC/blob/master/2-Draft-Accepted/RFC0018-Get-StringHash.md

1. According to the RFC the type of Encoding parameter is `FileSystemCmdletProviderEncoding` but I had to use `EncodingConversion class` because the API contains useful methods and can be easily _enhanced_ in future (to support full set of codepages). Now multiple Utility cmdlets uses it. In contrast, a code using `FileSystemCmdletProviderEncoding` type is distributed in FileProvider and it's hard to reuse it although it is designated as the preferred (?) in https://github.com/PowerShell/PowerShell-RFC/blob/master/1-Draft/RFC0020-DefaultFileEncoding.md We also talked about that we should rationalize the use of encoding so `EncodingConversion class` looks like the best option.
2. Added `ValueFromPipelineByPropertyName` to `Path` parameter that is skipped in the RFC 0018.
3. ~~After the implementation we can really see that `Path` and `InputString` parameters have the same type and cannot be both `ValueFromPipeline`. Perhaps here we understand that it is better for UX to have both options and we should split on `Get-FileHash` and `Get-StringHash` still. Until concluding on this I haven't added pipeline tests.~~ We follow `Select-String` pattern as suggested by @rkeithhill.
4. ~~I wanted to implement `ValidateEncoding` attribute but was blocked so how to implement Tab Completion. I need help.~~ We should follow #4119.

PR #6430, opened 2018-03-19
Change `Get-FileHash` tests to use raw bytes

rjmholt:
## PR Summary

Tests of `Get-FileHash` currently rely on a static asset, whose encoding depends on platform configurations. This causes hash output to be platform-dependent and makes tests fail.

This PR changes the tests to write raw bytes to the test file first, so that the platform does not influence the outcome of the test.

Partially addresses #6413 and possibly #5960 and #6369.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/Get-FileHash.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/Get-FileHash.Tests.ps1
@@ -4,15 +4,22 @@ Describe "Get-FileHash" -Tags "CI" {
 
     BeforeAll {
         $testDocument = Join-Path -Path $TestDrive -ChildPath "hashable.txt"
-        # Encode bytes of "<BOM>Get-Module`n" directly to avoid platform issues
-        [System.IO.File]::WriteAllBytes($testDocument, @(0xef, 0xbb, 0xbf, 0x47, 0x65, 0x74, 0x2d, 0x4d, 0x6f, 0x64, 0x75, 0x6c, 0x65, 0x0a))
+        $utf8NoBOM = [System.Text.UTF8Encoding]::new($false)
+        [System.IO.File]::WriteAllText($testDocument, "Get-Module`n", $utf8NoBOM)
+    }
+
+    Context "Validate platform encoding" {
+        # If this test fails, then the problem lies outside the Get-FileHash implementation
+        It "Should have the bytes in the file as expected" {
+            [System.IO.File]::ReadAllBytes($testDocument) | Should -BeExactly @(0x47, 0x65, 0x74, 0x2d, 0x4d, 0x6f, 0x64, 0x75, 0x6c, 0x65, 0x0a)
+        }
     }
 
     Context "Default result tests" {
         It "Should default to correct algorithm, hash and path" {
             $result = Get-FileHash $testDocument
             $result.Algorithm | Should Be "SHA256"
-            $result.Hash | Should Be "8129a08e5d748ffb9361375677785f96545a1a37619a27608efd76a870787a7a"
+            $result.Hash | Should Be "41620f6c9f3531722efe90aed9abbc1d1b31788aa9141982030d3dde199f770c"
             $result.Path | Should Be $testDocument
         }
     }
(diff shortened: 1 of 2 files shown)


test/powershell/Modules/Microsoft.PowerShell.Utility/Get-FileHash.Tests.ps1
```diff
@@ -3,8 +3,10 @@
 Describe "Get-FileHash" -Tags "CI" {
 
     BeforeAll {
-        $testDocument = Join-Path -Path $PSScriptRoot -ChildPath assets testablescript.ps1
+        $testDocument = Join-Path -Path $TestDrive -ChildPath "hashable.txt"
         Write-Host $testDocument
+        # Encode bytes directly to avoid platform issues
+        [System.IO.File]::WriteAllBytes($testDocument, @(0xef, 0xbb, 0xbf, 0x47, 0x65, 0x74, 0x2d, 0x4d, 0x6f, 0x64, 0x75, 0x6c, 0x65, 0x0a))
```
SteveL-MSFT: You should have a comment on what those bytes represent

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-FileHash.Tests.ps1
```diff
@@ -3,8 +3,10 @@
 Describe "Get-FileHash" -Tags "CI" {
 
     BeforeAll {
-        $testDocument = Join-Path -Path $PSScriptRoot -ChildPath assets testablescript.ps1
+        $testDocument = Join-Path -Path $TestDrive -ChildPath "hashable.txt"
         Write-Host $testDocument
```
SteveL-MSFT: We shouldn't have a `Write-Host` here, seems like a leftover when this was written for debugging purposes.  Just delete this line.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-FileHash.Tests.ps1
```diff
@@ -3,8 +3,10 @@
 Describe "Get-FileHash" -Tags "CI" {
 
     BeforeAll {
-        $testDocument = Join-Path -Path $PSScriptRoot -ChildPath assets testablescript.ps1
```
SteveL-MSFT: I don't see that script being used anywhere else, so you should delete that script if it's no longer needed.

SteveL-MSFT (changes requested):

rjmholt:
Just to document: the actual byte string used is "<BOM>Get-Module\n" encoded in UTF-8. The BOM may be controversial, but that's what matches the hashes that were already there, so is the simplest solution.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-FileHash.Tests.ps1
rjmholt: Deleted in latest commit

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-FileHash.Tests.ps1
```diff
@@ -3,8 +3,9 @@
...
-        Write-Host $testDocument
+        $testDocument = Join-Path -Path $TestDrive -ChildPath "hashable.txt"
+        # Encode bytes of "<BOM>Get-Module`n" directly to avoid platform issues
+        [System.IO.File]::WriteAllBytes($testDocument, @(0xef, 0xbb, 0xbf, 0x47, 0x65, 0x74, 0x2d, 0x4d, 0x6f, 0x64, 0x75, 0x6c, 0x65, 0x0a))
```
SteveL-MSFT: Based on your explanation, we are moving towards BOM-less UTF-8, so perhaps we should remove the BOM, get the bytes at runtime making the code more readable, and updating the expected hash for validation.

SteveL-MSFT (dismissed):

rjmholt:
I've restarted the Travis CI job, since none of the tests failed.

The problem was this:
Travis CI failure

SteveL-MSFT (approved):

adityapatwardhan (approved):

rjmholt:
@dantraMSFT with this merged, are #6369 and #5960 still issues?

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-FileHash.Tests.ps1
```diff
@@ -3,15 +3,23 @@
 Describe "Get-FileHash" -Tags "CI" {
 
     BeforeAll {
-        $testDocument = Join-Path -Path $PSScriptRoot -ChildPath assets testablescript.ps1
-        Write-Host $testDocument
+        $testDocument = Join-Path -Path $TestDrive -ChildPath "hashable.txt"
+        $utf8NoBOM = [System.Text.UTF8Encoding]::new($false)
+        [System.IO.File]::WriteAllText($testDocument, "Get-Module`n", $utf8NoBOM)
```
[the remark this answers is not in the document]
rjmholt: Excellent questions!

The default Git configuration on Windows machines means that LF is checked out as CRLF -- which added a CR byte to `testablescript.ps1` and broke the test. Also, opening testablescript.ps1 in text editors caused the BOM to get deleted in some cases.

I would have been happy with leaving the BOM, or even just writing some arbitrary bytes to a file, but we decided that if .NET Core is trying to get away from using BOMs in UTF-8, we should encourage that everywhere. Also, it makes the tests easier to verify against other hashsum calculators in BOM-less UTF-8 environments, which are arguably more prevalent and more likely to have an independent implementation from ours.

Finally, the WriteAllText we use is [`[System.IO.File]::WriteAllText(String path, String text, Encoding fileEncoding)`](https://msdn.microsoft.com/en-us/library/ms143376(v=vs.110).aspx), rather than [`[Microsoft.VisualBasic.FileIO.FileSystem]::WriteAllText(String file, String text, bool append)`](https://msdn.microsoft.com/en-us/library/ms128035%28v=vs.110%29.aspx). Which I also got caught up by a couple of times.

rjmholt:
Also, I was waiting til the end to squash my commits -- should I have done that earlier?

iSazonov:
Clear about WriteAllText :-)

I wonder - I was sure we excluded these assert files from the Git recoding 😕 But it doesn't matter anymore - I agree that it is better to create them in runtime.
As for the BOM I believe we should keep these tests because the BOM is still widely used and this cmdlet should work with it correctly. I've PR #3395 and I fear that I can ruin something without these tests.

iSazonov:
I think if we copy a file (with BOM too) from Windows 7 to Windows Server 2016 - the hash should be the same on both platforms.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-FileHash.Tests.ps1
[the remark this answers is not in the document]
SteveL-MSFT: The hash should be based on the bytes of the file and the file encoding including BOM is opaque to the cmdlet, so don't think there's value in having a test with BOM.

SteveL-MSFT:
@rjmholt don't worry about squashing, it happens at merge time

rjmholt:
@SteveL-MSFT Awesome

@iSazonov I agree, we should add some more tests that have strings with BOMs and line endings in. One because that helps us verify the `Get-FileHash` cmdlet, but also because having that test fail is often a good indicator that other things could be happening.

iSazonov:
@rjmholt Have you plans to add the tests? Otherwise please open new tracking Issue.

PR #6434, opened 2018-03-20
Throw better parsing error when statements should be put in named block

### LINKED ISSUE #6431: Mistakenly using a Dynamicparam block in single-body functions fails obscurely
This is a minor problem and the behavior is as designed, but perhaps the error message can be improved.

A `dynamicparam {}` block requires an advanced function's other code to  be in `begin {}`, `process{}` and `end {}` blocks.

Instead trying to use `dynamicparam` in the function's main block _alongside top-level code_ fails obscurely:

Steps to reproduce
------------------

```powershell
Function foo {
    [CmdletBinding()] param()
    DynamicParam {
    }

    # Placing code here, outside of begin {}, process {} or end {} fails obscurely.
    'hi'
}

```

Expected behavior
-----------------

Actual behavior
---------------

```none
At /Users/mklement/Desktop/pg/pg.ps1:33 char:14
+ Function foo {
+              ~
Missing closing '}' in statement block or type definition.
```

It is not obvious from this error message what the problem is.

Enclosing `'hi'` in `end { ... }` fixes the problem.

Aside from a more helpful error message, perhaps _allowing_ this code is an option (where the top-level code would run in an implied `end {}` block), 
though perhaps that requires special-cased parsing behavior that is not worth the trouble.

Environment data
----------------

```powershell
PowerShell Core v6.0.2 on macOS 10.13.3
PowerShell Core v6.0.2 on Ubuntu 16.04.3 LTS
PowerShell Core v6.0.2 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
Windows PowerShell v5.1.15063.674 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
```

daxian-dbw:
## PR Summary

Fix #6431 
Throw better parsing error when statements should be put in named block.
Now the error looks like this:

```powershell
PS:24> Function foo {
>>     [CmdletBinding()] param()
>>     DynamicParam {
>>     }
>>
>>     'hi'
At line:7 char:5
+     'hi'
+     ~~~~
unexpected token ''hi'', expected 'begin', 'process', 'end', or 'dynamicparam'.
+ CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
+ FullyQualifiedErrorId : MissingNamedBlocks
```

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/parser/Parser.cs
+++ b/src/System.Management.Automation/engine/parser/Parser.cs
@@ -1631,10 +1631,17 @@ namespace System.Management.Automation.Language
                     default:
                         // Next token is unexpected.
                         // ErrorRecovery: if 'lCurly' is present, pretend we saw a closing curly; otherwise, eat the unexpected token.
-                        if (lCurly != null) { UngetToken(blockNameToken); }
+                        if (lCurly != null)
+                        {
+                            UngetToken(blockNameToken);
+                            scriptBlockExtent = ExtentOf(startExtent, endExtent);
+                        }
+                        else
+                        {
+                            // If "lCurly == null", then it's a ps1/psm1 file, and thus the extent is the whole file.
+                            scriptBlockExtent = _tokenizer.GetScriptExtent();
+                        }
 
-                        // If 'lCurly == null", then it's a ps1/psm1 file, and thus the extent is the whole file.
-                        scriptBlockExtent = lCurly != null ? ExtentOf(startExtent, endExtent) : _tokenizer.GetScriptExtent();
                         // Report error about the unexpected token.
                         ReportError(blockNameToken.Extent, () => ParserStrings.MissingNamedBlocks, blockNameToken.Text);
                         goto return_script_block_ast;
--- a/src/System.Management.Automation/engine/parser/Parser.cs
+++ b/src/System.Management.Automation/engine/parser/Parser.cs
@@ -1512,7 +1512,7 @@ namespace System.Management.Automation.Language
             return elementType;
         }
 
-        private bool CompleteScriptBlockBody(Token lCurly, ref IScriptExtent bodyExtent, out IScriptExtent fullBodyExtent, bool calledFromNamedBlockRule)
+        private bool CompleteScriptBlockBody(Token lCurly, ref IScriptExtent bodyExtent, out IScriptExtent fullBodyExtent)
         {
             // If the caller passed in the open curly, then they expect us to consume the closing curly
             // and include that in the extent.
@@ -1526,10 +1526,7 @@ namespace System.Management.Automation.Language
 
                     UngetToken(rCurly);
                     endScriptBlock = bodyExtent ?? lCurly.Extent;
-                    var errorExpr = calledFromNamedBlockRule
-                        ? (Expression<Func<string>>)(() => ParserStrings.MissingEndCurlyBraceOrNamedBlocks)
-                        : (Expression<Func<string>>)(() => ParserStrings.MissingEndCurlyBrace);
-                    ReportIncompleteInput(lCurly.Extent, rCurly.Extent, errorExpr);
+                    ReportIncompleteInput(lCurly.Extent, rCurly.Extent, () => ParserStrings.MissingEndCurlyBrace);
                 }
                 else
                 {
@@ -1595,7 +1592,7 @@ namespace System.Management.Automation.Language
                     statementListExtent = ExtentOf(statementListExtent, extent);
                 }
 
-                if (CompleteScriptBlockBody(lCurly, ref statementListExtent, out scriptBlockExtent, calledFromNamedBlockRule: false))
+                if (CompleteScriptBlockBody(lCurly, ref statementListExtent, out scriptBlockExtent))
                 {
                     break;
                 }
@@ -1623,7 +1620,8 @@ namespace System.Management.Automation.Language
                                             ? lCurly.Extent
                                             : (paramBlockAst != null) ? paramBlockAst.Extent : null;
             IScriptExtent endExtent = null;
-            IScriptExtent extent;
+            IScriptExtent extent = null;
+            IScriptExtent scriptBlockExtent = null;
 
             while (true)
             {
@@ -1632,6 +1630,20 @@ namespace System.Management.Automation.Language
                 {
                     default:
                         UngetToken(blockNameToken);
+                        extent = ExtentOf(startExtent, endExtent);
+
+                        // If 'lCurly' is null, then handle the next token in 'CompleteScriptBlockBody'.
+                        if (lCurly == null) { goto finished_named_block_list; }
+                        // Otherwise, we handle the unexpected next token here.
+                        ReportError(blockNameToken.Extent, () => ParserStrings.MissingNamedBlocks, blockNameToken.Text);
+                        scriptBlockExtent = extent;
+                        goto return_script_block_ast;
+
+                    case TokenKind.RCurly:
+                    case TokenKind.EndOfInput:
+                        // If the next token is RCurly or <eof>, handle it in 'CompleteScriptBlockBody'.
+                        UngetToken(blockNameToken);
+                        extent = ExtentOf(startExtent, endExtent);
                         goto finished_named_block_list;
 
                     case TokenKind.Dynamicparam:
@@ -1690,11 +1702,9 @@ namespace System.Management.Automation.Language
                 SkipNewlinesAndSemicolons();
             }
         finished_named_block_list:
+            CompleteScriptBlockBody(lCurly, ref extent, out scriptBlockExtent);
 
-            IScriptExtent scriptBlockExtent;
-            extent = ExtentOf(startExtent, endExtent);
-            CompleteScriptBlockBody(lCurly, ref extent, out scriptBlockExtent, calledFromNamedBlockRule: true);
-
+        return_script_block_ast:
             return new ScriptBlockAst(scriptBlockExtent, usingStatements, paramBlockAst, beginBlock, processBlock, endBlock,
                 dynamicParamBlock);
         }
--- a/src/System.Management.Automation/engine/parser/Parser.cs
+++ b/src/System.Management.Automation/engine/parser/Parser.cs
@@ -1632,11 +1632,10 @@ namespace System.Management.Automation.Language
                         UngetToken(blockNameToken);
                         extent = ExtentOf(startExtent, endExtent);
 
-                        // If 'lCurly' is null, then handle the next token in 'CompleteScriptBlockBody'.
-                        if (lCurly == null) { goto finished_named_block_list; }
-                        // Otherwise, we handle the unexpected next token here.
+                        // If 'lCurly == null", then it's a ps1/psm1 file, then the extent is the whole file.
+                        scriptBlockExtent = lCurly != null ? extent : _tokenizer.GetScriptExtent();
+                        // Handle the unexpected next token.
                         ReportError(blockNameToken.Extent, () => ParserStrings.MissingNamedBlocks, blockNameToken.Text);
-                        scriptBlockExtent = extent;
                         goto return_script_block_ast;
 
                     case TokenKind.RCurly:
--- a/src/System.Management.Automation/engine/parser/Parser.cs
+++ b/src/System.Management.Automation/engine/parser/Parser.cs
@@ -1629,11 +1629,12 @@ namespace System.Management.Automation.Language
                 switch (blockNameToken.Kind)
                 {
                     default:
-                        UngetToken(blockNameToken);
-                        extent = ExtentOf(startExtent, endExtent);
+                        // Next token is unexpected.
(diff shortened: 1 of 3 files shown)


src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -1526,7 +1526,12 @@ private bool CompleteScriptBlockBody(Token lCurly, ref IScriptExtent bodyExtent,
 
                     UngetToken(rCurly);
                     endScriptBlock = bodyExtent ?? lCurly.Extent;
-                    ReportIncompleteInput(lCurly.Extent, rCurly.Extent, () => ParserStrings.MissingEndCurlyBrace);
+                    Expression<Func<string>> errorExpr = () => ParserStrings.MissingEndCurlyBrace;
+                    if (calledFromNamedBlockRule)
```
iSazonov: Maybe use `calledFromNamedBlockRule ? ParserStrings.MissingEndCurlyBraceOrNamedBlock: ParserStrings.MissingEndCurlyBrace`

src/System.Management.Automation/resources/ParserStrings.resx
```diff
@@ -493,6 +493,9 @@ The correct form is: foreach ($a in $b) {...}</value>
   <data name="DuplicateScriptCommandClause" xml:space="preserve">
     <value>Script command clause '{0}' has already been defined.</value>
   </data>
+  <data name="MissingEndCurlyBraceOrNamedBlock" xml:space="preserve">
+    <value>Missing closing '}' in script block. If you meant to continue with the script block, please note that new statements should be put in named blocks after a named block has been declared.</value>
```
iSazonov: Sorry, it's hard to understand 😕

test/powershell/Language/Parser/Parser.Tests.ps1
```diff
@@ -915,4 +915,14 @@ foo``u{2195}abc
...
         { ExecuteCommand "`$herestr=@`"`n'`"'`n`"@" } | Should Not Throw
     }
+
+    It "Throw better error message when statement should be put in named blocks" {
+        try {
+            ExecuteCommand "Function foo { [CmdletBinding()] param() DynamicParam {} Hi"
+            throw "Execution OK"
+        }
+        catch {
+            $_.Exception.InnerException.ErrorRecord.FullyQualifiedErrorId | Should -Be "MissingEndCurlyBraceOrNamedBlock"
```
iSazonov: Why we don't use `ShouldBeErrorId`?

daxian-dbw: Because the the `FullyQualifiedErrorId` is from a nested exception, not `$_.FullyQualifiedErrorId`.

src/System.Management.Automation/engine/parser/Parser.cs
daxian-dbw: The `Expression<Func<string>>` passed in needs to have a body of `MemberExpression` for the parser to get the correct error id. See [`SaveError`](https://github.com/daxian-dbw/PowerShell/blob/24599fa413485ad65738f606bd06c97b727bc3f3/src/System.Management.Automation/engine/parser/Parser.cs#L6837).

src/System.Management.Automation/resources/ParserStrings.resx
daxian-dbw: How about `Missing closing '}' in script block. If you meant to add more statements to the script block, put the statements in one or more named blocks.`

test/powershell/Language/Parser/Parser.Tests.ps1
iSazonov: Thanks for clarify. Please replace "Execution OK" with "No Execution!!!".

iSazonov (approved):

test/powershell/Language/Parser/Parser.Tests.ps1
```diff
@@ -915,4 +915,14 @@ foo``u{2195}abc
         # Issue #2780
         { ExecuteCommand "`$herestr=@`"`n'`"'`n`"@" } | Should Not Throw
     }
+
+    It "Throw better error message when statement should be put in named blocks" {
+        try {
```
kalgiz: Instead of try and catch we can use: 
`$err = { <command> } | Should -Throw -PassThru
$err.Exception.InnerException.ErrorRecord.FullyQualifiedErrorId | Should -BeExactly "MissingEndCurlyBraceOrNamedBlocks" `

daxian-dbw: Will update to use the Pester V4 syntax.

src/System.Management.Automation/resources/ParserStrings.resx
```diff
@@ -493,6 +493,9 @@ The correct form is: foreach ($a in $b) {...}</value>
   <data name="DuplicateScriptCommandClause" xml:space="preserve">
     <value>Script command clause '{0}' has already been defined.</value>
   </data>
+  <data name="MissingEndCurlyBraceOrNamedBlocks" xml:space="preserve">
+    <value>Missing closing '}' in script block. If you meant to add more statements to the script block, put the statements in one or more named blocks.</value>
```
SteveL-MSFT: I don't think `named blocks` is a common term outside of the code, perhaps change the last part to: put the statements in begin, process, or end blocks.

test/powershell/Language/Parser/Parser.Tests.ps1
```diff
@@ -915,4 +915,9 @@ foo``u{2195}abc
         # Issue #2780
         { ExecuteCommand "`$herestr=@`"`n'`"'`n`"@" } | Should Not Throw
     }
+
+    It "Throw better error message when statement should be put in named blocks" {
+        $err = { ExecuteCommand "Function foo { [CmdletBinding()] param() DynamicParam {} Hi" } | Should -Throw -PassThru
```
SteveL-MSFT: Can you also check the parent errorid: Should -Throw -ErrorId "errorid expected" -PassThru

SteveL-MSFT (changes requested):

src/System.Management.Automation/resources/ParserStrings.resx
daxian-dbw: Fixed.

test/powershell/Language/Parser/Parser.Tests.ps1
daxian-dbw: Sure thing. Fixed.

SteveL-MSFT (approved):

src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -1690,7 +1693,7 @@ private ScriptBlockAst NamedBlockListRule(Token lCurly, List<UsingStatementAst>
 
             IScriptExtent scriptBlockExtent;
             extent = ExtentOf(startExtent, endExtent);
-            CompleteScriptBlockBody(lCurly, ref extent, out scriptBlockExtent);
+            CompleteScriptBlockBody(lCurly, ref extent, out scriptBlockExtent, calledFromNamedBlockRule: true);
```
lzybkr: Instead of reporting here, I think the error should be reported in this function before the `goto finished_named_block_list`.

And the proposed error text might be confusing in some cases, it might be better to just report `unexpected token, expected 'begin', 'process', 'end', or 'dynamicparam'.

src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -1526,7 +1526,10 @@ private bool CompleteScriptBlockBody(Token lCurly, ref IScriptExtent bodyExtent,
 
                     UngetToken(rCurly);
                     endScriptBlock = bodyExtent ?? lCurly.Extent;
-                    ReportIncompleteInput(lCurly.Extent, rCurly.Extent, () => ParserStrings.MissingEndCurlyBrace);
+                    var errorExpr = calledFromNamedBlockRule
```
lzybkr: Not needed in the PR, but I'd suggest removing my ugly hack of using LINQ expression trees and replace with `nameof(ParserStrings.MissingEndCurlyBrace)`.

When that code was written, `nameof` was not in C#, so the ugly and inefficient code was used because:
* It ensured no typos in the `FullyQualifiedErrorId`.
* It was easy to search for all references to an error message via the property.
* Performance was not deemed important because they were errors.

Performance might matter more than I originally assumed given editor scenarios - and given the `nameof` language feature, it's reasonable (but annoying and a little verbose) to change error reporting to look more like:

```C#
ReportIncompleteInput(lCurly.Extent, rCurly.Extent,
    ParserStrings.MissingEndCurlyBrace, nameof(ParserStrings.MissingEndCurlyBrace));
```

In other words, pass the error text **and** the error id.

Alternatively you could just pass the error id and use reflection of some sort to load the error string, but that would not be good for performance.

src/System.Management.Automation/engine/parser/Parser.cs
daxian-dbw: If we handle the error before `goto finished_named_block_list`, then we will have to duplicate some of the code from `CompleteScriptBlockBody`, for example, the code to handle `lCurly == null`, and the code to handle `blockNameToken == <eof>`.

I will make some changes to `CompleteScriptBlockBody` to make it less confusing.

src/System.Management.Automation/engine/parser/Parser.cs
daxian-dbw: Agreed. Open #6444 to track this work.

src/System.Management.Automation/engine/parser/Parser.cs
lzybkr: I agree you don't want to duplicate unnecessarily. What about adding a case for `TokenKind.LCurly`, then consume any other token.

In the common case, you would not see a second error because there will just be a single token there, e.g. if you are in the middle of typing `process` or whatever.

daxian-dbw: @lzybkr I moved the error handling to `NamedBlockListRule`, can you please take another look?

src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -1629,6 +1630,20 @@ private ScriptBlockAst NamedBlockListRule(Token lCurly, List<UsingStatementAst>
                 {
                     default:
                         UngetToken(blockNameToken);
+                        extent = ExtentOf(startExtent, endExtent);
+
+                        // If 'lCurly' is null, then handle the next token in 'CompleteScriptBlockBody'.
+                        if (lCurly == null) { goto finished_named_block_list; }
```
lzybkr: This feels wrong - I'd expect the same error in a ps1/psm1 file as in a function/anonymous script block.

lzybkr (approved):

src/System.Management.Automation/engine/parser/Parser.cs
daxian-dbw: Fixed. But now it seems to me that this code block in `CompleteScriptBlockBody` will never be hit:
https://github.com/PowerShell/PowerShell/blob/cdc2015418e92f321311587d61570648ca392038/src/System.Management.Automation/engine/parser/Parser.cs#L1548-L1555

What shall we do with this code block?

daxian-dbw: I take that back. This block of code still can be hit when `lCurly == null` and `nextToken.Kind == RCurly`.

daxian-dbw:
@lzybkr Can you please take another look at the new commits? The additional change is to make sure we eat the unexcepted token in case that `lCurly == null` and `nextToken.Kind != EndOfInput`, just as what we do today in `CompleteScriptBlockBody`.

iSazonov:
@daxian-dbw Could you please add new error sample in the PR description?

src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -1620,15 +1620,30 @@ private ScriptBlockAst NamedBlockListRule(Token lCurly, List<UsingStatementAst>
...
                 Token blockNameToken = NextToken();
                 switch (blockNameToken.Kind)
                 {
                     default:
+                        // Next token is unexpected.
+                        // ErrorRecovery: if 'lCurly' is present, pretend we saw a closing curly; otherwise, eat the unexpected token.
+                        if (lCurly != null) { UngetToken(blockNameToken); }
+
+                        // If 'lCurly == null", then it's a ps1/psm1 file, and thus the extent is the whole file.
+                        scriptBlockExtent = lCurly != null ? ExtentOf(startExtent, endExtent) : _tokenizer.GetScriptExtent();
```
iSazonov: We already check `lCurly != null` in line 1634. Maybe use one "if"?

daxian-dbw: Fixed.

daxian-dbw:
@iSazonov The example in PR summary has already been updated.

iSazonov:
LGTM

lzybkr (approved):

Issue #3803, opened 2017-05-17
#requires on interactive console results in unhelpful error message


SteveL-MSFT:
Expected:
```powershell
PS C:\Users\slee\Repos\PowerShell> #requires
PS C:\Users\slee\Repos\PowerShell> 
```

Actual:
```powershell
PS C:\Users\slee\Repos\PowerShell> #requires
An error occurred while creating the pipeline.
    + CategoryInfo          : NotSpecified: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : RuntimeException
```

```powershell
Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
BuildVersion                   3.0.0.0
CLRVersion
GitCommitId                    v6.0.0-beta.1
OS                             Microsoft Windows 10.0.16193
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

rjmholt:
I've opened [an RFC](https://github.com/PowerShell/PowerShell-RFC/pull/122) on updating the behaviour of `#requires`. Please feel free to comment in the PR.

PR #6469, opened 2018-03-23
Add error handling for interactive #requires

rjmholt:
## PR Summary

Fix for #3803. Restores the tokenizer state more fully to prevent the null dereference causing the original error. Also changed the name of a variable to express that it doesn't actually mean that the PowerShell session is interactive or not.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/ex.ps1
+++ b/ex.ps1
@@ -1,4 +1,0 @@
-Write-Host "Hello"
-#requires
-7
-@($^, $$)
--- a/ex.ps1
+++ b/ex.ps1
@@ -1,0 +1,4 @@
+Write-Host "Hello"
+#requires
+7
+@($^, $$)
--- a/test/powershell/Language/Parser/Parser.Tests.ps1
+++ b/test/powershell/Language/Parser/Parser.Tests.ps1
@@ -915,4 +915,17 @@ foo``u{2195}abc
         # Issue #2780
         { ExecuteCommand "`$herestr=@`"`n'`"'`n`"@" } | Should Not Throw
     }
+
+    Context "#requires nested scan tokenizer tests" {
+        $testCases = @(
+            @{ script = "#requires"; firstToken = $null; lastToken = $null },
+            @{ script = "#requires -Version 5.0`n10"; firstToken = "10"; lastToken = "10" },
+            @{ script = "Write-Host 'Hello'`n#requires -Version 5.0`n7"; firstToken = "Write-Host"; lastToken = "7" },
+            @{ script = "Write-Host 'Hello'`n#requires -Version 5.0"; firstToken = "Write-Host"; lastToken = "'Hello'"}
+        )
+
+        It "Correctly resets the first and last tokens in the tokenizer after nested scan in script" -TestCases $testCases {
+            param($script, $firstToken, $lastToken)
+        }
+    }
 }
--- a/test/powershell/Language/Parser/Parser.Tests.ps1
+++ b/test/powershell/Language/Parser/Parser.Tests.ps1
@@ -917,15 +917,33 @@ foo``u{2195}abc
     }
 
     Context "#requires nested scan tokenizer tests" {
+        BeforeAll {
+            $settings = [System.Management.Automation.PSInvocationSettings]::new()
+            $settings.AddToHistory = $true
+
+            $ps = [powershell]::Create()
+        }
+
+        AfterEach {
+            $ps.Commands.Clear()
+        }
+
         $testCases = @(
             @{ script = "#requires"; firstToken = $null; lastToken = $null },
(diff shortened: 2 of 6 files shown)


rjmholt:
I've opened [an RFC](https://github.com/PowerShell/PowerShell-RFC/pull/122) on updating the behaviour of `#requires`. Please feel free to comment in the PR.

daxian-dbw (approved):

lzybkr (dismissed):
Can you add a test?

lzybkr:
@rjmholt - First/Last token are exposed through PowerShell variables `$^` and `$$`, so your test should check the variables.

`Parser.Tests.ps1` seems like a good place for this test.

test/powershell/Language/Parser/Parser.Tests.ps1
```diff
@@ -915,4 +915,35 @@ foo``u{2195}abc
...
+            @{ script = "#requires -Version 5.0`n10"; firstToken = "10"; lastToken = "10" },
+            @{ script = "Write-Host 'Hello'`n#requires -Version 5.0`n7"; firstToken = "Write-Host"; lastToken = "7" },
+            @{ script = "Write-Host 'Hello'`n#requires -Version 5.0"; firstToken = "Write-Host"; lastToken = "Hello"}
+        )
+
+        It "Correctly resets the first and last tokens in the tokenizer after nested scan in script" -TestCases $testCases {
+            param($script, $firstToken, $lastToken)
+
+            $ps.AddScript($script)
+            $ps.AddScript("(`$^,`$`$)")
```
lzybkr: I'm a little surprised this works, 2 calls to AddScript is I think like piping the first call to the second. That said, it neatly drops any results from the first call.

To be slightly more like the interactive scenario, I think you would call `$ps.Invoke` twice, clearing the commands between calls.

ex.ps1
```diff
@@ -0,0 +1,4 @@
+Write-Host "Hello"
```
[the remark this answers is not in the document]
rjmholt: Gah, I thought I checked that first. One sec

test/powershell/Language/Parser/Parser.Tests.ps1
rjmholt: Yeah, I tried that and it didn't seem to do what I wanted. To be honest, I wasn't entirely certain how I should be using the embedded PowerShell session to recreate the input. But I spent some time playing with it, tried the "script, invoke, tokens, invoke" way and seemed to get `$null` for everything.

daxian-dbw:
@rjmholt Can you please address @lzybkr's comments (_if there is any pending_)?

rjmholt:
@daxian-dbw I think this is all addressed. I just tried `$ps.Invoke()` again, but it's still as I describe in the comment above.

test/powershell/Language/Parser/Parser.Tests.ps1
```diff
@@ -915,4 +915,35 @@ foo``u{2195}abc
         # Issue #2780
         { ExecuteCommand "`$herestr=@`"`n'`"'`n`"@" } | Should Not Throw
     }
+
+    Context "#requires nested scan tokenizer tests" {
+        BeforeAll {
+            $settings = [System.Management.Automation.PSInvocationSettings]::new()
+            $settings.AddToHistory = $true
+
+            $ps = [powershell]::Create()
```
daxian-dbw: Please have a `AfterAll` block to dispose `$ps`.

test/powershell/Language/Scripting/Requires.Tests.ps1
```diff
@@ -27,4 +27,14 @@ Describe "Requires tests" -Tags "CI" {
...
-}
\ No newline at end of file
+    Context "Interactive requires" {
+        It "Successfully does nothing when given '#requires' interactively" {
+            $settings = [System.Management.Automation.PSInvocationSettings]::new()
+            $settings.AddToHistory = $true
+
+            $ps = [powershell]::Create()
+            $ps.AddScript("#requires`n10")
+            $ps.Invoke(@(), $settings) | Should -BeExactly 10
```
daxian-dbw: Please dispose `$ps` after you done with it.
Also, I think it's not necessary to use `PSInvocationSettings` in this test. Just `$ps = [powershell]::Create(); $ps.AddScript("#requires`n10").Invoke() | Should -BeExactly 10` will do.

daxian-dbw (changes requested):

test/powershell/Language/Scripting/Requires.Tests.ps1
daxian-dbw: I see, we need to hit the $^/$$ setting code. Never mind.

daxian-dbw (approved):

Issue #6305, opened 2018-03-03
get-childitem env: fails on Windows when an environment variable is defined twice, cased differently


ericsciple:
The problem:
`get-childitem env:` fails on Windows when an environment variable exists twice, each instance cased differently.

The scenario:
On Windows, set the environment variable `NPM_CONFIG_PATH` to control where the npm cache is created. And use npm to run unit tests, use powershell within a unit test, and from powershell invoke `get-childitem env:`. Note, it is common to use `npm test` to build and test a nodejs project.

It appears npm internally creates environment variables to manage state, and a lower-case variable `npm_config_path` is created. When powershell is invoked from the unit test, the environment variables `NPM_CONFIG_PATH` and `npm_config_path` are both set.

Steps to reproduce
------------------

On Windows, tested with npm v5.6.0 in my PATH.

```powershell
Write-Host "******************** SETTING NPM_CONFIG_PATH ********************"
$null = mkdir env-test
cd env-test
$env:NPM_CONFIG_PATH = $PWD
Write-Host "******************** CREATING PACKAGE.JSON ********************"
& npm init --force --yes
Write-Host "******************** MODIFYING PACKAGE.JSON ********************"
$package = Get-Content package.json | ConvertFrom-Json
$package.scripts.test = "pwsh -command Get-ChildItem env:"
Set-Content -LiteralPath package.json -Encoding UTF8 -Value (ConvertTo-Json -InputObject $package -Depth 99)
Get-Content package.json
Write-Host "******************** TESTING ENV PS DRIVE, VIA NPM ********************"
& npm test
```

Expected behavior
-----------------

PowerShell should gracefully handle the duplicate environment variable. The `set` builtin in cmd.exe handles it gracefully. .NET `Environment.GetEnvironmentVariables()` handles it gracefully.

Actual behavior
---------------

Get-ChildItem fails with \"An item with the same key has already been added. Key: NPM_CONFIG_PATH\"

Environment data
----------------

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.0.1
PSEdition                      Core
GitCommitId                    v6.0.1
OS                             Microsoft Windows 10.0.16299
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

TravisEz13:
Does not repro on macOS

[...]

Issue #6460, opened 2018-03-22
#6320 (fix for #6305) is flawed, because it may expose incorrect environment-variable values via Get-Item


mklement0:
A follow-up from https://github.com/PowerShell/PowerShell/issues/6305#issuecomment-375025005, as requested by @TravisEz13:

#6320 attempts to fix #6305, but only does so superficially:

* while it fixes the _immediate_ problem by preventing an _error_ from surfacing when you use `Get-item env:` while case-variant environment variables are unexpectedly defined,

* it can end up exposing an effectively _hidden_ value that neither `$env:<name>` nor `[environment]::GetEnvironmentVariable('<name>')` see _intermittently_; in other words: **`Get-Item env:foo` may intermittently report an incorrect value for `$env:foo`**

Steps to reproduce
------------------

Note: Run this in PowerShell _Core_ on _Windows_,  with Node.js installed and #6320 applied (without it, `Get-Item` will fail, complaining about a duplicate key).

```powershell
$env:FOO = 'A'
$tries = 0
do {
  Write-Host '.' -NoNewline
  ++$tries
  $res = node -pe @'
    env = {}
    env.fOo = 'B' // redefine with a case variant name and different value
    env.FOO = process.env.FOO // also include the original case variant with its original value.
    // Contrast $env:foo with Get-Item env:foo
    require('child_process').execSync(\"pwsh -nop -command $env:foo; (Get-Item env:foo).Value\", { env: env }).toString()
'@
} while ($res[0] -eq $res[1])

"
Values differed after $tries tries: '$($res[0])' vs. '$($res[1])'"
```

Expected behavior
-----------------

The loop should never exit, because `$env:foo` and  `Get-Item env:foo` should always report the same value.

Actual behavior
---------------

Because `Get-Item env:foo` intermittently reports `B` rather than `A`, the loop exits after a few tries with, e.g.:

```
Values differed after 3 tries: 'A' vs. 'B'
```

Fix
----

As proposed in https://github.com/PowerShell/PowerShell/issues/6305#issuecomment-371721480, in https://github.com/PowerShell/PowerShell/pull/6320/files#diff-1bee4b279bb56aca25fb479994641529, replace:

```csharp
  // Windows only: duplicate key (variable name that differs only in case)
  // NOTE: Even though this shouldn't happen, it can, e.g. when npm
  //       creates duplicate environment variables that differ only in case -
  //       see https://github.com/PowerShell/PowerShell/issues/6305.
[...]

PR #6320, opened 2018-03-06
Fix error in windows provider when the environment as an existing set of variables name that only differs by case

TravisEz13:
## PR Summary

fix error in windows provider when the environment as an existing set of variables name that only differs by case
    - make the provider storage for the environment on windows ignore duplicates
    - add tests to verify existing environment get-item behavior

fixes #6305 

## PR Checklist

- [x] [PR has a meaningful title](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - Use the present tense and imperative mood when describing your changes
- [x] [Summarized changes](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [x] [Change is not breaking](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes)
- [x] [Make sure all `.h`, `.cpp`, `.cs`, `.ps1` and `.psm1` files have the correct copyright header](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [x] This PR is ready to merge and is not [Work in Progress](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---work-in-progress).
    - If the PR is work in progress, please add the prefix `WIP:` to the beginning of the title and remove the prefix when the PR is ready.
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - [ ] User-facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
        - [ ] Issue filed - Issue link:
- **Testing - New and feature**
    - [x] Not Applicable or can only be tested interactively
    - **OR**
    - [ ] [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
        - [ ] [Add `[feature]` if the change is significant or affects feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

PR #6489, opened 2018-03-24
Fix error in windows provider when the environment has accidental duplicates that differ only by case

mklement0:
## PR Summary

Fix error in windows provider when the environment has accidental duplicates that differ only by case.

- Make the provider storage for the environment on windows ignore duplicates  and only report the _effective_ value.
- Add tests to verify existing environment get-item behavior and to ensure that `Get-Item env:<var>` reports the same as `$env:<var>`, namely the effective value.

Fixes #6305 and supersedes #6320 based on discussion in #6460.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/namespaces/EnvironmentProvider.cs
+++ b/src/System.Management.Automation/namespaces/EnvironmentProvider.cs
@@ -198,15 +198,33 @@ namespace Microsoft.PowerShell.Commands
             IDictionary environmentTable = Environment.GetEnvironmentVariables();
             foreach (DictionaryEntry entry in environmentTable)
             {
-                // Windows only: duplicate key (variable name that differs only in case)
-                // NOTE: Even though this shouldn't happen, it can, e.g. when npm
-                //       creates duplicate environment variables that differ only in case -
-                //       see https://github.com/PowerShell/PowerShell/issues/6305.
-                //       However, because retrieval *by name* later is invariably
-                //       case-Insensitive, in effect only a *single* variable exists.
-                //       We simply ask Environment.GetEnvironmentVariable() which value is
-                //       the effective one, and use that.
-                providerTable.TryAdd((string)entry.Key, entry);
+                try
+                {
+                    providerTable.Add((string)entry.Key, entry);
+                }
+                catch (System.ArgumentException)
+                {   // Windows only: duplicate key (variable name that differs only in case)
+                    // NOTE: Even though this shouldn't happen, it can, e.g. when npm
+                    //       creates duplicate environment variables that differ only in case -
+                    //       see https://github.com/PowerShell/PowerShell/issues/6305.
+                    //       However, because retrieval *by name* later is invariably
+                    //       case-INsensitive, in effect only a *single* variable exists.
+                    //       We simply ask Environment.GetEnvironmentVariable() for the effective value 
+                    //       and use that as the only entry, because for a given key 'foo' (and all its case variations),
+                    //       that is guaranteed to match what $env:FOO and [environment]::GetEnvironmentVariable('foo') return. 
+                    //       (If, by contrast, we just used `entry` as-is every time a duplicate is encountered,
+                    //        it could - intermittently - represent a value *other* than the effective one.)
+                    string effectiveValue = Environment.GetEnvironmentVariable((string)entry.Key);
+                    if (((string)entry.Value).Equals(effectiveValue, StringComparison.Ordinal)) { // We've found the effective definition.
+                        // Note: We *recreate* the entry so that the specific name casing of the
+                        //       effective definition is also reflected. However, if the case variants
+                        //       define the same value, it is unspecified which name variant is reflected
+                        //       in Get-Item env: output; given the always case-insensitive nature of the retrieval, 
+                        //       that shouldn't matter.
+                        providerTable.Remove((string)entry.Key);
+                        providerTable.Add((string)entry.Key, entry);
+                    }
+                }
             }
 
             return providerTable;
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/namespaces/EnvironmentProvider.cs
```diff
@@ -198,15 +198,33 @@ internal override IDictionary GetSessionStateTable()
...
-                //       However, because retrieval *by name* later is invariably
-                //       case-Insensitive, in effect only a *single* variable exists.
-                //       We simply ask Environment.GetEnvironmentVariable() which value is
-                //       the effective one, and use that.
-                providerTable.TryAdd((string)entry.Key, entry);
+                try
+                {
+                    providerTable.Add((string)entry.Key, entry);
+                }
+                catch (System.ArgumentException)
```
iSazonov: We could use
```c#
if (!providerTable.Add((string)entry.Key, entry))
{
    string effectiveValue = Environment.GetEnvironmentVariable((string)entry.Key); 
   ...
}

File: test/powershell/Modules/Microsoft.PowerShell.Management/Get-Item.Tests.ps1 [hunk 2]
```diff
@@ -141,5 +141,24 @@ Describe "Get-Item" -Tags "CI" {
 
             (get-item env:\testvar).Value | Should -BeExactly $expectedValue
         }
+
+        It "get-item reports the effective value among accidental case-variant duplicates on Windows" -skip:(-not $isWindows) {
+            if (-not (Get-Command -ErrorAction Ignore node.exe)) {
+                Write-Warning "Test skipped, because Node.js is required to run it."
```
iSazonov: We should not depend on Node.js - please implement a mock/help function.

mklement0: @iSazonov: Node.js is the only way that I know of to create this scenario - all Windows APIs seemingly - and sensibly - prevent the creation of case-variant duplicates.

src/System.Management.Automation/namespaces/EnvironmentProvider.cs
mklement0: @iSazonov: Assuming you meant `.TryAdd()`: good idea; done.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Item.Tests.ps1
mklement0: So, unless we do come up with a third-party-free way to mock this up, we're faced with the choice between leaving it in with the warning or removing the test altogether.

iSazonov: I wonder - what api Node.js use on _Windows_?

iSazonov: https://stackoverflow.com/questions/573817/where-are-environment-variables-stored-in-registry

mklement0: @iSazonov: Re what API Node.js uses - that's an excellent question, but, honestly, not one I feel personally inspired to investigate. As for the SO link: The Windows registry is inherently case-INSENSITIVE, so by definition it cannot be involved here - we're talking about dynamically created process-level environment variables.

TravisEz13: @iSazonov It's definitely not the registry as it is updating the process environment, which is in-memory.

TravisEz13: I agree the node requirement should not be added to a CI test.  Can you move this to a different describe and make it a Feature test?

TravisEz13 (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Item.Tests.ps1
mklement0: @TravisEz13: Happy to do it, but I need guidance: you mean a separate `Describe` block with `-tags 'Scenario'`? In the same file? Anything else (more formal description of the context and what's needed to run the test)?

TravisEz13: @yes, the implementation details you describe look correct.  The goal is just to prevent everyone from having to have Node.js.  We can easily enough ensure it's available on the machine we run the scenario tests.

TravisEz13 (approved):

iSazonov (approved):

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Item.Tests.ps1
```diff
@@ -141,5 +141,33 @@ Describe "Get-Item" -Tags "CI" {
 
             (get-item env:\testvar).Value | Should -BeExactly $expectedValue
         }
+
```
iSazonov: Please remove extra line.

test/powershell/Modules/Microsoft.PowerShell.Management/Get-Item.Tests.ps1
```diff
@@ -141,5 +141,33 @@ Describe "Get-Item" -Tags "CI" {
...
+    }
+
```
iSazonov: Please remove extra line.

iSazonov (approved):

iSazonov:
@mklement0 Thanks for the fix!

TravisEz13:
@mklement0  Yes, thanks for all your effort.

mklement0:
@TravisEz13, @iSazonov: My pleasure; I'm glad we arrived at a shared understanding.

Issue #6482, opened 2018-03-23
$PSVersionTable.PSVersion throws error System.ArgumentOutOfRangeException,Microsoft.PowerShell.Commands.OutLineOutputCommand


adityapatwardhan:
$PSVersionTable.PSVersion.ToString() and $PSVersionTable still works.

The error can be reproduced on master branch.

Steps to reproduce
------------------

```powershell

$PSVersionTable.PSVersion
out-lineoutput : Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: index
+ CategoryInfo          : NotSpecified: (:) [out-lineoutput], ArgumentOutOfRangeException
+ FullyQualifiedErrorId : System.ArgumentOutOfRangeException,Microsoft.PowerShell.Commands.OutLineOutputCommand

```

Expected behavior
-----------------

```none
Get the PSVersion string
```

Actual behavior
---------------
Error - System.ArgumentOutOfRangeException,Microsoft.PowerShell.Commands.OutLineOutputCommand

```none

```

Environment data
----------------

```powershell
> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.0.1
PSEdition                      Core
GitCommitId                    v6.0.1-290-g942832c13a3c325851e9873bd25036775c85a18c
OS                             Microsoft Windows 10.0.16299
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

```

MaximoTrinidad:
Hum! So, this is for the future release of PSCore 6.1.0.  Right!

It's find on 6.0.1: Win 10, Win 10 Bash, and Ubuntu.
:)

SteveL-MSFT:
This was found during 6.1.0-Preview1 release.  I know the issue, just working on some tests.

stknohg:
Workaround :

### use Format-Table -AutoSize

```powershell
PS C:\> $PSVersionTable.PSVersion | ft -AutoSize

Major Minor Patch PreReleaseLabel BuildLabel
----- ----- ----- --------------- ----------
6     1     0     preview.1
```

### use other formatting

```powershell
PS C:\> $PSVersionTable.PSVersion | fl


Major           : 6
Minor           : 1
Patch           : 0
PreReleaseLabel : preview.1
BuildLabel      :
```

PR #6487, opened 2018-03-24
fix for #6482 - Make sure that the width of the column is a least the size of the label for Semantic Version

JamesWTruher:
## PR Summary
which in this case is the same as the property name

When the label is too short, an unhelpful error will be emitted
this fixes https://github.com/PowerShell/PowerShell/issues/6482
 
## PR Checklist

- [x] [PR has a meaningful title](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - Use the present tense and imperative mood when describing your changes
- [x] [Summarized changes](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [x] [Change is not breaking](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes)
- [x] [Make sure all `.h`, `.cpp`, `.cs`, `.ps1` and `.psm1` files have the correct copyright header](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [x] This PR is ready to merge and is not [Work in Progress](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---work-in-progress).
    - If the PR is work in progress, please add the prefix `WIP:` to the beginning of the title and remove the prefix when the PR is ready.
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - [ ] User-facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
        - [x] Issue filed - Issue link: https://github.com/PowerShell/PowerShell/issues/6482
- **Testing - New and feature**
    - [ ] Not Applicable or can only be tested interactively
    - **OR**
    - [x] [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
        - [ ] [Add `[feature]` if the change is significant or affects feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

PR #6491, opened 2018-03-24
Fix error when 'Format-Wide -AutoSize | Out-String' is called.

### LINKED ISSUE #6471: "Get-ChildItem | Format-Wide -AutoSize | Out-String" can't work on Powershell 6.0.1 & 6.0.2
Steps to reproduce
------------------

```powershell
Get-ChildItem | Format-Wide -AutoSize | Out-String
```

or
```powershell
Get-ChildItem | Format-Wide -AutoSize | Out-String -Stream
```

Expected behavior
-----------------

```none


    Directory: C:\Users\andot\Documents\WindowsPowerShell\Modules\lolcat\1.0.1



.git                        LICENSE
lolcat.psd1                        lolcat.psm1
Out-Rainbow.psm1                   Out-StripAnsi.psm1
README.md                   screenshot.png

```

Actual behavior
---------------

```none
out-lineoutput : Array dimensions exceeded supported range.
At line:1 char:1
+ Get-ChildItem | Format-Wide -AutoSize | Out-String -Stream
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : NotSpecified: (:) [out-lineoutput], OutOfMemoryException
+ FullyQualifiedErrorId : System.OutOfMemoryException,Microsoft.PowerShell.Commands.OutLineOutputCommand
```

Environment data
----------------

```powershell
> $PSVersionTable
```

```powershell
> $PSVersionTable                                                 
PSVersion                      6.0.2
GitCommitId                    v6.0.2
OS                             Linux 4.4.0-43-Microsoft #1-Microsoft Wed Dec...
```

If no ` -AutoSize` param, the result will be fine. And it works well in WindowsPowerShell  5.1 with or without ` -AutoSize` param.

stknohg:
## PR Summary

Fix #6471.

This pull request fixes the error when  `Format-Wide -AutoSize | Out-String` is called.  

From #5193,  `int.MaxValue` is adopted as output width when `Out-String -Width` parameter is not specified.  
`OutCommandInner.TableOutputContext` class is checking output width at initializing.

https://github.com/PowerShell/PowerShell/blob/36b71ba39e36be3b86854b3551ef9f8e2a1de5cc/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs#L923-L939

This pull request applies the same method to `OutCommandInner.WideOutputContext` class.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [x] Issue filed - Issue link: #6471
    - [] Not Applicable or can only be tested interactively
    - **OR**

### DIFF
--- a/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
@@ -743,6 +743,8 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             _formattingHint = hint;
         }
 
+        static private int _failedToReadConsoleWidth = 0;
+
         static private int GetConsoleWindowWidth(int columnNumber)
         {
             if (columnNumber == int.MaxValue)
@@ -752,13 +754,18 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 // If we detect that int.MaxValue is used, first we try to get the current console window width.
                 // However, if we can't read that (for example, implicit remoting has no console window), we default
                 // to something reasonable: 120 columns.
+                if (_failedToReadConsoleWidth > 0)
+                {
+                    return _failedToReadConsoleWidth;
+                }
                 try
                 {
                     return Console.WindowWidth;
                 }
                 catch
                 {
-                    return 120;
+                    _failedToReadConsoleWidth = 120;
+                    return _failedToReadConsoleWidth;
                 }
             }
             return columnNumber;
--- a/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
@@ -743,10 +743,14 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             _formattingHint = hint;
         }
 
-        static private int _failedToReadConsoleWidth = 0;
+        // In cases like implicit remoting, there is no console so reading the console width results in an exception.
+        // We cache this value rather than handling the exception (on the target) everytime as a performance improvement.
+        static private bool _noConsole = false;
 
         static private int GetConsoleWindowWidth(int columnNumber)
         {
+            const int defaultConsoleWidth = 120;
+
             if (columnNumber == int.MaxValue)
             {
                 // Tables and Wides need to use spaces for padding to maintain table look even if console window is resized.
@@ -754,9 +758,9 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 // If we detect that int.MaxValue is used, first we try to get the current console window width.
                 // However, if we can't read that (for example, implicit remoting has no console window), we default
                 // to something reasonable: 120 columns.
-                if (_failedToReadConsoleWidth > 0)
+                if (_noConsole)
                 {
-                    return _failedToReadConsoleWidth;
+                    return defaultConsoleWidth;
                 }
                 try
                 {
@@ -764,8 +768,8 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 }
                 catch
                 {
-                    _failedToReadConsoleWidth = 120;
-                    return _failedToReadConsoleWidth;
+                    _noConsole = true;
+                    return defaultConsoleWidth;
                 }
             }
             return columnNumber;
--- a/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
@@ -743,21 +743,25 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             _formattingHint = hint;
         }
 
-        // In cases like implicit remoting, there is no console so reading the console width results in an exception.
-        // We cache this value rather than handling the exception (on the target) everytime as a performance improvement.
+        /// <summary>
+        /// In cases like implicit remoting, there is no console so reading the console width results in an exception. 
+        /// Instead of handling exception every time we cache this value to increase performance. 
+        /// </summary>
         static private bool _noConsole = false;
 
+        /// <summary>
+        /// Tables and Wides need to use spaces for padding to maintain table look even if console window is resized.
+        /// For all other output, we use int.MaxValue if the user didn't explicitly specify a width.
+        /// If we detect that int.MaxValue is used, first we try to get the current console window width.
+        /// However, if we can't read that (for example, implicit remoting has no console window), we default
+        /// to something reasonable: 120 columns.
+        /// </summary>
         static private int GetConsoleWindowWidth(int columnNumber)
         {
             const int defaultConsoleWidth = 120;
 
             if (columnNumber == int.MaxValue)
             {
-                // Tables and Wides need to use spaces for padding to maintain table look even if console window is resized.
-                // For all other output, we use int.MaxValue if the user didn't explicitly specify a width.
-                // If we detect that int.MaxValue is used, first we try to get the current console window width.
-                // However, if we can't read that (for example, implicit remoting has no console window), we default
-                // to something reasonable: 120 columns.
                 if (_noConsole)
                 {
                     return defaultConsoleWidth;
--- a/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
@@ -743,6 +743,27 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             _formattingHint = hint;
         }
 
+        static private int TryGetColumnsOnTheScreen(int columnNumber)
+        {
+            // Tables and Wides need to use spaces for padding to maintain table look even if console window is resized.
+            // For all other output, we use int.MaxValue if the user didn't explicitly specify a width.
+            // If we detect that int.MaxValue is used, first we try to get the current console window width.
+            // However, if we can't read that (for example, implicit remoting has no console window), we default
+            // to something reasonable: 120 columns.
+            if (columnNumber == int.MaxValue)
+            {
+                try
+                {
+                    return Console.WindowWidth;
+                }
+                catch
+                {
+                    return 120;
+                }
+            }
+            return columnNumber;
+        }
+
         /// <summary>
         /// base class for all the formatting hints
         /// </summary>
@@ -910,23 +931,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                     columnWidthsHint = tableHint.columnWidths;
                 }
 
-                int columnsOnTheScreen = this.InnerCommand._lo.ColumnNumber;
-                // Tables need to use spaces for padding to maintain table look even if console window is resized.
-                // For all other output, we use int.MaxValue if the user didn't explicitly specify a width.
-                // If we detect that int.MaxValue is used, first we try to get the current console window width.
-                // However, if we can't read that (for example, implicit remoting has no console window), we default
-                // to something reasonable: 120 columns.
-                if (columnsOnTheScreen == int.MaxValue)
-                {
-                    try
-                    {
-                        columnsOnTheScreen = Console.WindowWidth;
-                    }
-                    catch
-                    {
-                        columnsOnTheScreen = 120;
-                    }
-                }
+                int columnsOnTheScreen = TryGetColumnsOnTheScreen(this.InnerCommand._lo.ColumnNumber);
 
                 int columns = this.CurrentTableHeaderInfo.tableColumnInfoList.Count;
                 if (columns == 0)
@@ -1132,18 +1137,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 // get the header info and the view hint
                 WideFormattingHint hint = this.InnerCommand.RetrieveFormattingHint() as WideFormattingHint;
 
-                int columnsOnTheScreen = this.InnerCommand._lo.ColumnNumber;
-                if (columnsOnTheScreen == int.MaxValue)
-                {
-                    try
-                    {
-                        columnsOnTheScreen = Console.WindowWidth;
-                    }
-                    catch
-                    {
-                        columnsOnTheScreen = 120;
-                    }
-                }
+                int columnsOnTheScreen = TryGetColumnsOnTheScreen(this.InnerCommand._lo.ColumnNumber);
 
                 // give a preference to the hint, if there
                 if (hint != null && hint.maxWidth > 0)
--- a/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
@@ -743,15 +743,15 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             _formattingHint = hint;
         }
 
-        static private int TryGetColumnsOnTheScreen(int columnNumber)
+        static private int GetConsoleWindowWidth(int columnNumber)
         {
-            // Tables and Wides need to use spaces for padding to maintain table look even if console window is resized.
-            // For all other output, we use int.MaxValue if the user didn't explicitly specify a width.
-            // If we detect that int.MaxValue is used, first we try to get the current console window width.
-            // However, if we can't read that (for example, implicit remoting has no console window), we default
-            // to something reasonable: 120 columns.
             if (columnNumber == int.MaxValue)
             {
+                // Tables and Wides need to use spaces for padding to maintain table look even if console window is resized.
+                // For all other output, we use int.MaxValue if the user didn't explicitly specify a width.
+                // If we detect that int.MaxValue is used, first we try to get the current console window width.
+                // However, if we can't read that (for example, implicit remoting has no console window), we default
+                // to something reasonable: 120 columns.
                 try
                 {
                     return Console.WindowWidth;
@@ -931,7 +931,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                     columnWidthsHint = tableHint.columnWidths;
                 }
 
-                int columnsOnTheScreen = TryGetColumnsOnTheScreen(this.InnerCommand._lo.ColumnNumber);
+                int columnsOnTheScreen = GetConsoleWindowWidth(this.InnerCommand._lo.ColumnNumber);
 
                 int columns = this.CurrentTableHeaderInfo.tableColumnInfoList.Count;
                 if (columns == 0)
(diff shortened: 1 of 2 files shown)


iSazonov:
Do this fix  #6482 (PR #6487) too?

test/powershell/Modules/Microsoft.PowerShell.Utility/Format-Wide.Tests.ps1
```diff
@@ -15,6 +15,8 @@ Describe "Format-Wide" -Tags "CI" {
 
     It "Should be able to use the autosize switch" {
         { Get-ChildItem | Format-Wide -Autosize } | Should -Not -Throw
+        # Test for #6471
```
iSazonov: Please add more useful comment why we should keep next test line.

test/powershell/Modules/Microsoft.PowerShell.Utility/Format-Wide.Tests.ps1
```diff
@@ -15,6 +15,8 @@ Describe "Format-Wide" -Tags "CI" {
...
+        { Get-ChildItem | Format-Wide -Autosize | Out-String } | Should -Not -Throw
```
iSazonov: What is current directory empty - do we get an exception without the fix?

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
```diff
@@ -1132,10 +1132,23 @@ internal override void Initialize()
...
                 WideFormattingHint hint = this.InnerCommand.RetrieveFormattingHint() as WideFormattingHint;
 
+                int columnsOnTheScreen = this.InnerCommand._lo.ColumnNumber;
+                if (columnsOnTheScreen == int.MaxValue)
+                {
+                    try
+                    {
+                        columnsOnTheScreen = Console.WindowWidth;
+                    }
+                    catch
```
iSazonov: `int.MaxValue` is default - can we exclude the exception?

stknohg:
@iSazonov This fix only #6471.

test/powershell/Modules/Microsoft.PowerShell.Utility/Format-Wide.Tests.ps1
stknohg: Other tests are also invoked on the premise that there are files in the folder.
I think that it is probably okay.

iSazonov: /cc @adityapatwardhan @JamesWTruher Can you approve the test?

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
stknohg: `OutCommandInner.TableOutputContext` class uses ` try-catch` statement assuming an environment without a console window such as implicit remoting.
Therefore, I do the same thing with `OutCommandInner.WideOutputContext` class.

>  // Tables need to use spaces for padding to maintain table look even if console window is resized. 
> // For all other output, we use int.MaxValue if the user didn't explicitly specify a width. 
>  // If we detect that int.MaxValue is used, first we try to get the current console window width. 
>  // However, if we can't read that (for example, implicit remoting has no console window), we default 
>  // to something reasonable: 120 columns.

iSazonov: We already have reports of poor performance of the console. It is too costly to handle the exception for each row. My question is whether it is possible to exclude it.

SteveL-MSFT: Since there are now at least two cases of this usage.  Perhaps we can have a static function get the windowwidth and if it caught an exception once, then have it return the default each time.

test/powershell/Modules/Microsoft.PowerShell.Utility/Format-Wide.Tests.ps1
SteveL-MSFT: I think we should make this more predictable and consistent.  How about `Get-ChildItem $TestDrive` (and create at least one file there).  And since the test is for formatting and not testing `Get-ChildItem`, I would suggest doing that call once in `BeforeAll` and saving it to be used in the other tests.

SteveL-MSFT (changes requested):

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
```diff
@@ -743,6 +743,27 @@ private void ProcessCachedGroupOnWide(WideViewHeaderInfo wvhi, List<PacketInfoDa
             _formattingHint = hint;
         }
 
+        static private int TryGetColumnsOnTheScreen(int columnNumber)
```
iSazonov: Usually `Try...()` fuctions return `bool`. 
Also we get mainly `Console.WindowWidth`. So I think we could use `GetConsoleWindowWidth` name.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
```diff
@@ -743,6 +743,27 @@ private void ProcessCachedGroupOnWide(WideViewHeaderInfo wvhi, List<PacketInfoDa
...
+        {
+            // Tables and Wides need to use spaces for padding to maintain table look even if console window is resized.
+            // For all other output, we use int.MaxValue if the user didn't explicitly specify a width.
+            // If we detect that int.MaxValue is used, first we try to get the current console window width.
+            // However, if we can't read that (for example, implicit remoting has no console window), we default
+            // to something reasonable: 120 columns.
+            if (columnNumber == int.MaxValue)
```
iSazonov: Seems the comment is more for the method then the `if`. I believe we sgould move the comment.

test/powershell/Modules/Microsoft.PowerShell.Utility/Format-Wide.Tests.ps1
```diff
@@ -1,35 +1,40 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 Describe "Format-Wide" -Tags "CI" {
+    BeforeAll {
+        1..10 | ForEach-Object { New-Item -Path ("TestDrive:\Testdir{0:00}" -f $_) -ItemType Directory }
+        1..10 | ForEach-Object { New-Item -Path ("TestDrive:\TestFile{0:00}.txt" -f $_) -ItemType File }
```
iSazonov: 10 is very large. 2 is enough.

test/powershell/Modules/Microsoft.PowerShell.Utility/Format-Wide.Tests.ps1
```diff
@@ -1,35 +1,40 @@
...
+    }
 
     It "Should have the same output between the alias and the unaliased function" {
-        $nonaliased = Get-ChildItem | Format-Wide
-        $aliased    = Get-ChildItem | fw
+        $nonaliased = Get-ChildItem $TestDrive | Format-Wide
+        $aliased = Get-ChildItem $TestDrive | fw
```
iSazonov: Please remove the alias and use full name.

My understanding @SteveL-MSFT request is that we should move `Get-ChildItem $TestDrive` to `BeforeAll` and reuse it in all `It` blocks:
```c#
$pathList = Get-ChildItem $TestDrive
...
$aliased = $pathList | Format-Wide 
```

test/powershell/Modules/Microsoft.PowerShell.Utility/Format-Wide.Tests.ps1
```diff
@@ -1,35 +1,40 @@
...
+        { Get-ChildItem $TestDrive | Format-Wide -Column 3 } | Should -Not -Throw
     }
 
     It "Should be able to use the autosize switch" {
-        { Get-ChildItem | Format-Wide -Autosize } | Should -Not -Throw
-        # Test for #6471
-        { Get-ChildItem | Format-Wide -Autosize | Out-String } | Should -Not -Throw
+        { Get-ChildItem $TestDrive | Format-Wide -Autosize } | Should -Not -Throw
+        # 'Format-Wide -AutoSize | Out-String' fails on PowerShell Core 6.0.1. (issue #6471)
+        # so we add a new test for that.
```
iSazonov: Please remove unneeded comment. Each test is important.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
stknohg: That's right. I will rename `TryGetColumnsOnTheScreen` to `GetConsoleWindowWidth`.

test/powershell/Modules/Microsoft.PowerShell.Utility/Format-Wide.Tests.ps1
stknohg: I will remove it.

test/powershell/Modules/Microsoft.PowerShell.Utility/Format-Wide.Tests.ps1
stknohg: I will fix it.

test/powershell/Modules/Microsoft.PowerShell.Utility/Format-Wide.Tests.ps1
```diff
@@ -1,33 +1,39 @@
...
+        1..2 | ForEach-Object { New-Item -Path ("TestDrive:\Testdir{0:00}" -f $_) -ItemType Directory }
+        1..2 | ForEach-Object { New-Item -Path ("TestDrive:\TestFile{0:00}.txt" -f $_) -ItemType File }
+        $pathList = Get-ChildItem $TestDrive
+    }
 
     It "Should have the same output between the alias and the unaliased function" {
-        $nonaliased = Get-ChildItem | Format-Wide
-        $aliased    = Get-ChildItem | fw
+        $nonaliased = $pathList | Format-Wide
+        $aliased = $pathList | fw
```
iSazonov: Please replace the alias.

stknohg: What does `replace` mean?
This test compares the results of alias and unaliased command.

iSazonov: Ah, sorry. We have special tests for aliases and shouldn't test it here. Pleas remove.

stknohg: I understand. I will remove this test.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
```diff
@@ -743,6 +743,27 @@ private void ProcessCachedGroupOnWide(WideViewHeaderInfo wvhi, List<PacketInfoDa
...
+                // If we detect that int.MaxValue is used, first we try to get the current console window width.
+                // However, if we can't read that (for example, implicit remoting has no console window), we default
+                // to something reasonable: 120 columns.
+                try
+                {
+                    return Console.WindowWidth;
+                }
+                catch
+                {
+                    return 120;
```
SteveL-MSFT: I think we can safely assert that if trying to get the WindowWidth ever fails, it will always fail.  So if we catch here, I think we can set a private flag like _failedToReadConsoleWidth and just return the default rather than incurring the cost of handling the exception each time (I think that was @iSazonov's original concern)

SteveL-MSFT (changes requested):

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
stknohg: Does it really always fail?

I will present a simple example.   
I simply add `_failedToReadConsoleWidth` flag as follows.

```csharp
static private int _failedToReadConsoleWidth = 0;

static private int GetConsoleWindowWidth(int columnNumber)
{
    if (columnNumber == int.MaxValue)
    {
        // Tables and Wides need to use spaces for padding to maintain table look even if console window is resized.
        // For all other output, we use int.MaxValue if the user didn't explicitly specify a width.
        // If we detect that int.MaxValue is used, first we try to get the current console window width.
        // However, if we can't read that (for example, implicit remoting has no console window), we default
        // to something reasonable: 120 columns.
        if (_failedToReadConsoleWidth > 0)
        {
            return _failedToReadConsoleWidth;
        }
        try
        {
            return Console.WindowWidth;
        }
        catch
        {
            _failedToReadConsoleWidth = 120;
            return _failedToReadConsoleWidth;
        }
    }
    return columnNumber;
}
```

then invoke four commands in order.

* command1 : Invoke local command (Console.WindowWidth is available)
* command2 : Invoke implict remoting command (Console.WindowWidth isn't available)
* command3 : Invoke implict remoting command (Console.WindowWidth isn't available)
* command4 : Invoke local command (Console.WindowWidth is available)

value of `_failedToReadConsoleWidth` is changed to 120 after command2, and this value is used in command3.
But when we invoke command4, I think we have to use `Console.WindowWidth`.

iSazonov: We should reset the flag.

stknohg: Hmm...

In order to reset the flag, we need to be able to determine if `Console.WindowWidth` can get.  
if we can determine whether we can get `Console.WindowWidth` from the start, we do not need to use `try-catch` statement.

I am sorry but I have no idea that how to determine if `Console.WindowWidth` can get...

SteveL-MSFT: In the implicit remoting case, the command would be executed on the remote side where it wouldn't be able to read the WindowWidth.  So the cached value would only be the remote side.  Local commands should still work.

stknohg: @SteveL-MSFT Thank you. My premise was wrong.
I will fix the code as above.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
```diff
@@ -743,6 +743,34 @@ private void ProcessCachedGroupOnWide(WideViewHeaderInfo wvhi, List<PacketInfoDa
...
+
+        static private int GetConsoleWindowWidth(int columnNumber)
+        {
+            if (columnNumber == int.MaxValue)
+            {
+                // Tables and Wides need to use spaces for padding to maintain table look even if console window is resized.
+                // For all other output, we use int.MaxValue if the user didn't explicitly specify a width.
+                // If we detect that int.MaxValue is used, first we try to get the current console window width.
+                // However, if we can't read that (for example, implicit remoting has no console window), we default
+                // to something reasonable: 120 columns.
```
iSazonov: I'd move the comment before the method and add clarification about `_failedToReadConsoleWidth` and remote session.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
```diff
@@ -743,6 +743,34 @@ private void ProcessCachedGroupOnWide(WideViewHeaderInfo wvhi, List<PacketInfoDa
...
+                if (_failedToReadConsoleWidth > 0)
```
iSazonov: I believe bool type is more suitable.
I'd rename this variable to reflect that a console is missing from the remote session.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
stknohg: OK.

However, I am not good at English and I do not have much confidence to make appropriate comments.
Is there any good comment plan?

iSazonov: @SteveL-MSFT Could you please help with the comment?

SteveL-MSFT: @stknohg Your English is actually quite good.  I believe @iSazonov is asking to move the comment to where line 748 right before the method.  You should probably just add some text before line 746 like:

```powershell
// In cases like implicit remoting, there is no console so reading the console width results in an exception.
// We cache this value rather than handling the exception (on the target) everytime as a performance improvement.
```

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
SteveL-MSFT: Something like `bool _noConsole`

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
stknohg: @SteveL-MSFT Thank you.I will add your comment.

SteveL-MSFT: @stknohg I think this PR is almost there.  Just need to move the comment from lines 756-760 to right before line 750 since the comment applies to the entire method and not just the `if` statement within the method

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
```diff
@@ -743,6 +743,38 @@ private void ProcessCachedGroupOnWide(WideViewHeaderInfo wvhi, List<PacketInfoDa
             _formattingHint = hint;
         }
 
+        // In cases like implicit remoting, there is no console so reading the console width results in an exception.
+        // We cache this value rather than handling the exception (on the target) everytime as a performance improvement.
+        static private bool _noConsole = false;
+
```
iSazonov: Sorry that I again feedback about comments.
We should follow the file pattern. Also I reword the comment.
```c#
/// <summary>
/// In cases like implicit remoting, there is no console so reading the console width results in an exception. 
/// Instead of handling exception every time we cache this value to increase performance. 
/// </summary>
static private bool _noConsole = false; 

/// <summary>
/// Tables and Wides need to use spaces for padding to maintain table look even if console window is resized.
/// For all other output, we use int.MaxValue if the user didn't explicitly specify a width.
/// If we detect that int.MaxValue is used, first we try to get the current console window width.
/// However, if we can't read that (for example, implicit remoting has no console window), we default
/// to something reasonable: 120 columns.
/// </summary>
static private int GetConsoleWindowWidth(int columnNumber)

stknohg: No problem. I appreciate it.

iSazonov (approved):

SteveL-MSFT (approved):
Thanks for the contribution!

iSazonov:
@stknohg Thanks for the contribution!

Thank you for being patient - we want to have clean and clear code and helpful comments.
Now you have a useful experience and we would be happy to see your new contributions in formatting subsystem and other areas.

PR #6498, opened 2018-03-26
Change Parser SaveError to use `nameof` instead of LINQ expressions

### LINKED ISSUE #6444: Update 'SaveError' in 'Parser.cs' to not depend on LINQ expression tree to get the error id
Quote Jason's comment at https://github.com/PowerShell/PowerShell/pull/6434/files/5b5be3122971552d2ce208858ac84868954b841c#r175933192

I'd suggest removing my ugly hack of using LINQ expression trees in `SaveError` and related helper methods, and replace with `nameof(ParserStrings.MissingEndCurlyBrace)`.

When that code was written, `nameof` was not in C#, so the ugly and inefficient code was used because:
* It ensured no typos in the `FullyQualifiedErrorId`.
* It was easy to search for all references to an error message via the property.
* Performance was not deemed important because they were errors.

Performance might matter more than I originally assumed given editor scenarios - and given the `nameof` language feature, it's reasonable (but annoying and a little verbose) to change error reporting to look more like:

```C#
ReportIncompleteInput(lCurly.Extent, rCurly.Extent,
    ParserStrings.MissingEndCurlyBrace, nameof(ParserStrings.MissingEndCurlyBrace));
```

In other words, pass the error text **and** the error id.

Alternatively you could just pass the error id and use reflection of some sort to load the error string, but that would not be good for performance.

rjmholt:
## PR Summary

Resolves #6444. Many error messages in PowerShell currently use a LINQ expression to pass both the name of the error and the message through in the error processing. This PR uses the `nameof` feature to carry the error name/ID and gets rid of LINQ expression reflection, hopefully improving performance, especially in editor scenarios.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/parser/Parser.cs
+++ b/src/System.Management.Automation/engine/parser/Parser.cs
@@ -7284,7 +7284,7 @@ namespace System.Management.Automation.Language
         private void SaveError(IScriptExtent extent, string errorId, string errorMsg, bool incompleteInput, params object[] args)
         {
             Diagnostics.Assert(String.Equals(errorMsg, typeof(ParserStrings).GetProperty(errorId, BindingFlags.Static | BindingFlags.NonPublic)?.GetValue(null) as string, StringComparison.CurrentCulture),
-                "ErrorMsg must correspond the the ErrorId in ParserStrings");
+                String.Format("ErrorMsg \"{0}\" must correspond the the ErrorId \"{1}\" in ParserStrings", errorMsg, errorId));
 
             if (args != null && args.Any())
             {
--- a/src/System.Management.Automation/engine/parser/Parser.cs
+++ b/src/System.Management.Automation/engine/parser/Parser.cs
@@ -752,35 +752,6 @@ namespace System.Management.Automation.Language
             return true;
         }
 
-        // Collect types that keep .resx file strings
-        // for errors thrown with "SaveError"
-        private static readonly Type[] s_resxTypes = new []
-        {
-            typeof(ParserStrings),
-            typeof(ExtendedTypeSystem),
-            typeof(DiscoveryExceptions)
-        };
-
-        /// <summary>
-        /// Given an errorId and an errorMsg, verify that they correspond to one another
-        /// in some .resx file
-        /// </summary>
-        /// <param name="errorId">the ID (or RESX key) of the error</param>
-        /// <param name="errorMsg">the message (or RESX value) of the error</param>
-        /// <returns>true if the ID and message are paired in some .resx file, false otherwise</returns>
-        private static bool ErrorIdCorrespondsToMsg(string errorId, string errorMsg)
-        {
-            foreach (Type resxType in s_resxTypes)
-            {
-                if (errorMsg.Equals(resxType.GetProperty(errorId)?.GetValue(null)))
-                {
-                    return true;
-                }
-            }
-
-            return false;
-        }
-
         #endregion Utilities
 
         #region Statements
@@ -7310,7 +7281,8 @@ namespace System.Management.Automation.Language
 
         private void SaveError(IScriptExtent extent, string errorId, string errorMsg, bool incompleteInput, params object[] args)
         {
-            Diagnostics.Assert(ErrorIdCorrespondsToMsg(errorId, errorMsg), "ErrorMsg must correspond the the ErrorId in ParserStrings");
+            Diagnostics.Assert(errorMsg.Equals(typeof(ParserStrings).GetProperty(errorId, BindingFlags.Static | BindingFlags.NonPublic, null, typeof(string), new Type[0], null)?.GetValue(null, null)),
+                "ErrorMsg must correspond the the ErrorId in ParserStrings");
 
             if (args != null && args.Any())
             {
--- a/src/System.Management.Automation/engine/parser/Parser.cs
+++ b/src/System.Management.Automation/engine/parser/Parser.cs
@@ -752,6 +752,35 @@ namespace System.Management.Automation.Language
             return true;
         }
 
+        // Collect types that keep .resx file strings
+        // for errors thrown with "SaveError"
+        private static readonly Type[] s_resxTypes = new []
+        {
+            typeof(ParserStrings),
+            typeof(ExtendedTypeSystem),
+            typeof(DiscoveryExceptions)
+        };
+
+        /// <summary>
+        /// Given an errorId and an errorMsg, verify that they correspond to one another
+        /// in some .resx file
+        /// </summary>
+        /// <param name="errorId">the ID (or RESX key) of the error</param>
+        /// <param name="errorMsg">the message (or RESX value) of the error</param>
+        /// <returns>true if the ID and message are paired in some .resx file, false otherwise</returns>
+        private static bool ErrorIdCorrespondsToMsg(string errorId, string errorMsg)
+        {
+            foreach (Type resxType in s_resxTypes)
+            {
+                if (errorMsg.Equals(resxType.GetProperty(errorId)?.GetValue(null)))
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
         #endregion Utilities
 
         #region Statements
@@ -7267,7 +7296,7 @@ namespace System.Management.Automation.Language
 
         private void SaveError(IScriptExtent extent, string errorId, string errorMsg, bool incompleteInput, params object[] args)
         {
-            Diagnostics.Assert(!String.IsNullOrEmpty(errorId), "Errors must have IDs");
+            Diagnostics.Assert(ErrorIdCorrespondsToMsg(errorId, errorMsg), "ErrorMsg must correspond the the ErrorId in ParserStrings");
 
             if (args != null && args.Any())
             {
--- a/src/System.Management.Automation/engine/parser/Parser.cs
+++ b/src/System.Management.Automation/engine/parser/Parser.cs
@@ -7283,8 +7283,7 @@ namespace System.Management.Automation.Language
 
         private void SaveError(IScriptExtent extent, string errorId, string errorMsg, bool incompleteInput, params object[] args)
         {
-            Diagnostics.Assert(String.Equals(errorMsg, typeof(ParserStrings).GetProperty(errorId, BindingFlags.Static | BindingFlags.NonPublic)?.GetValue(null) as string, StringComparison.CurrentCulture),
-                String.Format("ErrorMsg \"{0}\" must correspond the the ErrorId \"{1}\" in ParserStrings", errorMsg, errorId));
+            AssertErrorIdCorrespondsToMsgString(errorId, errorMsg);
 
             if (args != null && args.Any())
             {
@@ -7293,6 +7292,40 @@ namespace System.Management.Automation.Language
 
             ParseError errorToSave = new ParseError(extent, errorId, errorMsg, incompleteInput);
             SaveError(errorToSave);
+        }
+
+        /// <summary>
+        /// Assertion to ensure that .resx resource files are where all errors
+        /// saved by the parser are stored. Goes through the known resource types
+        /// and checks that the given error ID and message match some entry in one of them
+        /// </summary>
+        /// <param name="errorId">The error ID string (.resx key)</param>
+        /// <param name="errorMsg">The error message, which may be a template string (.resx value)</param>
+        [System.Diagnostics.Conditional("DEBUG")]
+        [System.Diagnostics.Conditional("ASSERTIONS_TRACE")]
+        private static void AssertErrorIdCorrespondsToMsgString(string errorId, string errorMsg)
+        {
+            Type[] resxTypes = new []
+            {
+                typeof(ParserStrings),
+                typeof(DiscoveryExceptions),
+                typeof(ExtendedTypeSystem),
+                typeof(MshSnapInCmdletResources),
+                typeof(ParameterBinderStrings)
+            };
+
+            bool msgCorrespondsToString = false;
+            foreach (Type resxType in resxTypes)
+            {
+                string resxErrorBody = resxType.GetProperty(errorId, BindingFlags.Static | BindingFlags.NonPublic)?.GetValue(null) as string;
+                if (String.Equals(errorMsg, resxErrorBody))
+                {
+                    msgCorrespondsToString = true;
+                    break;
+                }
+            }
+
+            Diagnostics.Assert(msgCorrespondsToString, String.Format("Parser error ID \"{0}\" must correspond to the error message \"{1}\"", errorId, errorMsg));
         }
 
         private static object[] arrayOfOneArg
--- a/src/System.Management.Automation/engine/parser/Parser.cs
+++ b/src/System.Management.Automation/engine/parser/Parser.cs
@@ -2952,7 +2952,8 @@ namespace System.Management.Automation.Language
                         // This shouldn't happen - the system classes should always be good, but just in case,
                         // we'll catch the exception and report it as an error.
                         ReportError(configurationKeywordToken.Extent,
-                            "ParserError",
+                            nameof(ParserStrings.ParserError),
+                            ParserStrings.ParserError,
                             e.ToString());
                         return null;
                     }
@@ -3140,7 +3141,8 @@ namespace System.Management.Automation.Language
             {
                 // In theory this should never happen so if it does, we'll report the actual exception rather than introducing a new message
                 ReportError(configurationKeywordToken.Extent,
-                    "ParserError",
+                    nameof(ParserStrings.ParserError),
+                    ParserStrings.ParserError,
                     "ConfigurationStatementToken: " + e);
                 return null;
             }
@@ -7281,7 +7283,7 @@ namespace System.Management.Automation.Language
 
         private void SaveError(IScriptExtent extent, string errorId, string errorMsg, bool incompleteInput, params object[] args)
         {
-            Diagnostics.Assert(errorMsg.Equals(typeof(ParserStrings).GetProperty(errorId, BindingFlags.Static | BindingFlags.NonPublic, null, typeof(string), new Type[0], null)?.GetValue(null, null)),
+            Diagnostics.Assert(String.Equals(errorMsg, typeof(ParserStrings).GetProperty(errorId, BindingFlags.Static | BindingFlags.NonPublic)?.GetValue(null) as string, StringComparison.CurrentCulture),
                 "ErrorMsg must correspond the the ErrorId in ParserStrings");
 
             if (args != null && args.Any())
--- a/src/System.Management.Automation/engine/parser/Parser.cs
+++ b/src/System.Management.Automation/engine/parser/Parser.cs
@@ -212,7 +212,9 @@ namespace System.Management.Automation.Language
                 }
                 else
                 {
-                    ReportError(_tokenizer.CurrentExtent(), () => ParserStrings.ScriptTooComplicated);
+                    ReportError(_tokenizer.CurrentExtent(),
+                        nameof(ParserStrings.ScriptTooComplicated),
+                        ParserStrings.ScriptTooComplicated);
                 }
             }
 
@@ -559,7 +561,9 @@ namespace System.Management.Automation.Language
             }
             else if (terminatorToken.Kind != TokenKind.EndOfInput)
             {
-                ReportIncompleteInput(terminatorToken.Extent, () => ParserStrings.MissingStatementTerminator);
+                ReportIncompleteInput(terminatorToken.Extent,
+                    nameof(ParserStrings.MissingStatementTerminator),
+                    ParserStrings.MissingStatementTerminator);
             }
         }
 
@@ -864,7 +868,8 @@ namespace System.Management.Automation.Language
                 UngetToken(rParen);
                 endExtent = Before(rParen);
                 ReportIncompleteInput(After(parameters != null && parameters.Any() ? parameters.Last().Extent : lparen.Extent),
-                    () => ParserStrings.MissingEndParenthesisInFunctionParameterList);
+                    nameof(ParserStrings.MissingEndParenthesisInFunctionParameterList),
+                    ParserStrings.MissingEndParenthesisInFunctionParameterList);
             }
 
             List<AttributeAst> attributes = new List<AttributeAst>();
@@ -882,7 +887,10 @@ namespace System.Management.Automation.Language
                         // ErrorRecovery: nothing to do, this is a semantic error that is caught in the parser
                         // because the ast only allows attributes, no type constraints.
 
-                        ReportError(attr.Extent, () => ParserStrings.TypeNotAllowedBeforeParam, attr.TypeName.FullName);
+                        ReportError(attr.Extent,
+                            nameof(ParserStrings.TypeNotAllowedBeforeParam),
+                            ParserStrings.TypeNotAllowedBeforeParam,
+                            attr.TypeName.FullName);
                     }
                 }
             }
@@ -907,8 +915,10 @@ namespace System.Management.Automation.Language
                     {
                         // ErrorRecovery: ??
 
-                        ReportIncompleteInput(After(commaToken), () => ParserStrings.MissingExpressionAfterToken,
-                                              commaToken.Kind.Text());
+                        ReportIncompleteInput(After(commaToken),
+                            nameof(ParserStrings.MissingExpressionAfterToken),
+                            ParserStrings.MissingExpressionAfterToken,
+                            commaToken.Kind.Text());
                     }
                     break;
                 }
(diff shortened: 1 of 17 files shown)


src/System.Management.Automation/engine/parser/SemanticChecks.cs
```diff
@@ -214,15 +241,19 @@ public override AstVisitAction VisitAttribute(AttributeAst attributeAst)
...
                 }
                 else
                 {
                     names.Add(name);
 
                     if (!namedArg.ExpressionOmitted && !IsValidAttributeArgument(namedArg.Argument, constantValueVisitor))
                     {
-                        _parser.ReportError(namedArg.Argument.Extent, GetNonConstantAttributeArgErrorExpr(constantValueVisitor));
+                        Tuple<string, string> errorInfo = GetNonConstantAttributeArgErrorExpr(constantValueVisitor);
+                        _parser.ReportError(namedArg.Argument.Extent, errorInfo.Item1, errorInfo.Item2);
```
iSazonov: `errorInfo.Item1, errorInfo.Item2` looks bad - maybe use out parameters in `GetNonConstantAttributeArgErrorExpr`?

src/System.Management.Automation/engine/parser/SemanticChecks.cs
```diff
@@ -727,7 +801,10 @@ private void CheckAssignmentTarget(ExpressionAst ast, bool simpleAssignment, Act
                                         var expectedType = SpecialVariables.AutomaticVariableTypes[specialIndex];
                                         if (expectedType != lastConvertType)
                                         {
-                                            _parser.ReportError(ast.Extent, () => ParserStrings.AssignmentStatementToAutomaticNotSupported, varPath.UnqualifiedPath, expectedType);
+                                            _parser.ReportError(ast.Extent,
+                                                nameof(ParserStrings.AssignmentStatementToAutomaticNotSupported),
+                                                ParserStrings.AssignmentStatementToAutomaticNotSupported,
+                                                varPath.UnqualifiedPath, expectedType);
```
iSazonov: Please move `expectedType); ` to new line.

src/System.Management.Automation/engine/parser/SemanticChecks.cs
```diff
@@ -1192,7 +1316,10 @@ public override AstVisitAction VisitDynamicKeywordStatement(DynamicKeywordStatem
                 }
                 catch (Exception e)
                 {
-                    _parser.ReportError(dynamicKeywordStatementAst.Extent, () => ParserStrings.DynamicKeywordSemanticCheckException, dynamicKeywordStatementAst.Keyword.ResourceName, e.ToString());
+                    _parser.ReportError(dynamicKeywordStatementAst.Extent,
+                        nameof(ParserStrings.DynamicKeywordSemanticCheckException),
+                        ParserStrings.DynamicKeywordSemanticCheckException,
+                        dynamicKeywordStatementAst.Keyword.ResourceName, e.ToString());
```
iSazonov: Please move last parameter to new line.

src/System.Management.Automation/engine/parser/SemanticChecks.cs
```diff
@@ -1208,13 +1335,15 @@ public override AstVisitAction VisitDynamicKeywordStatement(DynamicKeywordStatem
                     if (propName == null)
                     {
                         _parser.ReportError(keyValueTuple.Item1.Extent,
-                            () => ParserStrings.ConfigurationInvalidPropertyName,
+                            nameof(ParserStrings.ConfigurationInvalidPropertyName),
+                            ParserStrings.ConfigurationInvalidPropertyName,
                             dynamicKeywordStatementAst.FunctionName.Extent, keyValueTuple.Item1.Extent);
```
iSazonov: Please move last parameter to new line.

src/System.Management.Automation/engine/parser/SemanticChecks.cs
```diff
@@ -1208,13 +1335,15 @@ public override AstVisitAction VisitDynamicKeywordStatement(DynamicKeywordStatem
...
                     }
                     else if (!keyword.Properties.ContainsKey(propName.Value))
                     {
                         _parser.ReportError(propName.Extent,
-                            () => ParserStrings.InvalidInstanceProperty,
+                            nameof(ParserStrings.InvalidInstanceProperty),
+                            ParserStrings.InvalidInstanceProperty,
                             propName.Value,
                             string.Join("', '", keyword.Properties.Keys.OrderBy(key => key, StringComparer.OrdinalIgnoreCase)));
```
iSazonov: Please move last parameter to new line.

src/System.Management.Automation/engine/parser/SymbolResolver.cs
```diff
@@ -393,12 +399,19 @@ public override AstVisitAction VisitAssignmentStatement(AssignmentStatementAst a
                             {
                                 var typeAst = _symbolTable.GetCurrentTypeDefinitionAst();
                                 Diagnostics.Assert(typeAst != null, "Method scopes can exist only inside type definitions.");
-                                _parser.ReportError(variableExpressionAst.Extent, () => ParserStrings.MissingTypeInStaticPropertyAssignment,
-                                    String.Format(CultureInfo.InvariantCulture, "[{0}]::", typeAst.Name), propertyMember.Name);
+                                _parser.ReportError(variableExpressionAst.Extent,
+                                    nameof(ParserStrings.MissingTypeInStaticPropertyAssignment),
+                                    ParserStrings.MissingTypeInStaticPropertyAssignment,
+                                    String.Format(CultureInfo.InvariantCulture, "[{0}]::", typeAst.Name),
```
iSazonov: Please move last parameter to new line.

iSazonov (approved):
LGTM with some minor comments.

rjmholt:
@iSazonov, I've fixed up the parameters I didn't newline. There were a couple that were parameters in a method call that was itself a parameter to `ReportError`, so I've pulled out the whole expression and assigned it to a variable, which hopefully improves clarity without polluting the namespace too much. Also did the `out` variable thing -- much better than Tuples as you say.

src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -6849,27 +7265,10 @@ private void SaveError(ParseError error)
...
-                    var getter = propertyInfo.GetMethod;
-                    if (getter != null && getter.IsStatic && getter.ReturnType == typeof(string))
-                    {
-                        errorMsg = (string)getter.Invoke(null, null);
-                        errorId = propertyInfo.Name;
-                    }
-                }
-            }
-            if (errorMsg == null)
+            if (String.IsNullOrEmpty(errorId))
```
lzybkr: This should not happen anymore - so maybe change to an assertion?

lzybkr (approved):
Sometimes I wouldn't mind limited macros in C# - this is a perfect example where I'd use them.

rjmholt:
@lzybkr I had the same thought about macros

src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -7267,10 +7267,7 @@ private void SaveError(ParseError error)
 
         private void SaveError(IScriptExtent extent, string errorId, string errorMsg, bool incompleteInput, params object[] args)
         {
-            if (String.IsNullOrEmpty(errorId))
-            {
-                errorId = "ParserError";
-            }
+            Diagnostics.Assert(!String.IsNullOrEmpty(errorId), "Errors must have IDs");
```
lzybkr: It might even be worth an assertion that the errorMsg is exactly ParserString.errorId, so something like (not tested, probably doesn't even compile):

```c#
Diagnostics.Assert(typeof(ParserStrings).GetProperty(errorId, null).Value.Equals(errorMsg), "ErrorId must match errorMsg from ParserStrings);
```

lzybkr: I'd consider moving the resources instead of writing a helper function.

Assertions disappear in non-debug builds - the arguments should not be evaluated - but to be honest, I've never tried a non-trivial argument like what I wrote - I'm just assuming the jit removes the argument evaluation.

src/System.Management.Automation/engine/parser/SemanticChecks.cs
```diff
@@ -970,13 +1072,23 @@ public override AstVisitAction VisitHashtable(HashtableAst hashtableAst)
...
+                            errorId = nameof(ParserStrings.DuplicateKeyInHashLiteral);
+                            errorMsg = ParserStrings.DuplicateKeyInHashLiteral;
+                        }
 
-                        _parser.ReportError(entry.Item1.Extent, errorMessageExpression,
-                                            keyStr);
+                        _parser.ReportError(entry.Item1.Extent,
+                            errorId,
+                            errorMsg,
+                            keyStr);
```
daxian-dbw: Better to have them on the same line.

src/System.Management.Automation/engine/parser/Parser.cs
rjmholt: There are also a couple of hand-rolled strings passed into `SaveError`. The ones I saw were "this should never happen" clauses. Perhaps I should just migrate everything to ParserStrings.resx and include a new "ParserError", "{0} is very bad and should never ever happen" entry?

src/System.Management.Automation/engine/parser/SymbolResolver.cs
```diff
@@ -631,10 +658,22 @@ private bool VisitTypeName(TypeName typeName, int genericArgumentCount, bool isA
...
+                                errorId = nameof(ParserStrings.TypeNotFound);
+                                errorMsg = ParserStrings.TypeNotFound;
+                            }
                             _parser.ReportError(typeName.Extent,
-                                isAttribute
-                                    ? (Expression<Func<string>>)(() => ParserStrings.CustomAttributeTypeNotFound)
-                                    : () => ParserStrings.TypeNotFound, typeName.Name);
+                                errorId,
+                                errorMsg,
+                                typeName.Name);
```
daxian-dbw: Maybe put them on the same line?

daxian-dbw (changes requested):

src/System.Management.Automation/engine/parser/SemanticChecks.cs
```diff
@@ -63,11 +63,18 @@ private bool IsValidAttributeArgument(Ast ast, IsConstantValueVisitor visitor)
             return (bool)ast.Accept(visitor);
         }
 
-        private Expression<Func<string>> GetNonConstantAttributeArgErrorExpr(IsConstantValueVisitor visitor)
+        private void GetNonConstantAttributeArgErrorExpr(IsConstantValueVisitor visitor, out string errorId, out string errorMsg)
```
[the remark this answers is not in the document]
rjmholt: I didn't know about that nicer tuple syntax being in C# though

src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -752,6 +752,35 @@ internal static bool TryParseAsConstantHashtable(string input, out Hashtable res
...
+        /// in some .resx file
+        /// </summary>
+        /// <param name="errorId">the ID (or RESX key) of the error</param>
+        /// <param name="errorMsg">the message (or RESX value) of the error</param>
+        /// <returns>true if the ID and message are paired in some .resx file, false otherwise</returns>
+        private static bool ErrorIdCorrespondsToMsg(string errorId, string errorMsg)
+        {
+            foreach (Type resxType in s_resxTypes)
+            {
+                if (errorMsg.Equals(resxType.GetProperty(errorId)?.GetValue(null)))
```
daxian-dbw: > resxType.GetProperty(errorId)

The properties in `resxType` are all internal, so this won't get back anything.

daxian-dbw: > I'd consider moving the resources instead of writing a helper function.

I tend to agree with Jason's comment. As for the two "ParserError" instances, I don't mind have the resource string ("ParserError" => "{0}") in `ParserStrings`.

rjmholt:
The outstanding problem is: there are a couple of freeform errors, and all the others come from at least three resource files. The whole point of all the work here is making sure that the resource files track all the error IDs and messages and ideally we'd make an assertion for that.

So, should we:
- [ ] Move all the errors passed into `SaveError` into `ParserStrings.resx`?
- [ ] Add a generic catchall error into `ParserStrings.resx` to deal with the hardcoded error strings?
- [ ] Use complex assertion to look up all the error messages and make sure they all come from a resource file and have a corresponding key/value?

src/System.Management.Automation/engine/parser/Parser.cs
rjmholt: Ok cool. I'll do that

src/System.Management.Automation/engine/parser/SemanticChecks.cs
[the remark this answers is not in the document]
daxian-dbw: Yeah, you were using the reference tuple previously. I'm proposing using the value tuple, which can have field names.

src/System.Management.Automation/engine/parser/SemanticChecks.cs
```diff
@@ -63,18 +63,16 @@ private bool IsValidAttributeArgument(Ast ast, IsConstantValueVisitor visitor)
             return (bool)ast.Accept(visitor);
         }
 
-        private void GetNonConstantAttributeArgErrorExpr(IsConstantValueVisitor visitor, out string errorId, out string errorMsg)
+        (string id, string msg) GetNonConstantAttributeArgErrorExpr(IsConstantValueVisitor visitor)
```
daxian-dbw: Why removing the `private` modifier?

src/System.Management.Automation/engine/parser/SemanticChecks.cs
```diff
@@ -248,9 +246,7 @@ public override AstVisitAction VisitAttribute(AttributeAst attributeAst)
 
                     if (!namedArg.ExpressionOmitted && !IsValidAttributeArgument(namedArg.Argument, constantValueVisitor))
                     {
-                        string errorId;
-                        string errorMsg;
-                        GetNonConstantAttributeArgErrorExpr(constantValueVisitor, out errorId, out errorMsg);
+                        (string errorId, string errorMsg) = GetNonConstantAttributeArgErrorExpr(constantValueVisitor);
```
daxian-dbw: No need to assign like this, just do `var error = ..` and then `_parser.ReportError(namedArg.Argument.Extent, error.id, error.msg)`

src/System.Management.Automation/engine/parser/SemanticChecks.cs
```diff
@@ -260,9 +256,7 @@ public override AstVisitAction VisitAttribute(AttributeAst attributeAst)
             {
                 if (!IsValidAttributeArgument(posArg, constantValueVisitor))
                 {
-                    string errorId;
-                    string errorMsg;
-                    GetNonConstantAttributeArgErrorExpr(constantValueVisitor, out errorId, out errorMsg);
+                    (string errorId, string errorMsg) = GetNonConstantAttributeArgErrorExpr(constantValueVisitor);
```
daxian-dbw: same here.

src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -2308,7 +2400,9 @@ private StatementAst IfStatementRule(Token ifToken)
                     // ErrorRecovery: assume the next token is a newline or part of something else,
                     // so stop parsing the statement and try parsing something else.
 
-                    ReportIncompleteInput(rParen.Extent, () => ParserStrings.MissingStatementBlock, keyword.Text);
+                    ReportIncompleteInput(rParen.Extent,
+                        nameof(ParserStrings.MissingStatementBlock),
+                        keyword.Text);
```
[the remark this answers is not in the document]
daxian-dbw: Errors like this one could easily happen in other places in a PR with so many changes. I suggest you to go through your changes again and make sure this is the only case.

src/System.Management.Automation/engine/parser/SemanticChecks.cs
```diff
@@ -63,11 +63,16 @@ private bool IsValidAttributeArgument(Ast ast, IsConstantValueVisitor visitor)
             return (bool)ast.Accept(visitor);
         }
 
-        private Expression<Func<string>> GetNonConstantAttributeArgErrorExpr(IsConstantValueVisitor visitor)
+        (string id, string msg) GetNonConstantAttributeArgErrorExpr(IsConstantValueVisitor visitor)
```
daxian-dbw: Why removing the private modifier?

src/System.Management.Automation/engine/parser/SemanticChecks.cs
```diff
@@ -214,15 +235,19 @@ public override AstVisitAction VisitAttribute(AttributeAst attributeAst)
...
+                        name);
                 }
                 else
                 {
                     names.Add(name);
 
                     if (!namedArg.ExpressionOmitted && !IsValidAttributeArgument(namedArg.Argument, constantValueVisitor))
                     {
-                        _parser.ReportError(namedArg.Argument.Extent, GetNonConstantAttributeArgErrorExpr(constantValueVisitor));
+                        (string errorId, string errorMsg) = GetNonConstantAttributeArgErrorExpr(constantValueVisitor);
```
daxian-dbw: No need to assign like this, just do `var error = ..` and then `_parser.ReportError(namedArg.Argument.Extent, error.id, error.msg)`

src/System.Management.Automation/engine/parser/Parser.cs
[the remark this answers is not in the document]
lzybkr: Ah, the assertion I suggested would catch this as well - assuming we ran a test build and test all errors :)

src/System.Management.Automation/engine/parser/SemanticChecks.cs
```diff
@@ -231,7 +256,8 @@ public override AstVisitAction VisitAttribute(AttributeAst attributeAst)
             {
                 if (!IsValidAttributeArgument(posArg, constantValueVisitor))
                 {
-                    _parser.ReportError(posArg.Extent, GetNonConstantAttributeArgErrorExpr(constantValueVisitor));
+                    (string errorId, string errorMsg) = GetNonConstantAttributeArgErrorExpr(constantValueVisitor);
```
daxian-dbw: Same here.

daxian-dbw:
I want to bring up the comment https://github.com/PowerShell/PowerShell/pull/6498#discussion_r177223334 again. 
Errors like this one could easily happen in other places in a PR with so many changes. I suggest you to go through your changes again and make sure this is the only wrong case.

Also, please add [feature] tag to your last commit, to trigger the full tests.

src/System.Management.Automation/engine/parser/Parser.cs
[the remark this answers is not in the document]
rjmholt: Yeah, I suspected this would be a risk. I'm intending to go back through and check all the changes before any merge. But I've included that assertion too to be surer.

src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -6849,29 +7279,10 @@ private void SaveError(ParseError error)
...
-                        errorId = propertyInfo.Name;
-                    }
-                }
-            }
-            if (errorMsg == null)
-            {
-                errorMsg = errorExpr.Compile().Invoke();
-                errorId = "ParserError";
-            }
+            Diagnostics.Assert(errorMsg.Equals(typeof(ParserStrings).GetProperty(errorId, BindingFlags.Static | BindingFlags.NonPublic, null, typeof(string), new Type[0], null)?.GetValue(null, null)),
```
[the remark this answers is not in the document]
rjmholt: Is `StringComparison` preferable? In which case, do I want `StringComparison.OrdinalCurrentCulture` since it's a resource file?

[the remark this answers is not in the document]
daxian-dbw: `System.StringComparer.CurrentCulture` sounds right to me.

rjmholt:
I've now read through all the changes to make sure all the errorIDs have a corresponding errorMsg, and that the errorMsg is the right one.

I'll also have another look to see if `ReportIncompleteInput`, `ReportError` or `SaveError` is used anywhere else that might send through an errorID or errorMsg that doesn't come from ParserStrings.resx.

src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -6849,29 +7281,10 @@ private void SaveError(ParseError error)
...
-                    }
-                }
-            }
-            if (errorMsg == null)
-            {
-                errorMsg = errorExpr.Compile().Invoke();
-                errorId = "ParserError";
-            }
+            Diagnostics.Assert(String.Equals(errorMsg, typeof(ParserStrings).GetProperty(errorId, BindingFlags.Static | BindingFlags.NonPublic)?.GetValue(null) as string, StringComparison.CurrentCulture),
+                "ErrorMsg must correspond the the ErrorId in ParserStrings");
```
daxian-dbw: Please specify the errorid and the errormsg that causes the failure.

daxian-dbw:
Strings like `ParameterAlreadyBound` and `InvalidPSSnapInName` should really stay in `ParameterBinderStrings.resx` and `MshSnapInCmdletResources.resx`.

Talked to @rjmholt offline. We think it's better to keep the resource strings in their existing resource files.
Rob will implement a comparison method with `ConditionalAttribute("Debug")` that searches string id from candidate resource types, and use it for the assertion.

rjmholt:
@lzybkr: After discussing with @daxian-dbw, I think that moving strings out of their resource files and into `ParserStrings.resx` so that we can support the assertion is going to end up with all resource strings in one big file. Just on the basis that I started touching strings in `MshSnapInCmdletResources.resx` and `ParameterBinderStrings.resx` which were rightly owned by those resource files.

Instead, I'm going to write a lookup method and wrap it in a `[System.Diagnostics.Conditional("DEBUG")]` attribute (and possibly an `"ASSERTIONS_TRACE"` one as well?) and use that in a `Diagnostics.Assert()` call. There's concern that we're not building debug builds to catch assertion failures often enough, but I imagine the answer to that problem isn't to stop writing assertions.

Hopefully this is the right way to go.

iSazonov:
Maybe we merge the PR and continue in new PR to keep PRs as simple as possible?

src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -7295,6 +7294,40 @@ private void SaveError(IScriptExtent extent, string errorId, string errorMsg, bo
...
+                typeof(ExtendedTypeSystem),
+                typeof(MshSnapInCmdletResources),
+                typeof(ParameterBinderStrings)
+            };
+
+            bool msgCorrespondsToString = false;
+            foreach (Type resxType in resxTypes)
+            {
+                string resxErrorBody = resxType.GetProperty(errorId, BindingFlags.Static | BindingFlags.NonPublic)?.GetValue(null) as string;
+                if (String.Equals(errorMsg, resxErrorBody))
```
iSazonov: Maybe use `Ordinal` comparision?

rjmholt: Oh -- quite right! We want `StringComparison.CurrentCulture` I think, since they're resource files. Does that sound right?

lzybkr: `String.Equals` would be correct (you want an exact comparison - the string should come from the resource, not user generated) but FxCop does complain when you omit the `StringComparison`, so you should use `StringComparison.Ordinal`.

src/System.Management.Automation/engine/parser/Parser.cs
```diff
@@ -6882,6 +7294,40 @@ private void SaveError(IScriptExtent extent, Expression<Func<string>> errorExpr,
             SaveError(errorToSave);
         }
 
+        /// <summary>
+        /// Assertion to ensure that .resx resource files are where all errors
+        /// saved by the parser are stored. Goes through the known resource types
```
iSazonov: Could you please make the comment more simple?

rjmholt: I'd prefer to document why this function exists, since it does something a bit odd and annotations will mean it only exists in Debug builds. Is there a way you'd like me to simplify it? I can just imagine someone coming across this function in a year or two and having to go through all the resource file stuff to work out what's going on

iSazonov: I means that it is very complex English sentence `... that .. are where ... by ... are stored`.

iSazonov (approved):

src/System.Management.Automation/engine/parser/Parser.cs
rjmholt: Yes, after rereading I agree. Does the new wording work better?

daxian-dbw (approved):
LGTM.

PR #6552, opened 2018-04-03
Engine: Fix several code cleanup issues

daxian-dbw:
## PR Summary

Some random fixes:
1. `BaseDotNetAdapterForAdaptedObjects` derives from `DotNetAdapter`. However, instead of directly calling the base methods, it calls the methods on a static instance of `DotNetAdapter`.
2. Fix a typo in `EventManager.cs`.
3. A clone is created but not used in `InitialSessionState.cs`.
4. The property `InternalBaseDotNetAdapter` points to `InternalAdapterSet.DotNetAdapter`, so use `InternalBaseDotNetAdapter` to make the code more readable.
5. The assignment is unnecessary in `SessionStateProviderAPIs.cs`.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/CoreAdapter.cs
+++ b/src/System.Management.Automation/engine/CoreAdapter.cs
@@ -4252,7 +4252,7 @@ namespace System.Management.Automation
             // are ignored.
             if (typeof(T) == typeof(PSMemberInfo))
             {
-                T returnValue = PSObject.dotNetInstanceAdapter.GetDotNetMethod<T>(obj, memberName);
+                T returnValue = base.GetDotNetMethod<T>(obj, memberName);
                 // We only return a method if there is no property by the same name
                 // to match the behavior we have in GetMembers
                 if (returnValue != null && property == null)
@@ -4263,7 +4263,7 @@ namespace System.Management.Automation
 
             if (IsTypeParameterizedProperty(typeof(T)))
             {
-                PSParameterizedProperty parameterizedProperty = PSObject.dotNetInstanceAdapter.GetDotNetProperty<PSParameterizedProperty>(obj, memberName);
+                PSParameterizedProperty parameterizedProperty = base.GetDotNetProperty<PSParameterizedProperty>(obj, memberName);
                 // We only return a parameterized property if there is no property by the same name
                 // to match the behavior we have in GetMembers
                 if (parameterizedProperty != null && property == null)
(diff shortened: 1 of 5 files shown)


src/System.Management.Automation/engine/CoreAdapter.cs
```diff
@@ -4252,7 +4252,7 @@ protected override T GetMember<T>(object obj, string memberName)
             // are ignored.
             if (typeof(T) == typeof(PSMemberInfo))
             {
-                T returnValue = PSObject.dotNetInstanceAdapter.GetDotNetMethod<T>(obj, memberName);
+                T returnValue = base.GetDotNetMethod<T>(obj, memberName);
```
lzybkr: Do you have an example that demonstrates this is a bug?

It might be a change in behavior, e.g. if the instance was a "property only" adapter, `GetMember` might no longer return any methods whereas it would have before.

PaulHigin: Shouldn't this also be changed at line 4607?  It seems like GetDotNetMethod at least should be a static method, and a static instance is not needed.

PaulHigin (changes requested):

src/System.Management.Automation/engine/CoreAdapter.cs
daxian-dbw: > if the instance was a "property only" adapter

This type is `BaseDotNetAdapterForAdaptedObjects` and it derives from `DotNetAdapter` directly. An instance of it won't be a "property only" adapter. It serves as the fall-back DotNet adapter for custom adapters.
I may miss an aspect of your comment, could you please explain your concern a bit more?

daxian-dbw: And performance shouldn't be a problem because the cache tables are static to the process, so every instance of `DotNetAdapter` can benefit from them.

daxian-dbw: Yes, I think the use of `PSObject.dotNetInstanceAdapter` in `PropertyOnlyAdapter` can be changed to `base` too. `PropertyOnlyAdapter` derives from `DotNetAdapter`.

lzybkr: I see, I thought this might be used by `ThirdPartyAdapter` or other `PropertyOnlyAdapter` implementations, but it's the fallback adapter.

After reviewing `BaseDotNetAdapterForAdaptedObjects` and `PropertyOnlyAdapter` and related methods, I don't understand why it exists at all.

daxian-dbw: I have made the similar change to `PropertyOnlyAdapter`.

daxian-dbw: I have made the similar change to `PropertyOnlyAdapter`.

PaulHigin (approved):
LGTM

PR #6558, opened 2018-04-04
Support user@host:port syntax for SSH Transport

### LINKED ISSUE #6284: Should `New-PSSession` using SSH accept user@hostname?
Currently, `New-PSSession` has a parameter set to use `-Hostname` and `-Username` to indicate SSHTransport is being used (instead of `-ComputerName` and `-Credential`).  SSH users are probably more familiar and comfortable with `ssh user@hostname`, so it seems we should support `New-PSSession -Hostname user@hostname` (unfortunately, it would be a breaking change to default to `-Hostname` instead of `-Computername` to support this as a positional parameter)

cc @PaulHigin @joeyaiello

SteveL-MSFT:
## PR Summary

For users familiar with using SSH typically the syntax for the target is `user@host`.  This PR
enables use of `-HostName` or `-SSHConnection` parameters to embed the username
and/or port as part of the hostname such as:

> new-pssession -h steve@mycomputer:22
> invoke-command -sshconnection @{hostname="steve@mycomputer"}

Removed use of validation method that explicitly disallowed anything more than just the host name.
Added method to create a uri from the hostname and parse out the username and port if specified.

Tests waiting on https://github.com/PowerShell/PowerShell/issues/3904

Fix https://github.com/PowerShell/PowerShell/issues/6284

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - [x] User-facing [https://github.com/PowerShell/PowerShell-Docs/pull/2298/](https://github.com/PowerShell/PowerShell-Docs/pull/2298/)
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/remoting/commands/PSRemotingCmdlet.cs
+++ b/src/System.Management.Automation/engine/remoting/commands/PSRemotingCmdlet.cs
@@ -826,6 +826,30 @@ namespace Microsoft.PowerShell.Commands
         #endregion
 
         /// <summary>
+        /// Parse a hostname used with SSH Transport to get embedded
+        /// username and/or port.
+        /// </summary>
+        /// <param name="hostname">host name to parse</param>
+        /// <param name="host">resolved target host</param>
+        /// <param name="userName">resolved target user name</param>
+        /// <param name="port">resolved target port</param>
+        protected void ParseSshHostName(string hostname, out string host, out string userName, out int port)
+        {
+            var uri = new System.Uri("ssh://" + hostname);
+            host = uri.Host;
+            userName = uri.UserInfo;
+            if (userName == String.Empty)
+            {
+                userName = this.UserName;
+            }
+            port = uri.Port;
+            if (port == -1)
+            {
+                port = this.Port;
+            }
+        }
+
+        /// <summary>
         /// Parse the Connection parameter HashTable array.
         /// </summary>
         /// <returns>Array of SSHConnection objects</returns>
@@ -856,8 +880,16 @@ namespace Microsoft.PowerShell.Commands
                     if (paramName.Equals(ComputerNameParameter, StringComparison.OrdinalIgnoreCase) || paramName.Equals(HostNameAlias, StringComparison.OrdinalIgnoreCase))
                     {
                         var resolvedComputerName = ResolveComputerName(GetSSHConnectionStringParameter(item[paramName]));
-                        ValidateComputerName(new string[] { resolvedComputerName });
-                        connectionInfo.ComputerName = resolvedComputerName;
+                        ParseSshHostName(resolvedComputerName, out string host, out string userName, out int port);
+                        connectionInfo.ComputerName = host;
+                        if (userName != String.Empty)
+                        {
+                            connectionInfo.UserName = userName;
+                        }
+                        if (port != -1)
+                        {
+                            connectionInfo.Port = port;
+                        }
                     }
                     else if (paramName.Equals(UserNameParameter, StringComparison.OrdinalIgnoreCase))
                     {
@@ -1337,11 +1369,11 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         protected void CreateHelpersForSpecifiedSSHComputerNames()
         {
-            ValidateComputerName(ResolvedComputerNames);
-
             foreach (string computerName in ResolvedComputerNames)
             {
-                var sshConnectionInfo = new SSHConnectionInfo(this.UserName, computerName, this.KeyFilePath, this.Port);
+                ParseSshHostName(computerName, out string host, out string userName, out int port);
+
+                var sshConnectionInfo = new SSHConnectionInfo(userName, host, this.KeyFilePath, port);
                 var typeTable = TypeTable.LoadDefaultTypeFiles();
                 var remoteRunspace = RunspaceFactory.CreateRunspace(sshConnectionInfo, this.Host, typeTable) as RemoteRunspace;
                 var pipeline = CreatePipeline(remoteRunspace);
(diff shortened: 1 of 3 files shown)


src/System.Management.Automation/engine/remoting/commands/PSRemotingCmdlet.cs
```diff
@@ -825,6 +825,30 @@ internal static void ValidateSpecifiedAuthentication(PSCredential credential, st
...
+        /// Parse a hostname used with SSH Transport to get embedded
+        /// username and/or port.
+        /// </summary>
+        /// <param name="hostname">host name to parse</param>
+        /// <param name="host">resolved target host</param>
+        /// <param name="userName">resolved target user name</param>
+        /// <param name="port">resolved target port</param>
+        protected void ParseSshHostName(string hostname, out string host, out string userName, out int port)
+        {
+            var uri = new System.Uri("ssh://" + hostname);
```
markekraus: I'd like for this to support absolute URIs `ssh://user@host:822`. I recommend trying to parse Hostname as a Uri, check for a schema, if one exists, use that uri. If one does not, do as you have here.

```csharp
Uri uri;
if (!Uri.TryCreate(hostname, UriKind.Absolute, out uri))
{
    uri = new Uri("ssh://" + hostname);
}
```

SteveL-MSFT: There is an existing `-ConnectionUri` I would love to use to support this.  However, that parameter has expectations that WSMan is used (like support `-Credential` which we currently aren't able to support with SSH due to how SSH works and prompts for credentials).  It may be worthwhile to have a separate PR to update `-ConnectionUri` to support an absolute uri for SSH and perhaps throw an unsupported exception if any WSMan only parameters are used.  I don't think it's necessary for this PR as the intent to have `hostname` syntax align with common `ssh.exe` usage.

iSazonov: So why we fix `hostname` if `ConnectionUri` is more suitable for the scenario?

PaulHigin: We should handle possible a possible UriFormatException here.

src/System.Management.Automation/engine/remoting/commands/PSRemotingCmdlet.cs
```diff
@@ -825,6 +825,30 @@ internal static void ValidateSpecifiedAuthentication(PSCredential credential, st
...
+            host = uri.Host;
+            userName = uri.UserInfo;
+            if (userName == String.Empty)
```
PaulHigin: Nitpick:
```
userName = (uri.UserInfo != string.Empty) ? uri.UserInfo : this.UserName;
port = (uri.Port != -1) ? uri.Port : this.Port;
```

src/System.Management.Automation/engine/remoting/commands/PSRemotingCmdlet.cs
```diff
@@ -856,8 +880,16 @@ internal SSHConnection[] ParseSSHConnectionHashTable()
                     if (paramName.Equals(ComputerNameParameter, StringComparison.OrdinalIgnoreCase) || paramName.Equals(HostNameAlias, StringComparison.OrdinalIgnoreCase))
                     {
                         var resolvedComputerName = ResolveComputerName(GetSSHConnectionStringParameter(item[paramName]));
-                        ValidateComputerName(new string[] { resolvedComputerName });
```
PaulHigin: Do we want to no longer validate the host name?

src/System.Management.Automation/engine/remoting/commands/PSRemotingCmdlet.cs
```diff
@@ -825,6 +825,30 @@ internal static void ValidateSpecifiedAuthentication(PSCredential credential, st
...
 
+        /// <summary>
+        /// Parse a hostname used with SSH Transport to get embedded
+        /// username and/or port.
+        /// </summary>
+        /// <param name="hostname">host name to parse</param>
+        /// <param name="host">resolved target host</param>
+        /// <param name="userName">resolved target user name</param>
+        /// <param name="port">resolved target port</param>
+        protected void ParseSshHostName(string hostname, out string host, out string userName, out int port)
```
PaulHigin: This can be a static method.

PaulHigin (changes requested):
We also need to change Enter-PSSession, since it uses its own parameter set (PushRunspaceCommand.cs).  We should make ParseSshHostName a static internal method, maybe a utility method.

src/System.Management.Automation/engine/remoting/commands/PSRemotingCmdlet.cs
SteveL-MSFT: `-ConnectionUri` requires URI syntax which means including a `scheme` like "ssh://".  This goes against the intent of providing consistent syntax with ssh.exe

src/System.Management.Automation/engine/remoting/commands/PSRemotingCmdlet.cs
SteveL-MSFT: I suppose I can call ValidateComputerName after extracting the hostname

src/System.Management.Automation/engine/remoting/commands/PSRemotingCmdlet.cs
SteveL-MSFT: Need access to `this` so can't be static

src/System.Management.Automation/engine/remoting/commands/PSRemotingCmdlet.cs
iSazonov: I don't understand why we can not do:
```powershell
invoke-command -ConnectionUri "https://steve@mycomputer"
invoke-command -ConnectionUri "ssh://steve@mycomputer"
```
https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml
https://tools.ietf.org/id/draft-salowey-secsh-uri-00.html

SteveL-MSFT: @iSazonov we can certainly support that, but I think that's a separate PR.

iSazonov: I'd prefer to keep `-hostname` simple and enhance `-ConnectionUri` - seems it is better UX.

PaulHigin (approved):

PR #6605, opened 2018-04-09
Make a temporary array static to exclude extra allocations

iSazonov:
## PR Summary

The `map` array is allocated every the method call and never changed. So replace it with `if` statements to exclude extra allocations.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/lang/parserutils.cs
+++ b/src/System.Management.Automation/engine/lang/parserutils.cs
@@ -529,15 +529,6 @@ namespace System.Management.Automation
 
         private static RegexOptions parseRegexOptions(SplitOptions options)
         {
-            int[][] map = {
-                new int[] { (int)SplitOptions.CultureInvariant, (int)RegexOptions.CultureInvariant },
-                new int[] { (int)SplitOptions.IgnorePatternWhitespace, (int)RegexOptions.IgnorePatternWhitespace },
-                new int[] { (int)SplitOptions.Multiline, (int)RegexOptions.Multiline },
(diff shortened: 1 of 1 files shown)


daxian-dbw (approved):

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -550,6 +541,15 @@ private static RegexOptions parseRegexOptions(SplitOptions options)
             return result;
         }
 
+        private static readonly int[][] map = {
```
lzybkr: While not as pretty, I'd expect removing the arrays and using 6 if statements would result in faster code.
With the arrays, there are between 6 and 12 array bounds checks in addition to the 6 tests.

daxian-dbw: How about using `Tuple<SplitOptions, RegexOptions>[]` instead? So we can keep it pretty and also a bit more performant.

iSazonov: It is again an array and a cycle - seems `if` will be more faster.

daxian-dbw: JIT eliminates the range check when looping on an array as IEnumerable, so the range check only happens to `entry[0]` and `entry[1]` in the original code. By replacing with a tuple, we can avoid the range change. But I'm fine changing it to if's too.

iSazonov:
@daxian-dbw I updated the PR summary.

lzybkr (approved):

Issue #6598, opened 2018-04-09
File Explorer context-menu integration fails with directory names with embedded apostrophes (single quotes)


mklement0:
Just as the current _Windows PowerShell_ File Explorer context-menu integration (with Shift-right-click) is broken with paths with embedded `'` chars. , as described [here](https://superuser.com/q/1310258/139307), so is PowerShell Core's.

The following two lines must be revised in accordance with the linked instructions:

https://github.com/PowerShell/PowerShell/blob/77f36498e613037babcc99029ca386e36f3b7b59/assets/Product.wxs#L204

https://github.com/PowerShell/PowerShell/blob/77f36498e613037babcc99029ca386e36f3b7b59/assets/Product.wxs#L212

Environment data
----------------

```powershell
PowerShell Core v6.1.0-preview.1
```

SteveL-MSFT:
@bergmeister do you want to take this?

bergmeister:
@SteveL-MSFT I will try to take a look at this but cannot promise anything at the moment (I will be at psconf.eu next week and after that on holiday for 2 weeks so if I cannot fix/investigate it until psconf, then others will need to take a look at it).

bergmeister:
@mklement0 @SteveL-MSFT  I have tried your suggestion of using `\"` instead of the `'` and various other variations with quotes and it does not work for me. This bug exist also with the native Windows PowerShell integration in Windows. The reason is because `%V` gets replaced with the actual path and I could not quickly find a way of making the command work because escaping a single quote inside a single quote means that it needs to appear twice. I also tried double quotes, no success but double quotes are not good from a security perspective because folder names can contain `$`  and therefore you would open up a hole for script injection...

bergmeister:
@mklement0 I am not sure if `-File` would solve it because how do you pass the literal string to the function then, I imagine you would run into the same problem. To me it seems to be rather a language problem because adding just one single quote inside a single quote make it invalid syntax. Maybe an expert in batch or Windows could help out?
What do the powershell language experts think @lzybkr or @BrucePay ?

mklement0:
> I imagine you would run into the same problem

No, you wouldn't - any argument passed to `-File` from the outside is a _literal_.

Thus, something like the following would be safe :

```cmd
powershell.exe -noexit -File  /path/to/helper-script.ps1 "%V"
```

[...]

PR #6612, opened 2018-04-10
Enable -WorkingDirectory parameter to pwsh

SteveL-MSFT:
## PR Summary

Some issues with PowerShell shortcut and context menu not correctly starting in right working directory.  Solution is to expose this parameter so that the shortcut and context menu can leverage it.

Related https://github.com/PowerShell/PowerShell/issues/6598
Related https://github.com/PowerShell/PowerShell/issues/5705

cc @bergmeister 

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - [X] User-facing https://github.com/PowerShell/PowerShell-Docs/pull/2320
    - **OR**

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
@@ -1752,6 +1752,33 @@ namespace Microsoft.PowerShell
             TelemetryAPI.ReportStartupTelemetry(this);
 #endif
 
+            // If working directory was specified, set it
+            if (s_cpp != null && s_cpp.WorkingDirectory != null)
+            {
+                Pipeline tempPipeline = exec.CreatePipeline();
+                Command c = new Command("Set-Location");
+                c.Parameters.Add("LiteralPath", s_cpp.WorkingDirectory);
+                tempPipeline.Commands.Add(c);
+
+                Exception e1;
+
+                if (IsRunningAsync)
+                {
+                    exec.ExecuteCommandAsyncHelper(tempPipeline, out e1, Executor.ExecutionOptions.AddOutputter);
+                }
+                else
+                {
+                    exec.ExecuteCommandHelper(tempPipeline, out e1, Executor.ExecutionOptions.AddOutputter);
+                }
+
+                if (e1 != null)
+                {
+                    // Remember last exception
+                    _lastRunspaceInitializationException = e1;
+                    ReportException(e1, exec);
+                }
+            }
+
             // If a file was specified as the argument to run, then run it...
             if (s_cpp != null && s_cpp.File != null)
             {
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
@@ -1760,21 +1760,20 @@ namespace Microsoft.PowerShell
                 command.Parameters.Add("LiteralPath", s_cpp.WorkingDirectory);
                 tempPipeline.Commands.Add(command);
 
-                Exception e1;
+                Exception exception;
                 if (IsRunningAsync)
                 {
-                    exec.ExecuteCommandAsyncHelper(tempPipeline, out e1, Executor.ExecutionOptions.AddOutputter);
+                    exec.ExecuteCommandAsyncHelper(tempPipeline, out exception, Executor.ExecutionOptions.AddOutputter);
                 }
                 else
                 {
-                    exec.ExecuteCommandHelper(tempPipeline, out e1, Executor.ExecutionOptions.AddOutputter);
+                    exec.ExecuteCommandHelper(tempPipeline, out exception, Executor.ExecutionOptions.AddOutputter);
                 }
 
-                if (e1 != null)
+                if (exception != null)
                 {
-                    // Remember last exception
-                    _lastRunspaceInitializationException = e1;
-                    ReportException(e1, exec);
+                    _lastRunspaceInitializationException = exception;
+                    ReportException(exception, exec);
                 }
             }
 
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
@@ -1756,12 +1756,11 @@ namespace Microsoft.PowerShell
             if (s_cpp != null && s_cpp.WorkingDirectory != null)
             {
                 Pipeline tempPipeline = exec.CreatePipeline();
-                Command c = new Command("Set-Location");
-                c.Parameters.Add("LiteralPath", s_cpp.WorkingDirectory);
-                tempPipeline.Commands.Add(c);
+                var command = new Command("Set-Location");
+                command.Parameters.Add("LiteralPath", s_cpp.WorkingDirectory);
+                tempPipeline.Commands.Add(command);
 
                 Exception e1;
-
                 if (IsRunningAsync)
                 {
                     exec.ExecuteCommandAsyncHelper(tempPipeline, out e1, Executor.ExecutionOptions.AddOutputter);
--- a/src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
+++ b/src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
@@ -187,7 +187,7 @@
     <value>Cannot process the command because -Configuration requires an argument that is a remote endpoint configuration name. Specify this argument and try again.</value>
   </data>
   <data name="MissingSettingsFileArgument" xml:space="preserve">
-    <value>Cannot process the command because -SettingsFile requires a file path. Supply a path for the -SettingsFile parameter and then try the command again.</value>
+    <value>Cannot process the command because -SettingsFile requires an argument that is a file path.</value>
   </data>
   <data name="InvalidSettingsFileArgument" xml:space="preserve">
     <value>Processing -SettingsFile '{0}' failed: {1}. Specify a valid path for the -SettingsFile parameter.</value>
@@ -202,6 +202,6 @@
     <value>Parameter -WindowStyle is not implemented on this platform.</value>
   </data>
   <data name="MissingWorkingDirectoryArgument" xml:space="preserve">
-    <value>Cannot process the command because -WorkingDirectory requires a directory path. Supply a path for the -WorkingDirectory parameter and then try the command again.</value>
+    <value>Cannot process the command because -WorkingDirectory requires an argument that is a directory path.</value>
   </data>
 </root>
--- a/src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
+++ b/src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
@@ -187,7 +187,7 @@
     <value>Cannot process the command because -Configuration requires an argument that is a remote endpoint configuration name. Specify this argument and try again.</value>
   </data>
   <data name="MissingSettingsFileArgument" xml:space="preserve">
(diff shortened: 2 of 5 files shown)


src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -1752,6 +1752,33 @@ private void DoRunspaceInitialization(bool skipProfiles, string initialCommand,
...
 
+            // If working directory was specified, set it
+            if (s_cpp != null && s_cpp.WorkingDirectory != null)
+            {
+                Pipeline tempPipeline = exec.CreatePipeline();
+                Command c = new Command("Set-Location");
+                c.Parameters.Add("LiteralPath", s_cpp.WorkingDirectory);
+                tempPipeline.Commands.Add(c);
+
+                Exception e1;
```
bergmeister: C#7 should allow you to inline the declaration, when you call the methods below:
````powershell
if (IsRunningAsync)
{
    exec.ExecuteCommandAsyncHelper(tempPipeline, out Exception e1, Executor.ExecutionOptions.AddOutputter);
}
else
{
    exec.ExecuteCommandHelper(tempPipeline, out Exception e1, Executor.ExecutionOptions.AddOutputter);
}
````

File: src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs [hunk 2]
```diff
@@ -1752,6 +1752,33 @@ private void DoRunspaceInitialization(bool skipProfiles, string initialCommand,
             TelemetryAPI.ReportStartupTelemetry(this);
 #endif
 
+            // If working directory was specified, set it
+            if (s_cpp != null && s_cpp.WorkingDirectory != null)
+            {
+                Pipeline tempPipeline = exec.CreatePipeline();
+                Command c = new Command("Set-Location");
```
bergmeister: Personally, I would prefer a better variable name and use var instead.
````c#
var command = new Command("Set-Location");
````

File: src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx [hunk 3]
```diff
@@ -201,4 +201,7 @@
   <data name="WindowStyleArgumentNotImplemented" xml:space="preserve">
     <value>Parameter -WindowStyle is not implemented on this platform.</value>
   </data>
+  <data name="MissingWorkingDirectoryArgument" xml:space="preserve">
+    <value>Cannot process the command because -WorkingDirectory requires a directory path. Supply a path for the -WorkingDirectory parameter and then try the command again.</value>
```
bergmeister: Is the 2nd part of the sentence really necessary? To me it does not add more valuable information than what the first sentence already states.

bergmeister (approved):
Some minor comments but looks good overall. I tested that this makes it usable for the explorer context menu. Using double quotes, this works now well with the explorer context menu (tested with paths containing spaces and/or ending in single quotes, or a backtick or 'backtick and n')

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
SteveL-MSFT: Will update

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
SteveL-MSFT: Will update

src/Microsoft.PowerShell.ConsoleHost/resources/CommandLineParameterParserStrings.resx
SteveL-MSFT: I borrowed it as a template from another message.  I'll make it more brief and update the other one as well.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
SteveL-MSFT: Not able to use this syntax as e1 is used outside the scope of the braces

bergmeister: Ah, I see. Sorry. I thought because it is an if/else expression, the compiler is intelligent enough to know that `e1` will always be declared.

bergmeister (approved):

daxian-dbw:
@andschwa Could you please review this PR?

daxian-dbw:
Mentioned the wrong Andrew :)
@anmenaga Could you please review this PR?

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
```diff
@@ -249,6 +250,11 @@ All parameters are case-insensitive.</value>
 
 -WindowStyle | -w
     Sets the window style to Normal, Minimized, Maximized or Hidden.
+
+-WorkingDirectory | -wd
+    Sets the working directory at the start of PowerShell given a valid PowerShell file path.
```
daxian-dbw: > given a valid PowerShell **file** path.

Maybe remove the word **file**?

SteveL-MSFT: Good catch.

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
```diff
@@ -249,6 +250,11 @@ All parameters are case-insensitive.</value>
 
 -WindowStyle | -w
     Sets the window style to Normal, Minimized, Maximized or Hidden.
+
+-WorkingDirectory | -wd
```
anmenaga: Looks like `-wo` should also be mentioned in the help text.

test/powershell/Host/ConsoleHost.Tests.ps1
```diff
@@ -561,6 +561,43 @@ foo
...
+            $folderName = (New-Guid).ToString();
+            New-Item -Path ~/$folderName -ItemType Directory
+            $ExitCodeBadCommandLineParameter = 64
+        }
+
+        AfterAll {
+            Remove-Item ~/$folderName -Force -ErrorAction SilentlyContinue
+        }
+
+        It "Can set working directory to '<value>'" -TestCases @(
```
anmenaga: It would be good to have a case where $value has a space in directory name.
Also a case for Windows-style path separators in directory path.

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
```diff
@@ -249,6 +250,11 @@ All parameters are case-insensitive.</value>
...
+    Sets the working directory at the start of PowerShell given a valid PowerShell directory path.
```
anmenaga: Probably should mention that this "valid PowerShell directory path" is used as a LiteralPath.

anmenaga (approved):
Non-blocking comments.

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
SteveL-MSFT: The pwsh parameters work the same as PowerShell parameters in that if there is sufficient difference in the name, it will figure it out.  So `-wi` will work for `-windowstyle` and `-wo` will work for `-workingdirectory`, although `-w` is reserved for `-windowstyle`.  So for that reason, I don't think it needs to be added explicitly to the help.

test/powershell/Host/ConsoleHost.Tests.ps1
SteveL-MSFT: Will add those cases

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
SteveL-MSFT: Will add

daxian-dbw (approved):

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
SteveL-MSFT: Should be true for all well written providers

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -1752,6 +1752,32 @@ private void DoRunspaceInitialization(bool skipProfiles, string initialCommand,
...
 
+            // If working directory was specified, set it
+            if (s_cpp != null && s_cpp.WorkingDirectory != null)
+            {
+                Pipeline tempPipeline = exec.CreatePipeline();
+                var command = new Command("Set-Location");
+                command.Parameters.Add("LiteralPath", s_cpp.WorkingDirectory);
+                tempPipeline.Commands.Add(command);
+
+                Exception e1;
```
iSazonov: Could we use more informative name?

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -1752,6 +1752,32 @@ private void DoRunspaceInitialization(bool skipProfiles, string initialCommand,
...
+                    exec.ExecuteCommandAsyncHelper(tempPipeline, out e1, Executor.ExecutionOptions.AddOutputter);
+                }
+                else
+                {
+                    exec.ExecuteCommandHelper(tempPipeline, out e1, Executor.ExecutionOptions.AddOutputter);
+                }
+
+                if (e1 != null)
+                {
+                    // Remember last exception
```
iSazonov: The comment looks not do useful - why we need remeber last exception?

src/Microsoft.PowerShell.ConsoleHost/resources/ManagedEntranceStrings.resx
iSazonov: Make sense add this to the help or online help?

test/powershell/Host/ConsoleHost.Tests.ps1
```diff
@@ -561,6 +561,44 @@ foo
             }
         }
     }
+
+    Context "-WorkingDirectory parameter" {
+        BeforeAll {
+            $folderName = (New-Guid).ToString() + " test";
+            New-Item -Path ~/$folderName -ItemType Directory
```
iSazonov: Maybe use `$pwd` instead of `~` alias? Below too.

iSazonov: I wonder - is this a suitable place for `~` tests?

SteveL-MSFT: The intent is to validate a path understood by the FileSystemProvider which abstracts `~` for Windows and Unix

iSazonov: I don't understand why we place FileSystemProvider tests in ConsoleHost.Tests.ps1.

SteveL-MSFT: The documentation for this parameter is that any file system provider path is accepted, and `~` is handled by the provider which is why I added that test case.

SteveL-MSFT:
@iSazonov are your concerns addressed?  Also note that I have a PR for the Docs repo linked in the PR description.

iSazonov:
@SteveL-MSFT My comment is not blocking - I want to get understanding where is right place for tests for providers.

SteveL-MSFT:
@iSazonov in general, I would agree that provider specific tests should be with the provider tests.  In this case, it's not really testing the provider (directly), but validating that the parameter is using a PS provider.

PR #6625, opened 2018-04-11
Use a cache for dash padding

iSazonov:
## PR Summary

Add a cache for dash padding to decrease allocations in the formatting system.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**
    - [] [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)

### DIFF
--- a/src/System.Management.Automation/utils/StringUtil.cs
+++ b/src/System.Management.Automation/utils/StringUtil.cs
@@ -81,6 +81,24 @@ namespace System.Management.Automation.Internal
 
             return result;
         }
+
+        private const int DashCacheMax = 120;
+        private static readonly string[] DashCache = new string[DashCacheMax];
+        internal static string DashPadding(int count)
+        {
+            if (count >= DashCacheMax)
+                return new string('-', count);
+
+            var result = DashCache[count];
+
+            if (result == null)
+            {
+                Interlocked.CompareExchange(ref DashCache[count], new string('-', count), null);
+                result = DashCache[count];
+            }
+
+            return result;
+        }
     }
 }   // namespace
 
--- a/src/System.Management.Automation/FormatAndOutput/common/TableWriter.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/TableWriter.cs
@@ -180,7 +180,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 // NOTE: we can do this because "-" is a single cell character
                 // on all devices. If changed to some other character, this assumption
                 // would be invalidated
-                breakLine[k] = new string('-', count);
+                breakLine[k] = StringUtil.DashPadding(count);
             }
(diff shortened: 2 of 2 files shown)


iSazonov:
Seem CI Appveyor fail is not related to the PR change.

iSazonov:
Reopen to restart CI Appveyor

src/System.Management.Automation/utils/StringUtil.cs
```diff
@@ -81,6 +81,24 @@ internal static string Padding(int countOfSpaces)
 
             return result;
         }
+
+        private const int DashCacheMax = 120;
+        private static readonly string[] DashCache = new string[DashCacheMax];
```
SteveL-MSFT: Seems like we can leverage the new Span<T> and Slice capabilities of C# 7.2

iSazonov: Good catch! I like the idea.

iSazonov: TableWriter expect strings, pass them through many methods and collect them in StringCollection.
So I can not implement the idea - it requires a complete rewriting of this TableWriter and it's a big risk - seems we haven't full set tests.
I am ready to pull another PR with some optimizations in TableWriter to decrease allocations.

I suggest use the cache until we're ready to rewrite TableWriter using `Memory<char>` or something else.

SteveL-MSFT (approved):

src/System.Management.Automation/utils/StringUtil.cs
```diff
@@ -81,6 +81,24 @@ internal static string Padding(int countOfSpaces)
...
+            if (count >= DashCacheMax)
+                return new string('-', count);
+
+            var result = DashCache[count];
+
+            if (result == null)
+            {
+                Interlocked.CompareExchange(ref DashCache[count], new string('-', count), null);
+                result = DashCache[count];
+            }
```
daxian-dbw: How about change this if block to the following? You can return the new array.
```
if (result == null)
{
    result = new string('-', count);
    Interlocked.CompareExchange(ref DashCache[count], result, null);
}
```

lzybkr: From a performance point of view, it shouldn't matter - the code should be hit just once per process in the normal case.

That said, returning the object from the cache is usually more desirable, and as a code pattern, I think that is reason enough to prefer the original code.

daxian-dbw: > the code should be hit just once per process in the normal case.

This if block may be hit 120 times per process, since the array has 120 slots.
I'm fine with the original code. This is not a blocking comment.

PR #6648, opened 2018-04-13
Reduce allocations in TableWriter

iSazonov:
## PR Summary

Use Span\<T\> and `stackalloc` to reduce allocations in `TableWriter`.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
@@ -1010,7 +1010,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
 
                 int fieldCount = tre.formatPropertyFieldList.Count;
 
-                for (int k = 0; k < values.Length; k++)
+                for (int k = 0; k < headerColumns; k++)
                 {
                     if (k < fieldCount)
                     {
--- a/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
@@ -744,8 +744,8 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
         }
 
         /// <summary>
-        /// In cases like implicit remoting, there is no console so reading the console width results in an exception. 
-        /// Instead of handling exception every time we cache this value to increase performance. 
+        /// In cases like implicit remoting, there is no console so reading the console width results in an exception.
+        /// Instead of handling exception every time we cache this value to increase performance.
         /// </summary>
         static private bool _noConsole = false;
 
@@ -955,8 +955,8 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 }
 
                 // create arrays for widths and alignment
-                int[] columnWidths = new int[columns];
-                int[] alignment = new int[columns];
+                Span<int> columnWidths = stackalloc int[columns];
+                Span<int> alignment = stackalloc int[columns];
                 int k = 0;
 
                 foreach (TableColumnInfo tci in this.CurrentTableHeaderInfo.tableColumnInfoList)
@@ -1006,11 +1006,11 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
 
                 // need to make sure we have matching counts: the header count will have to prevail
                 string[] values = new string[headerColumns];
-                int[] alignment = new int[headerColumns];
+                Span<int> alignment = stackalloc int[headerColumns];
 
                 int fieldCount = tre.formatPropertyFieldList.Count;
 
-                for (int k = 0; k < headerColumns; k++)
+                for (int k = 0; k < values.Length; k++)
                 {
                     if (k < fieldCount)
                     {
@@ -1168,8 +1168,8 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 _buffer = new StringValuesBuffer(itemsPerRow);
 
                 // initialize the writer
-                int[] columnWidths = new int[itemsPerRow];
-                int[] alignment = new int[itemsPerRow];
+                Span<int> columnWidths = stackalloc int[itemsPerRow];
+                Span<int> alignment = stackalloc int[itemsPerRow];
 
                 for (int k = 0; k < itemsPerRow; k++)
                 {
--- a/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
@@ -940,6 +940,8 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             {
                 TableFormattingHint tableHint = this.InnerCommand.RetrieveFormattingHint() as TableFormattingHint;
                 int[] columnWidthsHint = null;
+                // We expect that console width is less then 120.
+                const int columnsThresHold = 120;
 
                 if (tableHint != null)
                 {
@@ -955,10 +957,10 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 }
 
                 // create arrays for widths and alignment
-                Span<int> columnWidths = stackalloc int[columns];
-                Span<int> alignment = stackalloc int[columns];
-                int k = 0;
+                Span<int> columnWidths = columns < columnsThresHold ? stackalloc int[columns] : new int[columns];
+                Span<int> alignment = columns < columnsThresHold ? stackalloc int[columns] : new int[columns];
 
+                int k = 0;
                 foreach (TableColumnInfo tci in this.CurrentTableHeaderInfo.tableColumnInfoList)
                 {
                     columnWidths[k] = (columnWidthsHint != null) ? columnWidthsHint[k] : tci.width;
--- a/src/System.Management.Automation/FormatAndOutput/common/ComplexWriter.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/ComplexWriter.cs
@@ -699,14 +699,9 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
 #endif
         internal static string TruncateAtNewLine(string s)
         {
-            if (s == null)
+            if (string.IsNullOrEmpty(s))
             {
                 return "";
-            }
-
-            if (s == string.Empty)
-            {
-                return s;
             }
 
             int lineBreak = s.IndexOfAny(s_lineBreakChars);
--- a/src/System.Management.Automation/FormatAndOutput/common/ComplexWriter.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/ComplexWriter.cs
@@ -699,13 +699,22 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
 #endif
         internal static string TruncateAtNewLine(string s)
         {
-            if (string.IsNullOrEmpty(s))
+            if (s == null)
+            {
+                return "";
+            }
+
+            if (s == string.Empty)
+            {
                 return s;
+            }
 
             int lineBreak = s.IndexOfAny(s_lineBreakChars);
 
             if (lineBreak < 0)
+            {
                 return s;
+            }
 
             return s.Substring(0, lineBreak) + PSObjectHelper.ellipses;
         }
--- a/src/System.Management.Automation/FormatAndOutput/common/ColumnWidthManager.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/ColumnWidthManager.cs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
+
 namespace Microsoft.PowerShell.Commands.Internal.Format
 {
     /// <summary>
@@ -29,7 +31,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
         /// with column elimination, starting from the right most column
         /// </summary>
         /// <param name="columnWidths">array of column widths to appropriately size</param>
-        internal void CalculateColumnWidths(int[] columnWidths)
+        internal void CalculateColumnWidths(Span<int> columnWidths)
         {
             if (AssignColumnWidths(columnWidths))
             {
@@ -47,7 +49,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
         /// </summary>
         /// <param name="columnWidths">columns to process</param>
         /// <returns>true if there was a fit, false if there is need for trimming</returns>
-        private bool AssignColumnWidths(int[] columnWidths)
+        private bool AssignColumnWidths(Span<int> columnWidths)
         {
             // run a quick check to see if all the columns have a specified width,
             // if so, we are done
@@ -127,7 +129,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
         /// trim columns if the total column width is too much for the screen.
         /// </summary>
         /// <param name="columnWidths">column widths to trim</param>
-        private void TrimToFit(int[] columnWidths)
+        private void TrimToFit(Span<int> columnWidths)
         {
             while (true)
             {
@@ -166,7 +168,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
         /// </summary>
         /// <param name="columnWidths">column widths array</param>
         /// <returns></returns>
-        private int CurrentTableWidth(int[] columnWidths)
+        private int CurrentTableWidth(Span<int> columnWidths)
         {
             int sum = 0;
             int visibleColumns = 0;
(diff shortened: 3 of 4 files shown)


src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
```diff
@@ -1006,11 +1006,11 @@ internal override void ProcessPayload(FormatEntryData fed)
 
                 // need to make sure we have matching counts: the header count will have to prevail
                 string[] values = new string[headerColumns];
-                int[] alignment = new int[headerColumns];
+                Span<int> alignment = stackalloc int[headerColumns];
 
                 int fieldCount = tre.formatPropertyFieldList.Count;
 
-                for (int k = 0; k < headerColumns; k++)
+                for (int k = 0; k < values.Length; k++)
```
SteveL-MSFT: Is there a reason for this change?  Seems like `headerColumns` is more readable

src/System.Management.Automation/FormatAndOutput/common/ComplexWriter.cs
```diff
@@ -699,13 +699,22 @@ internal static string StripNewLines (string s)
 #endif
         internal static string TruncateAtNewLine(string s)
         {
-            if (string.IsNullOrEmpty(s))
+            if (s == null)
```
SteveL-MSFT: Why not:

```c#
if (string.IsNullOrEmpty(s))
{
  return string.Empty;
}
```

SteveL-MSFT (changes requested):

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
iSazonov: In samples I always see `values.Length` - I guess the pattern Roslyn can recognize and optimize to remove bounder checks.

src/System.Management.Automation/FormatAndOutput/common/ComplexWriter.cs
iSazonov: I don't know whether .Net allocate `string.Empty` or `""` everytime, if don't allocate we should use `IsNullOrEmpty`.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
SteveL-MSFT: Ok.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
```diff
@@ -955,8 +955,8 @@ internal override void Initialize()
                 }
 
                 // create arrays for widths and alignment
-                int[] columnWidths = new int[columns];
-                int[] alignment = new int[columns];
+                Span<int> columnWidths = stackalloc int[columns];
+                Span<int> alignment = stackalloc int[columns];
```
daxian-dbw: When it comes to using `stackalloc`, I think there are two things we need to be careful:
1. It's OK to avoid allocating heap memory when we just need small sizes, but we need to be sure `length` is small before using `stackalloc int[length]`
2. memory allocated with `stackalloc` should be used for operations that only happen in the current stack frame. The memory is associated with a given stack frame, so it can be dangerous to pass it around, because that may enable referring to memory that's no longer valid (e,g. it's passed to another method as parameter argument, and then gets returned from that method).

Here, I see `columnWidths` and `alignment` are passed to another method `Initialize`. This is not a safe usage of `stackalloc`. Plus, `Writer.Initialize` takes array arguments, so the `Span<int>` will be converted to array again.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
```diff
@@ -1006,11 +1006,11 @@ internal override void ProcessPayload(FormatEntryData fed)
 
                 // need to make sure we have matching counts: the header count will have to prevail
                 string[] values = new string[headerColumns];
-                int[] alignment = new int[headerColumns];
+                Span<int> alignment = stackalloc int[headerColumns];
```
daxian-dbw: Same here, `alignment` is passed to another method `GenerateRow`, so it's not safe to use `stackalloc` here.
Plus, `GenerateRow` takes array argument, so another conversion will happen.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
daxian-dbw: I would prefer to use `headerColumns` here for better readability.

JIT is now able to detect and eliminate bounds check in the following scenario (see [this article](https://blogs.msdn.microsoft.com/dotnet/2017/10/16/ryujit-just-in-time-compiler-optimization-enhancements/)):
```
    int length = data.Length;
    byte x = 0;

    for (var idx = 0; idx < length; idx++)
    {
        x ^= data[idx];  // <---- bounds check here removed
    }
```
So I guess JIT is possibly able to detect the following too:
```
   string[] values = new string[headerColumns];
   int[] alignment = new int[headerColumns];
   for (int k = 0; k < headerColumns; k++)
   {
      values[k] = ...
      alignment[k] = ...
   }
```

Even JIT cannot do that yet, I'm not too concerned about the bounds check here, for two reasons:
1. We cannot eliminate all bounds check.
2. The bottleneck of formatting code is reflection, so micro-optimization like this should come after the readability.

daxian-dbw (dismissed):

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
iSazonov: I agree about Length. Will review.

I don't understand why stackalloc is unsafe - we use Span<T> - it is just introduced to exclude unsafe code and unify working with heap and stack buffers. It is strong typed and has strong boundary checks. Roslyn and runtime do many checks to achieve this.

daxian-dbw: The `Span` struct is just referencing the memory, in this case, the memory is created on the stack. See [this article](https://msdn.microsoft.com/en-us/magazine/mt814808.aspx), and I quote the words about "Span usage validation":

> Span usage validation. Because spans can refer to data that might be associated with a given stack frame, it can be dangerous to pass spans around in a way that might enable referring to memory that’s no longer valid.

iSazonov: And below:
>Thankfully the C# compiler detects such invalid usage with ref structs and fails the compilation with an error:

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
iSazonov: I should add that Span<T> is _managed_ code.
Before C# 7.2 we have some protections for Span<T> only in runtime.
Since C# 7.2 Roslyn does additional checks at compile time and exclude unapproved usage..

daxian-dbw: I understand the C# compiler can catch an issue like that, but the person who is facing that compilation error might not be the one who used `stackalloc` in the first place.

Imagine I will change the implementation of `GenerateRow` at some later time and my change would actually cause the unsafe reference to happen, and that means I will see the compilation error but have no idea why it fails. After finally figuring out why this could happen, I will need to search all callers of this methods and maybe the callers of those callers and so on to find "Ah, there is a stackalloc here".

Therefore, I think we should have a guideline about when to use `stackalloc`, so things can be less complex.

TravisEz13: I agree that the error might be confusing to another contributor.  

If we have guidance on when `span` and `stackalloc` can be used it should be clear and easy to follow.

src/System.Management.Automation/FormatAndOutput/common/ComplexWriter.cs
lzybkr: This pattern is so common I think it's safe to trust .Net to generate good code.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
iSazonov: I reverted the change.

But I believe the patter is not so safe as `values.Lenght` because we could change to `string[] values = new string[headerColumns - 1];` and catch an error.

src/System.Management.Automation/FormatAndOutput/common/ComplexWriter.cs
iSazonov: Fixed.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
iSazonov: >who is facing that compilation error might not be the one who used stackalloc in the first place.

I still have small experience with Span and again don't understand. The compilation error is not related to stackalloc - it is related to Span. My understanding is that Span is managed code and full safe.
The pattern is widely used in CoreFX: 
```c#
Span<int> arr;
if ( len < threashold )
{
    arr = stackalloc int[len]
}
else
{
    arr = new int[len]
}
```
We can safely use the arr in the method, pass it to sub-methods, get a slice of the arr from sub-methods. We'll get errors only if the arr will not be used properly - in an iterator, assign to a heap entity and so on. So if we see Span in parameter definition we explicitly see the limitations.

iSazonov: [Guidance](https://github.com/dotnet/corefx/blob/e50ddc30deca8fa7434fcc32991354d5a38b5efd/Documentation/api-guidelines/System.Memory.md#guidance)

SteveL-MSFT (approved):

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
daxian-dbw: > The pattern is widely used in CoreFX

@iSazonov Thanks for pointing out it. I quickly searched in corefx code base and did uses like that. It looks I'm being overcautious here :)

`Span<T>` cannot be boxed, so yes, it's more accurate to say the limitation is on `Span<T>`, not directly stemming from `stackalloc`.

The scenario that I can think of where the compilation error actually stems from a `stackalloc` statement would be the pesudo code as follows
```
Span<int> MethodOne(...) {
    Span<int> arr = stackalloc int[len];
    return MethodTwo(arr); // <-- return the span returned from another method
}

Span<int> MethodTwo(Span<int> span, ...) {
    ....
    return span.Slice(...); // <-- return a slice of the passed-in span
}
```
In this example, when `stackalloc` was initially applied, it might work fine because `MethodTwo` didn't directly return a slice of the passed-in span. But if a later change to `MethodTwo` makes it `return span.Slice(...)`, then a compilation error will be thrown, in `MethodOne` I guess (didn't really try myself).

What should be the right use in this example? 
- Not returning a slice of a passed-in span argument?
- Or not directly returning a span result returned from another method when you pass in stack memory as argument? 

We are all new to the new language feature, and this is why I would like some guidance around using `stackalloc` and `Span<>`. I think this discussion is very useful for us all to learn more about this new feature and get a better understanding.

SteveL-MSFT: Related to Span<T>, it looks like it's not currently supported on Alpine Linux so our build script for Alpine is currently broken...

iSazonov: > What should be the right use in this example?

We should be thinking about Span<T> as _managed_ code - it is really managed code. Roslyn (and runtime) strongly track all references on Span and block us if we'll try to return (directly or not directly) Span (slice of the span) to above scope from one where it was defined with stackalloc.
So in your example `return MethodTwo(arr); ` will throw an error in any case.

iSazonov: @SteveL-MSFT They don't support Alpine in [2.0](https://blogs.msdn.microsoft.com/dotnet/2017/08/14/announcing-net-core-2-0/) and support in 2.1 [Prview1](https://blogs.msdn.microsoft.com/dotnet/2018/02/27/announcing-net-core-2-1-preview-1/) and perhaps in [Preview2](https://blogs.msdn.microsoft.com/dotnet/2018/04/11/announcing-net-core-2-1-preview-2/)

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
iSazonov: I see CoreFX use the protection in most cases. [Max threshold](https://github.com/dotnet/corefx/blob/e5bcd9aec536946e8053474475a0994436b4d5f4/src/Common/src/CoreLib/System/IO/Path.Windows.cs#L131) I found is 260. And sometimes they allocates 512 byte buffers.
Also I found [one sample](https://github.com/dotnet/corefx/blob/103639b6ff5aa6ab6097f70732530e411817f09b/src/Common/src/CoreLib/System/Globalization/CultureData.cs#L639) without protection. Seems they trust the fact that Culture name is short.

Can we assume that the number of columns is always small (<256, <512)? If no what is good threshold?

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
daxian-dbw: > So in your example return MethodTwo(arr); will throw an error in any case (Span and Slice).

I think you didn't get my point.

What I try to point out is **the use of `stackalloc` might make future changes raise compilation error at an unexpected location.**

Use the same example, say initially methods `MethodOne` and `MethodTwo` looked like this:
```c#
Span<int> MethodOne(...) {
    int[] arr = new int[len];
    return MethodTwo(arr); // <-- return the span returned from another method
}

Span<int> MethodTwo(Span<int> span, ...) {
    int[] array = new int[length];
    ...
    return array.slice(...); // <-- return a slice of an array
}
```
So when you changed to use `stackalloc` in `MethodOne`, it would compiles fine, and the methods became:
```c#
Span<int> MethodOne(...) {
    Span<int> arr = stackalloc int[len];
    return MethodTwo(arr); // <-- return the span returned from another method
}

Span<int> MethodTwo(Span<int> span, ...) {
    int[] array = new int[length];
    ...
    return array.slice(...); // <-- return a slice of an array
}
```
Then later, another contributor changes `MethodTwo` to return a sclice of the pass-in argument `span` like this:
```c#
Span<int> MethodTwo(Span<int> span, ...) {
    ....
    return span.Slice(...); // <-- return a slice of the passed-in span
}
```
And then the contributor **will see the unexpected compilation error from `MethodOne`.**

This is the situation that I want to avoid with some coding guidelines. So what should the guideline be in this situation? Should it be:
- Do not return a slice of a passed-in span argument in a method? (_attempt to prevent this situation from `MehtodTwo`_)
- Or, do not directly return a span result returned from another method when you pass in stack memory as argument to that method? (_attempt to prevent this situation from `MethodOne`_)

Hope I made myself clear this time :)

iSazonov: I see your point. I believe we could compare this with changing in a base class - we can catch tons compilation errors in derived classes - this is the same situation. Another example is a changing in a helper class - we can get compilation errors in each call point. 
And it seems like your example is too incredible if we try to imagine a real situation - such a change in the code completely changes its contract as if `Array. Sort` would start to do a `clone-then-sort`. By the way, in the latter case we may get an unpredictable run-time error or nothing worse. So Span gives even more reliable code.

powercode: I think this is something that we will have to get used to. It is a new and very welcome addition to the .net runtime, and we as developers need to learn to troubleshoot the issues that arises. 
I say let the compilation error happen, and add answers to StackOverflow on how to resolve and understand them.

iSazonov: If we use a threshold to select `stackalloc` vs `new` we could use one of two patterns
https://github.com/dotnet/corefx/blob/58af9adc8c6c70df4b7fd158c984508d96baca59/src/Common/src/System/Security/Cryptography/ECDiffieHellmanOpenSsl.Derive.cs#L165
https://github.com/dotnet/corefx/blob/8b70122cef4a165def5224001b38f10967619669/src/Common/src/CoreLib/System/Globalization/CompareInfo.Windows.cs#L141

Also I found interesting sample https://github.com/dotnet/corefx/blob/fcfb556ebf99ed551c02d0146c805e792b264cf2/src/System.IO.FileSystem/src/System/IO/Enumeration/FileSystemName.cs#L197 - using `Span<int> temp = stackalloc int[0];`  disable expose outside of the method.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
iSazonov: The edge case is that `psobject` have up to `int.MaxValue` properties and hence columns.

Usually we accept that the width of the console is less than 120 so we can use this value as threshold.

daxian-dbw (approved):
LGTM
Thanks @iSazonov for the helpful discussion!

Issue #6117, opened 2018-02-06
`using namespace` in two dot-sourced .ps1 files causes the namespace named in the other file to be used


alx9r:
This arose from #5942.

Steps to reproduce
------------------

#### UsingTest.psm1

```PowerShell
@(
    'fileA.ps1'
    'fileB.ps1'
) |
    % { . "$(Split-Path -Path $MyInvocation.MyCommand.Path)\$_" }
```

#### fileA.ps1

```PowerShell
using namespace UsingTest.A

Add-Type @'
namespace UsingTest { namespace A {
    public enum MyEnum {
        One = 1,
        Two
    }
}
}
'@

function GetA1 { return [int][MyEnum]::One }
```

#### fileB.ps1

```PowerShell
using namespace UsingTest.B

Add-Type @'
namespace UsingTest { namespace B {
    public enum MyEnum {
        One = 101,
        Two
    }
}
}
'@

function GetB1 { [int][MyEnum]::One }
```

#### Invoke these Commands

```PowerShell
Import-Module UsingTest; GetA1; GetB1
```
Expected behavior
-----------------

```none
1
101
```

Actual behavior
---------------

```none
101
1
```

Environment data
----------------

```powershell
> $PSVersionTable

Name                           Value                                            
----                           -----                                            
PSVersion                      6.0.0                                            
PSEdition                      Core                                             
GitCommitId                    v6.0.0                                           
OS                             Microsoft Windows 6.3.9600                       
Platform                       Win32NT                                          
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                          
PSRemotingProtocolVersion      2.3                                              
SerializationVersion           1.1.0.1                                          
WSManStackVersion              3.0                                              
```

BrucePay:
Hi @alx9r - dot sourcing a file results in that file executing in the current scope so any definitions in the dot sourced file will overwrite existing definitions in the current scope. It's like concatenating the two files then executing the result. Later definitions override earlier definitions.

alx9r:
>It's like concatenating the two files then executing the result. Later definitions override earlier definitions.

@BrucePay I thought that at first, too.  That explanation, however, is inconsistent with `GetB1` returning `1` in the repro.  If it were simply a matter of later definitions overriding earlier definitions, then `GetB1` would return `101`.

Issue #4713, opened 2017-08-31
PowerShell class methods cannot invoke non-exported functions.


alx9r:
Steps to reproduce
------------------

```powershell
New-Module m {
    class c { 
        [object] Invoke () { return Invoke-Something }
    }
    function Invoke-Something { 'Invoke-Something' }
    function New-InstanceOfC { [c]::new() }
    function Invoke-MethodOfC { (New-InstanceOfC).Invoke() }
    Export-ModuleMember *OfC
} | Out-Null

(New-InstanceOfC).Invoke()
Invoke-MethodOfC
```

Expected behavior
-----------------

I expected both calls to `Invoke-Something` to succeed and output
```none
Invoke-Something
Invoke-Something
```

Actual behavior
---------------

Neither call to `Invoke-Something` succeeds.

```none
Invoke-Something : The term 'Invoke-Something' is not recognized as the name of a cmdlet, function, script file, or
operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try
again.
At C:\Users\un1\Desktop\test.ps1:3 char:37
+         [object] Invoke () { return Invoke-Something }
+                                     ~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (Invoke-Something:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException

Invoke-Something : The term 'Invoke-Something' is not recognized as the name of a cmdlet, function, script file, or
operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try
again.
At C:\Users\un1\Desktop\test.ps1:3 char:37
+         [object] Invoke () { return Invoke-Something }
+                                     ~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (Invoke-Something:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

Environment data
----------------

```powershell
> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.0.0-beta
PSEdition                      Core
GitCommitId                    v6.0.0-beta.6
OS                             Microsoft Windows 6.3.9600
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

mklement0:
[...]

PR #6677, opened 2018-04-19
WIP: Make Import-Module -Force unload nested modules and make changing files invalidate the analysis cache

### LINKED ISSUE #2505: Reloading module does not reload submodules.
## Steps to reproduce

``` powershell
#Sub.psm1:
  class Sub {
    [int] $Value
  }

#modrepo.psm1:
  using module .\Sub.psm1

  function Get-Sub { 
     [Sub]@{
         Value = 42
     }
  }

Import-Module modrepo
Get-Sub
```

`Value : 1`

Change the definition of Sub:

``` powershell
#Sub.psm1:
  class Sub {  
    [string] $Name # added field
    [int] $Value
  }

#modrepo.psm1:
  using module .\Sub.psm1

  function Get-Sub{ 
     [Sub]@{
         Name = ‘Staffan’ # added field
         Value = 42
     }
  }

#Then

Remove-Module -force modrepo
Import-Module -force modrepo
Get-Sub 
```
## Expected behavior

output of object of class Sub with 

```
Name : 'Staffan'
Value : 42
```
## Actual behavior

```
Cannot create object of type "Sub". The Name property was not found for the Sub object. The available property is: [Value <System.Int32>]
At C:\Users\<user>\Documents\WindowsPowerShell\Modules\modrepo\modrepo.psm1:9 char:3
+   [Sub] @{
+   ~~~~~~~~
    + FullyQualifiedErrorId : ObjectCreationError
```
## Environment data

``` powershell

```

rjmholt:
## PR Summary

Fixes #2505.

This PR:
- Makes `Import-Module -Force` recursively unload nested modules (like `Remove-Module -Force` does)
- Causes a script module cache refresh when the `LastWriteTime` does not match
- Updates the module caches to use a `ConcurrentDictionary` rather than a `Dictionary` in a `lock` block
- Adds tests for `Import-Module`, which I want to expand on in a later PR to drive up coverage of all the module cmdlets so that we can refactor them safely and improve performance and consistency.

This fixes the scenarios where:
- A class or function is in a submodule and exposed in the top-level module, but updating the submodule definition and calling `Import-Module -Force` gives back the old definition results
- A class is defined in a module, loaded in with `Import-Module` and then the class definition is changed. Subsequent calls to `Import-Module` or `using module` then do not use the updated class definition because the cache is not improperly invalidated.

Note that, as discussed in #2505, making `Import-Module -Force` recursively reload nested modules is a **breaking change**. I can open an RFC for this if desired, but feel that it has been discussed in #2505 and that not reloading nested modules is not necessarily expected by users. However, I know that @BrucePay has indicated that leaving the nested modules is by design.

Also, I know that both @powercode and @alx9r have written tests for this (@powercode's [here](https://github.com/PowerShell/PowerShell/files/544410/Modrepo.Tests.ps1.txt) and @alx9r's [here](https://gist.github.com/alx9r/7d3c220163bf55daca6fcc544c07ee27)). I think I've essentially included functionality like that in @powercode's, but @alx9r's tests were considerably more complicated. I think it would be good to rewrite them to some extent and include them in this PR (the only problem being that they are a bit too self-referential for me to pick apart currently).

## PR Checklist

- [] This PR is ready to merge and is not [Work in Progress](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---work-in-progress).
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**


rjmholt:
I've found a heisenbug in the `"Uses updated class definitions in force-reloaded submodules"` test (second last) in `ModuleCmdlet.Tests.ps1`.

(If you want to reproduce: ` for ($i=0;$i-lt20;$i++) { Start-PSPester -Path .\test\powershell\engine\Module\ModuleCmdlet.Tests.ps1 }`. I'm picking it up about 14% of the time).

rjmholt:
I've caught the heisenbug with a `Wait-Debugger` when `Test-SubClassMain` returns `"FIRST"` instead of `"SECOND"`.

The interesting part is:
```powershell
[DBG]: PS C:\Users\roholt\Documents\Dev\PowerShell>> $mod = Get-Module subClassMod
[DBG]: PS C:\Users\roholt\Documents\Dev\PowerShell>> $mod.NestedModules[0].GetExportedTypeDefinitions().SubObj


Name           : SubObj
Attributes     : {}
BaseTypes      : {}
Members        : {Id, SubObj}
TypeAttributes : Class
IsEnum         : False
IsClass        : True
IsInterface    : False
Extent         : class SubObj
                 {
                     [string]$Id

                     SubObj()
                     {
                         $this.Id = "SECOND"
                     }
                 }
Parent         : class SubObj
                 {
                     [string]$Id

                     SubObj()
                     {
                         $this.Id = "SECOND"
                     }
                 }
```

rjmholt:
Did a bit more digging to look at `s_scriptModuleAnalysisCache` from `ModuleCmdletBase`:
```powershell
[DBG]: PS C:\Users\roholt\Documents\Dev\PowerShell>> $smCache = $fieldInfo.GetValue($null)
[DBG]: PS C:\Users\roholt\Documents\Dev\PowerShell>> $smCache.Count
0
```

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,506 @@
...
+        It "Imports a simple module" {
+            $modItem = New-SimpleModule
+            Import-Module $modItem.PSPath
+            Test-FirstModuleFunction | Should -BeExactly "TESTSTRING"
+        }
+
+        It "Passes the imported module out when -PassThru is used" {
+            $modItem = New-SimpleModule
+            $mod = Import-Module $modItem.PSPath -PassThru
+            $mod.ExportedFunctions.ContainsKey("Test-FirstModuleFunction") | Should -BeExactly $true
```
iSazonov: Please use `-BeTrue`.

Below too. And `-BeFalse`.

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,506 @@
...
+
+            $modItem = New-SimpleModule
+            try
+            {
+                $ps.AddScript("Import-Module $($modItem.PSPath) -$paramName `$null")
+                $ps.Invoke()
+            }
+            catch
+            {
+                $_.FullyQualifiedErrorId | Should -BeExactly "ParameterArgumentValidationError,Microsoft.PowerShell.Commands.ImportModuleCommand"
```
iSazonov: What do the test check? Exception? If so we should use  `{ ... } | Should -Throw -ErrorId " ... "`

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,506 @@
...
+    }
+
+    It "Imports a module by moduleinfo" {
+        try
+        {
+            Test-FirstModuleFunction
+        }
+        catch
+        {
+            $_.Exception | Should -BeOfType [System.Management.Automation.CommandNotFoundException]
```
iSazonov: What do the test check? Exception? If so we should use  `{ ... } | Should -Throw -ErrorId " ... "` Also we could add `-Throu` to get the exception.

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,506 @@
...
+    It "Passes the imported module out as a custom object when -AsCustomObject is used" {
+        $mod = Import-Module $modInfo -AsCustomObject
+        $mod.GetType() | Should -BeExactly "PSCustomObject"
+        $mod.'Test-FirstModuleFunction'() | Should -BeExactly "TESTSTRING"
+    }
+
+    It "Re-imports modules from file with new members when -Force is used" {
+        $module = Import-Module $modInfo -PassThru
+
+        $module.ExportedFunctions.Count | Should -BeExactly 1
```
iSazonov: `BeExactly ` is for string. Here we should use `-Be`.

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,506 @@
...
+
+        Test-MainModuleFunc | Should -BeExactly "SUBMODULESTRING"
+
+        try
+        {
+            Test-SubModuleFunc
+        }
+        catch
+        {
+            $_.Exception | Should -BeOfType [System.Management.Automation.CommandNotFoundException]
```
iSazonov: The same.

alx9r:
>...but @alx9r's tests were considerably more complicated. I think it would be good to rewrite them to some extent and include them in this PR (the only problem being that they are a bit too self-referential for me to pick apart currently).

IIRC, that test code emerged organically when I was prospecting for a repro of unexpectedly stale modules.  I apologize it's not more readable.  

What would be important to me (as a PowerShell user) is that there be regression tests in place for the full matrix of relevant permutations.  I _think_ all of the following are relevant for testing module reloading:

| Thing I _think_ is Relevant | Important Test Cases | Notes |
|----------------------|------------------------|--------|
| class or not | module contains PowerShell class; module does not contain PowerShell class |
| module dot-sources or not | module is a single `.psm1`; `.psm1` dot-sources  `.ps1` files |  (1) |
| nested module import location | `Import-Module` in `.psm1`; `Import-Module` in dot-sourced `.ps1` |
| module reload method | `Import-Module -Force`; `Remove-Module; Import-Module`; `Remove-Module -Force; Import-Module` |
| reloaded module kind | dynamic; script-file backed | (2) |
| file cache invalidation method | touch file; append line to file |
| files checked during cache validation | `.psm1`; `.psd1`; `.ps1` dot-sourced from `.psm1` |
| module arguments | second import of module with different arguments from first |

1. #6117 is an example of unexpected results when dot-sourcing files from a `.psm1` file.
2.  #4713 is an example of unexpected differences between dynamic and file-backed files.

I _think_ the combinatorics for such a matrix would work out to somewhere between single-digit hundreds and single-digit thousands of test cases.  My knowledge of the module reloading implementation is very limited.  I suspect someone with more knowledge of the internals could collapse that down a bit.

Note that [my prospecting tests](https://gist.github.com/alx9r/7d3c220163bf55daca6fcc544c07ee27) do not include many of these permutations.  One example of a missing permutation is that there is no test for cache invalidation of dot-sourced `.ps1` files.  I understand that to be a distinct equivalence partition for cache invalidation when such a `.ps1` contains a class, is edited, and the module that dot-sources it is reloaded.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -5807,17 +5851,26 @@ private void ClearAnalysisCaches()
...
 
-            lock (s_lockObject)
+            s_binaryAnalysisCache.TryGetValue(path, out cacheEntry);
+
+            // Check if the DLL file has changed since last caching
+            DateTime lastFileWriteTime = DateTime.MinValue;
+            try
             {
-                s_binaryAnalysisCache.TryGetValue(path, out tuple);
+                lastFileWriteTime = new FileInfo(path).LastWriteTimeUtc;
```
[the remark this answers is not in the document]
rjmholt: I based the `try`/`catch` thing here on [this](https://github.com/PowerShell/PowerShell/blob/5a326135ec2598cb430d4a1c8cf1533631161361/src/System.Management.Automation/engine/Modules/AnalysisCache.cs#L567) example in `AnalysisCache.cs`.

I don't know if we can trust the path and don't want to present an opportunity to crash PowerShell.

I didn't really want to use exceptions for what is essentially control flow logic, but it seemed more efficient than testing the path and then getting the writetime.

Also, you might note that I've used `LastWriteTimeUtc` unlike in `AnalysisCache` -- which is because I don't want the cache to break because the user changes their clock. (But if my assumption about how that works is wrong, can easily change it too).

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,506 @@
...
+function New-SimpleModule
+{
+    $moduleContent = @'
+function Test-FirstModuleFunction
+{
+    Write-Output "TESTSTRING"
+}
+'@
+
+    New-Item -Force -Path $TestDrive -Name "simple.psm1" -Value $moduleContent
```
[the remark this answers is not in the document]
rjmholt: I'm using the output of the function later.

To be honest, I would be open to advice on how to refactor the whole test file to get more reuse.

rjmholt:
@iSazonov Thanks so much for your feedback

rjmholt:
@alx9r Sorry, didn't mean at all to slight your code. Thank you for that table, it's really helpful. I think we are very much on the same wavelength about identifying equivalence classes for `Import-Module` cases and really pinning its behaviour down with regression tests.

I'd like to build on this PR (or just submit a separate one) soon to really build out the testing for `Import-Module`. Reading through the code it looks like there are a bunch of naive cases:
- `Import-Module` from `.psm1`
- `Import-Module` from `.psd1` (there's a [single method](https://github.com/PowerShell/PowerShell/blob/5a326135ec2598cb430d4a1c8cf1533631161361/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs#L1425) that does this, and its about 2000 lines long)
- `Import-Module` from path (where it tries all the extensions in a particular order)
- `Import-Module` from a `PSModuleInfo` object
- `Import-Module` with a binary module
- `Import-Module` on the other [strange PowerShell extensions](https://github.com/PowerShell/PowerShell/blob/5a326135ec2598cb430d4a1c8cf1533631161361/src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs#L486)
- `Import-Module` [over PSRP](https://github.com/PowerShell/PowerShell/blob/5a326135ec2598cb430d4a1c8cf1533631161361/src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs#L768)
- And a couple of CIM remoting cases

alx9r:
>I think we are very much on the same wavelength about identifying equivalence classes for Import-Module cases and really pinning its behaviour down with regression tests.
>
>I'd like to build on this PR (or just submit a separate one) soon to really build out the testing for Import-Module. 

This is really great news for PowerShell.  Thank you @rjmholt.

rjmholt:
Well my thinking is that there are a number of core functionalities in PowerShell that are heavily relied on, but over the years have become large enough that we don't have a good grasp on their semantics or API contacts. For us to extend them, make them more consistent with each other or make them more performant, we're going to need to refactor them. But without extensive regression tests in place that can't be done safely.

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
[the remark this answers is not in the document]
iSazonov: Sorry, I don't understand your comment.
I meant
```powershell
New-Item -Force -Path $TestDrive -Name "simple.psm1" -Value $moduleContent  | Out-Null 
```
to exclude extra output in Pester log.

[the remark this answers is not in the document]
rjmholt: Won't piping to `Out-Null` mean I can't do:
```powershell
$modItem = New-SimpleModule
Set-Content -Path $modItem.PSPath -Value "... new module content ..."
```
Or something similar?

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,487 @@
...
+        $mod.GetType() | Should -BeExactly "PSCustomObject"
+        $mod.'Test-FirstModuleFunction'() | Should -BeExactly "TESTSTRING"
+    }
+
+    It "Re-imports modules from file with new members when -Force is used" {
+        $module = Import-Module $modInfo -PassThru
+
+        $module.ExportedFunctions.Count | Should -Be 1
+        $module.ExportedFunctions.ContainsKey("Test-FirstModuleFunction") | Should -BeTrue
+        $module.ExportedFunctions.ContainsKey("Test-SecondModuleFunction") | Should -BeFalse
```
iSazonov: Maybe we could do like
```powershell
 $module.ExportedFunctions | Should -Contain "Test-SecondModuleFunction"
```
in error case we get better message. Below too.

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,487 @@
...
+    }
+
+    It "Uses updated class definitions in force-reloaded submodules" {
+        $modName = "subClassMod"
+        $subModName = "SubObj"
+        $modPath = "$TestDrive\$modName"
+
+        Get-Module $modName | Remove-Module -Force
+        Get-Module $subModName | Remove-Module -Force
+        if (Test-Path $modPath) { Remove-Item $modPath -Force -Recurse }
```
iSazonov: Please use common multiline formatting.

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,487 @@
...
+        $subModPath = "$modDir\sub.psm1"
+    }
+
+    BeforeEach {
+        New-Item -Path $modDir -ItemType Directory
+    }
+
+    AfterEach {
+        Get-Module $modName | Remove-Module -Force
+        Remove-Item -Path $modDir -Recurse -Force
```
iSazonov: Can we exclude BeforeEach and AfterEach and generate new module name and path for every test?

rjmholt:
@iSazonov I've refactored my tests a bit after the advice you gave me. Please let me know if there's anything else I can improve.

Also, I've marked two tests as skipped because I thought they should be passing but they look like bugs in `Import-Module`. Worth taking a look.

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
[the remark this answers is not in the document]
iSazonov: Ah, sorry - I skipped that it is a function.

I think we should use `return` to exclude confusing.
```powershell
$module = New-Item -Force -Path $TestDrive -Name "simple.psm1" -Value $moduleContent 
return $module
```

rjmholt:
Closing in favour of #6710.

PR #6678, opened 2018-04-19
Add some optimizations in formatting subsystem

iSazonov:
## PR Summary

Continue #6648
Use safe pattern 'columns <= columnsThresHold ? stackalloc int[columns] : new int[columns];'.
Remove duplicate code (we already have it in `GetConsoleWindowWidth()` with a cache).
Use common 'StackAllocThreshold ' constant.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
@@ -249,8 +249,8 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
         /// </summary>
         private enum PreprocessingState { raw, processed, error }
 
-        const int defaultConsoleWidth = 120;
-        internal const int columnsThresHold = 120;
+        const int DefaultConsoleWidth = 120;
+        internal const int StackAllocThreshold = 120;
 
         /// <summary>
         /// test if an object coming from the pipeline needs to be
@@ -765,7 +765,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             {
                 if (_noConsole)
                 {
-                    return defaultConsoleWidth;
+                    return DefaultConsoleWidth;
                 }
                 try
                 {
@@ -774,7 +774,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 catch
                 {
                     _noConsole = true;
-                    return defaultConsoleWidth;
+                    return DefaultConsoleWidth;
                 }
             }
             return columnNumber;
@@ -957,8 +957,8 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 }
 
                 // create arrays for widths and alignment
-                Span<int> columnWidths = columns <= columnsThresHold ? stackalloc int[columns] : new int[columns];
-                Span<int> alignment = columns <= columnsThresHold ? stackalloc int[columns] : new int[columns];
+                Span<int> columnWidths = columns <= StackAllocThreshold ? stackalloc int[columns] : new int[columns];
+                Span<int> alignment = columns <= StackAllocThreshold ? stackalloc int[columns] : new int[columns];
 
                 int k = 0;
                 foreach (TableColumnInfo tci in this.CurrentTableHeaderInfo.tableColumnInfoList)
@@ -1008,7 +1008,7 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
 
                 // need to make sure we have matching counts: the header count will have to prevail
                 string[] values = new string[headerColumns];
-                Span<int> alignment = headerColumns <= columnsThresHold ? stackalloc int[headerColumns] : new int[headerColumns];
+                Span<int> alignment = headerColumns <= StackAllocThreshold ? stackalloc int[headerColumns] : new int[headerColumns];
 
                 int fieldCount = tre.formatPropertyFieldList.Count;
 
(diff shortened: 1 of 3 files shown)


src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
```diff
@@ -249,6 +249,9 @@ private enum FormattingState
         /// </summary>
         private enum PreprocessingState { raw, processed, error }
 
+        const int defaultConsoleWidth = 120;
```
daxian-dbw: This variable is only used in the method `GetConsoleWindowWidth`. Maybe it should be kept as a local const variable in that method.

I know both const vars here are with the value 120, but the naming suggests different things. `return defaultConsoleWidth`  in `GetConsoleWindowWidth` is more readable than `return StackAllocThreshold` in my opinion :)

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
```diff
@@ -249,6 +249,9 @@ private enum FormattingState
...
+        internal const int columnsThresHold = 120;
```
daxian-dbw: > columns**ThresHold**

`threshold` is a single word, so maybe change it to `columnsThreshold`?

src/System.Management.Automation/FormatAndOutput/common/ListWriter.cs
```diff
@@ -36,6 +37,8 @@ internal class ListWriter
         /// <param name="dc">instance of the DisplayCells helper object</param>
         internal void Initialize(string[] propertyNames, int screenColumnWidth, DisplayCells dc)
         {
+            const int propertiesThresHold = OutCommandInner.columnsThresHold;
```
daxian-dbw: I don't think you need this; just use `OutCommandInner.columnsThresHold` instead. C# compiler will use the int value directly in the IL. See an example here:

C# code
---------

```
void Main()
{
	int a = Console.Read();
	string[] b = a <= Test.columnsThresHold ? new string[2] : new string[3];
}

// Define other methods and classes here

internal class Test
{
	internal const int columnsThresHold = 120;
}
```

IL code
--------
```
IL_0000:  nop         
IL_0001:  call        System.Console.Read
IL_0006:  stloc.0     // a
IL_0007:  ldloc.0     // a
IL_0008:  ldc.i4.s    78  // 0x78 == 120
IL_000A:  ble.s       IL_0014
IL_000C:  ldc.i4.3    
IL_000D:  newarr      System.String
IL_0012:  br.s        IL_001A
IL_0014:  ldc.i4.2    
IL_0015:  newarr      System.String
IL_001A:  stloc.1     // b
IL_001B:  ret         

Test..ctor:
IL_0000:  ldarg.0     
IL_0001:  call        System.Object..ctor
IL_0006:  nop         
IL_0007:  ret         
```

src/System.Management.Automation/FormatAndOutput/common/TableWriter.cs
```diff
@@ -82,18 +82,6 @@ internal void Initialize(int leftMarginIndent, int screenColumns, Span<int> colu
...
-            {
-                try
-                {
-                    screenColumns = System.Console.WindowWidth;
-                }
-                catch
-                {
-                    screenColumns = 120;
-                }
-            }
```
daxian-dbw: Why are we removing this part?

src/System.Management.Automation/FormatAndOutput/common/TableWriter.cs
```diff
@@ -596,5 +584,7 @@ private static string GenerateRowField(string val, int width, int alignment, Dis
         private bool _hideHeader = false;
 
         private int _startColumn = 0;
+
+        const int columnsThresHold = OutCommandInner.columnsThresHold;
```
daxian-dbw: Same here. No need to declare another const int variable.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
daxian-dbw: This threshold is more about using `stackalloc` or managed heap allocation, so maybe we can give it a more general name, like `StackAllocThreshold`.

As for the naming, you probably should use `StackAllocThreshold`, accoding to the naming convention [code-guidline](https://github.com/PowerShell/PowerShell/blob/master/docs/dev-process/coding-guidelines.md#naming-conventions)

> Use `PascalCase` to name constant local variables and fields. The only exception is for interop code where the constant should exactly match the name and value of the code you are calling via interop (e.g. `const int ERROR_SUCCESS = 0`).

daxian-dbw (changes requested):

src/System.Management.Automation/FormatAndOutput/common/ListWriter.cs
iSazonov: Thanks for the in-depth explanation!

Fixed.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
iSazonov: Fixed.

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
iSazonov: I put both constants next to each other specially because if we decide to change the first value, we probably want to change the second value too.

src/System.Management.Automation/FormatAndOutput/common/TableWriter.cs
iSazonov: It duplicates a code in `GetConsoleWindowWidth` so we should remove this (or add cache as in `GetConsoleWindowWidth`).

src/System.Management.Automation/FormatAndOutput/common/TableWriter.cs
iSazonov: Fixed.

src/System.Management.Automation/FormatAndOutput/common/TableWriter.cs
daxian-dbw: OK, got it. #close

src/System.Management.Automation/FormatAndOutput/common/BaseOutputtingCommand.cs
daxian-dbw: I added the `private` modifier to 'DefaultConsoleWidth'.

daxian-dbw (approved):

PR #6691, opened 2018-04-20
Pretty print Export-FormatData XML output

### LINKED ISSUE #6681: Default Export-FormatData to pretty-printing  or add an option
<sup>Related: #6680</sup>

One use case for `Export-FormatData` - even though [`about_Format.ps1xml` doesn't currently mention it](https://github.com/PowerShell/PowerShell-Docs/issues/2341) - is to use existing formatting definitions as a starting point for creating new ones.

To that end, outputting _pretty-printed_ XML would be much more convenient - currently, a single-line XML string is output.

* If changing the existing behavior is not a backward-compatibility concern:

  * Simply output pretty-printed XML by default (it will consume more space, however)

* Otherwise:
  
  * Introduce a switch such as `-Format` (or, to follow the  - inverted - logic of `ConvertTo-Json`, perhaps `-Uncompress`?) that results in pretty-printed output.

Environment data
----------------

Written as of:

```powershell
PowerShell Core v6.0.2
```

iSazonov:
## PR Summary

Fix #6681.

Well format Export-FormatData XML output.

Refactor  Export-FormatData tests, remove duplications.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**


src/System.Management.Automation/FormatAndOutput/common/FormatXMLWriter.cs
```diff
@@ -41,7 +41,12 @@ internal class FormatXmlWriter
 
             try
             {
-                using (XmlWriter xmlWriter = XmlWriter.Create(streamWriter))
+                var settings = new XmlWriterSettings();
```
markekraus: Could we use a static XmlWriterSettings?

markekraus:
I'm curious if this would have any significant impact with regards to storage and memory usage. I havne't used this command much myself, but something tells me we should include a way to revert to the previous settings at the same time as changing the default to pretty print. It's technically not a breaking change, but it may suddenly increase disk space and ram usage if users have this  used in some great capacity. Perhaps `-Compress`?

test/powershell/Modules/Microsoft.PowerShell.Utility/Export-FormatData.Tests.ps1
```diff
@@ -14,33 +21,70 @@ Describe "Export-FormatData DRT Unit Tests" -Tags "CI" {
...
+        $piped[0] | Should -BeExactly "<"
     }
 
-    Context "Check that the output is in the correct format" {
-	    It "Should not return an empty xml file" {
-	        Get-FormatData | Export-FormatData -Path $testOutput
-	        $piped = Get-Content $testOutput
-	        $piped | Should -Not -BeNullOrEmpty
-	    }
+    It "Should well format output xml" {
```
SteveL-MSFT: I think the test description should be `Should pretty print xml output` as `pretty print` is a well known term and `well format` may be confused with `well formed` which is another well known xml term.

test/powershell/Modules/Microsoft.PowerShell.Utility/Export-FormatData.Tests.ps1
```diff
@@ -14,33 +21,70 @@ Describe "Export-FormatData DRT Unit Tests" -Tags "CI" {
...
+              </TableColumnItem>
+            </TableColumnItems>
+          </TableRowEntry>
+        </TableRowEntries>
+      </TableControl>
+    </View>
+  </ViewDefinitions>
+</Configuration>
+"@
+        $testfilename = [guid]::NewGuid().ToString('N')
```
SteveL-MSFT: Would be nice if we could just use `New-TemporaryFile`, maybe we need to add a `-Path` parameter so we can pass in $TestDrive.  I'm fine leaving this as-is.

test/powershell/Modules/Microsoft.PowerShell.Utility/Export-FormatData.Tests.ps1
```diff
@@ -14,33 +21,70 @@ Describe "Export-FormatData DRT Unit Tests" -Tags "CI" {
...
+        $testfile = Join-Path -Path $TestDrive -ChildPath "$testfilename.ps1xml"
+        Set-Content -Path $testfile -Value $xmlContent
+        Update-FormatData -Append $testfile
```
SteveL-MSFT: I believe `Update-FormatData` affects the global sessionstate, so best to do this in a new runspace

SteveL-MSFT (changes requested):

iSazonov:
@markekraus In Issue discussion a conclusion was that it is not a breaking change and the cmdlet is commonly used for more human and not for scripts. If we want address script scenarios we could add `-AsXML` or/and `-AsXMLString` parameters. If you agree we could open new Issue to track this.

test/powershell/Modules/Microsoft.PowerShell.Utility/Export-FormatData.Tests.ps1
iSazonov: Yes, I see. Will fix.

SteveL-MSFT:
Since this was agreed not to be a breaking change, I think we can defer adding a `-Compress` type switch until someone actually needs it.  In the future, I want to replace ps1xml with a DSL anyways.

markekraus:
@iSazonov I am well aware that this is technically not a breaking change. However, this change does have an impact on the size of the output. Making such a change without allowing for the previous format is something that should be challenged. When developers make changes that result in increased memory and disk usage, such a change can have a significant impact on the user's environment even though the change doesn't impact operation of the application. It would be irresponsible for us not to at least consider the `-Compress` option to allow for the previous behavior.

@SteveL-MSFT There is no RFC in on that and no hard written commitment to move forward with a DSL that I have seen. Until such time as there is a hard commitment, we should not be making changes that increase consumption of user's resources without providing a way to revert to the previous behavior. This is especially true when doing so is, as far as I can tell, trivial.

test/powershell/Modules/Microsoft.PowerShell.Utility/Export-FormatData.Tests.ps1
iSazonov: Fixed.

test/powershell/Modules/Microsoft.PowerShell.Utility/Export-FormatData.Tests.ps1
iSazonov: Fixed.

iSazonov:
I discovered duplicate file `formatdata.tests.ps1` and I moved tests in one file.

test/powershell/Modules/Microsoft.PowerShell.Utility/Export-FormatData.Tests.ps1
```diff
@@ -14,33 +71,86 @@ Describe "Export-FormatData DRT Unit Tests" -Tags "CI" {
...
-	    }
+            $runspace.CreatePipeline("Update-FormatData -prependPath $testfile").Invoke()
+            $runspace.CreatePipeline("Get-FormatData -TypeName 'ExportFormatDataTypeName' | Export-FormatData -Path $testOutput").Invoke()
 
-	    It "Should have a valid xml tag at the start of the file" {
-	        Get-FormatData | Export-FormatData -Path $testOutput
-	        $piped = Get-Content $testOutput
-	        $piped[0] | Should -BeExactly "<"
-	    }
+            $content = (Get-Content $testOutput -Raw) -replace "`r`n?|`n", ""
```
mklement0: Thanks for taking this on. Quibble: `"`r?`n"` will do (or `'\r?\n'`)

SteveL-MSFT:
@markekraus in general, I would agree with you.  However, this is specifically about `Export-FormatData` which seems to be only used to produce something that one would modify to create their own format.ps1xml file which is why we agreed that defaulting to pretty print makes sense.

BrucePay:
@markekraus @SteveL-MSFT Being cautious about resource consumption is certainly a worthy goal, but in this case, taking something that is pretty much useless and making it work at a small cost of additional disk space caused by explicit user action seems worthwhile. But resource consumption is definitely something we need to keep track of. Historically it was a non-goal and the user experience trumped any concern about resources (the cmdlet cache being a good example of additional hidden resource consumption.) So thank-you for thanking the time to express your concerns.

markekraus:
All good then! I just wanted to to ensure the this was addressed and reasoned through.

SteveL-MSFT (approved):

PR #6708, opened 2018-04-23
Cleanup uses of `CommandTypes.Workflow` and `WorkflowInfo`

daxian-dbw:
## PR Summary

Clean up code related to the uses of `CommandTypes.Workflow` and `WorkflowInfo` in `System.Management.Automation`. This change mainly affects help provider code.

This PR includes a few minor breaking changes:
- Change the `public` constructors of `WorkflowInfo` to `internal`. We don't support workflow in PSCore, so I think it makes sense to not allow people to create `Workflow` instances.
- Remove the type `System.Management.Automation.DebugSource` since it's only used for workflow debugging.
- Remove the overload of `SetParent` from the abstract class `Debugger` that is only used for workflow debugging.
- Remove the same overload of `SetParent` from the derived class `RemotingJobDebugger`.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**


src/System.Management.Automation/FormatAndOutput/DefaultFormatters/Help_format_ps1xml.cs
```diff
@@ -413,7 +402,6 @@ internal static IEnumerable<ExtendedTypeDefinition> GetFormatData()
                 MamlPossibleValueControl,
                 MamlTrueFalseShortControl,
                 MamlIndentedSyntaxControl,
-                CommonWorkflowParametersControl,
```
daxian-dbw: This entry is with the index 16 in the array `sharedControls`, so for an index that is greater than 16, we need to adjust it by subtracting 1. This is why you will see many index changes below.

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
```diff
@@ -1166,24 +1166,12 @@ private static void SortAndRemoveDuplicates<T>(List<T> input, Func<T, string> ke
...
-                            sessionState.ExportedWorkflows.Add((WorkflowInfo)entry.Value);
-                        }
-                        else
-                        {
-                            message = StringUtil.Format(Modules.ExportingFunction, entry.Key);
-                            sessionState.ExportedFunctions.Add(entry.Value);
-                        }
-
+                        sessionState.ExportedFunctions.Add(entry.Value);
+                        string message = StringUtil.Format(Modules.ExportingFunction, entry.Key);
```
lzybkr: Not for this PR - but the result of this format call is typically thrown away. It should be possible to check if methods like `WriteVerbose` will do anything *before* formatting, or at least add overloads that do the formatting only after deciding the formatted string is needed.

src/System.Management.Automation/engine/debugger/debugger.cs
```diff
@@ -2971,7 +2971,7 @@ private void AddToJobRunningList(PSJobStartEventArgs jobArgs, DebuggerResumeActi
                     startAction,
                     _context.EngineHostInterface.ExternalHost,
                     _context.SessionState.Path.CurrentLocation,
-                    GetFunctionToSourceMap());
+                    new Dictionary<string, DebugSource>());
```
lzybkr: It's not obvious from the diffs, but it looks like this dictionary might not be needed anymore.
If it's needed and clear elsewhere in the code, then fine, but if it's not needed or unclear, a comment would be good.

src/System.Management.Automation/FormatAndOutput/DefaultFormatters/Help_format_ps1xml.cs
lzybkr: You could leave `null` in this entry.
In another PR - you could introduce names for each shared control to make the file more friendly to human editing now that we no longer use a tool to generate it.

src/System.Management.Automation/engine/debugger/debugger.cs
daxian-dbw: Yes, I noticed that too. The implementation of `SetParent` in `RemotingJobDebugger` just throws away all the arguments. However, `SetParent` is a virtual method, theoretically, someone could derive the `Debugger` class and implement `SetParent` that actually use the dictionary ... So, it's a breaking change that I'm not sure if I can make.

@PaulHigin Is it OK to remove the `functionSourceMap` parameter from `public virtual void SetParent`?
If it's OK, I can clean up the `DebugSource` type as a whole.

src/System.Management.Automation/FormatAndOutput/DefaultFormatters/Help_format_ps1xml.cs
daxian-dbw: Will do in a separate PR #6712

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
daxian-dbw: I think it would be the best to add a few overloads to `WriteVerbose`, which take a format string and arguments. So we can defer the construction of `message` until we know for sure we are going to write out verbose messages. What do you think?

lzybkr: That might be easiest to adopt.

Best is to use an `if` before the call, but folks tend to not like that style because it's verbose.

Accepting a delegate is also a clean solution, but that can allocate as well, so might not be an improvement.

src/System.Management.Automation/engine/debugger/debugger.cs
daxian-dbw: @PaulHigin Could you please take a look and weigh in? Thanks!

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
daxian-dbw: Open #6740 to track this investigation.

iSazonov:
Seems if we return an empty array it will be a breaking change - in the case a throw is better then a warning.
If we want remove this in future we could mark this with [obsolete] and write a warning in scripts.

daxian-dbw:
Currently, `ModuleInfo.ExportedWorkflows` returns an empty `Dictionary<string, FunctionInfo>` (_not changed in this PR_). I think the current behavior makes sense. We don't support `XAML` module and don't support the `workflow` keyword in the script, so there is no way to import a module that declares any workflow in it.

As to `Get-Module -list`, for a script module that has `workflow foo {}` defined in it, the module analysis does find `foo` as an exported module member, but consider `foo` to be a function (_this seems wrong to me, module analysis probably shouldn't return any exported commands in case a `workflow` definition is found in AST, because that module won't be able to be loaded_). The `ExportedWorkflows` property of the returned `PSModuleInfo` object is still an empty dictionary.

adityapatwardhan (approved):

src/System.Management.Automation/engine/debugger/debugger.cs
PaulHigin: Yes, the GetFunctionToSourceMap() is only for the workflow debugger.  You can remove the dictionary from SetParent() and also the GetFunctionSource() private method.

PaulHigin (changes requested):

src/System.Management.Automation/engine/debugger/debugger.cs
daxian-dbw: Thanks @PaulHigin for the insight!
I have addressed your comment. Also removed the `DebugSource` type, which is not used anywhere else.

PaulHigin (approved):

PR #6710, opened 2018-04-23
Invalidate script module cache when the underlying module file is changed

### LINKED ISSUE #2505: Reloading module does not reload submodules.

rjmholt:
## PR Summary

Fix the cache invalidation issue from #2505.

Currently, importing a module for the second time will always import from the cache, even if the underlying file has been changed.

This change makes the cache check the last write time on the underlying file first before proceeding, and if it is different, updates the cache and returns the new module file.

I have also changed the analysis cache dictionaries to use `ConcurrentDictionary` rather than needing to manually use a locking object.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell/6710)

### DIFF
--- a/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
@@ -5984,11 +5984,11 @@ namespace Microsoft.PowerShell.Commands
             DateTime lastModuleWriteTimeUtc = GetFileLastWriteTimeUtc(filePath);
 
             // If we have a cache entry and it's up to date, return that
-            if (s_scriptAnalysisCache.TryGetValue(filePath, out ScriptModuleCacheEntry cacheEntry) && lastModuleWriteTimeUtc == cacheEntry.LastFileWriteTimeUtc)
+            if (s_scriptAnalysisCache.TryGetValue(filePath, out ScriptModuleCacheEntry cacheEntry) && lastModuleWriteTimeUtc == cacheEntry.lastFileWriteTimeUtc)
             {
                 // We need to return a cloned version here.
                 // This is because the Get-Module -List -All modifies the returned module info and we do not want the original one changed.
-                return cacheEntry.ModuleInfo.Clone();
+                return cacheEntry.moduleInfo.Clone();
             }
 
             // fake/empty manifestInfo for processing in (!loadElements) mode
@@ -7268,19 +7268,19 @@ namespace Microsoft.PowerShell.Commands
             /// <param name="moduleInfo">the module info describing the module; the analysis result we are caching</param>
             public ScriptModuleCacheEntry(DateTime lastFileWriteTimeUtc, PSModuleInfo moduleInfo)
             {
-                LastFileWriteTimeUtc = lastFileWriteTimeUtc;
-                ModuleInfo = moduleInfo;
+                this.lastFileWriteTimeUtc = lastFileWriteTimeUtc;
+                this.moduleInfo = moduleInfo;
             }
 
             /// <summary>
             /// The last time the file defining the module was written to when we cached the analysis result
             /// </summary>
-            public DateTime LastFileWriteTimeUtc { get; }
+            public readonly DateTime lastFileWriteTimeUtc;
 
             /// <summary>
             /// The actual analysis result describing the script module we analyzed
             /// </summary>
-            public PSModuleInfo ModuleInfo { get; }
+            public readonly PSModuleInfo moduleInfo;
         }
 
     } // end ModuleCmdletBase
--- a/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Diagnostics.CodeAnalysis;
@@ -5782,20 +5783,14 @@ namespace Microsoft.PowerShell.Commands
             return shouldProcessModule;
         }
 
-        private static object s_lockObject = new object();
-
         private void ClearAnalysisCaches()
         {
-            lock (s_lockObject)
-            {
-                s_binaryAnalysisCache.Clear();
-                s_scriptAnalysisCache.Clear();
-            }
+            s_scriptAnalysisCache.Clear();
         }
 
         // Analyzes a binary module implementation for its cmdlets.
-        private static Dictionary<string, Tuple<BinaryAnalysisResult, Version>> s_binaryAnalysisCache =
-            new Dictionary<string, Tuple<BinaryAnalysisResult, Version>>();
+        private static ConcurrentDictionary<string, Tuple<BinaryAnalysisResult, Version>> s_binaryAnalysisCache =
+            new ConcurrentDictionary<string, Tuple<BinaryAnalysisResult, Version>>();
 
 #if CORECLR
         /// <summary>
@@ -5809,10 +5804,7 @@ namespace Microsoft.PowerShell.Commands
         {
             Tuple<BinaryAnalysisResult, Version> tuple;
 
-            lock (s_lockObject)
-            {
-                s_binaryAnalysisCache.TryGetValue(path, out tuple);
-            }
+            s_binaryAnalysisCache.TryGetValue(path, out tuple);
 
             if (tuple != null)
             {
@@ -5830,10 +5822,7 @@ namespace Microsoft.PowerShell.Commands
 
             var resultToReturn = analysisResult ?? new BinaryAnalysisResult();
 
-            lock (s_lockObject)
-            {
-                s_binaryAnalysisCache[path] = Tuple.Create(resultToReturn, assemblyVersion);
-            }
+            s_binaryAnalysisCache[path] = Tuple.Create(resultToReturn, assemblyVersion);
 
             return resultToReturn;
         }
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -5809,10 +5805,7 @@ private static BinaryAnalysisResult GetCmdletsFromBinaryModuleImplementation(str
         {
             Tuple<BinaryAnalysisResult, Version> tuple;
 
-            lock (s_lockObject)
-            {
-                s_binaryAnalysisCache.TryGetValue(path, out tuple);
-            }
+            s_binaryAnalysisCache.TryGetValue(path, out tuple);
 
             if (tuple != null)
```
iSazonov: Seems we could do
```c#
if (s_binaryAnalysisCache.TryGetValue(path, out tuple))
```

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -5845,10 +5835,7 @@ private BinaryAnalysisResult GetCmdletsFromBinaryModuleImplementation(string pat
         {
             Tuple<BinaryAnalysisResult, Version> tuple = null;
 
-            lock (s_lockObject)
-            {
-                s_binaryAnalysisCache.TryGetValue(path, out tuple);
-            }
+            s_binaryAnalysisCache.TryGetValue(path, out tuple);
 
             if (tuple != null)
```
iSazonov: The same.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -5986,29 +5970,42 @@ private static void AnalyzeSnapinDomainHelper()
...
+            ScriptModuleCacheEntry cacheEntry = null;
+            s_scriptAnalysisCache.TryGetValue(filePath, out cacheEntry);
 
-            lock (s_lockObject)
+            // Now check the file to see if its changed since we cached it
+            DateTime lastModuleWriteTimeUtc;
+            TryGetFileLastWriteTimeUtc(filePath, out lastModuleWriteTimeUtc);
+
+            // Return here if the cache is up to date
+            if (cacheEntry?.ModuleInfo != null && lastModuleWriteTimeUtc == cacheEntry.LastFileWriteTimeUtc)
```
iSazonov: About `TryGetFileLastWriteTimeUtc` see my comment below. We could do:
```c#
var lastModuleWriteTimeUtc = GetFileLastWriteTimeUtc(filePath)
if (s_scriptAnalysisCache.TryGetValue(filePath, out ScriptModuleCacheEntry cacheEntry) && cacheEntry.ModuleInfo != null && lastModuleWriteTimeUtc == cacheEntry.LastFileWriteTimeUtc)
{

}
```

And I wonder why here `ModuleInfo` could be null?

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -7252,6 +7243,58 @@ internal static PSSnapInInfo GetEngineSnapIn(ExecutionContext context, string na
...
+            try
+            {
+                lastWriteTimeUtc = File.GetLastAccessTimeUtc(path);
+                return true;
+            }
+            catch (Exception e)
+            {
+                ModuleIntrinsics.Tracer.WriteLine("Exception checking LastWriteTimeUtc on module {0}: {1}", path, e.Message);
+                lastWriteTimeUtc = DateTime.MinValue;
+                return false;
```
iSazonov: We return `lastWriteTimeUtc` in any case. So I suggest rename to `GetFileLastWriteTimeUtc`, don't return a bool and directly return `GetFileLastWriteTimeUtc`
```c#
internal static DateTime GetFileLastWriteTimeUtc(string path) 
```

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,518 @@
...
+    $modDir = New-Item -Path $modBasePath -ItemType Directory -Force
+    $null = New-Item -Path $modPath -Value $Content -Force
+
+    $null = $script:ModuleNames.Add($modName)
+
+    @{
+        "BaseDir" = $modDir;
+        "Name" = $modName;
+        "Path" = $modPath
+    }
```
iSazonov: Could we use `return` in all helper functions?

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,518 @@
...
+            $module = Import-Module $modData.Path -PassThru -Force
+            Test-SecondModuleFunction | Should -BeExactly "SECONDSTRING"
+        }
+
+        It "Re-imports modules from file with new members when -Force is used with -PassThru" {
+            $modData = New-TestModule -Content $simpleModContent
+            $module = Import-Module $modData.Path -PassThru
+
+            $module.ExportedFunctions.Count                                    | Should -Be 1
+            $module.ExportedFunctions.ContainsKey("Test-FirstModuleFunction")  | Should -BeTrue
```
iSazonov: Please use the same pattern from Line 128 with `Should -Contain`. Below too.

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,518 @@
...
+
+            $mod1 = Import-Module $modData.Path -PassThru
+
+            Set-Content -Path $modData.Path -Value $modSrc2
+            $mod2 = Import-Module $modData.Path -PassThru
+
+            $firstTypes = $mod1.GetExportedTypeDefinitions()
+            $secondTypes = $mod2.GetExportedTypeDefinitions()
+
+            $firstTypes.MyObj.Equals($secondTypes.Sub) | Should -BeFalse
```
daxian-dbw: Where does the `Sub` come from?

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,518 @@
...
+            )
+
+            $simpleModContent = @'
+function Test-FirstModuleFunction
+{
+    Write-Output "TESTSTRING"
+}
+'@
+
+            $ps = [powershell]::Create()
```
daxian-dbw: Why creating the powershell instance if it's never used?

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,518 @@
...
+    }
+
+    It "Re-imports modules from file with new members when -Force is used" {
+        $module = Import-Module $modInfo -PassThru
+
+        $module.ExportedFunctions.Count | Should -Be 1
+        $module.ExportedFunctions.Keys  | Should -Contain "Test-FirstModuleFunction"
+        $module.ExportedFunctions.Keys  | Should -Not -Contain "Test-SecondModuleFunction"
+
+        $newModuleContent = (Get-Content $modData.Path | Out-String) + "`nfunction Test-SecondModuleFunction { Write-Output 'SECONDSTRING' }"
```
daxian-dbw: > Get-Content $modData.Path | Out-String

Why not use `$simpleModContent`? Or having `$newModuleContent` declared in `beforeall` like `$simpleModContent`?

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,518 @@
...
+    }
+
+    It "Re-imports modules from file with new members when -Force is used with -PassThru" -Skip {
+        $module = Import-Module $modInfo -PassThru
+
+        $module.ExportedFunctions.Count | Should -Be 1
+        $module.ExportedFunctions.Keys  | Should -Contain "Test-FirstModuleFunction"
+        $module.ExportedFunctions.Keys  | Should -Not -Contain "Test-SecondModuleFunction"
+
+        $newModuleContent = (Get-Content $modData.Path | Out-String) + "`nfunction Test-SecondModuleFunction { Write-Output 'SECONDSTRING' }"
```
daxian-dbw: Same comment here.

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,518 @@
...
+    }
+
+    It "Re-imports modules from file with new members when -Force is used with -AsCustomObject" -Skip {
+        $module = Import-Module $modInfo -PassThru
+
+        $module.ExportedFunctions.Count | Should -Be 1
+        $module.ExportedFunctions.Keys  | Should -Contain "Test-FirstModuleFunction"
+        $module.ExportedFunctions.Keys  | Should -Not -Contain "Test-SecondModuleFunction"
+
+        $newModuleContent = (Get-Content $modData.Path | Out-String) + "`nfunction Test-SecondModuleFunction { Write-Output 'SECONDSTRING' }"
```
daxian-dbw: Ditto

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,518 @@
...
+Describe "Import-Module with nested modules" -Tag "Feature" {
+    BeforeAll {
+        SetupModules
+    }
+
+    AfterAll {
+        TearDownModules
+    }
+
+    It "Resolves submodule classes with 'using module'" {
```
daxian-dbw: I don't think `using module` is the right scenario for testing `import-module -force` doesn't refresh nested modules.
`using module` internally loads the module without `-Force`, so by definition forcing reloading the outer module doesn't refresh the ones pulled in by `using module`.

The right scenario should be nested modules or required modules declared in the module manifest I think.

daxian-dbw (changes requested):

daxian-dbw:
This code should enable a developer scenario, where I have the following script in editor:
```powershell
using module F:\tmp\test.psm1

class Zoo { [Foo]$Param }
```
`F:\tmp\test.psm1` has the following content:
```powershell
class Foo { [string]$Name }
```
Then I add another class to `test.psm1` as follows:
```powershell
class Foo { [string]$Name }
class Bar { [string]$Path }
```
And then I modify `Zoo` in the editor to add a property `[Bar]$bar`:
```powershell
using module F:\tmp\test.psm1

class Zoo { [Foo]$Param; [Bar]$bar }
```
The `[Bar]` should be resolved in this scenario, which means the new content in `test.psm1` should be picked up by the parser.
Be noted, no module is loaded in this scenario.

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
rjmholt: Well my intent here isn't to test `-Force` behaviour, just how `Import-Module` works in any scenarios with nested modules (seeing as how this seems to be a complicated area). The `using module` tests are just there to characterise the behaviour. We could move them to their own describe block though. @alx9r had a really nice breakdown of what tests he thought should be written in the old PR.

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
rjmholt: It does with `-Force` though

rjmholt: Is that the behaviour we are looking for?

test/powershell/engine/Module/ModuleCmdlet.Tests.ps1
```diff
@@ -0,0 +1,518 @@
...
+'@
+
+            $modData = New-TestModule -Content $content
+
+            Import-Module $modData.Path -ArgumentList 'value1'
+            $rootVal = Get-PassedArgsRoot
+            $noRootVal = Get-PassedArgsNoRoot
+            $rootVal | Should -BeExactly $noRootVal
+
+            Import-Module $modData.Path -ArgumentList 'value2'
```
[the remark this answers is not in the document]
rjmholt: You're quite right that it doesn't.

I've duplicated the test to describe the normal and `-Force` cases (with a test to confirm the value of `$rootVal`).

rjmholt:
@daxian-dbw I'm going to write another test in line with the example you give above

rjmholt:
That test failure seems to be Linux-specific, and only occurs about 10% of the time.

powercode:
I don't think this should be forgotten. Is a comment enough to keep it alive?

rjmholt:
@powercode I promise I haven't forgotten about this PR... Just had to cut it short and haven't had a good reason to return to it. I'm about to get back to it - it looks like the failure is consistent on Linux, so I'm hoping to be able to get to the bottom of it,

rjmholt:
Going to close this and continue working on it in an updated branch

PR #6718, opened 2018-04-24
Update PowerShell to build with .NET Core SDK 2.1.300-rc1-008662

iSazonov:
## PR Summary

- TargetFramework changed to netcoreapp21 and removed unnecessary RuntimeFrameworkVersion from PowerShell.Common.props
- Update sdk to 2.1.300-rc1-008662
- Updated TypeGen temporary proj target in `Build.psm1` to work with 2.1
- Rename macOS runtime to just `osx-x64` as the current build logic expects 10.12 and breaks running on 10.13 system.
- Remove assemblyreference to System.Memory as it's part of dotnetcore 2.1
- Update packaging.psm1 to dotnetcore 2.1
- ??? Update search for crossgen executable to find matching version to sdk

### Test updates:
- Update test tools (WebListener and TestExe) to latest asp.net core
- (!) Marked AuthHeader Redirect tests as `Pending` due to change in corefx
- Using [string]::contains("string") instead of [string]::contains(int) due to overloads introduced in 2.1

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**


markekraus:
You probably want to update test/Test.Common.props in a similar fashion as PowerShell.Common.props.

iSazonov:
@SteveL-MSFT @daxian-dbw 
I fixed TypeGet but compilation still completes unexpectedly:
- on CI Appveyor `dotnet xunit` failed
- on CI Travis `crossgen` failed.

I couldn't fix it today.
Fill free to continue.

SteveL-MSFT:
Working with dotnetcore team to figure out crossgen issue

SteveL-MSFT:
A bunch of webcmdlet tests are failing because the User-Agent string has a regression in dotnetcore 2.1-preview2.  Looks like they already [fixed it](https://github.com/dotnet/corefx/pull/28994) though.

SteveL-MSFT:
A bunch of other webcmdlet failures may be due to regressions or change in behavior of HttpClient

SteveL-MSFT:
@iSazonov for any new commit, we should make sure to run [feature]

iSazonov:
Locally I tried use the older HttpClientHandler adding
```c#
AppContext.SetSwitch("System.Net.Http.UseSocketsHttpHandler", false);
```
and other web cmdlet tests failed. It seems better not to use the former HttpClientHandler and investigate current test failures. @markekraus Could you please help?

markekraus:
This may be a bit painful. Since there are changes in kestrel and HttpClient, I suspect we will hit regressions or differences on the web cmdlets and weblistener. I will take a look, but I won't have much bandwidth until this weekend.

And no, we shouldn't set any tests to pending in our project due to regressions in CoreFX. We need those to continue to fail in our project until CoreFX fixes them in theirs. Doing so would potentially introduce regressions in our project should they be overlooked.

iSazonov:
We need to understand where the regression is - in WebListener (Kestrel) or in HttpClient or in both.

markekraus:
@iSazonov yes. that's why I said it may be painful. 😆 The only good way I know of tracking down which one is at fault is to build on the three major platforms, test the web cmdlets against httpbin.org and test the WebListener with curl/browser. The other option is to iterate them separately: increase the pwsh runtime version first, file/fix all regressions, then iterate WebListener and repeat.

iSazonov:
I see that [another test](https://github.com/PowerShell/PowerShell/blob/master/test/powershell/Language/Classes/Scripting.Classes.BasicParsing.Tests.ps1#L577) completes abnormally.
They added new overlaps:
- before 2.1
```
$tableOutput.Contains

OverloadDefinitions
-------------------
bool Contains(string value)
```
- in 2.1
```
$
tableOutput.Contains

OverloadDefinitions
-------------------
bool Contains(string value)
bool Contains(string value, System.StringComparison comparisonType)
bool Contains(char value)
bool Contains(char value, System.StringComparison comparisonType)
```
I'm fixing the test, but it seems that it can cause problems in scripts and not only in this case because other APIs can also be extended.

markekraus:
@iSazonov @SteveL-MSFT 

I looked into the authorization issues. It looks  like there is a change in HttpClient's behavior between 2.0 and 2.1 WRT authorization headers on redirect.

using this code in 6.0.2 and a build from @iSazonov branch:

```powershell
$Handler = [System.Net.Http.HttpClientHandler]::new()
$client = [System.Net.Http.HttpClient]::new($Handler)
$req = [System.Net.Http.HttpRequestMessage]::new()
$req.Headers.Add('Authorization','test')
$req.RequestUri = 'http://127.0.0.1:8083/Redirect?type=MultipleChoices&'
$req.RequestUri = 'https://httpbin.org/redirect-to?url=https://httpbin.org/get'
$res = $client.SendAsync($req).GetAwaiter().GetResult()
$res.Content.ReadAsStringAsync().GetAwaiter().GetResult()
```

in 6.0.2 (using dotnet 2.0):

```json
{
  "args": {},
  "headers": {
    "Authorization": "test",
    "Connection": "close",
    "Host": "httpbin.org"
  },
  "origin": "173.239.232.139",
  "url": "https://httpbin.org/get"
}
```

in the build of this PR (dotnet 2.1):

```json
{
  "args": {},
  "headers": {
    "Connection": "close",
    "Host": "httpbin.org"
  },
  "origin": "173.239.232.139",
  "url": "https://httpbin.org/get"
}
```

So it appears HttpClient no longer sends Authorization headers on redirect.

We can handle this in the web cmdlets with some change in logic. But I'd like to find out if this is a regression or `by design` breaking change in CoreFX. I might not have any free time until this weekend. if one of you could do a quick look or talk with the CoreFX folks that would be great. if not, i will look into deeper this weekend.

iSazonov:
@markekraus Many thanks! Could you please look https://stackoverflow.com/questions/28564961/authorization-header-is-lost-on-redirect ? My understanding is that a http client remove authorization headers by default because of _security_.

markekraus:
@iSazonov Yes. In .NET Framework (which is what that link is about) HttpClient did strip authorization by default. You can see this for yourself in Windows PowerShell 5.1 with the same code after running `Add-Type -AssemblyName 'System.Net.Http'`

In .NET Core 1.0-2.0, however, Authorization was not stripped by default. Code was added to the cmdlets to handle that. By default, we intercept the response if the authorization header is present. The pwsh web cmdlets do that by default and it can be overridden with `-PreserveAuthorizationOnRedirect`. This is a new feature we added to Core and it is something that we will need to continue to support as there were even recent issues opened about APIs requiring authorization headers on redirect. 

If the change in 2.1 is `by design` we need to switch our logic around and instead intercept the redirect when the user wants to preserve the auth header.

markekraus:
@iSazonov @SteveL-MSFT ugh.. the change in the `String.Contains()` method in 2.1 is going to cause some pain in PowerShell.

https://docs.microsoft.com/en-us/dotnet/api/system.string.contains?view=netcore-2.1
vs
https://docs.microsoft.com/en-us/dotnet/api/system.string.contains?view=netcore-2.0

I hope there is something we can point users to in our release notes for 6.1 to changes in 2.1 APIs.

iSazonov:
Clear Authorization Headers on Redirect https://github.com/dotnet/corefx/pull/26864

iSazonov:
TabCompletion test failed on MacOs - fixed.

markekraus:
@iSazonov thanks for finding that. It seems clear this is an intentional change so we will need to modify or logic.

The good news is that the current way was fragile (we fixed 2 issues with it since 6.0.0)  and since it was default it ran more often. Now we can rely on .NET and the fragile code will run less often.

markekraus:
@iSazonov I agree we can mark the authorization tests as pending and fix in another PR BUT only under the condition that we do not release 6.1 until the issue for it is closed.

As for the user-agent one.. since that is a CoreFX regression, we should not merge this PR, IMO. We should wait until the next preview release that has the fix before merging.

markekraus:
created #6728 to track the authorization logic changes needed.

iSazonov:
> BUT only under the condition that we do not release 6.1 until the issue for it is closed.

Yes, we can not release without fixing all open Issues and regressions. Maybe we discover new ones later. So I'd want to merge the PR to get feedback with PowerShell Core 6.1 Preview2.

markekraus:
I'm giving direct feedback: Preview2 is currently broken, don't use it. 😆  We have a confirmed CoreFX regression. We should wait until the next Preview build/release that include the fix @SteveL-MSFT linked.

markekraus:
@iSazonov we should not be marking the User-Agent tests as pending.. that is a CoreFX issue, not a PowerShell issue. That means CoreFX is defective and we should not merge it until they have a preview build that includes the changes.

Also, I have not investigated the environment proxy issue. A change in error means a change in behavior. so no.. we cannot simply change the errorid. We need to investigate it.

iSazonov:
@markekraus  User-Agent string fixed as @SteveL-MSFT mentioned above.

markekraus:
@iSazonov The build we are using in this PR is at least the 3rd one I have seen since the blog announcing Preview2. They should have another build of Preview2 that includes it, right? if not.. what is the point of their preview2 iterations if not fix regressions they find?

iSazonov:
If so then it's even better for us - we get fixes faster.

SteveL-MSFT:
My thought is that we shouldn't move to dotnetcore until PSCore6+2.1 is as good or better than current PSCore6 build.  So I would wait until the fixes are available to pick up in this PR than get into a state where some things work and some things don't and we accidentally miss something on release of 6.1.  I would be fine moving ahead to something newer than 2.1-Preview2 since it appears dotnetcore is moving towards RC and the window of opportunity to get them to fix issues in 2.1 before they declare GA is getting smaller.  For this reason, I prefer not to mark tests as `Pending` specifically because of 2.1.

iSazonov:
>  I would wait until the fixes are available to pick up in this PR 

If we are talking about an User-Agent, we can delete the last commit.
And we could try 2.1.0-rc1-26423-06.

SteveL-MSFT:
Trying 2.1.300-rc1-008662 and also updated WebListener to latest asp.net core

SteveL-MSFT:
On Linux/macOS, only 4 failures for HTTP[S]_Proxy env var tests

SteveL-MSFT:
Results looking good.  @iSazonov I updated the PR description with the changes we've made in this PR, please review

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -480,8 +480,8 @@ Describe "TabCompletion" -Tags CI {
                 $beforeTab = 'filesystem::{0}\Wind' -f $env:SystemDrive
                 $afterTab = 'filesystem::{0}\Windows' -f $env:SystemDrive
             } else {
-                $beforeTab = 'filesystem::/us' -f $env:SystemDrive
-                $afterTab = 'filesystem::/usr' -f $env:SystemDrive
+                $beforeTab = 'filesystem::/sbi' -f $env:SystemDrive
+                $afterTab = 'filesystem::/sbin' -f $env:SystemDrive
```
SteveL-MSFT: `$env:SystemDrive` plays no role here and isn't even defined on non-Windows, so that whole part should be removed.  Seems like maybe we should use the testdrive for the filesystem path as something predictable and can be used on all systems.  The `Windows` folder isn't anything special here with regards to tab completion.

test/powershell/Language/Classes/Scripting.Classes.BasicParsing.Tests.ps1
```diff
@@ -573,9 +573,9 @@ Describe 'Hidden Members Test ' -Tags "CI" {
 
         # Formatting should not include hidden members by default
         $tableOutput = $instance | Format-Table -HideTableHeaders -AutoSize | Out-String
-        It "Table formatting should not have included hidden member hiddenZ - should contain 10" { $tableOutput.Contains(10) | Should -BeTrue}
-        It "Table formatting should not have included hidden member hiddenZ- should contain 12" { $tableOutput.Contains(12) | Should -BeTrue}
-        It "Table formatting should not have included hidden member hiddenZ - should not contain 42" { $tableOutput.Contains(42) | Should -BeFalse}
+        It "Table formatting should not have included hidden member hiddenZ - should contain 10" { $tableOutput.Contains("10") | Should -BeTrue}
```
SteveL-MSFT: Perhaps we can update this test similar to what we have in the Format-Table.tests.ps1 which is instead of these 3 checks, we can have:

```powershell
# hiddenZ should not show up in table
$expectedTable = @"

visibleX visibleY
-------- --------
      10       12



"@

$tableOutput = $instance | Format-Table -AutoSize | Out-String
$tableOutput | Should -Be -Exactly $expectedTable
```

SteveL-MSFT (changes requested):

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
SteveL-MSFT: Updating this test

test/powershell/Language/Classes/Scripting.Classes.BasicParsing.Tests.ps1
SteveL-MSFT: Updating this test

iSazonov:
I updated the PR description.
Should we remove "Update search for crossgen executable to find matching version to sdk"?

test/powershell/Language/Classes/Scripting.Classes.BasicParsing.Tests.ps1
iSazonov: I created  tracking Issue #6738 to consider this in more detail.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
iSazonov: Formatting doesn't work in the test at all and we should remove it. I believe we leave this "as is" now and fix the test later.
Tracking Issue #6739

iSazonov:
Rebased to resolve a merge conflict and test with latest merged PRs.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -519,7 +519,7 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         $result.Error.FullyQualifiedErrorId | Should -Be "System.Threading.Tasks.TaskCanceledException,Microsoft.PowerShell.Commands.InvokeWebRequestCommand"
     }
 
-    It "Validate Invoke-WebRequest error with environment proxy set - '<name>'" -TestCases $testCase {
+    It "Validate Invoke-WebRequest error with environment proxy set - '<name>'" -Pending -TestCases $testCase {
```
markekraus: We should not be marking these tests as pending unless we have determined they are issues that need to be addressed in PowerShell and not a regression in CoreFX. I have not had time to troubleshoot this and wont until this weekend.

markekraus (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: It is temporary - I have local problems and I want load the builds from CIs. 
I just want to investigate why these tests don't work

iSazonov: Remoting test failed on Windows too - I skip them too.

iSazonov:
Need fix [Files.wxs](https://ci.appveyor.com/project/PowerShell/powershell/build/v6.1.0-preview.9137#L5344)
```
{$(env.ProductSourcePath)\System.Runtime.CompilerServices.Unsafe.dll} is no longer in product and should be removed from {C:\projects\powershell\tools\packaging\..\..\assets\Files.wxs}
 5339{$(env.ProductSourcePath)\sos_amd64_amd64_4.6.26212.01.dll} is no longer in product and should be removed from {C:\projects\powershell\tools\packaging\..\..\assets\Files.wxs}
 5340{$(env.ProductSourcePath)\mscordaccore_amd64_amd64_4.6.26212.01.dll} is no longer in product and should be removed from {C:\projects\powershell\tools\packaging\..\..\assets\Files.wxs}
 5341{$(env.ProductSourcePath)\System.ComponentModel.Composition.dll} is no longer in product and should be removed from {C:\projects\powershell\tools\packaging\..\..\assets\Files.wxs}
 5342{$(env.ProductSourcePath)\ref\System.ComponentModel.Composition.dll} is no longer in product and should be removed from {C:\projects\powershell\tools\packaging\..\..\assets\Files.wxs}
 5343new file {$(env.ProductSourcePath)\mscordaccore_amd64_amd64_4.6.26424.05.dll} need to be added to {C:\projects\powershell\tools\packaging\..\..\assets\Files.wxs}
 5344new file {$(env.ProductSourcePath)\sos_amd64_amd64_4.6.26424.05.dll} need to be added to {C:\projects\powershell\tools\packaging\..\..\assets\Files.wxs}
 5345new file {$(env.ProductSourcePath)\System.IO.Compression.Brotli.dll} need to be added to {C:\projects\powershell\tools\packaging\..\..\assets\Files.wxs}
 5346new file {$(env.ProductSourcePath)\System.IO.Pipes.AccessControl.dll} need to be added to {C:\projects\powershell\tools\packaging\..\..\assets\Files.wxs}
 5347new file {$(env.ProductSourcePath)\System.Memory.dll} need to be added to {C:\projects\powershell\tools\packaging\..\..\assets\Files.wxs}
 5348new file {$(env.ProductSourcePath)\ref\System.IO.Compression.Brotli.dll} need to be added to {C:\projects\powershell\tools\packaging\..\..\assets\Files.wxs}
 5349new file {$(env.ProductSourcePath)\ref\System.Memory.dll} need to be added to {C:\projects\powershell\tools\packaging\..\..\assets\Files.wxs}
```

SteveL-MSFT:
Fixed the proxy tests, they were fundamentally wrong.  I validated the behavior against curl manually.  The new tests basically expect the request to be sent to the proxy (which is the weblistener) instead of the external URL (as a real proxy will forward the request) and validate against that.

I already made the changes to tabcomplete and classes tests so included those here.

Also updated weblistener to use 2.1.0-rc1 of aspnetcore as the listener fails to start on my macBook Pro (10.13) w/o it.

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
```diff
@@ -97,7 +97,8 @@ Describe "Remoting loopback tests" -Tags @('CI', 'RequireAdminOnWindows') {
 
         $originalDefaultParameterValues = $PSDefaultParameterValues.Clone()
 
-        if ( ! $IsWindows )
+        #if ( ! $IsWindows )
+        if ( $true )
```
SteveL-MSFT: Why are we making this change?

SteveL-MSFT (changes requested):

markekraus:
@SteveL-MSFT I need to look over those tests in more detail. I recall seeing them and originally thinking they were fundamentally wrong as well, but after investigating I discovered they were testing a code path that presents itself in those specific conditions and with an unreachable proxy.

test/tools/WebListener/WebListener.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.App" Version="2.1.0-*" />
+    <PackageReference Include="Microsoft.AspNetCore.All" Version="2.1.0-rc1-30676" />
```
iSazonov: Maybe we should use "Microsoft.AspNetCore.App"?
https://andrewlock.net/exploring-the-microsoft-aspnetcore-app-shared-framework-in-asp-net-core-2-1-preview-1/

test/tools/WebListener/NuGet.config
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
+  </packageSources>
+</configuration>
```
iSazonov: Should we add this for TestExe?

iSazonov (commented):
@SteveL-MSFT Please remove my temporary commits

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
iSazonov: Please remove my temporary commits.

test/tools/WebListener/NuGet.config
SteveL-MSFT: I only needed to add this as the latest aspnetcore isn't on nuget.org.  No changes to TestExe are needed.

SteveL-MSFT:
@markekraus unreachable proxy should be a different set of tests (which kind of explains why the test seemed odd as it wasn't symmetric between invoke-webrequest and invoke-restmethod)

test/tools/WebListener/WebListener.csproj
SteveL-MSFT: Thanks for the link, will update

build.psm1
```diff
@@ -1928,12 +1930,13 @@ function Start-TypeGen
...
 
     $GetDependenciesTargetPath = "$PSScriptRoot/src/Microsoft.PowerShell.SDK/obj/Microsoft.PowerShell.SDK.csproj.TypeCatalog.targets"
+#            <_RefAssemblyPath Include="%(_ReferencesFromRAR)%3B" Condition=" '%(_ReferencesFromRAR.Type)' == 'assembly' And '%(_ReferencesFromRAR.PackageName)' != 'Microsoft.Management.Infrastructure' " />
     $GetDependenciesTargetValue = @'
 <Project>
     <Target Name="_GetDependencies"
             DependsOnTargets="ResolveAssemblyReferencesDesignTime">
         <ItemGroup>
-            <_RefAssemblyPath Include="%(_ReferencesFromRAR.ResolvedPath)%3B" Condition=" '%(_ReferencesFromRAR.Type)' == 'assembly' And '%(_ReferencesFromRAR.PackageName)' != 'Microsoft.Management.Infrastructure' " />
+            <_RefAssemblyPath Include="%(_ReferencesFromRAR.HintPath)%3B"  Condition=" '%(_ReferencesFromRAR.NuGetPackageId)' != 'Microsoft.Management.Infrastructure' "/>
```
SteveL-MSFT: @iSazonov for my own education, how did you figure out this change?

iSazonov: 1. Search "_ReferencesFromRAR" in "C:\Program Files\dotnet\sdk\2.1.300-preview2-008530\" for file filter '*.t*' (targets and tasks)
No useful results.
Star monitoring :-)
2. Run `dotnet msbuild /fileLogger /noconsolelogger /verbosity:diag`
3. Search "_ReferencesFromRAR" in `msbuild1.log`
Look assigned properties

SteveL-MSFT:
I'm investigating the Windows issues, but it'll take some time

SteveL-MSFT:
I think I fixed the Windows issues, but need to publish new pwrshplugin nupkg to be able to validate

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
```diff
@@ -1,5 +1,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
+
```
iSazonov: Seems we wanted to fix all test files.

SteveL-MSFT: Yes, I only fixed this one because I was in it trying to figure out why the tests were failing.  We should do the rest later.

iSazonov:
> but need to publish new pwrshplugin nupkg to be able to validate

Will it a preview build too? No conflicts with current version?

SteveL-MSFT:
@iSazonov pwrshplugin.dll is purely native code used by WinRM to start PowerShell.  The change is to add the new assemblies (like System.Memory.dll) to the trusted assembly list so that they can be loaded.  Since it's extending the list, there's no conflict.  Need to get it signed first, publish to MyGet, then update the csproj to use the new version.

SteveL-MSFT (approved):
Need some additional reviewers on this.

build.psm1
```diff
@@ -1464,8 +1466,8 @@ function Start-PSxUnit {
 
         # '-fxversion' workaround required due to https://github.com/dotnet/cli/issues/7901#issuecomment-343323674
```
adityapatwardhan: Remove the comment as well.

build.psm1
```diff
@@ -1928,12 +1930,13 @@ function Start-TypeGen
     Find-Dotnet
 
     $GetDependenciesTargetPath = "$PSScriptRoot/src/Microsoft.PowerShell.SDK/obj/Microsoft.PowerShell.SDK.csproj.TypeCatalog.targets"
+#            <_RefAssemblyPath Include="%(_ReferencesFromRAR)%3B" Condition=" '%(_ReferencesFromRAR.Type)' == 'assembly' And '%(_ReferencesFromRAR.PackageName)' != 'Microsoft.Management.Infrastructure' " />
```
adityapatwardhan: This can be removed right?

src/powershell-win-core/NuGet.config
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
```
adityapatwardhan: Do we need this nuget.config? We have one at project root.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -476,16 +476,21 @@ Describe "TabCompletion" -Tags CI {
...
-            if ($IsWindows) {
-                $beforeTab = 'filesystem::{0}\Wind' -f $env:SystemDrive
-                $afterTab = 'filesystem::{0}\Windows' -f $env:SystemDrive
-            } else {
-                $beforeTab = 'filesystem::/us' -f $env:SystemDrive
-                $afterTab = 'filesystem::/usr' -f $env:SystemDrive
+            $tempFolder = [System.IO.Path]::GetTempPath()
+            try
+            {
+                New-Item -ItemType Directory -Path "$tempFolder/helloworld"
```
adityapatwardhan: `New-Item … > $null`

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -505,31 +505,33 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
...
+        $command = "Invoke-WebRequest -Uri ${protocol}://httpbin.org -Proxy '${protocol}://${proxy_address}' -SkipCertificateCheck"
         $result = ExecuteWebCommand -command $command
-        $result.Error.FullyQualifiedErrorId | Should -Be "System.Threading.Tasks.TaskCanceledException,Microsoft.PowerShell.Commands.InvokeWebRequestCommand"
+        $command = "Invoke-WebRequest -Uri '${protocol}://${proxy_address}' -NoProxy"
+        $expectedResult = ExecuteWebCommand -command $command
+        $result.Output.Content | Should -BeExactly $expectedResult.Output.Content
     }
 
-    It "Validate Invoke-WebRequest error with environment proxy set - '<name>'" -TestCases $testCase {
+    It "Validate Invoke-WebRequest with environment proxy set - '<name>'" -Skip:$IsWindows -TestCases $testCase {
```
adityapatwardhan: Please add a comment about why this is skipped on Windows.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -505,31 +505,33 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
...
         param($proxy_address, $name, $protocol)
 
         # Configure the environment variable.
         New-Item -Name ${name} -Value ${proxy_address} -ItemType Variable -Path Env: -Force
```
adityapatwardhan: `New-Item … > $null`

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -776,7 +778,7 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
     #region Redirect tests
 
-    It "Validates Invoke-WebRequest with -PreserveAuthorizationOnRedirect preserves the authorization header on redirect: <redirectType> <redirectedMethod>" -TestCases $redirectTests {
+    It "Validates Invoke-WebRequest with -PreserveAuthorizationOnRedirect preserves the authorization header on redirect: <redirectType> <redirectedMethod>" -Pending -TestCases $redirectTests {
```
adityapatwardhan: Please add a comment why this is marked as pending

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -785,7 +787,7 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         $response.Content.Headers."Authorization" | Should -BeExactly "test"
     }
 
-    It "Validates Invoke-WebRequest preserves the authorization header on multiple redirects: <redirectType>" -TestCases $redirectTests {
+    It "Validates Invoke-WebRequest preserves the authorization header on multiple redirects: <redirectType>" -Pending -TestCases $redirectTests {
```
adityapatwardhan: Same as above.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -794,7 +796,7 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
         $response.Content.Headers."Authorization" | Should -BeExactly "test"
     }
 
-    It "Validates Invoke-WebRequest strips the authorization header on various redirects: <redirectType>" -TestCases $redirectTests {
+    It "Validates Invoke-WebRequest strips the authorization header on various redirects: <redirectType>" -Pending -TestCases $redirectTests {
```
adityapatwardhan: Same as above.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -808,7 +810,7 @@ Describe "Invoke-WebRequest tests" -Tags "Feature" {
 
     # NOTE: Only testing redirection of POST -> GET for unique underlying values of HttpStatusCode.
     # Some names overlap in underlying value.
-    It "Validates Invoke-WebRequest strips the authorization header redirects and switches from POST to GET when it handles the redirect: <redirectType> <redirectedMethod>" -TestCases $redirectTests {
+    It "Validates Invoke-WebRequest strips the authorization header redirects and switches from POST to GET when it handles the redirect: <redirectType> <redirectedMethod>" -Pending -TestCases $redirectTests {
```
adityapatwardhan: Same as above.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -1877,32 +1879,33 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
...
+        $command = "Invoke-RestMethod -Uri ${protocol}://httpbin.org -Proxy '${protocol}://${proxy_address}'"
         $result = ExecuteWebCommand -command $command
-        $result.Error.FullyQualifiedErrorId | Should -Be "WebCmdletWebResponseException,Microsoft.PowerShell.Commands.InvokeRestMethodCommand"
+        $command = "Invoke-RestMethod -Uri '${protocol}://${proxy_address}' -NoProxy"
+        $expectedResult = ExecuteWebCommand -command $command
+        $result.Output | Should -BeExactly $expectedResult.Output
     }
 
-    It "Validate Invoke-RestMethod error with environment proxy set - '<name>'" -TestCases $testCase {
+    It "Validate Invoke-RestMethod with environment proxy set - '<name>'" -Skip:$IsWindows -TestCases $testCase {
```
adityapatwardhan: Same as above.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -2110,7 +2113,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
 
     #region Redirect tests
 
-    It "Validates Invoke-RestMethod with -PreserveAuthorizationOnRedirect preserves the authorization header on redirect: <redirectType> <redirectedMethod>" -TestCases $redirectTests {
+    It "Validates Invoke-RestMethod with -PreserveAuthorizationOnRedirect preserves the authorization header on redirect: <redirectType> <redirectedMethod>" -Pending -TestCases $redirectTests {
```
adityapatwardhan: Same as above.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -2120,7 +2123,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
         $response.Content.Headers."Authorization" | Should -BeExactly "test"
     }
 
-    It "Validates Invoke-RestMethod preserves the authorization header on multiple redirects: <redirectType>" -TestCases $redirectTests {
+    It "Validates Invoke-RestMethod preserves the authorization header on multiple redirects: <redirectType>" -Pending -TestCases $redirectTests {
```
adityapatwardhan: Same as above.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -2130,7 +2133,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
         $response.Content.Headers."Authorization" | Should -BeExactly "test"
     }
 
-    It "Validates Invoke-RestMethod strips the authorization header on various redirects: <redirectType>" -TestCases $redirectTests {
+    It "Validates Invoke-RestMethod strips the authorization header on various redirects: <redirectType>" -Pending -TestCases $redirectTests {
```
adityapatwardhan: Same as above.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
```diff
@@ -2144,7 +2147,7 @@ Describe "Invoke-RestMethod tests" -Tags "Feature" {
 
     # NOTE: Only testing redirection of POST -> GET for unique underlying values of HttpStatusCode.
     # Some names overlap in underlying value.
-    It "Validates Invoke-RestMethod strips the authorization header redirects and switches from POST to GET when it handles the redirect: <redirectType> <redirectedMethod>" -TestCases $redirectTests {
+    It "Validates Invoke-RestMethod strips the authorization header redirects and switches from POST to GET when it handles the redirect: <redirectType> <redirectedMethod>" -Pending -TestCases $redirectTests {
```
adityapatwardhan: Same as above.

adityapatwardhan (changes requested):

src/powershell-win-core/NuGet.config
SteveL-MSFT: I think this can be removed, was trying to figure out why it wasn't picking up latest psrp.windows pkg but it turned out I built the pkg wrong.

build.psm1
iSazonov: Yes.

build.psm1
iSazonov: Fixed.

src/powershell-win-core/NuGet.config
iSazonov: Removed.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
iSazonov: Fixed.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: Fixed.

test/powershell/Modules/Microsoft.PowerShell.Utility/WebCmdlets.Tests.ps1
iSazonov: I believe we should open new tracking issues for proxy and authorization header tests - we can not release 6.1.0 without addressing this. These comments are superfluous.

iSazonov:
I removed some dummy comment to speed up refreshing the page.

adityapatwardhan (approved):

iSazonov:
@daxian-dbw Could you please review the important PR too and merge?

src/powershell-native/nativemsh/pwrshcommon/pwrshcommon.cpp
```diff
@@ -732,6 +732,7 @@ namespace NativeMsh
         "System.Globalization.Extensions",
         "System.IO",
         "System.IO.Compression",
```
daxian-dbw: It seems `System.ComponentModel.Composition.dll` is gone. Then it should be removed from this list now.

daxian-dbw: A note about `Microsoft.CodeAnalysis.VisualBasic.dll` we added from the `Add-Type` refactoring. Since we plan to remove the VB support, we shouldn't add that dll in this list.

markekraus (approved):

test/tools/WebListener/NuGet.config
daxian-dbw: This `nuget.config` file doesn't introduce a new feed. Do we need it only because we need to try `dotnet-core` before `nuget.org`?

tools/install-powershell.ps1
```diff
@@ -117,7 +117,7 @@ try {
         } elseif ($IsLinuxEnv) {
             "powershell-linux-x64"
         } elseif ($IsMacOSEnv) {
-            "powershell-osx.10.12-x64"
+            "powershell-osx-x64"
```
daxian-dbw: @TravisEz13 Does this renaming require packaging script changes?

tools/packaging/packaging.psm1
```diff
@@ -106,7 +106,7 @@ function Start-PSPackage {
             -not $PSModuleRestoreCorrect -or                        ## Last build didn't specify '-PSModuleRestore' correctly
             $Script:Options.Runtime -ne $Runtime -or                ## Last build wasn't for the required RID
             $Script:Options.Configuration -ne $Configuration -or    ## Last build was with configuration other than 'Release'
-            $Script:Options.Framework -ne "netcoreapp2.0")          ## Last build wasn't for CoreCLR
+            $Script:Options.Framework -ne "netcoreapp2.1")          ## Last build wasn't for CoreCLR
```
daxian-dbw: I see other places using `netcoreapp2.0`. They all need to be updated.
```powershell
PS:103> dir *.psm1 -Recurse | sls -SimpleMatch netcoreapp2.0

test\tools\OpenCover\OpenCover.psm1:627:        [parameter()]$PowerShellExeDirectory =
"${script:psRepoPath}/src/powershell-win-core/bin/CodeCoverage/netcoreapp2.0/win7-x64/publish",
tools\packaging\packaging.psm1:2221:        New-MSIPackage -Verbose -ProductCode (New-Guid) -ProductSourcePath
'.\src\powershell-win-core\bin\Debug\netcoreapp2.0\win7-x64\publish' -ProductTargetArchitecture x64 -ProductVersion
'1.2.3'
build.psm1:560:    # publish netcoreapp2.0 reference assemblies
```

build.psm1
```diff
@@ -2215,9 +2212,12 @@ function Start-CrossGen {
         throw "crossgen is not available for this platform"
     }
 
+    $dotnetRuntimeVersion = $dotnetCLIRequiredVersion -match "(\d\.\d)\..*" | ForEach-Object { $matches[1] }
```
daxian-dbw: This will give us the Major.Minor version of the sdk, not the runtime. The sdk version is different from the runtime version, e.g. we use sdk 2.1.4 currently with runtime version 2.0. It may happen in future that sdk moved to 2.2 while we are still using 2.1 runtime.

build.psm1
```diff
@@ -2215,9 +2212,12 @@ function Start-CrossGen {
...
+
     # Get the CrossGen.exe for the correct runtime with the latest version
     $crossGenPath = Get-ChildItem $script:Environment.nugetPackagesRoot $crossGenExe -Recurse | `
                         Where-Object { $_.FullName -match $crossGenRuntime } | `
+                        Where-Object { $_.FullName -match $dotnetRuntimeVersion } | `
```
daxian-dbw: Is this really necessary? I didn't try on Linux and macOS, but on Windows, the 2.1 folder is the first one after sort (I have 2.0.6, 2.0.5 and 2.0.0 runtime package installed):
```
PS:107> dir crossgen.exe -Recurse | ? FullName -Match win-x64 | sort -Property FullName -Descending | select -First 1

    Directory: C:\Users\dongbow\.nuget\packages\runtime.win-x64.microsoft.netcore.app\2.1.0-rc1-26423-06\tools

Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----        4/23/2018   9:07 AM        2916176 crossgen.exe
```

daxian-dbw (changes requested):

tools/packaging/packaging.psm1
iSazonov: Fixed.

build.psm1
iSazonov: Seems we need review this after .Net Core 2.1 release because it introduce new policy - always use latest version. It is still not work for Preview versions. And seems they want adjust this versions for simplicity.

build.psm1
daxian-dbw: You don't have to build powershell to find out this. Use a simple project that targets netcoreapp2.1, then run this:
```
dotnet msbuild .\temp.csproj /t:ResolveAssemblyReferencesDesignTime /fileLogger /noconsolelogger /v:diag
```
Then you can inspect the properties of `_ReferencesFromRAR` in `msbuild.log`. I think we should add comment here to describe this process.

tools/install-powershell.ps1
TravisEz13: The package is already named like this.  This is an unrelated bug fix.

build.psm1
daxian-dbw: > It is still not work for Preview versions.

What is not working for preview versions?

src/powershell-native/nativemsh/pwrshcommon/pwrshcommon.cpp
SteveL-MSFT: I'll take care of this

test/tools/WebListener/NuGet.config
SteveL-MSFT: Yes, to pick up newer version of aspnetcore than what is on nuget.org currently (otherwise it didn't work correctly on my macBook).  We should get rid of this in the future.

build.psm1
SteveL-MSFT: The problem we had in CI w/o this is that we pick up the first version that matches which was crossgen 2.2 (some preview version) which is not compatible with the 2.1 built binaries.

build.psm1
SteveL-MSFT: This was happening in CI

build.psm1
daxian-dbw: I wonder how come we have the 2.2 runtime packages in our CI runs ... I wish we can figure this out.

This line of script gets the `Major.Minor` of the sdk version, not the runtime version, which is not what we really want (not future-proof). Maybe explicitly use `2.1*` for now (which is also not future-proof, but at least intuitive than the regex matching)?

daxian-dbw:
We also need to update the versions of the dotnet NuGet packages referred by `PackageReference` in our `.csproj` files. For example, `Microsoft.Win32.Registry.AccessControl` in `system.management.automation.csproj` should be updated to `4.5.0-rc1-26423-06`.

Note that, we should pick the `rc1-26423` ones, which are from the same build of the `netcoreapp` we are using with the new sdk (`AssemblyFileVersion("4.6.26423.06")`).

daxian-dbw:
Pushed commit to update all dotnet NuGet package versions used in `<PackageReference>` nodes.

daxian-dbw:
Pushed commit to update all dotnet NuGet package versions used in `<PackageReference>` nodes.
@TravisEz13, would this change require any update to `files.wxs`? (I can run a package build, but it's faster to ask 😄)

build.psm1
SteveL-MSFT: Perhaps we should rename the variable to `$dotnetSdkVersion`.  Keeping this regex would mean that when we move to 2.2, but there's a 2.3 preview out we won't be broken.

daxian-dbw: It also means that when we move to sdk 2.2 but still building against runtime 2.1 (similar to what we have today: sdk 2.1.4, runtime 2.0), we will be broken.

SteveL-MSFT: Using `dotnet --list-runtimes` shows `Microsoft.NETCore.App 2.1.0-rc1-26423-06` which matches the runtime version of crossgen we need.  So perhaps we can use `dotnet --list-runtimes` and select the newest one and match that runtime version to find appropriate crossgen.

daxian-dbw: We can use `$Script:Options.Framework` to get the runtime version. The framework will always be `netcoreapp2.1` until we update it again.
```
$dotnetRuntimeVersion = $script:Options.Framework -replace 'netcoreapp'
```

daxian-dbw (approved):

build.psm1
iSazonov: @daxian-dbw 
> What is not working for preview versions?

I meant that .Net Core 2.1 has new model and always uses (will use) latest runtime versions with `dotnet publish` but it is not working for preview versions.

iSazonov:
LGTM.

- Seems we don't open a tracking issue for web cmdlet proxy tests.
- Need review and make cleanups our workarounds for `curl`.
- Need review packaging.

daxian-dbw:
@iSazonov I'm not familiar with the issue regarding web cmdlet proxy tests and workaround for curl. You probably have more context, so can you please open a tracking issue for that?

As to packaging review, do you have any specific concerns about how this change would affect packing?
Both the `[Feature]` CI run and `[Package]` CI run passed. We can kick off a release build after merging this PR to test the release build pipeline. /cc @TravisEz13, feel free to weigh in.

@SteveL-MSFT, @adityapatwardhan and @markekraus, since new commits were pushed after you signed off, could you please take another look at this PR? Thanks!

TravisEz13:
@daxian-dbw I agree with starting a release build after this is merged.

daxian-dbw:
About the Visual Studio build, I don't know if it will still work after this merge ... @TravisEz13 Does VS2017 allow configuring which dotnet sdk to use?

SteveL-MSFT:
This has been a long thread, @iSazonov can you summarize the proxy issue with the webcmdlet tests?  I changed the tests as they were originally just wrong and incomplete.  I think we should probably remove the tests for http_proxy env var which we skip on Windows because that is a side effect of corefx taking dependency on libcurl for non-Windows and not something supported explicitly by corefx nor the webcmdlets.

iSazonov:
The PR LGTM.

SteveL-MSFT (approved):
LGTM

adityapatwardhan (approved):

daxian-dbw:
PR merged. Thanks everyone for getting this done!

Issue #6749, opened 2018-04-26
cd-ing to a new drive and back does not restore path


rjmholt:
Steps to reproduce
------------------

```powershell
C:\Users\me\Documents> cd Z:
Z:\> cd C:
```

Expected behavior
-----------------

```powershell
Z:\> cd C:
C:\Users\me\Documents>
```

Actual behavior
---------------

```powershell
Z:\> cd C:
C:\>
```

Environment data
----------------

```powershell
> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.1.0-preview.1
PSEdition                      Core
GitCommitId                    v6.1.0-preview.1-99-g35d8de927b8e408461a5b4fc40ba20016c68433e
OS                             Microsoft Windows 10.0.16299
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

```

It looks like https://github.com/PowerShell/PowerShell/pull/6600 caused a regression. It seems to be Windows only. I tested with https://github.com/PowerShell/PowerShell/commit/a9781dedc29978fd4b41d02d1a59826461d14b4c with no repro and then with https://github.com/PowerShell/PowerShell/commit/35d8de927b8e408461a5b4fc40ba20016c68433e to successfully repro.

Credit to @jianyunt.

iSazonov:
We need add new tests for this but I guess CI has only one filesystem drive.

GeeLaw:
@iSazonov We could use `New-PSDrive -Name TestSetLocation -Provider FileSystem -Root Path\To\Root` and later `Remove-PSDrive -Name TestSetLocation` later. Alternatively, we could use `subst.exe` for Windows.

iSazonov:
@GeeLaw Thanks! I guess we could use `New-PSDrive -Name "Z" -Provider FileSystem -Root Path\To\Root` - seems one char name can be important in the test.

mcbobke:
Hoping to have some time this weekend to take a closer look at this regression. Thanks for filing the issue!

mcbobke:
Dug into this tonight and I think [this](https://github.com/mcbobke/PowerShell/commit/fdac5e09f2968a10ee87a379924949680a97a974) may be a move in the right direction. The expected behavior as noted by @rjmholt is produced.

iSazonov:
[...]

PR #6600, opened 2018-04-09
Fix for files/folders not created when given path is drive root (#5228)

ghost:
## PR Summary

If the `Path` argument to `New-Item` is the root of a drive such as `C:\`, the LocationGlobber attempts to generate a drive root relative path from the path `C:`. `GetDriveRootRelativePathFromPSPath` determines that this is an absolute path, and strips the drive name and colon from the path; originally, this would result in `GenerateRelativePath` determining that the current working directory should be given as the relative path. Now, a check is made on whether stripping the drive name and colon results in an empty string; if it does, the root of the current PSDrive `context.Drive.Root` is returned as the relative path.

This fixes #5228.

## PR Checklist

- [X] [PR has a meaningful title](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - Use the present tense and imperative mood when describing your changes
- [X] [Summarized changes](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [X] [Change is not breaking](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes)
- [X] [Make sure all `.h`, `.cpp`, `.cs`, `.ps1` and `.psm1` files have the correct copyright header](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [X] This PR is ready to merge and is not [Work in Progress](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---work-in-progress).
    - If the PR is work in progress, please add the prefix `WIP:` to the beginning of the title and remove the prefix when the PR is ready.
- **User-facing changes**
    - [X] Not Applicable
    - **OR**
    - [ ] User-facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
        - [X] Issue filed - Issue link: https://github.com/PowerShell/PowerShell/issues/5228#
- **Testing - New and feature**
    - [ ] Not Applicable or can only be tested interactively
    - **OR**
    - [X] [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
[...]

Issue #5228, opened 2017-10-25
New-Item ignore's Path param if Name param is specified


affieuk:
Steps to reproduce
------------------

```powershell
Set-Location ~
New-Item -Path C:\ -Name Test -ItemType Directory -Verbose
```

Expected behavior
-----------------

```none
Test folder created in Root C:\
```

Actual behavior
---------------

```none
Folder created in current path
PS C:\Users\Aftab> New-Item -Path C:\ -Name Test -ItemType Directory -Verbose
VERBOSE: Performing the operation "Create Directory" on target "Destination: C:\Users\Aftab\Test".


    Directory: C:\Users\Aftab


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----       25/10/2017     13:51                Test


PS C:\Users\Aftab> Get-Item C:\Test
Get-Item : Cannot find path 'C:\Test' because it does not exist.
At line:1 char:1
+ Get-Item C:\Test
+ ~~~~~~~~~~~~~~~~
    + CategoryInfo          : ObjectNotFound: (C:\Test:String) [Get-Item], ItemNotFoundException
    + FullyQualifiedErrorId : PathNotFound,Microsoft.PowerShell.Commands.GetItemCommand

PS C:\Users\Aftab> Get-Item ~\Test


    Directory: C:\Users\Aftab


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----       25/10/2017     13:51                Test
```

Environment data
----------------

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.0.0-beta.8
PSEdition                      Core
GitCommitId                    v6.0.0-beta.8
OS                             Microsoft Windows 10.0.15063
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

affieuk:
This also occurs in a mounted a vhdx.
```none
PS E:\Users> New-Item -Path E:\ -Name Test -ItemType Directory -Verbose
VERBOSE: Performing the operation "Create Directory" on target "Destination: E:\Users\Test".


    Directory: E:\Users


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----       25/10/2017     14:09                Test


PS E:\Users> Get-Item E:\Test
[...]

PR #6751, opened 2018-04-27
Revert "Fix for files/folders not created when given path is drive root (#5228)"

rjmholt:
Reverts PowerShell/PowerShell#6600.

This caused a regression in drive behaviour where drive paths were not restored. See https://github.com/PowerShell/PowerShell/issues/6749.

We will need to reopen https://github.com/PowerShell/PowerShell/issues/5228.

### DIFF
--- a/src/System.Management.Automation/namespaces/LocationGlobber.cs
+++ b/src/System.Management.Automation/namespaces/LocationGlobber.cs
@@ -2013,9 +2013,6 @@ namespace System.Management.Automation
             // Check to see if the path is relative or absolute
             bool isPathForCurrentDrive = false;
 
-            // Check to see if the path is to the root of a drive
-            bool isPathForRootOfDrive = false;
-
             if (IsAbsolutePath(path, out driveName))
             {
                 Dbg.Diagnostics.Assert(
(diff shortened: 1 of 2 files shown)


daxian-dbw:
@mcbobke We need to revert the change from #6600 because it caused a regression. We will revisit that fix later.

mcbobke:
@daxian-dbw understood

iSazonov:
Why so many commits reverted?

rjmholt:
I'm not sure, I just clicked on GitHub's revert button

iSazonov:
We merged only one commit https://github.com/PowerShell/PowerShell/commit/35d8de927b8e408461a5b4fc40ba20016c68433e

rjmholt:
Yes, I think this is wrong because it seems to want to revert several commits.

But I just went to #6600 and clicked on this:
revert

iSazonov:
Maybe use the command line `git revert 35d8de9`?

rjmholt:
Ok, I think that's worked.

daxian-dbw (approved):

iSazonov (approved):

PR #6782, opened 2018-04-30
Quote Multipart/Form-Data Field Names

### LINKED ISSUE #6780: Invoke-WebRequest multipart form-data fails file upload when field names not quoted
This happens using the new `-Form` parameter that takes a hashtable of values and creates the multi-part form data.

Steps to reproduce
------------------
1. Find URL to webservice that accepts POST of multi-part form data
2. Run Fiddler or Wireshark to observe the request body
3. Execute IWR similar to below:

```powershell
Invoke-WebRequest -Uri $Uri -Method POST -Form @{
    APPLICATION_NAME = "Name"
    APPLICATION_VERION = "3.0"
    UPLOADED_FILE = Get-Item .\foo.zip
}
```

Expected behavior
-----------------
Upload works as it does when using Python script to upload the file.  Observe that in this Fiddler trace, that the form-data field values are not quoted:

image

Actual behavior
---------------
PowerShell fails and after some experimenting it is because the name values are not quoted:

image

With just a few extra quotes e.g.:

```powershell
Invoke-WebRequest -Uri $Uri -Method POST -Form @{
    '"APPLICATION_NAME"' = "Name"
    '"APPLICATION_VERION"' = "3.0"
    '"UPLOADED_FILE"' = Get-Item .\foo.zip
}
```

I can get the PowerShell request to quote all the field data values **except** the file's `filename` field.  And that is enough to break the upload.  Besides that, the above workaround is ugly.

As an aside, is there a way to control the filename value?  If there was, I could work-around this by supplying the quotes myself.  Also, I could see someone wanting to change the filename to something other than the original file's filename.

This also makes me wonder if we could use a new parameter like `-RequestFilter` that took a scriptblock with a signature like `{param($requestText) ... }` where you could get the request text just before it is sent.  Then you could manipulate it however you needed to and return that text from this scriptblock.  With such a feature I might be able to do this:

```powershell
Invoke-WebRequest -Uri $Uri -Method POST `
    -RequesetFilter {param($r) $r -replace '(?<=filename=)(\S+)','"$1"' } `
    -Form @{
        '"APPLICATION_NAME"' = "Name"
        '"APPLICATION_VERION"' = "3.0"
        '"UPLOADED_FILE"' = Get-Item .\foo.zip
    }
```

Environment data
----------------

```powershell
> $PSVersionTable
PSVersion                      6.1.0-preview.2
GitCommitId                    v6.1.0-preview.2
```

markekraus:
## PR Summary

* fixes  #6780
* When `-Form` is used, the field names and the filename value will be enclosed in quotes
* fixes multipart/form-data submission issues with some APIs
* brings functionality closer to modern browsers and curl

Since `-Form` has not been released in an actual release version, this is not a breaking change. 

There is no decent way to test this as the metadata for text fields is lost when the form is processed in ASP.NET.  Only way would be to return the raw request body and parse it with regex.

## PR Checklist

- **User-facing changes**
    - [X] Not Applicable
    - **OR**
    - [X] Not Applicable or can only be tested interactively
    - **OR**

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
@@ -1817,7 +1817,8 @@ namespace Microsoft.PowerShell.Commands
         private StringContent GetMultipartStringContent(Object fieldName, Object fieldValue)
         {
             var contentDisposition = new ContentDispositionHeaderValue("form-data");
-            contentDisposition.Name = LanguagePrimitives.ConvertTo<String>(fieldName);
+            // .NET does not enclose field names in quotes, however, modern browsers and curl do.
+            contentDisposition.Name = $"\"{LanguagePrimitives.ConvertTo<String>(fieldName)}\"";
 
             var result = new StringContent(LanguagePrimitives.ConvertTo<String>(fieldValue));
             result.Headers.ContentDisposition = contentDisposition;
@@ -1833,7 +1834,8 @@ namespace Microsoft.PowerShell.Commands
         private StreamContent GetMultipartStreamContent(Object fieldName, Stream stream)
         {
             var contentDisposition = new ContentDispositionHeaderValue("form-data");
-            contentDisposition.Name = LanguagePrimitives.ConvertTo<String>(fieldName);
+            // .NET does not enclose field names in quotes, however, modern browsers and curl do.
+            contentDisposition.Name = $"\"{LanguagePrimitives.ConvertTo<String>(fieldName)}\"";
 
             var result = new StreamContent(stream);
             result.Headers.ContentDisposition = contentDisposition;
@@ -1850,7 +1852,8 @@ namespace Microsoft.PowerShell.Commands
         private StreamContent GetMultipartFileContent(Object fieldName, FileInfo file)
         {
             var result = GetMultipartStreamContent(fieldName: fieldName, stream: new FileStream(file.FullName, FileMode.Open));
-            result.Headers.ContentDisposition.FileName = file.Name;
+            // .NET does not enclose field names in quotes, however, modern browsers and curl do.
+            result.Headers.ContentDisposition.FileName = $"\"{file.Name}\"";
 
             return result;
         }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/Common/WebRequestPSCmdlet.Common.cs
@@ -1818,7 +1818,7 @@ namespace Microsoft.PowerShell.Commands
         {
             var contentDisposition = new ContentDispositionHeaderValue("form-data");
             // .NET does not enclose field names in quotes, however, modern browsers and curl do.
-            contentDisposition.Name = $"\"{LanguagePrimitives.ConvertTo<String>(fieldName)}\"";
+            contentDisposition.Name = "\"" + LanguagePrimitives.ConvertTo<String>(fieldName) + "\"";
 
             var result = new StringContent(LanguagePrimitives.ConvertTo<String>(fieldValue));
             result.Headers.ContentDisposition = contentDisposition;
@@ -1835,7 +1835,7 @@ namespace Microsoft.PowerShell.Commands
         {
             var contentDisposition = new ContentDispositionHeaderValue("form-data");
             // .NET does not enclose field names in quotes, however, modern browsers and curl do.
-            contentDisposition.Name = $"\"{LanguagePrimitives.ConvertTo<String>(fieldName)}\"";
+            contentDisposition.Name = "\"" + LanguagePrimitives.ConvertTo<String>(fieldName) + "\"";
 
             var result = new StreamContent(stream);
             result.Headers.ContentDisposition = contentDisposition;
@@ -1853,7 +1853,7 @@ namespace Microsoft.PowerShell.Commands
         {
             var result = GetMultipartStreamContent(fieldName: fieldName, stream: new FileStream(file.FullName, FileMode.Open));
             // .NET does not enclose field names in quotes, however, modern browsers and curl do.
-            result.Headers.ContentDisposition.FileName = $"\"{file.Name}\"";
+            result.Headers.ContentDisposition.FileName = "\"" + file.Name + "\"";
 
             return result;
         }


SteveL-MSFT:
Should we open an issue against corefx?  If corefx supports/fixes this in the future, will we have double quotes?

markekraus:
At this point it would be a breaking change in CoreFX. This is the same behavior for `System.Net.Http.MultipartFormDataContent` in .NET Framework 4.5 when it was added. I found a [blog article form 2013](https://blog.xdumaine.com/uploading-a-photo-as-multipart-form-data-in-c/) that references this issue.  If they did add this, they would probably do so through a new property in which case we wouldn't be broken.

markekraus:
I found this https://github.com/dotnet/corefx/issues/26886#issuecomment-381697212 though, this begs the question why every browser and curl wraps the name and filename fields values in quotes and .NET does not?

SteveL-MSFT (approved):
LGTM

iSazonov:
Seems it is important fix so we need to add tests. Maybe use output with `-WhatIf`?

iSazonov:
I believe we should have tests for the uploading.
I assumed that we could do verbose output of the form but seems it is impossible. So we need full feature tests. Perhaps @JamesWTruher and @adityapatwardhan have ideas about this tests.

SteveL-MSFT:
@markekraus I think it would be ok to add a test to weblistener to specifically return the field names and values as raw text.  Since the values are controlled by the Pester test, it should know exactly what should be returned without having to resort to regex

markekraus:
@SteveL-MSFT The problem is the ASP.NET does not provide whether or not the feild names are quoted when you consume the response as a form. The only way to test that is to echo back the raw HTTP request as a string.

SteveL-MSFT:
@markekraus I see, so you either get just the name value pairs stripped of the quotes or the entire raw request.  It seems the risk is low and the cost is high.  I would be fine creating a new issue to add such a test in the future, particularly if it's useful beyond this specific issue.

markekraus:
Added #6794 to track possible adding an echo controller to WebListener.

iSazonov:
>the cost is high

@markekraus implemented WebListener  so fast that I'm still amazed :-)

iSazonov (dismissed):

iSazonov:
@daxian-dbw Do you agree that we use the `$""` string interpolation instead of explicit formatting?

daxian-dbw:
I don't have a strong opinion. I don't want to block people from using C# language features :)
When converting an interpolated string to a string instance, the current culture is used, which seems fine in this change. But be careful if you want a culture-invariant string.

iSazonov:
We could
```c#
 = "\"" + ... + "\"";
```

iSazonov:
Sorry for wrong close.

daxian-dbw:
You can also choose culture with the string interpolation:
```c#
FormattableString message = $"The speed of light is {speedOfLight:N3} km/s.";
string messageInInvariantCulture = FormattableString.Invariant(message);
```
See https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated

iSazonov:
I think that the double formatting superfluous here. 
Concatenating enough.
Or we could use `String.Create()` from .Net Core 2.1 (later).

markekraus:
@iSazonov Does the culture on these even matter? I'm not doing any format transformations, I'm just shoving one string inside another.

iSazonov:
I remember that we have already corrected misuses of culture in formatting. So I'm trying to be careful here. 
If culture is not important here, we should not use this pattern at all.
Now we have already merged #6718 and can use .Net Core 2.1 features - it is very optimized for string manipulations. 
I believe here we could use `String.Create()`. Samples: 
https://msdn.microsoft.com/en-us/magazine/mt814808.aspx
http://source.dot.net/#System.Private.Uri/System/Uri.cs,1412

iSazonov:
>So I'm asking.. does it even matter here?

No. The discussion is about _code style_. Bad style is a open door for camouflaged bugs. Therefore, we require explicit curly braces in the `if` and explicit culture in formattings - after fixing and discussing the latter we even added manuals to our documentation #4983 
Here I wanted to understand whether we can use `$""` and now believe that we should not use this.

markekraus:
... why is this bad style? what is wrong with `$""`?  I really don't understand what your objecttion is. You don't like it?

markekraus:
@iSazonov Also, I cannot figure out how you intend to use String.Create() here. Can you provide me with an example relevant to this PR? It seems like it is overkill for string concatenation.

markekraus:
ok.. so `$""` is banned even when used properly? So what is the best way to concat strings in this project then? using `+` has a perf hit so that's out. `String.Concat()` is in the same boat and needs multiple calls. `String.Create()` is confusing (seriously.. what is a `TState`? why is that not documented anywhere? how exactly would you even use this to concat string?) In this case, `String.Format()` and `$""` are identical. I could use `String.Format()` with the explicit formatter, but that seems really verbose for just enclosing a string in quotes.

iSazonov:
>ok.. so $"" is banned even when used properly?

If other maintainers agree that we should always use _explicit_ culture then yes.

Previously we use + for concatinations for few arguments (~< 5). Othervise StringBuilder is preferable.
Currently we can and should use .Net Core 2.1 and C# 7.2 features to decrease allocations.
Now we can use `ValueStringBuilder`, `String.Create()`, Span<T> and many new string method overloads.

In the case I'd use `String.Create()`. Although you can ignore it now and use +.
Sample:
```c#
            var name = LanguagePrimitives.ConvertTo<String>(fieldName);
            contentDisposition.Name = String.Create(2+name.Length, name, (Span<char> destination, string str) =>
            {
                destination[0] = '"';
                str.AsSpan().CopyTo(destination.Slice(1));
                destination[destination.Length-1] = '"';
            });
```

markekraus:
> If other maintainers agree that we should always use explicit culture then yes.

Why? Why always use explicit culture? String concatenation does not require it. It _is_ relvant when doing formatting or transformations, but just combining strings it's senseless.

> Previously we use + for concatinations for few arguments (~< 5). 

If this will unblock this PR so it will be merged, I will change it to use `+`. I still challenge the logic behind this debate.

> In the case I'd use String.Create(). 

Why? That is a seriously over-complicated verbose and tough to follow piece code just to put `"` around a string. I get it if we were working with concatenating some large strings or this code was being called so many times that performance really matters. but this is a fairly esoteric feature that wont be used often or heavily and it's more important the the code be readable here than to squeeze every ounce of performance out of it.

iSazonov:
> this is a fairly esoteric feature that wont be used often or heavily

We never know exactly what users will do.

markekraus:
> We never know exactly what users will do.

We do know that this is a form submission style that is not as widely adopted as standard forms. We do know that the intended purpose of this submission method is GUI browser based communications. We do know that some APIs do use this for file uploads, but not all. We do know that while there have been asks for this feature to be added, not many. 

If I didn't have my own personal need for it, it may never have even entered consideration to be adopted in the web cmdlets.

Yes, we are not clairvoyant and we cannot know 100% what users will do, but, let's be realistic here: the string concatenation for quoting field names filename values and of this feature is NEVER going to be a performance concern. Especially when it is already associated with slow, long running, and memory intense tasks (uploading data). The few fractions of a millisecond shaved off and memory usage saved by using `String.Create()` here is severely outweighed by a multitude of other critical performance losses happening at the same time in the same feature.

What I do know for certain is that I will have to look at this code in the future. And very likely, someone else will too. I do know that String.Create() is complex and not easy to read. I do know that I can skim the `$""` method and see very quickly what is being done. I do know that It takes me a few seconds to figure out that the code you provided is just wrapping a string in quotes.

If you want to add a string concatenation utility method that allows us to abstract `String.Create()` throughout the project, I would consider using that. But I don't see the need to add such a thing directly in the web cmdlets. I also don't see any benefit at all in using direct `String.Create()` calls here either.

iSazonov:
Forget about `String.Create()` - you could use `String.Concat()` or +.

>I do know that I can skim the $"" method and see very quickly what is being done. I do know that It takes me a few seconds to figure out that the code you provided is just wrapping a string in quotes.

This is not obviously for code reviewer. My first impression was that it evidently. Then there were questions. Why we use formatting if we need quoting? Why we use `InvariantCulture` in `ConvertTo` and `CurrentCulture` in `$""`? What's the trick? Should we worry about allocations? - we just had a discussion that PowerShell sometimes consumes too much resources and we should be more thrifty. As a result, we received a good exchange of views. It was useful for me to know your point of view.

Summing up this discussion I think `String.Concat()` will be the best pattern here.

rkeithhill:
In this case, the compiler will translate the `+` concatenated strings to a call to String.Concat:

image

iSazonov:
@rkeithhill What plugin you use in VS Code to see IL? In what will be compiled several operators `"a"+"b"+"c"`?

rkeithhill:
I use JetBrains' dotPeek.  Since `String.Concat()` has an overload that takes `params string[]`, I assume it take as many as you supply it.  But I haven't tested that. Here it is with 8 concats:
image

rkeithhill:
It would be much appreciated if this PR makes it in for Preview3.  Just sayin'.  :-)

PR #6839, opened 2018-05-08
Fix tab completions for a hash table

### LINKED ISSUE #5322: Completion of hash table key has invalid replacement index/length
Steps to reproduce
------------------

```powershell
$line = 'gps | format-table -Property @{'
$exp = TabExpansion2 $line $line.Length
$exp.ReplacementIndex
$exp.ReplacementLength
$exp.CompletionMatches.Count
```

Expected behavior
-----------------
ReplacementIndex should be `$line.Length`, the length should be 0.

```none
31
0
5
```
If you change the repro to add a space at the end, then you will see 32, 0, 5
```powershell
$line = 'gps | format-table -Property @{ '
```
Actual behavior
---------------
The replacement index and length are negative.

```none
-1
-1
5
```
This raises an exception in PSReadLine (https://github.com/lzybkr/PSReadLine/issues/534)
In Windows PowerShell, if you remove PSReadLine and try tab interactively (instead of manually calling `TabExpansion2`), PowerShell will crash - as in exit after reporting a crash to Watson.
PowerShell ISE doesn't crash, but it also doesn't complete anything.

Environment data
----------------

```powershell
> $PSVersionTable

PSVersion                      5.1.16299.19                                                                            
BuildVersion                   10.0.16299.19                                                                           



```

iSazonov:
## PR Summary

Fix #5322.

Now working:
- Get-Date | Sort-Object @{Expression=\<tab>
- Get-Date | Sort-Object @{Expression=...;\<tab>

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
+++ b/src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
@@ -473,6 +473,13 @@ namespace System.Management.Automation
                         replacementLength = 0;
                         break;
 
+                    case TokenKind.Semi:
+                        // Handle scenarios such as 'gci | Format-Table @{Label=...;<tab>'
+                        result = GetResultForHashtable(completionContext);
+                        replacementIndex += 1;
+                        replacementLength = 0;
+                        break;
+
                     case TokenKind.Number:
                         // Handle scenarios such as Get-Process -Id 5<tab> || Get-Process -Id 5210, 3<tab> || Get-Process -Id: 5210, 3<tab>
                         if (lastAst is ConstantExpressionAst &&
@@ -531,6 +538,18 @@ namespace System.Management.Automation
                     case TokenKind.AtParen:
                     case TokenKind.LParen:
                         {
+                            if (tokenAtCursor.Kind == TokenKind.Equals && lastAst is HashtableAst hashTableAst && CheckForPendingAssignment(hashTableAst))
+                            {
+                                if (lastAst.Parent is DynamicKeywordStatementAst)
+                                {
+                                    // Handle scenarios such as 'configuration foo { File ab { Attributes ='
+                                    bool unused;
+                                    result = GetResultForEnumPropertyValueOfDSCResource(completionContext, string.Empty, ref replacementIndex, ref replacementLength, out unused);
+                                    break;
+                                }
+                                // Handle scenarios such as 'gci | Format-Table @{Label=<tab>'
+                                return null;
+                            }
                             if (lastAst is AttributeAst)
                             {
                                 completionContext.ReplacementIndex = replacementIndex += tokenAtCursor.Text.Length;
--- a/src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
+++ b/src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
@@ -547,10 +547,15 @@ namespace System.Management.Automation
                                 completionContext.ReplacementLength = replacementLength = 0;
                                 result = GetResultForAttributeArgument(completionContext, ref replacementIndex, ref replacementLength);
                             }
-                            else if (lastAst.Parent is DynamicKeywordStatementAst || lastAst.Parent is CommandExpressionAst)
+                            else if (lastAst is HashtableAst hashTableAst && !(lastAst.Parent is DynamicKeywordStatementAst) && CheckForPendingAssignment(hashTableAst))
                             {
-                                // 1. Handle scenarios such as 'configuration foo { File ab { Attributes ='
-                                // 2. Handle auto completion for enum/dependson property of DSC resource,
+                                // Handle scenarios such as 'gci | Format-Table @{Label=<tab>' if incomplete parsing of the assignment.
+                                return null;
+                            }
+                            else
+                            {
+                                // Handle scenarios such as 'configuration foo { File ab { Attributes ='
+                                // (auto completion for enum/dependson property of DSC resource),
                                 // cursor is right after '=', '(' or '@('
                                 //
                                 // Configuration config
@@ -563,11 +568,6 @@ namespace System.Management.Automation
                                 //
                                 bool unused;
                                 result = GetResultForEnumPropertyValueOfDSCResource(completionContext, string.Empty, ref replacementIndex, ref replacementLength, out unused);
-                            }
-                            else if (lastAst is HashtableAst hashTableAst && CheckForPendingAssignment(hashTableAst))
-                            {
-                                // Handle scenarios such as 'gci | Format-Table @{Label=<tab>' if incomplete parsing of the assignment.
-                                return null;
                             }
                             break;
                         }
--- a/src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
+++ b/src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
@@ -475,9 +475,12 @@ namespace System.Management.Automation
 
                     case TokenKind.Semi:
                         // Handle scenarios such as 'gci | Format-Table @{Label=...;<tab>'
-                        result = GetResultForHashtable(completionContext);
-                        replacementIndex += 1;
-                        replacementLength = 0;
+                        if (lastAst is HashtableAst)
+                        {
+                            result = GetResultForHashtable(completionContext);
+                            replacementIndex += 1;
+                            replacementLength = 0;
+                        }
                         break;
 
                     case TokenKind.Number:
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
```diff
@@ -473,6 +473,13 @@ internal List<CompletionResult> GetResultHelper(CompletionContext completionCont
                         replacementLength = 0;
                         break;
 
+                    case TokenKind.Semi:
```
lzybkr: Does adding this case affect `get-something; e<TAB>` - in other words, no hashtable, just simple command completion?

iSazonov: No, I don't see the side affect.

iSazonov: I re-check `get-something; e<Tab>` - in Windows PowerShell 5.1 and in PowerShell Core 6.0 simple command completion works - nothing new added in the PR.

iSazonov:
@SteveL-MSFT @daxian-dbw Could you please review too?

src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
```diff
@@ -473,6 +473,13 @@ internal List<CompletionResult> GetResultHelper(CompletionContext completionCont
...
+                        // Handle scenarios such as 'gci | Format-Table @{Label=...;<tab>'
+                        result = GetResultForHashtable(completionContext);
```
daxian-dbw: `get-something;<tab>` does go through this path. It's better to check `if (lastAst is HashtableAst)` before calling `GetResultForHashtable`.

src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
```diff
@@ -531,6 +538,18 @@ internal List<CompletionResult> GetResultHelper(CompletionContext completionCont
...
+                                if (lastAst.Parent is DynamicKeywordStatementAst)
+                                {
+                                    // Handle scenarios such as 'configuration foo { File ab { Attributes ='
+                                    bool unused;
+                                    result = GetResultForEnumPropertyValueOfDSCResource(completionContext, string.Empty, ref replacementIndex, ref replacementLength, out unused);
+                                    break;
+                                }
+                                // Handle scenarios such as 'gci | Format-Table @{Label=<tab>'
+                                return null;
+                            }
```
daxian-dbw: This duplicates the call to `GetResultForEnumPropertyValueOfDSCResource`.
How about changing to this:
```
case TokenKind.LParen:
    {
        if (lastAst is AttributeAst)
        {
            ...
        }
        else if (lastAst.Parent is DynamicKeywordStatementAst)  // <-- change here
        {
            ...
        }
        else if (lastAst is HashtableAst hashTableAst && CheckForPendingAssignment(hashTableAst)) // <-- change here
        {
            return result;
        }
        break;
    }
```

test/powershell/Host/TabCompletion/BugFix.Tests.ps1
```diff
@@ -65,4 +65,19 @@ Describe "Tab completion bug fix" -Tags "CI" {
...
+        $result.CompletionMatches.Count | Should -Be 0
+    }
+
+    It "Issue#5322 - 'Get-Date | Sort-Object @{Expression=...;' should work" {
+        $cmd = "Get-Date | Sort-Object @{Expression=...;"
+        $result = TabExpansion2 -inputScript $cmd -cursorColumn $cmd.Length
+        $result.CompletionMatches.Count | Should -Be 3
+        $result.CompletionMatches[0].CompletionText | Should -Be 'Expression'
+        $result.CompletionMatches[1].CompletionText | Should -Be 'Ascending'
+        $result.CompletionMatches[2].CompletionText | Should -Be 'Descending'
```
daxian-dbw: This test succeeds even without changes in this PR. You need to check `ReplacementIndex` and `ReplacementLength`, as they were -1 previously and changes in this PR should fix them.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
iSazonov: Second condition doesn't work for `configuration foo { File ab { Attributes =(` test - really `lastAst.Parent` is `CommandExpressionAst`.

Fixed.

src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
iSazonov: Fixed.

test/powershell/Host/TabCompletion/BugFix.Tests.ps1
iSazonov: Fixed.

src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
```diff
@@ -537,10 +547,10 @@ internal List<CompletionResult> GetResultHelper(CompletionContext completionCont
                                 completionContext.ReplacementLength = replacementLength = 0;
                                 result = GetResultForAttributeArgument(completionContext, ref replacementIndex, ref replacementLength);
                             }
-                            else
+                            else if (lastAst.Parent is DynamicKeywordStatementAst || lastAst.Parent is CommandExpressionAst)
                             {
-                                //
-                                // Handle auto completion for enum/dependson property of DSC resource,
+                                // 1. Handle scenarios such as 'configuration foo { File ab { Attributes ='
```
daxian-dbw: > Handle auto completion for **enum/dependson** property of DSC resource

The example `configuration foo { File ab { Attributes =` is a completion for enum property, so it should not be marked as `1.`.

src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
```diff
@@ -537,10 +547,10 @@ internal List<CompletionResult> GetResultHelper(CompletionContext completionCont
                                 completionContext.ReplacementLength = replacementLength = 0;
                                 result = GetResultForAttributeArgument(completionContext, ref replacementIndex, ref replacementLength);
                             }
-                            else
+                            else if (lastAst.Parent is DynamicKeywordStatementAst || lastAst.Parent is CommandExpressionAst)
```
daxian-dbw: We are making two special checks here. Maybe we should only have one special check, like the following. What do you think?

```c#
if (lastAst is AttributeAst)
{
    // the original attribute handling
}
else if (lastAst is HashtableAst hashTableAst &&
         !(lastAst.Parent is DynamicKeywordStatementAst) &&
         CheckForPendingAssignment(hashTableAst))
{
    return null;
}
else
{
    // the original else block
}
```

test/powershell/Host/TabCompletion/BugFix.Tests.ps1
```diff
@@ -65,4 +65,21 @@ Describe "Tab completion bug fix" -Tags "CI" {
...
+        $cmd = "Get-Date | Sort-Object @{Expression="
+        $result = TabExpansion2 -inputScript $cmd -cursorColumn $cmd.Length
+        $result.CompletionMatches.Count | Should -Be 0
+    }
+
+    It "Issue#5322 - 'Get-Date | Sort-Object @{Expression=...;' should work" {
+        $cmd = "Get-Date | Sort-Object @{Expression=...;"
+        $result = TabExpansion2 -inputScript $cmd -cursorColumn $cmd.Length
+        $result.CurrentMatchIndex | Should -Be -1
+        $result.ReplacementIndex | Should -Be 40
```
daxian-dbw: And `ReplacementLength`?

test/powershell/Host/TabCompletion/BugFix.Tests.ps1
```diff
@@ -65,4 +65,21 @@ Describe "Tab completion bug fix" -Tags "CI" {
...
+        $result.CompletionMatches[2].CompletionText | Should -Be 'Descending'
+        $result.CompletionMatches[0].CompletionText | Should -Be 'Expression'
+        $result.CompletionMatches[1].CompletionText | Should -Be 'Ascending'
+        $result.CompletionMatches[2].CompletionText | Should -Be 'Descending'
```
daxian-dbw: `$result.CompletionMatches[2].CompletionText` is checked twice.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
iSazonov: Fixed.

test/powershell/Host/TabCompletion/BugFix.Tests.ps1
iSazonov: Fixed.

test/powershell/Host/TabCompletion/BugFix.Tests.ps1
iSazonov: It was for ReplacementLength :-)

Fixed.

src/System.Management.Automation/engine/CommandCompletion/CompletionAnalysis.cs
iSazonov: Looks good. Thanks!

Fixed.

iSazonov:
The while code is very tricky and sensetive :confused: and crush PowerShell if something wrong. Maybe we should make it more robust.

daxian-dbw (approved):
LGTM.
@iSazonov What `while` code were you referring to?

iSazonov:
I means our Completion code - if I return wrong index PowerShell is crashed. Asts can be combined in many ways. Can we be sure that we tested all their? The code is complex and tricky - can we create simple algorithm? Using visitors?

iSazonov:
@lzybkr @SteveL-MSFT Do you have any comments?

daxian-dbw:
@iSazonov Sorry for the late reply. I agree the tab completion code is a bit hard to organize now. Many changes have been made that are very specific to the scenarios they try to solve, so you can see checks in `CompletionAnalysis` that are tied to one particular scenario which is hard to understand just by reading the code.

However, the crash is possibly not related to tab completion, but the code rending results to console. If you mess up the index, the rendering code might do weird bad things. I don't think a visitor can help much here, as many checks need to inspect ancestors of an ast.

iSazonov:
@daxian-dbw Thanks for great comment!

PR #6523, opened 2018-03-28
Clean up workflow logic in the module loading component

daxian-dbw:
## PR Summary

Clean up workflow logic in the module loading component.
Workflow module is not supported in PSCore.
Currently, `Import-Module` throws a **terminating** error when seeing `.xaml` modules. After the change, `Import-Module` throws a **non-terminating** error when seeing `.xaml` modules.

## PR Checklist

- [x] [PR has a meaningful title](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - Use the present tense and imperative mood when describing your changes
- [x] [Summarized changes](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [x] [Change is not breaking](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes)
- [x] [Make sure all `.h`, `.cpp`, `.cs`, `.ps1` and `.psm1` files have the correct copyright header](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [x] This PR is ready to merge and is not [Work in Progress](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---work-in-progress).
    - If the PR is work in progress, please add the prefix `WIP:` to the beginning of the title and remove the prefix when the PR is ready.
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - [x] User-facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
        - [ ] Issue filed - Issue link:
- **Testing - New and feature**
    - [ ] Not Applicable or can only be tested interactively
    - **OR**
    - [x] [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
        - [x] [Add `[feature]` if the change is significant or affects feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

PR #6843, opened 2018-05-08
Fix a module-loading regression that caused an infinite loop

### LINKED ISSUE #6813: Process is terminating due to StackOverflowException
Since installing 6.1.0-preview.2 I get an StackOverflowException soon after I start PowerShell. This occurs on Mac OS X and Windows 10. This did not occur on 6.1.0-preview.1.

I suspect it has to do with my Module `OnCommand-Insight` as the issue did not occur when the module was not installed.

Steps to reproduce
------------------

Install 6.1.0-preview.2, then

```powershell
Install-Module -Name "OnCommand-Insight" -Scope CurrentUser
```

Restart PowerShell

```shell
Florians-MacBook-Pro:~ florianfeldhaus$ pwsh
PowerShell v6.1.0-preview.2

https://aka.ms/pscore6-docs
Type 'help' to get help.

PS /Users/florianfeldhaus> Process is terminating due to StackOverflowException.
Abort trap: 6

```

Environment data
----------------

Mac OS X

```powershell
PSVersion                      6.1.0-preview.2
GitCommitId                    v6.1.0-preview.2
OS                             Darwin 17.5.0 Darwin Kernel Version 17.5.0: Mon Mar  5 22:24:32 PST 2018; root:xnu-4570.51.1~1/RELEASE_X86_64
```

On Windows I was not fast enough, thus I edited the output from the 6.0.0 installation which is running fine:

```powershell
PS C:\Users\Florian\Documents\GitHub> $PSVersionTable

PSVersion                      6.1.0-preview.2
GitCommitId                    v6.1.0-preview.2
```

daxian-dbw:
## PR Summary

Fix #6813

This regression was introduced by #6523, in [`PSModuleInfo.cs`](https://github.com/PowerShell/PowerShell/pull/6523/files#diff-9bbf82b18c0321d75bdb0e7693a44040L582). A circular nested module check was removed because the comment there suggested it happens only with a deprecated workflow module. This causes a `StackOverflow` exception when running into circular nested modules.

Circular nested modules could happen for a module that is not well structured. For example, the module folder `test` contains two files: `test.psd1` and `test.psm1`, and `test.psd1` has the following content:
```
@{ ModuleVersion = '0.0.1'; RootModule = 'test'; NestedModules = @('test') }
```
The same value `test` is put in both RootModule and NestedModule, which will end up with a module whose nested module points to itself.

There are two changes in this PR:
1. Add back the circular nested module check in `PSModuleInfo.cs`.
2. Remove a wrong `Dbg.Assert` in `ModuleCmdletBase.cs` and two checks before it.
   - For the assertion `Dbg.Assert(newManifestInfo.SessionState == ss`, when facing the example above, the nested module will first be loaded with a different session state, and then when trying to load the root module, the same loaded nested module will be reused for it. So 'newManifestInfo.SessionState' is not `ss`. The assersion will fail in that case.
   - For the two checks before the assersion, they are not needed anymore based on the comments there. 

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**


src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -2976,30 +2976,16 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
                     BaseCmdletPatterns = oldCmdletPatterns;
                 }
 
-                // If there is an existing session state and the new module info
-                // session state is empty, then use the existing session state.
-                // This will be the case when ModuleToProcess is a binary module
-                // and there were nested modules.
-                if (newManifestInfo.SessionState == null && ss != null)
```
daxian-dbw: When it's not module analysis (get-module -list), `ss` will not be null. I verified that binary root module also use `ss` as the session state.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -2976,30 +2976,16 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
...
-                {
-                    newManifestInfo.SessionState = ss;
-                    ss.Internal.Module = newManifestInfo;
-                }
-                // If the new module info session state is not empty but there is no
-                // existing session state, then use the new module info's sessions state;
-                // This will be the case if the module to process is a script module
-                // but there where no nested modules.
-                else if (newManifestInfo.SessionState != null && ss == null)
```
daxian-dbw: `ss == null` only if we are doing module analysis, but when we are doing module analysis, `newManifestInfo.SessionState` is always `null`, because we don't really load anything.

rjmholt (approved):
These changes look good to me.

One question is: is there a possibility of a second-order nested module cycle?

Say A is a top level module, with nested module B, and B has in *its* nested modules A, then we'll still get caught in an infinite recursion? Unless that's not possible.

If it is possible, it seems like the only way to fix it is to carry a `HashSet` of module infos that we've seen with us down the recursive calls and ignore anything that's already in it. That would also prevent us reanalysing shared nested modules.

src/System.Management.Automation/engine/Modules/PSModuleInfo.cs
```diff
@@ -577,6 +577,15 @@ private bool IsScriptModuleFile(string path)
             var res = new Dictionary<string, TypeDefinitionAst>(StringComparer.OrdinalIgnoreCase);
             foreach (var nestedModule in this.NestedModules)
             {
+                if (nestedModule == this)
+                {
+                    // Circular nested modules could happen with ill-organized module structure.
+                    // For example, module folder 'test' has two files: 'test.psd1' and 'test.psm1', and 'test.psd1' has the following content:
+                    //    "@{ ModuleVersion = '0.0.1'; RootModule = 'test'; NestedModules = @('test') }"
+                    // Then, 'Import-Module test.psd1 -PassThru' will return a ModuleInfo object with circular nested modules.
+                    continue;
```
iSazonov: Based on Issue which we fix I believe it make sense to inform users about the situation or even terminated the cmdlet.

test/powershell/engine/Module/Module.Basic.Tests.ps1
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+Describe "Module basic tests" -tags "CI" {
```
iSazonov: Maybe put the _single_ test in `test\powershell\Modules\Microsoft.PowerShell.Core\Import-Module.Tests.ps1`?

src/System.Management.Automation/engine/Modules/PSModuleInfo.cs
daxian-dbw: Turning something that works to an error is a hard breaking change. As for the warning, I'm not sure if we should do it, since technically import-module is doing what it's asked to do ... (have itself as the nested module).

daxian-dbw:
@rjmholt I tried out the scenario you mentioned above and powershell handles it (yay!):
```
PS:7> Import-Module mtest
Import-Module : Cannot load the module 'F:\tmp\dump\mtest\blah' because the module nesting limit has been exceeded. Modules can only be nested to 10 levels. Evaluate and change the order in which you are loading modules to prevent exceeding the nesting limit, and then try running your script again.
At line:1 char:1
+ Import-Module mtest
+ ~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : InvalidOperation: (F:\tmp\dump\mtest\blah:String) [Import-Module], InvalidOperationException
+ FullyQualifiedErrorId : Modules_ModuleTooDeeplyNested,Microsoft.PowerShell.Commands.ImportModuleCommand
```

src/System.Management.Automation/engine/Modules/PSModuleInfo.cs
iSazonov: Can ScriptAnalyzer catch this?

iSazonov (approved):

src/System.Management.Automation/engine/Modules/PSModuleInfo.cs
daxian-dbw: Issue opened: https://github.com/PowerShell/PSScriptAnalyzer/issues/997

PR #6890, opened 2018-05-17
Allow Basic Auth over HTTPS

### LINKED ISSUE #6779: Explicitly disallow Basic Auth over HTTP with PSRP on Linux
A security review of the proposed fix for https://github.com/PowerShell/psl-omi-provider/issues/114 (Basic Auth over HTTP does not work from PowerShell) has determined that Basic Auth over HTTP should not be supported Linux. the primary reasons include:

1: Both credentials and payload are sent un-encrypted.
2: Neither the client nor the server (omi) provide a method for disabling this combination.
3: It's too easy for users to use PSRP in an insecure manner.

As the above issue implies, this combination is currently broken in PowerShell on Linux. Instead of fixing it, the proposed change is to explicitly disable it and report a more appropriate error message.

The recommendation going forward will be to use Negotiate which results in the following:

From the OMI/Setup NTLM doc... 
- The password is hashed using an irreversable algorithm, so the password is more secure than Basic.
 - The client and server support encryption of your data over http connections, so SSL certificates are not required.

dantraMSFT:
## PR Summary

Fix https://github.com/PowerShell/PowerShell/issues/6779 regression.

The change to disallow Basic Auth over HTTP was not checking the scheme for the -Uri parameter. The result is Basic Auth connections over HTTPS via the -Uri parameter were disallowed.

The change adds the check for the scheme as well as useSSL.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
+++ b/src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
@@ -1528,9 +1528,8 @@ namespace System.Management.Automation.Remoting.Client
             }
 
 #if UNIX
-            bool isHttps = StringComparer.OrdinalIgnoreCase.Compare("https", connectionUri.Scheme) == 0;
             // explicitly disallow Basic auth over HTTP on Unix.
-            if (connectionInfo.AuthenticationMechanism == AuthenticationMechanism.Basic && !isSSLSpecified && !isHttps)
+            if (connectionInfo.AuthenticationMechanism == AuthenticationMechanism.Basic && !isSSLSpecified && connectionUri.Scheme != Uri.UriSchemeHttps)
             {
                 throw new PSRemotingTransportException(PSRemotingErrorId.ConnectFailed, RemotingErrorIdStrings.BasicAuthOverHttpNotSupported);
             }
--- a/src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
+++ b/src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
@@ -1528,8 +1528,9 @@ namespace System.Management.Automation.Remoting.Client
             }
 
 #if UNIX
+            bool isHttps = StringComparer.OrdinalIgnoreCase.Compare("https", connectionUri.Scheme) == 0;
             // explicitly disallow Basic auth over HTTP on Unix.
-            if (connectionInfo.AuthenticationMechanism == AuthenticationMechanism.Basic && !isSSLSpecified)
+            if (connectionInfo.AuthenticationMechanism == AuthenticationMechanism.Basic && !isSSLSpecified && !isHttps)
             {
                 throw new PSRemotingTransportException(PSRemotingErrorId.ConnectFailed, RemotingErrorIdStrings.BasicAuthOverHttpNotSupported);
             }
--- a/test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
+++ b/test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
@@ -25,7 +25,7 @@ Describe "Basic Auth over HTTP not allowed on Unix" -Tag @("CI") {
         $credential = [PSCredential]::new('username', $password)
 
         # use a Uri that specifies HTTPS to test Basic Auth logic.
-        # NOTE: The connection is expected to fail but not with a  PSSessionOpenFailed exception
+        # NOTE: The connection is expected to fail but not with a  ConnectFailed exception
         $uri = "https://localhost"
         New-PSSession -Uri $uri -Credential $credential -Authentication Basic -ErrorVariable err
         $err.Exception | Should -BeOfType [System.Management.Automation.Remoting.PSRemotingTransportException]
--- a/test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
+++ b/test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
@@ -20,7 +20,7 @@ Describe "Basic Auth over HTTP not allowed on Unix" -Tag @("CI") {
         $err.Exception.ErrorCode | Should -Be 801
     }
 
-   # Marked as pending due tohttps://github.com/Microsoft/omi/issues/502
+   # Marked as pending due to https://github.com/Microsoft/omi/issues/502
    # It "New-PSSession should NOT throw a ConnectFailed exception when specifying Basic Auth over HTTPS on Unix" -skip:($IsWindows) {
     It "New-PSSession should NOT throw a ConnectFailed exception when specifying Basic Auth over HTTPS on Unix" -Pending {
         $password = ConvertTo-SecureString -String "password" -AsPlainText -Force
--- a/test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
+++ b/test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
@@ -16,10 +16,22 @@ Describe "Basic Auth over HTTP not allowed on Unix" -Tag @("CI") {
         $err = ({New-PSSession -ComputerName 'localhost' -Credential $credential -Authentication Basic}  | Should -Throw -PassThru  -ErrorId 'System.Management.Automation.Remoting.PSRemotingDataStructureException,Microsoft.PowerShell.Commands.NewPSSessionCommand')
         $err.Exception | Should -BeOfType [System.Management.Automation.Remoting.PSRemotingTransportException]
         # Should be PSRemotingErrorId.ConnectFailed
-        # Ensures we are looking at teh expected instance
+        # Ensures we are looking at the expected instance
         $err.Exception.ErrorCode | Should -Be 801
     }
 
+    It "New-PSSession should NOT throw a ConnectFailed exception when specifying Basic Auth over HTTPS on Unix" -skip:($IsWindows) {
+        $password = ConvertTo-SecureString -String "password" -AsPlainText -Force
+        $credential = [PSCredential]::new('username', $password)
+
+        # use a Uri that specifies HTTPS to test Basic Auth logic.
+        # NOTE: The connection is expected to fail but not with a  PSSessionOpenFailed exception
+        $uri = "https://localhost"
+        New-PSSession -Uri $uri -Credential $credential -Authentication Basic -ErrorVariable err
+        $err.Exception | Should -BeOfType [System.Management.Automation.Remoting.PSRemotingTransportException]
+        $err.FullyQualifiedErrorId | Should -Be '1,PSSessionOpenFailed'
+        $err.Exception.HResult | Should -Be -2146233087
+    }
 }
 
 Describe "JEA session Transcript script test" -Tag @("Feature", 'RequireAdminOnWindows') {
--- a/test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
+++ b/test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
@@ -32,7 +32,7 @@ Describe "Basic Auth over HTTP not allowed on Unix" -Tag @("CI") {
         New-PSSession -Uri $uri -Credential $credential -Authentication Basic -ErrorVariable err
         $err.Exception | Should -BeOfType [System.Management.Automation.Remoting.PSRemotingTransportException]
         $err.FullyQualifiedErrorId | Should -Be '1,PSSessionOpenFailed'
-        $err.Exception.HResult | Should -Be -2146233087
+        $err.Exception.HResult | Should -Be 0x80131501
     }
 }
 
--- a/test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
+++ b/test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
@@ -20,7 +20,9 @@ Describe "Basic Auth over HTTP not allowed on Unix" -Tag @("CI") {
         $err.Exception.ErrorCode | Should -Be 801
     }
 
-    It "New-PSSession should NOT throw a ConnectFailed exception when specifying Basic Auth over HTTPS on Unix" -skip:($IsWindows) {
+   # Marked as pending due tohttps://github.com/Microsoft/omi/issues/502
+   # It "New-PSSession should NOT throw a ConnectFailed exception when specifying Basic Auth over HTTPS on Unix" -skip:($IsWindows) {
+    It "New-PSSession should NOT throw a ConnectFailed exception when specifying Basic Auth over HTTPS on Unix" -Pending {
         $password = ConvertTo-SecureString -String "password" -AsPlainText -Force
         $credential = [PSCredential]::new('username', $password)
 


src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
```diff
@@ -1528,8 +1528,9 @@ private void Initialize(Uri connectionUri, WSManConnectionInfo connectionInfo)
             }
 
 #if UNIX
+            bool isHttps = StringComparer.OrdinalIgnoreCase.Compare("https", connectionUri.Scheme) == 0;
             // explicitly disallow Basic auth over HTTP on Unix.
-            if (connectionInfo.AuthenticationMechanism == AuthenticationMechanism.Basic && !isSSLSpecified)
+            if (connectionInfo.AuthenticationMechanism == AuthenticationMechanism.Basic && !isSSLSpecified && !isHttps)
```
SteveL-MSFT: I think you can get rid of the `isHttps` variable and just use:

```c#
connectionUri.Scheme != Uri.UriSchemeHttps
```

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
```diff
@@ -20,10 +20,22 @@ Describe "Basic Auth over HTTP not allowed on Unix" -Tag @("CI") {
         $err = ({New-PSSession -ComputerName 'localhost' -Credential $credential -Authentication Basic}  | Should -Throw -PassThru  -ErrorId 'System.Management.Automation.Remoting.PSRemotingDataStructureException,Microsoft.PowerShell.Commands.NewPSSessionCommand')
         $err.Exception | Should -BeOfType [System.Management.Automation.Remoting.PSRemotingTransportException]
         # Should be PSRemotingErrorId.ConnectFailed
-        # Ensures we are looking at teh expected instance
+        # Ensures we are looking at the expected instance
         $err.Exception.ErrorCode | Should -Be 801
     }
 
+    It "New-PSSession should NOT throw a ConnectFailed exception when specifying Basic Auth over HTTPS on Unix" -skip:($IsWindows) {
+        $password = ConvertTo-SecureString -String "password" -AsPlainText -Force
```
SteveL-MSFT: In the `HelpersRemoting.psm1` module, I believe you can get real creds to validate it works.

SteveL-MSFT (changes requested):

src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
dantraMSFT: @SteveL-MSFT: does connectionUri guarantee that it normalizes the case of the Scheme property?

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
dantraMSFT: I actually don't want to establish a connection; I'm verifying that Basic Auth verification doesn't reject the request.

src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
iSazonov: I believe @SteveL-MSFT means:
```c#
if (connectionInfo.AuthenticationMechanism == AuthenticationMechanism.Basic && !isSSLSpecified && StringComparer.OrdinalIgnoreCase.Compare("https", connectionUri.Scheme) != 0)

File: test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1 [hunk 3]
```diff
@@ -20,10 +20,22 @@ Describe "Basic Auth over HTTP not allowed on Unix" -Tag @("CI") {
...
+        $credential = [PSCredential]::new('username', $password)
+
+        # use a Uri that specifies HTTPS to test Basic Auth logic.
+        # NOTE: The connection is expected to fail but not with a  PSSessionOpenFailed exception
+        $uri = "https://localhost"
+        New-PSSession -Uri $uri -Credential $credential -Authentication Basic -ErrorVariable err
+        $err.Exception | Should -BeOfType [System.Management.Automation.Remoting.PSRemotingTransportException]
+        $err.FullyQualifiedErrorId | Should -Be '1,PSSessionOpenFailed'
```
iSazonov: Comment in Line 32 says "not with a  PSSessionOpenFailed". Seems the comment should be corrected (or the test is wrong?).

src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
markekraus: @SteveL-MSFT is correct. `Uri.Scheme` is always normalized to lowercase (see the notes on the property value here https://docs.microsoft.com/en-us/dotnet/api/system.uri.scheme?view=netcore-2.1)

Line 1531 is unneeded and on line 1533 `&& !isHttps` can be changed to `&& connectionUri.Scheme != Uri.UriSchemeHttps`

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
SteveL-MSFT: Is there a separate test that validates it works?

dantraMSFT: @SteveL-MSFT: No, we don't have any automated remoting tests for PSRP at this point.

src/System.Management.Automation/engine/remoting/fanin/WSManTransportManager.cs
dantraMSFT: Fixed

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
dantraMSFT: Fixed

markekraus (approved):

dantraMSFT:
The test failure appears to be a variation of https://github.com/Microsoft/omi/issues/508 that is not covered by the fix.

This PR cannot be merged until the libmi issue is fixed since the new test hits this path. It appears to hit sporadically in release builds and consistently in debug builds.

SteveL-MSFT (approved):

echobb8:
What's the status on this? I need to connect to Office 365 from my Mac, but it fails: "New-PSSession : Basic authentication is not supported over HTTP on Unix." 

Is there a workaround?

vors:
> Is there a workaround?

Use PS version before 6.1.0-Preview3, i.e. 6.1.0-Preview2
https://github.com/PowerShell/PowerShell/releases

dantraMSFT:
I will see if I can get the PR through today with the failing test marked as pending.

dantraMSFT:
I removed the blocking label and disabled the test that hits the libmi issue (https://github.com/Microsoft/omi/issues/502 )

dantraMSFT:
@adityapatwardhan Fixed

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
```diff
@@ -16,10 +16,24 @@ Describe "Basic Auth over HTTP not allowed on Unix" -Tag @("CI") {
         $err = ({New-PSSession -ComputerName 'localhost' -Credential $credential -Authentication Basic}  | Should -Throw -PassThru  -ErrorId 'System.Management.Automation.Remoting.PSRemotingDataStructureException,Microsoft.PowerShell.Commands.NewPSSessionCommand')
         $err.Exception | Should -BeOfType [System.Management.Automation.Remoting.PSRemotingTransportException]
         # Should be PSRemotingErrorId.ConnectFailed
-        # Ensures we are looking at teh expected instance
+        # Ensures we are looking at the expected instance
         $err.Exception.ErrorCode | Should -Be 801
     }
 
+   # Marked as pending due tohttps://github.com/Microsoft/omi/issues/502
```
adityapatwardhan: nitpick: space between `to` and `https`

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
```diff
@@ -16,10 +16,24 @@ Describe "Basic Auth over HTTP not allowed on Unix" -Tag @("CI") {
...
+   # It "New-PSSession should NOT throw a ConnectFailed exception when specifying Basic Auth over HTTPS on Unix" -skip:($IsWindows) {
+    It "New-PSSession should NOT throw a ConnectFailed exception when specifying Basic Auth over HTTPS on Unix" -Pending {
+        $password = ConvertTo-SecureString -String "password" -AsPlainText -Force
+        $credential = [PSCredential]::new('username', $password)
+
+        # use a Uri that specifies HTTPS to test Basic Auth logic.
+        # NOTE: The connection is expected to fail but not with a  ConnectFailed exception
+        $uri = "https://localhost"
+        New-PSSession -Uri $uri -Credential $credential -Authentication Basic -ErrorVariable err
```
adityapatwardhan: Can you use the pattern below?

```PowerShell
{ New-PSSession -Uri $uri -Credential $credential -Authentication Basic -ErrorVariable err } | Should -Throw -ErrorId '1,PSSessionOpenFailed' -ExceptionType 'System.Management.Automation.Remoting.PSRemotingTransportException'
$err.Exception.HResult | Should -Be -2146233087
```

adityapatwardhan (changes requested):

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
dantraMSFT: No, when -ErrorVariable is used, pester doesn't have an exception to catch.

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
dantraMSFT: fixed

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
adityapatwardhan: This cannot be done as when `err` is set no exception is thrown.

adityapatwardhan (approved):

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
SteveL-MSFT: If this is a negative test case where you need to do more validation of the exception, you should use the pattern in our test guidance doc with `-PassThru`: https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/WritingPesterTests.md#writing-pester-tests

SteveL-MSFT (changes requested):

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
dantraMSFT: @SteveL-MSFT: new-pssession doesn't throw, it writes to the error stream.

dantraMSFT: @SteveL-MSFT: do you have any additional feedback?

kai-h:
Can we please explicitly have a test to connect to Office 365, or even Azure.
We just need to test that the connection attempt succeeds, it doesn't need to be given valid credentials as long as it doesn't throw a ConnectFailed error.
The remote session can be accessed via the following Connection URI for Office 365:
```PowerShell
-ConnectionUri https://outlook.office365.com/powershell-liveid/
```

iSazonov:
> Can we please explicitly have a test to connect to Office 365

All tests using external resources is unstable. :confused: and daily/night builds frequently failed.

Perhaps for the cases we could have optional test set in separate job.

kai-h:
I suppose that the main thing is that PowerShell actually attempts to connect to a remote session, be it locally or on an actual remote system.
It would be wonderful however if the testing was able to qualify that it will connect successfully to Office 365, Azure or some other Microsoft endpoint that customers will be using PowerShell to manage.

dantraMSFT:
@kai-h I will be creating tests that verify we can successfully connect to o365 but it won't be in the nightly build.  One reason is mentioned above but the primary one, for now,  is I don't want to expose o365 account information; even if only for testing purposes.

The tests will be checked in but will likely be conditioned on a custom JSON file that is not present in the public CI builds.

kai-h:
@dantraMSFT - Thanks, I understand given the reasons above.
In my testing, I've found that it's not necessary to supply valid credentials in testing if PowerShell can connect to Office 365. Even if you're supplying invalid credentials (e.g. user@example.com with password Password123), if PowerShell can at least initiate the connection attempt to Office 365, even if Office365 says invalid username or password, this shows that PowerShell can connect, however it can't test if any remote commands can be executed.

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
```diff
@@ -16,10 +16,24 @@ Describe "Basic Auth over HTTP not allowed on Unix" -Tag @("CI") {
...
+        $err.Exception | Should -BeOfType [System.Management.Automation.Remoting.PSRemotingTransportException]
+        $err.FullyQualifiedErrorId | Should -Be '1,PSSessionOpenFailed'
+        $err.Exception.HResult | Should -Be -2146233087
```
SteveL-MSFT: Nit: prefer the hex version particularly with a HResult:  0x80131501

SteveL-MSFT (changes requested):

BrucePay (approved):

adityapatwardhan:
@dantraMSFT  Can you respond to @SteveL-MSFT  comment.

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
dantraMSFT: Fixed

dantraMSFT:
@adityapatwardhan Addressed Steve's comment.

test/powershell/engine/Remoting/RemoteSession.Basic.Tests.ps1
SteveL-MSFT: I think you should follow the convention for now to use `-ErrorAction Stop` and then use Pester's `Should -ThrowErrorId -PassThru` to do additional validation for consistency until we have capability to validate non-terminating errors

SteveL-MSFT (changes requested):

dantraMSFT:
@SteveL-MSFT Hmm, that's not how I interpreted the guidelines. 
That said, forcing a throw isn't testing the cmdlet the way it's normally used. Additionally, using the guideline prevents testing non-terminating errors.

SteveL-MSFT:
@dantraMSFT let's defer this to the other discussion on validating terminating vs non-terminating.  I'll approve this PR as this isn't blocking.

SteveL-MSFT (approved):

PR #6898, opened 2018-05-18
Make Select/ForEach/Where see dynamic properties

### LINKED ISSUE #6567: The PSObject implementation does not handle classes derived from DynamicObject properly
This repro’s on 5.1 and 6. The problem is that if I have a DynamicObject that has a dynamic property “Foo”, things like `$obj.Foo` work, but `Select-Object -ExpandProperty Foo` do not.

Steps to reproduce
------------------

```powershell
Add-Type @"
using System;
using System.Collections.Generic;
using System.Dynamic;

public class MyDynamicObject : DynamicObject
{
    public override IEnumerable<string> GetDynamicMemberNames()
    {
        return new string[] { "Foo" };
    }

    public override bool TryGetMember( GetMemberBinder binder, out object result )
    {
        result = null;
        bool succeeded = false;

        if( binder.Name == "Foo" )
        {
            result = 123;
            succeeded = true;
        }

        return succeeded;
    }
}
"@

$myObj = [MyDynamicObject]::new()

$myObj.Foo                                  # <-- works
$myObj | %{ $_.Foo }                        # <-- works
$myObj | Select-Object -ExpandProperty Foo  # <-- DOES NOT WORK
```

Expected behavior
-----------------

```none
The last three lines should all work (yield 123).
```

Actual behavior
---------------

```none
The Select-Object statement throws. ("Select-Object : Property "Foo" cannot be found.")
```

Environment data
----------------

```powershell
> $PSVersionTable
PSVersion                      6.1.0-preview.1
GitCommitId                    v6.1.0-preview.1
```

jazzdelightsme:
## PR Summary

Dynamic (DLR) objects work in some places today, but not others.  This
change expands that support to ForEach-Object, Where-Object and the
family of cmdlets that use MshExpression (Select-Object, etc.). (see the
linked issue for the simple repro and explanation: fixes #6567)

This change addresses both wildcard and non-wildcard cases. In wildcard
cases, it uses the existing support of generating PSDynamicMember
objects for names returned by GetDynamicMemberNames.

In non-wildcard cases, a dynamic property access is attempted whether or
not the name shows up in GetDynamicMemberNames, but a truly "blind"
access is only attempted if we see that the base object is an IDMOP. If
the dynamic access fails, you'll get the same or a similar error
experience as before ("The property 'Blarg' cannot be found", or no
error at all, depending on the cmdlet and the strict mode setting).

The included test coverage includes a stub for the .ForEach
operator--once people are happy with this change, I can continue by
adding support there.

This change should allegedly also have positive perf impact, though I
have not tested that assertion.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
@@ -4,9 +4,11 @@
 using System;
 using System.Management.Automation;
 using System.Management.Automation.Internal;
+using System.Management.Automation.Language;
 using System.Collections;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Runtime.CompilerServices;
 
 namespace Microsoft.PowerShell.Commands.Internal.Format
 {
@@ -123,13 +125,20 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             {
                 // get the members first: this will expand the globbing on each parameter
                 members = target.Members.Match(_stringValue,
-                                            PSMemberTypes.Properties | PSMemberTypes.PropertySet);
+                                            PSMemberTypes.Properties | PSMemberTypes.PropertySet | PSMemberTypes.Dynamic);
             }
             else
             {
                 // we have no globbing: try an exact match, because this is quicker.
                 PSMemberInfo x = target.Members[_stringValue];
 
+                if ((x == null) && (target.BaseObject is System.Dynamic.IDynamicMetaObjectProvider))
+                {
+                    // We could check if GetDynamicMemberNames includes the name...  but
+                    // GetDynamicMemberNames is only a hint, not a contract, so we'd want
+                    // to attempt the binding whether it's in there or not.
+                    x = new PSDynamicMember(_stringValue);
+                }
                 List<PSMemberInfo> temp = new List<PSMemberInfo>();
                 if (x != null)
                 {
@@ -165,10 +174,14 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                             }
                         }
                     }
-                    continue;
                 }
                 // it can be a property
-                if (member is PSPropertyInfo)
+                else if (member is PSPropertyInfo)
+                {
+                    temporaryMemberList.Add(member);
+                }
+                // it can be a dynamic member
+                else if (member is PSDynamicMember)
                 {
                     temporaryMemberList.Add(member);
                 }
@@ -225,11 +238,13 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
 
         #region Private Members
 
+        private CallSite<Func<CallSite, object, object>> _getValueDynamicSite;
+
         private MshExpressionResult GetValue(PSObject target, bool eatExceptions)
         {
             try
             {
-                object result;
+                object result = null;
 
                 if (Script != null)
                 {
@@ -243,12 +258,16 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
                 }
                 else
                 {
-                    PSMemberInfo member = target.Properties[_stringValue];
-                    if (member == null)
+                    if (_getValueDynamicSite == null)
                     {
-                        return new MshExpressionResult(null, this, null);
+                        _getValueDynamicSite =
+                            CallSite<Func<CallSite, object, object>>.Create(
+                                    PSGetMemberBinder.Get(
+                                        _stringValue,
+                                        classScope: (Type) null,
+                                        @static: false));
                     }
-                    result = member.Value;
+                    result = _getValueDynamicSite.Target.Invoke(_getValueDynamicSite, target);
                 }
 
                 return new MshExpressionResult(result, this, null);
--- a/src/System.Management.Automation/engine/InternalCommands.cs
+++ b/src/System.Management.Automation/engine/InternalCommands.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Dynamic;
+using System.Linq;
 using System.Linq.Expressions;
 using System.Runtime.CompilerServices;
 using System.Text;
@@ -16,6 +18,38 @@ using Dbg = System.Management.Automation.Diagnostics;
 
 namespace Microsoft.PowerShell.Commands
 {
+    /// <summary>
+    /// A thin wrapper over a property-getting Callsite, to allow reuse when possible.
+    /// </summary>
+    struct DynamicPropertyGetter
+    {
+        private CallSite<Func<CallSite, object, object>> _getValueDynamicSite;
+
+        // For the wildcard case, lets us know if we can reuse the callsite:
+        private string _lastUsedPropertyName;
+
+        public object GetValue(PSObject inputObject, string propertyName)
+        {
+            Dbg.Assert(!WildcardPattern.ContainsWildcardCharacters(propertyName), "propertyName should be pre-resolved by caller");
+
+            // If wildcards are involved, the resolved property name could potentially
+            // be different on every object... but probably not, so we'll attempt to
+            // reuse the callsite if possible.
+
+            if (!propertyName.Equals(_lastUsedPropertyName, StringComparison.OrdinalIgnoreCase))
+            {
+                _lastUsedPropertyName = propertyName;
+                _getValueDynamicSite = CallSite<Func<CallSite, object, object>>.Create(
+                        PSGetMemberBinder.Get(
+                            propertyName,
+                            classScope: (Type) null,
+                            @static: false));
+            }
+
+            return _getValueDynamicSite.Target.Invoke(_getValueDynamicSite, inputObject);
+        }
+    }
+
     #region Built-in cmdlets that are used by or require direct access to the engine.
 
     /// <summary>
@@ -136,6 +170,7 @@ namespace Microsoft.PowerShell.Commands
         }
         private string _propertyOrMethodName;
         private string _targetString;
+        private DynamicPropertyGetter _propGetter;
 
         /// <summary>
         /// The arguments passed to a method invocation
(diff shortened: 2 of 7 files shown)


test/powershell/Modules/Microsoft.PowerShell.Core/Where-Object.Tests.ps1
```diff
@@ -1,5 +1,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
+. (Join-Path -Path $PSScriptRoot -ChildPath ..\..\Add-TestDynamicType.ps1)
```
iSazonov: Please put the help code in BeforeAll block with check that the class is not already added.

jazzdelightsme:
About the CodeFactor checks: it looks like most of these are not caused by my commit (although I exacerbated an existing issue--added another small helper class to a file that already contains multiple classes). What is the strategy for dealing with these? Ignore them? Take an IOU to fix them in a separate commit?

test/powershell/Modules/Microsoft.PowerShell.Utility/Select-Object.Tests.ps1
```diff
@@ -242,6 +243,30 @@ Describe "Select-Object DRT basic functionality" -Tags "CI" {
...
+        $dynObj = [TestDynamic]::new()
+        $results = $dynObj, $dynObj | Select-Object -ExpandProperty HiddenProp
+        $results.Count | Should -Be 2
+        $results[0] | Should -Be 789
+        $results[1] | Should -Be 789
+    }
+
+    It "Select-Object should handle wildcarded dynamic (DLR) properties when hinted by GetDynamicMemberNames"{
+        $dynObj = [TestDynamic]::new()
+        $results = $dynObj, $dynObj | Select-Object -ExpandProperty Foo*
```
lzybkr: Maybe also test `*Prop` (should be error with `-ExpandProperty`, maybe also test without `-ExpandProperty`.)

src/System.Management.Automation/engine/InternalCommands.cs
```diff
@@ -351,80 +386,107 @@ protected override void ProcessRecord()
...
 
-                        if (member == null)
-                        {
-                            errorRecord = GenerateNameParameterError("Name", InternalCommandStrings.PropertyOrMethodNotFound,
-                                                                     "PropertyOrMethodNotFound", _inputObject,
-                                                                     _propertyOrMethodName);
-                        }
-                        else
+                        // member is a method
+                        if ((member != null) && (member is PSMethodInfo))
```
lzybkr: Why was the `null` test added? The `C#` operator `is` can be safely used with `null` values.

src/System.Management.Automation/engine/InternalCommands.cs
```diff
@@ -439,16 +501,26 @@ protected override void ProcessRecord()
...
-                                    catch (Exception)
+                                    catch (Exception ex)
                                     {
-                                        // When the property is not gettable or it throws an exception.
-                                        // e.g. when trying to access an assembly's location property, since dynamic assemblies are not backed up by a file,
-                                        // an exception will be thrown when accessing its location property. In this case, return null.
-                                        WriteObject(null);
+                                        if (isBlindDynamicAccess)
+                                        {
+                                            errorRecord = new ErrorRecord(ex,
```
lzybkr: I know you're not a fan of silent errors, but I think consistency here is important - you wouldn't want an error just because the object implemented `IDMOP`, e.g. if some object didn't in the past, but a new version does, scripts would break when they silently "worked" before.

src/System.Management.Automation/engine/InternalCommands.cs
```diff
@@ -16,6 +18,38 @@
...
+
+        // For the wildcard case, lets us know if we can reuse the callsite:
+        private string _lastUsedPropertyName;
+
+        public object GetValue(PSObject inputObject, string propertyName)
+        {
+            Dbg.Assert(!WildcardPattern.ContainsWildcardCharacters(propertyName), "propertyName should be pre-resolved by caller");
+
+            // If wildcards are involved, the resolved property name could potentially
+            // be different on every object... but probably not, so we'll attempt to
```
lzybkr: I don't have any intuition for how often wildcards are used to specify multiple properties, but I know *I* use them that way for formatting commands.

I'm not sure perf matters much here, but it would be good to understand the impact on something like `ps | select *mem*`.

lzybkr (changes requested):
Changes look great, just some minor feedback.

It'd be great if you can measure the improvement using dynamic sites.

test/powershell/Modules/Microsoft.PowerShell.Utility/Select-Object.Tests.ps1
jazzdelightsme: Good ideas; will do.

src/System.Management.Automation/engine/InternalCommands.cs
jazzdelightsme: D'oh! Thanks.

src/System.Management.Automation/engine/InternalCommands.cs
jazzdelightsme: The problem here is that I can't tell the difference between "it failed because the property does not exist" (let's call this case 1) and "it failed because accessing it actually threw some exception" (let's call that case 2).

Existing PowerShell behavior is different for these two cases: case 1 gets an error (which is possible because the ETS tells us up front if the property exists or not), and case 2 does not.

For non-IDMOP objects, there's no problem; but for IDMOPs, we have the chance to attempt a "blind" access, but the cost is that we must have the same response to both cases (because we cannot distinguish between the two). So we have to make a choice: we can either swallow ALL errors (including "The property 'Blarg' does not exist"), or expose them all.  

I felt like it was more important to preserve the behavior of showing "The property 'Blarg' does not exist" (case 1) errors than to suppress "FooException thrown when accessing Bloop property" (case 2) errors. This decision certainly must be influenced by my predisposition against silent errors, but actually I was going for usability: a typo or otherwise trying to access a property that does not exist is a common mistake, and generally easily fixed when presented with the error, and if I had to pick one of "error suppression" or "tell the user s/he goofed", I'd choose the latter.

What do you think?

jazzdelightsme: (I can at least add a comment about this)

lzybkr: You can make some assumptions though:
* non-dynamic members must exist
* if `GetDynamicMemberNames` says the member exists, trust it

jazzdelightsme: Yes--in fact, those are exactly the assumptions I am making: if it had shown up in `GetDynamicMemberNames`, it would show up as a dynamic member, and `isBlindDynamicAccess` would be false. (And it would be false for normal properties as well, of course.) I'll add more comments about this.

src/System.Management.Automation/engine/InternalCommands.cs
jazzdelightsme: For `select`, the wildcard gets resolved into individual `MshExpressionResult` objects, so no problems there. So this case here really is for the heterogenous-objects-with-different-matching-names case, which I think is probably a rare enough case not to care too much about (though it will work, of course).

test/powershell/Add-TestDynamicType.ps1
```diff
@@ -32,5 +44,22 @@ public class TestDynamic : DynamicObject
...
     }
+
+    public override bool TryInvokeMember(InvokeMemberBinder binder, object[] args, out object result)
+    {
+        if (binder.Name == "FooMethod")
+        {
+            result = "yes";
+            return true;
+        }
+        else if (binder.Name == "HiddenMethod") // Not presented in GetDynamicMemberNames
```
iSazonov: Please put the comment on separate line.

test/powershell/Add-TestDynamicType.ps1
```diff
@@ -32,5 +44,22 @@ public class TestDynamic : DynamicObject
...
+        {
+            result = m_serialNumber;
+            return true;
+        }
+
+        result = null;
```
iSazonov: I'd move the line to line 50 - it is more clear.

test/powershell/Modules/Microsoft.PowerShell.Utility/Select-Object.Tests.ps1
```diff
@@ -262,11 +262,28 @@ Describe "Select-Object DRT basic functionality" -Tags "CI" {
...
+        $results.Count | Should -Be 2
+        $results[0].FooProp | Should -Be 123
+        $results[0].BarProp | Should -Be 456
+        $results[1].FooProp | Should -Be 123
+        $results[1].BarProp | Should -Be 456
+    }
+
+    It "Select-Object -ExpandProperty should yield errors if multiple dynamic (DLR) properties match"{
+        $dynObj = [TestDynamic]::new()
+        $e = { $results = $dynObj, $dynObj | Select-Object -ExpandProperty *Prop -EA Stop} |
```
iSazonov: Please use full names instead of aliases like -EA.

test/powershell/Add-TestDynamicType.ps1
```diff
@@ -4,11 +4,18 @@ using System.Dynamic;
 
 public class TestDynamic : DynamicObject
 {
-    public TestDynamic() { }
+    private static int s_lastSerialNumber;
+
+    private int m_serialNumber;
```
iSazonov: We don't use "m" prefix - right pattern is `_serialNumber`.

test/powershell/Add-TestDynamicType.ps1
```diff
@@ -4,11 +4,18 @@ using System.Dynamic;
...
+
+    public TestDynamic()
+    {
+        m_serialNumber = ++s_lastSerialNumber;
+    }
 
     public override IEnumerable<string> GetDynamicMemberNames()
     {
-        return new string[] { "FooProp", "BarProp" };
+        return new string[] { "FooProp", "BarProp", "FooMethod", "SerialNumber" };
```
iSazonov: Can we make it static?

src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
```diff
@@ -123,13 +125,21 @@ internal List<MshExpression> ResolveNames(PSObject target, bool expand)
...
-                                            PSMemberTypes.Properties | PSMemberTypes.PropertySet);
+                                            PSMemberTypes.Properties | PSMemberTypes.PropertySet | PSMemberTypes.Dynamic);
             }
             else
             {
                 // we have no globbing: try an exact match, because this is quicker.
                 PSMemberInfo x = target.Members[_stringValue];
 
+                if ((x == null) &&
+                    (target.BaseObject is System.Dynamic.IDynamicMetaObjectProvider))
```
iSazonov: Please move to line 135.

src/System.Management.Automation/engine/InternalCommands.cs
```diff
@@ -351,80 +386,107 @@ protected override void ProcessRecord()
...
 
-                                PSMethodInfo targetMethod = member as PSMethodInfo;
-                                Dbg.Assert(targetMethod != null, "targetMethod should not be null here.");
-                                try
-                                {
-                                    // should process
-                                    string methodAction = String.Format(CultureInfo.InvariantCulture,
-                                        InternalCommandStrings.ForEachObjectMethodActionWithoutArguments, targetMethod.Name);
+                            PSMethodInfo targetMethod = member as PSMethodInfo;
+                            Dbg.Assert(targetMethod != null, "targetMethod should not be null here.");
```
iSazonov: Maybe throw here with user friendly message then in line 413 with standard null exception?

src/System.Management.Automation/engine/InternalCommands.cs
```diff
@@ -351,80 +386,107 @@ protected override void ProcessRecord()
...
+                            }
+                            catch (PipelineStoppedException)
+                            {
+                                // PipelineStoppedException can be caused by select-object
+                                throw;
+                            }
+                            catch (Exception ex)
+                            {
+                                MethodException mex = ex as MethodException;
+                                if (mex != null && mex.ErrorRecord != null && mex.ErrorRecord.FullyQualifiedErrorId == "MethodCountCouldNotFindBest")
```
iSazonov: Please use C# 7 syntax
```c#
if ( ex is MethodException mex ...
```

src/System.Management.Automation/engine/InternalCommands.cs
```diff
@@ -439,16 +501,53 @@ protected override void ProcessRecord()
...
+                                            // an exception will be thrown when accessing its location property. In this case, return null.
+                                            WriteObject(null);
+                                        }
                                     }
                                 }
-                            } // end of member is a property
-                        } // member is not null
+                            } // end of resolvedPropertyName != null
+                        } // end of member is a [presumed] property
                     } // no args provided
```
iSazonov: Please remove the 3 comments - modern editors work well with large files.

test/powershell/Modules/Microsoft.PowerShell.Core/Where-Object.Tests.ps1
iSazonov: Please remove. We can put the code in the file. Or in HelpCommon module.

test/powershell/Modules/Microsoft.PowerShell.Utility/Select-Object.Tests.ps1
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 . (Join-Path -Path $PSScriptRoot -ChildPath Test-Mocks.ps1)
+. (Join-Path -Path $PSScriptRoot -ChildPath ..\..\Add-TestDynamicType.ps1)
```
iSazonov: The same.

test/powershell/engine/ETS/Adapter.Tests.ps1
```diff
@@ -187,13 +187,26 @@ Describe "Adapter Tests" -tags "CI" {
                 } -PassThru -Force
             $x.ForEach(5) | Should -Be 10
         }
+
+        # Pending: https://github.com/PowerShell/PowerShell/issues/6567
+        It "ForEach magic method works for dynamic (DLR) things" -Pending:$true {
+            . (Join-Path -Path $PSScriptRoot -ChildPath ..\..\Add-TestDynamicType.ps1)
```
iSazonov: The same.

src/System.Management.Automation/engine/InternalCommands.cs
jazzdelightsme: I'm not going to change this; it's totally unrelated to my change.

src/System.Management.Automation/engine/InternalCommands.cs
jazzdelightsme: This code is also unrelated to my change (harder to tell for this one, but it's just moved a bit).

test/powershell/Modules/Microsoft.PowerShell.Core/Where-Object.Tests.ps1
```diff
@@ -1,5 +1,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
+Import-Module HelpersCommon
```
iSazonov: The module is auto-loaded. Please remove the line.

test/powershell/Modules/Microsoft.PowerShell.Utility/Select-Object.Tests.ps1
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 . (Join-Path -Path $PSScriptRoot -ChildPath Test-Mocks.ps1)
+Import-Module HelpersCommon
```
iSazonov: The same.

test/powershell/engine/ETS/Adapter.Tests.ps1
```diff
@@ -187,13 +187,27 @@ Describe "Adapter Tests" -tags "CI" {
                 } -PassThru -Force
             $x.ForEach(5) | Should -Be 10
         }
+
+        # Pending: https://github.com/PowerShell/PowerShell/issues/6567
+        It "ForEach magic method works for dynamic (DLR) things" -Pending:$true {
+            Import-Module HelpersCommon
```
iSazonov: The same.

src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
iSazonov: Please address the comment.

```c#
if ((x == null) && (target.BaseObject is System.Dynamic.IDynamicMetaObjectProvider))
```

src/System.Management.Automation/engine/InternalCommands.cs
iSazonov: I leave this open - we should fix this in the PR or later.

src/System.Management.Automation/engine/InternalCommands.cs
iSazonov: Please remove the unneeded comments.

TravisEz13:
@lzybkr Can you update your review?

lzybkr (approved):

jazzdelightsme:
Thanks for the reviews, people! @TravisEz13 what is the process for getting this merged? Is there any transparency to how you track proposed merges, or is it a "it happens when it happens" sort of thing?

I don't mean to be impatient; I'm just wondering how it works, and I want to make sure there aren't any extra steps I need to do or anything like that.

iSazonov:
@jazzdelightsme MSFT team has announced that they is actively working on porting Windows PowerShell modules with minimal activity on GitHub. So now some PRs is waiting reviews from MSFT experts.

daxian-dbw:
Hey @jazzdelightsme, thanks for your contribution and patience!
Can you please push another commit with  the `[feature]` tag at the beginning of the commit message? That way it will trigger the daily build run which runs more tests. I want the chagnes to be validated by the daily test run. Thanks!

jazzdelightsme:
> Can you please push another commit with the [feature] tag at the beginning of the commit message?

Did it work (to trigger the extra testing)?

TravisEz13:
@jazzdelightsme  The feature tag did work.

src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
```diff
@@ -123,13 +125,20 @@ internal List<MshExpression> ResolveNames(PSObject target, bool expand)
...
+                                            PSMemberTypes.Properties | PSMemberTypes.PropertySet | PSMemberTypes.Dynamic);
             }
             else
             {
                 // we have no globbing: try an exact match, because this is quicker.
                 PSMemberInfo x = target.Members[_stringValue];
 
+                if ((x == null) && (target.BaseObject is System.Dynamic.IDynamicMetaObjectProvider))
+                {
+                    // We could check if GetDynamicMemberNames includes the name...  but
```
daxian-dbw: About this comment, we have already checked `GetDynamicMemberNames` when reaching here (`target.Members` includes the dynamic members returned from `GetDynamicMemberNames`). So maybe we can change comment to: `GetDynamicMemberNames does not return this name, but GetDynamicMemberNames is only a hint ...`

daxian-dbw (approved):
LGTM.
@jazzdelightsme I only have one comment about a comment in the code.
I can update the comment for you if you are OK with what I suggest.

src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
jazzdelightsme: Your comment is not quite correct. (Maybe it *should* be correct and there is another bug; I'm not sure what the intended behavior is.)

If you iterate through `.Members` you will indeed see the dynamic members. However, if you do `.Members[propName]`, where `propName` is the name of one of those dynamic members... you will get `null` back.

Sample output:

```powershell
PS C:\src\PowerShell> $myObj.PSObject.Members | where Name -eq FooProp


MemberType      : Dynamic
Value           :
TypeNameOfValue : dynamic
Name            : FooProp
IsInstance      : True



PS C:\src\PowerShell> $myObj.PSObject.Members['FooProp']
PS C:\src\PowerShell> $myObj.PSObject.Members['FooProp'] -eq $null
True
PS C:\src\PowerShell>
```

I'm not sure if that's a bug or By Design. Because you can't do much with PSDynamicMembers (for instance, the getter/setter just throws), maybe it is intentionally excluded when indexing directly in. But it does seem odd.

daxian-dbw: I see. That does seem odd. I will look into it tomorrow. Thanks for bringing it up!.

iSazonov:
@jazzdelightsme Thanks for the great contribution!

PR #6816, opened 2018-05-03
Remove empty branch and add Dbg.Assert in ExportCsvHelper.

sethvs:
## PR Summary

Remove empty branch and add Dbg.Assert in ExportCsvHelper.
Fix #6809 

## PR Checklist

- [x] [PR has a meaningful title](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - Use the present tense and imperative mood when describing your changes
- [x] [Summarized changes](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [x] [Change is not breaking](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes)
- [x] [Make sure all `.h`, `.cpp`, `.cs`, `.ps1` and `.psm1` files have the correct copyright header](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [x] This PR is ready to merge and is not [Work in Progress](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---work-in-progress).
    - If the PR is work in progress, please add the prefix `WIP:` to the beginning of the title and remove the prefix when the PR is ready.
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - [ ] User-facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
        - [ ] Issue filed - Issue link:
- **Testing - New and feature**
    - [x] Not Applicable or can only be tested interactively
    - **OR**
    - [ ] [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
        - [ ] [Add `[feature]` if the change is significant or affects feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

PR #6910, opened 2018-05-21
Replace "Dbg.Assert" with 'if () throw' in CSVCommands.cs.

sethvs:
## PR Summary

As of discussion in #6816 

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
@@ -331,7 +331,7 @@ namespace Microsoft.PowerShell.Commands
         {
             if (_path == null)
             {
-                throw new InvalidOperationException("FileName is mandatory parameter");
+                throw new InvalidOperationException(CsvCommandStrings.FileNameIsAMandatoryParameter);
             }
 
             string resolvedFilePath = PathUtils.ResolveFilePath(this.Path, this, _isLiteralPath);
@@ -421,12 +421,12 @@ namespace Microsoft.PowerShell.Commands
         {
             if (!_isActuallyAppending)
             {
-                throw new InvalidOperationException("This method should only get called when appending");
+                throw new InvalidOperationException(CsvCommandStrings.ReconcilePreexistingPropertyNamesMethodShouldOnlyGetCalledWhenAppending);
             }
 
             if (_preexistingPropertyNames == null)
             {
-                throw new InvalidOperationException("This method should only get called when we have successfully read preexisting property names");
+                throw new InvalidOperationException(CsvCommandStrings.ReconcilePreexistingPropertyNamesMethodShouldOnlyGetCalledWhenPreexistingPropertyNamesHaveBeenReadSuccessfully);
             }
 
             HashSet<string> appendedPropertyNames = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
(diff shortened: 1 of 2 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -331,6 +331,11 @@ private void CreateFileStream()
         {
             Dbg.Assert(_path != null, "FileName is mandatory parameter");
```
iSazonov: I think we can remove the asserts now.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -331,6 +331,11 @@ private void CreateFileStream()
...
 
+            if (_path == null)
+            {
+                throw new InvalidOperationException("FileName is mandatory parameter");
```
dantraMSFT: Since the strings are now user visible, they need to be moved to a resx file. This applies to all the changes that now throw exceptions.

dantraMSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
iSazonov: Shall we localize the strings? In the case we should review all code base.

dantraMSFT: @iSazonov I don't believe we localize assert strings because they are for debug only usage. Exception strings are handled like any other display string.

iSazonov: @dantraMSFT In this case, I think there is no point in transferring these throw debug strings to the resource files.

dantraMSFT: @iSazonov: I believe there is. We are effectively creating new strings that are displayed to the user. The fact that they were previously assert strings doesn't change that, IMHO.

iSazonov: @dantraMSFT What is a guideline? Will we mess the strings with others in the same resx files or we should create new resx files for assert strings? And where we should document this? In CONTRIBUTING.md?

iSazonov: We just discuss this becouse never do this previously.

sethvs: Moved.

dantraMSFT: @iSazonov: There already is a guideline in docs\dev-process\coding-guidelines.

Resource strings used for errors or UI should be put in resource files (`.resx`) so that they can be localized later.

With few, ahem, exceptions, strings passed as the message to an exception are error strings and should be in the resx file.  The fact that hit used to be a dbg.assert doesn't change this (IMHO).

iSazonov: @dantraMSFT Thanks for clarify!

iSazonov:
@dantraMSFT Could you please update your review?

dantraMSFT (approved):

PR #6913, opened 2018-05-21
Add Jumplist 'Run as Administrator' to Taskbar on Windows

### LINKED ISSUE #6649: PowerShell Core Installer for Windows - add the "Run as Administrator" action to taskbar button.
With Windows PowerShell pinned to my taskbar, I can easily run it as an administrator with the action at the top:

image

But with PowerShell Core, this action is missing:

image

Yeah, I know I can right click on PowerShell-6.x.x and then get the option to run as admin but that is buried UI and is harder to find than it should be.

bergmeister:
## PR Summary

~~**travisez13: marking as WIP while I run compliance tools to make sure this doesn't break any compliance rules.  Don't remove WIP until I've signed off on compliance issues**~~

Closes #6649 

This is a port of existing C++ Windows PowerShell code from [MainEntry.cpp](https://github.com/PowerShell/PowerShell/blob/master/src/powershell-native/nativemsh/pwrshexe/MainEntry.cpp)
Some of the code has been copied and minified from the [WindowsApiPack](https://github.com/contre/Windows-API-Code-Pack-1.1).

The code is not compiled for `Linux` (not sure also another condition is needed for Windows on ARM?).
The code checks if the PowerShell process has a window handle by checking the startupinfo and only then tries to populate the list (and also checks if there is a slot available in the jumplist).
Tested on Windows 10 1803, jumpLists have been supported in Windows since Windows 7, which matches what PowerShell Core supports.

This is how it looks like:

image

Note that I found that on WinServer 2012 R2, the icon does not recognised and is empty instead, I tried using the `SetIconLocation` method to point it to the ico file instead (which works on Win10) but it did not work on WinServer2012R2, therefore I suspect I have either an odd VM or it is an OS bug. Since this is only a minor optical and non-functional detail on a less common OS, I think this is OK.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/Microsoft.PowerShell.ConsoleHost.csproj
+++ b/src/Microsoft.PowerShell.ConsoleHost/Microsoft.PowerShell.ConsoleHost.csproj
@@ -16,7 +16,7 @@
     <DefineConstants>$(DefineConstants);CORECLR</DefineConstants>
   </PropertyGroup>
 
-  <ItemGroup Condition=" '$(Configuration)' == 'Linux' ">
+  <ItemGroup Condition=" '$(IsWindows)' != 'true' ">
     <Compile Remove="WindowsTaskbarJumpList\ComInterfaces.cs" />
     <Compile Remove="WindowsTaskbarJumpList\HResult.cs" />
     <Compile Remove="WindowsTaskbarJumpList\PropertyKey.cs" />
--- a/src/Microsoft.PowerShell.ConsoleHost/Microsoft.PowerShell.ConsoleHost.csproj
+++ b/src/Microsoft.PowerShell.ConsoleHost/Microsoft.PowerShell.ConsoleHost.csproj
@@ -17,11 +17,7 @@
   </PropertyGroup>
 
   <ItemGroup Condition=" '$(IsWindows)' != 'true' ">
-    <Compile Remove="WindowsTaskbarJumpList\ComInterfaces.cs" />
-    <Compile Remove="WindowsTaskbarJumpList\HResult.cs" />
-    <Compile Remove="WindowsTaskbarJumpList\PropertyKey.cs" />
-    <Compile Remove="WindowsTaskbarJumpList\PropVariant.cs" />
-    <Compile Remove="WindowsTaskbarJumpList\TaskbarJumpList.cs" />
+    <Compile Remove="WindowsTaskbarJumpList\*.cs" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Microsoft.PowerShell.ConsoleHost/Microsoft.PowerShell.ConsoleHost.csproj
+++ b/src/Microsoft.PowerShell.ConsoleHost/Microsoft.PowerShell.ConsoleHost.csproj
@@ -16,6 +16,14 @@
     <DefineConstants>$(DefineConstants);CORECLR</DefineConstants>
   </PropertyGroup>
 
+  <ItemGroup Condition=" '$(Configuration)' == 'Linux' ">
+    <Compile Remove="WindowsTaskbarJumpList\ComInterfaces.cs" />
+    <Compile Remove="WindowsTaskbarJumpList\HResult.cs" />
+    <Compile Remove="WindowsTaskbarJumpList\PropertyKey.cs" />
+    <Compile Remove="WindowsTaskbarJumpList\PropVariant.cs" />
+    <Compile Remove="WindowsTaskbarJumpList\TaskbarJumpList.cs" />
+  </ItemGroup>
+
   <ItemGroup>
     <Compile Remove="singleshell\installer\EngineInstaller.cs" />
     <Compile Remove="singleshell\installer\MshHostMshSnapin.cs" />
--- a/src/Microsoft.PowerShell.ConsoleHost/WindowsTaskbarJumpList/TaskbarJumpList.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/WindowsTaskbarJumpList/TaskbarJumpList.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
 using System.Reflection;
 using static Microsoft.PowerShell.ComInterfaces;
 
--- a/src/Microsoft.PowerShell.ConsoleHost/WindowsTaskbarJumpList/TaskbarJumpList.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/WindowsTaskbarJumpList/TaskbarJumpList.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Diagnostics;
 using System.Reflection;
 using static Microsoft.PowerShell.ComInterfaces;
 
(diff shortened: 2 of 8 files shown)


TravisEz13:
`[feature]` only works for commits

src/Microsoft.PowerShell.ConsoleHost/Microsoft.PowerShell.ConsoleHost.csproj
```diff
@@ -16,6 +16,14 @@
     <DefineConstants>$(DefineConstants);CORECLR</DefineConstants>
   </PropertyGroup>
 
+  <ItemGroup Condition=" '$(Configuration)' == 'Linux' ">
```
TravisEz13: Configuration `Linux` has been removed.  See [Common Props](https://github.com/PowerShell/PowerShell/blob/948532ac839f3bc7e8bbbd3c9f5f292868ac8687/PowerShell.Common.props#L138) for how to conditionally remove for Non-Windows.

TravisEz13 (commented):
I reviewed, but since this a WIP... I'm just marking my place.

src/Microsoft.PowerShell.ConsoleHost/Microsoft.PowerShell.ConsoleHost.csproj
bergmeister: Thanks, I addressed that.

bergmeister:
@TravisEz13 Thanks for having a quick look. I finished the last cleanup items and it is not WIP any more.

TravisEz13:
@bergmeister Would performance be better if you checked if we already are registered and don't register?

Also, I need to run compliance tools due to some changes you made.  I'm going to mark this as WIP again until the compliance tools say this change is good.

bergmeister:
@TravisEz13 By 'registered', do you mean if the jumplist has already been populated? As far as I am aware the Windows APIs for the jumplist, do not have 'Get' methods to check their current state (but I might be wrong). Any pointers are welcome.
There is no hurry for the review although getting it into the next preview might be good for testing exposure. So take your time, I will be on a conference from Wednesday on anyway.

TravisEz13:
@bergmeister No compliance issue with your new code
> As far as I am aware the Windows APIs for the jumplist, do not have 'Get' methods to check their current state 
That answered my question.

TravisEz13:
restarted macOS due to https://github.com/PowerShell/PowerShell/issues/6915

src/Microsoft.PowerShell.ConsoleHost/WindowsTaskbarJumpList/TaskbarJumpList.cs
```diff
@@ -0,0 +1,69 @@
...
+                    pCustDestList.CommitList();
+                }
+            }
+        }
+
+        private static void CheckAndReturn(HResult hResult)
+        {
+            if (hResult < 0)
+            {
+                throw new Exception($"HResult from COM call was negative, which indicates a failure. Was: '{hResult}'");
```
iSazonov: We use explicit formating in our code and never use the operator `$()`.

Is the message is user-faced? If so we should move it in resource file.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
```diff
@@ -1350,7 +1353,7 @@ internal TextWriter ConsoleTextWriter
 
         private uint Run(CommandLineParameterParser cpp, bool isPrestartWarned)
         {
-            Dbg.Assert(null != cpp, "CommandLine parameter parser cannot be null.");
+            Dbg.Assert(cpp != null, "CommandLine parameter parser cannot be null.");
```
iSazonov: Should we replace the assert with throw?

src/Microsoft.PowerShell.ConsoleHost/WindowsTaskbarJumpList/TaskbarJumpList.cs
bergmeister: Good point. I was thinking about this more after I opened the PR. Initially I wrote it as a check to make sure all COM calls were successful at each stage during development. Looking at it again now, it is probably better to not throw but instead make the function return/stop prematurely (which should not be a problem because we haven't commited the transaction)

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHost.cs
bergmeister: No strong opinion on this one, this line was only changed because the code style checker flagged it up that null was on the left hand side (which is only necessary in C/C++ but not C#)

iSazonov:
Does that slow down the PowerShell Core start?

bergmeister:
I have not measured it but I could not feel a difference. It also only affects pwsh when it is being opened in an actual window.

iSazonov:
I guess we get this easily with .Net Core 3.0 - right?

bergmeister:
Maybe, but .Net Core 3 development has only just started and realistically speaking, the chance that they will  be delayed is high and they are probably going to deliver a MVP first. But it is not a reason against this PR because one can just delete the folder that I added, so there is no deep integration that would make it difficult to take out in the future

iSazonov:
.Net Framework has [System.Windows.Shell.JumpList ](https://msdn.microsoft.com/en-us/library/system.windows.shell.jumplist(v=vs.110).aspx)

bergmeister:
Yes, but it does not support elevation.

TravisEz13:
@SteveL-MSFT Do we want to measure perf before accepting this?

bergmeister:
I could give you a measurement using the Stopwatch class when being compiled in Release and CrossGen'd to give you a rough number. Hardware would be either a 7th Gen i7 Notebook processor or an Azure VM size of your choice.
If you are worried, then we could potentially also implement the option of disabling the feature if a special environment variable is set. Otherwise it is a case of 'there is no such thing as free lunch'.

iSazonov:
The code remove one extra mouse click in GUI. This seems to be a very expensive solution, given that there are many non-interactive scenarios.

bergmeister:
Expensive in terms of what? In non-interactive scenarios, the only overhead is one call to get the startupinfo and then the algorithm will not proceed.

iSazonov:
> Expensive in terms of what?

Over 600 code lines to add one menu item.

bergmeister:
The main code is just the 90 lines in the TaskbarJumpList file, which was a port of the existing Windows PowerShell C++ code in the referenced MainEntry.cpp
The other classes were taken from the WindowsApiPack and already minimized to include only the smallest set of required classes and methods
Is the concern more about the review or maintenance/support of this PR?

iSazonov:
@bergmeister I'm trying to understand the need for this change, the difficulty in supporting and the negative impact on the start. So far I can take it but I don't see much benefits.

bergmeister:
The benefit of this feature is that it halves the time to open am elevated shell, brings consistency with Windows PowerShell and having one less click to do reduces muscle strain in the long term.

The difficulty in writing/supporting this feature is the lack of support for this scenario from the Microsoft/.Net side:
The COM APIs are not very user friendly and require memory structs that are awkard to work with in .Net (especially `PropVariant`). Therefore the Windows team made the WindowsApiPack in .Net 3.5 and stopped supporting it because it got merged into WPF, but both implementations are missing overloads to support elevation and do not provide access to the low level shell object pointer)
Hence why I ripped out the minimum necessary code for us and added a call to the COM Api that enables elevation. But the whole algorithm is the same that Windows PowerShell uses as well.

bergmeister:
@TravisEz13 @iSazonov @SteveL-MSFT I measured the performance impact using the `StopWatch` class: The first time it took 70 ms and after that the times were around **55 ms**. I did notice however that the times were the same when being launched in a new window or in the same windows. By adding debug statements I found that this is due to the startupInfo check not being sufficient to make it stop early. I think the last case could be improved to reduce the impact for starting new non-interactive runs on Windows or when launching pwsh as part of an existing shell window. 

- OS: Windows 10 1803 (10.0.17134 Build 17134)
- CPU: [Intel i7-7700HQ](https://ark.intel.com/products/97185/Intel-Core-i7-7700HQ-Processor-6M-Cache-up-to-3_80-GHz) (4 core Laptop version with 45 W TDP)
- Disk: SSD - Samsung EVO 850

src/Microsoft.PowerShell.ConsoleHost/WindowsTaskbarJumpList/TaskbarJumpList.cs
```diff
@@ -12,83 +12,87 @@ internal static class TaskbarJumpList
     {
         internal static void CreateElevatedEntry(string title)
         {
-            // check if the current shell owns a window
-            GetStartupInfo(out StartUpInfo startupInfo);
-            var STARTF_TITLEISLINKNAME = 0x00000800;
-            if (startupInfo.lpTitle == null || (startupInfo.dwFlags & STARTF_TITLEISLINKNAME) != STARTF_TITLEISLINKNAME)
+            if (Environment.UserInteractive)
```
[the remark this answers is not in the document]
bergmeister: Lol. I will remove it then. The idea behind this was to not even have a single COM call in non-interactive mode (but it will should stop after the call to getstartupinfo then)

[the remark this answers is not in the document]
iSazonov: I think it should be PowerShell internal flag - whether it interacts with an user.

src/Microsoft.PowerShell.ConsoleHost/Microsoft.PowerShell.ConsoleHost.csproj
```diff
@@ -16,6 +16,14 @@
     <DefineConstants>$(DefineConstants);CORECLR</DefineConstants>
   </PropertyGroup>
 
+  <ItemGroup Condition=" '$(IsWindows)' != 'true' ">
```
TravisEz13: explicit is better.

TravisEz13 (approved):

iSazonov:
@bergmeister Thanks for your contribution! I am delighted that you have found this solution.

Issue #6916, opened 2018-05-22
Transcription of Write-Information works if preference is 'SilentlyContinue'


hubuk:
Related issue: #4645

run.ps1
------------------

```powershell 
Start-Transcript -Path (Join-Path $PSScriptRoot 'transcriptdata.txt')
$InformationPreference = 'SilentlyContinue'
Write-Information 'SilentlyContinue'
$InformationPreference = 'Continue'
Write-Information 'LoudlyContinue'
Stop-Transcript
```

Steps to reproduce
------------------

```powershell 
.\run.ps1
```

Expected behavior
-----------------

```none
transcriptdata.txt should contains only LoudlyContinue
```

Actual behavior
---------------

```none
transcriptdata.txt contains only SilentlyContinue
```

Environment data
----------------

```powershell
> $PSVersionTable
Name                           Value
----                           -----
PSVersion                      6.1.0-preview.1
PSEdition                      Core
GitCommitId                    v6.1.0-preview.1
OS                             Microsoft Windows 10.0.16299
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

Investigation
----------------

The problem is in if/else statement which starts here:
https://github.com/PowerShell/PowerShell/blob/948532ac839f3bc7e8bbbd3c9f5f292868ac8687/src/System.Management.Automation/engine/MshCommandRuntime.cs#L757-L758
and has a continuation here:
https://github.com/PowerShell/PowerShell/blob/948532ac839f3bc7e8bbbd3c9f5f292868ac8687/src/System.Management.Automation/engine/MshCommandRuntime.cs#L824-L828

Transcription of Write-Host commands are handled in InternalUI.Write* methods as the comment suggests. These commands are distinguished by "PSHOST" tag.
Transcription of Write-Information commands is supposed to take place in the else statement but code in this statement is being executed only if 
``` C#
(!record.Tags.Contains("PSHOST") || record.Tags.Contains("FORWARDED"))
&&
preference != ActionPreference.Continue)
```
This condition is invalid. Correct condition is:
``` C#
!record.Tags.Contains("PSHOST")
&&
preference == ActionPreference.Continue)
```

PR pending...

iSazonov:
@mklement0 Could you please comment?

mklement0:
[...]

PR #6917, opened 2018-05-22
Fix conditions for transcription of Write-Information command.

hubuk:
## PR Summary

Fix for #6916.

This change makes transcription of Write-Information command consistent with `$InfomrationPreference` variable.
Currently Write-Information command is being transcribed if `$InfomrationPreference` is set to `'SilentlyContinue'` and not transcribed if set to `'Continue'`.

The problem is in if/else statement which starts here:
https://github.com/PowerShell/PowerShell/blob/948532ac839f3bc7e8bbbd3c9f5f292868ac8687/src/System.Management.Automation/engine/MshCommandRuntime.cs#L757-L758
and has a continuation here:
https://github.com/PowerShell/PowerShell/blob/948532ac839f3bc7e8bbbd3c9f5f292868ac8687/src/System.Management.Automation/engine/MshCommandRuntime.cs#L824-L828

Transcription of Write-Host commands are handled in InternalUI.Write* methods as the comment suggests. These commands are distinguished by "PSHOST" tag.
Transcription of Write-Information commands is supposed to take place in the else statement but code in this statement is being executed only if 
``` C#
(!record.Tags.Contains("PSHOST") || record.Tags.Contains("FORWARDED"))
&&
preference != ActionPreference.Continue)
```
This condition is invalid. Correct condition is:
``` C#
!record.Tags.Contains("PSHOST")
&&
preference == ActionPreference.Continue)
```

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [x] Issue filed - Issue link: #6916
    - **OR**

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WriteConsoleCmdlet.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WriteConsoleCmdlet.cs
@@ -137,7 +137,6 @@ namespace Microsoft.PowerShell.Commands
             }
 
             this.WriteInformation(informationMessage, new string[] { "PSHOST" });
-            this.Host.UI.TranscribeResult(result);
         }
 
         private Boolean _notAppendNewline = false;
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfacePromptForChoice.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfacePromptForChoice.cs
@@ -417,7 +417,12 @@ namespace Microsoft.PowerShell
             result = ReadLineResult.endedOnEnter;
             return InternalTestHooks.ForcePromptForChoiceDefaultOption
                    ? string.Empty
-                   : ReadLine(false, string.Empty, out result, true, true);
+                   : ReadLine(
+                       endOnTab: false,
+                       initialContent: string.Empty,
+                       result: out result,
+                       calledFromPipeline: true,
+                       transcribeResult: true);
         }
 
         private void ShowChoiceHelp(Collection<ChoiceDescription> choices, string[,] hotkeysAndPlainLabels)
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfacePromptForChoice.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfacePromptForChoice.cs
@@ -107,7 +107,7 @@ namespace Microsoft.PowerShell
                     WriteChoicePrompt(hotkeysAndPlainLabels, defaultChoiceKeys, false);
 
                     ReadLineResult rlResult;
-                    string response = ReadLine(false, string.Empty, out rlResult, true, true);
+                    string response = ReadChoiceResponse(out rlResult);
 
                     if (rlResult == ReadLineResult.endedOnBreak)
                     {
@@ -253,7 +253,7 @@ namespace Microsoft.PowerShell
                     WriteToConsole(PromptColor, RawUI.BackgroundColor, WrapToCurrentWindowWidth(choiceMsg));
 
                     ReadLineResult rlResult;
-                    string response = ReadLine(false, string.Empty, out rlResult, true, true);
+                    string response = ReadChoiceResponse(out rlResult);
 
                     if (rlResult == ReadLineResult.endedOnBreak)
                     {
@@ -412,6 +412,14 @@ namespace Microsoft.PowerShell
             WriteToConsole(fg, bg, trimEnd ? text.TrimEnd(null) : text);
         }
 
+        private string ReadChoiceResponse(out ReadLineResult result)
+        {
+            result = ReadLineResult.endedOnEnter;
+            return InternalTestHooks.ForcePromptForChoiceDefaultOption
+                   ? string.Empty
+                   : ReadLine(false, string.Empty, out result, true, true);
+        }
+
         private void ShowChoiceHelp(Collection<ChoiceDescription> choices, string[,] hotkeysAndPlainLabels)
         {
             Dbg.Assert(choices != null, "choices: expected a value");
@@ -473,4 +481,3 @@ namespace Microsoft.PowerShell
         }
     }
 }   // namespace
-
--- a/src/System.Management.Automation/engine/MshCommandRuntime.cs
+++ b/src/System.Management.Automation/engine/MshCommandRuntime.cs
@@ -826,7 +826,7 @@ namespace System.Management.Automation
                 if (!record.Tags.Contains("PSHOST") && (preference == ActionPreference.Continue))
                 {
                     // Only transcribe informational messages here. Transcription of PSHost-targeted messages is done in the InternalUI.Write* methods.
-                    CBhost.InternalUI.TranscribeResult(StringUtil.Format(InternalHostUserInterfaceStrings.InformationFormatString, record.ToString()));
+                    CBhost.InternalUI.TranscribeResult(record.ToString());
                 }
             }
 
--- a/src/System.Management.Automation/engine/MshCommandRuntime.cs
+++ b/src/System.Management.Automation/engine/MshCommandRuntime.cs
@@ -821,12 +821,12 @@ namespace System.Management.Automation
                             CBhost.InternalUI.WriteLine(record.ToString());
                         }
                     }
+                }
                     
-                    if (!record.Tags.Contains("PSHOST") && (preference == ActionPreference.Continue))
-                    {
-                        // Only transcribe informational messages here. Transcription of PSHost-targeted messages is done in the InternalUI.Write* methods.
-                        CBhost.InternalUI.TranscribeResult(StringUtil.Format(InternalHostUserInterfaceStrings.InformationFormatString, record.ToString()));
-                    }
+                if (!record.Tags.Contains("PSHOST") && (preference == ActionPreference.Continue))
+                {
+                    // Only transcribe informational messages here. Transcription of PSHost-targeted messages is done in the InternalUI.Write* methods.
+                    CBhost.InternalUI.TranscribeResult(StringUtil.Format(InternalHostUserInterfaceStrings.InformationFormatString, record.ToString()));
                 }
             }
 
--- a/src/System.Management.Automation/engine/MshCommandRuntime.cs
+++ b/src/System.Management.Automation/engine/MshCommandRuntime.cs
@@ -822,8 +822,8 @@ namespace System.Management.Automation
                         }
                     }
                 }
-                    
-                if (!record.Tags.Contains("PSHOST") && (preference == ActionPreference.Continue))
+
+                if (!record.Tags.Contains("PSHOST") && (preference != ActionPreference.SilentlyContinue))
                 {
                     // Only transcribe informational messages here. Transcription of PSHost-targeted messages is done in the InternalUI.Write* methods.
                     CBhost.InternalUI.TranscribeResult(record.ToString());
--- a/src/System.Management.Automation/engine/MshCommandRuntime.cs
+++ b/src/System.Management.Automation/engine/MshCommandRuntime.cs
@@ -821,7 +821,8 @@ namespace System.Management.Automation
                             CBhost.InternalUI.WriteLine(record.ToString());
                         }
                     }
-                    else
+                    
+                    if (!record.Tags.Contains("PSHOST") && (preference == ActionPreference.Continue))
                     {
                         // Only transcribe informational messages here. Transcription of PSHost-targeted messages is done in the InternalUI.Write* methods.
                         CBhost.InternalUI.TranscribeResult(StringUtil.Format(InternalHostUserInterfaceStrings.InformationFormatString, record.ToString()));
--- a/src/System.Management.Automation/engine/MshCommandRuntime.cs
+++ b/src/System.Management.Automation/engine/MshCommandRuntime.cs
@@ -748,8 +748,7 @@ namespace System.Management.Automation
                     //
                     if (null == Host || null == Host.UI)
                     {
-                        Diagnostics.Assert(false, "No host in CommandBase.WriteInformation()");
-                        throw PSTraceSource.NewInvalidOperationException();
+                        throw PSTraceSource.NewInvalidOperationException("No host in CommandBase.WriteInformation()");
                     }
 
                     CBhost.InternalUI.WriteInformationRecord(record);
--- a/src/System.Management.Automation/engine/MshCommandRuntime.cs
+++ b/src/System.Management.Automation/engine/MshCommandRuntime.cs
@@ -748,7 +748,7 @@ namespace System.Management.Automation
                     //
                     if (null == Host || null == Host.UI)
                     {
-                        Diagnostics.Assert(false, "No host in CommandBase.WriteVerbose()");
+                        Diagnostics.Assert(false, "No host in CommandBase.WriteInformation()");
                         throw PSTraceSource.NewInvalidOperationException();
                     }
 
@@ -823,9 +823,11 @@ namespace System.Management.Automation
                     }
                 }
 
-                if (!record.Tags.Contains("PSHOST") && (preference != ActionPreference.SilentlyContinue))
+                // Both informational and PSHost-targeted messages are transcribed here.
+                // The only difference between these two is that PSHost-targeted messages are transcribed
+                // even if InformationAction is SilentlyContinue.
+                if (record.Tags.Contains("PSHOST") || (preference != ActionPreference.SilentlyContinue))
                 {
-                    // Only transcribe informational messages here. Transcription of PSHost-targeted messages is done in the InternalUI.Write* methods.
                     CBhost.InternalUI.TranscribeResult(record.ToString());
                 }
             }
--- a/test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
@@ -53,6 +53,7 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
 
     AfterEach {
         Remove-Item $transcriptFilePath -ErrorAction SilentlyContinue
+        [System.Management.Automation.Internal.InternalTestHooks]::SetTestHook('ForcePromptForChoiceDefaultOption', $False)
     }
 
     It "Should create Transcript file at default path" {
@@ -122,7 +123,7 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
             }
         }
 
-        Test-Path $transcriptFilePath | Should -BeTrue
+        $transcriptFilePath | Should -Exist
         $transcriptFilePath | Should -FileContentMatch "After Dispose"
     }
 
@@ -131,7 +132,7 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
         $powerShellCommand = $powerShellPath + ' -c "start-transcript $transcriptFilePath; Write-Host ''Before Dispose'';"'
         Invoke-Expression $powerShellCommand
 
-        Test-Path $transcriptFilePath | Should -BeTrue
+        $transcriptFilePath | Should -Exist
         $transcriptFilePath | Should -FileContentMatch "Before Dispose"
         $transcriptFilePath | Should -FileContentMatch "PowerShell transcript end"
     }
@@ -145,7 +146,7 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
 
         & $script
 
-        Test-Path $transcriptFilePath | Should -BeTrue
+        $transcriptFilePath | Should -Exist
         $machineName = [System.Environment]::MachineName
         $transcriptFilePath | Should -FileContentMatch $machineName
     }
@@ -160,7 +161,7 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
 
         & $script
 
-        Test-Path $transcriptFilePath | Should -BeTrue
+        $transcriptFilePath | Should -Exist
         $transcriptFilePath | Should -Not -FileContentMatch "INFO: "
         $transcriptFilePath | Should -FileContentMatch $message
     }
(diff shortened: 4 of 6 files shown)


iSazonov:
@hubuk Thanks for your contribution. I believe we need to get a conclusion in Issue before we continue the PR.

hubuk:
@iSazonov Sure. I need to check why a new test is failing. It was OK on my local machine.

hubuk:
I found that tests are executed with redirection to a pipe.
In this case all Write-* commands are always logged in the transcript file no matter what preference is set.
According to this I removed the second test (no transcription for 'SilentlyContinue') as it cannot be satisfied by current test environment.

iSazonov:
@mklement0 Could you please review the PR?

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
```diff
@@ -147,4 +147,16 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
...
+    It "Transcription should record Write-Information output when preference is set to Continue" {
+        $script = {
+            Start-Transcript -Path $transcriptFilePath
+            $InformationPreference = 'Continue'
+            Write-Information 'Continue'
+            Stop-Transcript }
+        & $script
+        Test-Path $transcriptFilePath | Should -BeTrue
+
+        $transcriptFilePath | Should FileContentMatch 'Continue'
```
iSazonov: Please use new syntax `Should -FileContentMatch`.

iSazonov (commented):
Leave a comment

mklement0 (commented):
Perhaps I'm missing something, but it it looks like you accidentally reverted the original fix.

hubuk:
@mklement0 You are right. My bad...
And it proved that newly written test is not that useful...
Will fix it shortly.

hubuk:
It should be OK now. Test was passing for the same reason why my previous test was failing.
All Write-* commands are transcribed if redirected to a pipe.

mklement0:
@hubuk:

> All Write-* commands are transcribed if redirected to a pipe.

Can you clarify what that means, specifically, and perhaps give an example?
Is that a problem in itself or by-design behavior?

iSazonov:
@hubuk Please don't fix CodeFactor issues in the PR (It's hard to review meaningful changes) - all style changes should be in new PR.

hubuk:
@iSazonov Reverted the change.

@mklement0 
> All Write-* commands are transcribed if redirected to a pipe.

Sorry, for wrong information I provided:
Tests was failing because of too simple data to check. 'Continue' was also part of the pwsh.exe command line:
`Host Application: ...\bin\Release\netcoreapp2.1\win7-x64\publish\pwsh.dll -noprofile -c $env:POWERSHELL_TELEMETRY_OPTOUT = 1;$ProgressPreference = 'silentlyContinue';...`
and this is part of the transcription file.
I still need to investigate my previous observations. For now lets assume that this is not an issue.

iSazonov:
Write-Host is related to this code so you have to add tests for it.

hubuk:
@iSazonov I added tests for `Write-Host` with `-InformationAction` `Continue` and `SilentlyContinue`. Currently `Ignore` does not write to console but is being transcribed, so I skipped test for it. Let me know if you would like to have this test included anyway. I can also spend some time to make it produce same result for console and for transcript file. Fixing console output rather than transcription will make `Write-Host` behave as documented.

Additionally removing `INFO: ` prefix mentioned by @mklement0 I easy. Should I proceed with it in this PR?

mklement0:
@hubuk:

> Fixing console output rather than transcription will make Write-Host behave as documented.

I'm currently only aware of problems with the _transcription_ behavior, not with _console output_, so the guiding principle should be to align transcription behavior with that of the current console-output behavior.

If, with respect to console-output behavior, you're referring to `Write-Host` being silenced by `-InformationAction Ignore`, that is not a _behavioral_ problem, it is a _documentation_ problem.

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
```diff
@@ -145,6 +145,66 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
...
+        Test-Path $transcriptFilePath | Should -BeTrue
+
+        $transcriptFilePath | Should -FileContentMatch $message
+    }
+
+    It "Transcription should record Write-Host output when InformationAction is set to SilentlyContinue" {
+        [String]$message = New-Guid
+        $script = {
+            Start-Transcript -Path $transcriptFilePath
+            Write-Host -Message $message -InformationAction SilentlyContinue
```
iSazonov: Please add test for Ignore.

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
```diff
@@ -145,6 +145,66 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
...
+
+    It "Transcription should record Write-Information output when InformationAction is set to Continue" {
+        [String]$message = New-Guid
+        $script = {
+            Start-Transcript -Path $transcriptFilePath
+            Write-Information -Message $message -InformationAction Continue
+            Stop-Transcript }
+        & $script
+        Test-Path $transcriptFilePath | Should -BeTrue
+
```
iSazonov: Please remove extra line.

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
```diff
@@ -145,6 +145,66 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
...
+    }
+
+    It "Transcription should record Write-Information output when InformationAction is set to Continue" {
+        [String]$message = New-Guid
+        $script = {
+            Start-Transcript -Path $transcriptFilePath
+            Write-Information -Message $message -InformationAction Continue
+            Stop-Transcript }
+        & $script
+        Test-Path $transcriptFilePath | Should -BeTrue
```
iSazonov: Please add new line.

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
```diff
@@ -145,6 +145,66 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
...
-        $transcriptFilePath | Should FileContentMatch $machineName
+        $transcriptFilePath | Should -FileContentMatch $machineName
+    }
+
+    It "Transcription should record Write-Information output when InformationAction is set to Continue" {
+        [String]$message = New-Guid
+        $script = {
+            Start-Transcript -Path $transcriptFilePath
+            Write-Information -Message $message -InformationAction Continue
+            Stop-Transcript }
```
iSazonov: Please use common formatting.
```powershell
$script = {  
    Start-Transcript -Path $transcriptFilePath  
    Write-Information -Message $message -InformationAction Continue  
    Stop-Transcript
}
```

iSazonov (commented):
Style comments for all added tests

iSazonov:
I'm not sure I need to remove the INFO: prefix. I suppose it was made that distinguish the output of the cmdlet. Although if we want to get the exact snapshot of the console then we need to remove this prefix

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
iSazonov: I see. I believe we should use the right pattern because It block and other code use right patterns.

hubuk:
There are two issues still present:
1. `Write-Host -InformationAction Ignore` is being transcribed but is should not be.
2. `Write-Host -InformationAction Inquire` has wrong order in transcript file:
```
> Write-Host 'Question?' -InformationAction Inquire
Question?

Confirm
Continue with this operation?
[Y] Yes  [A] Yes to All  [H] Halt Command  [S] Suspend  [?] Help (default is "Y"):
```
in transcript file:
```
Confirm
Continue with this operation?
&Yes  Yes to &All  &Halt Command  &Suspend
Y
Question?
```

The reason of 2. is that `Write-Host` is being transcribed separately in WriteConsoleCmdlet.cs file after the whole command has been handled.

Current transcript implementation for `Write-Information` can also be used also for `Write-Host` which will slightly simplify the code and provide fix for both 1. and 2.
May I include this fix in this PR?

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
hubuk: Fixed

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
hubuk: Fixed

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
iSazonov: Seems it is an isolation. I'am not sure that we should remove thos.

mklement0:
@hubuk: 

If there are no side effects, it seems like a good change to make (but it's not my call).

On a side note: the transcription of the prompt choices with the raw "accelerator chars." (`&Yes` instead of `[Y] Yes`) is unexpected (though not a big deal).

@iSazonov:

> I'm not sure I need to remove the INFO: prefix. I suppose it was made that distinguish the output of the cmdlet. Although if we want to get the exact snapshot of the console then we need to remove this prefix

Yes, consistency with console output is important, I think.

However, you could argue that the `INFO: ` prefix should be present in the console (as well), for consistency with the `Write-Verbose`, `Write-Warning`, and `Write-Debug` cmdlets, and that it is therefore console output that needs fixing in this case.

While changing that wouldn't technically be a _breaking_ change - _for-display_ output is not part of the contract with the user - it would be a _noticeable_ one.

Either way, let's discuss it in a separate issue: #6951

mklement0:
Now that we know that the `INFO: ` prefix is not printed to the console by design - see https://github.com/PowerShell/PowerShell/issues/6951#issuecomment-393290377 - we should remove it from the transcript too.

hubuk:
@mklement0 @iSazonov 
Made changes for two issues I mentioned.

Unfortunately "accelerator chars." are handled by `PromptForChoice` method in `ConsoleHostUserInterfacePromptForChoice.cs` file. Fix could be done by making part of this method public.

It seems like there could be much more cases in which transcription file is different from the console output. For example there is no transcription of the invalid accelerator keys user pressed.
Currently transcription code is spreed among entire codebase and it is hard to make it consistent with console output because host code may make changes to the data just before displaying them.

Better approach would be to make a transcript part of host classes which will make it easier to ensure consistency but it would be a major refactoring.

src/System.Management.Automation/engine/MshCommandRuntime.cs
```diff
@@ -822,10 +822,9 @@ internal void WriteInformation(InformationRecord record, bool overrideInquire =
                         }
                     }
                 }
-
-                if (!record.Tags.Contains("PSHOST") && (preference != ActionPreference.SilentlyContinue))
+                    
+                if (record.Tags.Contains("PSHOST") || (preference != ActionPreference.SilentlyContinue))
                 {
-                    // Only transcribe informational messages here. Transcription of PSHost-targeted messages is done in the InternalUI.Write* methods.
```
iSazonov: Seems new comment about 'Inquire' and PSHost-targets will be useful.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WriteConsoleCmdlet.cs
```diff
@@ -137,7 +137,6 @@ protected override void ProcessRecord()
             }
 
             this.WriteInformation(informationMessage, new string[] { "PSHOST" });
-            this.Host.UI.TranscribeResult(result);
```
iSazonov: Have we a test for this?

mklement0:
@hubuk: Thanks for the update; my sense is that at least for now we needn't worry about transcribing 100% faithfully as long as all the information is ultimately there.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WriteConsoleCmdlet.cs
hubuk: I already created three test for transcription of Write-Host command (Continue, SilentlyContinue, Ignore).
I have no idea how costly would it be to create a test for Inquire action as it makes the test interactive.

src/System.Management.Automation/engine/MshCommandRuntime.cs
hubuk: Most important change is that transcription of `Write-Host` and `Write-Information` is handled in the same place now.
The part about fixing `Inquire` has only historical value and does not introduce any new information about current behavior.
Anyway, take a look at my comment and let me know if you want it stated differently as English is not my native language.

src/System.Management.Automation/engine/MshCommandRuntime.cs
```diff
@@ -822,7 +821,13 @@ internal void WriteInformation(InformationRecord record, bool overrideInquire =
...
                     }
                 }
-                    
+
+                // Both informational and PSHost-targeted messages are transcribed here.
+                // The only difference betwen these two is that PSHost-targeted messages are transcribed
+                // even if InformationAction is SilentlyContinue.
+                // Previously messages targetting PSHost were transcribed in InternalUI.Write* methods.
+                // Because of that for -InformationAction Inquire the order of written message and choice prompt
+                // was different in host output and transcript file.
```
iSazonov: @mklement0 Could you please help and reword the comment?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WriteConsoleCmdlet.cs
adityapatwardhan: Is it possible to add a test hook?

iSazonov: Not clear where the hook should be to emulate the user interaction. Maybe in PromptForChoice method in ConsoleHostUserInterfacePromptForChoice.cs file?

src/System.Management.Automation/engine/MshCommandRuntime.cs
mklement0: @iSazonov: It sounds good to me in general, is there a specific aspect you're not happy with / that you think is missing?

On a copy-editing note, I suggest placing commas after "Previously" and "Because of that" and perhaps say "the  output sequencing of the message and the prompt choices differed between the host and the transcript".

Taking a step back:  we could put the information about how it _used to_ work in the commit message rather than  in a code comment.

iSazonov: @mklement0 Thanks! (English is not my native)
@hubuk Olease address @mklement0 comments (and update PR description) and correct typo in "betwen".

hubuk: Done

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WriteConsoleCmdlet.cs
iSazonov: @hubuk You can add hook here https://github.com/PowerShell/PowerShell/blob/c1c5344a8897262433141ecbc13bb06ac2c4bbef/src/System.Management.Automation/engine/Utils.cs#L1492, inject in PromptForChoice and use in the test.

hubuk: That's great. Thanks.

hubuk: Tests added.

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
```diff
@@ -1,5 +1,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
+
+using namespace System.Management.Automation.Internal
```
iSazonov: Previously we don't use this in our tests so maybe remove this and use full names in tests.

hubuk: Done.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfacePromptForChoice.cs
```diff
@@ -412,6 +412,14 @@ private void WriteChoiceHelper(string text, ConsoleColor fg, ConsoleColor bg, re
             WriteToConsole(fg, bg, trimEnd ? text.TrimEnd(null) : text);
         }
 
+        private string ReadChoiceResponse(out ReadLineResult result)
+        {
+            result = ReadLineResult.endedOnEnter;
+            return InternalTestHooks.ForcePromptForChoiceDefaultOption
+                   ? string.Empty
+                   : ReadLine(false, "", out result, true, true);
```
iSazonov: Sorry, I skipped this - please use string.Empty as in previous line.
Thank you for your patience. Seems this is my last comment.

hubuk: Changed. Not a problem.

iSazonov (approved):

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfacePromptForChoice.cs
```diff
@@ -412,6 +412,14 @@ private void WriteChoiceHelper(string text, ConsoleColor fg, ConsoleColor bg, re
             WriteToConsole(fg, bg, trimEnd ? text.TrimEnd(null) : text);
         }
 
+        private string ReadChoiceResponse(out ReadLineResult result)
+        {
+            result = ReadLineResult.endedOnEnter;
+            return InternalTestHooks.ForcePromptForChoiceDefaultOption
+                   ? string.Empty
+                   : ReadLine(false, string.Empty, out result, true, true);
```
SteveL-MSFT: Can you use [named parameters](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments) here?

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
```diff
@@ -132,19 +132,143 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
...
+        $transcriptFilePath | Should -Not -FileContentMatch "INFO: "
+        $transcriptFilePath | Should -Not -FileContentMatch $message
+    }
+
+    It "Transcription should record Write-Information output in correct order when InformationAction is set to Inquire" {
+        [String]$message = New-Guid
+        $newLine = [System.Environment]::NewLine
+        $expectedContent = "$message$($newLine)Confirm$($newLine)Continue with this operation?"
+        $script = {
+            [System.Management.Automation.Internal.InternalTestHooks]::SetTestHook('ForcePromptForChoiceDefaultOption', $True)
```
SteveL-MSFT: Seems like there should be an `AfterEach{}` that resets the test hook

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
```diff
@@ -132,19 +132,143 @@ Describe "Start-Transcript, Stop-Transcript tests" -tags "CI" {
...
+        [String]$message = New-Guid
+        $script = {
+            Start-Transcript -Path $transcriptFilePath
+            Write-Information -Message $message -InformationAction Continue
+            Stop-Transcript
+        }
+
+        & $script
+
+        Test-Path $transcriptFilePath | Should -BeTrue
```
SteveL-MSFT: Perhaps a search and replace to:

> $transcriptFilePath | Should -Exist

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.Powershell.Host/Start-Transcript.Tests.ps1
hubuk: Done.

hubuk:
@SteveL-MSFT Is there anything else you would like to be changed?

iSazonov:
@SteveL-MSFT Please update your review.

SteveL-MSFT (approved):

iSazonov:
I have a fear of removal FORWARDED condition - do it really not used?

/cc @SteveL-MSFT Who can confirm?

hubuk:
@iSazonov 
FORWARDED condition is still there: https://github.com/PowerShell/PowerShell/blob/c67c20c5c3caa7615028cb2cafea1cdfa9fe9b26/src/System.Management.Automation/engine/MshCommandRuntime.cs#L756-L757 Only logic for the else statement has changed from:
https://github.com/PowerShell/PowerShell/blob/e177fcaef610f874aa31cd3d85f97162d2859e72/src/System.Management.Automation/engine/MshCommandRuntime.cs#L824-L828 to: https://github.com/PowerShell/PowerShell/blob/c67c20c5c3caa7615028cb2cafea1cdfa9fe9b26/src/System.Management.Automation/engine/MshCommandRuntime.cs#L828-L831

**Before** my changes `TranscribeResult` was being executed for the following conditions:
1. In `WriteConsoleCmdlet.cs` which is when `record.Tags.Contains("PSHOST")`
2. `(preference != ActionPreference.Continue) && !record.Tags.Contains("PSHOST")` as the first part of the evaluation of condition for else statement.
3. `(preference != ActionPreference.Continue) && record.Tags.Contains("PSHOST") && record.Tags.Contains("FORWARDED")` as the second part.
_So transcription for FORWARDED tag was being executed only for Write-Host. But Write-Host had its own transcription code. This resulted in duplicated transcription for one Write-Host command execution and made FORWARDED tag irrelevant from transcription point of view._

**After** my changes `TranscribeResult` is being executed only for the following cases:
1. `record.Tags.Contains("PSHOST")`
2. `!record.Tags.Contains("PSHOST") && preference != ActionPreference.SilentlyContinue`

I hope this is somehow readable.

iSazonov:
@hubuk Sorry, sometimes GitHub shows diffs in very strange way.

iSazonov:
@hubuk Thank you for your contribution and your patience! 
I hope you will make still many useful contributions.

hubuk:
Happy to contribute. Thanks!

zoidhddhdd:
cool

PR #6923, opened 2018-05-23
Allow indentation of here-string closing characters.

### LINKED ISSUE #2337: here string syntax
Issue
----------------
it would be nice it here strings could close without necessarily needing to be at the start of a line.  Specifically it would be nice if I could tab out the closing side of a here string to to match the opening indentation.

ex:
```powershell
  @"
    Some wacky
    Multiline
    String
  "@
```
instead of this:
```powershell
  @"
    Some wacky
    Multiline
    String
"@
```

Environment data
----------------

```powershell
CLRVersion

```

sethvs:
## PR Summary

Fix #2337 

## PR Checklist

- **User-facing changes**
    - [] Not Applicable
    - **OR**
        - [x] Issue filed - Issue link: PowerShell/PowerShell-Docs#2471
    - **OR**

### DIFF
--- a/test/powershell/Language/Parser/Parser.Tests.ps1
+++ b/test/powershell/Language/Parser/Parser.Tests.ps1
@@ -784,6 +784,11 @@ foo``u{2195}abc
         { ExecuteCommand "`$herestr=@`"`n'`"'`n`"@" } | Should -Not -Throw
     }
 
+    It "A here string closing characters can be indented with spaces" {
+        { ExecuteCommand "@`"`n   here`n   string`n   `"@"} | Should -Not -Throw
+        ExecuteCommand "@`"`n   here`n   string`n   `"@" | Should -BeExactly "   here`n   string"
+    }
+
     It "Throw better error when statement should be put in named blocks - <name>" -TestCases @(
         @{ script = "Function foo { [CmdletBinding()] param() DynamicParam {} Hi"; name = "function" }
         @{ script = "{ begin {} Hi"; name = "script-block" }
--- a/test/powershell/Language/Parser/Parser.Tests.ps1
+++ b/test/powershell/Language/Parser/Parser.Tests.ps1
@@ -785,7 +785,6 @@ foo``u{2195}abc
     }
 
     It "A here string closing characters can be indented with spaces" {
-        { ExecuteCommand "@`"`n   here`n   string`n   `"@"} | Should -Not -Throw
         ExecuteCommand "@`"`n   here`n   string`n   `"@" | Should -BeExactly "   here`n   string"
     }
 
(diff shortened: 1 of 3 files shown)


iSazonov:
@sethvs Please add tests.

sethvs:
Added tests and docs.

test/powershell/Language/Parser/Parser.Tests.ps1
```diff
@@ -784,6 +784,11 @@ foo``u{2195}abc
         { ExecuteCommand "`$herestr=@`"`n'`"'`n`"@" } | Should -Not -Throw
     }
 
+    It "A here string closing characters can be indented with spaces" {
+        { ExecuteCommand "@`"`n   here`n   string`n   `"@"} | Should -Not -Throw
```
iSazonov: Please remove the line - next test do this.

sethvs: Agree. Removed.

iSazonov (approved):

lzybkr:
As mentioned in the issue, this **is** a potential breaking change, so the committee must review.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and the benefit doesn't justify the breaking change as the PR is currently defined so we cannot accept this as-is.  @sethvs if you want to pursue this, you would need to consider other needs of Here Strings which may require a new token so that this doesn't break existing scripts.

joeyaiello:
To add a couple other points: 

* this doesn't buy you much from a readability perspective when you can't indent the lines between the opening and closing operators, e.g.:

```powershell
function foo {
        $a = @"
Some text
that can't
be indented.
        "@
}
```

* it may be a reasonable user expectation that if they had a trailing amount of whitespace, that it'd be included in the here-string, e.g.

```powershell
function foo {
        $a = @"
Some text
that can't
be indented.
        "@ # why wouldn't the whitespace on this line be included?
}
```

* given that we don't get additional indentation in the content of the here-string, "good enough" readability is achievable with this syntax:

```powershell
function foo {
        $a = 
# inline python code cares about whitespace
@"
if 1 == 1:
    print("yes")
else:
    print("no")
"@
}
```

lzybkr:
My original design years ago (prototyped but never shipped) would strip as much indentation on each line as we found before the terminator - this made here strings quite readable indented, but it is potentially confusing.

rkeithhill:
@lzybkr That's kind of how I would expect any such change to behave. That said, I'm not sure it is worth it at this point.

iSazonov:
@sethvs Thanks for your contribution! 

I close the PR based on PowerShell Committee conclusion. We should continue the discussion in related Issue.

SteveL-MSFT:
@lzybkr we discussed that option (to use the terminator as indicator of stripping leading whitespace), but the main discussion was it would be confusing to the end user.  Perhaps a new token would resolve that.  Personally, I would like to do something like this:

```powershell
  $content = @"
    This is
    my text.
    "@
```

iSazonov:
We could use quotes in last line to set size of the desired indentation:
```c#
$content = @"
        This is
        my text.
"""""@
```

PR #6909, opened 2018-05-21
Cleanup FileSystem provider

iSazonov:
## PR Summary

Related #6874. The PR is coming from a code review to cleanup psl. Here we clean up the side branch of the code that will allow later to clean up a branch which uses psl.

- IO.FileInfo does not make system calls in constructor.  So we can create the object and then use the required attributes without direct call IO.FileInfo.GetAttributes() ( SafeGetFileAttributes() ). This allow us to exclude some p/invoke calls in our code in later cleanups. Also we get unified code for both Windows and Unix.
- Remove SafeGetFileAttributes() and WinSafeGetFileAttributes(). Currently .Net Core support file attributes on all platforms in fastest way and we can remove our workaround. We get a regression in rare case (for files like pagefile.sys). Fix is ready in [CoreFX](https://github.com/dotnet/corefx/issues/29782), we get it in 2.1.1. I suggest ignore the regression because this is a very-very rare situation (Get-ChildItem c:\pagefile.sys -Hidden). The .Net Core team was not even able to create an artificial test for such files and uses a real pagefile.sys file for the test. Also the enumeration is still working (dir c:\ -hidden).
- Re-add test which we lost in #4050. The test is pending because of the regression.

## PR Checklist

- [x] [PR has a meaningful title](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - Use the present tense and imperative mood when describing your changes
- [x] [Summarized changes](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [ ] [Change is not breaking](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes)
- [x] [Make sure all `.h`, `.cpp`, `.cs`, `.ps1` and `.psm1` files have the correct copyright header](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [x] This PR is ready to merge and is not [Work in Progress](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---work-in-progress).
    - If the PR is work in progress, please add the prefix `WIP:` to the beginning of the title and remove the prefix when the PR is ready.
- **User-facing changes**
    - [] Not Applicable
    - **OR**
[...]

Issue #6874, opened 2018-05-15
Need psl-native library code review


iSazonov:
psl-native library appeared at the time of the first version .Net Core 1.0 which had many limitations.
Currently we moved to .Net Core 2.1 and we need to review psl-native library code and remove unneeded p/invokes. Ideally we could remove this library altogether.

TODO Tasks (Please check when it's finished):
- [ ] Remove the native .cpp code related to the changes in #6909 from psl-native
- [ ] Remove the native .cpp code related to the changes in #6948 from psl-native
- [ ] Remove the native .cpp code related to the changes in #6929 from psl-native

TravisEz13:
I reviewed and SetDate is still needed

iSazonov:
We could use other glibc function (without tm struct) to fix SetDate..

iSazonov:
I started reviewing and already found APIs not used. Will push PR.

Some APIs use struct-s and we need review this in depth. We should either use .Net API or fix as recommended by CoreFX team or use another Unix API.

@PaulHigin Can we remove the unused code?
https://github.com/PowerShell/PowerShell/blob/master/src/libpsl-native/src/createprocess.cpp#L79

PaulHigin:
Hi @iSazonov   This is still needed for Linux platforms for SSH remoting.
https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs#L2368

We use this native code to create a client SSH process with the needed attributes, and this is not support by DotNet.  Also they indicated they would not support it in the future.

TravisEz13:
@iSazonov I reviewed for structs that cross into managed code and I only found the one and have PR #6881 to fix as DotNet recommended.

dantraMSFT:
psl-native is also needed for syslog and os_log.

adityapatwardhan:
The native code is going to be moved to its own repo: https://github.com/PowerShell/PowerShell-Native

Please make changes here. I am working on creating builds and setting up CI system for it.

PR #6929, opened 2018-05-24
Cleanup NativeItemExists()

iSazonov:
## PR Summary

Related #6874. After the PR we can remove the files - isfile.cpp, isfile.h, isdirectory.cpp, isdirectory.h - from psl.

- Use File.GetAttributes() instead of p/invokes. The method throws access exceptions (unlike File.Exists and Directory.Exists) which we use for better error messages.
- Remove 'Native' from method names.
- As stated in #6909 we get a regression in rare case (for files like pagefile.sys). Fix is ready in [CoreFX](https://github.com/dotnet/corefx/issues/29782), we get it in 2.1.1. I suggest ignore the regression because this is a very-very rare situation (Get-ChildItem c:\pagefile.sys -Hidden). The .Net Core team was not even able to create an artificial test for such files and uses a real pagefile.sys file for the test. Also the enumeration is still working (dir c:\ -hidden).

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**


iSazonov:
Reopen to restart CIs.

iSazonov:
@TravisEz13 Could you please help - I can not find why CIs failed?

iSazonov:
Feature tests failed with Pester not found. :confused:

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -895,113 +895,49 @@ internal static bool IsAdministrator()
...
-            return NativeItemExists(path, out unusedIsDirectory, out unusedException);
+            bool result = false;
+            try
+            {
+                result =  ItemExists(path, out bool unused);
+            }
+            catch
+            {
+            }
+            return false;
```
SteveL-MSFT: should be

>return result;

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -895,113 +895,49 @@ internal static bool IsAdministrator()
...
+        internal static bool ItemExists(string path)
         {
-            bool unusedIsDirectory;
-            Exception unusedException;
-
-            return NativeItemExists(path, out unusedIsDirectory, out unusedException);
+            bool result = false;
+            try
+            {
+                result =  ItemExists(path, out bool unused);
```
SteveL-MSFT: extra whitespace after equals sign

SteveL-MSFT: perhaps use named parameter to make it more clear what `unused` is?

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -895,113 +895,49 @@ internal static bool IsAdministrator()
...
+            }
+            catch
+            {
+            }
+            return false;
         }
 
-        // This is done through P/Invoke since File.Exists and Directory.Exists pay 13% performance degradation
```
SteveL-MSFT: Are we sure we no longer have a perf penalty with this change?

SteveL-MSFT (changes requested):

src/System.Management.Automation/engine/Utils.cs
iSazonov: We have still a perf penalty (on windows because normalization(!) - one extra kernel call in .Net code) but it is tracked in CoreFX https://github.com/dotnet/corefx/issues/19717
We have already mentioned that we do up to 7 the normalizations of the same path per cmdlet call. So perf penalty is more in PowerShell code. I'm trying to find how we can reduce the path normalization perf penalty. 

Also please see https://github.com/dotnet/corefx/issues/27049#issuecomment-392826352

src/System.Management.Automation/engine/Utils.cs
iSazonov: `unused` here is self explaination - we don't use the parameter at all.

src/System.Management.Automation/engine/Utils.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/Utils.cs
iSazonov: Extra space fixed.

TravisEz13:
The Linux failure does not seem related to your change
https://travis-ci.org/PowerShell/PowerShell/jobs/385527642#L903
image
I'm restarting the CI run

iSazonov:
@TravisEz13 Thanks! But CIs failed again. I rebased - perhaps it resolve..

SteveL-MSFT (approved):

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -4111,7 +4111,7 @@ private ExternalScriptInfo FindLocalizedModuleManifest(string path)
 
                 String filePath = stringBuilder.ToString();
 
-                if (Utils.NativeFileExists(filePath))
+                if (File.Exists(filePath))
```
daxian-dbw: Why not use `Utils.FileExists` ?

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -233,7 +233,7 @@ internal static IEnumerable<string> GetDefaultAvailableModuleFiles(string topDir
                 foreach (string ext in ModuleIntrinsics.PSModuleExtensions)
                 {
                     string moduleFile = Path.Combine(directoryToCheck, proposedModuleName) + ext;
-                    if (!Utils.NativeFileExists(moduleFile))
+                    if (!File.Exists(moduleFile))
```
daxian-dbw: Same here.
Actually, I see `Utils.NativeFileExists` and `File.Exists` mixed in this file (possibly in that files too). Do we really care about the exception on failure in those cases?

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -895,113 +895,61 @@ internal static bool IsAdministrator()
...
+            catch
+            {
+            }
 
-            return NativeItemExists(path, out unusedIsDirectory, out unusedException);
+            return result;
         }
 
-        // This is done through P/Invoke since File.Exists and Directory.Exists pay 13% performance degradation
-        // through the CAS checks, and are terribly slow for network paths.
```
daxian-dbw: I'm fine with changing to `File.GetAttribute()` because `CAS` is not supported in .NET Core, so the 13% penalty is gone. So I think the biggest concern of using the .NET API is not there anymore.

However, can you check the `"network paths"` scenario?

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -895,113 +895,61 @@ internal static bool IsAdministrator()
...
 
-        internal static bool NativeItemExists(string path)
+        internal static bool ItemExists(string path)
         {
-            bool unusedIsDirectory;
-            Exception unusedException;
+            bool result = false;
+            try
+            {
+                result = ItemExists(path, out bool unused);
```
daxian-dbw: > ItemExists(path, out bool unused)

I suggest changing it to `ItemExists(path, out _)`, using the "discard" variable.

daxian-dbw: And, maybe

```
internal static bool ItemExists(string path)
{
    try { return ItemExists(path, out _) } catch { }
    return false;
}
```

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -895,113 +895,61 @@ internal static bool IsAdministrator()
...
-                ((int)NativeMethods.FileAttributes.Directory);
-
-            return true;
-#endif
         }
 
-        // This is done through P/Invoke since we pay 13% performance degradation
-        // through the CAS checks required by File.Exists and Directory.Exists
-        internal static bool NativeFileExists(string path)
+        internal static bool FileExists(string path)
```
daxian-dbw: I don't think all uses of this method expects an exception to be thrown. Some uses might just believe that it's faster than `File.Exists` but expect it to behave like `File.Exists()`.

I think you can inspect the uses of `Utils.FileExists` and `Utils.DirectoryExists` and change to `File.Exists` and `Directory.Exists` as appropriate.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -2921,15 +2891,15 @@ private void RemoveDirectoryInfoItem(DirectoryInfo directory, bool recurse, bool
...
                 }
 
-                bool isDirectory;
-                Exception accessException;
-
-                if (!Utils.NativeItemExists(directory.FullName, out isDirectory, out accessException))
+                try
                 {
-                    return;
+                    if (!Utils.ItemExists(directory.FullName, out bool unused))
```
daxian-dbw: Use the "discard" variable.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -2921,15 +2891,15 @@ private void RemoveDirectoryInfoItem(DirectoryInfo directory, bool recurse, bool
...
+                    {
+                        // Directory does not exist
+                        return;
+                    }
                 }
-
-                if (accessException != null)
+                catch (Exception accessException)
                 {
                     ErrorRecord errorRecord = new ErrorRecord(accessException, "RemoveFileSystemItemUnAuthorizedAccess", ErrorCategory.PermissionDenied, directory);
```
daxian-dbw: Looks to me it's expecting an `UnauthorizedAccessException`, shall we narrow the catch to only catch this type of exception?

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -3249,19 +3219,7 @@ private bool ItemExists(string path, out ErrorRecord error)
...
-                if (Utils.NativeItemExists(path, out notUsed, out accessException))
-                {
-                    result = true;
-                }
-
-                if (accessException != null)
-                {
-                    throw accessException;
-                }
+                result = Utils.ItemExists(path, out bool unused);
```
daxian-dbw: "discard" variable.

src/System.Management.Automation/utils/ClrFacade.cs
```diff
@@ -137,10 +137,10 @@ private static void EncodingRegisterProvider()
         internal static SecurityZone GetFileSecurityZone(string filePath)
         {
             Diagnostics.Assert(Path.IsPathRooted(filePath), "Caller makes sure the path is rooted.");
-            Diagnostics.Assert(Utils.NativeFileExists(filePath), "Caller makes sure the file exists.");
+            Diagnostics.Assert(Utils.FileExists(filePath), "Caller makes sure the file exists.");
             string sysRoot = System.Environment.GetEnvironmentVariable("SystemRoot");
             string urlmonPath = Path.Combine(sysRoot, @"System32\urlmon.dll");
-            if (Utils.NativeFileExists(urlmonPath))
+            if (Utils.FileExists(urlmonPath))
```
daxian-dbw: At least for these two uses, they do not expect exception to be thrown. So change them to `File.Exists` should be fine.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -895,113 +895,61 @@ internal static bool IsAdministrator()
...
-            }
+            bool itemExists = ItemExists(path, out bool isDirectory);
 
             return (itemExists && (!isDirectory));
         }
 
-        // This is done through P/Invoke since we pay 13% performance degradation
-        // through the CAS checks required by File.Exists and Directory.Exists
-        internal static bool NativeDirectoryExists(string path)
+        internal static bool DirectoryExists(string path)
```
daxian-dbw: I think we should change the name of this one and `FileExists` to something that indicate it might throw exception, so it's easy to tell whether you should use them instead of `File.Exists` and `Directory.Exists` when writing code.

I suggest rename them to `CheckFileExistsAndThrowOnError` and `CheckDirectoryExistsAndThrowOnError`.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -895,113 +895,61 @@ internal static bool IsAdministrator()
...
+            }
 
-            return NativeItemExists(path, out unusedIsDirectory, out unusedException);
+            return result;
         }
 
-        // This is done through P/Invoke since File.Exists and Directory.Exists pay 13% performance degradation
-        // through the CAS checks, and are terribly slow for network paths.
-        internal static bool NativeItemExists(string path, out bool isDirectory, out Exception exception)
+        internal static bool ItemExists(string path, out bool isDirectory)
```
daxian-dbw: Similarly, maybe changing this to `CheckItemExistsAndThrowOnError`.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -895,113 +895,61 @@ internal static bool IsAdministrator()
...
             {
-                int errorCode = Marshal.GetLastWin32Error();
-                if (errorCode == 5)
-                {
-                    // Handle "Access denied" specifically.
-                    Win32Exception win32Exception = new Win32Exception(errorCode);
-                    exception = new UnauthorizedAccessException(win32Exception.Message, win32Exception);
-                }
-                else if (errorCode == 32)
-                {
```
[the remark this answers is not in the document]
iSazonov: Please see the PR description (Fix is ready in [CoreFX](https://github.com/dotnet/corefx/issues/29782), we get it in 2.1.1.)

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
iSazonov: Utils.FileExists behavior is the same (no exception is thrown). So it is better to use the standard method.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
iSazonov: Utils.FileExists behavior is the same (no exception is thrown). So it is better to use the standard method.

Yes, we should review all these cases but there are dozens of them to do in this PR or in a new?

src/System.Management.Automation/engine/Utils.cs
iSazonov: The code come from Windows PowerShell and I don't know why it existed. I could not find any way where we would have throw on "network path not found" and not throw on "local path not found".
If we continue migration from ItemExists to File.Exists/Directory.Exists perhaps we found the amazing case.
Even if it is necessary we should eliminate the exception and replace it with regular check.
Now I suppose it's contrived

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -895,113 +895,61 @@ internal static bool IsAdministrator()
...
-                isDirectory = false;
                 return false;
             }
-#if UNIX
-            isDirectory = Platform.NonWindowsIsDirectory(path);
-            return Platform.NonWindowsIsFile(path);
-#else
-
+#if !UNIX
             if (IsReservedDeviceName(path))
```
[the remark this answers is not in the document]
iSazonov: Seems method in line 898 (internal static bool ItemExists(string path)) is used on Unix.

src/System.Management.Automation/engine/Utils.cs
```diff
@@ -895,113 +895,61 @@ internal static bool IsAdministrator()
...
-
-            int result = NativeMethods.GetFileAttributes(path);
-            if (result == -1)
+#endif
+            try
+            {
+                // Use 'File.GetAttributes()' because we want to get access exceptions.
+                // TODO: we should review the tricky logic
+                // (historically we throw 'UnauthorizedAccessException' here) and migrate
+                // to standard methods 'File.Exist()' and 'Directory.Exists()' where possible.
```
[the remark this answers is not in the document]
iSazonov: Seems my comment is not clear.
I meant that we should migrate from the  ItemExists method to File.Exists() and Directory.Exits() which don't throw UnauthorizedAccessException.

src/System.Management.Automation/engine/Utils.cs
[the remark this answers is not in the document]
iSazonov: Please see the PR description - we'll get the fix in .Net Core 2.1.1.

src/System.Management.Automation/engine/Utils.cs
iSazonov: I'll rename if we will not migrate to File.Exists/Directory.Exists in the PR.

src/System.Management.Automation/engine/Utils.cs
iSazonov: I'll rename if we will not migrate to File.Exists/Directory.Exists in the PR.

src/System.Management.Automation/namespaces/FileSystemProvider.cs

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: Here I only preserve the logic.
Really we should clean up the code too because .Net Core support reparse points and recursion https://github.com/dotnet/corefx/blob/f25eb288a449010574a6e95fe298f3ad880ada1e/src/System.IO.FileSystem/src/System/IO/FileSystem.Windows.cs#L359

Seems we should clean up all code in FileSystemProvider - perhaps we don't need FileSystemProviderV2 :-)

src/System.Management.Automation/namespaces/FileSystemProvider.cs

src/System.Management.Automation/utils/ClrFacade.cs
iSazonov: Fixed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: @daxian-dbw I opened #7017 to clean up the code.

src/System.Management.Automation/engine/Utils.cs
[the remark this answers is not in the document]
daxian-dbw: OK, so this will only affect `c:\pagefile.sys`?

src/System.Management.Automation/engine/Utils.cs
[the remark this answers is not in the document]
daxian-dbw: Yeah, I agree. I also pointed that out in my comments. Open an issue to track that migration instead of this comment. This comment is not useful in the context of this method, which is designed to throw on error.

daxian-dbw:
@iSazonov No, you don't need to do it in this PR. Open an issue to track the analysis and migration work (Utils.FileExists/DirectoryExists => File.Exists/Directory.Exists).

src/System.Management.Automation/engine/Utils.cs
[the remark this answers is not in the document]
daxian-dbw: Take a look at the method `IsReservedDeviceName`, it's body already takes care of the `#if/def`.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
daxian-dbw: It's better to make all the `Utils.NativeFileExists -> File/Directory.Exists` changes in a separate PR. So for now, just do the simple replacement of `Utils.NativeFileExists -> Utils.FileExists`.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
daxian-dbw: Yeah. let's do it in a separate PR. For this PR, just do the simple rename replacement: `Utils.NativeFileExists -> Utils.FileExists`.

src/System.Management.Automation/engine/Utils.cs
daxian-dbw: No, I was saying about the "terribly slow for network paths" part. Is "File.Exists and Directory.Exists" still terribly slow on network paths? I guess not as they all checking for attribute now, just like our implementation here. But better to check out.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
daxian-dbw: I'm fine leaving this catch as is.

src/System.Management.Automation/engine/Utils.cs
daxian-dbw: No, let's use a separate PR to do the migration changes. Go ahead rename this method.

src/System.Management.Automation/utils/ClrFacade.cs
daxian-dbw: Maybe revert these two changes too, since we agreed to use a separate PR for the migration changes.

src/System.Management.Automation/engine/Utils.cs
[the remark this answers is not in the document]
iSazonov: Three files - pagefile.sys, hiberfil.sys, swapfile.sys. In the CoreFX issue a comment says that they can not emulate the behavior and so use the real file c:\pagefile.sys in test. I think users are unlikely to find something like this

iSazonov:
Tracking Issue #7031

src/System.Management.Automation/engine/Utils.cs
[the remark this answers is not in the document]
iSazonov: Done #7031

src/System.Management.Automation/engine/Utils.cs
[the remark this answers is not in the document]
iSazonov: :-) I looked the other way.
Fixed.

src/System.Management.Automation/engine/Utils.cs
iSazonov: I see a delay only if I use a non-existent server name. I suppose this usual behavior with DNS name resolving.

src/System.Management.Automation/utils/ClrFacade.cs
iSazonov: Reverted.

src/System.Management.Automation/engine/Utils.cs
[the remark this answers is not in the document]
iSazonov: TODO comment was removed.

iSazonov:
@daxian-dbw File.Exists() and Directory.Exists() return false for device names so we can remove IsReservedDeviceName call from ItemExists() method at all. What do you think?

daxian-dbw:
@iSazonov `ItemExists` doesn't call `FileExists` or `Directory.Exists`. Do you mean `File.GetAttributes` returns -1 for reserved device names?

daxian-dbw:
The changes LGTM.
@iSazonov Are you going to rename `FileExists`, `DirectoryExists` and `ItemExists` to
`CheckFileExistsAndThrowOnError`, `CheckDirectoryExistsAndThrowOnError` and `CheckItemExistsAndThrowOnError` in this PR?

iSazonov:
> Are you going to rename FileExists, DirectoryExists and ItemExists to
CheckFileExistsAndThrowOnError, CheckDirectoryExistsAndThrowOnError and CheckItemExistsAndThrowOnError in this PR?

No, I'll remove the methods in follow PR.

>ItemExists doesn't call FileExists or Directory.Exists. Do you mean File.GetAttributes returns -1 for reserved device names?

Never mind - I hope to remove ItemExists() altogether in follow PR.

iSazonov:
@daxian-dbw Should I fix something else?

daxian-dbw (approved):
@iSazonov If you plan to remove those methods in your next PR, then it's fine to leave their names unchanged.

daxian-dbw:
@iSazonov It's preferred that you remove the native code from psl-native along with your managed code changes in your following PRs. For the PRs you already finished, I listed the tasks to track removing the corresponding native code in #6874.

iSazonov:
@daxian-dbw My thoughts was that if we have to re-publish psl-native after every change 
it is better to finish first the cleanup in the repo and only then cleanup in psl-native repo.

Issue #6855, opened 2018-05-11
Allow Measure-Object to accept ad-hoc script-block arguments in lieu of property names


mklement0:
<sup>This issue arose out of #6837.</sup>

Motivation:

`Sort-Object`'s `-Property` not only supports property _names_ (the typical case), but also _script blocks_ that are evaluated for each input object and thereby enable sorting criteria _derived_ from properties, which is a powerful feature (also found, in essence, in `Select-Object`); e.g.:

```powershell
# Sort by only the numerical part of the input, via a script-block parameter argument.
PS> 'z1', 'a0', 'b10' | Sort-Object -Property { [int] $_.Substring(1) }
a0
z1
b10
```

Suggestion:

It would be handy if `Measure-Object` supported analogous use; e.g.:

```powershell
# Count the  odd numbers in the input.
PS> 1, 2, 3 | Measure-Object -Sum -Property { $_ % 2 }  # wishful thinking
2
```

To that end:

* The `-Property` parameter's data type would have to change from`[string[]]` to `[object[]]`
* In the event of a script block getting passed, it would have to be evaluated for each input object and used in lieu of a simple property access on the input object.

Environment data
----------------

Written as of:

```powershell
PowerShell Core v6.0.2
```

BrucePay:
It's actually just a 3-line change with Property going from [string[]] to [MshExpression[]]. The downside is that MshExpression is an internal (and unfortunately named) class. So about 30 files need to be changed to do this properly (rename MshExpression -> PSPropertyExpression and make it public.)

mklement0:
Thanks, @BrucePay, but wouldn't changing the type to `[object[]]` do, accompanied by internal logic to see if the object passed is either a `[scriptblock]` or `[string]` (or something that can be coerced to a string)?

That's how `Sort-Object` seems to do it.

mklement0:
Thanks, @BrucePay. 

My point was that if `Sort-Object` types its `-Property` parameter `[object[]]` _publicly_ and uses `[MshExpression]` _internally_, can't `Measure-Object` do the same, thereby saving us from having to make [a renamed] `[MshExpression]`  public?

mklement0:
That's good to hear, @SteveL-MSFT.

We should consider doing the same thing not only for `Sort-Object`, but also for `Select-Object`.

Note, however, that additional considerations apply, because **both commands accept a _hashtable_ in addition to a script block**:

[...]

PR #6934, opened 2018-05-25
Fix for #6855 - Measure-Object should handle scriptblock properties.

BrucePay:
## PR Summary

Fixed issue #6855 by renaming the class that does all the work from `MshExpression` to `PSPropertyExpression`, making the class public and then in `MeasureObjectCommand`, lifting it up to the parameter level. Previously the implementation exposed the `-Property` as `string` then wrapped it internally as a `PSPropertyExpression`. Now the parameter type is `PSPropertyExpression` directly allowing for both wildcard strings and scriptblocks.  Note that this is technically a breaking change (the public type of the parameter has changed) but this was reviewed by the committee and approved. (The vast majority of changes in this PR are due to the renaming. The code changes are relatively small.)

`PSPropertyExpression` now lives in a public namespace where it can be used by cmdlet and script authors to easily add the same type of functionality to their commands. 

I also modified `PSPropertyExpression` to treat hashtables as objects so that
```
@{prop = 3} | measure-object prop
```
and
```
@{prop = 3} | measure-object {$_.prop}
```
will work the identically. (Previously the example using just the property name would just fail.)

#### Breaking Change Summary

This PR brings in a breaking change in an edge case:
> This could be a breaking change when it comes to the ConstrainedLanguage mode.
@{prop = 3} | measure-object Count works in ConstrainedLanguage mode today, but with this change, the conversion from Hashtable to PSCustomObject will throw exception.

It has been reviewed by the committee and agreed to accept this breaking change.
Please see the discussion at https://github.com/PowerShell/PowerShell/pull/6934#discussion_r193243793 for more information.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
+++ b/src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
@@ -10,47 +10,65 @@ using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Runtime.CompilerServices;
 
-namespace Microsoft.PowerShell.Commands.Internal.Format
+namespace Microsoft.PowerShell.Commands
 {
     /// <summary>
-    /// class to hold results
-    /// NOTE: we should make it an PSObject eventually
+    /// class that represents the results from evaluating a PSPropertyExpression against an object.
     /// </summary>
-    internal class MshExpressionResult
+    public class PSPropertyExpressionResult
     {
-        internal MshExpressionResult(object res, MshExpression re, Exception e)
+
+        /// <summary>
+        /// Create a property expression result containing the original object, matching property expression
+        /// and any exception generated during the match process.
+        /// </summary>
+        public PSPropertyExpressionResult(object res, PSPropertyExpression re, Exception e)
         {
             Result = res;
             ResolvedExpression = re;
             Exception = e;
         }
 
-        internal object Result { get; } = null;
+        /// <summary>
+        /// The value of the object property matched by this property expression.
+        /// </summary>
+        public object Result { get; } = null;
 
-        internal MshExpression ResolvedExpression { get; } = null;
+        /// <summary>
+        /// The original property expression fully resolved.
+        /// </summary>
+        public PSPropertyExpression ResolvedExpression { get; } = null;
 
-        internal Exception Exception { get; } = null;
+        /// <summary>
+        /// Any exception thrown while evaluating the expression.
+        /// </summary>
+        public Exception Exception { get; } = null;
     }
 
-    internal class MshExpression
+    /// <summary>
+    /// PSPropertyExpression class. This class is used to get the names and/or values of properties
+    /// on an object. A property expression can be constructed using either a wildcard expression string
+    /// or a scriptblock to use to get the property value.
+    /// </summary>
+    public class PSPropertyExpression
     {
         /// <summary>
         /// constructor
         /// </summary>
         /// <param name="s">expression</param>
         /// <exception cref="ArgumentNullException"></exception>
-        internal MshExpression(string s)
+        public PSPropertyExpression(string s)
             : this(s, false)
         {
         }
 
         /// <summary>
-        /// constructor
+        /// Create a property expression with a wildcard pattern.
         /// </summary>
-        /// <param name="s">expression</param>
+        /// <param name="s">Property name pattern to match.</param>
         /// <param name="isResolved"><c>true</c> if no further attempts should be made to resolve wildcards</param>
         /// <exception cref="ArgumentNullException"></exception>
-        internal MshExpression(string s, bool isResolved)
+        public PSPropertyExpression(string s, bool isResolved)
         {
             if (string.IsNullOrEmpty(s))
             {
@@ -61,11 +79,11 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
         }
 
         /// <summary>
-        /// constructor
+        /// Create a property expression with a ScriptBlock
         /// </summary>
-        /// <param name="scriptBlock"></param>
+        /// <param name="scriptBlock">ScriptBlock to evaluate when retrieving the property value from an object.</param>
         /// <exception cref="ArgumentNullException"></exception>
-        internal MshExpression(ScriptBlock scriptBlock)
+        public PSPropertyExpression(ScriptBlock scriptBlock)
         {
             if (scriptBlock == null)
             {
@@ -74,8 +92,14 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             Script = scriptBlock;
         }
 
+        /// <summary>
+        /// The ScriptBlock for this expression to use when matching.
+        /// </summary>
         public ScriptBlock Script { get; } = null;
 
+        /// <summary>
+        /// ToString() implementation for the property expression.
+        /// </summary>
         public override string ToString()
         {
             if (Script != null)
@@ -84,12 +108,20 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             return _stringValue;
         }
 
-        internal List<MshExpression> ResolveNames(PSObject target)
+        /// <summary>
+        /// Resolve the names matched by this the expression.
+        /// </summary>
+        /// <param name="target">The object to apply the expression against.</param>
+        public List<PSPropertyExpression> ResolveNames(PSObject target)
         {
             return ResolveNames(target, true);
         }
 
-        internal bool HasWildCardCharacters
+        /// <summary>
+        /// Indicates if the pattern has wildcard characters in it. If the supplied pattern was
+        /// a scriptblock, this will be false.
+        /// </summary>
+        public bool HasWildCardCharacters
         {
             get
             {
@@ -99,9 +131,14 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             }
         }
 
-        internal List<MshExpression> ResolveNames(PSObject target, bool expand)
+        /// <summary>
+        /// Resolve the names matched by this the expression.
+        /// </summary>
+        /// <param name="target">The object to apply the expression against.</param>
+        /// <param name="expand">If the matched properties are property sets, expand them.</param>
+        public List<PSPropertyExpression> ResolveNames(PSObject target, bool expand)
         {
-            List<MshExpression> retVal = new List<MshExpression>();
+            List<PSPropertyExpression> retVal = new List<PSPropertyExpression>();
 
             if (_isResolved)
             {
@@ -112,12 +149,16 @@ namespace Microsoft.PowerShell.Commands.Internal.Format
             if (Script != null)
             {
                 // script block, just add it to the list and be done
-                MshExpression ex = new MshExpression(Script);
+                PSPropertyExpression ex = new PSPropertyExpression(Script);
 
                 ex._isResolved = true;
                 retVal.Add(ex);
                 return retVal;
             }
+
+            // If the object passed in is a hashtable, then turn it into a PSCustomObject so
+            // that property expressions can work on it.
+            target = IfHashtableWrapAsPSCustomObject(target);
 
             // we have a string value
             IEnumerable<PSMemberInfo> members = null;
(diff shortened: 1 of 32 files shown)


iSazonov:
@BrucePay Thanks! I am extecting that the PR will marked by the label.
Seems we lost all MSFT team in recent weeks. I wonder.

test/powershell/Modules/Microsoft.PowerShell.Utility/Measure-Object.Tests.ps1
```diff
@@ -349,6 +349,72 @@ Describe "Measure-Object DRT basic functionality" -Tags "CI" {
...
+        $result.Count    | Should -Be 10
+        $result.Average  | Should -Be 55
+        $result.Sum      | Should -Be 550
+        $result.Minimum  | Should -Be 10
+        $result.Maximum  | Should -Be 100
+        $result.Property | Should -Be '$_ * 10'
+    }
+
+    It "Measure-Object with ScriptBlock properties should work with -word" {
+        $result = "a,b,c" | Measure-Object -Word  {$_ -split ','}
```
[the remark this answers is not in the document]
BrucePay: Fixed.

BrucePay:
Oops - I had the wrong issue number in the comment so you might have looked at the wrong issue. #6855 is, in fact, marked "Committee Reviewed".

test/powershell/Modules/Microsoft.PowerShell.Utility/Measure-Object.Tests.ps1
```diff
@@ -349,6 +349,72 @@ Describe "Measure-Object DRT basic functionality" -Tags "CI" {
...
+        $htables = @{foo = 1}, @{foo = 3}, @{foo = 10}
+        $result = $htables | Measure-Object -Sum {$_.foo * 10 }
+        $result.Sum | Should -Be 140
+    }
+
+    #
+    # Since PSPropertyExtression is now a public type, this function is used to test its
+    # operation as a parameter on a PowerShell function, independent of Measure-Object
+    #
+    function Test-PSPropertyExpression {
```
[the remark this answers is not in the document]
BrucePay: Moved into a separate Describe.

test/powershell/Modules/Microsoft.PowerShell.Utility/Measure-Object.Tests.ps1
```diff
@@ -349,6 +349,72 @@ Describe "Measure-Object DRT basic functionality" -Tags "CI" {
...
+        $result = 1..10 | Test-PSPropertyExpression {$_}
+        $result | Should -Be 55
+    }
+
+    It "Test-PropertyExpression function with a scriptblock property expression should be able to transform input" {
+        # Count the number of 'e's in the words.
+        $result = "one", "two", "three", "four", "five" | Test-PSPropertyExpression {($_.ToCharArray() -match 'e').Count}
+        $result | Should -Be 4
+    }
     It "Measure-Object with multiple lines should work"{
```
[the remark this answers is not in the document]
BrucePay: Added.

src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
```diff
@@ -246,25 +301,37 @@ private MshExpressionResult GetValue(PSObject target, bool eatExceptions)
...
                 }
             }
         }
+        
+        private PSObject IfHashtableWrapAsPSCustomObject(PSObject target)
+        {
+            // If the object passed in is a hashtable, then turn it into a PSCustomObject so
+            // that property expressions can work on it.
+            Hashtable targetAsHash = PSObject.Base(target) as Hashtable;
+            if (targetAsHash != null)
```
iSazonov: Could you please use C# 7 syntax?
```c#
if (PSObject.Base(target) as Hashtable targetAsHash) 
```

iSazonov (commented):
LGTM with one minor comment.

iSazonov (approved):

BrucePay:
@SteveL-MSFT Says we're ignoring CodeFactor issues for now.

SteveL-MSFT:
@iSazonov FYI - Most of my team has been involved with helping port existing Windows PowerShell modules owned by other teams to work with PSCore6 so you may find us not as active.  Hopefully we'll be back to fully engaged by end of June (or mostly engaged).

iSazonov:
I'm glad that this process began. I do not see this leading to corrections in PowerShell Core - it convinces me that PowerShell Core is very good! :smile:

src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
```diff
@@ -193,30 +234,44 @@ internal List<MshExpression> ResolveNames(PSObject target, bool expand)
...
+        /// <param name="expand">If the matched properties are parameter sets, expand them.</param>
+        /// <param name="eatExceptions">If true, any exceptions that occur during the match process are ignored.</param>
+        public List<PSPropertyExpressionResult> GetValues(PSObject target, bool expand, bool eatExceptions)
         {
-            List<MshExpressionResult> retVal = new List<MshExpressionResult>();
+            List<PSPropertyExpressionResult> retVal = new List<PSPropertyExpressionResult>();
+
+            // If the object passed in is a hashtable, then turn it into a PSCustomObject so
+            // that property expressions can work on it.
+            target = IfHashtableWrapAsPSCustomObject(target);
```
daxian-dbw: This could be a breaking change when it comes to the `ConstrainedLanguage` mode.
`@{prop = 3} | measure-object Count` works in `ConstrainedLanguage` mode today, but with this change, the conversion from `Hashtable` to `PSCustomObject` will throw exception.

src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
```diff
@@ -246,25 +301,36 @@ private MshExpressionResult GetValue(PSObject target, bool eatExceptions)
...
+                    return new PSPropertyExpressionResult(null, this, e);
                 }
                 else
                 {
                     throw;
                 }
             }
         }
+        
+        private PSObject IfHashtableWrapAsPSCustomObject(PSObject target)
```
daxian-dbw: Can you please change the name to `WrapHashtableAsPSCustomObject`? C# method naming conversion suggests using verbs or verb phrases to name methods.

src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
```diff
@@ -246,25 +301,36 @@ private MshExpressionResult GetValue(PSObject target, bool eatExceptions)
...
+        {
+            // If the object passed in is a hashtable, then turn it into a PSCustomObject so
+            // that property expressions can work on it.
+            if (PSObject.Base(target) is Hashtable targetAsHash)
```
daxian-dbw: If we want to handle hashtable specially, shall we expand that to `IDictionary`? `LanguagePrimitives.ConvertPSObjectToType` is able to convert `IDictionary` to `PSObject`.

daxian-dbw (changes requested):
/cc @PaulHigin Now that `PSPropertyExpression` is publically exposed and it has the ability to run scripts, Paul you may want to check if this will introduce any security concerns to the `ConstrainedLanguage` mode.

src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
BrucePay: The point of the rule is that the names of methods should be imperative which this is. There are other examples of (type 1) conditional clauses in .NET such as  `ThrowIfCancellationRequested` on `System.Threading.CancellationToken`.

src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
BrucePay: The goal is to match the behaviour of `[PSCustomObject] @{a=1; b=2}`. Currently casting `Dictionary<T,T>` to `PSCustomObject` doesn't result in  a custom object so I don't think we should introduce that behaviour here.

src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
BrucePay: Hmm - true - good catch. It's pretty obscure and currently the only useful scenario for piping hashtables into `measure-object` (other than instance-counting). Still, we should probably get a committee review on this.

src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
daxian-dbw: Understood. Good to know that we are matching the behavior of `[PSCustomObject] @{a=1; b=2}`.

src/System.Management.Automation/FormatAndOutput/common/Utilities/Mshexpression.cs
daxian-dbw: OK. #Closed

daxian-dbw (approved):
Waiting on the committee's review.

PaulHigin:
@daxian-dbw @BrucePay  Sorry for the late review.  To answer Dongbo's question, from a security point of view I think this is Ok.  MshExpression (now PSPropertyExpression) takes a ScriptBlock in its constructor which it will invoke.  But as long as the passed in ScriptBlock is created in a constrained environment it will be marked accordingly and the script block will be invoked with CL mode constraints.  

A constructor also takes a string "expression", but as near as I can tell it is never evaluated as script and is used to look up properties.  Is this correct?

However, we should include a CL mode test to ensure the expression scriptblock is run in CL mode.

daxian-dbw:
> A constructor also takes a string "expression", but as near as I can tell it is never evaluated as script and is used to look up properties. Is this correct?

@PaulHigin Yes, that's correct. It's for property lookup.

src/System.Management.Automation/engine/parser/TypeResolver.cs
```diff
@@ -756,6 +756,7 @@ internal static class CoreTypes
                     { typeof(PSListModifier),                              new[] { "pslistmodifier" } },
                     { typeof(PSObject),                                    new[] { "psobject", "pscustomobject" } },
                     { typeof(PSPrimitiveDictionary),                       new[] { "psprimitivedictionary" } },
+                    { typeof(PSPropertyExpression),                        new[] { "pspropertyexpression"} },
```
daxian-dbw: @PaulHigin I would like you to take another look at this change. `PSPropertyExpression` is added to `CoreTypes`, and that means users can call its methods in `ConstrainedLanguage` mode. Do you have any concerns about it?

PaulHigin: Thanks, I didn't see that.  If this type is going to be white listed then it needs to go through a separate/full security review.  I see that some internal methods are now public, and these along with constructors need to be reviewed for vulnerabilities.  Please set up a meeting for this.

PaulHigin (changes requested):

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and is fine with this breaking change which is likely only an edge case, but the positive impact for most users is beneficial

src/System.Management.Automation/engine/parser/TypeResolver.cs
BrucePay: No - it's only intended to be a type accelerator. I added it in the wrong place. I'll move it to the `TypeAccelerators()` method where it should be.

daxian-dbw: Yeah, I agree it should be moved :)

daxian-dbw:
@PaulHigin New commits have been pushed. Can you please update your review? Thanks

PaulHigin (approved):
LGTM

daxian-dbw:
@BrucePay I just noticed that we didn't run feature tests for this PR CI. Can you please push another commit with `[feature]` in the commit message? That will trigger the feature test run. You can also update the current head commit by running `git commit --amend` in your local `brucepay_MeasureObject` branch, and then forcing push to the remote by `git push --force-with-lease <your fork repo> brucepay_MeasureObject`. Thanks!

daxian-dbw:
@BrucePay I have submitted a dummy commit to trigger the feature test run for this PR.

Issue #6944, opened 2018-05-27
Speed up AppVeyor builds by using build matrix to split into separate, parallel tasks such as unit tests, system tests and full build of artifacts


bergmeister:
In the early days when `PowerShell` was still in Alpha state, the `AppVeyor` build took only 10 minutes, now it is nearly half an hour. This was very convenient for me as an early contributor where I could make minor changes without having to setup a dev environment locally and just download the MSI.
![image](https://user-images.githubusercontent.com/9250262/40586019-2a413df8-61b3-11e8-8ea5-371c4f2ef703.png)

The free version of AppVeyor can run only 1 job at a time, but using the paid account that is used for PR builds, we can run multiple jobs in parallel when using a [build matrix](https://www.appveyor.com/docs/build-configuration/#build-matrix). They were originally designed for easy building and testing against different configurations and platforms (`PSScriptAnalyzer` uses them as well [here](https://github.com/PowerShell/PSScriptAnalyzer/blob/development/appveyor.yml)) but we could still leverage that by splitting the whole build into independent chunks that will then run in parallel on separate AppVeyor VMs:
- Quick build and running the Pester unit test suite
- Running the XUnit tests
- Running the Pester system tests (e.g. JEA, Admin tests, etc.) that take quite a while
- Running the expensive feature tests
- Full Build (CrossGen, MSI , etc) that produces artifacts.

Using this approach, I guess we should be able to go back to times of 10-15 minutes and also receive more test/artifact results faster due to the parallel nature.

iSazonov:
A problem is that we should first build a crossgened release and then run tests with the build.

markekraus:
@bergmeister I'm Ok with this if it means that feature tests are run on every single PR. I think it presents a decent compromise between the need to ensure things aren't broken and making quick changes in markdown. Though, we'd need to ensure it doesn't make becoming a first time contributor to the project an overly burdensome experience too.

bergmeister:
@iSazonov 
> A problem is that we should first build a crossgened release and then run tests with the build.

As far as I know AppVeyor does not support a similar concept to VSTS with Build and Release tasks where the Release (test) tasks use the artifact from the build. To do [Parallel testing](https://www.appveyor.com/docs/parallel-testing/) in such a way, AppVeyor requires a Premium or Enterprise plan, hence why the free alternative would be to use a build matrix.
[...]

PR #6945, opened 2018-05-27
Use Appveyor matrix for faster PR builds

bergmeister:
## PR Summary

Related: #6944

Reduce PR build time by 5 minutes by:
- Having Packaging as a separate build job in a matrix -> runs in parallel in PR builds because the Microsoft account is a paid account that allows that (at no additional costs)
- Not caching the dotnet folder any more, which is too large and the overhead of zip/unzip/upload/download does not pay off (and fails in forked builds that are on a free AppVeyor account due to the size).
- Setting the environment variable `DOTNET_SKIP_FIRST_TIME_EXPERIENCE` to `1` because the initialization of the dotnet cli cache (1 minute) does not pay off for the whole build.

The total build time of builds on a fork that is on a free AppVeyor account and therefore do not have parallelism, remains the same due to the time saving of redundant caching.

This is just a simple example of what we can easily achieve, we could continue this pattern and split the test runs as per the referenced issue to bring PR builds down to 10 minutes (but this will incur an increase for fork builds on free AppVeyor accounts)

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**


iSazonov:
Seems we don't want publish packages if tests failed.

iSazonov:
Restoring build cache take 4 minutes! Unzipping - can we cache unzipped data?

bergmeister:
@iSazonov `fast_finish: true` would still be helpful in the sense that tests don't continue to run if the packaging build is broken. I don't see a reason for not publishing the packages even if some tests fail later on because if it is something that is just a prototype/experiment then I would not want to be blocked on that.
I do think that trying to cache dotnet does not pay off the zip/unzip/upload/download costs, instead of just installing the required dotnet SDK anyway.
I now set an environment variable `DOTNET_SKIP_FIRST_TIME_EXPERIENCE` to `1` to avoid unnecessary cache initialisation in this case, which saves around a 1 minute

iSazonov:
Make sense set DOTNET_SKIP_FIRST_TIME_EXPERIENCE in second job too?

iSazonov:
Seems we could decouple still two job - for admin tests and xUnit tests. If first we start CI-tagged tests and then packaging, admin-tagged tests and xUnit tests we get minimal time about 14 minutes.

bergmeister:
Yes, that's why this PR is not supposed to close the issue but yes, this is the way to go, I have outlined ideas like that in the issue. This PR is just the first step to show the direction without disadvantages for builds on a fork with a free account. Bear in mind that for every additional build job, people on non-free AppVeyor accounts incur a 2 minute penalty on their fork builds (i.e. their own builds before they have submitted a PR) due to it having to bootstrap  the environment.

TravisEz13:
I'm not sure this will speed things up.  It may actually slow things down as we often hit our maximum number of jobs.  We would need more budget to make this actually work.  @SteveL-MSFT   What do you think?

appveyor.yml
```diff
@@ -1,33 +1,37 @@
...
+    Purpose: BuildingAndTesting
 
-image: Visual Studio 2017
+# Stop all jobs in the matrix if any one them fail
+matrix:
+  fast_finish: true
 
-# cache version - netcoreapp.2.1-sdk.2.1.300-rc1
 cache:
-  - '%LocalAppData%\Microsoft\dotnet -> appveyor.yml'
```
TravisEz13: This is here for reliability.  Please don't remove

TravisEz13:
Does `DOTNET_SKIP_FIRST_TIME_EXPERIENCE` mean dotnet cli doesn't download the files?  We weren't caching the dotnet cache for perf but for reliability.

appveyor.yml
bergmeister: Are you saying the download of the SDK sometimes fails? Usually cache makes things less reliable due to leftovers. The reason for removal was the amount of time it was taking and the fact that it often also fails due to its size.

bergmeister:
`DOTNET_SKIP_FIRST_TIME_EXPERIENCE` skips the dotnet internal initial cache initialization (that is only useful for long term performance but not when one uses dotnet only a couple of times in a build) that always happens on the very first dotnet command after a fresh install. This process takes about a minute, which is time that we never get back.
For details see e.g. here:
http://donovanbrown.com/post/Stop-wasting-time-during-NET-Core-builds

appveyor.yml
bergmeister: Ok, but then shouldn't we rather cache just what we have downloaded instead of caching the appdata folder of dotnet?

appveyor.yml
```diff
@@ -1,33 +1,37 @@
 # version is set in tools\appveyor.psm1 - Invoke-AppVeyorInstall
+environment:
+  POWERSHELL_TELEMETRY_OPTOUT: 1
+  DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
```
TravisEz13: can we add this link somewhere near here
http://donovanbrown.com/post/Stop-wasting-time-during-NET-Core-builds

TravisEz13 (commented):
looks good if we add a comment about what the skip first-time experience variable means.

bergmeister:
@TravisEz13 Ok, done. What are your thoughts on the concern about reliability due to not caching the CLI any more?

appveyor.yml
bergmeister: cc @DarqueWarrior 😄

appveyor.yml
```diff
@@ -12,7 +12,7 @@ environment:
 matrix:
   fast_finish: true
 
-cache:
+# cache version - netcoreapp.2.1-sdk.2.1.300-rc1cache:
```
TravisEz13: You commented out the cache line but not the data

TravisEz13:
@bergmeister I think we can try it and see what happens.

appveyor.yml
```diff
@@ -1,33 +1,37 @@
...
+    Purpose: BuildingAndPackaging
+  - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+    Purpose: BuildingAndTesting
 
-image: Visual Studio 2017
+# Stop all jobs in the matrix if any one them fail
+matrix:
+  fast_finish: true
 
-# cache version - netcoreapp.2.1-sdk.2.1.300
```
daxian-dbw: Can you please leave this comment in place? It's needed to flush the nuget package cache when we update the sdk or runtime for the build.

daxian-dbw (approved):

appveyor.yml
TravisEz13: sorry, I didn't notice this.  Yes, please add this back

appveyor.yml
```diff
@@ -1,33 +1,37 @@
 # version is set in tools\appveyor.psm1 - Invoke-AppVeyorInstall
+environment:
+  POWERSHELL_TELEMETRY_OPTOUT: 1
+  DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1  # avoid expensive initialization of dotnet cli, see: http://donovanbrown.com/post/Stop-wasting-time-during-NET-Core-builds
```
iSazonov: I believe it is better to place comments on new line.

appveyor.yml
```diff
@@ -1,33 +1,37 @@
...
+  matrix:
+  - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+    Purpose: BuildingAndPackaging
+  - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+    Purpose: BuildingAndTesting
 
-image: Visual Studio 2017
+# Stop all jobs in the matrix if any one them fail
+matrix:
+  fast_finish: true
```
iSazonov: If we split jobs should we fast stop?

bergmeister: The idea was: if the build itself is broken, then both jobs will fails at the same stage because they both need to build. The only undesired case would be when adding/updating new 3rd party packages and packaging is broken then the tests would abort early.

appveyor.yml
bergmeister: Ok, added back

appveyor.yml
iSazonov: Thanks! 

Closed.

iSazonov (approved):

iSazonov:
@bergmeister Thanks for the contribution!

PR #6998, opened 2018-06-05
Use PSReadLine 2.0.0-beta2 from PSGallery

SteveL-MSFT:
## PR Summary

Removed building PSReadLine from repo.  Changed to downloading from PSGallery.

Addresses part of https://github.com/PowerShell/PowerShell/issues/1979

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**


test/powershell/engine/Help/UpdatableHelpSystem.Tests.ps1
```diff
@@ -148,8 +148,8 @@ $testCases = @{
         HelpFiles            = "Microsoft.PowerShell.PSReadline.dll-help.xml"
```
lzybkr: This should be `Microsoft.PowerShell.PSReadLine.dll-help.xml`.

build.psm1
```diff
@@ -2282,7 +2282,6 @@ function Start-CrossGen {
         "Microsoft.PowerShell.Security.dll",
         "Microsoft.PowerShell.CoreCLR.Eventing.dll",
         "Microsoft.PowerShell.ConsoleHost.dll",
-        "Microsoft.PowerShell.PSReadLine.dll",
```
lzybkr: I'd imagine this will hurt startup, but maybe it can't be cross-gen'd.

lzybkr: I think you should also remove the Windows specific dll `System.Runtime.InteropServices.RuntimeInformation.dll` from the nuget package - PowerShell Core ships the correct version of this file and this copy is just wasting disk space and I suppose could cause problems in the future.

test/powershell/engine/Help/UpdatableHelpSystem.Tests.ps1
SteveL-MSFT: Will fix

build.psm1
SteveL-MSFT: Had to remove from crossgen list as the crossgen happens before psmodulerestore.  We can reorder it if startup time is an issue.  The code to not include `System.Runtime.InteropServices.RuntimeInformation.dll` is already there at line 2395

lzybkr (approved):

kilasuit:
Whilst I am **delighted** that you are following up on #1979 and finally getting some traction on it, my only small concern is that the is a preview version of PSReadline. 

Would I be right in thinking that this is would be changed for the next non-preview release of pwsh to include the non-preview version of PSReadline with a future PR up the line?

daxian-dbw:
@SteveL-MSFT Console host tests are failing. I could be some recent changes broke them.

SteveL-MSFT:
@daxian-dbw my suspicion is that AppVeyor has PSReadLine 2.0.0-beta1 installed which is getting loaded first which has the bug that causes these tests to fail.  I added a test to check the path of where PSReadLine is imported from to validate this.  Not sure yet how to fix if this is the case since the semver string of the version gets dropped.

SteveL-MSFT:
My experiment shows that it is in fact 2.0.0-beta2 being used in AppVeyor.  So not sure why the test is failing in AppVeyor, but passes locally other than the console is probably different.  I think we still want to get 2.0.0-beta2 into Preview3 so we can get more feedback on PSReadLine 2.0.  I propose we make these tests pending for now.

iSazonov:
Therefore, these scripts will not work in environments without a console. What are these scenarios? Task Scheduler?

kilasuit:
Whilst I am **delighted** that you are following up on #1979 and finally getting some traction on it, my only small concern is that the is a preview version of PSReadline. 

Would I be right in thinking that this is would be changed for the next non-preview release of pwsh to include the non-preview version of PSReadline with a future PR up the line?

SteveL-MSFT:
@iSazonov Not entirely sure about the root cause yet, but just hypothesizing that it's due to console.  If so, Task Scheduler would have similar issue, but redirecting stdin isn't a common scenario.

iSazonov:
@SteveL-MSFT Thanks! If we plan to release next Preview I agree to merge the PR.

PR #7021, opened 2018-06-07
AppVeyor build matrix: more efficient build job split to reduce total time by another 5 minutes

### LINKED ISSUE #6944: Speed up AppVeyor builds by using build matrix to split into separate, parallel tasks such as unit tests, system tests and full build of artifacts

bergmeister:
## PR Summary

Closes #6944 

Following PR #6945 , further reduce the total build time (without any disadvantages) by around **5 minutes** by making sure there is a more even split between the 2 build jobs (the 2nd build job used to be much shorter).
Therefore this PR moves also the `xUnit` and `Pester`-Admin tests into the 2nd build job. If it is a daily/feature test commit, then the feature tests will also happen (only) in the 2nd build job. Because both jobs now run tests, the `failfast` option was removed. The final question from my side is whether running tests in 2 build jobs is OK for the daily build, which uploads code coverage results?

The time to wait for the AppVeyor build results is now 15 +/- 2 minutes, which is a huge improvement to what used to be around 28 minutes before the build matrix was introduced.

In the future, we could possibly run the feature tests in an optional 3rd build job once `AppVeyor` implements conditional build matrices here: https://github.com/appveyor/ci/issues/2299

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,10 @@ environment:
     Purpose: BuildingAndPackaging
   - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
     Purpose: BuildingAndTesting
+    TestCategory: PesterCI
+  - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
+    Purpose: BuildingAndTesting
+    TestCategory: PesterAdminAndOptionallyFeatureTests_xUnit
 
 # Stop all jobs in the matrix if any one them fail
 matrix:
@@ -29,11 +33,11 @@ build_script:
   - ps: if ($env:Purpose -eq 'BuildingAndTesting'){ Invoke-AppveyorBuild }
 
 test_script:
-  - ps: if ($env:Purpose -eq 'BuildingAndTesting'){ Invoke-AppveyorTest }
+  - ps: if ($env:Purpose -eq 'BuildingAndTesting'){ Invoke-AppveyorTest $env:TestCategory }
 
 after_test:
-  - ps: if ($env:Purpose -eq 'BuildingAndTesting'){ Invoke-AppVeyorAfterTest }
+  - ps: if ($env:Purpose -eq 'BuildingAndTesting'){ Invoke-AppVeyorAfterTest $env:TestCategory }
 
 # Run the finish command and delete files which should not be in the cache
 on_finish:
-  - ps: if ($env:Purpose -eq 'BuildingAndPackaging'){ Invoke-AppveyorFinish }
+  - ps: if ($env:Purpose -eq 'BuildingAndPackaging'){ Invoke-AppveyorFinish $env:TestCategory }
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,10 +29,10 @@ build_script:
   - ps: Invoke-AppveyorBuild
 
 test_script:
-  - ps: Invoke-AppveyorTest -TestCategory $env:TestCategory
+  - ps: Invoke-AppveyorTest -Purpose $env:Purpose
 
 after_test:
-  - ps: Invoke-AppVeyorAfterTest -TestCategory $env:TestCategory
+  - ps: Invoke-AppVeyorAfterTest -Purpose $env:Purpose
 
 # Packaging
 on_finish:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,7 +28,7 @@ test_script:
   - ps: Invoke-AppveyorTest -Purpose $env:Purpose
 
 after_test:
-  - ps: Invoke-AppVeyorAfterTest -Purpose $env:Purpose
+  - ps: Invoke-AppVeyorAfterTest
 
 # Packaging
 on_finish:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -40,4 +40,4 @@ after_test:
 
 # Run the finish command and delete files which should not be in the cache
 on_finish:
-  - ps: if ($env:Purpose -eq 'BuildingAndPackaging'){ Invoke-AppveyorFinish $env:TestCategory }
+  - ps: if ($env:Purpose -eq 'BuildingAndPackaging'){ Invoke-AppveyorFinish }
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,10 +9,6 @@ environment:
   - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
     Purpose: AdvancedPesterTests_xUnit_Packaging
 
-# Stop all jobs in the matrix if any one them fail
-matrix:
-  fast_finish: true
-
 # cache version - netcoreapp.2.1-sdk.2.1.300
 cache:
   - '%HOMEDRIVE%%HOMEPATH%\.nuget\packages -> appveyor.yml'
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,9 +5,9 @@ environment:
   DOTNET_SKIP_FIRST_TIME_EXPERIENCE: 1
   matrix:
   - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
-    Purpose: PesterCI
+    Purpose: UnelevatedPesterTests
   - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
-    Purpose: AdvancedPesterTests_xUnit_Packaging
+    Purpose: ElevatedPesterTests_xUnit_Packaging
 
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -5,13 +5,9 @@ environment:
...
   - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
-    Purpose: BuildingAndPackaging
+    Purpose: PesterCI
   - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
-    Purpose: BuildingAndTesting
-
-# Stop all jobs in the matrix if any one them fail
-matrix:
-  fast_finish: true
+    Purpose: AdvancedPesterTests_xUnit_Packaging
```
daxian-dbw: > AdvancedPesterTests

How about `ElevatedPesterTests`

bergmeister: Could be but keep in mind that it is not only the admin tests, when the `feature` keyword is specified in a commit then it it also `'Slow', 'Feature', 'Scenario'` categories.

daxian-dbw: Yes, it only indicates running Pester tests in an elevated process, not implying what tags of tests are run.

daxian-dbw: Actually, `PesterCI` part only runs a sub set of PR CI tests -- it doesn't run tests with tag `-Tags
@('CI', 'RequireAdminOnWindows', 'RequireSudoOnUnix')`, or when `[Feature]` is sepecifed, it runs `'slow'`, `'feature'`, `'scenario'` tests as long as they don't have `'RequireAdminOnWindows'`. So to be precise, this purpose probably can be renamed to `UnelevatePesterTests`.

TravisEz13:
@bergmeister  can you push a commit with `[Feature]` at this beginning on it.  This will cause the CI to run like our nightly build which runs more tests.

bergmeister:
OK, it seems, when using `[feature]`, there are only around 800 tests in the 2nd build matrix but there should be nearly 2000, I need to look into that, maybe I have missed one condition in the test function

TravisEz13:
@bergmeister  thanks for investigating the results

bergmeister:
@TravisEz13 I found the reason for the missing unelevated Pester feature tests and I think I fixed it (currently waiting for results).
One important question that might be a blocker: In `Invoke-AppVeyorAfterTest`, code coverage results get uploaded. Is this a problem now with test results being spread across 2 build jobs?

TravisEz13:
@bergmeister  The last feature build was still 678 tests short... I'm trying to take a quick look why...
[Feature build with 7975 tests](https://ci.appveyor.com/project/PowerShell/powershell/build/v6.1.0-preview.9647)
[Your feature build with 7296 test](https://ci.appveyor.com/project/PowerShell/powershell/build/v6.1.0-preview.9645)

TravisEz13:
With the Feature tag, the elevated tests should be 633, while only 114 where run with your PR

bergmeister:
Thanks for pointing this out, I will look into this.

bergmeister:
@TravisEz13 I pushed another `[feature]` commit and this time the build results [here](https://ci.appveyor.com/project/PowerShell/powershell/build/v6.1.0-preview.9664) were fine (i.e. had roughly the expected number of tests with 7974).
I can only speculate what went wrong but the feature build that you referenced was triggered by a push that included 3 commits (a merge commit to pull upstream changes in, a commit from myself and then the empty feature commit at last), which might have confused AppVeyor...
What do you think about my [question](https://github.com/PowerShell/PowerShell/pull/7021#issuecomment-395853603) about code coverage of the daily build?

TravisEz13:
@bergmeister Code coverage does not use the AppVeyor workflow so this change should not be an issue.

TravisEz13:
@bergmeister  I was discussing the impact of FailFast with other team members.  @JamesWTruher  Mentioned that we would like to always have logs of tests results (even if it is that tests were canceled.)   FailFast defeats this and we don't get logs to any canceled tests.  Could you disable FailFast?  We don't view this as a critical part of this optimization.

bergmeister:
@TravisEz13 OK. Sounds good. This PR already has `FailFast` removed, I am thinking the same way, especially now with tests running in both build jobs.

TravisEz13:
@bergmeister sorry I messed that.

TravisEz13 (approved):

TravisEz13:
@bergmeister 
Travis-CI has a similar feature https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix
We are considering refactoring Travis-CI in a similar way and enabling the feature tests by default on both systems.  This would eliminate the most of the time gains you have given us but give us better coverage in the CI system.  This is just an FYI in case you are interested in contributing.

TravisEz13:
@bergmeister  And thanks for this work.

bergmeister:
@TravisEz13 You're welcome, I quite enjoy working with CI systems and tweaking them. I also like it now to be able to get complete results now again in 15 minutes, which is close to how it was when PowerShell was in Alpha state.
I have not worked with `Travis` yet myself and having `AppVeyor` fast is sufficient for me because I am now more interested on Windows as the platform and rarely have PRs that break non-Windows platforms. Therefore don't expect a PR from my side at the moment.

If you think about always running feature tests, then I think it would be the best to do that in an optional 3rd build job once `AppVeyor` implements conditional build matrices here: https://github.com/appveyor/ci/issues/2299

It would be interesting if you guys could setup a Travis CI build for PSSA to get coverage for Mac (AppVeyor now officially supports Ubuntu images and I can tell from experience that they work very well, are very fast and the AppVeyor support is awesome, just a few days ago they fixed a bug in production for me in less than an hour or me raising it). You could therefore consider moving the Linux tests to AppVeyor.

TravisEz13:
@bergmeister 
> If you think about always running feature tests, then I think it would be the best to do that in an optional 3rd build job once AppVeyor implements conditional build matrices here: appveyor/ci#2299

I don't think we would wait for conditional matrices.  We would just add another job with no condition that would run all the time.

> AppVeyor now officially supports Ubuntu images and I can tell from experience that they work very well ...

I have discussed with other team members moving the PowerShell Core Linux tests from Travis-CI to AppVeyor.  I agree that they seem faster and I prefer the AppVeyor feature set and support over Travis-CI.  If we were to move to AppVeyor, we would probably run both for some time to prove that they are reliable.  Currently, I have the PowerShell-Docker tests all running in AppVeyor as testing on Linux should be equivalent of testing on macOS for what we are delivering in that repo.

bergmeister:
@TravisEz13 Yes, just running the feature tests in a 3rd job would be OK for me as well.
Just as an aside/idea, one could also consider dogfooding VSTS hosted images that support all 3 platforms and VSTS currently has a preview feature of offering a public read-only view. What would be great about VSTS is that one only needs to build once and can then kick off multiple release definitions with the artifacts.

TravisEz13:
@bergmeister We have considered VSTS.  I have already reserved `powershell.visualstudio.com`.  One issue I'm aware of, VSTS Hosted Linux uses an Ubuntu container not a VM and our tests currently have issues with containers.  There is a PR to get this partially working https://github.com/PowerShell/PowerShell/pull/6953.  I doubt the author will do more work.  I'm fine as long as the build works, we tag tests somehow as not working with docker and file an issue to fix the tests (hopefully with some sort of breakout into smaller pieces.)

I plan to move PowerShell-Docker to VSTS when I get the issues worked out (the CI builds aren't running even though the webhook fires.)

Issue #7022, opened 2018-06-08
Cursor is not set properly so lots of spaces are left out there after showing progress bar


jianyunt:
Unlike the filesystem which all the files and directories already exist in our system, SHiPS Azure provider initially does not have any data in the local system to search for . When a user is navigating around, it reaches out to Azure for fetching data and builds up the navigation tree.  In the SHiPS case, when a user cd and then tab, or get-azurerm tab, the progress bar shows up while fetching data. After write-progress completes,  it lefts a large spaces.

![image](https://user-images.githubusercontent.com/10171679/41133134-f8676f9e-6a79-11e8-8f77-70d269b495a6.png)

Initial Analysis
---------
Both the PowerShell and PSReadline are contributing to this issue.  PowerShell does not set the cursor to the place where it starts while  the PSReadline does not update the cursor position in case the screen has been scrolled up for displaying progress messages.

Steps to reproduce
------------------
portal.azure.com
select bash
type pwsh
cd tab

Expected behavior
-----------------
After write-progress completes the cursor should be the initial command line place, in this case "cd".

Actual behavior
---------------

Left lots of spaces used for showing progress bar.

Environment data
----------------

6.1.0-preview.2
and 
PSReadLine 2.0.0-Beta.1 or PSReadLine 2.0.0-Beta.2

jianyunt:
This issue is fixed in 6.1 GA.

PR #7023, opened 2018-06-08
Set the cursor to the place where a user hits tab key

jianyunt:
## PR Summary

Addressed the issue https://github.com/PowerShell/PowerShell/issues/7022
For the PSReadLine part of issue, see PR https://github.com/lzybkr/PSReadLine/pull/708

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfaceProgress.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfaceProgress.cs
@@ -99,7 +99,7 @@ namespace Microsoft.PowerShell
                 _progPane.Show(_pendingProgress);
 
                 // Reset the cursor back to where it started
-                if(record.RecordType == ProgressRecordType.Completed)
+                if (record.RecordType == ProgressRecordType.Completed)
                 {
                     _progPane.Hide();
                 }
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfaceProgress.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfaceProgress.cs
@@ -97,6 +97,12 @@ namespace Microsoft.PowerShell
                 // Update the progress pane only when the timer set up the update flag or WriteProgress is completed.
                 // As a result, we do not block WriteProgress and whole script and eliminate unnecessary console locks and updates.
                 _progPane.Show(_pendingProgress);
+
+                // Reset the cursor back to where it started
+                if(record.RecordType == ProgressRecordType.Completed)
+                {
+                    _progPane.Hide();
+                }
             }
         }
 
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ProgressPane.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ProgressPane.cs
@@ -99,12 +99,20 @@ namespace Microsoft.PowerShell
 
                 //if the cursor is at the bottom, create screen buffer space by scrolling
                 int scrollRows = rows - ((_rawui.BufferSize.Height - 1) - _location.Y);
-                for (int i = 0; i < rows; i++)
-                {
-                    Console.Out.Write('\n');
-                }
                 if (scrollRows > 0)
                 {
+                    // The following can be possibly replaced by Console.Write ("\x1b[" + scrollRows + "S");
+                    // For details, see https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences
+
+                    // Scroll the console screen up by 'scrollRows'
+                    var bottomLocation = _location;
+                    bottomLocation.Y = _rawui.BufferSize.Height;
+                    _rawui.CursorPosition = bottomLocation;
+                    for (int i = 0; i < scrollRows; i++)
+                    {
+                        Console.Out.Write('\n');
+                    }
+
                     _location.Y -= scrollRows;
                     _savedCursor.Y -= scrollRows;
                 }
--- a/src/System.Management.Automation/engine/MshCommandRuntime.cs
+++ b/src/System.Management.Automation/engine/MshCommandRuntime.cs
@@ -2252,7 +2252,7 @@ namespace System.Management.Automation
             {
                 // Only generate these exceptions if a pipeline has already been declared as the 'writing' pipeline.
                 // Otherwise, these are probably infrastructure messages and can be ignored.
-                if (this.PipelineProcessor._permittedToWrite != null)
+                if (this.PipelineProcessor?._permittedToWrite != null)
                 {
                     throw PSTraceSource.NewInvalidOperationException(
                         PipelineStrings.WriteNotPermitted);


src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfaceProgress.cs
```diff
@@ -97,6 +97,12 @@ class ConsoleHostUserInterface : System.Management.Automation.Host.PSHostUserInt
                 // Update the progress pane only when the timer set up the update flag or WriteProgress is completed.
                 // As a result, we do not block WriteProgress and whole script and eliminate unnecessary console locks and updates.
                 _progPane.Show(_pendingProgress);
+
+                // Reset the cursor back to where it started
+                if(record.RecordType == ProgressRecordType.Completed)
+                {
+                    _progPane.Hide();
```
iSazonov: We do the hide at a pipeline end.

jianyunt: You are right. I observed that the hide() gets called when write-progress completes for cmdlets. However for the cloudshell case, the SHiPS provider calls write-progress when a user hits get tab for example;  it then calls a Azure cmdlet that triggers write-progress too. When the Azure cmdlet completes, the hide() gets called. PowerShell handles it properly. But when the WriteProgress for ProgressRecordType.Completed gets called from the SHiPS provider, [Show()](https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfaceProgress.cs#L99) gets called again. At that time, record.RecordType is ProgressRecordType.Completed. After that, no more hide() gets called. 

image

Call stack:
Microsoft.PowerShell.ConsoleHost.dll!Microsoft.PowerShell.ProgressPane.Show() Line 95
Microsoft.PowerShell.ConsoleHost.dll!Microsoft.PowerShell.ProgressPane.Show(Microsoft.PowerShell.PendingProgress pendingProgress) Line 250
Microsoft.PowerShell.ConsoleHost.dll!Microsoft.PowerShell.ConsoleHostUserInterface.HandleIncomingProgressRecord(long sourceId, System.Management.Automation.ProgressRecord record) Line 99
Microsoft.PowerShell.ConsoleHost.dll!Microsoft.PowerShell.ConsoleHostUserInterface.WriteProgress(long sourceId, System.Management.Automation.ProgressRecord record) Line 1324
System.Management.Automation.dll!System.Management.Automation.Internal.Host.InternalHostUserInterface.WriteProgress(long sourceId, System.Management.Automation.ProgressRecord record) Line 609
System.Management.Automation.dll!System.Management.Automation.MshCommandRuntime.WriteProgress(long sourceId, System.Management.Automation.ProgressRecord progressRecord, bool overrideInquire) Line 413
System.Management.Automation.dll!System.Management.Automation.MshCommandRuntime.WriteProgress(System.Management.Automation.ProgressRecord progressRecord, bool overrideInquire) Line 354
System.Management.Automation.dll!System.Management.Automation.MshCommandRuntime.WriteProgress(System.Management.Automation.ProgressRecord progressRecord) Line 324
System.Management.Automation.dll!System.Management.Automation.Cmdlet.WriteProgress(System.Management.Automation.ProgressRecord progressRecord) Line 560
System.Management.Automation.dll!System.Management.Automation.CmdletProviderContext.WriteProgress(System.Management.Automation.ProgressRecord record) Line 851
System.Management.Automation.dll!System.Management.Automation.Provider.CmdletProvider.WriteProgress(System.Management.Automation.ProgressRecord progressRecord) Line 1794
Microsoft.PowerShell.SHiPS.dll!Microsoft.PowerShell.SHiPS.ProgressTracker.End(CodeOwls.PowerShell.Provider.PathNodeProcessors.IProviderContext context) Line 56
Microsoft.PowerShell.SHiPS.dll!Microsoft.PowerShell.SHiPS.PSScriptRunner.InvokeScriptBlock(CodeOwls.PowerShell.Provider.PathNodeProcessors.IProviderContext context, Microsoft.PowerShell.SHiPS.SHiPSDirectory node, Microsoft.PowerShell.SHiPS.SHiPSDrive drive) Line 87
Microsoft.PowerShell.SHiPS.dll!Microsoft.PowerShell.SHiPS.ContainerNodeService.GetNodeChildrenInternal(CodeOwls.PowerShell.Provider.PathNodeProcessors.IProviderContext context) Line 153

iSazonov: @jianyunt Thanks for great clarify! If I understood correctly SHiPS provider does not create a pipeline. I'm still afraid it's a wrong fix - it might break something - we don't have tests to catch this. Perhaps the fix should be in SHiPS provider.

In general 'ProgressRecordType.Completed' status can be important to show a final message to user. With the fix we remove the message. I guess we shouldn't do that.

/cc @daxian-dbw @lzybkr Could you please look the issue?

lzybkr: Maybe completion, or even the call to `PSConsoleHostReadLine` (which is how PSReadLine is called from the console host) should set `$ProgressPreference = 'SilentlyContinue'`.

jianyunt: We thought about turning on & off echo for Write-Progress.  @BrucePay  could you please remind me the story about it?
Fetching data from Azure takes time,  so we need to show the progress.

jianyunt: Actually turning echo off, in this case, does not help.  Turn off Progress will help but we cannot do it because of network calls to Azure.

iSazonov: >Perhaps the fix should be in SHiPS provider.

@jianyunt Could you please clarify - how does SHiPS provider implement the progress bar? Why it doesn't remove the bar itself after finishing current operation?

jianyunt:
@iSazonov SHiPS is written in C#. It's a PowerShell provider. it calls into PowerShell scripts for fetching data. For example, Get-AzureRMWebApp.  Before calling PowerShell script from C#, it creates [ProgressRecord and call WriteProgress](https://github.com/PowerShell/SHiPS/blob/development/src/Microsoft.PowerShell.SHiPS/Utility/ProgressTracker.cs#L27), then Get-AzureRMWebApp get called which generates a few more write-progress from it's own cmdlet. When the cmdlet completes, SHiPS gets the data returned from Get-AzureRMWebApp. It then calls WriteProgress with ProgressRecordType.Completed from [here](https://github.com/PowerShell/SHiPS/blob/development/src/Microsoft.PowerShell.SHiPS/SHiPSScriptRu)nner.cs#L85.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfaceProgress.cs
jianyunt: @iSazonov SHiPS is written in C#. It's a PowerShell provider. it calls into PowerShell scripts for fetching data. For example, Get-AzureRMWebApp. Before calling PowerShell script from C#, it creates ProgressRecord and call WriteProgress, then Get-AzureRMWebApp get called which generates a few more write-progress from it's own cmdlet. When the cmdlet completes, SHiPS gets the data returned from Get-AzureRMWebApp. It then calls WriteProgress with ProgressRecordType.Completed from herenner.cs#L85.

jianyunt: The progress bar is removed, but the cursor does not set back to the right location.  Before the progressbar shows, we save the [cursor position](https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.ConsoleHost/host/msh/ProgressPane.cs#L87). Hide() restores the cursor position. The Hide() I added here is mainly for restoring the cursor position.

I think for the cmdlets  like Get-AzureRMWebApp that has multiple children of write-progress, we see the problem. I noticed the cmdlet that has only one write-progress (no child progress) does not have the problem. 

Also I do not think this is SHiPS specific. I guess any app written in C#, writeprogress(), call powershell script (parent, child, grant child write-progess), will hit the issue.

---
In reply to: [195180383](https://github.com/PowerShell/PowerShell/pull/7023#discussion_r195180383) [](ancestors = 195180383)

iSazonov: @jianyunt Thanks for clarify. Previously I thought only about pipelines.
[Docs](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/write-progress?view=powershell-6#optional-parameters) says that `Completed` should hide the progress bar. And seems we do this - with `_pendingProgress.Update(sourceId, record);` we remove the progress bar from buffer and with `_progPane.Show(_pendingProgress);` we hide this from screen if render context == null - it should be null after previous update if we haven't nested progress bars.

BrucePay (approved):

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfaceProgress.cs
BrucePay: @iSazonov The docs you referenced describe the behaviour of the `Write-Progress` cmdlet, not the APIs. Internally, `Show()` and `Hide()` get called a lot (like every time you write a line to the console.) @jianyunt's change to call `Hide()` makes sense since the `ProgressRecordType` is `Completed`. Note that this code path doesn't get hit in the cmdlet scenario so I think it's actually fixing a latent bug.

iSazonov: I agree that Completed must remove the progress bar. And our code do this right.  The problem is "the cursor does not set back to the right location". I'm still not sure if this fix is correct. If we look [Show()](https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.ConsoleHost/host/msh/ProgressPane.cs#L131) it restore the cursor location on Unix but not on Windows - I guess the fix must be here.
Also we call Interlocked.CompareExchange(), then Show() and then immediately Hide() - seems we could use more short code path for Completed.

daxian-dbw: Oops, it looks I merged it too earlier ... there are still concerns about this PR that I didn't notice.
Please continue the discussion. If needed, we can reopen the issue and revert this change.

jianyunt: @iSazonov I think here is what [Show()](https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.ConsoleHost/host/msh/ProgressPane.cs#L131) is doing for Unix. 

- Save the current cursor position to _savedCursor  so it can be restored once writeprogress completes
-  calculate spaces needed for displaying the progress messages. If needed, it will scroll screen which will change the cursor position, right?
- clear the region for progress message 
- [set the cursor position to the start of the progress pane](https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.ConsoleHost/host/msh/ProgressPane.cs#L131). This is not set the cursor to the original command line a user typed.
- show the progress messages.

The _savedCursor  saved in Show() gets restored in Hide().

iSazonov: Thanks for clarify. You are right.

>The _savedCursor saved in Show() gets restored in Hide().

Does this mean that after SHow() we have an erroneous cursor position until we call Hide()? If so it is a bug.

And for Completed we call Hide() in the line
https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.ConsoleHost/host/msh/ProgressPane.cs#L206

jianyunt: @iSazonov yes `if (contents == null)` is false in this case, so we end up in line [Show()](https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.ConsoleHost/host/msh/ProgressPane.cs#L247). After that no one calls Hide() to restore the cursor position. We possibly do not need to call Show() if Complete. But that will be more risky since the code has ben there for a long time and maybe for reasons I do not know. Calling Hide() when progress complete is safer, I think.

iSazonov: > We possibly do not need to call Show() if Complete. 

We should call because of nested progress bar.
The code below shows two problems with current fix:
1. After complete nested progress bar we do Hide() and if we don't update parent progress bar long time users will not see the progress bar on screen. It is on `Sleep 5` in code sample.
2. If we press Ctrl-C when code sample is _after_ `Sleep 5` and when parent bar is on screen again - the pipe exit and the screen is _not cleared_ from parent progress bar.
```powershell
0..10 | % {Write-Progress -Id 0 -Activity 11 -PercentComplete $_; Sleep 1; 0..10 | % { Write-Progress -Activity 22 -Id 1 -ParentId 0 -PercentComplete $_; Sleep 1 };Write-Progress -Activity 22 -Id 1 -Completed }
```
I think we should revert this change. 
I guess new fix should be here
https://github.com/PowerShell/PowerShell/blob/master/src/Microsoft.PowerShell.ConsoleHost/host/msh/ProgressPane.cs#L241-L252

jianyunt: I have tried your sample code. Indeed there is a time window press Ctrl-C the screen does not cleared. This means we should not call [_propane.Hide()](https://github.com/PowerShell/PowerShell/pull/7023/files#diff-9ef9b6a033cd10812212cf4c872787c8R104). Other changes  in this PR are good. But I am ok to revert them all. @daxian-dbw, can you take care of it?

daxian-dbw: I will submit a PR to revert changes in this PR.
@iSazonov Thanks for being so thorough in code review! I appreciate it 👍

jianyunt: @iSazonov, please go ahead for the fix as I am tied up with other stuff now. Thank you again for the thorough code review.

iSazonov: I can not repo this:
>When the Azure cmdlet completes, the hide() gets called. PowerShell handles it properly. But when the WriteProgress for ProgressRecordType.Completed gets called from the SHiPS provider, Show() gets called again. At that time, record.RecordType is ProgressRecordType.Completed. 

with script:
```powershell
0..1 | % {Write-Progress -Id 0 -Activity 11 -PercentComplete $_;
          Sleep 1;
          0..1 | % { Write-Progress -Activity 22 -Id 1 -PercentComplete $_; Sleep 1 };
          Write-Progress -Activity 22 -Id 1 -Completed };
Write-Progress -Activity 11 -Id 0 -Completed
```

jianyunt: No. won't repro. As @BrucePay also mentioned above, this code path doesn't get hit in the cmdlet scenario. It will repro with APIs.

jianyunt:
You may write an app in C#, call WriteProgress(), then call the above PowerShell script from C#.

Or you may try the following:
On your Linux, install-module azurerm.netcore
install-module azurepsdrive
Login-AzureRmAccount
new-psdrive -name az -psprovider SHiPS -root 'Azurepsdrive#Azure'
cd az:
Get- then tab
you will the problem
cd az:\AutomationTeam\WebApps
Get- tab, you will the nested progress issue.

src/Microsoft.PowerShell.ConsoleHost/host/msh/ConsoleHostUserInterfaceProgress.cs
jianyunt: You may write an app in C#, call WriteProgress(), then call the above PowerShell script from C#.

Or you may try the following:
On your Linux, install-module azurerm.netcore
install-module azurepsdrive
Login-AzureRmAccount
new-psdrive -name az -psprovider SHiPS -root 'Azurepsdrive#Azure'
cd az:
Get- then tab
you will the problem
cd az:\AutomationTeam\WebApps
Get- tab, you will the nested progress issue.

PR #6939, opened 2018-05-25
Skip compiling the non-supported cmdlets in System.Management.Automation.dll

daxian-dbw:
## PR Summary

Fix #6886 
Our configuration definition 'Linux' caused our Linux build to be unoptimized, and we didn't realize until recently. This was fixed in #6853, which makes our Linux build adopts the same configuration settings as on Windows -- debug and release.

`Start-PSBuild -Clean` on Linux starts to use the debug configuration by default, which reveals another issue on Linux build -- we use a cache in `InitialSessionState` to initialize cmdlets from S.M.A.dll in order to speed up powershell startup and items in cache drifts from actual cmdlet types available in S.M.A.dll in Unix platforms.

This is because there are some cmdlets not supported on Unix, which are if/def'ed out in the cache code, but they are still compiled in S.M.A.dll on Unix. We have some code that only runs in debug configuration to check if the cache is in sync with the real status of S.M.A.dll, but we never got to run in debug configuration because of the 'Linux' configuration we used before.

This PR skips compiling the non-supported cmdlets in S.M.A.dll, to keep the cache in InitialSessionState in sync with the actually available cmdlet types in System.Management.Automation.dll.

## Change Details

**Note that** some of the cmdlet source files contain codes that are used in other places outside the cmdlet. Those codes are moved out to `PSRemotingCmdlet.cs`, a more general place where base classes and helper classes for remoting command are located. The following type/methods are moved:
1. Class `QueryRunspaces` and enum `SessionFilterState` from `ConnectPSSession.cs`
2. Class `PSSessionOption` and enum `ProxyAccessType` from `NewPSSessionOptionCommand.cs`
3. Methods `CheckIfPowerShellVersionIsInstalled` and `CheckPSVersion` from `CustomShellCommands.cs` to class `RemotingCommandUtils` in `PSRemotingCmdlet.cs`

## PR Checklist

- [x] [PR has a meaningful title](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - Use the present tense and imperative mood when describing your changes
- [x] [Summarized changes](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [x] [Change is not breaking](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes)
[...]

PR #7029, opened 2018-06-09
Cleanup: remove the unneeded type 'RemotingCommandUtils'

daxian-dbw:
## PR Summary

I added the type `Microsoft.PowerShell.Commands.RemotingCommandUtils` in PR #6939, but it turned out we already have a type named `Microsoft.PowerShell.Commands.RemotingCommandUtil` for the same purpose. So this PR merge these two and remove the newly added type `RemotingCommandUtils`.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/remoting/commands/remotingcommandutil.cs
+++ b/src/System.Management.Automation/engine/remoting/commands/remotingcommandutil.cs
@@ -3,10 +3,10 @@
 
 using System;
 using System.Management.Automation;
+using System.Management.Automation.Internal;
+using System.Management.Automation.Remoting;
 using System.Management.Automation.Runspaces;
 using Microsoft.Win32;
-using Dbg = System.Management.Automation.Diagnostics;
-using System.Management.Automation.Internal;
 
 namespace Microsoft.PowerShell.Commands
 {
--- a/src/System.Management.Automation/engine/remoting/commands/remotingcommandutil.cs
+++ b/src/System.Management.Automation/engine/remoting/commands/remotingcommandutil.cs
@@ -31,12 +31,6 @@ namespace Microsoft.PowerShell.Commands
     /// </summary>
     internal static class RemotingCommandUtil
     {
-        /// <summary>
-        /// The existence of the following registry confirms that the host machine is a WinPE
-        /// HKLM\System\CurrentControlSet\Control\MiniNT
-        /// </summary>
-        internal static string WinPEIdentificationRegKey = @"System\CurrentControlSet\Control\MiniNT";
-
         internal static bool HasRepeatingRunspaces(PSSession[] runspaceInfos)
         {
             if (runspaceInfos == null)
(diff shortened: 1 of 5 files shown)


src/System.Management.Automation/engine/remoting/commands/remotingcommandutil.cs
```diff
@@ -170,6 +164,61 @@ internal static void CheckHostRemotingPrerequisites()
...
+#if CORECLR
+                // PowerShell 2.0 is not available for CoreCLR
+                throw new ArgumentException(
+                    PSRemotingErrorInvariants.FormatResourceString(
+                        RemotingErrorIdStrings.PowerShellNotInstalled,
+                        version, "PSVersion"));
+#else
+                // Because of app-compat issues, in Win8, we will have PS 2.0 installed by default but not .NET 2.0
+                // In such a case, it is not enough if we check just PowerShell registry keys. We also need to check if .NET 2.0 is installed.
+                try
```
iSazonov: Seems we can remove the code after "else".

rjmholt: And if it gets removed, this method will just throw when given v2 and do nothing otherwise. I assume that's all the method should do for PowerShell 6?

rjmholt (approved):
This looks good to me

src/System.Management.Automation/engine/remoting/commands/remotingcommandutil.cs
daxian-dbw: Hmm, throwing that exception in `#if CORECLR` was a decision when porting the code to NanoServer targeting .NET Core 1.0. Back then, PSv2 was definitely not available on NanoServer, which the only place where PowerShell Core edition is available, so throwing the exception had no problem. Now PowerShell Core is on the desktop, so I'm not sure if we should just throw the exception there (win 10 has removed PSv2, but downlevel OS still has it as an optional windows feature). Besides, if we decide to go with the exception, we probably need to change the error message.

I have updated #3565 to keep track of this cleanup work.

rjmholt: Is this the behaviour designed to support `powershell.exe -version 2`?

daxian-dbw: @rjmholt I didn't look into why the PSv2 is taken into consideration here, but possibly related to the scenario where the remote server is PSv2.

rjmholt: Yes that makes sense! Well I suppose the behaviour without the #ifdef makes sense then -- no-op unless trying to remote into v2, in which case, **exception**.

iSazonov: I believe we don't support PSv2 at all.

rjmholt: Yes -- so I'm guessing an exception makes sense then? Something that is red and tells the user we don't support this, but doesn't crash the shell

iSazonov: Seems the method is used to run jobs. I guess the remoting (with both local and remote hosts) with PSv2 doesn't work at all.
/cc @PaulHigin Could you please comment too?

Issue #2136, opened 2016-08-30
Modules like PowerShell Get need the version bumped up when modified


kilasuit:
https://github.com/PowerShell/PowerShell/blob/master/src/Modules/Shared/PowerShellGet/PowerShellGet.psd1 - This still shows 1.0.0.1 for the module version (same version as the module that was bundled with Win10 10586) even though there have been a number of changes to this module since July 2015

This follows on from the request in #1979 to decouple any included modules like PowerShell Get so that with these modules we can have a build & release process that we have confidence in and this is becoming more and more important to get right as the number of changes to this Repository grows.

zipzit:
If you are a user, is there any way to determine which version of powershell you are running? 
`powershell help` and `powershell help version` reveal nothing.   `powershell --version` and `powershell -ver` reveal nothing.   There is no menu select for "About" or anything related.  And why isn't such a command available by an observer starting at `powershell help` ? 

Why the big secret?   Is the version level such a big secret, or am I just doing this wrong?

kilasuit:
@zipzit you would have to start powershell and then run
```powershell
$PSVersionTable.PSVersion
```
that is the correct way to find the version of PowerShell that you are running. 

If you are new to using powershell then take a look at the following learning powershell page at https://github.com/PowerShell/PowerShell/blob/master/docs/learning-powershell

bmanikm:
Fixed in #2372

kilasuit:
@bmanikm It may be fixed for PowerShellGet however this needs to be **enforced on all inbox shipped modules** until #1979 has been completed and each module has decoupled and has a reliable and robust build process that all users can trust in which is especially truthful of all MS released modules

Currently I haven't got that level of trust that any change in any of the modules shipped inside of this repository, regardless of the size of the change, couldn't end up bringing an unforeseen breaking change somewhere else down the line.

Don't get me wrong as now PowerShell has been Open Sourced the direction is going the right way to being able to actually have that level of trust however there is still work to be done to get us there so could you re-open this until #1979 has been completed

bmanikm:
Thanks @kilasuit!, I have updated my above comment.

SteveL-MSFT:
@zipzit getting the powershell version using -version without argument makes sense, can you open as a new issue?

joeyaiello:
We should do this, but I think it's probably more in line with whenever we do the module refactoring.

kilasuit:
[...]

PR #7083, opened 2018-06-15
Add 'CompatiblePSEditions' to PowerShell Core built-in modules

daxian-dbw:
## PR Summary

Add `CompatiblePSEditions = @("Core")` to PowerShell Core built-in modules.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**


rjmholt:
Do we control the PSReadLine psd1? I noticed trying it with my compatibility checking prototype that it also doesn't have this field 😱

src/Modules/Shared/Microsoft.PowerShell.Host/Microsoft.PowerShell.Host.psd1
```diff
@@ -4,10 +4,8 @@ Author="PowerShell"
 CompanyName="Microsoft Corporation"
 Copyright="Copyright (c) Microsoft Corporation. All rights reserved."
 ModuleVersion="3.0.0.0"
```
SteveL-MSFT: There's a request to update our in-package ModuleVersion https://github.com/PowerShell/PowerShell/issues/2136, I think we can just have the ModuleVersion match the PSCore version "6.1.0.0" for this release until we can actually decouple them.

SteveL-MSFT:
@rjmholt That needs to be addressed in the PSReadLine repo.  However, since PSReadLine supports PSv3 and higher which doesn't understand the `CompatiblePSEditions` property, we can't add it.

rjmholt:
Also, given that the modules have been changing in sync with PowerShell and that PowerShell is now version 6.1, should the `PowerShellVersion` field be `6.1`?

It's unlikely that someone would even manage to get to the point where loading old PowerShell modules into a newer version is possible, but it's conceivable, and in principle should be protected against maybe?

rjmholt:
@SteveL-MSFT New PSReadLine will only support v5+ right? So that can have the field? But I'm assuming that until then, or in cases using older PSReadLine versions, we take advantage of implicit compatibility due to path location?

SteveL-MSFT:
@rjmholt regarding `PowerShellVersion`, yes it would be more accurate to set it to `6.1`.  As for PSReadLine, I think even 2.0.0 still supports v3, not sure.  I would suggest following up on the PSReadLine repo.

daxian-dbw:
@rjmholt @SteveL-MSFT Your comments were addressed, please take another look when you have time. Thanks!

rjmholt:
@SteveL-MSFT Looks like PSReadLine 2.0 is now v5+: https://github.com/lzybkr/PSReadLine/blob/46960b1fe442024923381eedd1c9100ded30d123/PSReadLine/Changes.txt#L32

rjmholt (approved):

SteveL-MSFT (approved):

Issue #7038, opened 2018-06-10
AppVeyor PR builds can hang sporadically and then timeout/abort after only 3 hours


bergmeister:
Not sure if this has happened before or is due to a new test, but there were recently 2 different `AppVeyor` PR builds that got stuck and then only aborted due to the 3 hour timeout.

I first saw it in [this ](https://ci.appveyor.com/project/PowerShell/powershell/build/v6.1.0-preview.9622) PR build of PR 7013 and then also in [this ](https://ci.appveyor.com/project/PowerShell/powershell/build/v6.1.0-preview.9662)PR build of PR 7033
The log stops at different tests in those 2 runs, therefore not sure if it is a problem of a specific test or aftereffect of a certain test suite or an AppVeyor problem

iSazonov:
I guess there was unhandled error in build process.

bergmeister:
It happened again (in master) yesterday:
https://ci.appveyor.com/project/PowerShell/powershell/build/v6.1.0-preview.9751

TravisEz13:
The last two hangs I saw where in this code: https://github.com/PowerShell/PowerShell/blob/master/tools/appveyor.psm1#L551

I'm guessing that AppVeyor is hanging after we throw the error back to appveyor.  
Finish is a really wierd place to be doing this type of stuff... building a package and testing it.  I would refactor this into a different bulid and test job if I had the time and see if it behaves better.

iSazonov:
Maybe don't throw - write an error and `exit -1`?

RDIL:
This is fixed now

PR #7120, opened 2018-06-19
Update AppVeyor cache

SteveL-MSFT:
## PR Summary

We've been seeing some AppVeyor failures after publishing new version of pwrshplugin.dll.  The cache name wasn't changed and suspicion is that the cache is invalid and needs to be updated.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,7 +9,7 @@ environment:
   - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
     Purpose: ElevatedPesterTests_xUnit_Packaging
 
-# cache version - netcoreapp.2.1-sdk.2.1.300
+# cache version - netcoreapp.2.1-sdk.2.1.300-pwrshplugin-6.1.0-preview3
 cache:
   - '%HOMEDRIVE%%HOMEPATH%\.nuget\packages -> appveyor.yml'
 


bergmeister:
@steveL-msft
Can you please provide links to the build failures? Can you also please explain how/why this PR would help? Only NuGet packages are being cached and I cannot see how this can be a problem?
Is it related to #7038 ?

SteveL-MSFT:
This should not be related to #7038.  Pwrshplugin.dll is published as psrp.windows nupkg.  Failure was first noticed in our nightly (https://ci.appveyor.com/project/PowerShell/powershell-f975h), but I also see it in my PRs.

bergmeister:
Ok. Thanks for the info. However, if you really suspect that the presence of an older NuGet package is really the culprit then there seems to be an underlying issue (in general or in the test) that should be solved rather than being swept under the carpet by flushing the cache. It might be OK as a quick workaround to get it back to work to unblock things but understanding the failure would still be essential IMHO

SteveL-MSFT:
That binary is pulled into the build from myget as a nupkg and required for WSMan remoting endpoint for PSCore6.  The test that's failing is trying to setup a remoting endpoint and that is failing because of that binary.  The cache needs to be updated everytime we update the nuget dependencies which is the case here.  I don't think anything is being swept under the carpet.

SteveL-MSFT:
Ok, just looked at the nupkg and it was published incorrectly.  May not need this cache change.

PR #7126, opened 2018-06-21
Refactor module specification logic

rjmholt:
## PR Summary

Refactors the code for `ModuleSpecification`s:
* Use `nameof` where appropriate
* Factor out a repeated `entry.Key.ToString()` call
* Use a `StringBuilder` in the `ToString()` method
* Dissect a trick comparison method into something simpler
* Use HashCode.Combine() method

Each of the above is its own commit, so we can take any change out.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
@@ -45,7 +45,7 @@ namespace Microsoft.PowerShell.Commands
         {
             if (string.IsNullOrEmpty(moduleName))
             {
-                throw new ArgumentNullException("moduleName");
+                throw new ArgumentNullException(nameof(moduleName));
             }
             this.Name = moduleName;
             // Alias name of miniumVersion
@@ -67,7 +67,7 @@ namespace Microsoft.PowerShell.Commands
         {
             if (moduleSpecification == null)
             {
-                throw new ArgumentNullException("moduleSpecification");
+                throw new ArgumentNullException(nameof(moduleSpecification));
             }
 
             var exception = ModuleSpecificationInitHelper(this, moduleSpecification);
@@ -166,7 +166,7 @@ namespace Microsoft.PowerShell.Commands
         {
             if (moduleInfo == null)
             {
-                throw new ArgumentNullException("moduleInfo");
+                throw new ArgumentNullException(nameof(moduleInfo));
             }
 
             this.Name = moduleInfo.Name;
--- a/src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
@@ -304,9 +304,6 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         /// <param name="obj">The module specification for the object.</param>
         /// <returns>A hashcode that is always the same for any module specification with the same properties.</returns>
-        /// <remarks>
-        /// The algorithm in this method is adapted from a StackOverflow answer from Jon Skeet: https://stackoverflow.com/a/263416/9944203
-        /// </remarks>
         public int GetHashCode(ModuleSpecification obj)
         {
             if (obj == null)
@@ -314,40 +311,7 @@ namespace Microsoft.PowerShell.Commands
                 return 0;
             }
 
-            // Initial prime value
-            int p = 23;
-
-            // Use 31 for multiplication. See https://computinglife.wordpress.com/2008/11/20/why-do-hash-functions-use-prime-numbers/
-            int q = 31;
-
-            int result = p; 
-
-            if (obj.Name != null)
-            {
-                result = unchecked(q * result + obj.Name.GetHashCode());
-            }
-
-            if (obj.Guid != null)
-            {
-                result = unchecked(q * result + obj.Guid.GetHashCode());
-            }
-
-            if (obj.RequiredVersion != null)
-            {
-                result = unchecked(q * result + obj.RequiredVersion.GetHashCode());
-            }
-
-            if (obj.Version != null)
-            {
-                result = unchecked(q * result + obj.Version.GetHashCode());
-            }
-
-            if (obj.MaximumVersion != null)
-            {
-                result = unchecked(q * result + obj.MaximumVersion.GetHashCode());
-            }
-
-            return result;
+            return HashCode.Combine(obj.Name, obj.Guid, obj.RequiredVersion, obj.Version, obj.MaximumVersion);
         }
     }
 
--- a/src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
@@ -91,31 +91,35 @@ namespace Microsoft.PowerShell.Commands
             {
                 foreach (DictionaryEntry entry in hashtable)
                 {
-                    if (entry.Key.ToString().Equals("ModuleName", StringComparison.OrdinalIgnoreCase))
+                    string field = entry.Key.ToString();
+
+                    if (field.Equals("ModuleName", StringComparison.OrdinalIgnoreCase))
                     {
                         moduleSpecification.Name = LanguagePrimitives.ConvertTo<string>(entry.Value);
                     }
-                    else if (entry.Key.ToString().Equals("ModuleVersion", StringComparison.OrdinalIgnoreCase))
+                    else if (field.Equals("ModuleVersion", StringComparison.OrdinalIgnoreCase))
                     {
                         moduleSpecification.Version = LanguagePrimitives.ConvertTo<Version>(entry.Value);
                     }
-                    else if (entry.Key.ToString().Equals("RequiredVersion", StringComparison.OrdinalIgnoreCase))
+                    else if (field.Equals("RequiredVersion", StringComparison.OrdinalIgnoreCase))
                     {
                         moduleSpecification.RequiredVersion = LanguagePrimitives.ConvertTo<Version>(entry.Value);
                     }
-                    else if (entry.Key.ToString().Equals("MaximumVersion", StringComparison.OrdinalIgnoreCase))
+                    else if (field.Equals("MaximumVersion", StringComparison.OrdinalIgnoreCase))
                     {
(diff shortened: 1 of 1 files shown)


iSazonov:
We could use https://stackoverflow.com/questions/263400/what-is-the-best-algorithm-for-an-overridden-system-object-gethashcode

rjmholt:
@iSazonov I actually was thinking of that exact StackOverflow answer. I didn’t want to change the actual numerical result of `GetHashCode()` without approval on the basis that it could break things...

But I definitely agree that that would be a better hash algorithm and can change it over in this PR if that is desired.

iSazonov:
GetHashCode result is not persistent.
From https://msdn.microsoft.com/en-us/library/system.object.gethashcode%28v=vs.110%29.aspx
> 
•You should not assume that equal hash codes imply object equality. 


•You should never persist or use a hash code outside the application domain in which it was created, because the same object may hash across application domains, processes, and platforms.

rjmholt:
Ah wasn't aware of the second principle! In that case I shall change it use Mr Skeet's implementation

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
```diff
@@ -266,92 +273,58 @@ public static bool TryParse(string input, out ModuleSpecification result)
...
+            }
+
+            // Two ordinary prime numbers for hashing
+            int p = 23;
+            int q = 59;
 
-            if (obj != null)
+            int result = p; 
+
+            foreach (object property in new object[] { obj.Name, obj.Guid, obj.RequiredVersion, obj.Version, obj.MaximumVersion })
```
iSazonov: Looking at this I remembered that we have
```c#
internal static int CombineHashCodes(int h1, int h2, int h3, int h4, int h5)
```
in Utils.cs.

rjmholt: Well on the one hand it would be nice to deduplicate our code.

On the other hand:
* The original intent of the refactor here was to roll the repeated operations into a loop. `CombineHashCodes()` doesn't have a `params int[] hs` overload, so we'd have to unroll the loop again.
* The implementation in `CombineHashCodes()` still uses XOR. [This StackOverflow answer](https://stackoverflow.com/a/34229665/9944203) discusses the implementation we borrowed from System.Web.Util.HashCodeCombiner and I notice that another answer is Jon Skeet again disrecommending XOR for hashcodes.

Not for this PR, but maybe it's worth:
* Changing `CombineHashCodes()` to use the add/multiply method, and
* Adding a `CombineHashCodes(int h, param int[] hs)` overload

If we did that it would probably then be worth moving the `ModuleSpecificationComparer.GetHashCode()` method over to use `Utils.CombineHashCodes()`.

iSazonov: I agree about XOR but we use it very frequently - I propose to don't block the PR and to postpone the change to a special PR where cleanup all code base. Please open new tracking Issue if you agree.

iSazonov (approved):

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
iSazonov: For reference. We have also:
```
internal static int SequenceGetHashCode<T>(this IEnumerable<T> xs)

public int GetHashCode(Tuple<ITypeName, TypeResolutionState> obj)

public override int GetHashCode(ICollection<T> obj)
```

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
```diff
@@ -266,92 +273,58 @@ public static bool TryParse(string input, out ModuleSpecification result)
...
-                    else if (result && (x.RequiredVersion != null || y.RequiredVersion != null))
-                    {
-                        result = false;
-                    }
-                }
+                return true;
             }
 
-            return result;
+            return x != null && y != null
```
anmenaga: I would like to see test results comparison 'before change' vs 'after change' for specific tests for:
- ToString()
- Equals(ModuleSpecification x, ModuleSpecification y)
- GetHashCode(ModuleSpecification obj)

If we don't have such tests already - please add them.

anmenaga (changes requested):

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
rjmholt: The tests for `GetHashCode()` will be pretty simple. Basically that `null` doesn't fail and that objects with some fields set have values. Because we can't guarantee the value for `GetHashCode()` for any particular object.

anmenaga: In addition:  create 2 separate but identical objects, ensure that Equals method returns true for them, verify that GetHashCode returns same value for them.

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
```diff
@@ -266,92 +273,60 @@ public static bool TryParse(string input, out ModuleSpecification result)
...
+                && String.Equals(x.Name, y.Name, StringComparison.OrdinalIgnoreCase)
+                && Guid.Equals(x.Guid, y.Guid)
+                && Version.Equals(x.RequiredVersion, y.RequiredVersion)
+                && Version.Equals(x.Version, y.Version)
+                && String.Equals(x.MaximumVersion, y.MaximumVersion);
         }
```
BrucePay: What happened to all of the null checks?

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
```diff
@@ -266,92 +273,60 @@ public static bool TryParse(string input, out ModuleSpecification result)
...
         {
-            int result = 0;
+            if (obj == null)
+            {
+                return 0;
+            }
+
+            // Initial prime value
+            int p = 23;
+
```
BrucePay: If the only use of p is to initialize `result` why not just initialize `result` directly.

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
```diff
@@ -266,92 +273,60 @@ public static bool TryParse(string input, out ModuleSpecification result)
...
+            // Use 31 for multiplication. See https://computinglife.wordpress.com/2008/11/20/why-do-hash-functions-use-prime-numbers/
+            int q = 31;
+
+            int result = p; 
 
-            if (obj != null)
+            foreach (object property in new object[] { obj.Name, obj.Guid, obj.RequiredVersion, obj.Version, obj.MaximumVersion })
             {
```
BrucePay: While the loop is nicer code it does mean that you doing an allocation where there wasn't one before.

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
rjmholt: The null checks are done by the static methods themselves (tooltip for the methods specify that). Figured it was better not to repeat them.

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
rjmholt: I figured it looks better than just throwing `23` in there (I noticed a few stack overflow answers on this tended to say "I dunno why we used that number", so I thought it was worth calling them `p` and `q`). But not attached to it -- can happily change them back to inline constants.

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
rjmholt: Yes I tried to look for a better way to do this (without the allocation), but I'm guessing it's not possible. Either we keep the allocation or we unroll the loop. Thought the loop was nicer initially, but can go either way.

iSazonov: I agree that we should avoid such allocations.
As we discussed above we could review all code base to clean up GetHashCode methods to use unified approach.

iSazonov: So everybody put the same questions I suggest now use 
`internal static int CombineHashCodes(int h1, int h2, int h3, int h4, int h5)`
and defer improvements to new PR.

rjmholt:
Ok, I've written some tests in #7140. Please take a look.

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
rjmholt: I imagine `CombineHashCodes` may not get improved for a while if ever and for large numbers of modules, a better distribution is helpful. How about if I unroll the loop but use the new hash algorithm?

iSazonov: Ok.

rjmholt: If we do end up updating the `CombineHashCodes` method, we should circle back here and call that.

iSazonov (approved):

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
rjmholt: @anmenaga I've opened another PR for tests in https://github.com/PowerShell/PowerShell/pull/7140.

anmenaga (approved):

iSazonov:
Waiting #7140

rjmholt:
Going to change the hashcode logic to use .NET Core's native implementation

iSazonov (approved):

iSazonov:
@rjmholt Please rebase to get new tests.

rjmholt:
It looks like this is passing all the new tests I wrote 😃

iSazonov:
@BrucePay @daxian-dbw Could you please review the PR?

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
iSazonov: New tests from #7140 was merged in master and added in the PR by rebase.

PR #7100, opened 2018-06-18
Remove Utils.FileExists(), Utils.DirectoryExists() and Utils.ItemExists() methods

iSazonov:
## PR Summary

Continue #6929

Cleanup code to replace Utils.FileExists(), Utils.DirectoryExists() and Utils.ItemExists() methods with .Net Core methods.

## PR Checklist

- [x] [PR has a meaningful title](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - Use the present tense and imperative mood when describing your changes
- [x] [Summarized changes](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [x] [Change is not breaking](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes)
- [x] [Make sure all `.h`, `.cpp`, `.cs`, `.ps1` and `.psm1` files have the correct copyright header](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [x] This PR is ready to merge and is not [Work in Progress](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---work-in-progress).
    - If the PR is work in progress, please add the prefix `WIP:` to the beginning of the title and remove the prefix when the PR is ready.
- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - [ ] User-facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
        - [ ] Issue filed - Issue link:
- **Testing - New and feature**
    - [x] Not Applicable or can only be tested interactively
    - **OR**
    - [x] [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
        - [x] [Add `[feature]` if the change is significant or affects feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

PR #7129, opened 2018-06-21
Cleanup Utils.FileExists(), Utils.DirectoryExists() and Utils.ItemExists() methods

iSazonov:
## PR Summary

Continue #6929
Replace #7100 

Cleanup code to replace Utils.FileExists(), Utils.DirectoryExists() and Utils.ItemExists() methods with .Net Core methods.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell/7129)


iSazonov:
WIP - still 6 commits upcoming...

iSazonov:
@daxian-dbw Please look the issue with Unix compiling.
```
843/home/travis/build/PowerShell/PowerShell/src/System.Management.Automation/help/UpdatableHelpSystem.cs(1650,106): error CS0117: 'EnumerationOptions' does not contain a definition for 'MatchCasing' [/home/travis/build/PowerShell/PowerShell/src/System.Management.Automation/System.Management.Automation.csproj]
```
Yestoday the commit was compiled without errors. Today it failed. I tryed compile the code on CI Appveyor and get the same error.
Maybe we get updated version of .Net Core with a bug?

Locally I can compile without problem and IntelliSense works well in VS Code.

iSazonov:
I have only one commit related to Unix but can not push it until travis-ci problem will be resolved.

iSazonov:
I tested locally all versions - msi and zip packages - 2.1.300 and 2.1.301 and the project is compiled well.
So problem is somewhere else.
I wonder how and why the problem is arised on both CIs.

iSazonov:
Rebased to move to .Net Core 2.1.1.

iSazonov:
@daxian-dbw Thanks for help!

The PR is ready for review.

iSazonov:
Rebased to resolve a merge conflict.

src/System.Management.Automation/engine/SessionStateDriveAPIs.cs
```diff
@@ -963,25 +963,9 @@ private bool IsAStaleVhdMountedDrive(PSDriveInfo drive)
...
-                            }
-                            // We don't want to have automounting cause an exception. We
-                            // rather it just fail silently as it wasn't a result of an
-                            // explicit request by the user anyway.
-                            // Following the same pattern as the Calling API.
-                            catch (IOException)
-                            {
-                            }
-                            catch (UnauthorizedAccessException)
+                            if (!Directory.Exists(drive.Root))
```
adityapatwardhan: Should this be
```CSharp
return !DirectoryExists(drive.Root)
```

adityapatwardhan (approved):
Signed of with 1 non-blocking comment.

src/System.Management.Automation/engine/SessionStateDriveAPIs.cs
iSazonov: Seems original code is easier to understand.

daxian-dbw:
@iSazonov Let me quickly go through the changes.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -5573,7 +5576,7 @@ protected override bool IsItemContainer(string path)
 
             path = NormalizePath(path);
 
-            return Utils.DirectoryExists(path);
+            return Directory.Exists(path);
```
daxian-dbw: same question, why we ignore all exceptions here?

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -6144,27 +6134,21 @@ public void SetProperty(string path, PSObject propertyToSet)
...
 
-            if (Utils.ItemExists(path, out bool isDirectory))
+            // Create a PSObject with either a DirectoryInfo or FileInfo object at its core.
+            if (fsinfo != null)
             {
                 if (isDirectory)
                 {
-                    isContainer = true;
-                    fileSystemInfoShell = PSObject.AsPSObject(new DirectoryInfo(path));
+                    fileSystemInfoShell = PSObject.AsPSObject((DirectoryInfo)fsinfo);
```
daxian-dbw: no need to cast here

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -6144,27 +6134,21 @@ public void SetProperty(string path, PSObject propertyToSet)
...
                 }
                 else
                 {
-                    // Maybe the path is a file name so try a FileInfo instead
-                    fileSystemInfoShell = PSObject.AsPSObject(new FileInfo(path));
+                    fileSystemInfoShell = PSObject.AsPSObject((FileInfo)fsinfo);
```
daxian-dbw: ditto.
Actually, this if/else block can be removed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -6144,27 +6134,21 @@ public void SetProperty(string path, PSObject propertyToSet)
...
                 }
-            }
 
-            if (fileSystemInfoShell != null)
-            {
```
daxian-dbw: `fileSystemInfoShell` still could be null, right?

src/System.Management.Automation/namespaces/LocationGlobber.cs
```diff
@@ -501,6 +502,14 @@ out providerInstance
 
             Collection<string> stringResult = new Collection<string>();
 
+            // If the directory exists, just return it.
+            // Otherwise fallback to previous behavior and let provider handle it.
+            if (Directory.Exists(userPath))
```
daxian-dbw: can you please explain why having this change? And this might not be the right place for the new code, because a new collection is just allocated above, but you might return without using the collection at all.

daxian-dbw (changes requested):

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1494,30 +1499,28 @@ protected override void InvokeDefaultAction(string path)
...
-                    DirectoryInfo directory = new DirectoryInfo(path);
                     InodeTracker tracker = null;
 
                     if (recurse)
                     {
                         GetChildDynamicParameters fspDynamicParam = DynamicParameters as GetChildDynamicParameters;
                         if (fspDynamicParam != null && fspDynamicParam.FollowSymlink)
                         {
-                            tracker = new InodeTracker(directory.FullName);
+                            tracker = new InodeTracker(((DirectoryInfo)fsinfo).FullName);
```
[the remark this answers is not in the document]
iSazonov: I see `FullName` is in different code paths in the classes:
https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/FileInfo.cs#L29
https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/DirectoryInfo.cs#L38

src/System.Management.Automation/help/UpdatableHelpSystem.cs
```diff
@@ -1627,32 +1627,36 @@ internal static string LoadStringFromPath(PSCmdlet cmdlet, string path, PSCreden
         /// <returns></returns>
         internal static string GetFilePath(string path)
         {
+            FileInfo item = new FileInfo(path);
+
+            // We use 'FileInfo.Attributes' (not 'FileInfo.Exist')
+            // because we want to get exceptions
+            // like UnauthorizedAccessException or IOException.
+            if ((int)item.Attributes != -1)
```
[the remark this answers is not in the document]
iSazonov: I tried to keep the logic of the previous code. And I tried to make as few changes as possible. I'm afraid that this PR will grow many times if we try to correct the previous logic in all places where `Utils.DirectoryExists` or `Utils.FileExists` exist.

Here I'm afraid that we'll get issues in `LoadStringFromPath()` .

src/System.Management.Automation/help/UpdatableHelpSystem.cs
```diff
@@ -1627,32 +1627,36 @@ internal static string LoadStringFromPath(PSCmdlet cmdlet, string path, PSCreden
...
                 {
-                    if (filePath.EndsWith(fileName, StringComparison.OrdinalIgnoreCase))
-                    {
-                        return filePath;
-                    }
+                    return fileList[0].FullName;
                 }
             }
-#else
-            if (Utils.FileExists(path))
```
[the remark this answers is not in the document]
iSazonov: I assumed that we will always receive the path with the right case. (Seems we get the path in code not from user). So the first block will always return a value on all platforms. The code below will only search the file on Unix with ignoring case.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1553,12 +1556,12 @@ protected override void InvokeDefaultAction(string path)
...
                             WriteItemObject(
-                                fileInfo.Name,
-                                fileInfo.FullName,
+                                ((FileInfo)fsinfo).Name,
+                                ((FileInfo)fsinfo).FullName,
                                 false);
                         }
                         else
-                            WriteItemObject(fileInfo, path, false);
+                            WriteItemObject((FileInfo)fsinfo, path, false);
```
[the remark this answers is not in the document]

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -2341,20 +2348,19 @@ public static string Mode(PSObject instance)
...
+                        pathDirInfo = GetFileSystemInfo(path, out isPathDirectory);
                     }
                     catch (Exception e)
                     {
                         WriteError(new ErrorRecord(e, "AccessException", ErrorCategory.PermissionDenied, strTargetPath));
                         return;
                     }
 
-                    DirectoryInfo pathDirInfo = new DirectoryInfo(path);
+                    bool pathExists = pathDirInfo != null;
```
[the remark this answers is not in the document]
iSazonov: There is two uses below.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -5401,13 +5406,12 @@ private Stack<string> NormalizeThePath(string basepath, Stack<string> tokenizedP
...
 
-                    Boolean isContainer = false;
-                    FileSystemInfo fsinfo = GetFileSystemInfo(currentPath, ref isContainer);
+                    var fsinfo = new FileInfo(currentPath);
 
                     // Clean up the child name to proper casing and short-path
                     // expansion if required. Also verify that .NET hasn't over-normalized
                     // the path
-                    if (fsinfo != null)
+                    if ((int)fsinfo.Attributes != -1)
```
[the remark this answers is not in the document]
iSazonov: `GetFileSystemInfo` method does unnecessary actions. Given that `NormalizeThePath` is called too often, we better exclude this call. Although if you want I can use `GetFileSystemInfo` - it is better to optimize the provider globally

src/System.Management.Automation/namespaces/FileSystemProvider.cs

src/System.Management.Automation/namespaces/FileSystemProvider.cs

src/System.Management.Automation/namespaces/FileSystemProvider.cs
iSazonov: Why? We call PSObject.AsPSObject(fsinfo) where fsinfo != null.

src/System.Management.Automation/namespaces/LocationGlobber.cs
iSazonov: In this PR I went through so many code ways that I do not remember how I got here.
As far as I see the idea was simple - do not perform an analysis if the path already exists.

Fixed - I return without using the collection.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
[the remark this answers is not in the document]
daxian-dbw: Here is a quick check of the properties, and it looks the casting is not needed.
```c#
void Main()
{
	FileSystemInfo info1 = new FileInfo(@"C:\arena\temp.txt");
	FileSystemInfo info2 = new FileInfo(@"C:\arena");
	
	Console.WriteLine("info1.Name: {0}", info1.Name);
	Console.WriteLine("info1.FullName: {0}", info1.FullName);
	
	Console.WriteLine("info2.Name: {0}", info2.Name);
	Console.WriteLine("info2.FullName: {0}", info2.FullName);
}
```
Console Output:
```
info1.Name: temp.txt
info1.FullName: C:\arena\temp.txt
info2.Name: arena
info2.FullName: C:\arena
```

src/System.Management.Automation/namespaces/FileSystemProvider.cs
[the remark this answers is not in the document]
daxian-dbw: Yes, I missed them. #Closed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
daxian-dbw: Ah, you moved the code within `if (fsinfo != null)`. I missed that. #Closed

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -6144,27 +6134,14 @@ public void SetProperty(string path, PSObject propertyToSet)
 
             PSObject results = new PSObject();
             PSObject fileSystemInfoShell = null;
```
daxian-dbw: This variable now should be moved in `if (fsinfo != null)`, since it's only used in the if block.

src/System.Management.Automation/namespaces/LocationGlobber.cs
daxian-dbw: From a quick browsing of the code after this, it seems the new code you added treats the `userPath` as a literal path without consulting `context.SuppressWildcardExpansion`. Say `userPath` is `/blah/a[bc]`, and it happens to have a folder named `a[bc]`.

I'm not that familiar with the file system code, so there could be other concerns on this early return. I suggest to focus on the topic of this PR, and leave optimization to a separate issue/PR.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
```diff
@@ -1553,12 +1556,12 @@ protected override void InvokeDefaultAction(string path)
                         if (nameOnly)
                         {
                             WriteItemObject(
-                                fileInfo.Name,
-                                fileInfo.FullName,
+                                ((FileInfo)fsinfo).Name,
+                                ((FileInfo)fsinfo).FullName,
```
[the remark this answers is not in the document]
iSazonov: Ah, we don't need casting because the value is evaluated in object creation time and here we only retrive the value.

Fixed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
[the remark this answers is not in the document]
iSazonov: Ah, we don't need casting because the value is evaluated in object creation time and here we only retrive the value.

Fixed.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
daxian-dbw: I see. I guess it makes sense to swallow all exceptions here. #closed

src/System.Management.Automation/namespaces/FileSystemProvider.cs

src/System.Management.Automation/namespaces/LocationGlobber.cs
iSazonov: The code was removed.

src/System.Management.Automation/help/UpdatableHelpSystem.cs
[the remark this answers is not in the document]
daxian-dbw: Yes, it makes sense to to always have that check on all plats. #closed

src/System.Management.Automation/help/UpdatableHelpSystem.cs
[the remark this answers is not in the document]
daxian-dbw: Why do you think this method `GetFilePath(string path)` should throw the exception from checking the file existence?

> I tried to keep the logic of the previous code

The previous code just used `Utils.FileExists` and `Utils.DirectoryExists` like everywhere else. Here, you didn't replace them to `File.Exists` and `Directory.Exists` but check the existence in a way that may throw exception. I want to know what your decision is based on.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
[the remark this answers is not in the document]
daxian-dbw: I prefer to have as less `.Attributes != -1` checks as possible. Ideally, there should be only one utility method doing this `!= -1` check, which gets called everywhere else.

src/System.Management.Automation/help/UpdatableHelpSystem.cs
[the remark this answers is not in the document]
iSazonov: Utils.FileExists and Utils.DirectoryExists can throw but File.Exists and Directory.Exists don't throw.
If calling code catch exceptions and ignore them we can safely replace Utils.FileExists and Utils.DirectoryExists with File.Exists and Directory.Exists.
Here calling code don't catch and ignore exceptions so I used `item.Attributes`.

src/System.Management.Automation/namespaces/FileSystemProvider.cs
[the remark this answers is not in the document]
iSazonov: I'm too. 
And I'm always thinking about extra operations in the code.

Fixed.

iSazonov:
@daxian-dbw Could you please continue?

TravisEz13:
___
*[src/System.Management.Automation/engine/SessionStateDriveAPIs.cs, line 966 at r1](https://reviewable.io/reviews/powershell/powershell/7129#-LHipQN46-frdEEmsyE5:-LHipQN46-frdEEmsyE6:b-nrmmp5) ([raw file](https://github.com/powershell/powershell/blob/5b04c073f16fd311a034a27fee8c9683052a42d2/src/System.Management.Automation/engine/SessionStateDriveAPIs.cs#L966)):*
> ```Smalltalk
>                         if (systemDriveInfo.DriveType == DriveType.NoRootDirectory)
>                         {
>                             if (!Directory.Exists(drive.Root))
> ```

Is this going to suppress IO and UnAuthorized Exceptions like the old code?

TravisEz13:
___
*[src/System.Management.Automation/namespaces/FileSystemProvider.cs, line 136 at r1](https://reviewable.io/reviews/powershell/powershell/7129#-LHiqEnlBrPTUf3-0i6K:-LHiqEnlBrPTUf3-0i6L:b-o2g79i) ([raw file](https://github.com/powershell/powershell/blob/5b04c073f16fd311a034a27fee8c9683052a42d2/src/System.Management.Automation/namespaces/FileSystemProvider.cs#L136)):*
> ```Smalltalk
>                 if (isContainer)
>                 {
>                     return new DirectoryInfo(path);
> ```

Why don't you cast fsinfo to a `DirectoryInfo` like you do in a later line?

`Dir((DirectoryInfo)fsinfo, recurse, depth, nameOnly, returnContainers, tracker);`

TravisEz13 (commented):
> Reviewed 9 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/powershell/powershell/7129)* status: :shipit: complete! all files reviewed

iSazonov:
>Is this going to suppress IO and UnAuthorized Exceptions like the old code?

Yes, we preserve old code logic.

>Why don't you cast fsinfo to a DirectoryInfo like you do in a later line?

Here we create objects of _different_ types.
In Dir() you referenced we have above `if (isContainer)` so we really have DirectoryInfo object and can cast to the type.

daxian-dbw:
@iSazonov Will do today. Sorry for being tardy in reviewing this PR (need to spend much time on releasing this week).

src/System.Management.Automation/help/UpdatableHelpSystem.cs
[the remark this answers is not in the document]
daxian-dbw: > If calling code catch exceptions and ignore them we can safely replace Utils.FileExists and Utils.DirectoryExists with File.Exists and Directory.Exists.
Here calling code don't catch and ignore exceptions so I used item.Attributes.

Does this mean for all the places you changed to `File.Exists` or `Directory.Exists`, you have checked that the calling code actually caches exceptions and ignore them?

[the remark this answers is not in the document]
daxian-dbw: > If calling code catch exceptions and ignore them we can safely replace Utils.FileExists and Utils.DirectoryExists with File.Exists and Directory.Exists.
Here calling code don't catch and ignore exceptions so I used item.Attributes.

Does this mean for all the places you changed to `File.Exists` or `Directory.Exists`, you have checked that the calling code actually caches exceptions and ignore them?

src/System.Management.Automation/help/UpdatableHelpSystem.cs
```diff
@@ -1627,32 +1627,36 @@ internal static string LoadStringFromPath(PSCmdlet cmdlet, string path, PSCreden
...
             {
                 // Get the list of files in the directory.
-                string[] fileList = Directory.GetFiles(directoryPath);
-                foreach (string filePath in fileList)
+                FileInfo[] fileList = dirInfo.GetFiles(searchPattern: fileName, new System.IO.EnumerationOptions { MatchCasing = MatchCasing.CaseInsensitive });
+
+                // We use 'FileInfo.Attributes' (not 'FileInfo.Exist')
+                // because we want to get exceptions
+                // like UnauthorizedAccessException or IOException.
+                if (fileList.Length > 0 && (int)fileList[0].Attributes != -1)
```
daxian-dbw: The original code doesn't check for existence for the files returned from `Direcotry.GetFiles`. Why do you add the check now? Can't we trust `direcotryInfo.GetFiles`?

daxian-dbw: Please also respond to this comment: https://github.com/PowerShell/PowerShell/pull/7129#discussion_r204191483

daxian-dbw: @iSazonov Can you please respond to the first comment in this thread (https://github.com/PowerShell/PowerShell/pull/7129#discussion_r204194315)?

iSazonov: Sorry, GitHub sometimes tricky hides comments and I can not find where to type response :-)

You are right - the check is dead code - I removed it.

daxian-dbw: Great, thanks! #Closed

daxian-dbw (approved):

PR #7135, opened 2018-06-21
Fix for #6976 -replace operator slowed down after adding scriptblock support

### LINKED ISSUE #6976: Combining the -replace operator with a lookbehind assertion slowed down by a factor of 30+ between v6.0.2 and v6.1.0-preview.2
Note: I have not tried to analyze whether the problem is in PowerShell or CoreFx; on the PowerShell side, new features were recently added to `-replace` - see #6029 - but I have no idea whether this symptom relates to that.

The problem appears to be specific to `-replace`; using `-match` does not exhibit the symptom.

Steps to reproduce
------------------

Run the following command line:

* first on v6.0.2 (the current stable release as of this writing).

* then on v6.1.0-preview.2 (the current preview release).

```powershell
$arr =  (, 'fooo') * 1e6; (Measure-Command { $arr -replace '(?<=.{3}).+' }).TotalSeconds
```

Expected behavior
-----------------

Similar timings.

Actual behavior
---------------

Here are sample timings, in seconds, from a Windows 10 VM:

```none
# v6.0.2
1.8444752

# v6.1.0-preview.2
# !! more than 30 times slower
61.5428622
```

Environment data
----------------

```powershell
PowerShell Core v6.0.2
PowerShell Core v6.1.0-preview.2
```

BrucePay:
## PR Summary

Fix #6976.

The performance of `-replace` was greatly reduced when scriptblock support was added. This is because of the way the code selected what type of replacement to do. In the old code, the first choice was string. In the modified code, the last choice was string and an earlier clause used `TryConvertTo()` to see if the substitution object was (or could be transformed into) a `MatchEvaluator` object. It was this (very expensive) call to `TryConvertTo()` that was slowing everything down. I added a clause for strings (the most common case) to the beginning of the switch statement so it's fast again. I also added a call to `PSObject.Base()` to make sure the test for strings is correct.  

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/lang/parserutils.cs
+++ b/src/System.Management.Automation/engine/lang/parserutils.cs
@@ -908,7 +908,7 @@ namespace System.Management.Automation
                     pattern = rList[0];
                     if (rList.Count > 1)
                     {
-                        substitute = rList[1];
+                        substitute = PSObject.Base(rList[1]);
                     }
                 }
             }
@@ -972,6 +972,9 @@ namespace System.Management.Automation
         {
             switch (substitute)
             {
+                case string replacementString:
+                    return regex.Replace(input, replacementString);
+
                 case ScriptBlock sb:
                     MatchEvaluator me = match => {
                         var result = sb.DoInvokeReturnAsIs(


daxian-dbw (approved):
LGTM.

src/System.Management.Automation/engine/lang/parserutils.cs
```diff
@@ -908,7 +908,7 @@ internal static object ReplaceOperator(ExecutionContext context, IScriptExtent e
                     pattern = rList[0];
                     if (rList.Count > 1)
                     {
-                        substitute = rList[1];
+                        substitute = PSObject.Base(rList[1]);
```
iSazonov: Could you please add a comment why it is needed? what we break without this?

BrucePay: Sure, This is actually a pretty fundamental operation in PowerShell. Anywhere you receive something of type `object`, you have to deal with the case where it might be wrapped in a `PSObject`. So before you can use it as `string` or `ScriptBlock`, you have to get the base object. The `PSObject.Base()` method encapsulates this operation. If it's passed a PSObject, it returns the core object otherwise it just returns the object.

iSazonov:
LGTM.

lzybkr (approved):

PR #7138, opened 2018-06-21
Fix for #7128 Methods with return type [object] should return null for an empty result

### LINKED ISSUE #7128: Returning nothing (Automation.Null) from an [object]-typed custom-class method breaks
Steps to reproduce
------------------

```powershell
class Foo { [object] Bar() { return & {} } }; (new Foo).Bar()
```

Note that `& {}` outputs "nothing", i.e., the "null collection" `[System.Management.Automation.Internal.AutomationNull]::Value`

Expected behavior
-----------------

Given that PowerShell allows casting _any_ value to `[object]` - including `$null` and `[System.Management.Automation.Internal.AutomationNull]::Value`, the operation should succeed and the `[System.Management.Automation.Internal.AutomationNull]::Value`) should be passed through.

@BrucePay's take on the desired behavior differs, however (taken from [this SO answer](https://stackoverflow.com/a/50957388/45375) of his):

> We should still generate an error but it should be "non-void methods must return a value" or some such

Actual behavior
---------------

The following error occurs:

```none
Index was out of range. Must be non-negative and less than the size of the collection.
Parameter name: Index
...
```

An analysis of the cause, courtesy of @BrucePay's previously linked answer:

> [...] the code that invokes the scriptblock expects there to be an object in the result List i.e. `result.Count == 1`. There are no objects in the pipeline case so you get an index-out-of-range error. [...]
> BTW - the [code in question is here](https://github.com/PowerShell/PowerShell/blob/c1c5344a8897262433141ecbc13bb06ac2c4bbef/src/System.Management.Automation/engine/lang/scriptblock.cs#L508).

Environment data
----------------

```powershell
PowerShell Core v6.1.0-preview.2 on macOS 10.13.4
PowerShell Core v6.1.0-preview.2 on Ubuntu 16.04.4 LTS
PowerShell Core v6.1.0-preview.2 on Microsoft Windows 10 Pro (64-bit; Version 1709, OS Build: 16299.371)
Windows PowerShell v5.1.17134.48 on Microsoft Windows 10 Pro (64-bit; Version 1709, OS Build: 16299.371)
```

BrucePay:
## PR Summary

Fix: #7128

If a method was typed to return [object] but no actual value was returned then you would get an index-out-of-range exception because the code expected a value in the result collection but there was none. The fix is to check to see if the array is empty. If it is, then return default(T).

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/lang/scriptblock.cs
+++ b/src/System.Management.Automation/engine/lang/scriptblock.cs
@@ -519,10 +519,18 @@ namespace System.Management.Automation
                                        invocationInfo: null,
                                        propagateAllExceptionsToTop: true,
                                        args: args);
+
+            // This is needed only for the case where the
+            // method returns [object]. If the argument to 'return'
+            // is a pipeline that emits nothing then result.Count will
+            // be zero so we catch that and "convert" it to null. Note that
+            // the return statement is still required in the method, it 
+            // just recieves nothing from it's argument.
             if (result.Count == 0)
             {
                 return default(T);
             }
+
             return (T)result[0];
         }
 
--- a/src/System.Management.Automation/engine/lang/scriptblock.cs
+++ b/src/System.Management.Automation/engine/lang/scriptblock.cs
@@ -519,7 +519,10 @@ namespace System.Management.Automation
                                        invocationInfo: null,
                                        propagateAllExceptionsToTop: true,
                                        args: args);
-            Diagnostics.Assert(result.Count == 1, "Code generation ensures we return the correct type");
+            if (result.Count == 0)
+            {
+                return default(T);
+            }
             return (T)result[0];
         }
 
--- a/test/powershell/Language/Classes/Scripting.Classes.BasicParsing.Tests.ps1
+++ b/test/powershell/Language/Classes/Scripting.Classes.BasicParsing.Tests.ps1
@@ -116,6 +116,16 @@ Describe 'Positive Parse Properties Tests' -Tags "CI" {
         class C9b { [System.Collections.Generic.List[C9b]] f() { return [C9b]::new() } }
         $c9b = [C9b]::new().f()
         It "Expected a System.Collections.Generic.List[C9b] returned" {  $c9b -is [System.Collections.Generic.List[C9b]] | Should -BeTrue }
+        It 'Methods returning object should return $null if no output was produced' {
+            class Foo {
+                [object] Bar1() { return & {} }
+                static [object] Bar2() { return & {} }
+            }
+            # Test instance method
+            [Foo]::new().Bar1() | Should -BeNullOrEmpty
+            # Test static method
+            [foo]::Bar2() | Should -BeNullOrEmpty
+        }
     }
 
     It 'Positive ParseProperty Attributes Test' {


test/powershell/Language/Classes/Scripting.Classes.BasicParsing.Tests.ps1
```diff
@@ -116,6 +116,16 @@ Describe 'Positive Parse Properties Tests' -Tags "CI" {
         class C9b { [System.Collections.Generic.List[C9b]] f() { return [C9b]::new() } }
         $c9b = [C9b]::new().f()
         It "Expected a System.Collections.Generic.List[C9b] returned" {  $c9b -is [System.Collections.Generic.List[C9b]] | Should -BeTrue }
+        It 'Methods returning object should return $null if no output was produced' {
+            class Foo {
+                [object] Bar1() { return & {} }
+                static [object] Bar2() { return & {} }
+            }
+            # Test instance method
+            [Foo]::new().Bar1() | Should -BeNullOrEmpty
```
iSazonov: We could explicitly check result type. I believe it make sense to use `testcase` for two return types (object and int/string) to ensure that default(T) works as expected.

src/System.Management.Automation/engine/lang/scriptblock.cs
```diff
@@ -519,7 +519,10 @@ internal T InvokeAsMemberFunctionT<T>(object instance, object[] args)
                                        invocationInfo: null,
                                        propagateAllExceptionsToTop: true,
                                        args: args);
-            Diagnostics.Assert(result.Count == 1, "Code generation ensures we return the correct type");
+            if (result.Count == 0)
```
lzybkr: I understand the change, but it feels really dirty because in an indirect way, it allows a method to not return a value, similar to the parser allowing the method to omit a `return` statement.

Unfortunately I can't think of a cleaner fix. I suppose it's worth adding a comment explaining that this is **not** to allow methods to omit a `return` statement.

test/powershell/Language/Classes/Scripting.Classes.BasicParsing.Tests.ps1
BrucePay: The *only* time this code path is hit is if the method returns `[object]`. All of the other cases are already handled elsewhere.  And the object returned is null. How do you propose checking it's type?

src/System.Management.Automation/engine/lang/scriptblock.cs
BrucePay: Yeah - it feels dirty to me but pipeline semantics make it necessary. I like the idea of adding a comment to explain it.

lzybkr (approved):
Note that a fix in the compiler is slightly nicer from a performance point of view - you would omit the check for non-object returning methods, but that fix is slightly more involved.

I would have asked for that **if** we could rely on the syntax, but the bug is still an issue with expressions:

```powershell
class C {
    static [object] f() {
        $o = & {} 
        return $o
    }
}
[C]::f()
```

test/powershell/Language/Classes/Scripting.Classes.BasicParsing.Tests.ps1
iSazonov: Sorry, it was my misunderstanding of default(T).

PR #7145, opened 2018-06-22
Refactor module code related to 'Get-Module -ListAvailable'

daxian-dbw:
## PR Summary

The major refactoring changes are:
- In `ModuleIntrisic.cs`, remove unneeded Windows-PowerShell-only code.
- In `ModuleUtils.cs`
   - use the new API `Directory.GetDirectories(string path, string searchPattern, EnumerationOptions enumerationOptions)` and `Directory.GetFiles(string path, string searchPattern, EnumerationOptions enumerationOptions)` to enumerate files and sub-directories within a directory path.
   - remove the unused parameter `bool force` from `GetDefaultAvailableModuleFiles(bool force, bool isForAutoDiscovery, ExecutionContext context)`
   - refactor the method `GetModuleVersionsFromAbsolutePath`. Add more comments and rename the method name.
- In `ModuleCmdletBase.cs`, refactor the method `GetModuleForNonRootedPaths` to `GetModuleForNames` to simply its implementation.
- In `PSModuleInfo.cs`
   - group the declarations of `Declared*Exports` fields together
   - rename `_detected*Exports` fields to `Detected*Exports` to group them together. They are internal fields and used outside `PSModuleInfo`.

### Pref improvement

> NOTE: Use Windows PowerShell default module paths to exercies the code with more modules.
> The following measurement are done without the crossgen'ed assemblies.

There is some perf improvement after this refactoring change:
- For `Get-Module -ListAvailable`, there is about 36% speed improvement for 94 default modules.
- For `Get-Module -ListAvailable -All`, there is about 14% speed improvement for totally 600 module files.
- For `Get-Module <name> -ListAvailable -List`, there is over 17x speed improvement for finding 13 modules from 600 modules. This is because we now filter names using the module file before creating a `PSMdouleInfo` object.

#### Before

```powershell
PS> $env:PSModulePath = "C:\Users\dongbow\Documents\WindowsPowerShell\Modules;C:\Program Files\WindowsPowerShell\Modules;C:\windows\system32\WindowsPowerShell\v1.0\Modules"
PS> $time = foreach ($i in 1..20) { Measure-Command { gmo -ListAvailable } | % TotalMilliseconds }
PS> $time | Measure-Object -Maximum -Minimum -Average | select Count, Maximum, Minimum, Average

Count  Maximum  Minimum    Average
-----  -------  -------    -------
   20 701.3139 623.5225 642.628685

PS> $time = foreach ($i in 1..20) { Measure-Command { gmo -ListAvailable -All } | % TotalMilliseconds }
PS> $time | Measure-Object -Maximum -Minimum -Average | select Count, Maximum, Minimum, Average

Count   Maximum   Minimum     Average
-----   -------   -------     -------
   20 1369.1893 1243.1195 1310.147175

PS> $time = foreach ($i in 1..20) { Measure-Command { gmo windows* -ListAvailable -All } | % TotalMilliseconds }
PS> $time | Measure-Object -Maximum -Minimum -Average | select Count, Maximum, Minimum, Average

Count   Maximum   Minimum    Average
-----   -------   -------    -------
   20 1535.3805 1292.2188 1353.62972
```

#### After

```powershell
PS> $env:PSModulePath = "C:\Users\dongbow\Documents\WindowsPowerShell\Modules;C:\Program Files\WindowsPowerShell\Modules;C:\windows\system32\WindowsPowerShell\v1.0\Modules"
PS> $time = foreach ($i in 1..20) { Measure-Command { gmo -ListAvailable } | % TotalMilliseconds }
PS> $time | Measure-Object -Maximum -Minimum -Average | select Count, Maximum, Minimum, Average

Count Maximum  Minimum  Average
----- -------  -------  -------
   20 438.338 393.7476 410.62353

PS> $time = foreach ($i in 1..20) { Measure-Command { gmo -ListAvailable -All } | % TotalMilliseconds }
PS> $time | Measure-Object -Maximum -Minimum -Average | select Count, Maximum, Minimum, Average

Count   Maximum   Minimum Average
-----   -------   -------   -------
   20 1341.5443 1055.1771 1120.4748

PS> $time = foreach ($i in 1..20) { Measure-Command { gmo windows* -ListAvailable -All } | % TotalMilliseconds }
PS> $time | Measure-Object -Maximum -Minimum -Average | select Count, Maximum, Minimum, Average

Count Maximum Minimum Average
----- ------- -------   -------
   20 78.0931 61.9288 66.194415
```

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
@@ -2635,8 +2635,8 @@ namespace Microsoft.PowerShell.Commands
                 bool needToAnalyzeScriptModules = usedWildcard;
 
                 // We can skip further analysis if 'FunctionsToExport', 'CmdletsToExport' and 'AliasesToExport'
-                // are all declared and no wildcard character is used for them. But if any of them are missing,
-                // we possibly need further analysis depending on the nested/root module types.
+                // are all declared and no wildcard character is used for them. But if any of 'FunctionsToExport',
+                // 'CmdletsToExport' or 'AliasesToExport' were not given, we must check to see if more analysis is needed.
                 if (!needToAnalyzeScriptModules && (!sawExportedCmdlets || !sawExportedFunctions || !sawExportedAliases))
                 {
                     foreach (var nestedModule in nestedModules)
--- a/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
@@ -875,28 +875,10 @@ namespace Microsoft.PowerShell.Commands
             // If no names were passed to this function, then this API will return list of all available modules
             if (names == null || moduleNames.Count > 0)
             {
-                modulesToReturn.AddRange(GetModuleForNonRootedPaths(moduleNames.ToArray(), all, refresh));
+                modulesToReturn.AddRange(GetModuleForNames(moduleNames.ToArray(), all, refresh));
             }
 
             return modulesToReturn;
-        }
-
-        private IEnumerable<PSModuleInfo> GetModuleForNonRootedPaths(string[] names, bool all, bool refresh)
-        {
-            const WildcardOptions wildcardOptions = WildcardOptions.IgnoreCase | WildcardOptions.CultureInvariant;
-            IEnumerable<WildcardPattern> patternList = SessionStateUtilities.CreateWildcardsFromStrings(names, wildcardOptions);
-
-            Dictionary<String, List<PSModuleInfo>> availableModules = GetAvailableLocallyModulesCore(names, all, refresh);
-            foreach (var entry in availableModules)
-            {
-                foreach (PSModuleInfo module in entry.Value)
-                {
-                    if (SessionStateUtilities.MatchesAnyWildcardPattern(module.Name, patternList, true))
-                    {
-                        yield return module;
-                    }
-                }
-            }
         }
 
         private IEnumerable<PSModuleInfo> GetModuleForRootedPaths(string[] modulePaths, bool all, bool refresh)
@@ -950,7 +932,7 @@ namespace Microsoft.PowerShell.Commands
 
                             var availableModuleFiles = all
                                 ? ModuleUtils.GetAllAvailableModuleFiles(resolvedModulePath)
-                                : ModuleUtils.GetModuleVersionsFromAbsolutePath(resolvedModulePath);
+                                : ModuleUtils.GetModuleFilesFromAbsolutePath(resolvedModulePath);
 
                             bool foundModule = false;
                             foreach (string file in availableModuleFiles)
@@ -998,38 +980,31 @@ namespace Microsoft.PowerShell.Commands
             return er;
         }
 
-        private Dictionary<String, List<PSModuleInfo>> GetAvailableLocallyModulesCore(string[] names, bool all, bool refresh)
+        private IEnumerable<PSModuleInfo> GetModuleForNames(string[] names, bool all, bool refresh)
         {
-            var modules = new Dictionary<string, List<PSModuleInfo>>(StringComparer.OrdinalIgnoreCase);
-            var modulePaths = ModuleIntrinsics.GetModulePath(false, Context);
-
+            IEnumerable<PSModuleInfo> allModules = null;
+            HashSet<string> modulePathSet = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
             bool cleanupModuleAnalysisAppDomain = Context.TakeResponsibilityForModuleAnalysisAppDomain();
 
             try
             {
                 foreach (string path in ModuleIntrinsics.GetModulePath(false, Context))
                 {
+                    string uniquePath = path.TrimEnd(Utils.Separators.Directory);
+
+                    // Ignore repeated module path.
+                    if (!modulePathSet.Add(uniquePath)) { continue; }
+
                     try
                     {
-                        var availableModules = all
-                            ? GetAllAvailableModules(path, refresh)
-                            : GetDefaultAvailableModules(names, path, refresh);
-
-                        // Add the path in $env:PSModulePath as the keys of the dictionary
-                        // If the paths are repeated, ignore the repetitions
-                        string uniquePath = path.TrimEnd(Utils.Separators.Directory);
-                        if (!modules.ContainsKey(uniquePath))
-                        {
-                            modules.Add(uniquePath, availableModules.OrderBy(m => m.Name).ToList());
-                        }
+                        IEnumerable<PSModuleInfo> modulesFound = GetModulesFromOneModulePath(
+                            names, uniquePath, all, refresh).OrderBy(m => m.Name);
+                        allModules = allModules == null ? modulesFound : allModules.Concat(modulesFound);
                     }
-                    catch (IOException)
+                    catch (Exception e) when (e is IOException || e is UnauthorizedAccessException)
                     {
-                        continue; // ignore directories that can't be accessed
-                    }
-                    catch (UnauthorizedAccessException)
-                    {
-                        continue; // ignore directories that can't be accessed
+                        // ignore directories that can't be accessed
+                        continue;
                     }
                 }
             }
@@ -1041,64 +1016,43 @@ namespace Microsoft.PowerShell.Commands
                 }
             }
 
-            return modules;
+            // Make sure we always return a non-null collection.
+            return allModules ?? Utils.EmptyArray<PSModuleInfo>();
         }
 
         /// <summary>
-        /// Get a list of all modules
-        /// which can be imported just by specifying a non rooted file name of the module
-        /// (Import-Module foo\bar.psm1;  but not Import-Module .\foo\bar.psm1)
+        /// Get modules based on the given names and module files.
         /// </summary>
-        private IEnumerable<PSModuleInfo> GetAllAvailableModules(string directory, bool refresh)
+        private IEnumerable<PSModuleInfo> GetModulesFromOneModulePath(string[] names, string modulePath, bool all, bool refresh)
         {
-            var availableModuleFiles = ModuleUtils.GetAllAvailableModuleFiles(directory);
-
-            foreach (string file in availableModuleFiles)
+            const WildcardOptions options = WildcardOptions.IgnoreCase | WildcardOptions.CultureInvariant;
+            IEnumerable<WildcardPattern> namePatterns = null;
+            if (names != null && names.Length > 0)
             {
-                PSModuleInfo module = CreateModuleInfoForGetModule(file, refresh);
-                if (module != null)
-                {
-                    yield return module;
-                }
+                namePatterns = SessionStateUtilities.CreateWildcardsFromStrings(names, options);
             }
-        }
 
-        /// <summary>
-        /// Get a list of the available modules
-        /// which can be imported just by specifying a non rooted directory name of the module
-        /// (Import-Module foo\bar;  but not Import-Module .\foo\bar or Import-Module .\foo\bar.psm1)
-        /// </summary>
-        private List<PSModuleInfo> GetDefaultAvailableModules(string[] name, string directory, bool refresh)
-        {
-            List<PSModuleInfo> availableModules = new List<PSModuleInfo>();
+            IEnumerable<string> moduleFiles = all
+                ? ModuleUtils.GetAllAvailableModuleFiles(modulePath)
+                : ModuleUtils.GetDefaultAvailableModuleFiles(modulePath);
 
-            const WildcardOptions wildcardOptions = WildcardOptions.IgnoreCase | WildcardOptions.CultureInvariant;
-            IEnumerable<WildcardPattern> patternList = SessionStateUtilities.CreateWildcardsFromStrings(name, wildcardOptions);
-
-            var availableModuleFiles = ModuleUtils.GetDefaultAvailableModuleFiles(directory);
-
-            foreach (string file in availableModuleFiles)
+            foreach (string file in moduleFiles)
             {
-                string actualModuleName = System.IO.Path.GetFileNameWithoutExtension(file);
-                if (SessionStateUtilities.MatchesAnyWildcardPattern(actualModuleName, patternList, true))
+                if (namePatterns == null ||
+                    SessionStateUtilities.MatchesAnyWildcardPattern(
+                        Path.GetFileNameWithoutExtension(file), namePatterns, defaultValue: true))
                 {
                     PSModuleInfo module = CreateModuleInfoForGetModule(file, refresh);
+                    if (module == null) { continue; }
 
-                    if (module != null)
+                    if (all || !ModuleUtils.IsModuleInVersionSubdirectory(file, out Version directoryVersion) || directoryVersion == module.Version)
                     {
-                        Version directoryVersion;
-                        if (!ModuleUtils.IsModuleInVersionSubdirectory(file, out directoryVersion)
-                            || directoryVersion == module.Version)
-                        {
-                            availableModules.Add(module);
-                        }
+                        yield return module;
                     }
                 }
             }
 
             ClearAnalysisCaches();
-
-            return availableModules;
         }
 
         /// <summary>
@@ -1191,10 +1145,10 @@ namespace Microsoft.PowerShell.Commands
                     moduleInfo = LoadModuleManifest(
                             scriptInfo,
                             flags /* - don't write errors, don't load elements */,
-                            null,
-                            null,
-                            null,
-                            null);
+                            minimumVersion: null,
+                            maximumVersion: null,
+                            requiredVersion: null,
+                            requiredModuleGuid: null);
                 }
                 else
                 {
@@ -1202,13 +1156,13 @@ namespace Microsoft.PowerShell.Commands
                     ImportModuleOptions options = new ImportModuleOptions();
                     bool found = false;
 
-                    moduleInfo = LoadModule(file, null, String.Empty, null, ref options, flags, out found);
+                    moduleInfo = LoadModule(file, moduleBase: null, prefix: String.Empty, ss: null, ref options, flags, out found);
                 }
 
                 // return fake PSModuleInfo if can't read the file for any reason
                 if (moduleInfo == null)
                 {
-                    moduleInfo = new PSModuleInfo(file, null, null);
+                    moduleInfo = new PSModuleInfo(file, context: null, sessionState: null);
                     moduleInfo.HadErrorsLoading = true;     // Prevent analysis cache from caching a bad module.
                 }
 
@@ -2677,9 +2631,13 @@ namespace Microsoft.PowerShell.Commands
                     }
                 }
 
-                var needToAnalyzeScriptModules = usedWildcard;
+                // We have to further analyze the module if any wildcard characters are used.
+                bool needToAnalyzeScriptModules = usedWildcard;
 
-                if (!needToAnalyzeScriptModules)
+                // We can skip further analysis if 'FunctionsToExport', 'CmdletsToExport' and 'AliasesToExport'
+                // are all declared and no wildcard character is used for them. But if any of them are missing,
+                // we possibly need further analysis depending on the nested/root module types.
+                if (!needToAnalyzeScriptModules && (!sawExportedCmdlets || !sawExportedFunctions || !sawExportedAliases))
                 {
                     foreach (var nestedModule in nestedModules)
                     {
@@ -3173,9 +3131,9 @@ namespace Microsoft.PowerShell.Commands
                     newManifestInfo.DeclaredCmdletExports = manifestInfo.DeclaredCmdletExports;
                 }
 
-                if (manifestInfo._detectedCmdletExports != null)
+                if (manifestInfo.DetectedCmdletExports != null)
                 {
-                    foreach (string detectedExport in manifestInfo._detectedCmdletExports)
+                    foreach (string detectedExport in manifestInfo.DetectedCmdletExports)
                     {
                         newManifestInfo.AddDetectedCmdletExport(detectedExport);
                     }
@@ -3187,9 +3145,9 @@ namespace Microsoft.PowerShell.Commands
                     newManifestInfo.DeclaredFunctionExports = manifestInfo.DeclaredFunctionExports;
                 }
 
-                if (manifestInfo._detectedFunctionExports != null)
+                if (manifestInfo.DetectedFunctionExports != null)
                 {
-                    foreach (string detectedExport in manifestInfo._detectedFunctionExports)
+                    foreach (string detectedExport in manifestInfo.DetectedFunctionExports)
                     {
                         newManifestInfo.AddDetectedFunctionExport(detectedExport);
                     }
@@ -3200,9 +3158,9 @@ namespace Microsoft.PowerShell.Commands
                     newManifestInfo.DeclaredAliasExports = manifestInfo.DeclaredAliasExports;
                 }
 
-                if (manifestInfo._detectedAliasExports != null)
+                if (manifestInfo.DetectedAliasExports != null)
                 {
-                    foreach (var pair in manifestInfo._detectedAliasExports)
+                    foreach (var pair in manifestInfo.DetectedAliasExports)
                     {
                         newManifestInfo.AddDetectedAliasExport(pair.Key, pair.Value);
                     }
@@ -3269,15 +3227,7 @@ namespace Microsoft.PowerShell.Commands
                         }
                         else
                         {
-                            Collection<string> v = new Collection<string>();
-                            if (manifestInfo.DeclaredFunctionExports != null)
-                            {
-                                foreach (var f in manifestInfo.DeclaredFunctionExports)
-                                {
-                                    v.Add(f);
-                                }
-                            }
-                            UpdateCommandCollection(v, exportedFunctions);
+                            UpdateCommandCollection(manifestInfo.DeclaredFunctionExports, exportedFunctions);
                         }
                     }
 
--- a/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
@@ -869,19 +869,19 @@ namespace Microsoft.PowerShell.Commands
                         moduleNames.Add(n);
                     }
                 }
-                modulesToReturn.AddRange(GetModuleForRootedPaths(modulePaths.ToArray(), all, refresh));
+                modulesToReturn.AddRange(GetModuleForRootedPaths(modulePaths, all, refresh));
             }
 
             // If no names were passed to this function, then this API will return list of all available modules
             if (names == null || moduleNames.Count > 0)
             {
-                modulesToReturn.AddRange(GetModuleForNames(moduleNames.ToArray(), all, refresh));
+                modulesToReturn.AddRange(GetModuleForNames(moduleNames, all, refresh));
             }
 
             return modulesToReturn;
         }
 
-        private IEnumerable<PSModuleInfo> GetModuleForRootedPaths(string[] modulePaths, bool all, bool refresh)
+        private IEnumerable<PSModuleInfo> GetModuleForRootedPaths(List<string> modulePaths, bool all, bool refresh)
         {
             // This is to filter out duplicate modules
             var modules = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
@@ -980,7 +980,7 @@ namespace Microsoft.PowerShell.Commands
             return er;
         }
 
-        private IEnumerable<PSModuleInfo> GetModuleForNames(string[] names, bool all, bool refresh)
+        private IEnumerable<PSModuleInfo> GetModuleForNames(List<string> names, bool all, bool refresh)
         {
             IEnumerable<PSModuleInfo> allModules = null;
             HashSet<string> modulePathSet = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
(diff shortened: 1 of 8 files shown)


src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -2677,9 +2630,13 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
                     }
                 }
 
+                // We have to further analyze the module if any wildcard characters are used.
                 var needToAnalyzeScriptModules = usedWildcard;
```
rjmholt: Can this be changed to `bool`? Just one extra character

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
```diff
@@ -715,14 +715,12 @@ private static string AddToPath(string basePath, string pathToAdd, int insertPos
 
         /// <summary>
         /// Check if the current powershell is likely running in following scenarios:
-        ///  - sxs ps started on windows [machine-wide env:PSModulePath will influence]
-        ///  - sxs ps started from full ps
-        ///  - sxs ps started from inbox nano/iot ps
-        ///  - full ps started from sxs ps
-        ///  - inbox nano/iot ps started from sxs ps
+        ///  - PSCore started on windows [machine-wide env:PSModulePath will influence]
```
rjmholt: This is really helpful. Thanks!

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
```diff
@@ -733,30 +731,19 @@ private static bool NeedToClearProcessModulePath(string currentProcessModulePath
...
+            // The machine-wide and user-wide environment variables are only meaningful for full ps,
+            // so if the current process module path contains any of them, it's likely that the sxs
+            // ps was started directly on windows, or from full ps. The same goes for the legacy personal
+            // and shared module paths.
+            string hklmModulePath = GetExpandedEnvironmentVariable(Constants.PSModulePathEnvVar, EnvironmentVariableTarget.Machine);
+            string hkcuModulePath = GetExpandedEnvironmentVariable(Constants.PSModulePathEnvVar, EnvironmentVariableTarget.User);
+            string legacyPersonalModulePath = personalModulePath.Replace(winSxSModuleDirectory, winLegacyModuleDirectory);
+            string legacyProgramFilesModulePath = sharedModulePath.Replace(winSxSModuleDirectory, winLegacyModuleDirectory);
+
+            return (!string.IsNullOrEmpty(hklmModulePath) && currentProcessModulePath.IndexOf(hklmModulePath, StringComparison.OrdinalIgnoreCase) != -1) ||
```
rjmholt: Any chance of breaking this expression up?

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
```diff
@@ -860,7 +844,7 @@ public static string GetModulePath(string currentProcessModulePath, string hklmM
             }
             // EVT.Process exists
             // Now handle the case where the environment variable is already set.
-            else if (runningSxS) // The running powershell is an SxS PS instance
+            else
```
rjmholt: Oh wow this makes things much nicer.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -11,6 +11,18 @@ namespace System.Management.Automation.Internal
...
     internal static class ModuleUtils
     {
+        // Default option for local file system enumeration:
+        //  - Ignore files/directories when access is denied;
+        //  - Search top directory only.
+        private readonly static EnumerationOptions s_defaultEnumOptions = new EnumerationOptions() { AttributesToSkip = 0 };
+
+        // Default option for UNC path enumeration. Same as above plus a large buffer size.
+        // For network shares, a large buffer may result in better performance as more results can be batched over the wire.
+        private readonly static EnumerationOptions s_uncPathEnumOptions = new EnumerationOptions() { AttributesToSkip = 0, BufferSize = 16384 };
```
rjmholt: Is there a reason for the buffer size to be this specific number (2^14)?

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -43,22 +54,25 @@ internal static bool IsPossibleModuleDirectory(string dir)
...
-        /// info only for the first one</remarks>
+        /// <remarks>
+        /// When obtaining all module files we return all possible combinations for a given file.
+        /// For example, for 'foo' we return both 'foo.psd1' and 'foo.psm1' if found.
+        /// Get-Module will create the module info only for the first one.
+        /// </remarks>
         internal static IEnumerable<string> GetAllAvailableModuleFiles(string topDirectoryToCheck)
         {
+            if (!Directory.Exists(topDirectoryToCheck)) { yield break; }
+            var options = Utils.PathIsUnc(topDirectoryToCheck) ? s_uncPathEnumOptions : s_defaultEnumOptions;
```
rjmholt: Would it be possible to use `EnumerationOptions` instead of `var` here?

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -43,22 +54,25 @@ internal static bool IsPossibleModuleDirectory(string dir)
...
-        /// foo.psd1 and foo.psm1 if found. Get-Module will create the module
-        /// info only for the first one</remarks>
+        /// <remarks>
+        /// When obtaining all module files we return all possible combinations for a given file.
+        /// For example, for 'foo' we return both 'foo.psd1' and 'foo.psm1' if found.
+        /// Get-Module will create the module info only for the first one.
+        /// </remarks>
         internal static IEnumerable<string> GetAllAvailableModuleFiles(string topDirectoryToCheck)
         {
+            if (!Directory.Exists(topDirectoryToCheck)) { yield break; }
```
rjmholt: Probably just personal preference, but on a single line I find this escape hatch fairly easy to miss (it just looks like any other statement next to other assignments). Maybe worth giving it a full block and newline?

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -43,22 +54,25 @@ internal static bool IsPossibleModuleDirectory(string dir)
...
+            var options = Utils.PathIsUnc(topDirectoryToCheck) ? s_uncPathEnumOptions : s_defaultEnumOptions;
+            HashSet<int> fileAtIndexWasReturned = new HashSet<int>();
```
rjmholt: This variable reads like a boolean to me. Maybe `returnedFileIndexes`? (Part of me wants to suggest `returnedFileIndices`, but clarity is better than "correctness" I think)

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -79,22 +93,30 @@ internal static IEnumerable<string> GetAllAvailableModuleFiles(string topDirecto
                 catch (IOException) { }
                 catch (UnauthorizedAccessException) { }
 
-                var files = Directory.GetFiles(directoryToCheck, "*", SearchOption.TopDirectoryOnly);
-                foreach (string moduleFile in files)
+                fileAtIndexWasReturned.Clear();
```
rjmholt: Is there a chance the `HashSet` won't be empty at this point? If it needs to be guaranteed empty, maybe it's worth moving the initialisation down to above the `foreach` where it's used

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -79,22 +93,30 @@ internal static IEnumerable<string> GetAllAvailableModuleFiles(string topDirecto
...
+                var files = Directory.GetFiles(directoryToCheck, "*", options);
```
rjmholt: Maybe use `FileInfo` instead of `var`?

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -146,59 +168,75 @@ internal static IEnumerable<string> GetDefaultAvailableModuleFiles(bool force, b
...
-            Version moduleVersion;
-            // if the user give the module path including version, we should be able to find the module as well
-            if (Version.TryParse(fileName, out moduleVersion) && Directory.Exists(Directory.GetParent(directory).ToString()))
-            {
-                fileName = Directory.GetParent(directory).Name;
-            }
-            foreach (var version in GetModuleVersionSubfolders(directory))
+
+            // If the given directory doesn't exist or it's the root folder, then return an empty list.
+            if (!Directory.Exists(directory) || fileName == string.Empty) { return result; }
```
rjmholt: MSDN [suggests using `String.IsNullOrEmpty()`](https://msdn.microsoft.com/en-us/library/system.string.empty(v=vs.110).aspx#Anchor_1) instead of `== String.Empty` here. Does a `null` filename mean something else?

rjmholt: Also, it might be worth moving these checks up above where we allocate the list

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -998,30 +969,38 @@ private ErrorRecord CreateModuleNotFoundError(string modulePath)
...
-                        string uniquePath = path.TrimEnd(Utils.Separators.Directory);
-                        if (!modules.ContainsKey(uniquePath))
-                        {
-                            modules.Add(uniquePath, availableModules.OrderBy(m => m.Name).ToList());
-                        }
+                        var availableModuleFiles = all
+                            ? ModuleUtils.GetAllAvailableModuleFiles(uniquePath)
+                            : ModuleUtils.GetDefaultAvailableModuleFiles(uniquePath);
+
+                        availableModules = GetModulesBasedOnGivenModuleFiles(names, availableModuleFiles, refresh, all).OrderBy(
```
rjmholt: This logic might actually be very useful as a method on `PSModuleInfo`. The formatter for `Get-Module -ListAvailable` [also does something like this](https://github.com/PowerShell/PowerShell/blob/bbb4f2ea84f6c4a8e8775978178826b61db20bda/src/System.Management.Automation/FormatAndOutput/DefaultFormatters/PowerShellCore_format_ps1xml.cs#L1187-L1203). I know that in PowerShellEditorServices we wanted a method or property like this.

I admit I'm also a bit averse to such a large inline delegate for a LINQ method.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -146,59 +168,75 @@ internal static IEnumerable<string> GetDefaultAvailableModuleFiles(bool force, b
...
-                string manifestPath = qualifiedPathWithVersion + StringLiterals.PowerShellDataFileExtension;
-                if (File.Exists(manifestPath))
-                {
-                    bool isValidModuleVersion = version.Equals(ModuleIntrinsics.GetManifestModuleVersion(manifestPath));
+                string parentDirPath = Path.GetDirectoryName(directory);
+                string parentDirName = Path.GetFileName(parentDirPath);
 
-                    if (isValidModuleVersion)
+                // If the parent directory is NOT a root folder, then it could be the module folder.
+                if (parentDirName != string.Empty)
```
rjmholt: Maybe `String.IsNullOrEmpty` is better here?

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -146,59 +168,75 @@ internal static IEnumerable<string> GetDefaultAvailableModuleFiles(bool force, b
...
+
+            // If the user give the module path including version, the module name could be the parent folder name.
+            if (Version.TryParse(fileName, out Version ver))
             {
-                var qualifiedPathWithVersion = Path.Combine(directory, Path.Combine(version.ToString(), fileName));
-                string manifestPath = qualifiedPathWithVersion + StringLiterals.PowerShellDataFileExtension;
-                if (File.Exists(manifestPath))
-                {
-                    bool isValidModuleVersion = version.Equals(ModuleIntrinsics.GetManifestModuleVersion(manifestPath));
+                string parentDirPath = Path.GetDirectoryName(directory);
```
rjmholt: I assume rather than throw this returns `null` or `""`?

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -146,59 +168,75 @@ internal static IEnumerable<string> GetDefaultAvailableModuleFiles(bool force, b
...
                     {
                         result.Add(manifestPath);
+                        return result;
                     }
                 }
             }
 
-            foreach (string ext in ModuleIntrinsics.PSModuleExtensions)
+            // If we reach here, then use the given directory as the module folder.
+            foreach (var version in GetModuleVersionSubfolders(directory))
```
rjmholt: Does this method return a `Version`?

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -146,59 +168,75 @@ internal static IEnumerable<string> GetDefaultAvailableModuleFiles(bool force, b
...
+                manifestPath += StringLiterals.PowerShellDataFileExtension;
+                if (File.Exists(manifestPath) && version.Equals(ModuleIntrinsics.GetManifestModuleVersion(manifestPath)))
                 {
-                    continue;
+                    result.Add(manifestPath);
                 }
+            }
 
-                result.Add(moduleFile);
+            foreach (string ext in ModuleIntrinsics.PSModuleExtensions)
```
rjmholt: This logic looks very similar to the `Import-Module` local load logic:
https://github.com/PowerShell/PowerShell/blob/bbb4f2ea84f6c4a8e8775978178826b61db20bda/src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs#L530-L760

Where we first check the path, then the folder, then the versions, then the extensions, etc.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -146,59 +168,75 @@ internal static IEnumerable<string> GetDefaultAvailableModuleFiles(bool force, b
...
         /// <summary>
-        /// Get a list of the available module files
-        /// which can be imported just by specifying a non rooted directory name of the module
-        /// (Import-Module foo\bar;  but not Import-Module .\foo\bar or Import-Module .\foo\bar.psm1)
+        /// Get a list of the available module files from the given directory.
+        /// Search all module folders under the specified directory, but do not search sub-directories under a module folder.
         /// </summary>
         internal static IEnumerable<string> GetDefaultAvailableModuleFiles(string topDirectoryToCheck)
         {
+            if (!Directory.Exists(topDirectoryToCheck)) { yield break; }
```
rjmholt: Same comments here as above

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
bergmeister: I agree, `NullOrEmpty()` is the better practice, makes it easier to read an less error prone since NullReferenceExceptions are the most expensive unintentional exceptions in this world.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -923,56 +905,45 @@ private IEnumerable<PSModuleInfo> GetModuleForRootedPaths(string[] modulePaths,
...
-                                if (!modules.Contains(resolvedModulePath))
+                                PSModuleInfo module = CreateModuleInfoForGetModule(resolvedModulePath, refresh);
+                                if (module != null)
                                 {
-                                    modules.Add(resolvedModulePath);
                                     yield return module;
                                 }
                             }
                         }
                         else
```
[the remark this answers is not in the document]
daxian-dbw: The method `GetModuleForRootedPaths` is massively refactored to change the behavior of `Get-Module <directory-path> -ListAvailable` and `Get-Module <directory-path> -ListAvailable -All`. The new behvaior will make them the same as if running `Get-Module -ListAvailable` and `Get-Module -ListAvailable -All` when `$env:PSModulePath == <directory-path>`.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -923,56 +905,45 @@ private IEnumerable<PSModuleInfo> GetModuleForRootedPaths(string[] modulePaths,
...
                         {
-                            // Given path is a directory, we first check if it is end with module version.
-                            Version version;
-                            if (Version.TryParse(moduleName, out version))
-                            {
-                                moduleName = Path.GetFileName(Directory.GetParent(resolvedModulePath).Name);
-                            }
-
                             var availableModuleFiles = all
```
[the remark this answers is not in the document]
daxian-dbw: Fixed.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -923,56 +905,45 @@ private IEnumerable<PSModuleInfo> GetModuleForRootedPaths(string[] modulePaths,
...
-                                        // We need to list all versions of the module.
-                                        string subModulePath = Path.GetDirectoryName(file);
-                                        if (!modules.Contains(subModulePath))
-                                        {
-                                            modules.Add(subModulePath);
-                                            yield return module;
-                                        }
-                                    }
+                                    foundModule = true;
+                                    yield return module;
```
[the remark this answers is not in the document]
daxian-dbw: The method `GetModuleForRootedPaths` is massively refactored. Please review again. Thanks!

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -998,30 +969,38 @@ private ErrorRecord CreateModuleNotFoundError(string modulePath)
...
-                        if (!modules.ContainsKey(uniquePath))
-                        {
-                            modules.Add(uniquePath, availableModules.OrderBy(m => m.Name).ToList());
-                        }
+                        var availableModuleFiles = all
+                            ? ModuleUtils.GetAllAvailableModuleFiles(uniquePath)
+                            : ModuleUtils.GetDefaultAvailableModuleFiles(uniquePath);
+
+                        availableModules = GetModulesBasedOnGivenModuleFiles(names, availableModuleFiles, refresh).OrderBy(
+                                                module => {
```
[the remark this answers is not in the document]
daxian-dbw: The `OrderBy` call was previous `OrderBy(m => m.Name)`. However, simply ordering by name will cause the modules from the same parent folder to be divided by some modules in more nested sub-folders. I have moved this delegate to a method in `ModuleUtils.cs`.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -998,30 +969,38 @@ private ErrorRecord CreateModuleNotFoundError(string modulePath)
...
+                                                    var moduleBase = module.ModuleBase;
```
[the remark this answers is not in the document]
daxian-dbw: Fixed.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -998,30 +969,38 @@ private ErrorRecord CreateModuleNotFoundError(string modulePath)
             return er;
         }
 
-        private Dictionary<String, List<PSModuleInfo>> GetAvailableLocallyModulesCore(string[] names, bool all, bool refresh)
+        private IEnumerable<PSModuleInfo> GetModuleForNames(string[] names, bool all, bool refresh)
```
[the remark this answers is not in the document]
daxian-dbw: It's better but still not strictly following the path order. Look at `ModuleUtils.GetDefaultAvailableModuleFiles`, we make `Microsoft.PowerShell.Management` and `Microsoft.PowerShell.Utility` more preferred than others ...

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
daxian-dbw: Sure. Fixed.

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
daxian-dbw: I looked at this expression, it's not as confusing as it looks. I also found the existing comments for the method and in the method very helpful.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
daxian-dbw: It's recommended in the comments of `BufferSize`. I add a comment about it.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
daxian-dbw: Sure, fixed.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
daxian-dbw: The style of putting a simple statement/expression in this way is used across the code base, and it makes the code concise to some degree. I will add a new line to separate it from the assignments statements below.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
daxian-dbw: I think `indexesOfReturnedFiles` may be better.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
daxian-dbw: Not sure I understand your comment. This is to make sure it's empty when we start to process all the files from the directory.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
daxian-dbw: It's `string[]`. Fixed.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
daxian-dbw: The `GetModuleFilesFromAbsolutePath` method is removed along with the refactor to `GetModuleForRootedPaths`. It's not needed for the proposed new behaviors.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
daxian-dbw: The method `GetModulesBasedOnGivenModuleFiles` was removed to further simply `GetModuleForNames`.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
daxian-dbw: This method is removed now.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
rjmholt: Just mean that there don't seem to be any references to `fileAtIndexWasReturned` before the call to `Clear()` — so is there a need to call `Clear()`?

If there's a strong need to ensure that `fileAtIndexWasReturned` is empty when it's used later, then maybe it should just be defined and created later, above the foreach loop.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
daxian-dbw: The method is removed now.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
daxian-dbw: It returns `List<Version>`. But the method is removed now.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
daxian-dbw: Add a new line.

daxian-dbw:
@rjmholt I have reverted the breaking change I made to the scenario of `Get-Module <Path-to-directory> -ListAvailable`. Now the changes introduce no change to the current behaviors. Can you please continue the review?

daxian-dbw:
Added tests for different scenarios of `Get-Module -ListAvailable` and `Get-Module -ListAvailable -All`. All tests pass in `pwsh` with or without changes in this PR.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -1041,64 +1016,43 @@ private ErrorRecord CreateModuleNotFoundError(string modulePath)
...
         }
 
         /// <summary>
-        /// Get a list of all modules
-        /// which can be imported just by specifying a non rooted file name of the module
-        /// (Import-Module foo\bar.psm1;  but not Import-Module .\foo\bar.psm1)
+        /// Get modules based on the given names and module files.
         /// </summary>
-        private IEnumerable<PSModuleInfo> GetAllAvailableModules(string directory, bool refresh)
+        private IEnumerable<PSModuleInfo> GetModulesFromOneModulePath(string[] names, string modulePath, bool all, bool refresh)
```
rjmholt: How hard is the requirement on `names` being an array? I only ask because I notice this part of the code tends to have collection types in signatures vary a lot (some `IEnumerable<string>`, some `string[]`, some `List<string>`) and we should probably work out a better rationale for what types we prefer.

If there is a particular performance implication from using `string[]` then that makes sense, but in general I think we should favour the principle of using an interface with only the properties required by each method; if we only want to go through entries sequentially, `IEnumerable<T>`, if we want to know how many there are up front, something like `ICollection<T>`, if we want indexing `IList<T>`.

I'm not making a particular suggestion here I suppose. I just have the impression that the various collection types used in the methods in the module cmdlets weren't used for any particular reason, and we should be more judicious about what collection types we use (and the semantics of each) given how intensively we handle them.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
daxian-dbw: This method returns an empty string. I changed to use `string.IsNullOrEmpty` below.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -2677,9 +2631,13 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
...
 
-                var needToAnalyzeScriptModules = usedWildcard;
+                // We have to further analyze the module if any wildcard characters are used.
+                bool needToAnalyzeScriptModules = usedWildcard;
 
-                if (!needToAnalyzeScriptModules)
+                // We can skip further analysis if 'FunctionsToExport', 'CmdletsToExport' and 'AliasesToExport'
+                // are all declared and no wildcard character is used for them. But if any of them are missing,
+                // we possibly need further analysis depending on the nested/root module types.
+                if (!needToAnalyzeScriptModules && (!sawExportedCmdlets || !sawExportedFunctions || !sawExportedAliases))
```
rjmholt: The comment above I read as saying "we can skip analysis if functions/cmdlets/aliases to export are given".

But I would translate that as the condition:
```csharp
!needToAnalyzeScriptModules && (sawExportedCmdlets && sawExportedFunctions && sawExportedAliases))
```

The expression here is equivalent to:
```csharp
!needToAnalyzeScriptModules && !(sawExportedCmdlets && sawExportedFunctions && sawExportedAliases))
```

Which I think would be "analysis is not required and not all of functions/cmdlets/aliases to export were present".

Do the variable names mean something converse to how I read them?

test/powershell/Modules/Microsoft.PowerShell.Core/Get-Module.Tests.ps1
```diff
@@ -0,0 +1,104 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
rjmholt: A much needed file!

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
daxian-dbw: For this method, it makes more sense to use `List<string>`, so the top-level caller doesn't need to cast the List to a string array.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
daxian-dbw: > The comment above I read as saying "we can skip analysis if all of functions/cmdlets/aliases to export are given".

That's correct.
That means if `needToAnalyzeScriptModules == false` when we reaching here, we might still change it to `true` if we didn't see one of "FunctionToExport", "CmdletToExport" and "AliasToExport".

One thing worth to note is the `if block` here is not doing further analysis, but checking if we really need to do further analysis -- it's changing the value of `needToAnalyzeScriptModules` to `true`.

test/powershell/Modules/Microsoft.PowerShell.Core/Get-Module.Tests.ps1
daxian-dbw: Hopefully, more tests can be added gradually.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
rjmholt: Oh right! The comment is slightly unclear to me because I didn't realise that the `if` is really a check to see if we maybe do need to do further analysis.

I would have worded it more as "If any of 'FunctionsToExport', 'CmdletsToExport' or 'AliasesToExport' were not given, we must check to see if more analysis is needed". Not saying the wording should be that way, just submitting an alternate description

rjmholt (approved):
Ok, this looks good to me now!

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
daxian-dbw: Good suggestion. I like your comment better. Will update.

PR #7151, opened 2018-06-22
Make switch statement report correct error position when it fails to evaluate the condition

### LINKED ISSUE #7150: switch statement report incorrect error position when it fails to evaluate the condition
Steps to reproduce
------------------

```powershell
PS> cat .\about.ps1
$Test = 1;
switch ($NeverUsedBefore[0]) {
    "a" {};
}
```

Expected behavior
-----------------
Run `about.ps1` reports the correct script location.
```none
PS> .\about.ps1
Cannot index into a null array.
At F:\tmp\about.ps1:2 char:9
+ switch ($NeverUsedBefore[0]) {
+         ~~~~~~~~~~~~~~~~~~~
+ FullyQualifiedErrorId : NullArray
```

Actual behavior
---------------
Run `about.ps1` reports the wrong script location.
```none
PS> .\about.ps1
Cannot index into a null array.
At F:\tmp\about.ps1:1 char:1
+ $Test = 1;
+ ~~~~~~~~~
+ FullyQualifiedErrorId : NullArray
```

Environment data
----------------

```powershell
> $PSVersionTable
```

daxian-dbw:
## PR Summary

Fix #7150
Make switch statement report correct error position when it fails to evaluate the condition

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/parser/Compiler.cs
+++ b/src/System.Management.Automation/engine/parser/Compiler.cs
@@ -4189,10 +4189,7 @@ namespace System.Management.Automation.Language
             // $foreach/$switch = GetEnumerator $enumerable
             var enumerable = NewTemp(typeof(object), "enumerable");
             temps.Add(enumerable);
-            if (generatingForeach)
-            {
-                exprs.Add(UpdatePosition(stmt.Condition));
-            }
+            exprs.Add(UpdatePosition(stmt.Condition));
             exprs.Add(
                 Expression.Assign(enumerable,
                                   GetRangeEnumerator(stmt.Condition.GetPureExpression())
--- a/test/powershell/Language/Scripting/ErrorPosition.Tests.ps1
+++ b/test/powershell/Language/Scripting/ErrorPosition.Tests.ps1
@@ -1,0 +1,51 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+Describe "Error position Tests" -Tags "CI" {
+    It "switch condition evaluation failure should report correct error position" {
+        $testFile = Join-Path $TestDrive "SwitchError1.ps1"
+        Set-Content -Path $testFile -Encoding Ascii -Value @'
+$test = 1
+switch ($nullVar[0]) {
+    "a" {};
+}
+'@
+        try { & $testFile } catch { $errorRecord = $_ }
+        $errorRecord | Should -Not -BeNullOrEmpty
+        $errorRecord.ScriptStackTrace | Should -Match "SwitchError1.ps1: line 2"
+    }
+
+    It "swtich condition MoveNext failure should report correct error position" {
+        $code = @'
+using System;
+using System.Collections.Generic;
+namespace SwitchTest
+{
+    public class Test
+    {
+        public static IEnumerable<string> GetName()
+        {
+            yield return "Hello world";
+            throw new ArgumentException();
+        }
+    }
+}
+'@
+        $testFile = Join-Path $TestDrive "SwitchError2.ps1"
+        Set-Content -Path $testFile -Encoding Ascii -Value @'
+$test = 1
+$enumerable = [SwitchTest.Test]::GetName()
+switch ($enumerable) {
+    "hello world" { $test = 1; $_ }
+    "Yay" { $test = 2; $_ }
+}
+'@
+        if (-not ("SwitchTest.Test" -as [type])) {
+            Add-Type -TypeDefinition $code
+        }
+        
+        try { & $testFile > $null } catch { $errorRecord = $_ }
+        $errorRecord | Should -Not -BeNullOrEmpty
+        $errorRecord.ScriptStackTrace | Should -Match "SwitchError2.ps1: line 3"
+    }
+}
--- a/test/powershell/Language/Scripting/ErrorPosition.Tests.ps1
+++ b/test/powershell/Language/Scripting/ErrorPosition.Tests.ps1
@@ -6,7 +6,7 @@ Describe "Error position Tests" -Tags "CI" {
         $testFile = Join-Path $TestDrive "SwitchError1.ps1"
         Set-Content -Path $testFile -Encoding Ascii -Value @'
 $test = 1
-switch ($nullVar[0]) {
+switch ($null[0]) {
     "a" {};
 }
 '@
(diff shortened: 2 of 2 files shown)


test/powershell/Language/Scripting/ErrorPosition.Tests.ps1
```diff
@@ -0,0 +1,51 @@
...
+    "a" {};
+}
+'@
+        try { & $testFile } catch { $errorRecord = $_ }
+        $errorRecord | Should -Not -BeNullOrEmpty
+        $errorRecord.ScriptStackTrace | Should -Match "SwitchError1.ps1: line 2"
+    }
+
+    It "swtich condition MoveNext failure should report correct error position" {
+        $code = @'
```
BrucePay: typo swtich -> switch

test/powershell/Language/Scripting/ErrorPosition.Tests.ps1
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+Describe "Error position Tests" -Tags "CI" {
+    It "switch condition evaluation failure should report correct error position" {
+        $testFile = Join-Path $TestDrive "SwitchError1.ps1"
+        Set-Content -Path $testFile -Encoding Ascii -Value @'
+$test = 1
+switch ($nullVar[0]) {
+    "a" {};
```
BrucePay: could use `$null` here  `switch ($null[0]) {`

test/powershell/Language/Scripting/ErrorPosition.Tests.ps1
```diff
@@ -0,0 +1,51 @@
...
+}
+'@
+        try { & $testFile } catch { $errorRecord = $_ }
+        $errorRecord | Should -Not -BeNullOrEmpty
+        $errorRecord.ScriptStackTrace | Should -Match "SwitchError1.ps1: line 2"
+    }
+
+    It "swtich condition MoveNext failure should report correct error position" {
```
SteveL-MSFT: typo: swtich

src/System.Management.Automation/engine/parser/Compiler.cs
```diff
@@ -4189,10 +4189,7 @@ private Expression GenerateDoLoop(LoopStatementAst loopStatement)
             // $foreach/$switch = GetEnumerator $enumerable
             var enumerable = NewTemp(typeof(object), "enumerable");
             temps.Add(enumerable);
-            if (generatingForeach)
-            {
-                exprs.Add(UpdatePosition(stmt.Condition));
-            }
+            exprs.Add(UpdatePosition(stmt.Condition));
```
BrucePay: So simple 👍

BrucePay (approved):

SteveL-MSFT (changes requested):

rjmholt (approved):
@BrucePay and @SteveL-MSFT seem to have addressed all concerns 👍

test/powershell/Language/Scripting/ErrorPosition.Tests.ps1
```diff
@@ -0,0 +1,51 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
```
iSazonov: Why we need new file? 
It seems we can use existing.

SteveL-MSFT (approved):

test/powershell/Language/Scripting/ErrorPosition.Tests.ps1
daxian-dbw: Fixed.

test/powershell/Language/Scripting/ErrorPosition.Tests.ps1
daxian-dbw: Fixed. Thanks

test/powershell/Language/Scripting/ErrorPosition.Tests.ps1
daxian-dbw: I used a new file because I couldn't find an existing test file that this category of issues can fall in.
Here we are not testing indexing into a null array, but the scenario where it fails to evaluate the switch statement's condition (it could fail for a different reason, and I just happen to use `$null[0]`).

As for the test you referenced, it looks more fit in `Language/Scripting` instead of `Language/Parser` because the error that is tested is a runtime error instead of a parsing one.

src/System.Management.Automation/engine/parser/Compiler.cs
daxian-dbw: We generate code to update position in the script that is executing at various places, so when an error happens, the error can be associated with a position in the script. Here we didn't update the position for the `condition` when evaluating the condition from a switch statement.

rjmholt: Interesting that the previous code went out of its way to only work for `foreach` loops

test/powershell/Language/Scripting/ErrorPosition.Tests.ps1
iSazonov: Yes, I don't found a file with runtime error tests too. And I wonder. If we need better code coverage perhaps we should rename the file from `ErrorPosition` to `RuntimeErrors`.

daxian-dbw: Runtime errors usually are tested along with the feature that they are associated with, no need for a separate test file only for validating runtime errors.
Error position is a feature used for reporting errors, so I think it's better to have a test file of its own category.

src/System.Management.Automation/engine/parser/Compiler.cs
iSazonov: Yes, I wonder why it was so? If we didn't have tests for that previously didn't we break something now?

daxian-dbw: Yeah, I did break something 😄
When stepping over the script in debugging, now the debugger stops on `switch ($b)` twice. I will revert this PR.

iSazonov: Don't add me to CI list :smile:

PR #7171, opened 2018-06-25
Add type inference for Select-Object command

### LINKED ISSUE #7170: Add type inference for Select-Object
Steps to reproduce
------------------

```powershell
get-childitem | Select-Object -first 1 | foreach <Tab>
```

Expected behavior
-----------------

```none
Tab completion of the properties of the pipeline type
```

Actual behavior
---------------

```none
No completion occurs
```

Environment data
----------------

```powershell
> $PSVersionTable

PSVersion                      6.0.2
GitCommitId                    v6.0.2
```

powercode:
## PR Summary

Add type inference for Select-Object command

Fixes #7170

## PR Checklist

- [x] [PR has a meaningful title]
- [x] [Summarized changes] 
This change add type inference for Select-Object in the cases when no projection is done.
I.E. when neither `-Property` or `-ExcludeProperty` is specified. 
- [x] [Change is not breaking]
- [x] [Make sure all `.h`, `.cpp`, `.cs`, `.ps1` and `.psm1` files have the correct copyright header] 
- [x] This PR is ready to merge and is not [Work in Progress]
- **User-facing changes**
    - [x] Not Applicable
    - [x] [Make sure you've added a new test if existing tests do not effectively test the code changed]
    - [ ] [Add `[feature]` if the change is significant or affects feature tests]

### DIFF
--- a/src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
+++ b/src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
@@ -786,9 +786,8 @@ namespace System.Management.Automation
                 yield break;
             }
 
-            AstParameterArgumentPair pathArgument;
             string pathParameterName = "Path";
-            if (!pseudoBinding.BoundArguments.TryGetValue(pathParameterName, out pathArgument))
+            if (!pseudoBinding.BoundArguments.TryGetValue(pathParameterName, out var pathArgument))
             {
                 pathParameterName = "LiteralPath";
                 pseudoBinding.BoundArguments.TryGetValue(pathParameterName, out pathArgument);
@@ -799,9 +798,9 @@ namespace System.Management.Automation
             // If you specified - Path, the list of OutputType can be refined, but we have to make a copy of the CmdletInfo object this way to get that refinement.
             var commandInfo = pseudoBinding.CommandInfo;
             var pathArgumentPair = pathArgument as AstPair;
-            if (pathArgumentPair?.Argument is StringConstantExpressionAst)
+            if (pathArgumentPair?.Argument is StringConstantExpressionAst ast)
             {
-                var pathValue = ((StringConstantExpressionAst)pathArgumentPair.Argument).Value;
+                var pathValue = ast.Value;
                 try
                 {
                     commandInfo = commandInfo.CreateGetCommandCopy(new object[] { "-" + pathParameterName, pathValue });
@@ -809,8 +808,7 @@ namespace System.Management.Automation
                 catch (InvalidOperationException) { }
             }
 
-            var cmdletInfo = commandInfo as CmdletInfo;
-            if (cmdletInfo != null)
+            if (commandInfo is CmdletInfo cmdletInfo)
             {
                 // Special cases
                 var inferTypesFromObjectCmdlets = InferTypesFromObjectCmdlets(commandAst, cmdletInfo, pseudoBinding).ToArray();
@@ -890,6 +888,14 @@ namespace System.Management.Automation
                     yield return foreachType;
                 }
             }
+
+            if (cmdletInfo.ImplementingType.FullName.EqualsOrdinalIgnoreCase("Microsoft.PowerShell.Commands.SelectObjectCommand"))
+            {
+                foreach (var foreachType in InferTypesFromSelectCommand(pseudoBinding, commandAst))
+                {
+                    yield return foreachType;
+                }
+            }
         }
 
         private static IEnumerable<PSTypeName> InferTypesFromCimCommand(PseudoBindingInfo pseudoBinding)
@@ -914,8 +920,7 @@ namespace System.Management.Automation
 
         private IEnumerable<PSTypeName> InferTypesFromForeachCommand(PseudoBindingInfo pseudoBinding, CommandAst commandAst)
         {
-            AstParameterArgumentPair argument;
-            if (pseudoBinding.BoundArguments.TryGetValue("MemberName", out argument))
+            if (pseudoBinding.BoundArguments.TryGetValue("MemberName", out AstParameterArgumentPair argument))
             {
                 var previousPipelineElement = GetPreviousPipelineCommand(commandAst);
                 if (previousPipelineElement == null)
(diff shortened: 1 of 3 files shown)


src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -157,9 +157,9 @@ public bool TryGetRepresentativeTypeNameFromExpressionSafeEval(ExpressionAst exp
             {
                 return false;
             }
-            object value;
+
             return expression != null &&
-                   SafeExprEvaluator.TrySafeEval(expression, ExecutionContext, out value) &&
+                   SafeExprEvaluator.TrySafeEval(expression, ExecutionContext, out var value) &&
```
iSazonov: The PR contains so many style changes that makes it unreasonably large.
Let's leave only significant changes - it will greatly facilitate and speed up the code viewing. Please move style changes in another PR.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -930,68 +918,91 @@ private IEnumerable<PSTypeName> InferTypesFromForeachCommand(PseudoBindingInfo p
...
+                }
+            }
+        }
+
+        private void InferTypesFromSelectCommand(PseudoBindingInfo pseudoBinding, CommandAst commandAst, List<PSTypeName> inferredTypes)
+        {
+            if (pseudoBinding.BoundArguments.TryGetValue("Property", out _)
+               || pseudoBinding.BoundArguments.TryGetValue("ExcludeProperty", out _))
+            {
+                return;
```
BrucePay: If the argument to `-Property` is a set of literal strings (most common case), then you could infer the "type" of the output object from them.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -777,18 +780,17 @@ object ICustomAstVisitor.VisitFileRedirection(FileRedirectionAst fileRedirection
             return TypeInferenceContext.EmptyPSTypeNameArray;
         }
 
-        private IEnumerable<PSTypeName> InferTypesFrom(CommandAst commandAst)
+        private void InferTypesFrom(CommandAst commandAst, List<PSTypeName> inferredTypes)
         {
```
BrucePay: Why the change from IEnumerable to List?

BrucePay (changes requested):
Over all it looks good. I'm still a bit curious about the change from `IEnumerable` -> `List`.  Also, i think the comments that talk about 'yield' need to be updated.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
iSazonov: I don't understand too why we add the allocations.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
powercode: Yes. I discussed this a bit with Jason. 
We don't currently have a good representation for this, but I would like to have that.
Same thing when you create a PSCustomObject from a hashtable.

But maybe something like "PSObject#Prop1/Prop2/Prop3"

It wouldn't say anything about the types of the properties, but that could maybe be encoded as well.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
powercode: Mostly debugability. It was a nightmare with all the nested state machines.

But since you have an allocation and virtual dispatch on every generated statemachine class that is generated when we use yield, my guess is that we actually have fewer allocations now.

And in general, I think the code got simpler. Several instances of loops that just yielded the loop object that just disappeared.

If you want to do the exercise, try debugging a tab completion before and after the change.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -414,8 +410,7 @@ private static bool IsMemberHidden(object member)
         private static bool IsConstructor(object member)
         {
             var psMethod = member as PSMethod;
-            var methodCacheEntry = psMethod?.adapterData as DotNetAdapter.MethodCacheEntry;
-            return methodCacheEntry != null && methodCacheEntry.methodInformationStructures[0].method.IsConstructor;
+            return psMethod?.adapterData is DotNetAdapter.MethodCacheEntry methodCacheEntry && methodCacheEntry.methodInformationStructures[0].method.IsConstructor;
```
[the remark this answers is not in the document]
powercode: Agree! A bit to fast there :)

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
iSazonov: Yes, debugging can be a headache here.

As for allocations my understanding is that it depends on how we access the object - if we use IEnumerable interface we allocate in both cases, if we cast to a collection we don't allocate I believe.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -350,8 +347,7 @@ private static bool TryGetRepresentativeTypeNameFromValue(object value, out PSTy
             type = null;
             if (value != null)
             {
-                var list = value as IList;
-                if (list != null && list.Count > 0)
+                if (value is IList list && list.Count > 0)
                 {
```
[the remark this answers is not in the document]
powercode: fixing.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
powercode: But there is no underlying collection to cast to. All the methods here that yield only has the generated state machine classes.

powercode:
I don't think the Travis build failure has anything to do with this PR.

BrucePay (approved):

iSazonov:
@powercode As for CodeFactor issues.
Many thanks that you fix ALL issues!!!

Really our policy is flexible:
- we don't ask fix all issues - only ones in new and changed code
- if you is _very busy_ you can ask maintainers to skip CodeFactor issues - we'll fix them ourselves later.
- if you do many changes it is better fix style issues in separate PR or add first separate commit with style changes.

Thanks again for your contribution!

iSazonov:
Restart CI Travis Linux.

anmenaga:
@TravisEz13 this looks ready; pls take a look when you can.

PR #7176, opened 2018-06-26
Fix tab expansion for `Get-Process` on macOS

### LINKED ISSUE #7174: No-argument tab-completion fails with Get-Process on macOS, possibly due to process objects lacking a ProcessName property
Steps to reproduce
------------------

On macOS, try the following:

```powershell
Get-Process  # < -- press <tab> here (whether once or multiple times makes no difference)
```

On Windows and Linux you get the names of all running processes, but on macOS nothing happens.

Expected behavior
-----------------

A  list (menu) of the names of all running processes.

Actual behavior
---------------

Nothing happens at all.

The problem goes away once you've typed at least prefix char. of the process name to match.

_Presumed_ cause:

On macOS, many processes output by `Get-Process` lack a `.Name` aka `.ProcessName` value.

As an aside: Is that expected or a bug in itself?

It seems that these entries without a `.Name` value cause the problem when the list of completions is constructed:

@lzybkr provided the following code to diagnose the problem, which simulates what happens when you tab-complete, while also surfacing exceptions:

```powershell
$line = "Get-Process "  # Simulate `Get-Process <tab>`
TabExpansion2 $line $line.Length
```

The above, rather than outputting a `[System.Management.Automation.CommandCompletion]` object with all  process names, reports the following error:

```
Exception calling "CompleteInput" with "3" argument(s): "Cannot process argument because the value of argument "listItemText" is null. Change the value of argument "listItemText"
...
```

Environment data
----------------

```powershell
PowerShell Core v6.1.0-preview.3 on macOS 10.13.5
```

SteveL-MSFT:
## PR Summary

macOS has a number of nameless processes.  When tab expansion gets called, the empty process names results in an exception which fails the entire completer.  Fix is to skip processes with empty name so that the named processes still show up in the result list.

Fix https://github.com/PowerShell/PowerShell/issues/7174

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
+++ b/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
@@ -3172,6 +3172,11 @@ namespace System.Management.Automation
                         completionText = quote + completionText + quote;
                     }
 
+                    if (string.IsNullOrEmpty(listItemText))
+                    {
+                        continue;
+                    }
+
                     result.Add(new CompletionResult(completionText, listItemText, CompletionResultType.ParameterValue, listItemText));
                 }
 
--- a/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
+++ b/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
@@ -3172,6 +3172,7 @@ namespace System.Management.Automation
                         completionText = quote + completionText + quote;
                     }
 
+                    // on macOS, system processes names will be empty if PowerShell isn't run as `sudo`
                     if (string.IsNullOrEmpty(listItemText))
                     {
                         continue;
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -3172,6 +3172,11 @@ private static void NativeCompletionProcessCommands(CompletionContext context, s
                         completionText = quote + completionText + quote;
                     }
 
+                    if (string.IsNullOrEmpty(listItemText))
+                    {
+                        continue;
```
iSazonov: If we now silently skip empty strings should we document this?

iSazonov (approved):

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
lzybkr: I'm surprised a process might not have a name - is this patching over a problem elsewhere, like in .Net?

At any rate, a comment on why this is expected would be useful.

SteveL-MSFT: @lzybkr if I run PowerShell with `sudo`, the process names are there.  So presumably these are system processes where some properties are protected.  I'll add a comment.

BrucePay (approved):
LGTM

adityapatwardhan (approved):

PR #7182, opened 2018-06-26
Revert "Make switch statement report correct error position when it fails to evaluate the condition"

daxian-dbw:
Reverts #7151

The fix causes a problem in debugging. When stepping over the scripts in debugging mode, the debugger will stop at the switch statement condition expression twice at the very beginning -- one for evaluating the condition, and one for iterating the condition.

Here is an example:
```
PS:4> .\about.ps1
Entering debug mode. Use h or ? for help.

Hit Line breakpoint on 'F:\tmp\about.ps1:1'

At F:\tmp\about.ps1:1 char:1
+ $Test = 1;
+ ~~~~~~~~~
PS:4>> v
At F:\tmp\about.ps1:2 char:1
+ $b = 1..10
+ ~~~~~~~~~~
PS:4>> v
At F:\tmp\about.ps1:3 char:9
+ switch ($b) {
+         ~~
PS:4>> v
At F:\tmp\about.ps1:3 char:9
+ switch ($b) {
+         ~~
PS:4>> v
At F:\tmp\about.ps1:4 char:5
+     "a" {};
+     ~~~
```


test/powershell/Language/Scripting/ErrorPosition.Tests.ps1
```diff
@@ -1,51 +0,0 @@
-# Copyright (c) Microsoft Corporation. All rights reserved.
```
rjmholt: The tests still seem relevant though -- they do cover an existing bug and after it's fixed would check for regressions. We could keep them with a `-Pending` flag and track it with an issue?

rjmholt (approved):

test/powershell/Language/Scripting/ErrorPosition.Tests.ps1
daxian-dbw: The tests will be added back along with the next attempt to fix the issue.

BrucePay (commented):
LGTM

Issue #6991, opened 2018-06-04
Respect CompatiblePSEditions field in module manifests, with sane default


rjmholt:
For Windows PowerShell/PowerShell Core modules to have a clear cross-compatibility story, we need PowerShell Core to respect the `CompatiblePSEditions` field.

This means, for PowerShell Core:
* `Get-Module -ListAvailable` lists only those modules on the module path that are Core compatible
* `Import-Module` will only import a module that is Core compatible and throw an error otherwise
* The absence of the `CompatiblePSEditions` field should imply a sane default. The current suggested default is `CompatiblePSEditions = @('Desktop')`, i.e. compatible with Windows PowerShell (since there is no such field in Windows PowerShell and modules written for it only would lack this field by default).

rjmholt:
See also: https://github.com/PowerShell/PowerShell/issues/6992

rjmholt:
Tagging @PowerShell/powershell-committee for discussion. I think we're also going to do an RFC for this, yes?

rjmholt:
We should also clarify the difference between `PSEdition` and `CompatiblePSEdition`

SteveL-MSFT:
@rjmholt yes, RFC to be published shortly

SteveL-MSFT:
https://github.com/PowerShell/PowerShell-RFC/pull/130

Issue #6990, opened 2018-06-04
Include Windows PowerShell System32 module directories on module path


rjmholt:
PowerShell Core 6 should try to use compatible Windows modules out of the box. Part of this is including those modules under the `\Windows\System32\` path on the module path to allow discovery and loading of those modules.

rjmholt:
See also:
* https://github.com/PowerShell/PowerShell/issues/6991
* https://github.com/PowerShell/PowerShell/issues/6992

rjmholt:
Tagging @PowerShell/powershell-committee for review

SteveL-MSFT:
See https://github.com/PowerShell/PowerShell-RFC/pull/130

PR #7183, opened 2018-06-26
Check CompatiblePSEditions manifest field for modules loaded from Windows PowerShell $PSHOME

### LINKED ISSUE #6992: Module discovery cmdlets should be able to override PSCompatibleEdition
Moving toward broader usage on Windows for PowerShell Core 6, we will need to honour the `PSCompatibleEditions` field of module manifests.

However, this field may be set conservatively or defaulted to a conservative value. As such we should allow a flag to allow the forced import of those modules for `Get-Module [-ListAvailable]` and `Import-Module`.

Ideally the name of this flag should be pithy, expressive about being a compatibility override behaviour, and dangerous-sounding so that users will understand that things may break as a result.

rjmholt:
## PR Summary

Resolves #6992, #6991, #6990. Also see https://github.com/PowerShell/PowerShell-RFC/pull/130.

- Add `%WINDIR%\System32\WindowsPowerShell\v1.0\Modules` (Windows PowerShell $PSHOME) to the end of the default PSCore 6 module path (i.e. the module path as initially set at startup).
- Cause an error to be thrown by `Import-Module` when a module with `CompatiblePSEditions` not containing `"Core"` is being loaded from the 'System32' module path.
- Suppress output of modules listed by `Get-Module -ListAvailable` from Windows PowerShell $PSHOME when `CompatiblePSEditions` does not contain `"Core"`.
- Introduce the `-SkipCompatibilityCheck` switch parameter on both `Import-Module` and `Get-Module` to respectively allow importing incompatible modules and listing incompatible modules.
- Adds a  `PSEdition` column to the `PSModuleInfo` table view format.
- Ensures that completions are not given for incompatible modules on the System32 module path.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**


test/powershell/Modules/Microsoft.PowerShell.Core/CompatiblePSEditions.Tests.ps1
```diff
@@ -0,0 +1,194 @@
...
+        AfterAll {
+            Pop-ModulePath
+        }
+
+        It "Successfully imports compatible modules from the module path with PSEdition <Edition>" -TestCases $successCases -Skip:(-not $IsWindows) {
+            param($Edition, $ModuleName, $Result)
+
+            Import-Module $ModuleName
+            [scriptblock]::Create("Test-$ModuleName").Invoke() | Should -Be $Result
+        }
```
BrucePay: Why not just do `& "Test-$ModuleName" | Should -Be $Result`

test/powershell/Modules/Microsoft.PowerShell.Core/CompatiblePSEditions.Tests.ps1
```diff
@@ -0,0 +1,194 @@
...
+
+        It "Fails to import incompatible modules from the module path with PSEdition <Edition>" -TestCases $failCases -Skip:(-not $IsWindows) {
+            param($Edition, $ModuleName, $Result)
+
+            [scriptblock]::Create("Import-Module $ModuleName -ErrorAction 'Stop'; Test-$ModuleName") | Should -Throw -ErrorId "Modules_PSEditionNotSupported,Microsoft.PowerShell.Commands.ImportModuleCommand"
```
BrucePay: Why not just do `{Import-Module $ModuleName -ErrorAction Stop;  & "Test-$ModuleName"}  | Should -Throw -ErrorId "Modules_PSEditionNotSupported,Microsoft.PowerShell.Commands.ImportModuleCommand"`

test/powershell/Modules/Microsoft.PowerShell.Core/CompatiblePSEditions.Tests.ps1
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        It "Imports an incompatible module from the module path with -SkipCompatibilityCheck with PSEdition <Edition>" -TestCases ($successCases + $failCases) -Skip:(-not $IsWindows) {
+            param($Edition, $ModuleName, $Result)
+
+            Import-Module $ModuleName -SkipCompatibilityCheck
+            [scriptblock]::Create("Test-$ModuleName").Invoke() | Should -Be $Result
+        }
```
BrucePay: Again `& "Test-$ModuleName" | Should -Be $Result`

BrucePay (changes requested):
I think you should fix the scriptblock creates in the tests. (Unless I'm missing something and there's a real reason for doing it that way.)

src/System.Management.Automation/engine/Modules/GetModuleCommand.cs
```diff
@@ -84,6 +84,16 @@ public sealed class GetModuleCommand : ModuleCmdletBase, IDisposable
         [ArgumentCompleter(typeof(PSEditionArgumentCompleter))]
         public string PSEdition { get; set; }
 
+        /// <summary>
+        /// When set, CompatiblePSEditions checking is disabled for modules in the System32 (Windows PowerShell) module directory.
+        /// </summary>
+        [Parameter]
+        public SwitchParameter SkipCompatibilityCheck
+        {
```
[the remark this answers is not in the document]
rjmholt: I agree with you here. My concern is that `SkipCompatibilityCheck` is [what is given in the RFC](https://github.com/SteveL-MSFT/PowerShell-RFC/blob/72ecfdaff97c59cd5eac3c01d4762193bf6c2f9b/2-Draft-Accepted/RFCNNNN-PSCore6-and-Windows-Modules.md#using-modules-from-system32-not-declared-as-compatible).

test/powershell/Modules/Microsoft.PowerShell.Core/CompatiblePSEditions.Tests.ps1
rjmholt: Ah much nicer!

src/System.Management.Automation/engine/Modules/GetModuleCommand.cs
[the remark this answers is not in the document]
SteveL-MSFT: I think it's fine to change it to `SkipEditionCheck` as it's more specific

src/System.Management.Automation/engine/Modules/GetModuleCommand.cs
```diff
@@ -341,6 +351,22 @@ protected override void ProcessRecord()
                 ThrowTerminatingError(error);
             }
 
+            // SkipCompatibilityCheck should not work with PSEdition specified
```
SteveL-MSFT: It seems that the two are not mutually exclusive.  `-PSEdition` is for filtering while `-SkipCompatibilityCheck` is for enforcement.

test/powershell/Modules/Microsoft.PowerShell.Core/assets/CompatiblePSEditionsModules/BothModule/BothModule.psd1
```diff
@@ -0,0 +1,122 @@
+#
```
SteveL-MSFT: These test modules are pretty basic.  I wonder if it's better to just generate these at runtime as test variations?

test/powershell/Modules/Microsoft.PowerShell.Core/CompatiblePSEditions.Tests.ps1
```diff
@@ -0,0 +1,194 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+$script:modulePathStack = [System.Collections.Generic.Stack[string]]::new()
+function Push-ModulePath
```
SteveL-MSFT: Seems overkill since it doesn't appear the tests need to push more than once before a pop.  Existing tests simply store `currentPSModulePath` in `Before` and restore in `After`

SteveL-MSFT (changes requested):

src/System.Management.Automation/resources/Modules.resx
```diff
@@ -606,4 +606,7 @@
   <data name="PrerequisiteForDesktopEditionOnly" xml:space="preserve">
     <value>This prerequisite is valid for the PowerShell Desktop edition only.</value>
   </data>
+  <data name="PSEditionNotSupported" xml:space="preserve">
+    <value>Module '{0}' does not support current PowerShell edition '{1}'. Its supported editions are '{2}'. Use 'Import-Module -SkipCompatibilityCheck' to ignore the compatibility of this module.</value>
```
iSazonov: Should we use imperative mood in ` Use 'Import-Module -SkipCompatibilityCheck' `? 
Also `{0}` can be long path.

SteveL-MSFT (changes requested):
The formatter for ModuleInfo needs to be updated to expose PSEditions column.

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
```diff
@@ -984,6 +1001,14 @@ internal static string SetModulePath()
 
             if (!string.IsNullOrEmpty(newModulePathString))
             {
+
+// If on Windows, we want to add the System32
+// Windows PowerShell module directory
+// so that Windows modules are discovered
+#if !UNIX
+                newModulePathString += Path.PathSeparator + GetWindowsPowerShellPSHomeModulePath();
```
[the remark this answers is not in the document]
rjmholt: This is to do `"C:\path\one;C:\path\two" + "C:\path\three"`; we're not combining paths into one path, but concatenating multiple paths into the PSModulePath using a path separator (`';'`). Is there a method for that?

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -2482,13 +2531,18 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
...
                 }
             }
+
             if (compatiblePSEditions != null)
             {
-                foreach (var psEdition in compatiblePSEditions)
-                {
-                    manifestInfo.AddToCompatiblePSEditions(psEdition);
-                }
+                manifestInfo.AddToCompatiblePSEditions(compatiblePSEditions);
```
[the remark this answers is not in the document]
rjmholt: My thinking is that the `PSModuleInfo` should reflect the manifest. But I think this particular point needs further discussion, because rewriting default logic everywhere is error prone. I'm reluctant to add another public field on to the `PSModuleInfo` (and that would probably merit an RFC anyway).

But so far, the way it works is:
* `PSModuleInfo.CompatiblePSEditions` is exactly what's in the manifest (or an empty array)
* Everywhere we check `CompatiblePSEditions` a default is used if it's null or empty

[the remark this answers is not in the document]
iSazonov: I meant that if `compatiblePSEditions == null` we should add default.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -2399,7 +2399,14 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
 
                 if (bailOnFirstError)
                 {
-                    // If we return null here, a fake module info will be created. Since
+                    // If we're trying to load the module, return null so that caches
+                    // are not polluted
+                    if ((manifestProcessingFlags & ManifestProcessingFlags.LoadElements) != 0)
```
iSazonov: We should use `HasFlag()` - now it is optimized.

src/System.Management.Automation/FormatAndOutput/DefaultFormatters/PowerShellCore_format_ps1xml.cs
```diff
@@ -1192,11 +1192,24 @@ private static IEnumerable<FormatViewDefinition> ViewsOf_ModuleInfoGrouping(Cust
                     .AddHeader(Alignment.Left, width: 10)
                     .AddHeader(Alignment.Left, width: 10)
                     .AddHeader(Alignment.Left, width: 35)
+                    .AddHeader(Alignment.Left, width: 10, label: "PSEdition")
```
iSazonov: Maybe 9?

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -3458,6 +3454,12 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
             }
 
             isOnSystem32ModulePath = true;
+
+            if (BaseSkipEditionCheck)
+            {
+                return true;
+            }
+
```
iSazonov: I wonder that we move the block - have we I side effect in lines 3450-3451?

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -3455,12 +3455,7 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
 
             isOnSystem32ModulePath = true;
 
-            if (BaseSkipEditionCheck)
-            {
-                return true;
-            }
-
-            return Utils.IsPSEditionSupported(compatiblePSEditions);
+            return BaseSkipEditionCheck || Utils.IsPSEditionSupported(compatiblePSEditions);
```
iSazonov: I wonder that we move the block - have we I side effect in lines 3450-3451?

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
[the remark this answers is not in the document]
rjmholt: Yeah, I'm thinking that we shouldn't.

If we do, then we are not reflecting what the manifest says. So there will be no way to differentiate between a manifest that has `CompatiblePSEditions = @("Desktop")` and no `CompatiblePSEditions` field at all -- all our tools to parse a manifest will incorrectly see them as the same.

For the purposes of this logic, yes we assume a default. But this is quite specific logic. So I think we need to preserve the information for other functionalities that may need to distinguish the scenarios.

But maybe we also need a way to record the default assumed value on the `PSModuleInfo`?

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -111,14 +111,14 @@ Describe "Import-Module for Binary Modules in GAC" -Tags 'CI' {
 
         It "Load PSScheduledJob from Windows Powershell Modules folder should fail" -Skip:(-not $IsWindows) {
             $modulePath = Join-Path $env:windir "System32/WindowsPowershell/v1.0/Modules/PSScheduledJob"
-            { Import-Module $modulePath -ErrorAction SilentlyContinue } | Should -Throw -ErrorId 'FormatXmlUpdateException,Microsoft.PowerShell.Commands.ImportModuleCommand'
+            { Import-Module $modulePath -SkipEditionCheck -ErrorAction SilentlyContinue } | Should -Throw -ErrorId 'FormatXmlUpdateException,Microsoft.PowerShell.Commands.ImportModuleCommand'
```
rjmholt: AppVeyor is failing this test, but it's passing on my machine. Any ideas?

iSazonov: Why we use `SilentlyContinue ` not `Stop`?

src/System.Management.Automation/engine/Modules/GetModuleCommand.cs
```diff
@@ -489,50 +489,55 @@ private void GetLoadedModules(string[] names, IDictionary<string, ModuleSpecific
...
-        /// directory that do not explicitly support the current PowerShell edition.
-        /// </returns>
-        private IEnumerable<PSModuleInfo> FilterModulesForEdition(IEnumerable<PSModuleInfo> modules)
+        /// <param name="modules">The modules to filter incompatible examples from.</param>
+        /// <returns>All modules that are either not on the checked path or are compatible with the current PowerShell edition.</returns>
+        private IEnumerable<PSModuleInfo> FilterModulesForIncompatibleOnEditionCheckedPath(IEnumerable<PSModuleInfo> modules)
         {
-            // If no PSEdition is given, we only need to check the System32 path inclusions
-            if (String.IsNullOrEmpty(PSEdition))
+            Dbg.Assert(!SkipEditionCheck, $"Caller to verify that {nameof(SkipEditionCheck)} is false");
```
iSazonov: What is benefits from moving the check on up level?

src/System.Management.Automation/engine/Modules/GetModuleCommand.cs
```diff
@@ -489,50 +489,55 @@ private void GetLoadedModules(string[] names, IDictionary<string, ModuleSpecific
...
-                yield break;
             }
+        }
+
+        /// <summary>
+        /// Filter an enumeration of modules based on what PowerShell editions they are compatible with.
+        /// If the PSEdition parameter has been given, filter modules based on that.
+        /// </summary>
+        /// <param name="modules">The modules to filter by edition.</param>
+        /// <returns> All modules meeting the PSEdition constraint.</returns>
```
iSazonov: Please remove first space before `All`.

src/System.Management.Automation/engine/Modules/GetModuleCommand.cs
```diff
@@ -489,50 +489,55 @@ private void GetLoadedModules(string[] names, IDictionary<string, ModuleSpecific
...
+        private IEnumerable<PSModuleInfo> FilterModulesForEdition(IEnumerable<PSModuleInfo> modules)
+        {
+            Dbg.Assert(!String.IsNullOrEmpty(PSEdition), $"Caller to verify that {nameof(PSEdition)} is not null or empty");
```
iSazonov: The same. What is benefits from moving the check on up level?

rjmholt: The callee assertion seems to be a common occurrence in the codebase -- I could have included a comment, but thought an assertion was better as "zero-cost executable documentation".

Putting the check inside the method means needing to have a separate `for`-loop to handle the unchecked enumeration (since we can't use both `return` and `yield return` in the same method). Which doubles the size of the method and I think makes it less readable too. Doing it like this, we can avoid the whole method call ahead of time and keep the method itself cohesive.

iSazonov: `return` -> `yield break`?

> we can avoid the whole method call 

I think a performance is not so important here.

rjmholt: Well `yield break` will return an empty enumeration. But if `PSEdition` is not set you don't want that, you want everything. So you need another `for`-loop just for the case where it's not set, which seems pretty ugly to me compared to just opting out of the filter.

Yes I very much agree that the performance gain is minimal and not needed. Just pointing out a separate virtue of the approach I took.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
rjmholt: Oh didn't see this before! Good idea!

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -2361,6 +2371,53 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
                 if (bailOnFirstError) return null;
             }
 
+// Add PSEdition check for Windows PowerShell
+// modules imported from the System32 path
+#if !UNIX
```
daxian-dbw: I think the `# if UNIX` part can be abstracted in `IsPSEditionCompatible`, and then code here don't need to be divided up.
The `IsPSEditionCompatible` would look like this:
```powershell
        private bool IsPSEditionCompatible(
            string moduleManifestPath,
            IEnumerable<string> compatiblePSEditions,
            out bool isOnSystem32ModulePath)
        {
            isOnSystem32ModulePath = false;
#if UNIX
            return false;
#else
            string windowsPSModulePath = ModuleIntrinsics.GetWindowsPowerShellPSHomeModulePath();
            if (!moduleManifestPath.StartsWith(windowsPSModulePath, StringComparison.OrdinalIgnoreCase))
            {
                return true;
            }

            isOnSystem32ModulePath = true;

            return BaseSkipEditionCheck || Utils.IsPSEditionSupported(compatiblePSEditions);
#endif
        }
```

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
rjmholt: The `if` block? My goal is for the `PSModuleInfo` to accurately reflect the module manifest being loaded. It seems to be working for our purposes at present -- I moved the block down because I realised `isOnSystem32ModulePath` wasn't always set when a module was actually found there.

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
```diff
@@ -600,6 +600,23 @@ private static string GetSharedModulePath()
...
+        /// </summary>
+        /// <returns>The path of the Windows PowerShell system module directory.</returns>
+        internal static string GetWindowsPowerShellPSHomeModulePath()
+        {
+            if (!String.IsNullOrEmpty(InternalTestHooks.TestWindowsPowerShellPSHomeLocation))
+            {
+                return InternalTestHooks.TestWindowsPowerShellPSHomeLocation;
+            }
+
+            return Path.Combine(System.Environment.SystemDirectory, "WindowsPowerShell", "v1.0", "Modules");
```
daxian-dbw: It's better to have a static field keeping the value, so we only need to calculate it once.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -399,11 +406,48 @@ internal static IEnumerable<CommandInfo> GetMatchingCommands(string pattern, Exe
...
+                        if (s_incompatibleEditionSystem32Modules.ContainsKey(modulePath))
+                        {
+                            // We have already identified the module as incompatible
+                            continue;
+                        }
+                        else
+                        {
+                            using (PowerShell pwsh = PowerShell.Create())
+                            {
+                                PSModuleInfo module = pwsh.AddCommand("Get-Module")
```
rjmholt: I quite dislike this -- not sure if there's a better way to do it?

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
rjmholt: That's what I wanted to do. I assume the environment variable can't change then?

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
rjmholt: Yes I thought I'd start with the "zero runtime overhead on UNIX" approach and see how everyone felt. I definitely think your way is cleaner!

src/System.Management.Automation/engine/Modules/GetModuleCommand.cs
```diff
@@ -482,14 +472,117 @@ private void GetLoadedModules(string[] names, IDictionary<string, ModuleSpecific
...
+            foreach (PSModuleInfo module in modules)
+            {
+                if (!module.IsLoadedFromCompatibilityCheckedPath)
+                {
+                    yield return module;
+                    continue;
+                }
+
+                IEnumerable<string> moduleCompatibleEditions = module.CompatiblePSEditions.Any() ? module.CompatiblePSEditions : DefaultCompatiblePSEditions;
+                if (moduleCompatibleEditions.Contains(currentPSEdition, StringComparer.OrdinalIgnoreCase))
```
rjmholt: Going to change this over to the `Utils` method and optimise that one slightly. Want to let @daxian-dbw finish his review first (GitHub has had a nasty habit of forcing me to rewrite comments when new commits are added recently...)

src/System.Management.Automation/engine/Modules/GetModuleCommand.cs
```diff
@@ -465,14 +459,10 @@ private void GetAvailableViaPsrpSession(string[] names, IDictionary<string, Modu
...
         private void GetAvailableLocallyModules(string[] names, IDictionary<string, ModuleSpecification> moduleSpecTable, bool all)
         {
-            var refresh = Refresh.IsPresent;
-            var modules = GetModule(names, all, refresh);
-
-            foreach (
-                var psModule in
-                    modules.Where(module => ModuleMatch(module, moduleSpecTable, PSEdition)).Select(module => new PSObject(module))
-                )
+            IEnumerable<PSModuleInfo> modules = GetModule(names, all, Refresh);
```
iSazonov: Why we use `IEnumerable`? We could use List<T> - seems it is more thriftily.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
iSazonov: I see we have `private IEnumerable<PSModuleInfo> GetModuleForRootedPaths(string[] modulePaths, bool all, bool refresh)`

test/powershell/Modules/Microsoft.PowerShell.Core/CompatiblePSEditions.Tests.ps1
```diff
@@ -0,0 +1,247 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
```
iSazonov: The file name doesn't contain `Module`. Maybe `ModuleCompatiblePSEditions.Tests.ps1`?

iSazonov (commented):
Leave a comment

src/System.Management.Automation/engine/Modules/GetModuleCommand.cs
rjmholt: Used `IEnumerable` here because:
* The interface type defines all the functionality we need without added constraints
* We don't require indexing or even size checking, just ordered enumeration
* The method is free to implement the collection as it wishes

*But*, perhaps there are benefits to using a `List` or at least an `IList`?
* Less type complexity (and binding to a concrete type not an issue with an internal API)?
* Provides strictness and ordering guarantees

rjmholt:
@iSazonov I couldn't reply to [one of your comments](https://github.com/PowerShell/PowerShell/pull/7183#discussion_r199331481) for some reason.

The `GetModuleForRootedPaths(string[] modulePaths, bool all, bool refresh)` method is a private member method of `ModuleCmdletBase` I think, so to use it from a static method in `ModuleUtils` we would need to find a runspace, instantiate the cmdlet object, get access to the method and fire it off.

After thinking about that, I thought that `PowerShell.Create()...Invoke()` seemed to do most of that. So I've done that here.

But maybe there's a more direct way to instantiate a cmdlet in a runspace and fire a method?

src/System.Management.Automation/engine/Modules/ModuleIntrinsics.cs
daxian-dbw: `System.Environment.SystemDirectory` doesn't use the environment variable. It calls the win32 API `GetSystemDirectoryW`. See https://msdn.microsoft.com/en-us/library/windows/desktop/ms724373(v=vs.85).aspx

rjmholt:
So the `Import-Module for Binary Modules in GAC` > `Modules are not loaded from GAC` test failure is something I can't reproduce. I notice the test is not idempotent; it will succeed the first time, and then fail on reruns after that.

After stepping through it all with the debugger, I can't find any difference in the behaviour...

daxian-dbw:
We can put off the rebase after the PR has been signed off.

src/System.Management.Automation/FormatAndOutput/DefaultFormatters/PowerShellCore_format_ps1xml.cs
```diff
@@ -1192,11 +1192,24 @@ private static IEnumerable<FormatViewDefinition> ViewsOf_ModuleInfoGrouping(Cust
...
+                        .AddScriptBlockColumn(@"
+                            $result = [System.Collections.ArrayList]::new()
+                            $editions = $_.CompatiblePSEditions
+                            if (-not $editions)
+                            {
+                                $editions = @('Desktop')
+                            }
+                            foreach ($edition in $editions)
+                            {
+                                $result += $edition.Substring(0,1).ToUpper()
```
SteveL-MSFT: Since the column is 9 characters wide, I wonder if using `Desk` and `Core` may be better than just `D` and `C`

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -256,6 +261,11 @@ internal List<WildcardPattern> MatchAll
             "ModuleVersion"
         };
 
+        internal static IReadOnlyList<string> DefaultCompatiblePSEditions { get; } = new string[]
```
SteveL-MSFT: This is the default if CompatiblePSEditions isn't declared in the module manifest, correct?  Should have a comment to clarify this as someone unfamiliar with the intent may thing this is what PSCore6 is by default compatible with.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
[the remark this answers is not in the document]
SteveL-MSFT: I agree with @rjmholt that the member should reflect what is actually in the manifest.

test/powershell/Modules/Microsoft.PowerShell.Core/CompatiblePSEditions.Tests.ps1
```diff
@@ -0,0 +1,247 @@
...
+        $basePath = Join-Path $TestDrive "EditionCompatibleModules"
+        New-TestModules -TestCases $successCases -BaseDir $basePath
+        New-TestModules -TestCases $failCases -BaseDir $basePath
+
+        # Emulate the System32 module path for tests
+        [System.Management.Automation.Internal.InternalTestHooks]::SetTestHook("TestWindowsPowerShellPSHomeLocation", $basePath)
+    }
+
+    AfterAll {
+        Remove-Item -Force -Recurse $basePath
```
SteveL-MSFT: You don't have to remove the modules as Pester will automatically clean out the `$TestDrive`.

SteveL-MSFT (changes requested):

rjmholt:
Ok, I believe I've fixed the test failure, but want to document it here.

I needed to rename my (CompatiblePSEditions) test file...

The problem is that the [test for disabling GAC loading](https://github.com/PowerShell/PowerShell/blob/3c079b9d42b5484bdfa6577ac5406bce05f8af02/test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1#L102-L116) will fail if you've already loaded the type it's supposed to fail on, because it caches it. You can verify this easily with:

```powershell
> Invoke-Pester ./test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
# All succeed
> Invoke-Pester ./test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
# GAC disable test fails
```

My tests came before this test, pre-caching types from the System32 module path, causing the load to succeed and thus this test to fail. Renaming the test file to @iSazonov's suggested name of `ModuleCompatiblePSEditions.Tests.ps1` fixed it because that gets executed after `Import-Module.Tests.ps1`.

I've included a note in my test file.

rjmholt:
Also, it looks like AppVeyor won't run the tests until I resolve the merge conflict.

But I think if I use GitHub's merge resolution, I won't be able to squash my commits properly?

iSazonov:
The PR was started as WIP and we already have over 67 comments - GitHub becomes very slow.
So I suggest close the PR, rebase, resolve merge conflits, fix style issues, squash commits and open new PR. add open questions to the new PR description.

> Renaming the test file to @iSazonov's suggested name of ModuleCompatiblePSEditions.Tests.ps1 fixed it because that gets executed after Import-Module.Tests.ps1.

Usually we create separate PowerShell instance to pass such tests without side effects.

iSazonov:
Sample to invoke cmdlet:

https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs#L4077-L4085

iSazonov:
Perhaps we could convert `GetModuleForRootedPaths()` to static.

iSazonov:
Seems the idea to cache available and compatible modules was good - this can extremely speed up completions.

rjmholt:
@iSazonov agreed on the caching

rjmholt:
It looks like that same test is still failing despite all my attempts to make it work -- assuming a `Get-Module -ListAvailable` is being fired elsewhere to load the PSScheduledJob assembly.

I'm pretty certain at this point that I've fixed the regressions, so I'm going to try and spin that test out into its own process...

BrucePay (approved):

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -2436,13 +2496,18 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
...
-                    manifestInfo.AddToCompatiblePSEditions(psEdition);
-                }
+                manifestInfo.AddToCompatiblePSEditions(compatiblePSEditions);
             }
+
+// Modules loaded from the System32 module path may
+// not have the CompatiblePSEditions fields ignored
+#if !UNIX
+            manifestInfo.IsLoadedFromCompatibilityCheckedPath = isOnSystem32ModulePath;
+#endif
```
daxian-dbw: I think it's not needed to put this in `if/def` here. On Unix plat, `isOnSystem32ModulePath` will always be false, and it's fine to set it here to keep a single code path.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -3010,6 +3075,12 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
                 newManifestInfo.IconUri = manifestInfo.IconUri;
                 newManifestInfo.RepositorySourceLocation = manifestInfo.RepositorySourceLocation;
 
+// On Windows, we need to copy over the field indicating
+// whether the module was imported from the System32 module path
+#if !UNIX
+                newManifestInfo.IsLoadedFromCompatibilityCheckedPath = manifestInfo.IsLoadedFromCompatibilityCheckedPath;
+#endif
```
daxian-dbw: Same here.

src/System.Management.Automation/engine/Modules/GetModuleCommand.cs
```diff
@@ -482,14 +472,120 @@ private void GetLoadedModules(string[] names, IDictionary<string, ModuleSpecific
...
+        /// </summary>
+        /// <param name="modules">The modules to filter through.</param>
+        /// <param name="moduleSpecificationTable">Module constraints, keyed by module name, to filter modules of that name by.</param>
+        /// <returns>All modules from the original input that meet both any module edition and module specification constraints provided.</returns>
+        private IEnumerable<PSModuleInfo> FilterModulesForEditionAndSpecification(
+            IEnumerable<PSModuleInfo> modules,
+            IDictionary<string, ModuleSpecification> moduleSpecificationTable)
+        {
+// Edition check only applies to Windows
+// System32 module path
```
daxian-dbw: It's better to put the comment right above `if (!SkipEditionCheck)`, like
```
     // Edition check only applies to Windows System32 module path
     if (!SkipEditionCheck)
```

rjmholt:
@SteveL-MSFT, @daxian-dbw this is now passing CI

src/System.Management.Automation/engine/Modules/GetModuleCommand.cs
rjmholt: I included this comment style sort of as a suggestion. We have a number of `#if`s in the code where it's not clear why the behaviour is version-specific. You're very right that in this case it's not needed, but in a number of the FullCLR/CoreCLR and UNIX/Windows cases, as time goes on we'll probably want to remove those conditionals -- the problem being that we'll have to spend half a day understanding the historical context of why they went in in the first place. But I suppose it's a bit late now 😄

iSazonov: Year ago we removed tons `#if`s after migrating to .Net Core 2.0.
> where it's not clear why the behaviour is version-specific

That's why we did not delete everything :smile:
In fact, we left only what we want to port ( since .Net Core 2.1, 3.0, 4.0, ... :-) )

rjmholt: Yeah that's what I'm thinking — it would be much easier to clear out the platform-specific code as .NET Core matures with documentation devoted to why we made it platform-specific in the first place!

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -99,26 +99,23 @@ Describe "Import-Module with ScriptsToProcess" -Tags "CI" {
...
-
         It "Load PSScheduledJob from Windows Powershell Modules folder should fail" -Skip:(-not $IsWindows) {
-            $modulePath = Join-Path $env:windir "System32/WindowsPowershell/v1.0/Modules/PSScheduledJob"
-            { Import-Module $modulePath -ErrorAction SilentlyContinue } | Should -Throw -ErrorId 'FormatXmlUpdateException,Microsoft.PowerShell.Commands.ImportModuleCommand'
+            $testScript = {
+                [System.Management.Automation.Internal.InternalTestHooks]::SetTestHook('DisableGACLoading', $true)
+                $modulePath = Join-Path $env:windir "System32/WindowsPowershell/v1.0/Modules/PSScheduledJob"
+                Import-Module $modulePath -SkipEditionCheck
+            }
+            $err = (& "$PSHOME\pwsh.exe" -c $testScript 2>&1 | Where-Object { $_ -is [System.Management.Automation.ErrorRecord] })[0]
```
SteveL-MSFT: You should add a comment why you are using a new PS process

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
rjmholt: Done

rjmholt:
I've just realised there's a problem when the Windows PowerShell module dir comes before the PSCore native $PSHOME on the module path -- you lose all the core (except Microsoft.PowerShell.Core) cmdlets...

Going to try and fix it today

src/System.Management.Automation/engine/Modules/TestModuleManifestCommand.cs
```diff
@@ -23,6 +23,14 @@ namespace Microsoft.PowerShell.Commands
     [OutputType(typeof(PSModuleInfo))]
     public sealed class TestModuleManifestCommand : ModuleCmdletBase
     {
+        /// <summary>
+        /// Creates an instance of the Test-ModuleManifest command
+        /// </summary>
+        public TestModuleManifestCommand()
+        {
+            BaseSkipEditionCheck = true;
```
daxian-dbw: For my education, can you please explain why this is needed?

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -2315,6 +2330,51 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
...
+                    // If we're trying to load the module, return null so that caches
+                    // are not polluted
+                    if (manifestProcessingFlags.HasFlag(ManifestProcessingFlags.LoadElements))
+                    {
+                        return null;
+                    }
+
+                    // If we return null with Get-Module, a fake module info will be created. Since
+                    // we want to suppress output of the module, we need to do that here.
+                    return new PSModuleInfo(moduleManifestPath, context: null, sessionState: null)
```
daxian-dbw: Why do we need to return a dummy module info object in case of `Get-Module`? If it's not compatible, `Get-Module` will not return it anyway, seems no need to return a dummy module info.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/Modules/TestModuleManifestCommand.cs
rjmholt: This calls `ModuleCmdletBase.LoadModuleManifest()` and throws if `BaseSkipEditionCheck` is not set.

On the basis that `Test-ModuleManifest` should never reject a manifest based on incompatibility (only manifest invalidity), I just set it to true.

However, if `Test-ModuleManifest` should have a more sophisticated functionality, let me know I and I'll change this accordingly.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
rjmholt: The calling method in the context of `Get-Module` will take a `null` result and decide to make a dummy `PSModuleInfo` object that doesn't say "I'm incompatible", so `Get-Module` will show it when it fails the check.

So this creates the fake `PSModuleInfo` earlier with the `IsLoadedFromCompatibilityCheckedPath` set.

But it's not a great piece of code, just the cheapest fix at the moment.

src/System.Management.Automation/engine/Modules/TestModuleManifestCommand.cs
daxian-dbw: OK, can you please add comment here explain why it's set to true here?

src/System.Management.Automation/engine/Modules/GetModuleCommand.cs
```diff
@@ -341,6 +353,18 @@ protected override void ProcessRecord()
                 ThrowTerminatingError(error);
             }
 
+            // -SkipEditionCheck only makes sense for -ListAvailable (otherwise the module is already loaded)
+            if (SkipEditionCheck && !ListAvailable)
+            {
+                ErrorRecord error = new ErrorRecord(
+                    new InvalidOperationException("-SkipEditionCheck can only be used with -ListAvailable"),
```
daxian-dbw: Exception message should be put in resource files.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -2376,6 +2391,52 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
...
+                        moduleManifestPath);
+
+                    WriteError(er);
+                }
+
+                if (bailOnFirstError)
+                {
+                    // If we're trying to load the module, return null so that caches
+                    // are not polluted
+                    if (manifestProcessingFlags.HasFlag(ManifestProcessingFlags.LoadElements))
```
daxian-dbw: I believe you can use `importingModule` here, instead of checking the flag again.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -66,9 +69,11 @@ internal static bool IsPossibleModuleDirectory(string dir)
         /// Get all module files by searching the given directory recursively.
         /// All sub-directories that could be a module folder will be searched.
         /// </summary>
-        internal static IEnumerable<string> GetAllAvailableModuleFiles(string topDirectoryToCheck)
+        internal static IEnumerable<string> GetAllAvailableModuleFiles(string topDirectoryToCheck, bool skipEditionCheck)
```
daxian-dbw: Let's defer solving the `Get-Module -List -All` problem. We are not sure what is the right user experience yet. We can solve it as a bug fix.

src/System.Management.Automation/engine/Modules/PSModuleInfo.cs
```diff
@@ -834,6 +834,19 @@ internal void AddToCompatiblePSEditions(string psEdition)
...
+            _compatiblePSEditions.AddRange(psEditions);
+        }
+
+        /// <summary>
+        /// Describes whether the module was considered compatible at load time.
+        /// Any module not on the System32 module path should have this as true.
+        /// Modules loaded from the System32 module path will have this as true if they
+        /// have declared edition compatibility with PowerShell Core.
+        /// </summary>
+        internal bool IsConsideredEditionCompatible { get; set; } = true;
```
daxian-dbw: With this default value, directly importing a non-psd1 module from the system32 module path will succeed.
This might not be the desired behavior, but let's defer solving that problem.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -3382,6 +3446,40 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
...
+        /// </summary>
+        /// <param name="moduleManifestPath">The path to the module manifest being checked.</param>
+        /// <param name="compatiblePSEditions">The value of the CompatiblePSEditions field of the module manifest.</param>
+        /// <param name="moduleDeclaresCoreCompatible">
+        /// True if the module explicitly declares compatiblity for the current PSEdition in its manifest, otherwise false.</param>
+        /// <param name="isOnSystem32ModulePath">
+        /// True if the module is on the system32 module path (where edition compatibility is checked), false otherwise.
+        /// </param>
+        /// <returns>True if the module is compatible with the running PowerShell edition, false otherwise.</returns>
+        private bool IsPSEditionCompatible(
```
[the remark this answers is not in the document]
daxian-dbw: Possibly we can use the same method for `AnalysisCache.ModuleIsEditionIncompatible` by moving this method to `ModuleUtils`

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
daxian-dbw: Discussed with @rjmholt offline on this. The current behavior for `Get-Module -ListAvailable` is that if we have a problem parsing the manifest, we just create a dummy `PSModuleInfo` using the file path, and still display that module to the user.

This behavior doesn't work with the new behavior where we want to hide modules that are incompatible -- say we got parsing error for a module manifest under system 32, should we return a dummy module for that file or not? We got a parsing error, so we don't know if that module has `Core` declared or not.

A more suitable behavior would be don't return a dummy module in case we got a parsing error, because practically, that module is not available since it cannot be imported. We defer solving this along with the other two issues described in my comments.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -2101,17 +2107,19 @@ internal static string ConcatenateStringPathArguments(CommandElementAst stringAs
...
                         bool loadedModulesOnly = boundArguments == null || !boundArguments.ContainsKey("ListAvailable");
-                        NativeCompletionModuleCommands(context, parameterName, loadedModulesOnly, /* isImportModule: */ false, result);
+                        bool skipEditionCheck = !loadedModulesOnly && boundArguments.ContainsKey("SkipEditionCheck");
+                        NativeCompletionModuleCommands(context, parameterName, result, loadedModulesOnly, skipEditionCheck: skipEditionCheck);
                         break;
                     }
                 case "Remove-Module":
                     {
-                        NativeCompletionModuleCommands(context, parameterName, /* loadedModulesOnly: */ true, /* isImportModule: */ false, result);
+                        NativeCompletionModuleCommands(context, parameterName, result, loadedModulesOnly: true);
```
iSazonov: Named parameters look more readable (what is "result" here?).

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -2101,17 +2107,19 @@ internal static string ConcatenateStringPathArguments(CommandElementAst stringAs
...
                         break;
                     }
                 case "Import-Module":
                     {
-                        NativeCompletionModuleCommands(context, parameterName, /* loadedModulesOnly: */ false, /* isImportModule: */ true, result);
+                        bool skipEditionCheck = boundArguments != null && boundArguments.ContainsKey("SkipEditionCheck");
+                        NativeCompletionModuleCommands(context, parameterName, result, isImportModule: true, skipEditionCheck: skipEditionCheck);
```
iSazonov: Named parameters look more readable (what is "result" here?).

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
iSazonov: Please open tracking and reference issues.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
iSazonov: Please open tracking and reference issue.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
rjmholt: Very much agree with you here. Sadly the parameter name is also `result`.

rjmholt:
@daxian-dbw I think this should be ready to merge now

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -2475,6 +2536,10 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
             manifestInfo.ProcessorArchitecture = requiredProcessorArchitecture;
             manifestInfo.Prefix = resolvedCommandPrefix;
 
+            // A module is considered compatible if it's not on the System32 module path, or
+            // if it is and declared "Core" as a compatible PSEdition.
+            manifestInfo.IsConsideredEditionCompatible = isConsideredCompatible;
```
daxian-dbw: `IsConsideredEditionCompatible` should reflect the truth about whether a module is considered compatible, which shouldn't be affected by `BaseSkipEditionCheck`.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
```diff
@@ -106,6 +111,110 @@ internal static IEnumerable<string> GetAllAvailableModuleFiles(string topDirecto
...
+        }
+
+        /// <summary>
+        /// Check if the CompatiblePSEditions field of a given module
+        /// declares compatibility with the running PowerShell edition.
+        /// </summary>
+        /// <param name="moduleManifestPath">The path to the module manifest being checked.</param>
+        /// <param name="compatiblePSEditions">The value of the CompatiblePSEditions field of the module manifest.</param>
+        /// <returns>True if the module is compatible with the running PowerShell edition, false otherwise.</returns>
+        internal static bool IsPSEditionCompatible(
```
daxian-dbw: The name should be `IsModuleConsideredCompatible`, because a module doesn't necessarily need to declare `CompatiblePSEditions` for it to be considered compatible.

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
daxian-dbw: I think we agreed to not solve `Get-Module -List -All` within this PR, because we are not sure what's the right UX.
Anything changes since we last discussion?

src/System.Management.Automation/engine/Modules/PSModuleInfo.cs
daxian-dbw: Please open an issue tracking this.

src/System.Management.Automation/engine/Modules/PSModuleInfo.cs
```diff
@@ -834,6 +837,22 @@ internal void AddToCompatiblePSEditions(string psEdition)
...
+        }
+
+        /// <summary>
+        /// Describes whether the module was considered compatible at load time.
+        /// Any module not on the System32 module path should have this as true.
+        /// Modules loaded from the System32 module path will have this as true if they
+        /// have declared edition compatibility with PowerShell Core. Currently, this field
+        /// is true for all non-psd1 module files, when it should not be. Being able to
+        /// load psm1/dll modules from the System32 module path without needing to skip
+        /// the edition check is considered a bug and should be fixed.
```
daxian-dbw: Please open an issue tracking this.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
rjmholt: Ah good catch!

daxian-dbw:
@rjmholt Can you please respond to all open comments?

src/System.Management.Automation/engine/Modules/ModuleUtils.cs
rjmholt: Please see https://github.com/PowerShell/PowerShell/issues/7297

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
rjmholt: Please see https://github.com/PowerShell/PowerShell/issues/7298

src/System.Management.Automation/engine/Modules/PSModuleInfo.cs
rjmholt: Please see https://github.com/PowerShell/PowerShell/issues/7296

daxian-dbw (approved):
LGTM. Will merge the PR once CI builds pass.

daxian-dbw:
This feature PR was squash-merged because:
1. The commit history is not reasonably clean.
2. We need to cherry-pick this changes to the `preview-4` release branch.

iSazonov:
@rjmholt Great work! Thanks!

PR #7191, opened 2018-06-27
Update remote prompt when using ssh to show username if different

### LINKED ISSUE #7156: Default prompt for SSH session should include USERNAME/LOGNAME
This is a suggestion that would bring PowerShell more into line with Bash.  For instance, if I ssh from my Windows 10 OpenSSH enabled system to a RHEL7 box, this is the prompt I get from Bash:
```
PS C:\> ssh mctcbld@es-kalbuild-rh.acme.com
mctcbld@es-kalbuild-rh.acme.com's password:
Last login: Thu Jun 21 09:51:35 2018 from stacy1.acme.com
Kickstarted on 2016-02-26
[mctcbld@AD.ACME.COM@es-kalbuild-rh ~]$
```
This system has PowerShell Core 6.0.1 installed on it and if I open a PSSession to it, I get the hostname in the default prompt but not the username/logname.  FWIW I never log into this system as myself so it is very useful to see what user I'm logged in as.
```
> Enter-PSSession -HostName es-kalbuild-rh.acme.com -UserName mctcbld
mctcbld@es-kalbuild-rh.acme.com's password:
[es-kalbuild-rh.acme.com]: PS /home/mctcbld>
```
The default prompt I'd like to see is:
```
[mctcbld@es-kalbuild-rh]: PS /home/mctcbld>
# or maybe even
[mctcbld@AD.ACME.COM@es-kalbuild-rh]: PS ~>
```
I don't need to see the DNS suffix - just the hostname.

Finally, the docs I see for configuring powershell as a subsystem of sshd, say to use `-NoProfile` but then how could you `configure/customize` this prompt for the SSH transport?

SteveL-MSFT:
## PR Summary

Currently, the remote prompt only shows the remote computer name:

> [computer] PS:\foo>

With this change, if you use PSRP over SSH and you specified a different username, it shows whom you logged in as:

> [user@computer] PS:\foo>

Fix is when the remote prompt is being created to check if using SSH, if so, check if the current user is different than the specified user and specified.  If different, use `[user@computer]`, otherwise fallback to default `[computer]` prompt.

Fix https://github.com/PowerShell/PowerShell/issues/7156

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
+++ b/src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
@@ -781,6 +781,7 @@ namespace System.Management.Automation
                     return string.Format(CultureInfo.InvariantCulture, "[{0}@{1}]: {2}", sshConnectionInfo.UserName, sshConnectionInfo.ComputerName, basePrompt);
                 }
             }
+
             return string.Format(CultureInfo.InvariantCulture, "[{0}]: {1}", runspace.ConnectionInfo.ComputerName, basePrompt);
         }
 
--- a/src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
+++ b/src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
@@ -774,9 +774,11 @@ namespace System.Management.Automation
             }
 
             SSHConnectionInfo sshConnectionInfo = runspace.ConnectionInfo as SSHConnectionInfo;
+
+            // Usernames are case-sensitive on Unix systems
             if (sshConnectionInfo != null &&
                 !string.IsNullOrEmpty(sshConnectionInfo.UserName) &&
-                !System.Environment.UserName.Equals(sshConnectionInfo.UserName, StringComparison.Ordinal)) // Usernames are case-sensitive on Unix
+                !System.Environment.UserName.Equals(sshConnectionInfo.UserName, StringComparison.Ordinal))
             {
                 return string.Format(CultureInfo.InvariantCulture, "[{0}@{1}]: {2}", sshConnectionInfo.UserName, sshConnectionInfo.ComputerName, basePrompt);
             }
--- a/src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
+++ b/src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
@@ -776,7 +776,7 @@ namespace System.Management.Automation
             SSHConnectionInfo sshConnectionInfo = runspace.ConnectionInfo as SSHConnectionInfo;
             if (sshConnectionInfo != null &&
                 !string.IsNullOrEmpty(sshConnectionInfo.UserName) &&
-                !System.Environment.UserName.Equals(sshConnectionInfo.UserName, StringComparison.OrdinalIgnoreCase))
+                !System.Environment.UserName.Equals(sshConnectionInfo.UserName, StringComparison.Ordinal)) // Usernames are case-sensitive on Unix
             {
                 return string.Format(CultureInfo.InvariantCulture, "[{0}@{1}]: {2}", sshConnectionInfo.UserName, sshConnectionInfo.ComputerName, basePrompt);
             }
--- a/src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
+++ b/src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
@@ -772,10 +772,16 @@ namespace System.Management.Automation
             {
                 return basePrompt;
             }
-            else
+            else if (runspace.ConnectionInfo is SSHConnectionInfo)
             {
-                return string.Format(CultureInfo.InvariantCulture, "[{0}]: {1}", runspace.ConnectionInfo.ComputerName, basePrompt);
+                SSHConnectionInfo sshConnectionInfo = runspace.ConnectionInfo as SSHConnectionInfo;
+                if (!string.IsNullOrEmpty(sshConnectionInfo.UserName) &&
+                    !System.Environment.UserName.Equals(sshConnectionInfo.UserName, StringComparison.OrdinalIgnoreCase))
+                {
+                    return string.Format(CultureInfo.InvariantCulture, "[{0}@{1}]: {2}", sshConnectionInfo.UserName, sshConnectionInfo.ComputerName, basePrompt);
+                }
             }
+            return string.Format(CultureInfo.InvariantCulture, "[{0}]: {1}", runspace.ConnectionInfo.ComputerName, basePrompt);
         }
 
         internal static bool IsProcessInteractive(InvocationInfo invocationInfo)
--- a/src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
+++ b/src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
@@ -772,10 +772,11 @@ namespace System.Management.Automation
             {
                 return basePrompt;
             }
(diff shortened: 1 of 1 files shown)


src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
```diff
@@ -772,10 +772,16 @@ internal static string GetRemotePrompt(RemoteRunspace runspace, string basePromp
             {
                 return basePrompt;
             }
-            else
+            else if (runspace.ConnectionInfo is SSHConnectionInfo)
             {
-                return string.Format(CultureInfo.InvariantCulture, "[{0}]: {1}", runspace.ConnectionInfo.ComputerName, basePrompt);
+                SSHConnectionInfo sshConnectionInfo = runspace.ConnectionInfo as SSHConnectionInfo;
+                if (!string.IsNullOrEmpty(sshConnectionInfo.UserName) &&
+                    !System.Environment.UserName.Equals(sshConnectionInfo.UserName, StringComparison.OrdinalIgnoreCase))
```
iSazonov: Seems we could remove `!string.IsNullOrEmpty(sshConnectionInfo.UserName)`.

src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
```diff
@@ -772,10 +772,16 @@ internal static string GetRemotePrompt(RemoteRunspace runspace, string basePromp
...
+                {
+                    return string.Format(CultureInfo.InvariantCulture, "[{0}@{1}]: {2}", sshConnectionInfo.UserName, sshConnectionInfo.ComputerName, basePrompt);
+                }
             }
+            return string.Format(CultureInfo.InvariantCulture, "[{0}]: {1}", runspace.ConnectionInfo.ComputerName, basePrompt);
```
iSazonov: Please add new line (from CodeFactor).

src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
SteveL-MSFT: In case where username is not explicitly used, at this point in the code `UserName` is empty so without this check it shows up as `[@computer]` since empty != current username.

src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
```diff
@@ -772,10 +772,17 @@ internal static string GetRemotePrompt(RemoteRunspace runspace, string basePromp
             {
                 return basePrompt;
             }
-            else
+            else if (runspace.ConnectionInfo is SSHConnectionInfo)
             {
-                return string.Format(CultureInfo.InvariantCulture, "[{0}]: {1}", runspace.ConnectionInfo.ComputerName, basePrompt);
+                SSHConnectionInfo sshConnectionInfo = runspace.ConnectionInfo as SSHConnectionInfo;
+                if (!string.IsNullOrEmpty(sshConnectionInfo.UserName) &&
+                    !System.Environment.UserName.Equals(sshConnectionInfo.UserName, StringComparison.OrdinalIgnoreCase))
```
PaulHigin: For Windows sshConnectionInfo.UserName can have a domain component (domainName\userName), so we need to handle that unless we make this Linux platform only.  We can create a common helper method from here: https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs#L2052.  Or maybe update SSHConnectionInfo to automatically parse and expose UserName and DomainName properties.

PaulHigin (changes requested):

PaulHigin (changes requested):

src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
PaulHigin: I am assuming we would want the prompt to look something like: `[user@domain@computer]: PS C:\>`

PaulHigin: Or maybe just drop the domain...

rkeithhill: I think you display it as `[<username>@<hostname>]` where `<username>` may or may not specify the additional `@domain`.  That's what I see from Bash.

SteveL-MSFT: Currently, it depends on how the user specified the domain name.  If they used `user@domain`, it would be `[user@domain@computer]`.  Alternatively, if they used `domain\user`, it shows up as `[domain\user@computer]`.  Personally, I think we should show the domain and fine with whichever form the user provided.

PaulHigin: We don't currently support `user@domain` format.  It seems like domain information is not needed, but I am fine with `[domain\user@computer`].  We can easily change it later if it is unpopular.

PaulHigin (approved):
Nitpick: I just realized looking at the whole method that we don't need the else block because we are doing early returns.  We can just remove it.

SteveL-MSFT:
@PaulHigin agree it's not needed.  Can you take a quick look and re-approve?

PaulHigin (approved):
LGTM

src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
```diff
@@ -772,10 +772,16 @@ internal static string GetRemotePrompt(RemoteRunspace runspace, string basePromp
             {
                 return basePrompt;
             }
-            else
+
+            SSHConnectionInfo sshConnectionInfo = runspace.ConnectionInfo as SSHConnectionInfo;
+            if (sshConnectionInfo != null &&
+                !string.IsNullOrEmpty(sshConnectionInfo.UserName) &&
+                !System.Environment.UserName.Equals(sshConnectionInfo.UserName, StringComparison.OrdinalIgnoreCase))
```
anmenaga: Aren't unix usernames case sensitive?

anmenaga (changes requested):

src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
SteveL-MSFT: You're right, Unix usernames are case-sensitive.  I'll change this to a case sensitive comparison.

src/System.Management.Automation/engine/hostifaces/HostUtilities.cs
```diff
@@ -772,10 +772,16 @@ internal static string GetRemotePrompt(RemoteRunspace runspace, string basePromp
             {
                 return basePrompt;
             }
-            else
+
+            SSHConnectionInfo sshConnectionInfo = runspace.ConnectionInfo as SSHConnectionInfo;
+            if (sshConnectionInfo != null &&
+                !string.IsNullOrEmpty(sshConnectionInfo.UserName) &&
+                !System.Environment.UserName.Equals(sshConnectionInfo.UserName, StringComparison.Ordinal)) // Usernames are case-sensitive on Unix
```
iSazonov: Please move the comment on separate line.

TravisEz13:
@anmenaga  Please update your review

rkeithhill (approved):
LGTM

anmenaga (approved):

iSazonov (approved):

iSazonov:
If our docs have related samples we should correct them.

PR #7210, opened 2018-06-29
Make Start-PSPester more user friendly by calling Restore-Pester automatically when needed and not having to specify -Path as a named parameter for Start-PSPester

bergmeister:
## PR Summary

- Make `Start-PSPester` automatically call `Restore-PSPester` when needed. This simplifies it and reduces manual steps when running it interactively (I tend to use `git clean -dfx` or `Start-PSBuild -Clean` relatively often as I do not work with the code base every day). I am running Pester tests usually using `Start-PSPester`
- Make `-Path` parameter on `Start-PSPester` the 0th parameter to so that `Start-PSPester` $Path works without needing to specify `-Path` as a named parameter (useful when using the cmdlet interactively for running tests locally). Yes, I know, named parameters should be encouraged but for cmdlets that are used interactively a lot and where mostly just one parameter is used, I think it is fine and I rather prefer the lazy option in this case (PSSA 'allows' for up to 3 parameters to be not named before warning btw).

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**


build.psm1
```diff
@@ -648,11 +648,6 @@ Fix steps:
     if ($PSModuleRestore) {
         Restore-PSModuleToBuild -PublishPath $publishPath
     }
-
-    # Restore the Pester module
-    if ($CI) {
-        Restore-PSPester -Destination (Join-Path $publishPath "Modules")
-    }
```
daxian-dbw: I think this is still useful -- I want to build powershell and then manually run some Pester tests, but not `Start-PSPester`. In this case, I can run `Start-PSBuild -Clean -CI` and Pester will be installed.

build.psm1
```diff
@@ -981,11 +976,12 @@ function Publish-PSTestTools {
 function Start-PSPester {
     [CmdletBinding(DefaultParameterSetName='default')]
     param(
+        [Parameter(Position=0)]
+        [string[]]$Path = @("$PSScriptRoot/test/common","$PSScriptRoot/test/powershell"),
```
daxian-dbw: In your PR description, you wrote:
> Make `-Path` parameter on `Start-PSPester` the 0th parameter to so that `Restore-PSPester $Path` works without needing to specify -Path as a named parameter.

I'm confused about the `Restore-PSPester $Path` part. You are not calling `Restore-PSPester` with `$Path` in your change.

daxian-dbw (changes requested):

build.psm1
bergmeister: This change was for easier interactive usage only

build.psm1
bergmeister: Maybe I misunderstood the usage of this line. Because I made `Start-PSPester` restore Pester if needed, I thought, I could remove this line. I would be happy to revert this change but would also be interested in why one would want/need Pester to be installed when building PowerShell (my guess is that your use case is to call `Start-DevPowerShell` and run the pester test from there?)

daxian-dbw: > would also be interested in why one would want/need Pester to be installed when building PowerShell (my guess is that your use case is to call Start-DevPowerShell and run the pester test from there?)

When I'm wringing Pester tests for some code changes and just want to run the tests I'm working on, I would build the code base, start `pwsh` from the bin folder, and run `Invoke-Pester -Script <path-of-my-test-file>`. Or when I'm investigating a test failure from daily build, I might want to sync master branch, build, start the `pwsh` produced from the build, run `Invoke-Pester` with the file where the test fails.

Sure, we can install Pester to user module path, so it will always be available, but install the Pester module while build powershell could be useful in case you don't have Pester in other module paths.

build.psm1
daxian-dbw: You mentioned about `Restore-PSPester $Path` in your description, but I don't see how making `-Path` positional affect your use of `Restore-PSPester`.

Do you mean `without needing to specify -Path for Start-PSPester` ?

bergmeister: Sorry, my bad. I meant to write `Start-PSPester $Path`, you are right

build.psm1
bergmeister: Ok, I see. I will revert this change then.
When I run tests locally, then my workflow is `Start-PSBuild ...` and then `Start-PSPester $TestFile`

daxian-dbw: Yeah, sometimes I expect some of my tests to fail initially, and it's easier for me to keep the `pwsh` instance around for some ad-hoc verifications.

build.psm1
daxian-dbw: Can you please update your PR description? Thanks!

build.psm1
```diff
@@ -1002,23 +1003,9 @@ function Start-PSPester {
     )
 
     $getModuleResults = Get-Module -ListAvailable -Name $Pester -ErrorAction SilentlyContinue
-    if (-not $getModuleResults)
+    if (-not ($getModuleResults | Where-Object { $_.Version -ge "4.2" } ))
```
daxian-dbw: How about move `Where-Object` part after the `Get-Module` call? Like `Get-Module ... | Where-Object ...`. Then the if check would be simpler to read here.

bergmeister: Ok, done.

build.psm1
bergmeister: Ok, done. Sorry for that mistake.

daxian-dbw (approved):

anmenaga:
Restarted Travis CI build because of this error on Mac:
`dotnet_install: Error: Could not download .NET Core SDK version 2.1.301`

PR #7221, opened 2018-07-02
Make TabCompletion case-insensitive for file names on Unix (2-nd attempt)

### LINKED ISSUE #1273: Location completion should be case-insensitive

iSazonov:
## PR Summary

Attempt to fix #1273.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
+++ b/test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
@@ -263,23 +263,23 @@ Describe "TabCompletion" -Tags CI {
         It "TabCompletion should be case-insensitive for file names on Windows and MacOS" -Skip:($IsLinux) {
             Push-Location -Path $tempDir
             $res = TabExpansion2 -inputScript $oneSubDirLowerTest -cursorColumn $oneSubDirLowerTest.Length
-            $res.CompletionMatches.Count | Should BeGreaterThan 0
-            $res.CompletionMatches[0].CompletionText | Should Be ".${separator}$oneSubDirTest"
+            $res.CompletionMatches.Count | Should -BeGreaterThan 0
+            $res.CompletionMatches[0].CompletionText | Should -BeExactly ".${separator}$oneSubDirTest"
 
             $res = TabExpansion2 -inputScript $oneSubDirUpperTest -cursorColumn $oneSubDirUpperTest.Length
-            $res.CompletionMatches.Count | Should BeGreaterThan 0
-            $res.CompletionMatches[0].CompletionText | Should Be ".${separator}$oneSubDirTest"
+            $res.CompletionMatches.Count | Should -BeGreaterThan 0
+            $res.CompletionMatches[0].CompletionText | Should -BeExactly ".${separator}$oneSubDirTest"
         }
 
         It "TabCompletion should be case-sensitive for file names on Unix" -Skip:(!$IsLinux) {
             Push-Location -Path $tempDir
             $res = TabExpansion2 -inputScript $oneSubDirLowerTest -cursorColumn $oneSubDirLowerTest.Length
-            $res.CompletionMatches.Count | Should BeGreaterThan 0
-            $res.CompletionMatches[0].CompletionText | Should BeExactly ".${separator}$oneSubDirTest"
+            $res.CompletionMatches.Count | Should -BeGreaterThan 0
+            $res.CompletionMatches[0].CompletionText | Should -BeExactly ".${separator}$oneSubDirTest"
 
             $res = TabExpansion2 -inputScript $oneSubDirUpperTest -cursorColumn $oneSubDirUpperTest.Length
-            $res.CompletionMatches.Count | Should BeGreaterThan 0
-            $res.CompletionMatches[0].CompletionText | Should BeExactly ".${separator}$oneSubDirTest2"
+            $res.CompletionMatches.Count | Should -BeGreaterThan 0
+            $res.CompletionMatches[0].CompletionText | Should -BeExactly ".${separator}$oneSubDirTest2"
         }
 
         It "Input '<inputStr>' should successfully complete" -TestCases $testCases {
--- a/test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
+++ b/test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
@@ -260,7 +260,7 @@ Describe "TabCompletion" -Tags CI {
             Pop-Location
         }
 
-        It "TabCompletion should be case-insensitive for file names on Windows" -Skip:(!$IsWindows) {
+        It "TabCompletion should be case-insensitive for file names on Windows and MacOS" -Skip:($IsLinux) {
             Push-Location -Path $tempDir
             $res = TabExpansion2 -inputScript $oneSubDirLowerTest -cursorColumn $oneSubDirLowerTest.Length
             $res.CompletionMatches.Count | Should BeGreaterThan 0
@@ -271,7 +271,7 @@ Describe "TabCompletion" -Tags CI {
             $res.CompletionMatches[0].CompletionText | Should Be ".${separator}$oneSubDirTest"
         }
 
-        It "TabCompletion should be case-sensitive for file names on Unix" -Skip:($IsWindows) {
+        It "TabCompletion should be case-sensitive for file names on Unix" -Skip:(!$IsLinux) {
             Push-Location -Path $tempDir
             $res = TabExpansion2 -inputScript $oneSubDirLowerTest -cursorColumn $oneSubDirLowerTest.Length
             $res.CompletionMatches.Count | Should BeGreaterThan 0
--- a/test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
+++ b/test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
@@ -206,8 +206,12 @@ Describe "TabCompletion" -Tags CI {
 
     Context "File name completion" {
         BeforeAll {
+            $oneSubDirLowerTest = "onesubd"
+            $oneSubDirUpperTest = "oneSubD"
+            $oneSubDirTest = "onesubdir"
+
             $tempDir = Join-Path -Path $TestDrive -ChildPath "baseDir"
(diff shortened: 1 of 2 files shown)


iSazonov:
@SteveL-MSFT @vors Can we review and merge the PR before 6.1?

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -250,6 +260,28 @@ Describe "TabCompletion" -Tags CI {
             Pop-Location
         }
 
+        It "TabCompletion should be case-insensitive for file names on Windows and MacOS" -Skip:($IsLinux) {
+            Push-Location -Path $tempDir
+            $res = TabExpansion2 -inputScript $oneSubDirLowerTest -cursorColumn $oneSubDirLowerTest.Length
+            $res.CompletionMatches.Count | Should BeGreaterThan 0
+            $res.CompletionMatches[0].CompletionText | Should Be ".${separator}$oneSubDirTest"
```
SteveL-MSFT: I would expect the result casing to match the actual casing.  Use -BeExactly

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -250,6 +260,28 @@ Describe "TabCompletion" -Tags CI {
...
+
+            $res = TabExpansion2 -inputScript $oneSubDirUpperTest -cursorColumn $oneSubDirUpperTest.Length
+            $res.CompletionMatches.Count | Should BeGreaterThan 0
```
SteveL-MSFT: Shouldn't we know exactly the number of expected completions?

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -250,6 +260,28 @@ Describe "TabCompletion" -Tags CI {
...
+            $res.CompletionMatches[0].CompletionText | Should Be ".${separator}$oneSubDirTest"
```
SteveL-MSFT: -BeExactly

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -250,6 +260,28 @@ Describe "TabCompletion" -Tags CI {
...
+        }
+
+        It "TabCompletion should be case-sensitive for file names on Unix" -Skip:(!$IsLinux) {
+            Push-Location -Path $tempDir
+            $res = TabExpansion2 -inputScript $oneSubDirLowerTest -cursorColumn $oneSubDirLowerTest.Length
+            $res.CompletionMatches.Count | Should BeGreaterThan 0
```
SteveL-MSFT: Shouldn't we know exactly the number of expected completions?

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -250,6 +260,28 @@ Describe "TabCompletion" -Tags CI {
...
+            $res.CompletionMatches[0].CompletionText | Should BeExactly ".${separator}$oneSubDirTest"
```
SteveL-MSFT: -BeExactly

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -250,6 +260,28 @@ Describe "TabCompletion" -Tags CI {
...
+
+            $res = TabExpansion2 -inputScript $oneSubDirUpperTest -cursorColumn $oneSubDirUpperTest.Length
+            $res.CompletionMatches.Count | Should BeGreaterThan 0
+            $res.CompletionMatches[0].CompletionText | Should BeExactly ".${separator}$oneSubDirTest2"
```
SteveL-MSFT: -BeExactly

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -250,6 +260,28 @@ Describe "TabCompletion" -Tags CI {
...
+        }
+
+        It "TabCompletion should be case-sensitive for file names on Unix" -Skip:(!$IsLinux) {
+            Push-Location -Path $tempDir
+            $res = TabExpansion2 -inputScript $oneSubDirLowerTest -cursorColumn $oneSubDirLowerTest.Length
+            $res.CompletionMatches.Count | Should BeGreaterThan 0
+            $res.CompletionMatches[0].CompletionText | Should BeExactly ".${separator}$oneSubDirTest"
+
+            $res = TabExpansion2 -inputScript $oneSubDirUpperTest -cursorColumn $oneSubDirUpperTest.Length
+            $res.CompletionMatches.Count | Should BeGreaterThan 0
```
SteveL-MSFT: Shouldn't we know exactly how many results we should get since we control the folder structure?

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -250,6 +260,28 @@ Describe "TabCompletion" -Tags CI {
             Pop-Location
         }
 
+        It "TabCompletion should be case-insensitive for file names on Windows and MacOS" -Skip:($IsLinux) {
+            Push-Location -Path $tempDir
+            $res = TabExpansion2 -inputScript $oneSubDirLowerTest -cursorColumn $oneSubDirLowerTest.Length
+            $res.CompletionMatches.Count | Should BeGreaterThan 0
```
SteveL-MSFT: Shouldn't we know exactly the number of expected completions?

SteveL-MSFT (changes requested):

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
iSazonov: Fixed.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
iSazonov: For this fix, it does not matter count (the check is only protection from exception in next line) - the order is important.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
iSazonov: Fixed.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
SteveL-MSFT: Since this test created the folder structure, the number of completions should be the same each time, right?

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -250,6 +260,28 @@ Describe "TabCompletion" -Tags CI {
...
+
+            $res = TabExpansion2 -inputScript $oneSubDirUpperTest -cursorColumn $oneSubDirUpperTest.Length
+            $res.CompletionMatches | Should -HaveCount 1
+            $res.CompletionMatches[0].CompletionText | Should -BeExactly ".${separator}$oneSubDirTest"
+        }
+
+        It "TabCompletion should be case-sensitive for file names on Unix" -Skip:(!$IsLinux) {
+            Push-Location -Path $tempDir
+            $res = TabExpansion2 -inputScript $oneSubDirLowerTest -cursorColumn $oneSubDirLowerTest.Length
+            $res.CompletionMatches | Should -HaveCount 2
```
SteveL-MSFT: This should still be `-HaveCount 1`?  You're testing that "onesubd" and "oneSubD" both return "onesubdir", right?

It seems we should have a test for Linux for a directory structure like:

```none
testfolder
testFolder
```

and tabcomplete off `test` should return two results

SteveL-MSFT (changes requested):

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
SteveL-MSFT: This additional test case is still needed.

iSazonov: @SteveL-MSFT It is already there - see line  226. So I removed last commit. Seems my solution is not full - the count should be 2 on . 
How can I get the artifact (compiled PowerShell) from CI Travis?

SteveL-MSFT: @iSazonov you're right, I misunderstood that test.  Did a little investigation and it appears that the change to GetFileSystemEntries() is not sufficient as a call to Resolve-Path earlier only returns case-sensitive matches on Linux.

iSazonov: Should we open new issue to discuss making `Resolve-Path` case-insensitive? This is most likely to affect globbing whole.

SteveL-MSFT: @iSazonov we should probably have a new issue for `Resolve-Path`.  May need to add a `-CaseInsensitive` switch to maintain backwards compatibility.

iSazonov:
I considered `Resolve-Path -CaseInsensitive` . Adding the parameter in APIs looks as hude change in engine (in public APIs too). Another way is to introduce new internal flag/property. I like this more but don't still get full understanding of the consequences (side affects).

PR #7231, opened 2018-07-03
Type inference for Select-Object, Group-Object, PSObject and Hashtable

### LINKED ISSUE #7230: Add type inference for PSObject and Hashtable
Constructs like 
```
@{
    Name = "Staffan",
    Size = 3
}
``` 
is quite easy to infer types from, and the same is true for the conversion expression to 

```
[pscustomobject] @{
    Name = "Staffan",
    Size = 3
}
```

The same mechanisms that is needed to support synthetic `PSTypeName`s could also be used to provide type inference for `Select-Object -Property` and the `Group` and `Value` propererties of the output of `Group-Object`

None of this is supported as of 6.1-preview 3.

powercode:
## PR Summary

Fixes #7230.

Adding the concept of a PSSyntheticTypeName, derived from PSTypeName,
that extends it with a list of synthetic members.

This allows us to express the inferred type of
```
[pscustomobject] @{
   A = 1
   B = "2"
}
```
as a "PSObject#A:B" and with information about the types of A and B.

This is also used to annotate the output of
```
Select-Object -Property
Select-Object -ExcludeProperty
Select-Object -ExpandProperty
```
Finally, it adds information about the types of the
`Group` and `Value` properties of the output of `Group-Object`

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [x] [Add `[feature]` if the change is significant or affects feature tests]

### DIFF
--- a/src/System.Management.Automation/engine/CommandInfo.cs
+++ b/src/System.Management.Automation/engine/CommandInfo.cs
@@ -3,11 +3,15 @@
 
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using System.Management.Automation.Language;
 using System.Management.Automation.Runspaces;
 using System.Reflection;
 using System.Runtime.ExceptionServices;
+using System.Text;
+
 using Microsoft.PowerShell.Commands;
 
 namespace System.Management.Automation
@@ -794,6 +798,17 @@ namespace System.Management.Automation
         }
 
         /// <summary>
+        /// This constructor is used when the creating a PSObject with a custom typename.
+        /// </summary>
+        /// <param name="name">The name of the type.</param>
+        /// <param name="type">The real type.</param>
+        internal PSTypeName(string name, Type type)
+        {
+            Name = name;
+            _type = type;
+        }
+
+        /// <summary>
         /// This constructor is used when the type is defined in PowerShell.
         /// </summary>
         /// <param name="typeDefinitionAst">The type definition from the ast.</param>
@@ -899,6 +914,95 @@ namespace System.Management.Automation
         {
             return Name ?? string.Empty;
         }
+    }
+
+    [DebuggerDisplay("{PSTypeName} {Name}")]
+    internal struct PSMemberNameAndType
+    {
+        public readonly string Name;
+
+        public readonly PSTypeName PSTypeName;
+
+        public readonly object Value;
+
+        public PSMemberNameAndType(string name, PSTypeName typeName, object value = null)
+        {
+            Name = name;
+            PSTypeName = typeName;
+            Value = value;
+        }
+    }
+
+    /// <summary>
+    /// Represents dynamic types such as <see cref="System.Management.Automation.PSObject"/>,
+    /// but can be used where a real type might not be available, in which case the name of the type can be used.
+    /// The type encodes the members of dynamic objects in the type name.
+    /// </summary>
+    internal class PSSyntheticTypeName : PSTypeName
+    {
+        internal static PSSyntheticTypeName Create(string typename, IList<PSMemberNameAndType> membersTypes) => Create(new PSTypeName(typename), membersTypes);
+
+        internal static PSSyntheticTypeName Create(Type type, IList<PSMemberNameAndType> membersTypes) => Create(new PSTypeName(type), membersTypes);
+
+        internal static PSSyntheticTypeName Create(PSTypeName typename, IList<PSMemberNameAndType> membersTypes)
+        {
+            var typeName = GetMemberTypeProjection(typename.Name, membersTypes);
+            var members = new List<PSMemberNameAndType>();
+            members.AddRange(membersTypes);
+            members.Sort((c1,c2) =>string.Compare(c1.Name, c2.Name, StringComparison.CurrentCultureIgnoreCase));
+            return new PSSyntheticTypeName(typeName, typename.Type, members);
+        }
+
+        private PSSyntheticTypeName(string typeName, Type type, IList<PSMemberNameAndType> membersTypes)
+        : base(typeName, type)
+        {
+            Members = membersTypes;
+            if (type != typeof(PSObject))
+            {
+                return;
+            }
+
+            for (int i = 0; i < Members.Count; i++)
+            {
+                var psMemberNameAndType = Members[i];
+                if (IsPSTypeName(psMemberNameAndType))
+                {
+                    Members.RemoveAt(i);
+                    break;
+                }
+            }
+        }
+
+        private static bool IsPSTypeName(PSMemberNameAndType member) => string.Compare("PSTypeName", member.Name, StringComparison.CurrentCultureIgnoreCase) == 0;
+
+        private static string GetMemberTypeProjection(string typename, IList<PSMemberNameAndType> members)
+        {
+            if (typename == typeof(PSObject).FullName)
+            {
+                foreach (var mem in members)
+                {
+                    if (IsPSTypeName(mem))
+                    {
+                        typename = mem.Value.ToString();
+                    }
+                }
+            }
+
+            var builder = new StringBuilder(typename, members.Count * 7);
+            builder.Append('#');
+            foreach (var m in members.OrderBy(m => m.Name))
+            {
+                if (!IsPSTypeName(m))
+                {
+                    builder.Append(m.Name).Append(":");
+                }
+            }
+
+            builder.Length--;
+            return builder.ToString();
+        }
+
+        public IList<PSMemberNameAndType> Members { get; }
     }
 
     internal interface IScriptCommandInfo
--- a/src/System.Management.Automation/engine/CommandInfo.cs
+++ b/src/System.Management.Automation/engine/CommandInfo.cs
@@ -802,7 +802,7 @@ namespace System.Management.Automation
         /// </summary>
         /// <param name="name">The name of the type.</param>
         /// <param name="type">The real type.</param>
-        internal PSTypeName(string name, Type type)
+        public PSTypeName(string name, Type type)
         {
             Name = name;
             _type = type;
@@ -973,7 +973,7 @@ namespace System.Management.Automation
             }
         }
 
-        private static bool IsPSTypeName(PSMemberNameAndType member) => string.Compare("PSTypeName", member.Name, StringComparison.CurrentCultureIgnoreCase) == 0;
+        private static bool IsPSTypeName(PSMemberNameAndType member) => member.Name.Equals(nameof(PSTypeName), StringComparison.CurrentCultureIgnoreCase);
 
         private static string GetMemberTypeProjection(string typename, IList<PSMemberNameAndType> members)
         {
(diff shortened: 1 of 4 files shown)


powercode:
I don't intend to fix the remaining CodeFactor issues.

powercode:
Ping

daxian-dbw:
Sorry for delaying so long. Will review in a day or two.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -515,6 +524,33 @@ object ICustomAstVisitor.VisitArrayLiteral(ArrayLiteralAst arrayLiteralAst)
...
+                foreach (var kv in hashtableAst.KeyValuePairs)
+                {
+                    if (SafeExprEvaluator.TrySafeEval(kv.Item1, _context.ExecutionContext, out object nameValue))
+                    {
+                        var name = nameValue.ToString();
+                        var typeName = InferTypes(kv.Item2).FirstOrDefault()?.Name ?? "System.Object";
+                        object value = null;
+                        if (kv.Item2 is PipelineAst pipelineAst
+                            && pipelineAst.PipelineElements[0] is CommandExpressionAst commandExpressionAst
+                            && SafeExprEvaluator.TrySafeEval(commandExpressionAst.Expression, _context.ExecutionContext, out object safeValue))
```
daxian-dbw: `pipelineAst` might have more than one pipeline elements. In that case, does it still make sense to evaluate the `commandExpressionAst`?

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -515,6 +524,33 @@ object ICustomAstVisitor.VisitArrayLiteral(ArrayLiteralAst arrayLiteralAst)
 
         object ICustomAstVisitor.VisitHashtable(HashtableAst hashtableAst)
         {
+            if (hashtableAst.KeyValuePairs.Count > 0)
+            {
+                var properties = new List<PSMemberNameAndType>();
+
+                foreach (var kv in hashtableAst.KeyValuePairs)
+                {
+                    if (SafeExprEvaluator.TrySafeEval(kv.Item1, _context.ExecutionContext, out object nameValue))
```
daxian-dbw: Shall we check for `StringConstantExpressionAst` first? There is a good chance we can avoid evaluating the expression.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -581,6 +617,15 @@ object ICustomAstVisitor.VisitUnaryExpression(UnaryExpressionAst unaryExpression
         object ICustomAstVisitor.VisitConvertExpression(ConvertExpressionAst convertExpressionAst)
         {
             var type = convertExpressionAst.Type.TypeName.GetReflectionType();
+
+            if (type == typeof(PSObject) && convertExpressionAst.Child is HashtableAst hashtableAst)
```
daxian-dbw: Do you mean to use `type == typeof(PSCustomObject)`? `[psobject]@{ a = 1 }` just wrap the hashtable into psobject, and it should still be `Hashtable` type.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -174,6 +175,14 @@ internal IList<object> GetMembersByInferredType(PSTypeName typename, bool isStat
             List<object> results = new List<object>();
```
daxian-dbw: Any reason why this is `new Linux<PSMemberInfo>`?

daxian-dbw (commented):
Sorry for being late on the code review. Left a few comments.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
powercode: Good catch - no in that case, I think we should just bail.

Adding a `&& pipelineAst.PipelineElements.Count == 1` condition.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
powercode: Yes, we should :) 
And probably check that for the value too. Chances are that it's either a number or a string, right?

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
daxian-dbw: Maybe use `var expression = pipelineAst.GetPureExpression()`. If the pipeline is not a pure expression the method will return `null`.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
daxian-dbw: Yes, the value could also be a constant. Maybe check for `ConstantExpressionAst` for value?

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
daxian-dbw: I guess `[psobject]@{a =1}` would be very rare and thus we don't need to differentiate `[pscustomeobject]@{a=1}` from `[psobject]@{a =1}`.

But can you please add a comment about `[PSCustomObject]` actually returns `PSObject` as reflection type?

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
powercode: Not that I can think of - I'll make the change and see what happens :)

powercode: I was wrong. It needs to be object, since we have expressions like
`results.AddRange(cimClass.CimClassProperties);`

powercode:
@daxian-dbw Thanks for your review! Very constructive as usual!

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -515,6 +524,64 @@ object ICustomAstVisitor.VisitArrayLiteral(ArrayLiteralAst arrayLiteralAst)
...
+                    else if (SafeExprEvaluator.TrySafeEval(kv.Item1, _context.ExecutionContext, out object nameValue))
+                    {
+                        name = nameValue.ToString();
+                    }
+                    if (name != null)
+                    {
+                        object value = null;
+                        if (kv.Item2 is PipelineAst pipelineAst
+                            && pipelineAst.PipelineElements.Count == 1
+                            && pipelineAst.PipelineElements[0] is CommandExpressionAst commandExpressionAst)
```
daxian-dbw: We can use `pipelineAst.GetPureExpression()` here, which does the same thing and return null if it's not a pure expression. pipelineAst.GetPureExpression()
```c#
if (kv.Item2 is PipelineAst pipelineAst
    && pipelineAst.GetPureExpression() is ExpressionAst expression)
{
      switch (expression)
      {
          ....
```

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -581,6 +648,15 @@ object ICustomAstVisitor.VisitUnaryExpression(UnaryExpressionAst unaryExpression
         object ICustomAstVisitor.VisitConvertExpression(ConvertExpressionAst convertExpressionAst)
         {
             var type = convertExpressionAst.Type.TypeName.GetReflectionType();
+
+            if (type == typeof(PSObject) && convertExpressionAst.Child is HashtableAst hashtableAst)
```
daxian-dbw: I guess `[psobject]@{a =1}` would be very rare and thus we don't need to differentiate `[pscustomeobject]@{a=1}` from `[psobject]@{a =1}`.

It would be great if you can add a comment about `[PSCustomObject]` actually returns `PSObject` as the reflection type.

daxian-dbw (approved):
LGTM. Left two non-blocking comments.
@powercode Thanks for the great contribution, again!

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
powercode: Good point!
I didn't even consider the `[psobject]@{a =1}` case. I'll revisit, now that I understand what you meant.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
powercode: Didn't know about PureExpression. Excellent!

powercode: That also covers the redirection case. Nice!

daxian-dbw (approved):
Thanks for addressing the non-blocking comments!

src/System.Management.Automation/engine/CommandInfo.cs
```diff
@@ -793,6 +797,17 @@ public PSTypeName(string name)
             _type = null;
         }
 
+        /// <summary>
+        /// This constructor is used when the creating a PSObject with a custom typename.
+        /// </summary>
+        /// <param name="name">The name of the type.</param>
+        /// <param name="type">The real type.</param>
+        internal PSTypeName(string name, Type type)
```
BrucePay: This looks like it should be public but we can do that in a future PR.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -1099,10 +1292,78 @@ private void InferTypesFromPreviousCommand(CommandAst commandAst, List<PSTypeNam
...
         private void InferTypesFromSelectCommand(PseudoBindingInfo pseudoBinding, CommandAst commandAst, List<PSTypeName> inferredTypes)
         {
-            if (pseudoBinding.BoundArguments.TryGetValue("Property", out _)
-                || pseudoBinding.BoundArguments.TryGetValue("ExcludeProperty", out _))
+            void InferFromSelectProperties(AstParameterArgumentPair astParameterArgumentPair, CommandBaseAst previousPipelineElementAst, bool includeMatchedProperties = true)
             {
-                return;
+                if (astParameterArgumentPair is AstPair astPair)
+                {
+                    object ToWildCardOrString(string value) => WildcardPattern.ContainsWildcardCharacters(value) ? (object)new WildcardPattern(value) : value;
```
BrucePay: Consider using PSPropertyExpression - it already handles wildcards in property names.

BrucePay (approved):
There are a few little things that we can change later. Otherwise it looks OK to me.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
powercode: Maybe even use the ParameterProcessor? So we can handle hashtable arguments also?
I'll leave that to a future PR.

TravisEz13:
Add Committee review for new public interface

SteveL-MSFT:
@PowerShell/powershell-committee reviewed the public constructor for PSTypeName and is ok with the proposal

PR #7240, opened 2018-07-06
Fix for issue 5752: pwsh cannot start in a directory with a name that includes a wildcard

### LINKED ISSUE #5752: Powershell does not open in current directory if its name contains a left square bracket (wildcard)

JamesWTruher:
## PR Summary

Fix #5752.

If you attempt to start pwsh in a directory like `/tmp/[T]est` the shell will default to $PSHOME

The issue was that we attempted to glob against the string returned by `Directory.GetCurrentDirectory()` which we should not do.

Technically, this is a breaking change - anyone that relies on starting in $PSHOME under these circumstances may be surprised by this behavior.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [x] Issue filed - Issue link: https://github.com/PowerShell/PowerShell/issues/5752
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/InitialSessionState.cs
+++ b/src/System.Management.Automation/engine/InitialSessionState.cs
@@ -3229,6 +3229,7 @@ namespace System.Management.Automation.Runspaces
 
                         try
                         {
+                            providerContext.SuppressWildcardExpansion = true;
                             context.EngineSessionState.SetLocation(Directory.GetCurrentDirectory(), providerContext);
                         }
                         catch (ItemNotFoundException)
--- a/test/powershell/Host/ConsoleHost.Tests.ps1
+++ b/test/powershell/Host/ConsoleHost.Tests.ps1
@@ -683,3 +683,24 @@ Describe "Pwsh exe resources tests" -Tag CI {
         $psversiontable.os | Should -MatchExactly "$($osversion.Major).$($osversion.Minor)"
     }
 }
+
+Describe 'Pwsh startup in directories that contain wild cards' -Tag CI {
+    BeforeAll {
+        $powershell = Join-Path -Path $PsHome -ChildPath "pwsh"
+        $dirnames = "[T]est","[Test","T][est","T*est","Test"
+        foreach ( $d in $dirnames ) { new-item -type Directory -path "${TESTDRIVE}/$d" }
+        $testcases = foreach ( $d in $dirnames ) { @{ Dirname = $d } }
+    }
+
+    It "pwsh can startup in a directory named <dirname>" -testcases $testcases {
+        param ( $dirname )
+        try {
+            Push-Location -LiteralPath "${TESTDRIVE}/${dirname}"
+            $result = & $powershell -c '(Get-Item .).Name'
+            $result | Should -BeExactly $dirname
+        }
+        finally {
+            Pop-Location
+        }
+    }
+}
--- a/test/powershell/Host/ConsoleHost.Tests.ps1
+++ b/test/powershell/Host/ConsoleHost.Tests.ps1
@@ -687,9 +687,12 @@ Describe "Pwsh exe resources tests" -Tag CI {
 Describe 'Pwsh startup in directories that contain wild cards' -Tag CI {
     BeforeAll {
         $powershell = Join-Path -Path $PsHome -ChildPath "pwsh"
-        $dirnames = "[T]est","[Test","T][est","T*est","Test"
-        foreach ( $d in $dirnames ) { new-item -type Directory -path "${TESTDRIVE}/$d" }
-        $testcases = foreach ( $d in $dirnames ) { @{ Dirname = $d } }
+        $dirnames = "[T]est","[Test","T][est","Test"
+        $testcases = @()
+        foreach ( $d in $dirnames ) {
+            $null = New-Item -type Directory -path "${TESTDRIVE}/$d"
+            $testcases += @{ Dirname = $d }
+        }
     }
 
     It "pwsh can startup in a directory named <dirname>" -testcases $testcases {
--- a/test/powershell/Modules/ThreadJob/ThreadJob.Tests.ps1
+++ b/test/powershell/Modules/ThreadJob/ThreadJob.Tests.ps1
@@ -275,6 +275,7 @@ Describe 'Basic ThreadJob Tests' -Tags 'CI' {
             $job2 = Start-ThreadJob -ScriptBlock { "Hello 2!" }
             $job3 = Start-ThreadJob -ScriptBlock { "Hello 3!" }
             $job4 = Start-ThreadJob -ScriptBlock { "Hello 4!" }
+            wait-forrunning $job4
 
             $job1,$job2,$job3,$job4 | Wait-Job
 
--- a/test/powershell/Modules/ThreadJob/ThreadJob.Tests.ps1
+++ b/test/powershell/Modules/ThreadJob/ThreadJob.Tests.ps1
@@ -275,7 +275,6 @@ Describe 'Basic ThreadJob Tests' -Tags 'CI' {
             $job2 = Start-ThreadJob -ScriptBlock { "Hello 2!" }
             $job3 = Start-ThreadJob -ScriptBlock { "Hello 3!" }
             $job4 = Start-ThreadJob -ScriptBlock { "Hello 4!" }
-            wait-forrunning $job4
 
             $job1,$job2,$job3,$job4 | Wait-Job
 


JamesWTruher:
it looks like the test that failed is unrelated to my change. In reviewing the threadjob test failure, there looks to be a timing issue. I'll follow up with @PaulHigin.

test/powershell/Host/ConsoleHost.Tests.ps1
```diff
@@ -683,3 +683,24 @@ Describe "Pwsh exe resources tests" -Tag CI {
         $psversiontable.os | Should -MatchExactly "$($osversion.Major).$($osversion.Minor)"
     }
 }
+
+Describe 'Pwsh startup in directories that contain wild cards' -Tag CI {
+    BeforeAll {
+        $powershell = Join-Path -Path $PsHome -ChildPath "pwsh"
+        $dirnames = "[T]est","[Test","T][est","Test"
+        foreach ( $d in $dirnames ) { new-item -type Directory -path "${TESTDRIVE}/$d" }
```
iSazonov: Please suppress output `new-item -type Directory -path "${TESTDRIVE}/$d" > $null`.
Please use common formatting style:
```powershell
 foreach ( $d in $dirnames ) {
    new-item -type Directory -path "${TESTDRIVE}/$d" > $null
}
```

test/powershell/Host/ConsoleHost.Tests.ps1
```diff
@@ -683,3 +683,24 @@ Describe "Pwsh exe resources tests" -Tag CI {
...
+        $testcases = foreach ( $d in $dirnames ) { @{ Dirname = $d } }
+    }
+
+    It "pwsh can startup in a directory named <dirname>" -testcases $testcases {
+        param ( $dirname )
+        try {
+            Push-Location -LiteralPath "${TESTDRIVE}/${dirname}"
+            $result = & $powershell -c '(Get-Item .).Name'
```
iSazonov: Should we use cmdlet? Maybe better .Net method to get current directory or -LiteralPath?

src/System.Management.Automation/engine/InitialSessionState.cs
```diff
@@ -3229,6 +3229,7 @@ internal static void SetSessionStateDrive(ExecutionContext context, bool setLoca
 
                         try
                         {
+                            providerContext.SuppressWildcardExpansion = true;
```
KillyMXI: (style) Maybe it's just my OCD, by why configuring providerContext inside try block rather than right after initialization?

test/powershell/Host/ConsoleHost.Tests.ps1
```diff
@@ -683,3 +683,24 @@ Describe "Pwsh exe resources tests" -Tag CI {
         $psversiontable.os | Should -MatchExactly "$($osversion.Major).$($osversion.Minor)"
     }
 }
+
+Describe 'Pwsh startup in directories that contain wild cards' -Tag CI {
+    BeforeAll {
+        $powershell = Join-Path -Path $PsHome -ChildPath "pwsh"
+        $dirnames = "[T]est","[Test","T][est","Test"
+        foreach ( $d in $dirnames ) { new-item -type Directory -path "${TESTDRIVE}/$d" }
+        $testcases = foreach ( $d in $dirnames ) { @{ Dirname = $d } }
```
adityapatwardhan: We iterate over `$dirname` twice, once for `new-item` and once for generating test cases. Should be do it in a single loop?

test/powershell/Host/ConsoleHost.Tests.ps1
JamesWTruher: sure, although this is slower than:
`$null = New-Item -type Directory -path "${TESTDRIVE}/$d"`

test/powershell/Host/ConsoleHost.Tests.ps1
JamesWTruher: it really doesn't matter, does it?

src/System.Management.Automation/engine/InitialSessionState.cs
JamesWTruher: I wanted the two operations to be visually close, putting the assignment in the block did that for me

test/powershell/Host/ConsoleHost.Tests.ps1
adityapatwardhan: I agree, maybe use 
```PowerShell
$result = & $powershell -c '([Environment]::CurrentDirectory)'
```

adityapatwardhan (dismissed):

test/powershell/Host/ConsoleHost.Tests.ps1
JamesWTruher: sure

test/powershell/Host/ConsoleHost.Tests.ps1
JamesWTruher: also, I want to check only the name,  ason some systems, Environment.CurrentDirectory will include a symlink which is a problem because PS does not, I'm I need to call split-path, I can use this form just as well.

BrucePay (approved):
LGTM

PaulHigin (approved):
Looking at these changes I realize that we really only need wait-forrunning function for the Job2 tests because they test job state and/or use job APIs with the assumption of the job running.  And also for the couple of other tests that wait for job state to go to running (with throttle limit).  Otherwise Wait-Job is used which will wait for an end state.  I don't think it hurts anything but it is not needed.

JamesWTruher:
@adityapatwardhan using the API approach will fail in some circumstances where `${TESTDRIVE}/...` in PowerShell is `/tmp/<guid>/...` but the API may return `/private/tmp/<guid>/...` as it follows the symlink. Testing the name of the directory is the easiest way to avoid the necessity of recognizing/constructing the "real" path. What I've done proves the behavior well enough.

adityapatwardhan:
@JamesWTruher Is the thread jobs test fix intended to be in this PR? If so could you update the title and description.

test/powershell/Host/ConsoleHost.Tests.ps1
iSazonov: My misunderstanding is that we use `Directory.GetCurrentDirectory()` in the code and `Set-Location` and `Get-Item` in the test - do we work with different current directories (first in .Net, second in FileSystem provider)?

anmenaga:
@JamesWTruher it looks like job test fix was moved to a separate PR #7270 (with some additions), so they can be excluded from this one.

daxian-dbw (approved):

ZeppLu:
Sorry to bother, but when will this fix be in Windows? I'm using PowerShell 5.1.17763.134 on Windows 1809, and issue 5752 still exists.

vexx32:
@ZeppLu the PS team have mentioned in previous issues and PRs that no more fixes are being backported to Windows PowerShell.

KillyMXI:
This fix was included into the version v6.1.0-preview.4 - 2018-07-19.
The question is, when the fresh enough version of PowerShell will be shipped with Windows.

ZeppLu:
@vexx32 Installing PowerShell Core does not seem a good solution for me, I'll try out other workarounds. Anyway, thanks for your explanation and advice.

Issue #6669, opened 2018-04-17
Using Enter-PSSession -ContainerId is hardcoded to powershell.exe


SteveL-MSFT:
The code to connect to a container is [hardcoded](https://github.com/PowerShell/PowerShell/blob/c1c5344a8897262433141ecbc13bb06ac2c4bbef/src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs#L3163) to start powershell.exe.  Need to expose a parameter to allow starting `pwsh`.

Also, `-ContainerId` requires full hash and not just the partial hash which is a usability problem.

Steps to reproduce
------------------

```powershell
docker run microsoft/windowsservercore
# get containerid
docker ps
enter-pssession -containerid <containerId>
```

Expected behavior
-----------------

```none
Ability to connect to PowerShell Core via PowerShell Direct
```

Actual behavior
---------------

```none
Can only connect to Windows PowerShell
```

Environment data
----------------

```powershell
Name                           Value
----                           -----
PSVersion                      6.1.0-preview.1
PSEdition                      Core
GitCommitId                    v6.1.0-preview.1
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

bergmeister:
The current cmdlet syntax is:
```powershell
Enter-PSSession [-ContainerId] <string> [-ConfigurationName <string>] [-RunAsAdministrator] [<CommonParameters>]
```
The proposal would be to add a switch `DoNotUseWindowsPowerShell` to be able to use PowerShell Core. I know, the naming does not sound good (double negative, etc), but such a switch parameter is more useful as existing scripts can't be broken by changing the default and having to incorporate the word `PowerShell Core` in a ValidateSet is also not good if it gets renamed back to just `PowerShell` in the future. I am happy to take any suggestions

MaximoTrinidad:
+1

That sounds interesting!  In the scenario of using winrm and having both PowerShell installed in Windows. I would see it practical to have a switch to pick which PowerShell to use.

:)

bergmeister:
Also, as far as I see, the same parameter would need to be added to the `Invoke-Command` cmdlet

SteveL-MSFT:
Closing this since #7241 provides a good enough solution

PR #7241, opened 2018-07-06
Update PowerShell Direct to try pwsh then fallback to powershell

### LINKED ISSUE #7237: Forward port PowerShell Direct into PS Core on Windows
In the Windows source tree, we implemented the ability for PowerShell Direct (i.e. PowerShell Remoting over Hyper-V sockets) to enter a VM or container with PowerShell Core (specifically pwsh.exe) instead of Windows PowerShell/powershell.exe.

Those PRs are tracked as 1376621 and 1174588, and they need to be forward-ported to PowerShell Core.

TravisEz13:
## PR Summary

Update PowerShell Direct to try pwsh then fallback to powershell
  - Forward port changes from Windows Powershell to fall back
  - change the order from powershell->pwsh to pwsh->powershell

fixes #7237

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
+++ b/src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
@@ -2956,61 +2956,61 @@ namespace System.Management.Automation.Runspaces
         private const uint FileNotFoundHResult = 0x80070002;
 
         // The list of executable to try in order
-        private static readonly string[] Executables = new string[]{"pwsh.exe","powershell.exe"};
+        private static readonly string[] Executables = new string[] { "pwsh.exe", "powershell.exe"};
 
         #endregion
 
         #region Properties
 
         /// <summary>
-        /// For Hyper-V container, it is Guid of utility VM hosting Hyper-V container.
+        /// Gets or Sets, for Hyper-V container, the Guid of utility VM hosting Hyper-V container.
         /// For Windows Server Container, it is empty.
         /// </summary>
         public Guid RuntimeId { get; set; }
 
         /// <summary>
-        /// OB root of the container.
+        /// Gets or sets the OB root of the container.
         /// </summary>
         public string ContainerObRoot { get; set; }
 
         /// <summary>
-        /// ID of the container.
+        /// Gets or sets the ID of the container.
         /// </summary>
         public string ContainerId { get; set; }
 
         /// <summary>
-        /// Process ID of the process created in container.
+        /// Gets or sets the process ID of the process created in container.
         /// </summary>
         internal int ProcessId { get; set; }
 
         /// <summary>
-        /// Whether the process in container should be launched as high privileged account
+        /// Gets or sets whether the process in container should be launched as high privileged account
         /// (RunAsAdmin being true) or low privileged account (RunAsAdmin being false).
         /// </summary>
         internal bool RunAsAdmin { get; set; } = false;
 
         /// <summary>
-        /// The configuration name of the container session.
+        /// Gets or sets the configuration name of the container session.
         /// </summary>
         internal string ConfigurationName { get; set; }
 
         /// <summary>
-        /// Whether the process in container has terminated.
+        /// Gets or sets whether the process in container has terminated.
         /// </summary>
         internal bool ProcessTerminated { get; set; } = false;
 
         /// <summary>
-        /// The error code.
+        /// Gets or sets the error code.
         /// </summary>
         internal uint ErrorCode { get; set; } = 0;
 
         /// <summary>
-        /// The error message for other errors.
+        /// Gets or sets the error message for other errors.
         /// </summary>
         internal string ErrorMessage { get; set; } = string.Empty;
 
         /// <summary>
-        /// The PowerShell executable being used to host the runspace.
+        /// Gets or sets the PowerShell executable being used to host the runspace.
         /// </summary>
         internal string Executable { get; set; } = string.Empty;
 
@@ -3223,7 +3223,7 @@ namespace System.Management.Automation.Runspaces
                     // expected to be PowerShell Core as it's inbox in the container.
                     // If `pwsh.exe` does not exist, fall back to `powershell.exe` which is Windows PowerShell.
                     //
-                    foreach(string executableToTry in Executables)
+                    foreach (string executableToTry in Executables)
                     {
                         cmd = GetContainerProcessCommand(executableToTry);
 
@@ -3237,8 +3237,10 @@ namespace System.Management.Automation.Runspaces
                         if (result == 0)
                         {
                             processId = Convert.ToInt32(ProcessInformation.ProcessId);
+
                             // Reset error to 0 in case this is not the first iteration of the loop.
                             error = 0;
+
                             // the process was started, exit the loop.
                             break;
                         }
@@ -3255,6 +3257,7 @@ namespace System.Management.Automation.Runspaces
                         {
                             processId = 0;
                             error = result;
+
                             // the executable was found but did not work
                             // exit the loop with the error state.
                             break;
@@ -3291,15 +3294,18 @@ namespace System.Management.Automation.Runspaces
         /// <summary>
         /// Get Command to launch container process based on instance properties.
         /// </summary>
+        /// <param name="executable">The name of the executable to use in the command.</param>
+        /// <returns>The command to launch the container process.</returns>
         private string GetContainerProcessCommand(string executable)
         {
             Executable = executable;
-            return string.Format(System.Globalization.CultureInfo.InvariantCulture,
+            return string.Format(
+                        System.Globalization.CultureInfo.InvariantCulture,
                         @"{{""CommandLine"": ""{0} {1} -NoLogo {2}"",""RestrictedToken"": {3}}}",
                         Executable, 
                         (RuntimeId != Guid.Empty) ? "-SocketServerMode -NoProfile" : "-NamedPipeServerMode",
-                        String.IsNullOrEmpty(ConfigurationName) ? String.Empty : String.Concat("-Config ", ConfigurationName),
-                        (RunAsAdmin) ? "false" : "true");
+                        string.IsNullOrEmpty(ConfigurationName) ? string.Empty : string.Concat("-Config ", ConfigurationName),
+                        RunAsAdmin ? "false" : "true");
         }
 
         /// <summary>
--- a/src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
+++ b/src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
@@ -2953,6 +2953,8 @@ namespace System.Management.Automation.Runspaces
         private const uint ContainersFeatureNotEnabled = 2;
         private const uint OtherError = 9999;
 
+        private const uint FileNotFoundHResult = 0x80070002;
+
         #endregion
 
         #region Properties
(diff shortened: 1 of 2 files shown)


bergmeister:
If I understand correctly, this partially resolves also #6669 but it also means this is a breaking change because the default is no longer Windows PowerShell?

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -3207,11 +3207,11 @@ private void CreateContainerProcessInternal()
                     //
                     // Hyper-V container (i.e., RuntimeId is not empty) uses Hyper-V socket transport.
                     // Windows Server container (i.e., RuntimeId is empty) uses named pipe transport for now.
-                    // This code executes `powershell.exe` as it exists in the container which currently is
-                    // expected to be Windows PowerShell as it's inbox in the container.
+                    // This code executes `pwsh.exe` as it exists in the container which currently is
+                    // expected to be PowerShell Core as it's inbox in the container.
                     //
                     cmd = string.Format(System.Globalization.CultureInfo.InvariantCulture,
```
dantraMSFT: Shouldn't this logic be in a separate function?

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -3223,14 +3223,35 @@ private void CreateContainerProcessInternal()
...
+                        else
+                        {
+                            processId = Convert.ToInt32(ProcessInformation.ProcessId);
+                        }
                     }
                     else
                     {
-                        processId = Convert.ToInt32(ProcessInformation.ProcessId);
+                        processId = 0;
+                        error = result;
```
dantraMSFT: I think you should set ErrorMessage here and in the failure path for failing to launch powershell.exe; otherwise, it won't be diagnosable (i.e., which path failed?  pwsh.exe or powershell.exe after not finding pwsh.exe)

dantraMSFT (changes requested):

BrucePay (commented):
Subject to Dan's comments, looks ok to me.

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -3223,14 +3225,32 @@ private void CreateContainerProcessInternal()
...
                     //
                     result = HcsCreateProcess(ComputeSystem, cmd, ref ProcessInformation, ref Process, ref resultString);
-                    if (result != 0)
+                    if (result == 0)
                     {
-                        processId = 0;
-                        error = result;
+                        processId = Convert.ToInt32(ProcessInformation.ProcessId);
+                    }
+                    else if (result == 2147942402) // 0x80070002 (2147942402) - The system cannot find the file specified.
```
SteveL-MSFT: Prefer to just have the number on RHS in hex as it's easier to perform a web search against a hex error code.  The comment can then be simplified.

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -3223,14 +3225,32 @@ private void CreateContainerProcessInternal()
...
+                    {
+                        // If we cannot find PowerShell Core, fall back to `powershell.exe` or Windows Powershell
+                        // We cannot do "cmd.exe /c powershell" because it returns process id of cmd instead powershell/pwsh.
+                        Executable = "powershell.exe";
+                        cmd = GetContainerProcessCommand();
+
+                        result = HcsCreateProcess(ComputeSystem, cmd, ref ProcessInformation, ref Process, ref resultString);
```
SteveL-MSFT: Seems like this can be simplified as a loop (having only one `HcsCreateProcess` and checking of the error).  Have `Executable` as an array containing `pwsh.exe`, `powershell.exe`.  Once process creation succeeds, break out of the loop, otherwise try the next executable.

SteveL-MSFT (changes requested):

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -3207,30 +3212,43 @@ private void CreateContainerProcessInternal()
...
+                        {
+                            // this executable was not find, try the next one
+                            // or exit the loop of none are left to try.
+                            continue;
+                        }
+                        else
+                        {
+                            processId = 0;
+                            error = result;
+                            // the executable was found but did not work
```
SteveL-MSFT: We should return the HRESULT within the exception message for future troubleshooting

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -3207,30 +3212,43 @@ private void CreateContainerProcessInternal()
...
+                        // Create PowerShell process inside the container.
+                        //
+                        result = HcsCreateProcess(ComputeSystem, cmd, ref ProcessInformation, ref Process, ref resultString);
+                        if (result == 0)
+                        {
+                            processId = Convert.ToInt32(ProcessInformation.ProcessId);
+                            // the process was started, exit the loop.
+                            break;
+                        }
+                        else if (result == 0x80070002) // 0x80070002 (2147942402) - The system cannot find the file specified.
```
SteveL-MSFT: remove the "0x80070002 (2147942402) -" as it's not needed now

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -3260,6 +3278,19 @@ private void CreateContainerProcessInternal()
...
 
+        /// <summary>
+        /// Get Command to launch container process based on instance properties.
+        /// </summary>
+        private string GetContainerProcessCommand()
+        {
+            return string.Format(System.Globalization.CultureInfo.InvariantCulture,
+                        @"{{""CommandLine"": ""{0} {1} -NoLogo {2}"",""RestrictedToken"": {3}}}",
+                        Executable, 
+                        (RuntimeId != Guid.Empty) ? "-so -NoProfile" : "-NamedPipeServerMode",
```
SteveL-MSFT: Can we expand `-so`?

SteveL-MSFT (changes requested):

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -3260,6 +3280,19 @@ private void CreateContainerProcessInternal()
             ErrorCode = error;
         }
 
+        /// <summary>
+        /// Get Command to launch container process based on instance properties.
+        /// </summary>
+        private string GetContainerProcessCommand()
+        {
+            return string.Format(System.Globalization.CultureInfo.InvariantCulture,
```
bergmeister: Using string interpolation and [FormattableString.Invariant](https://docs.microsoft.com/en-us/dotnet/api/system.formattablestring.invariant?view=netcore-2.1#System_FormattableString_Invariant_System_FormattableString_) would make this string construction easier to read:
```csharp
using static System.FormattableString;
private string foo()
{
   return Invariant($"bar {expression} {baz == null ? 1 : 0}");
}
```

dantraMSFT (approved):

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -3207,30 +3214,43 @@ private void CreateContainerProcessInternal()
...
+                    // This code executes `pwsh.exe` as it exists in the container which currently is
+                    // expected to be PowerShell Core as it's inbox in the container.
+                    // If `pwsh.exe` does not exist, fall back to `powershell.exe` which is Windows PowerShell.
                     //
-                    cmd = string.Format(System.Globalization.CultureInfo.InvariantCulture,
-                        @"{{""CommandLine"": ""powershell.exe {0} -NoLogo {1}"",""RestrictedToken"": {2}}}",
-                        (RuntimeId != Guid.Empty) ? "-so -NoProfile" : "-NamedPipeServerMode",
-                        String.IsNullOrEmpty(ConfigurationName) ? String.Empty : String.Concat("-Config ", ConfigurationName),
-                        (RunAsAdmin) ? "false" : "true");
+                    string[] executables = new string[]("pwsh.exe"."powershell.exe")
```
PaulHigin: I am unfamiliar with this syntax.  Shouldn't the string array initializer be?:
```
var executables = new string[] {"pwsh.exe", "powershell.exe"}
```
Also this can be a static readonly string array.

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -3260,6 +3280,19 @@ private void CreateContainerProcessInternal()
...
         }
 
+        /// <summary>
+        /// Get Command to launch container process based on instance properties.
+        /// </summary>
+        private string GetContainerProcessCommand()
+        {
+            return string.Format(System.Globalization.CultureInfo.InvariantCulture,
+                        @"{{""CommandLine"": ""{0} {1} -NoLogo {2}"",""RestrictedToken"": {3}}}",
+                        Executable,
```
PaulHigin: It feels cleaner if the Executable loop variable is passed in as a parameter.  I know we need the property set for error handling, but this code is a bit confusing without realizing Executable is a loop variable.

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -3207,30 +3214,43 @@ private void CreateContainerProcessInternal()
...
+                        // Create PowerShell process inside the container.
+                        //
+                        result = HcsCreateProcess(ComputeSystem, cmd, ref ProcessInformation, ref Process, ref resultString);
+                        if (result == 0)
+                        {
+                            processId = Convert.ToInt32(ProcessInformation.ProcessId);
+                            // the process was started, exit the loop.
+                            break;
+                        }
+                        else if (result == 0x80070002)
```
PaulHigin: Can we create a const variable for this expected error, with a descriptive name?

PaulHigin (changes requested):

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
```diff
@@ -3207,30 +3219,41 @@ private void CreateContainerProcessInternal()
...
+                        {
+                            processId = Convert.ToInt32(ProcessInformation.ProcessId);
+                            // the process was started, exit the loop.
+                            break;
+                        }
+                        else if (result == FileNotFoundHResult)
+                        {
+                            // "The system cannot find the file specified", try the next one
+                            // or exit the loop of none are left to try.
+                            continue;
```
PaulHigin: If both pwsh.exe and powershell.exe files are not found then this will exit without error.  It is unlikely but not impossible and I feel we should handle this error case so as not to change behavior.

PaulHigin (changes requested):

src/System.Management.Automation/engine/remoting/common/RunspaceConnectionInfo.cs
TravisEz13: Fixed, it is actually fairly likely as NanoServer containers don't have PowerShell by default if they are 1709 or newer.

PaulHigin (approved):
LGTM

SteveL-MSFT (approved):

TravisEz13:
Restarted CI due to https://github.com/PowerShell/PowerShell/issues/7260

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and the order makes sense.  There was some discussion about supporting pwsh-preview, however, we'll defer that until that is customer feedback requiring that change.

PR #7299, opened 2018-07-16
Set the cursor to the place where a user hits tab key

jianyunt:
## PR Summary

To address the issue https://github.com/PowerShell/PowerShell/issues/7022, the changes are among the following repos:
- PSReadLine, see PR https://github.com/lzybkr/PSReadLine/pull/708
  For example, in the cloudshell, when a user types `get-azurermvmac` and then hits `tab`, we reach out to Azure to fetch the data and pwsh will show progress in the meantime, which in turn changes the cursor position. That requires PSReadLine to adjust the cursor position after calling GetCompletions().
- This PR is for fixing the cursor position calculation issue in PSCore when scrolling up screen is necessary so the PSReadLine can set the cursor to the right location
- With the above two fixes, the nested progress still does not work. I had a long conversation with  @iSazonov in my previous PR  https://github.com/PowerShell/PowerShell/pull/7023. We cannot call hide() because there is a small window where the progress won't get cleaned if ctrl+c.  Looking through the code, I think the easier way is to use the workaround `$ProgressPreference ='SilentlyContinue' `  at the AzurePSDrive module level for nested case, see https://github.com/PowerShell/AzurePSDrive/pull/53.
 With that, we can still show the fetching Azure data progress written by SHiPS,  suppress the  progress in the Get-AzureRmWebApp , but it does not change the ProgressPreference setting in the cloudshell. This means when a user type Azure cmdlet, the progress will still show.  That is the behavior we want.

Please note that we cannot use `$ProgressPreference ='SilentlyContinue' for both SHiPS and its provider AzurePSDrive because fetching data from Azure can take really long time. Showing progress is needed but keep (try keeping) one progress only. Thus the above fixes are all needed.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ProgressPane.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ProgressPane.cs
@@ -101,8 +101,6 @@ namespace Microsoft.PowerShell
                 int scrollRows = rows - ((_rawui.BufferSize.Height - 1) - _location.Y);
                 if (scrollRows > 0)
                 {
-                    // The following can be possibly replaced by Console.Write ("\x1b[" + scrollRows + "S");
-                    // For details, see https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences
                     // Scroll the console screen up by 'scrollRows'
                     var bottomLocation = _location;
                     bottomLocation.Y = _rawui.BufferSize.Height;
--- a/src/Microsoft.PowerShell.ConsoleHost/host/msh/ProgressPane.cs
+++ b/src/Microsoft.PowerShell.ConsoleHost/host/msh/ProgressPane.cs
@@ -99,12 +99,19 @@ namespace Microsoft.PowerShell
 
                 //if the cursor is at the bottom, create screen buffer space by scrolling
                 int scrollRows = rows - ((_rawui.BufferSize.Height - 1) - _location.Y);
-                for (int i = 0; i < rows; i++)
-                {
-                    Console.Out.Write('\n');
-                }
                 if (scrollRows > 0)
                 {
+                    // The following can be possibly replaced by Console.Write ("\x1b[" + scrollRows + "S");
+                    // For details, see https://docs.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences
+                    // Scroll the console screen up by 'scrollRows'
+                    var bottomLocation = _location;
+                    bottomLocation.Y = _rawui.BufferSize.Height;
+                    _rawui.CursorPosition = bottomLocation;
+                    for (int i = 0; i < scrollRows; i++)
+                    {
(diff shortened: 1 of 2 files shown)


src/Microsoft.PowerShell.ConsoleHost/host/msh/ProgressPane.cs
```diff
@@ -99,12 +99,19 @@ class ProgressPane
 
                 //if the cursor is at the bottom, create screen buffer space by scrolling
                 int scrollRows = rows - ((_rawui.BufferSize.Height - 1) - _location.Y);
-                for (int i = 0; i < rows; i++)
-                {
-                    Console.Out.Write('\n');
-                }
                 if (scrollRows > 0)
                 {
+                    // The following can be possibly replaced by Console.Write ("\x1b[" + scrollRows + "S");
```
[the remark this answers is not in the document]
jianyunt: Removed the comment. Thanks.

iSazonov:
@jianyunt I do not have the ability to check this on Unix, but definitely we have to do many _manual_ tests to make sure everything works without regression.

BrucePay (approved):
LGTM

src/System.Management.Automation/engine/MshCommandRuntime.cs
```diff
@@ -2254,7 +2254,7 @@ internal void ThrowIfWriteNotPermitted(bool needsToWriteToPipeline)
             {
                 // Only generate these exceptions if a pipeline has already been declared as the 'writing' pipeline.
                 // Otherwise, these are probably infrastructure messages and can be ignored.
-                if (this.PipelineProcessor._permittedToWrite != null)
+                if (this.PipelineProcessor?._permittedToWrite != null)
```
adityapatwardhan: Could you add comment why did change was made?

adityapatwardhan (changes requested):

src/System.Management.Automation/engine/MshCommandRuntime.cs
jianyunt: The [line 2249](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/MshCommandRuntime.cs#L2249) has 

if (this.PipelineProcessor == null  || _thisCommand != this.PipelineProcessor._permittedToWrite ...

When I debug the code, i saw bunch of nullref exceptions. 
Yes the change has nothing to do with the cursor position. But it's good to get rid of it.

jianyunt: @adityapatwardhan In this case I don't think any comments are needed.

adityapatwardhan (approved):

anmenaga (approved):

anmenaga:
@adityapatwardhan This looks ready for merge.

PR #7308, opened 2018-07-18
Convert Newtonsoft.Json metadata properties in ConvertFrom-Json

### LINKED ISSUE #7307: ConvertFrom-Json does not convert Newtonsoft.Json specific metadata properties in root object
Steps to reproduce
------------------

```powershell
> '{"$id": 1, "$type": "System.String", "id": "value"}' | ConvertFrom-Json
```
1. Add `$id` and `$type` properties to a json object
2. Pipe object to `ConvertFrom-Json` Cmdlet
3. Notice that `$id` and `$type` properties are missing from resulting `PSCustomObject`

Expected behavior
-----------------
`ConvertFrom-Json` does not ignore properties when deserializing into generic `PSCustomObject`, regardless of their name.
```powershell
> '{"$id": 1, "$type": "System.String", "id": "value"}' | ConvertFrom-Json

$id $type         id
--- -----         --
  1 System.String value
```

Actual behavior
---------------

```powershell
> '{"$id": 1, "$type": "System.String", "id": "value"}' | ConvertFrom-Json

id
--
value
```

Additional Notes
---------------
- Issue does not occur when metadata properties are deeper in the hierarchy.
- Other `Newtonsoft.Json` metadata properties include `$values` and `$ref`.

Environment data
----------------

```powershell
> $PSVersionTable

PSVersion                      6.1.0-preview.3
GitCommitId                    v6.1.0-preview.3
```

adamgauthier:
## PR Summary
Use `MetadataPropertyHandling.Ignore` in `JsonSerializerSettings` when Deserializing instead of the default (`MetadataPropertyHandling.Default`). Documentation on this option can be found [here](https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_JsonSerializerSettings_MetadataPropertyHandling.htm) and [here](https://www.newtonsoft.com/json/help/html/T_Newtonsoft_Json_MetadataPropertyHandling.htm). Basically `Newtonsoft.Json` specific metadata properties like `$id`/`$type`/`$ref` are by default read and used by `Newtonsoft.Json` to deserialize into types, use the same instance for different properties, etc.

`ConvertFrom-Json` deserializes into the generic type `PSCustomObject`, I *think* it doesn't make sense that these specific properties would not be converted as they are part of the json object. This restrict people from using these properties if they are not using `Newtonsoft.Json`, but even more likely, it restrict them from manipulating the entire raw json data of an object serialized with the library (which is probably what they want to do, otherwise they would be using the library directly to deserialize).

I don't know how "breaking" of a change this is considered to be, I would assume not many people rely on the current behavior, but it is changing with this PR.

Fixes #7307

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/JsonObject.cs
@@ -73,7 +73,14 @@ namespace Microsoft.PowerShell.Commands
                     // we just continue the deserialization.
                 }
 
-                obj = JsonConvert.DeserializeObject(input, new JsonSerializerSettings() { TypeNameHandling = TypeNameHandling.None, MaxDepth = 1024 });
+                obj = JsonConvert.DeserializeObject(
+                    input,
+                    new JsonSerializerSettings
+                    {
+                        TypeNameHandling = TypeNameHandling.None,
+                        MetadataPropertyHandling = MetadataPropertyHandling.Ignore,
+                        MaxDepth = 1024
+                    });
 
                 // JObject is a IDictionary
                 var dictionary = obj as JObject;
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertFrom-Json.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertFrom-Json.Tests.ps1
@@ -1,7 +1,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 Describe 'ConvertFrom-Json' -tags "CI" {
-    
+
     BeforeAll {
         $testCasesWithAndWithoutAsHashtableSwitch = @(
             @{ AsHashtable = $true  }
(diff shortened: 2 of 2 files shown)


dantraMSFT:
@louistio Would you mind rewording the title and the description for this PR.  The associated issue states that the metaproperties are not being converted yet the PR title appears to state that they should be ignored.

Thanks,

louistio:
@dantraMSFT Yes my bad, it is confusing, this PR stated to ignore them as in to tell the serializer not to do anything special with them, not to ignore them as in not convert them. I changed the title now.

iSazonov:
@louistio Seems the header should be "Don't convert ..."

iSazonov (approved):
Leave a comment

markekraus (approved):
My comments are not blocking. Otherwise LGTM

markekraus:
Upon reviewing the issue, I do not believe this to be a breaking change as this appears to be a regression from 5.1. The behavior in 5.1 was to serialize the meta properties as object properties which is now broekn in 6.0. This PR corrects that regression and IMO is not a breaking change.

louistio:
@iSazonov I think there is confusion about the naming. The issue is that some values are not converted from `json` to `PSCustomObject`, because `Newtonsoft.Json` reads them and thinks they are metadata properties only useful for deserialization. This change makes `Newtonsoft.Json` "ignore" those properties so they are **converted** correctly from `json` to `PSCustomObject`. I'm not sure how naming this PR "Don't convert" would make much sense.

iSazonov:
@dantraMSFT Do you have any comments?

iSazonov:
@louistio Thanks for your contribution!

PR #7329, opened 2018-07-21
Fix formatting, vt100 support, help uri and positional parameter issues for Markdown cmdlets

adityapatwardhan:
Fix #7283 

## PR Summary

* Formatting changes moved from types file to formatting file.
* Add logic to check VT100 support before adding VT100 escape sequences.
* Add online help uris.
* Add positional parameters for Convert-FromMarkdown and Set-MarkdownOption

## PR Checklist

- **User-facing changes**
    - [X] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
@@ -57,23 +57,24 @@ namespace Microsoft.PowerShell.Commands
         private const string LiteralPathParameterSet = "LiteralParamSet";
         private const string InputObjParamSet = "InputObjParamSet";
         private MarkdownConversionType conversionType = MarkdownConversionType.HTML;
-        private MarkdownOptionInfo mdOption = null;
+        private PSMarkdownOptionInfo mdOption = null;
 
         /// <summary>
-        /// Read the MarkdownOptionInfo set in SessionState.
+        /// Read the PSMarkdownOptionInfo set in SessionState.
         /// </summary>
         protected override void BeginProcessing()
         {
-            mdOption = SessionState.PSVariable.GetValue("MarkdownOptionInfo", new MarkdownOptionInfo()) as MarkdownOptionInfo;
-
-            if (!this.Host.UI.SupportsVirtualTerminal)
-            {
-                mdOption.EnableVT100Encoding = false;
-            }
+            mdOption = SessionState.PSVariable.GetValue("PSMarkdownOptionInfo", new PSMarkdownOptionInfo()) as PSMarkdownOptionInfo;
 
             if (mdOption == null)
             {
                 throw new InvalidOperationException();
+            }
+
+            if (this.Host != null && this.Host.UI.SupportsVirtualTerminal)
+            {
+                // If EnableVT100Encoding is true and SupportsVirtualTerminal is true then enable, else disable.
+                mdOption.EnableVT100Encoding = mdOption.EnableVT100Encoding && this.Host.UI.SupportsVirtualTerminal;
             }
 
             if (AsVT100EncodedString)
@@ -128,7 +129,7 @@ namespace Microsoft.PowerShell.Commands
             }
         }
 
-        private void ConvertEachFile(IEnumerable<string> paths, MarkdownConversionType conversionType, bool isLiteral, MarkdownOptionInfo optionInfo)
+        private void ConvertEachFile(IEnumerable<string> paths, MarkdownConversionType conversionType, bool isLiteral, PSMarkdownOptionInfo optionInfo)
         {
             foreach (var path in paths)
             {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
@@ -75,7 +75,7 @@ namespace Microsoft.PowerShell.Commands
 
             // supportsVT100 == null if the host is null.
             // supportsVT100 == false if host does not support VT100.
-            if (supportsVT100 == null || supportsVT100 == false)
+            if (supportsVT100 != true)
             {
                 mdOption.EnableVT100Encoding = false;
             }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
@@ -64,7 +64,7 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         protected override void BeginProcessing()
         {
-            mdOption = SessionState.PSVariable.GetValue("PSMarkdownOptionInfo", new PSMarkdownOptionInfo()) as PSMarkdownOptionInfo;
+            mdOption = this.CommandInfo.Module.SessionState.PSVariable.GetValue("PSMarkdownOptionInfo", new PSMarkdownOptionInfo()) as PSMarkdownOptionInfo;
 
             if (mdOption == null)
             {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
@@ -36,6 +36,7 @@ namespace Microsoft.PowerShell.Commands
         /// Gets or sets the path to the file to convert from Markdown to MarkdownInfo.
         /// </summary>
         [ValidateNotNullOrEmpty]
+        [Alias("PSPath", "LP")]
         [Parameter(ParameterSetName = LiteralPathParameterSet, Mandatory = true)]
         public string[] LiteralPath { get; set; }
 
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
@@ -21,7 +21,7 @@ namespace Microsoft.PowerShell.Commands
     [Cmdlet(
         VerbsData.ConvertFrom, "Markdown",
         DefaultParameterSetName = PathParameterSet,
-        HelpUri = "TBD")]
+        HelpUri = "https://go.microsoft.com/fwlink/?linkid=2006503")]
     [OutputType(typeof(Microsoft.PowerShell.MarkdownRender.MarkdownInfo))]
     public class ConvertFromMarkdownCommand : PSCmdlet
     {
@@ -29,7 +29,7 @@ namespace Microsoft.PowerShell.Commands
         /// Gets or sets path to the file to convert from Markdown to MarkdownInfo.
         /// </summary>
         [ValidateNotNullOrEmpty]
-        [Parameter(ParameterSetName = PathParameterSet, Mandatory = true)]
+        [Parameter(ParameterSetName = PathParameterSet, Mandatory = true, Position = 0)]
         public string[] Path { get; set; }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


src/Microsoft.PowerShell.MarkdownRender/VT100EscapeSequences.cs
```diff
@@ -295,7 +308,14 @@ public string FormatHeader1(string headerText)
         /// <returns>Formatted Header 2 string.</returns>
         public string FormatHeader2(string headerText)
         {
-            return string.Concat(Esc, options.Header2, headerText, endSequence);
+            if (options.EnableVT100Encoding)
```
SteveL-MSFT: There's a lot of similar code, perhaps have a helper that does the `EnableVT100Encoding` check and concats or just returns the string?

SteveL-MSFT (changes requested):

iSazonov:
Usually we use
```c#
[Alias("PSPath", "LP")]
```
for LiteralPath parameters.

iSazonov (commented):
Leave a comment

rkeithhill:
👍 for positional parameter support. When `ConvertFrom-Markdown .\README.md` generated an error, I was bummed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
```diff
@@ -65,6 +65,11 @@ protected override void BeginProcessing()
         {
             mdOption = SessionState.PSVariable.GetValue("MarkdownOptionInfo", new MarkdownOptionInfo()) as MarkdownOptionInfo;
```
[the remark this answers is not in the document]
adityapatwardhan: I prefer `MarkdownOptionInfo` since the cmdlet is `Get-MarkdownOption`. @daxian-dbw  your thoughts?

src/Microsoft.PowerShell.MarkdownRender/VT100EscapeSequences.cs
adityapatwardhan: Fixed for header types. For others the code is slightly different for each so kept it as is.

adityapatwardhan:
@SteveL-MSFT  please have another look.

PaulHigin (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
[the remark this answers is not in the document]
SteveL-MSFT: I think for any variables that are part of PowerShell itself, it should have the PS prefix to avoid name collision.  So this should be `PSMarkdownOptionInfo`.  cc @joeyaiello

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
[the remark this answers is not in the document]
daxian-dbw: The variable `MarkdownOptionInfo` should be saved in the module scope of the Utility module, instead of the current scope of the current session state. Doing the latter will make the persistence of markdown option unreliable/inconsistent.

And, did you consider other options to persist the markdown option? I don't have an alternative off the top of my mind, but would like to check if you considered anything else.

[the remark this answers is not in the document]
joeyaiello: @SteveL-MSFT if we agree it should be moved to the module state, does it matter? 

As for persistence, I think it's okay to tell people to put this stuff in their profile until we hear why that's not good enough (seems to work just fine for PSReadline).

SteveL-MSFT:
On macOS using the Terminal app, it doesn't seem to display the VT100 inverted sequence so the text becomes unreadable since the color is supposed to be black on inverted background but just appears as black.  A basic web search doesn't find anything on how to support this so we may need to special case macOS to use some other colors.

SteveL-MSFT:
I take back part of what I said.  Inverse works for Headers like

```markdown
# header
```

But this only renders as black text on black background

```markdown
`text`
```

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
```diff
@@ -65,6 +66,11 @@ protected override void BeginProcessing()
...
 
+            if (!this.Host.UI.SupportsVirtualTerminal)
```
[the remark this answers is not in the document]
adityapatwardhan: Fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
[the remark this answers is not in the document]
adityapatwardhan: @daxian-dbw  The committee feedback was to have it in session state so it is valid throughout the session. The preferences can be set in profile for persistence.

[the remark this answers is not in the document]
daxian-dbw: I'm OK to have it in session state, but it should be the module session state of `Microsoft.PowerShell.Utility` instead of the currently active session state, which could be the global session state or another module's session state. And the scope to save the variable should be the module scope, not the current scope.

For example, if I'm running `Set-MarkdownOptions` in a function from another module, then the variable will be saved in the function scope under the session state of that module, which will be lost once the function finishes execution.

adityapatwardhan:
@PaulHigin @SteveL-MSFT @daxian-dbw  Please have another look.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
```diff
@@ -57,23 +57,24 @@ public class ConvertFromMarkdownCommand : PSCmdlet
...
+        private PSMarkdownOptionInfo mdOption = null;
 
         /// <summary>
-        /// Read the MarkdownOptionInfo set in SessionState.
+        /// Read the PSMarkdownOptionInfo set in SessionState.
         /// </summary>
         protected override void BeginProcessing()
         {
-            mdOption = SessionState.PSVariable.GetValue("MarkdownOptionInfo", new MarkdownOptionInfo()) as MarkdownOptionInfo;
+            mdOption = SessionState.PSVariable.GetValue("PSMarkdownOptionInfo", new PSMarkdownOptionInfo()) as PSMarkdownOptionInfo;
```
daxian-dbw: Same comment as in https://github.com/PowerShell/PowerShell/pull/7329#discussion_r204597710

daxian-dbw (changes requested):

src/Microsoft.PowerShell.MarkdownRender/VT100EscapeSequences.cs
```diff
@@ -204,6 +205,8 @@ public MarkdownOptionInfo()
         private const string Header5Dark = "[4;96m";
         private const string Header6Dark = "[4;97m";
         private const string CodeDark = "[48;2;155;155;155;38;2;30;30;30m";
+
+        private const string CodeMacOS = "[107;95m";
```
SteveL-MSFT: Should add a comment why we are special casing macOS

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs
```diff
@@ -56,20 +57,26 @@ public class ConvertFromMarkdownCommand : PSCmdlet
...
 
             if (mdOption == null)
             {
                 throw new InvalidOperationException();
             }
 
+            if (this.Host != null && this.Host.UI.SupportsVirtualTerminal)
```
PaulHigin: `this.Host.UI.SupportsVirtualTerminal` is not needed twice here.  But it seems like the logic should be:
```
mdOption.EnableVT100Encoding = mdOption.EnableVT100Encoding && (this.Host != null) && (this.Host.UI.SupportsVirtualTerminal);

If there is no host I assume EnableVT100Encoding option should be disabled?

**Review by PaulHigin — CHANGES_REQUESTED** [review]

File: src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs (hunk 6 above)
Comment (adityapatwardhan) [reply src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs]: @PaulHigin improved it a bit by using nullable bool.

File: src/Microsoft.PowerShell.MarkdownRender/VT100EscapeSequences.cs (hunk 5 above)
Comment (adityapatwardhan) [reply src/Microsoft.PowerShell.MarkdownRender/VT100EscapeSequences.cs]: Added comment.

File: src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs (hunk 4 above)
Comment (adityapatwardhan) [reply src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs]: Fixed.

**Comment by adityapatwardhan** [comment]
@PaulHigin @SteveL-MSFT @daxian-dbw  Please have a look.

File: src/Microsoft.PowerShell.Commands.Utility/commands/utility/ConvertFromMarkdownCommand.cs [hunk 7]
```diff
@@ -71,10 +71,13 @@ protected override void BeginProcessing()
                 throw new InvalidOperationException();
             }
 
-            if (this.Host != null && this.Host.UI.SupportsVirtualTerminal)
+            bool? supportsVT100 = this.Host?.UI.SupportsVirtualTerminal;
+
+            // supportsVT100 == null if the host is null.
+            // supportsVT100 == false if host does not support VT100.
+            if (supportsVT100 == null || supportsVT100 == false)
```
PaulHigin: Nit.  I think you can just do `if (supportsVT100 != true)` here.

PaulHigin (approved):
LGTM

SteveL-MSFT (approved):

SteveL-MSFT:
@adityapatwardhan can you review the CodeFactor issues?

adityapatwardhan:
@SteveL-MSFT I have fixed a few CodeFactor issues. Others are either following a pattern which already exists or a rule we want to disable.

@daxian-dbw  Can you have another look?

test/powershell/Modules/Microsoft.PowerShell.Utility/MarkdownCmdlets.Tests.ps1
```diff
@@ -375,6 +375,34 @@ bool function()`n{`n}
...
+            $options.Link | Should -BeExactly "[4;38;5;117m"
+            $options.Image | Should -BeExactly "[33m"
+            $options.EmphasisBold | Should -BeExactly "[1m"
+            $options.EmphasisItalics | Should -BeExactly "[36m"
+        }
+
+        It "Verify PSMarkdownOptionInfo is defined in module scope" {
+
+            $mod = Get-Module Microsoft.PowerShell.Utility
+            $options =  & $mod { $PSMarkdownOptionInfo }
```
daxian-dbw: Better add one check for `$PSMarkdownOptionInfo -eq $null` before this line, to make sure that the variable is not leaked into the global scope or current scope.

daxian-dbw: Better add one check for `$PSMarkdownOptionInfo -eq $null` before this line, to make sure that the variable is not leaked into the global scope or current scope.

daxian-dbw (changes requested):
One more comment. Otherwise look good to me.

SteveL-MSFT (approved):

test/powershell/Modules/Microsoft.PowerShell.Utility/MarkdownCmdlets.Tests.ps1
adityapatwardhan: Fixed.

daxian-dbw (approved):

TravisEz13 (approved):

iSazonov:
@daxian-dbw Is your https://github.com/PowerShell/PowerShell/pull/7329#discussion_r205183077 closed?

daxian-dbw:
@iSazonov Yes, it has been changed to use the module scope to save the variable.

PR #7347, opened 2018-07-24
Throw error message when MaxumimVersion is less than Version when using Import-Module cmdlet with -FullyQualifiedName parameter.

### LINKED ISSUE #7337: Get-Module accepts nonsensical combinations of ModuleVersion and MaximumVersion entries for -FullyQualifiedName
Steps to reproduce
------------------

Request a minimum version of 1.6.0 and - nonsensically - a maximum version of 1.5.0:

```powershell
Get-Module -ListAvailable -FullyQualifiedName @{ ModuleName='PowerShellGet'; ModuleVersion='1.6.0'; MaximumVersion='1.5.0' }
```

Expected behavior
-----------------

I think a statement-terminating error indicating that `MaximumVersion`  must be greater or equal to `ModuleVersion` (which uses _minimum_ logic) would be more helpful than the current behavior.

If feasible, this validation should be tied to fully qualified module names themselves, wherever they are accepted.

Actual behavior
---------------

The argument is quietly accepted and therefore results in no output (because, by definition, no modules match).

Environment data
----------------

```powershell
PowerShell Core v6.1.0-preview.3 on macOS 10.13.6
PowerShell Core v6.1.0-preview.3 on Ubuntu 16.04.4 LTS
PowerShell Core v6.1.0-preview.3 on Microsoft Windows 10 Pro (64-bit; Version 1803, OS Build: 17134.165)
\Windows PowerShell v5.1.17134.165 on Microsoft Windows 10 Pro (64-bit; Version 1803, OS Build: 17134.165)
```

sethvs:
## PR Summary

Fix #7337
 
Also fixing following error message to include 'MaximumVersion' in the list of 'valid members':
```
  <data name="InvalidModuleSpecificationMember" xml:space="preserve">
    <value>The hashtable describing a module contains one or more members that are not valid.  The valid members are ({0}). Remove the members that are not valid ({1}), then try again.</value>
  </data>
```
Concerning if the change is breaking, I would refer it to Unlikely Grey Area.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**


src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
```diff
@@ -136,7 +136,7 @@ internal static Exception ModuleSpecificationInitHelper(ModuleSpecification modu
             string message;
             if (badKeys.Length != 0)
             {
-                message = StringUtil.Format(Modules.InvalidModuleSpecificationMember, "ModuleName, ModuleVersion, RequiredVersion, GUID", badKeys);
+                message = StringUtil.Format(Modules.InvalidModuleSpecificationMember, "ModuleName, ModuleVersion, MaximumVersion, RequiredVersion, GUID", badKeys);
```
anmenaga: Please add to PR Summary:
```
Also fixing following error message to include 'MaximumVersion' in the list of 'valid members':
  <data name="InvalidModuleSpecificationMember" xml:space="preserve">
    <value>The hashtable describing a module contains one or more members that are not valid.  The valid members are ({0}). Remove the members that are not valid ({1}), then try again.</value>
  </data>
```

anmenaga:
@daxian-dbw @SteveL-MSFT please evaluate if this technically breaking (however unlikely) change has to go to PS Committee for review.

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
sethvs: Done.

SteveL-MSFT:
@PowerShell/powershell-committee agrees this is a breaking change.  For example using this in a pipeline would terminate the pipeline now whereas a $null being returned would not execute the rest of the pipeline.  We agree that this should have been an error and would accept returning a non-terminating error and not a terminating one.

sethvs:
Can you guide me a little bit?
How can I use WriteError in there?
https://github.com/PowerShell/PowerShell/blob/321537c6e52b24e9187afd267bb0b18b242bbed3/src/System.Management.Automation/engine/Modules/ModuleSpecification.cs#L167-L172

Or is there another way to return non-terminating error?

SteveL-MSFT:
@sethvs I took a look at the code.  Since the validation is in the ModuleSpecification class that should continue to return an exception that gets thrown.  You'll probably need to update GetModuleCommand.cs to handle this specific case.

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
rjmholt: Ideally this should use `$"{nameof(...)}"` for the bad members

src/System.Management.Automation/resources/Modules.resx
```diff
@@ -165,6 +165,9 @@
   <data name="InvalidModuleSpecificationMember" xml:space="preserve">
     <value>The hashtable describing a module contains one or more members that are not valid.  The valid members are ({0}). Remove the members that are not valid ({1}), then try again.</value>
   </data>
+  <data name="ModuleSpecificationMemberIsLessThanOther" xml:space="preserve">
```
rjmholt: The error message should ideally be slightly more specific. Something like:
> The hashtable describing a module cannot have its ModuleVersion ({0}) greater than its MaximumVersion ({1}).

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
```diff
@@ -56,6 +56,10 @@ Describe "Import-Module" -Tags "CI" {
         Import-Module TestModule -RequiredVersion 1.1
         (Get-Module TestModule).Version | Should -BeIn "1.1"
     }
+
+    It "should throw if 'MaximumVersion' is less than 'Version' when using -FullyQualifiedName parameter" {
```
rjmholt: In addition to this test, we should add a more specific one using the `[ModuleSpecification]` constructor.

You should find those tests [here](https://github.com/PowerShell/PowerShell/blob/master/test/powershell/engine/Module/ModuleSpecification.Tests.ps1) (I know that's not an obvious place at all).

Note to self: once https://github.com/PowerShell/PowerShell/pull/7499 is merged, this test will be included in that.

rjmholt (commented):
Apologies, I should have reviewed this sooner. Looks like a good change! Just left a couple of minor comments.

Thanks for your contribution @sethvs!

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
rjmholt: Let me know if you need any help on creating that test btw, I'm always happy to

rjmholt:
@SteveL-MSFT is it possible for a parameter binding exception to be non-terminating? The error when called with `Get-Module` will occur when the hashtable is cast to a `ModuleSpecification`; execution will never reach `Get-Module`.

rjmholt:
Also for reference, here is `Import-Module`'s version check:
https://github.com/PowerShell/PowerShell/blob/c94fc31f9767fe5bc835ee4921896b6331eaad4d/src/System.Management.Automation/engine/Modules/ImportModuleCommand.cs#L1686-L1690

I think that's a terminating error, no?

src/System.Management.Automation/engine/Modules/ModuleSpecification.cs
sethvs: nameof(badKeys) returns 'badKeys', but we need its actual value.
Or am I missing something?

src/System.Management.Automation/resources/Modules.resx
sethvs: Done.

test/powershell/Modules/Microsoft.PowerShell.Core/Import-Module.Tests.ps1
sethvs: Done.

sethvs:
Is this OK with VSTS: PowerShell-CI-macos?

anmenaga (approved):

sethvs:
Close and reopen to restart tests.

SteveL-MSFT:
@PowerShell/powershell-committee re-reviewed this.  The conclusion is that well-formed filters should return nothing if the result set is nothing and not return an error.  In this case, the filter is well-formed, although invalid.  Since this is also a breaking change, our recommendation is to not take this change as it's inconsistent with the desired filtering behavior and introduces a breaking change.

PR #7366, opened 2018-07-26
Update error message when using an auto-discovered module that is not compatible with current PSEdition

### LINKED ISSUE #7352: Improve error message when autoloaded module fails the PSCompatibleEditions check
Module auto-loading knows about the existence of incompatible commands, even when `Get-Command` does not, so when you try to execute a command in an Edition-incompatible module (without loading it first), it will complain about being unable to load the module. 

Also note that `Get-Command Add-WindowsCapability` returns nothing here.

We may also want to extend `Get-Command` to support `-SkipEditionCheck`, but I'll file that issue as this one plays out.

Steps to reproduce
------------------

```powershell
C:\WINDOWS\system32> Add-WindowsCapability
```

Expected behavior
-----------------

One of these two:

```none
Add-WindowsCapability : The 'Add-WindowsCapability' command was found in the module 'Dism', but 'Dism' does not support current PowerShell edition 'Core'. Its supported editions are 'Desktop'. Use 'Import-Module -SkipEditionCheck 'Dism'' to ignore the compatibility of this module.
At line:1 char:1
+ Add-WindowsCapability
+ ~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : ObjectNotFound: (Add-WindowsCapability:String) [], CommandNotFoundException
+ FullyQualifiedErrorId : CouldNotAutoloadMatchingModule
```

less desirable, but still acceptable:

```none
Add-WindowsCapability : The term 'Add-WindowsCapability' is not recognized as the name of a cmdlet, function, script file, or operable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
At line:1 char:1
+ Add-WindowsCapability
+ ~~~~~
+ CategoryInfo          : ObjectNotFound: (Add-WindowsCapability:String) [], CommandNotFoundException
+ FullyQualifiedErrorId : CommandNotFoundException
```

Actual behavior
---------------

```none
Add-WindowsCapability : The 'Add-WindowsCapability' command was found in the module 'Dism', but the module could not be loaded. For more information, run 'Import-Module Dism'.
At line:1 char:1
+ Add-WindowsCapability
+ ~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : ObjectNotFound: (Add-WindowsCapability:String) [], CommandNotFoundException
+ FullyQualifiedErrorId : CouldNotAutoloadMatchingModule
```

Environment data
----------------

```powershell
> $PSVersionTable

PSVersion                      6.1.0-preview.4
GitCommitId                    6.1.0-preview.4
```

SteveL-MSFT:
## PR Summary

Update error message when using module auto-discovery to more closely resemble the error from import-module.

Fix https://github.com/PowerShell/PowerShell/issues/7352

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/resources/DiscoveryExceptions.resx
+++ b/src/System.Management.Automation/resources/DiscoveryExceptions.resx
@@ -204,7 +204,7 @@ The #requires statement must be in one of the following formats:
     <value>The script '{0}' cannot be run because the following modules that are specified by the "#requires" statements of the script are missing: {1}.</value>
   </data>
   <data name="CouldNotAutoImportMatchingModule" xml:space="preserve">
-    <value>The '{0}' command was found in the module '{1}', but the module could not be loaded. For more information, run 'Import-Module {1}'.</value>
+    <value>The '{0}' command was found in the module '{1}', but '{1}' does not support current PowerShell edition 'Core'. Use 'Import-Module -SkipEditionCheck '{1}''' to ignore the compatibility of this module.</value>
   </data>
   <data name="CouldNotAutoImportModule" xml:space="preserve">
     <value>The module '{0}' could not be loaded. For more information, run 'Import-Module {0}'.</value>


src/System.Management.Automation/resources/DiscoveryExceptions.resx
```diff
@@ -204,7 +204,7 @@ The #requires statement must be in one of the following formats:
     <value>The script '{0}' cannot be run because the following modules that are specified by the "#requires" statements of the script are missing: {1}.</value>
   </data>
   <data name="CouldNotAutoImportMatchingModule" xml:space="preserve">
-    <value>The '{0}' command was found in the module '{1}', but the module could not be loaded. For more information, run 'Import-Module {1}'.</value>
+    <value>The '{0}' command was found in the module '{1}', but '{1}' does not support current PowerShell edition 'Core'. Use 'Import-Module -SkipEditionCheck '{1}''' to ignore the compatibility of this module.</value>
```
daxian-dbw: This error message is too specific. It could be the module is compatible with PSCore but has a problem that prevents it from being loaded. And, I think @rjmholt has made the changes to prevent a "desktop" module being cached for command discovery, so commands from a "desktop" module shouldn't be discoverable. If you find otherwise, we should open an issue to fix the root cause as soon as possible (before the user's command discovery cache gets further polluted).

daxian-dbw (changes requested):

src/System.Management.Automation/resources/DiscoveryExceptions.resx
SteveL-MSFT: That's a good point.  Only running `Import-Module` will you get the detailed error message.  @joeyaiello you ok with keeping the current error?

anmenaga: @daxian-dbw what is the behavior of auto-discovered modules that don't have PSEdition specified?

daxian-dbw: If a module lives in System32 module path and it doesn't have PSEdition specified, then it should not be discoverable to PowerShell.

For modules out of System32 module path, they are all considered compatible to PSCore, and thus all discoverable, regardless of PSEdition.

rjmholt: I'm late to this, but unless we are certain this error can only be thrown because of PSEdition compatibility, I feel like changing the error message to only address PSEdition problems won't help us in the long term.

The big problem is that it's possible to get this error, run `Import-Module <module-that-just-failed-to-autoload>` and get no error, but running the import doesn't fix the error either.

To my mind, the ideal solution is actually reflecting the true cause of the autoload failure in the outer error message and then providing an `InnerException` with more information. That way subprocesses and remote scenarios still get enough information at the top level to begin diagnosis.

But the fact that this error is encountered at all for PSEdition compatibility failure is a bug. If we've got a minimal repro, I'll work on fixing it (@joeyaiello).

I would go as far as to say that we should never throw an error like this one; if we don't try to autoload PSEdition-incompatible modules, we shouldn't try to autoload modules that are incompatible in any other way.

rjmholt: Discussed the above further with @joeyaiello offline. Summary:

- We think the error message should give as much information as practical for the user to debug themselves, even if it breaks the abstraction (say, of an incompatible module being "as if it doesn't exist")
- There are three scenarios when trying to autoload a module based on a command usage:
  1. No module exists on the module path with the required command:
      * Throw `CommandNotFoundException`
  2. A compatible module with the command is found on the module path:
      * Load the first compatible module
      * Execute the command
  3. The only modules on the module path with the command are *incompatible* (maybe due to edition, but really for any reason where the module manifest is valid but not importable in the current PowerShell session):
      * Continue searching for compatible modules and if we find one, do behaviour (ii)
      * If none is found, throw an error with some detail in the top level message as to why the autoload failed (such as `IncompatiblePSEditionException` or whatever it's called)

joeyaiello: To reiterate from the customer experience side: if I try to load an incompatible module, I'm given a one step solution that might fix me. If I try to run an incompatible cmdlet (*and* it's confirmed that there is no compatible cmdlet anywhere in PSModulePath), I should get a one step solution that might fix me.

SteveL-MSFT:
Closing this PR as based on the discussion this PR doesn't solve the underlying problem

PR #7396, opened 2018-07-29
[Feature] Fix path handling with wildcard char

kwkam:
## PR Summary

_*System.Management.Automation/engine/CommandCompletion/CompletionCompleters:_
Use WildcardPattern::Escape to escape completion text of filename.

_*System.Management.Automation/engine/SessionStateContainer:_
@Get-ChildItem
Unescape non-literal, non-glob path before existence checking.
This solve the issue where Get-GhildItem does not behave correctly
when -Path contains special characters. For example,
```pwsh
Get-ChildItem -Path './`[dir`]'
```
will complain with error "Cannot find path ..." while
```pwsh
Get-ChildItem -Path './`[dir`]' -Depth 0
```
will work fine.

@New-Item
Unescape non-literal path when -Name is not set.
This works around for New-Item treating -Path as literal path while
it can also be globbable. For example, assuming there is ```"[file]1"```
in current directory, and tab completion suggests ```'./`[file`]1'```
for the -Path argument, but
```pwsh
New-Item -Path './`[file`]2'
```
will create a file named ```"`[file`]2"``` instead of ```"[file]2"```.

_*System.Management.Automation/engine/regex:_
WildcardPattern::Escape should also escape ```"`"``` since
WildcardPattern::Unescape would unescape it,
and the matcher parse it as escape character.

_*System.Management.Automation/namespaces/LocationGlobber:_
Do not escape CWD when LiteralPath is used. This causes issues when
both -LiteralPath and CWD contains special characters.
Do not pass regex escaped string to WildcardPattern::Get. This fails
tab completion when doing ```"./path/to/`[f"<TAB>```.

_*System.Management.Automation/utils/PathUtils:_
Unescape non-literal, non-glob path before calling
GetUnresolvedProviderPathFromPSPath since it only accepts literal path.
This solve the issue where some commands depending on that method
will treat -Path as literal unintentionally. For example,
```pwsh
Out-File -Path './`[out`].txt' -Append
```
will create a new file named ```"`[out`].txt"``` instead of writing to ```"[out].txt"```.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
+++ b/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
@@ -4303,6 +4303,12 @@ namespace System.Management.Automation
                         if (CompletionRequiresQuotes(completionText, !useLiteralPath))
                         {
                             var quoteInUse = quote == string.Empty ? "'" : quote;
+
+                            if (!useLiteralPath)
+                            {
+                                completionText = WildcardPattern.Escape(completionText);
+                            }
+
                             if (quoteInUse == "'")
                             {
                                 completionText = completionText.Replace("'", "''");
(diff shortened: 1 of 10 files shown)


kwkam:
Further split into #7403 #7404 #7405 #7406 #7407

PR #7399, opened 2018-07-29
CommandSearcher: treat invalid wildcard path as literal

kwkam:
## PR Summary

Do not glob path with invalid wildcard pattern (eg. "./[.ps1").

## PR Motivation

Merge the Qualified Path Search and Relative Path Search into a single function (`GetNextFromPath` into `ResolvePSPath`).

Add an internal method `ContainsValidWildcardPattern` to the WildcardPattern API to make the checking be more sensitive to invalid bracket patterns, to allow faster fallback from wildcard search for fully qualified path to a literal search.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**


src/System.Management.Automation/engine/GetCommandCommand.cs
```diff
@@ -705,7 +705,7 @@ private void AccumulateMatchingCommands(IEnumerable<string> commandNames)
                         moduleName = this.Module[0];
                     }
 
-                    bool isPattern = WildcardPattern.ContainsWildcardCharacters(plainCommandName);
+                    bool isPattern = WildcardPattern.MayBeWildcardPattern(plainCommandName);
```
iSazonov: ( Perhaps the best way would be to delegate this to a filesystem provider or even to .Net Core although it looks too difficult and big change. )

src/System.Management.Automation/engine/regex.cs
```diff
@@ -281,6 +281,55 @@ public static bool ContainsWildcardCharacters(string pattern)
...
+        /// <param name="pattern">
+        /// String which needs to be checked if it may be wildcard pattern
+        /// </param>
+        /// <returns> true if the string may be wildcard pattern, false otherwise. </returns>
+        /// <remarks>
+        /// Currently string contains { '*', '?' } or both { '[', ']' } is
+        /// considered to be possibly a wildcard pattern and
+        /// '`' is the escape character.
+        /// </remarks>
+        public static bool MayBeWildcardPattern(string pattern)
```
iSazonov: Shouldn't we enhance `ContainsWildcardCharacters` method?

iSazonov: Shouldn't we enhance `ContainsWildcardCharacters` method? 
Seem it would be better. And I'd consider using ReadOnlySpan.

iSazonov (commented):
Leave a comment

src/System.Management.Automation/engine/regex.cs
kwkam: I believe `test if a string contains wildcard characters` and `test if a string looks like a wildcard pattern` is two different thing.
I can enhance `ContainsWildcardCharacters`, but I am not sure whether it will break other components...

iSazonov: Yes, it can affect many things but I think it is worth it to take a deeper - split the code just because we have not learned it enough is a bad practice.
I believe that if `ContainsWildcardCharacters` method takes into account _escapes_, then it should do the same for square brackets.

kwkam: I tried to merge them, but I am not sure whether it is good enough.

src/System.Management.Automation/engine/CommandSearcher.cs
```diff
@@ -458,84 +458,19 @@ private CommandInfo GetNextFromPath()
...
 
-                    CommandDiscovery.discoveryTracer.WriteLine(
-                        "Path resolved to: {0}",
-                        path);
+            string resolvedPath = ResolvePSPath(_commandName);
 
-                    result = GetInfoFromPath(path);
-                }
-            } while (false);
+            if (resolvedPath != null && File.Exists(resolvedPath))
```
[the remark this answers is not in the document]
kwkam: Removed.

iSazonov:
@SteveL-MSFT do we have interest in the change?

SteveL-MSFT:
@iSazonov yes, I think this change is something we want

iSazonov:
@kwkam Please rebase to pass CIs.

kwkam:
@iSazonov OK, rebased to latest master.

src/System.Management.Automation/engine/regex.cs
```diff
@@ -258,24 +258,33 @@ public static bool ContainsWildcardCharacters(string pattern)
...
+                switch (pattern[index])
                 {
-                    ++index;
+                    case '[':
+                        ++bracket;
+                        break;
+
+                    case ']':
+                        --bracket;
+                        break;
```
daxian-dbw: @kwkam By looking at the code, as long as the given string contains unpaired `[` or `]`, it's considered not having wildcard characters, even if the string contains `*` or `?`, is that what you are after? If that's the case, you probably shouldn't call `IsWildcardChar`, but use `switch` only instead:
```c#
switch (pattern[index])
{
    case '*':
    case '?':
        result = true; break;
    case '[':
        result = true;
        ++bracket;
        break;
    case ']':
        result = true;
        --bracket;
        break;
    case escapeChar:
        ++index;
        break;
```
But still, this is not good enough, what if you have even number of `[` and `]` but the order is wrong? Like `./][.ps1`?

daxian-dbw (changes requested):
`ContainsWildcardCharacters` is a public method in `WildcardPattern`. Changing to the behavior is a breaking change. The method name clearly indicates that it only checks if the string contains any wildcard characters, but not if it contains **_valid wildcard characters_**. If we need to check for valid wildcard characters, a new method should be introduced.

iSazonov:
> but not if it contains valid wildcard characters.

Does the comment only refer to this method or its use too?

daxian-dbw:
> Does the comment only refer to this method or its use too?

The comment only applies to the method, not to its uses. If a caller assumes differently, then another method should be added for checking if a path contains invalid wildcard characters.

iSazonov:
I think the breaking change is in gray area because using of the method implies that we want exactly working wildcards and not something similar ']['. Also I'd said that passing '][' in the method is a bug.

daxian-dbw:
@iSazonov This is more of a unacceptable changes in public contract. Quoted from the [breaking-chagne-doc](https://github.com/PowerShell/PowerShell/blob/master/docs/dev-process/breaking-change-contract.md#unacceptable-changes):

> A code change that results in a change to the existing behavior observed for a given input with an API, protocol or the PowerShell language.

daxian-dbw:
Add the "Review-Committee" label.
To @PowerShell/powershell-committee: This PR proposes a behavior change to the public method `WildcardPattern.ContainsWildcardCharacters(string pattern)`. Currently, `ContainsWildcardCharacters(string pattern)` only checks if the string contains any wildcard characters. The proposed change is to make it check for **valid** wildcard patterns. That means, for the string `[abc` or `][abc`, currently the method returns `true` and it would return `false` with the change.

The proposed change is a breaking change that falls in bucket 1, so we need committee to review.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this.  Recommendation is to deprecate this api and introduce `ContainsValidWildcardPattern()` API and fix all callsites to use the new method (separate PR).

kwkam:
Should the new method be a strict checking (e.g. try/catch with the WildcardPatternParser::Parse)?

iSazonov:
At first glance it should be `TryParse()`. Although we should review all callsites (131) - huge work!

daxian-dbw:
The method `ContainsWildcardCharacters` is called at many places in powershell, and I believe at least some of the callers only need a hint about whether a string *might* contains wildcard pattern, instead of an accurate answer, so for those caller, a perf penalty will be charged as `[WildcardPatternParser]::Parse` is relatively expensive.

Even for the callers that do want the accurate answer, another call to `Parse` will be made when creating/using the `WildcardPattern` from the string, so we will be paying the tax twice.

We can still add the new method `ContainsValidWildcardPattern` wrapping a call to `WildcardPatternParser.Parse`, so we can accurately know if a string contains valid wildcard patterns. But it's probably not right to blindly change all callers of `ContainsWildcardCharacters` to the new method.

daxian-dbw:
@kwkam Changes regarding `ContainsValidWildcardPattern` and `ContainsValidWildcardPattern` would be a bit tricky and likely involve many changes in other places. How about we revert the changes to `ContainsValidWildcardPattern` in this PR and make it only about the refactoring changes in `CommandSearcher.cs`? We can merge the refactoring changes, and open a separate issue/PR to discuss the right fix for `ContainsValidWildcardPattern`.

iSazonov:
I full agree with @daxian-dbw.

iSazonov:
@kwkam Do you agree with @daxian-dbw suggestion to split the PR?

kwkam:
@iSazonov Yes, but since the added method only affects CommandSearcher, I believe a partial fix is better than no fix (the original purpose of this PR is to fix the CommandSearcher). Could we open another PR to follow this up (improve/rewrite `ContainsValidWildcardPattern` and review other callsites)?

src/System.Management.Automation/engine/regex.cs
rjmholt: Yeah, this is a context-free grammar -- you'll need a stack to recognise it.

daxian-dbw:
> Could we open another PR to follow this up (improve/rewrite ContainsValidWildcardPattern and review other callsites)

Sure, please go ahead open a PR for that specific issue. Thank you!

src/System.Management.Automation/engine/regex.cs
```diff
@@ -278,6 +278,64 @@ public static bool ContainsWildcardCharacters(string pattern)
...
+
+                    // If it is an escape character then advance past
+                    // the next character
+                    case escapeChar:
+                        ++index;
+                        break;
+                }
+            }
+            return result && !opened;
```
daxian-dbw: Please correct me if I'm wrong, but this will return `true` for a string like `"]abc"`.
For this PR, let's make this method able to discover unpaired `[` and `]` and incorrect order of `[` and `]`.

```c#
public static bool ContainsValidWildcardPattern(string pattern)
{
    if (string.IsNullOrEmpty(pattern))
    {
        return false;
    }

    bool result = false;
    int openBracketCount = 0;

    for (int index = 0; index < pattern.Length; ++index)
    {
        switch (pattern[index])
        {
            case '*':
            case '?':
                result = true;
                break;

            case '[':
                result = true;
                openBracketCount++;
                break;

            case ']':
                result = true;
                openBracketCount--;
                break;

            // If it is an escape character then advance past
            // the next character
            case escapeChar:
                index++;
                break;
        }

        if (openBracketCount < 0)
        {
            // Found ']' appears before '[', so it's not a valid wildcard pattern.
            return false;
        }
    }
    return result && openBracketCount == 0;
}
```

src/System.Management.Automation/engine/regex.cs
```diff
@@ -278,6 +278,64 @@ public static bool ContainsWildcardCharacters(string pattern)
...
+        /// <returns>
+        /// true if the string does not contain invalid wildcard pattern,
+        /// false otherwise.
+        /// </returns>
+        /// <remarks>
+        /// Currently string contains { '*', '?' } or both { '[', ']' } is
+        /// considered to be properbly a wildcard pattern and
+        /// '`' is the escape character.
+        /// </remarks>
+        public static bool ContainsValidWildcardPattern(string pattern)
```
daxian-dbw: Before finalizing the behavior of this method, let's make it internal.

daxian-dbw (changes requested):
I agree to moving toward deprecating/replacing `ContainsWildcardCharacters` in powershell code incrementally.

src/System.Management.Automation/engine/regex.cs
rjmholt: Is the intent of this method to return `false` if no wildcards are used at all? (i.e. the string is "valid", but contains no wildcard usage). Currently it will return false for the input `"string"`. It seems simpler to me for the method to just check that any wildcard usage is valid. That way the monotonic `result` variable can be thrown out.

Also the string `"duck["` will cause an `IndexOutOfRangeException` here.

Possible implementation:
```csharp
internal static bool ContainsInvalidWildcardUsage(string pattern)
{
    // Empty patterns are valid
    if (pattern == null)
    {
        return true; // Possibly throw an exception -- depends on semantic interpretation
    }

    int bracketDepth = 0;
    for (int i = 0; i < pattern.Length; i++)
    {
        switch (pattern[i])
        {
            case escapeChar:
                 i++; // Skip the next character
                 continue;

            case '[':
                if (i + 1 < pattern.Length && pattern[i + 1] == ']')
                {
                    return true;
                }
                bracketDepth++;
                continue;

            case ']':
                bracketDepth--;
                if (bracketDepth < 0)
                {
                     return true;
                }
                continue;
        }
    }

    return bracketDepth == 0;
}
```

rjmholt: Then you can get the `ContainsValidWildcardPattern` behaviour with:
```csharp
internal static bool ContainsValidWildcardPattern(string pattern)
{
    if (ContainsInvalidWildcardUsage(pattern))
    {
        return false;
    }

    foreach (char c in pattern)
    {
        case '[':
        case ']':
        case '*':
        case '?':
              return true;
    }

    return false;
}
```

daxian-dbw: > Is the intent of this method to return false if no wildcards are used at all?

Yes, the purpose is to have an relatively accurate answer for "Is this string contains any wildcard pattern".

> Also the string `"duck["` will cause an IndexOutOfRangeException here.

Good catch! Length needs to be checked against 'index + 1' in `case '[':`.

> Then you can get the `ContainsValidWildcardPattern` behaviour with:

There are 2 concerns with this:
- We don't really need `ContainsInvalidWildcardUsage` for the issue in discussion.
- You may need to loop through the string twice in worse case, and 1.5 times in average case.

rjmholt: Yes, the second loop is the price of composability -- and the average pattern is not going to be long, so a pretty low price. I would argue that the fact that we don't already have a method that can identify a string with invalid wildcards in it is surprising to say the least.

iSazonov: > if (pattern[index + 1] == ']')

Boundary overflow.

daxian-dbw: > I would argue that the fact that we don't already have a method that can identify a string with invalid wildcards in it is surprising to say the least.

I guess that's because we haven't got a need for it until now 😄

> Boundary overflow.

Exactly. Boundary check is needed.

daxian-dbw: Hmm, good catch, it is valid. So the simple dumb check for `[]` in my code above is not right.
I have updated my example code above to not catch invalid `[]`.

I'm fine we only check for unpaired and out-of-ordered open/close brackets for now, but we do need to make the method internal until the behavior is finalized.

src/System.Management.Automation/engine/regex.cs
kwkam: Yes.

src/System.Management.Automation/engine/regex.cs
kwkam: Since `[[]` is also a valid pattern, I updated the code to not consider lone ']' as wildcard character. Please let me know if there is any problem.

src/System.Management.Automation/engine/regex.cs
kwkam: Reverted.

src/System.Management.Automation/engine/regex.cs
iSazonov: Seems It still doesn't work well. Try "[ [ ]".
I suggest to make the method public temporary, collect possible patterns and create Pester tests that help to design right algorithm. Before merge we make the method internal and hide/pending the tests.

kwkam: I tested the algorithm with a simple script:
```PowerShell
function ContainsValidWildcardPattern
{
    param (
        [string] $pattern
    )

    if ([string]::IsNullOrEmpty($pattern))
    {
        return $false;
    }

    $hasWildcard = $false;
    $openBracket = $false;

    for ($index = 0; $index -lt $pattern.Length; ++$index)
    {
        switch ($pattern[$index])
        {
            '*' {
                $hasWildcard = $true
            }

            '?' {
                $hasWildcard = $true
            }

            '[' {
                $hasWildcard = $true
                $openBracket = $true
            }

            ']' {
                # ']' is wildcard only if '[' exists
                # so we do not set hasWildcard here
                $openBracket = $false
            }

            # If it is an escape character then advance past
            # the next character
            '`' {
                ++$index
            }
        }
    }

    if ($openBracket) {
        return $false
    }

    return $hasWildcard
}

@(
    @{input = '['; IsPattern = $false}
    @{input = '[]'; IsPattern = $false}
    @{input = ']['; IsPattern = $false}
    @{input = '][]'; IsPattern = $false}
    @{input = '[[]'; IsPattern = $true}
    @{input = '[]]'; IsPattern = $true}
    @{input = '[][]'; IsPattern = $true}
    @{input = ']abc'; IsPattern = $false}
    @{input = '[]abc'; IsPattern = $false}
    @{input = '[abc[]abc]'; IsPattern = $true}
) | % {
    $out = if ($(ContainsValidWildcardPattern $_.input) -ne $_.IsPattern) {
        @{code = 31; verb = if ($_.IsPattern) { 'should be' } else { 'should not be' }}
    } else {
        @{code = 32; verb = if ($_.IsPattern) { 'is' } else { 'is not' }}
    }
    "`e[$($out.code)m'$($_.input)' $($out.verb) wildcard pattern`e[0m"
}
```
And the result is:
```
'[' is not wildcard pattern
'[]' should not be wildcard pattern
'][' is not wildcard pattern
'][]' should not be wildcard pattern
'[[]' is wildcard pattern
'[]]' is wildcard pattern
'[][]' is wildcard pattern
']abc' is not wildcard pattern
'[]abc' should not be wildcard pattern
'[abc[]abc]' is wildcard pattern
```
I will try to improve if this is still not good enough.

src/System.Management.Automation/engine/CommandSearcher.cs
```diff
@@ -446,84 +446,19 @@ private CommandInfo GetNextFromPath()
...
-                    File.Exists(resolvedPaths[0]))
-                {
-                    string path = resolvedPaths[0];
+            CommandDiscovery.discoveryTracer.WriteLine(
+                "Trying to resolve the path as an PSPath");
 
-                    CommandDiscovery.discoveryTracer.WriteLine(
-                        "Path resolved to: {0}",
-                        path);
+            string resolvedPath = ResolvePSPath(_commandName);
```
TravisEz13: This part of the change seems to allow faster fallback from wildcard search for fully qualified path to a literal search.  Can you update your PR summary to include these detail of the change.

TravisEz13 (changes requested):
I see two changes:

1. Merging the Qualified Path search with the Relative Path Search.
2. Making the function that checks for wildcards be more sensitive to invalid bracket patterns.

Can you update your PR summary making sure you describe these two separately?

Also, We should add tests for 1. to make sure that it did not regress.  I'll submit a PR to add tests for the existing behavior and you or I can rebase your PR when we are ready.

daxian-dbw:
@kwkam I apologize for neglecting (unintentionally) this PR for so long. I rebase this PR after @TravisEz13 have his tests changes merged. Once this PR is merged, I will work on improve the implementation of `IsValidWildcardPattern` as well as visit all the call sites of `ContainsWildcardCharacters`.

src/System.Management.Automation/engine/regex.cs
```diff
@@ -278,6 +278,70 @@ public static bool ContainsWildcardCharacters(string pattern)
...
+            {
+                return false;
+            }
+
+            bool hasWildcard = false;
+            bool openBracket = false;
+
+            for (int index = 0; index < pattern.Length; ++index)
+            {
+                switch (pattern[index])
```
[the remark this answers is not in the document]
kwkam: I believe there is little benefit as `IsWildcardChar` only tells whether the character is one of the wildcard characters, while in this function, we need to know exactly which character it is to determine the result.

kwkam:
@TravisEz13 Please see if the PR summary is appropriate.

kwkam:
@daxian-dbw No problem. If you have spare time would you please also take a look at #9257 #9258 ?

TravisEz13:
I have a PR that is adding a lot of tests to verify all the behaviors I've found, unfortunately it refactors the code some.

src/System.Management.Automation/engine/regex.cs
```diff
@@ -278,6 +278,70 @@ public static bool ContainsWildcardCharacters(string pattern)
...
+                {
+                    case '*':
+                    case '?':
+                        hasWildcard = true;
+                        break;
+
+                    case '[':
+                        hasWildcard = true;
```
rjmholt: `ContainsValidWildcardPattern("du[[ck]")` will return `true`.

We probably need something like:

```csharp
case '[':
    if (openBracket)
    {
        return false;
    }
```

src/System.Management.Automation/engine/regex.cs
```diff
@@ -278,6 +278,70 @@ public static bool ContainsWildcardCharacters(string pattern)
...
+        /// <returns>
+        /// true if the string does not contain invalid wildcard pattern,
+        /// false otherwise.
+        /// </returns>
+        /// <remarks>
+        /// Currently string contains { '*', '?' } or both { '[', ']' } is
+        /// considered to be properbly a wildcard pattern and
+        /// '`' is the escape character.
+        /// </remarks>
+        internal static bool ContainsValidWildcardPattern(string pattern)
```
rjmholt: A couple of extra cases this method should reject:

- `"[a-mt]ook"`
- `"[ab-z]ook"`
- `"[a?b]"`
- `"[a*]x"`
- `"goose]"`

rjmholt: Worth putting these into test cases

test/powershell/Modules/Microsoft.PowerShell.Core/Pester.Commands.Cmdlets.GetCommand.Tests.ps1
```diff
@@ -7,11 +7,13 @@ Describe "Tests Get-Command with relative paths and wildcards" -Tag "CI" {
         # Create temporary EXE command files
         $file1 = Setup -f WildCardCommandA.exe -pass
         $file2 = Setup -f WildCardCommand[B].exe -pass
+        $file3 = Setup -f [.exe -pass
```
rjmholt: These are probably best as Pester TestCases:

```powershell
BeforeAll {
    $testFileNames = @(
        'WildCardCommandA'
        'WildCardCommand[B]'
        '['
        '[a-mt]ook'
        '[ab-z]ook'
        '[a?b]'
        '[a*]x'
        'goose]'
    ) | ForEach-Object { "$_.exe" }

    foreach ($testFileName in $testFileNames)
    {
        Setup -f $testFileName -pass
        if ($IsLinux -or $IsMacOS)
        {
            /bin/chmod 777 $testFileName
        }
    }

    $fileTestCases = $testFileNames | ForEach-Object { @{ Filename = $_ } }
}

...

It "Should find the file <Filename>" -TestCases $fileTestCases {
    param([string]$Filename)

    $result = Get-Command -Name $Filename -Type Application
    $result | Should -Not -BeNullOrEmpty
    $result | Should -BeExactly $Filename
}
```

iSazonov: I'd add "du[[ck]" too.

src/System.Management.Automation/engine/regex.cs
kwkam: `du[[ck]` should be a valid wildcard pattern.

src/System.Management.Automation/engine/regex.cs
kwkam: I'm confused now. All patterns suggested in the list are valid, would you please explain why should `ContainsValidWildcardPattern` reject them?

test/powershell/Modules/Microsoft.PowerShell.Core/Pester.Commands.Cmdlets.GetCommand.Tests.ps1
rjmholt: @kwkam you're absolutely right that all the cases I suggested *are* valid wildcard patterns (except `goose]`).

I think they're worth adding as positive test cases if possible

rjmholt (approved):

src/System.Management.Automation/engine/regex.cs
daxian-dbw: `"goose]"` is a not a pattern, even though it's valid (not throw when calling `WildCardPatter.IsMatch`)

daxian-dbw:
Referencing to the algorithm and tests in https://github.com/PowerShell/PowerShell/pull/7399#discussion_r232455452

I updated the test cases as follows:
```
@(
    @{input = '['; IsPattern = $false}
    @{input = '[]'; IsPattern = $false}
    @{input = ']['; IsPattern = $false}
    @{input = '][]'; IsPattern = $false}
    @{input = '[[]'; IsPattern = $true}
    @{input = '[]]'; IsPattern = $true}
    @{input = '[][]'; IsPattern = $true}
    @{input = ']abc'; IsPattern = $false}
    @{input = '[]abc'; IsPattern = $false}
    @{input = '[abc[]abc]'; IsPattern = $true}
    @{input = '[a-mt]ook'; IsPattern = $true}
    @{input = '[ab-z]ook'; IsPattern = $true}
    @{input = '[a?b]'; IsPattern = $true}
    @{input = '[a*]x'; IsPattern = $true}
    @{input = 'goose]'; IsPattern = $false}
    @{input = '[az-]'; IsPattern = $true}
    @{input = '[!--]a'; IsPattern = $true}
    @{input = '[e--]a'; IsPattern = $false}
    @{input = '[az-b]'; IsPattern = $false}
) | % {
    $out = if ($(ContainsValidWildcardPattern $_.input) -ne $_.IsPattern) {
        @{code = 31; verb = if ($_.IsPattern) { 'should be' } else { 'should not be' }}
    } else {
        @{code = 32; verb = if ($_.IsPattern) { 'is' } else { 'is not' }}
    }
    "`e[$($out.code)m'$($_.input)' $($out.verb) wildcard pattern`e[0m"
}
```
And this is how the current algorithm works with those test cases:
image

It's not perfect, but is a good start. I will submit a separate PR to improve the algorithm.

--------------------------------------

As for the refactoring changes in `CommandDiscovery.cs`, @TravisEz13's PR #9202 has been merged, which refactored this file a lot. So the refactoring changes in this PR is not relevant anymore and can be reverted now.
I will clean up this PR today.

daxian-dbw:
I have rebased the PR and reverted the refactoring changes in `CommandSearcher.cs`.

I think about it more when doing the rebase, and think this might not be the right fix, for two reasons:
1. `Get-Command .\[.exe -Type Application` works as expected, but `Get-Command .\[.exe`  under the same folder doesn't.
2. I think the behavior of 'Get-Command' probably should be similar to what #9202 has made to the command line resolution:
    - First try to resolve it as a literal name
    - Then, if the name contains wildcards, try to resolve it as wildcard name
       - when the wildcard is invalid
          - if it was successfully resolved as literal path, then don't report any error.
          - if it was not successfully resolved as literal path before, then report wildcard pattern error.

Having a method `ContainsValidWildcardPattern` to quickly tell if a string actually represents valid wildcard pattern is certainly useful, especially in terms of perf.
But sometimes, the user does want to specify a wildcard pattern as input, but just doesn't make it right. In those case, if our method `ContainsValidWildcardPattern` is very accurate, PowerShell will not throw error about "invalid wildcard pattern" because it believes the user input is a literal string. That might cause problems to user experience and also might cause behavior changes.

iSazonov:
> I think the behavior of 'Get-Command' probably should be similar to what #9202 has made

I agree that the behavior should be the same. Otherwise, we get amazing behavior when the `Get-Command` cmdlet finds a command, but when we type it, we get "not found" error.

> if our method ContainsValidWildcardPattern is very accurate, PowerShell will not report "invalid wildcard pattern" error because it believes the user input is a literal string. That might cause problems to user experience and also might cause behavior changes.

Maybe updated error message will be enough.

iSazonov:
It seems we need to review all cmdlets with -Pat/LiteralPath to get the same behavior for all.

TravisEz13:
rebased PR

TravisEz13:
@PoshChan Please remind me in 1 hour

PoshChan:
@TravisEz13, this is the reminder you requested 1 hour ago

TravisEz13:
@daxian-dbw  Please update your review

TravisEz13:
@daxian-dbw Please update your review

PR #7407, opened 2018-07-30
Fix tab completion for filename contains literal wildcard character

kwkam:
## PR Summary

_*System.Management.Automation/engine/CommandCompletion/CompletionCompleters:_
Use WildcardPattern::Escape to escape completion text of filename.

_*System.Management.Automation/engine/regex:_
WildcardPattern::Escape should also escape ```"`"``` since
WildcardPattern::Unescape would unescape it,
and the matcher parse it as escape character.

_*System.Management.Automation/namespaces/LocationGlobber:_
Do not pass regex escaped string to WildcardPattern::Get. This fails
tab completion when doing ```"./path/to/`[f"<TAB>```.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**


iSazonov:
@kwkam Please remove "[Feature]" from title in your PRs. We need this prfix only in _commit header_ to force run feature tests.

src/System.Management.Automation/engine/regex.cs
```diff
@@ -204,9 +204,10 @@ internal static string Escape(string pattern, char[] charsNotToEscape)
                 char ch = pattern[i];
 
                 //
-                // if it is a wildcard char, escape it
+                // if it is a wildcard char or escape char, escape it
                 //
-                if (IsWildcardChar(ch) && !charsNotToEscape.Contains(ch))
+                if ((IsWildcardChar(ch) || ch == escapeChar) &&
+                    !charsNotToEscape.Contains(ch))
```
iSazonov: Why we always should escape the escape char? What scenario don't work without fix?

iSazonov:
@kwkam Sorry for delay - it is very sensetive aria and we should very careful.

src/System.Management.Automation/engine/regex.cs
kwkam: 1. When we accidentally created a file named ``` `[test`]2 ``` due to the incorrect handling of wildcard character, we cannot remove it using the -Path parameter with the tab completed name.
2. When I escaped string A and get B, and try to unescape B, I would expected that A will be returned, but it is not when A contains escape character.
3. The wildcard pattern parser always treat escape character as escape character.

test/powershell/Language/Parser/LanguageAndParser.TestFollowup.Tests.ps1
```diff
@@ -303,3 +303,16 @@ Describe "Hash expression with if statement as value" -Tags "CI" {
...
+Describe "WildcardPattern" -Tags "CI" {
+    It "Unescaping escaped string '<inputStr>' should get the original" -TestCases @(
+        @{inputStr = '*This'}
+        @{inputStr = 'Is?'}
+        @{inputStr = 'Real[ly]'}
+        @{inputStr = 'Ba`sic'}
+        @{inputStr = 'Test `[more`]?'}
+    ) {
+        param($inputStr)
+        [WildcardPattern]::Unescape([WildcardPattern]::Escape($inputStr)) | Should -BeExactly $inputStr
```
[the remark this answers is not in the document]
kwkam: Updated test.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
```diff
@@ -739,6 +739,9 @@ dir -Recurse `
                 @{ inputStr = "Get-Process >'.\My ``[Path``]\'"; expected = "'.${separator}My ``[Path``]${separator}test.ps1'" }
                 @{ inputStr = "Get-Process >${tempDir}\My"; expected = "'${tempDir}${separator}My ``[Path``]'" }
                 @{ inputStr = "Get-Process > '${tempDir}\My ``[Path``]\'"; expected = "'${tempDir}${separator}My ``[Path``]${separator}test.ps1'" }
+                @{ inputStr = "cd 'My ``["; expected = "'.${separator}My ``[Path``]'" }
```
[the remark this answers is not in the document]
kwkam: 1. Updated test.
2. I disagree. User should type escape for `-Path` (which support globbing), just like we do in bash (eg. `touch \[test\]`). But if you typed escape and auto completion still did not work, that is because by default, powershell treats "unquoted string without space" like "double-quoted string", that "double-quoted string" will consume escape character. For file name `[test]`, either ````Get-Content -Path ```[```` or ``Get-Content -Path '`[`` will work.
3. Would you please let me know what did not work for `-LiteraPath`?
I tried `Get-Item -LiteralPath [`(TAB) for a file named `[test]` and it did auto complete.

kwkam:
@iSazonov Yes, that's why this PR was placed behind the cmdlet ones. Please let me know if something is not correct/not working.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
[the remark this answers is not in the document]
iSazonov: 2. Consider interactive UX. User do `dir` and then type directory or file name as one see on screen - without escapes. In the scenario both the user and PowerShell know right name. And more, we can have both names - with escape char too.
3. Get-Item '.\My `[

src/System.Management.Automation/engine/regex.cs
iSazonov: My understanding is that we should test if a file/directory/path exists before trying escape. If exist then return the path. Then continue with escaping wildcard chars. Then continue with expansion wildcards if -Path.
Also what if there is some special symbols in the path? Seems we should process them in cycle.

kwkam: Okay, I know what is wrong here. This part is to fix the `WildcardPattern::Escape` function itself, which is not necessary related to the file system.
When `WildcardPattern::Get` parse a pattern string, it always treats the escape character as escape character of an escape character. That means, ``` `` ``` will be parsed to `` ` ``.
And so is the `WildcardPattern::Unescape` method, `[WildcardPattern]::Unescape('``')` will return `` ` `` instead of ``` `` ```.
But, ``[WildcardPattern]::Escape('`')`` does not escape the escape character, which conflicts with the others.

test/powershell/Host/TabCompletion/TabCompletion.Tests.ps1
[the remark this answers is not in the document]
kwkam: 2. The problem is, `-Path` support wildcard. When a path can be an expression, the special characters should always be escaped otherwise it will become too complicated. For example, in a directory there are `A.txt`, `B.txt` and `[AB].txt`, what should `Get-Item -Path [AB].txt` return?
3. I tested both `-Path` and `-LiteralPath` and both will complete the file name. eg.
`New-Item -Name 'My [test]'`
``Get-Item '.\My `[`` will be completed to ``Get-Item '.\My `[test`]'``, and
``Get-Item -LiteralPath '.\My [`` will be completed to ``Get-Item -LiteralPath '.\My [test]'``
Did you use the build here? https://ci.appveyor.com/project/PowerShell/powershell/build/v6.1.0-rc.10960/job/3i76g2ar1wqlvw7x/artifacts

src/System.Management.Automation/engine/regex.cs
iSazonov: @kwkam I see your point.
We have the public API
```c#
public static string Escape(string pattern)
```
So the change is a breaking change and it is not desirable to change this behavior until  it is  established that this is a significant bug.

kwkam: @iSazonov Yes, we should be careful with changes in public API, but I doubt if this is not a significant bug. For example, there is a list of text:
```
`[HIGH] patch 1`
`[MED] update 1`
`[LOW] patch 1`
`[MED] patch 2`
`[HIGH] hotfix 3`
`[LOW] patch 2`
`[LOW] hotfix 3`
```
Now, I want to search for `` `[LOW] patch 1` `` and `` `[LOW] patch 2` ``, and I am using WildcardPattern:
```PowerShell
$prefix = '`[LOW] patch '
$pattern = '[1-2]'
$suffix = '`'
$sarg = [WildcardPattern]::Escape($prefix) + $pattern + [WildcardPattern]::Escape($suffix)
$opts = [Management.Automation.WildcardOptions]::Compiled
$matcher = [WildcardPattern]::Get($sarg, $opts)
$ListOfText | ? {$matcher.IsMatch($_)}
```
No result will be return because of the inconsistency between the Escape method and the WildcardPattern parser. (And this is also the cause of the issue in auto-completed filename which contains escape character)

iSazonov: Thanks for good example!

We need PowerShell Committee conclusion.
/cc @SteveL-MSFT

SteveL-MSFT:
cc @BrucePay to follow-up

iSazonov:
@SteveL-MSFT We have too many frozen PRs. How we can resolve the problem?

SteveL-MSFT:
@BrucePay will follow-up on this specific PR

iSazonov:
@SteveL-MSFT @BrucePay Could you please make a conclusion about the PR?

kwkam:
@BrucePay Erm, I am not sure how should I react, this is not about the quoting behaviour. You could see the conversation here:
https://github.com/PowerShell/PowerShell/pull/7407/files/931f073c743ace3f39e6d17cd116030a5db1eb15#diff-c1b737e0eb21f0203a20baae71a1c372
and here:
https://github.com/PowerShell/PowerShell/pull/7407/files/931f073c743ace3f39e6d17cd116030a5db1eb15#diff-4b52c367ceb4e4ce73e2616e9cc98b22

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this, we agree that this is a fragile area to change, however, would accept putting this behind an `ExperimentalFeature` flag to validate potential for regression and collect community feedback.

kwkam:
@SteveL-MSFT I have no idea how this should be done, is there any example/reference for the `ExperimentalFeature` implementation?

iSazonov:
@kwkam Please see #8696 and comment https://github.com/PowerShell/PowerShell/pull/8995#issuecomment-469053815

kwkam:
@iSazonov Thanks for the reference. I checked the test output, the linux one looks ok but the windows one is not showing the ExperimentalFeature-Enabled checking result, did I do something wrong?

iSazonov:
@kwkam I found the results under Artefacts -> testResults
So no problem.

iSazonov:
@SteveL-MSFT What is the fate of experimental features? I'd suggest to create an tracking issue (with label "Area-Experimental"+"Issue-Discussion") for every one to collect feedback and discuss. I believe this help us to move an experimental feature to common code or reject.

SteveL-MSFT:
@iSazonov if the question is about how does an experimental feature become non-experimental, then my current thinking is to have it as part of the `Send-PSFeedback` cmdlet proposed in https://github.com/PowerShell/PowerShell-RFC/pull/158.  For now, we can certainly create a discussion issue in this repo to get feedback, but it limits feedback to GitHub users (although they are the ones most likely to use Experimental Features...)

iSazonov:
> (although they are the ones most likely to use Experimental Features...)

If we want get more feedback we could add first-experience feature to inform users:
- write to console a message at first start of preview
- write the same with `help -FirstExperience`

SteveL-MSFT:
@iSazonov An about_whatsnew.md might be a good idea.  This is more intended for experienced users moving to a newer version.  If you're intent for "first experience" is for beginners, I think that's something else that would be equally interesting, but different.

iSazonov:
@SteveL-MSFT Should we re-open and re-assigned the PR?

test/powershell/engine/Api/WildcardPattern.Tests.ps1
```diff
@@ -0,0 +1,65 @@
+Describe "WildcardPattern Escape - Experimental-Feature-Disabled" -Tags "CI" {
```
TravisEz13: Add copyright header

src/System.Management.Automation/namespaces/LocationGlobber.cs
```diff
@@ -3363,13 +3363,11 @@ internal static string RemoveProviderQualifier(string path)
                 StringContainsGlobCharacters(leafElement) ||
                 isLastLeaf)
             {
-                string regexEscapedLeafElement = ConvertMshEscapeToRegexEscape(leafElement);
```
TravisEz13: Can you explain this change?

TravisEz13 (changes requested):

src/System.Management.Automation/namespaces/LocationGlobber.cs
kwkam: As explained in the PR Summary, `regexEscapedLeafElement` is passed to `WildcardPattern::Get`, which does not make sense since in PowerShell, regex escaped string is `\[abc\]`, while wildcard escaped string is ``` `[abc`] ```.
Moreover, why should we escape/alter the pattern string in methods `GenerateNew*PathsWithGlobLeaf` which is supposed to glob? This behaviour is causing issue on tab completion when filename contains bracket.

test/powershell/engine/Api/WildcardPattern.Tests.ps1
```diff
@@ -0,0 +1,67 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+Describe "WildcardPattern Escape - Experimental-Feature-Disabled" -Tags "CI" {
```
iSazonov: Please insert blank line.
```suggestion

Describe "WildcardPattern Escape - Experimental-Feature-Disabled" -Tags "CI" {
```

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -4383,6 +4383,12 @@ internal static IEnumerable<CompletionResult> CompleteFilename(CompletionContext
                         if (CompletionRequiresQuotes(completionText, !useLiteralPath))
                         {
                             var quoteInUse = quote == string.Empty ? "'" : quote;
+
```
TravisEz13: should this change be behind the experimental feature as well?

msftrncs:
I tested a variation of this PR, while testing the PR I am preparing to fix all argument completion patterns along with variable and member completions, #10226.

Previously, I could not use Get-Content to read the content of a file named ``` `[hello`].txt ```, the file would NOT be read, but there would also be no error.

I did not modify the WildCardPattern.Escape() method, but instead added `.Replace("`", "``")` to every single instance in the PowerShell code base of the Escape() method.

Now Get-Content reads the content of that file.

Even if changes to WildCardPattern.Escape() are considered too breaking, at least manually escape the backticks, in each call.

Unfortunately this doesn't fix the issue with redirection, which accepts a wildcard path, but then sends it to out-file as a literal.

kwkam:
> Unfortunately this doesn't fix the issue with redirection, which accepts a wildcard path, but then sends it to out-file as a literal.

@msftrncs It was in another PR #9258

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
TravisEz13: This was a condition of the committee

iSazonov:
@kwkam Please resolve merge conflicts and address comments.

src/System.Management.Automation/namespaces/LocationGlobber.cs
kwkam: Can I mark this resolved?

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
kwkam: Except filename completion will now also escape `*` and `?` in non-literal path, the behaviour should be identical unless the experimental feature (the breaking part) is enabled.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -4398,6 +4398,12 @@ internal static IEnumerable<CompletionResult> CompleteFilename(CompletionContext
                         if (CompletionRequiresQuotes(completionText, !useLiteralPath))
```
PoshChan, line 4398: @kwkam, your last commit had 2 failures in `PowerShell-CI-static-analysis`
Verify Markdown Links.Verify links in /home/vsts/work/1/s/ADOPTERS.md.https://docs.universaldashboard.io should work
```powershell
Name or service not known
at <ScriptBlock>, /home/vsts/work/1/s/test/common/markdown/markdown-link.tests.ps1: line 114
114:                                 $null = Invoke-WebRequest -Uri $url -RetryIntervalSec 10 -MaximumRetryCount 6

```
Verify Markdown Links.Verify links in /home/vsts/work/1/s/README.md.https://powershell.visualstudio.com/PowerShell/_build/latest?definitionId=32 should work
```powershell
Failed to complete request to "https://powershell.visualstudio.com/PowerShell/_build/latest?definitionId=32". Response status code does not indicate success: 404 (Not Found).
at <ScriptBlock>, /home/vsts/work/1/s/test/common/markdown/markdown-link.tests.ps1: line 119
119:                                     throw "Failed to complete request to `"$url`". $($_.Exception.Message)"

```

test/powershell/engine/Api/WildcardPattern.Tests.ps1
```diff
@@ -0,0 +1,68 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
TravisEz13, line 1:
```suggestion
# Copyright (c) Microsoft Corporation.
```

TravisEz13 (changes requested):

TravisEz13:
@daxian-dbw Can you review this?

floh96:
@iSazonov could this pr be reopened?
If I'm seeing this correctly, it was closed because of a copyright header change....

iSazonov:
@floh96 It makes no sense to reopen. The work could be continue if anybody grabbed the work and MSFT team agreed to review. Now you could open new issue to discuss and vote.

PR #7410, opened 2018-07-30
Improve performance of Group-Object

### LINKED ISSUE #7409: Improve performance of Group-Object
As of pwsh 6.1, preview-4, we have quadratic performance when the number of unique values -> n.

By gathering the input first, sorting it, and then only each new value to the last group, we can come much closer to `n * log(n)` instead of `n * n`.

I have a prototype with the following perf measurements:

Count| Unique| OldImpl | newImpl | Speedup | Command
------| ---------|--------- | ----------| ----------|-----------
10689| 8220 | 00:00:06.81 | 00:00:00.23|  29,1 | $allItemsInPowerShellSrcTree \| group {[io.path]::GetFileName($_)}
1690765 |3761|00:02:30.34 | 00:00:22.32 |  6,7 | $u \| group

where `$u` is a dataset of string values out of which 3700 is unique.

The only downside I have seen is that the both the order of the output objects are different, ~~and so is the order within the groups~~.

Is that part of the public contract?
~~Is it worth a PR?~~

powercode:
Fixes #7409.

## PR Summary

code cleanup
object[] -> IList<T> in PSType.ArrayToTuple signature to reduce allocations.
Sort input before grouping. Use sorting to only consider last group when comparing.
adding DebuggerDisplay for GroupInfo.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**


powercode:
Appveyor error seem unrelated to my changes.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
```diff
@@ -113,10 +116,10 @@ private static string BuildName(List<ObjectCommandPropertyValue> propValues)
             StringBuilder sb = new StringBuilder();
             foreach (ObjectCommandPropertyValue propValue in propValues)
             {
-                if (propValue != null && propValue.PropertyValue != null)
+                var propValuePropertyValue = propValue?.PropertyValue;
+                if (propValuePropertyValue != null)
```
iSazonov: Is the check really needed?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
```diff
@@ -246,53 +248,54 @@ public SwitchParameter NoElement
         internal static void DoGrouping(OrderByPropertyEntry currentObjectEntry, bool noElement, List<GroupInfo> groups, Dictionary<object, GroupInfo> groupInfoDictionary,
             OrderByPropertyComparer orderByPropertyComparer)
         {
-            if (currentObjectEntry != null && currentObjectEntry.orderValues != null && currentObjectEntry.orderValues.Count > 0)
+            var currentObjectorderValues = currentObjectEntry.orderValues;
```
iSazonov: Should it be `currentObjectOrderValues`?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
```diff
@@ -246,53 +248,54 @@ public SwitchParameter NoElement
...
-                GroupInfo currentGroupInfo = null;
-                if (groupInfoDictionary.TryGetValue(currentTupleObject, out currentGroupInfo))
+                if (groupInfoDictionary.TryGetValue(currentTupleObject, out var currentGroupInfo))
                 {
-                    if (currentGroupInfo != null)
-                    {
-                        //add this inputObject to an existing group
-                        currentGroupInfo.Add(currentObjectEntry.inputObject);
-                    }
+                    //add this inputObject to an existing group
```
iSazonov: Please add space after //.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
```diff
@@ -336,25 +341,31 @@ protected override void ProcessRecord()
...
                 if (AsHashTable)
                 {
-                    Hashtable _table = CollectionsUtil.CreateCaseInsensitiveHashtable();
+                    Hashtable table = CollectionsUtil.CreateCaseInsensitiveHashtable();
                     try
                     {
-                        foreach (GroupInfo _grp in _groups)
+                        foreach (GroupInfo grp in _groups)
                         {
                             if (AsString)
```
iSazonov: Make sense to move the cycle into the  `if (AsString)` ?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
powercode: yes, it is dereferenced in the else clause.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
powercode: yes

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
powercode: Not sure i understand what you suggest here.

powercode: Yes, I think that is clearer.

powercode:
I would like to fix the parameter sets too. 
```
[Parameter(ParameterSetName = "HashTable", Mandatory = true)]
public SwitchParameter AsHashTable { get; set; }

[Parameter(ParameterSetName = "HashTable")]
public SwitchParameter AsString { get; set; }
```

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
iSazonov: I thought more about performancethough :-) From this point of view it seems `for()` better than `foreach` - less allocations. Is the _groups so much that we can get a win?

iSazonov:
I believe we shouldn't mess performance changes with other and it is better fix parameter sets in new PR.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
iSazonov: I don't see the fix.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
```diff
@@ -302,22 +359,41 @@ protected override void ProcessRecord()
...
+                    {
+                        ArgumentException ex = new ArgumentException(UtilityCommonStrings.GroupObjectWithHashTable);
+                        ErrorRecord er = new ErrorRecord(ex, "ArgumentException", ErrorCategory.InvalidArgument, AsString);
+                        ThrowTerminatingError(er);
+                    }
+
+                    if (AsHashTable
+                        && !AsString
+                        && (Property != null && Property.Length > 1
+                            || _orderByProperty.MshParameterList.Count > 1))
```
iSazonov: We'd move this on previous line.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
```diff
@@ -83,7 +113,7 @@ internal static object ArrayToTuple<T>(IList<T> inputObjects, int startIndex)
     public sealed class GroupInfoNoElement : GroupInfo
     {
         internal GroupInfoNoElement(OrderByPropertyEntry groupValue)
-            : base(groupValue)
+        : base(groupValue)
```
iSazonov: Please move to previous line 115.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
```diff
@@ -224,6 +250,7 @@ public class GroupObjectCommand : ObjectBase
         /// <summary>
         /// Gets or sets the AsString parameter.
         /// </summary>
+        /// <value></value>
```
iSazonov: Please remove the empty XML comment.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
iSazonov: Please address the comment and fix case.

iSazonov (commented):
Leave a comment

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
```diff
@@ -131,15 +165,16 @@ private static string BuildName(List<ObjectCommandPropertyValue> propValues)
                     }
                     else
                     {
-                        sb.Append(string.Format(CultureInfo.InvariantCulture, "{0}, ", propValue.PropertyValue.ToString()));
+                        sb.Append(string.Format(CultureInfo.InvariantCulture, "{0}, ", propValuePropertyValue.ToString()));
```
iSazonov: Seems string.Format was optimized in .Net Core 2 but we could still use `AppendFormat()` here and in line 160 to get more readable code and exclude extra method calls.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
```diff
@@ -238,25 +269,26 @@ public SwitchParameter NoElement
...
-        internal static void DoGrouping(OrderByPropertyEntry currentObjectEntry, bool noElement, List<GroupInfo> groups, Dictionary<object, GroupInfo> groupInfoDictionary,
+        internal static void DoGrouping(
+            OrderByPropertyEntry currentObjectEntry,
+            bool noElement,
+            List<GroupInfo> groups,
+            Dictionary<object, GroupInfo> groupInfoDictionary,
             OrderByPropertyComparer orderByPropertyComparer)
         {
-            if (currentObjectEntry != null && currentObjectEntry.orderValues != null && currentObjectEntry.orderValues.Count > 0)
+            var currentObjectorderValues = currentObjectEntry.orderValues;
```
iSazonov: `currentObjectorderValues` should be `currentObjectOrderValues`.

iSazonov (approved):
LGTM with two minor comments.

daxian-dbw (approved):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
```diff
@@ -238,25 +269,26 @@ public SwitchParameter NoElement
...
+            if (currentObjectorderValues != null && currentObjectorderValues.Count > 0)
             {
-                object currentTupleObject = PSTuple.ArrayToTuple(currentObjectEntry.orderValues.ToArray());
+                object currentTupleObject = PSTuple.ArrayToTuple(currentObjectorderValues.ToArray());
 
-                GroupInfo currentGroupInfo = null;
-                if (groupInfoDictionary.TryGetValue(currentTupleObject, out currentGroupInfo))
+                if (groupInfoDictionary.TryGetValue(currentTupleObject, out var currentGroupInfo))
```
dantraMSFT: What's the use case for currentGroupInfo being null in the groupInfoDictionary?  If it is expected, a comment is needed; otherwise, it looks like currentObjectEntry.inputObject is dropped on the floor.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
```diff
@@ -286,12 +318,71 @@ public SwitchParameter NoElement
...
+        {
+            var currentObjectOrderValues = currentObjectEntry.orderValues;
+            if (currentObjectOrderValues != null && currentObjectOrderValues.Count > 0)
+            {
+                object currentTupleObject = PSTuple.ArrayToTuple(currentObjectOrderValues);
+
+                if (groupInfoDictionary.TryGetValue(currentTupleObject, out var currentGroupInfo))
+                {
+                    // add this inputObject to an existing group
+                    currentGroupInfo?.Add(currentObjectEntry.inputObject);
```
dantraMSFT: same as above, when is currentTobpleObject present but currentGroupInfo is null.

dantraMSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
```diff
@@ -22,28 +24,30 @@ internal static class PSTuple
...
+        /// <typeparam name="T">The first generic type parameter.</typeparam>
+        /// <param name="inputObjects">Input objects used to create a tuple.</param>
         /// <param name="startIndex">Start index of the array from which the objects have to considered for the tuple creation.</param>
         /// <returns>Tuple object.</returns>
-        internal static object ArrayToTuple(object[] inputObjects, int startIndex)
+        internal static object ArrayToTuple<T>(IList<T> inputObjects, int startIndex)
         {
             Diagnostics.Assert(inputObjects != null, "inputObjects is null");
-            Diagnostics.Assert(inputObjects.Length > 0, "inputObjects is empty");
+            Diagnostics.Assert(inputObjects.Count > 0, "inputObjects is empty");
```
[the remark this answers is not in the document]
powercode: This is a fairly common pattern in the engine, that there are asserts, but no backing runtime checks. I have always wondered about it...

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
```diff
@@ -286,12 +318,71 @@ public SwitchParameter NoElement
...
+
+        /// <summary>
+        /// Utility function called by Group-Object to create Groups.
+        /// </summary>
+        /// <param name="currentObjectEntry">Input object that needs to be grouped.</param>
+        /// <param name="noElement">True if we are not accumulating objects.</param>
+        /// <param name="groups">List containing Groups.</param>
+        /// <param name="groupInfoDictionary">Dictionary used to keep track of the groups with hash of the property values being the key.</param>
+        /// <param name="orderByPropertyComparer">The Comparer to be used while comparing to check if new group has to be created.</param>
+        internal static void DoOrderedGrouping(
```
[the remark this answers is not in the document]
powercode: It was internal, and I just kept it so, but I'm happy to make it private.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
powercode: I don't think that is a case that will ever occur. Just defensive programming.

Both cases are behaving as the original implementation. 

On could argue for removing the null check as it is a condition that shouldn't occur, and if it does, it is a bug that should be fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
powercode: Yeah, I'm removing the null checks.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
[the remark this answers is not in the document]
powercode: I guess the thinking is that since it's an internal class, the usage can be demonstrated to be safe and the error checking skipped.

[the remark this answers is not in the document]
dantraMSFT: Since we don't have an official policy around Assert, I'll not hold up the PR for it and you can ignore the rest. :)

I've been involved in three different threads recently where Dbg.Assert in internal code is hit in debug builds.  In two cases, crashes were occurring the release build and the third the lack of the check was masking a bug in another component.

IMHO: Since we don't test debug builds, Dbg.Assert is a placebo at best without a release check.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/group-object.cs
iSazonov: >  if it does, it is a bug that should be fixed.

So should we have runtime check and throw?

daxian-dbw:
@dantraMSFT Can you please take another look?

dantraMSFT (approved):
LGTM

iSazonov:
@powercode Many thanks for the great perf improvement!

PR #7413, opened 2018-07-31
Improving performance of Import-CSV

### LINKED ISSUE #7112: Performance of Windows PowerShell is better then PowerShell Core
I get very strange and disturbing performance results for Import-Csv cmdlet.

Import 111 Mb csv log file:
- On Windows PowerShell always ~ 3.7 sec. Memory ~ 10 Mb.
-On PowerShell Core from 6 sec up to 27 sec - each subsequent launch increases the time by 1-2-3 seconds. Memory grow from 1 Mb to 105 Mb. After waiting some minutes memory decreases to ~1 MB and agaim the cmdlet take ~6 sec.

I suppose we get the same for any script.

powercode:
## PR Summary
Fixes #7112 

Speed up the creation of PSObjects in CSV cmdlets by

- Use the overload Add(PSMember, bool prevalidated) for all objects except the first when adding the `NoteProperty` members.
- Add a new constructor to PSObject that preallocates the `_instanceMembers` collection with an initial capacity.
- Improve performance of AddToTypesXmlCache by avoiding an expensive copying of members just to check for the existence of one of them, perf is significantly increased, and allocations are reduced.
- Reduce allocations and GC pressure by preallocating and reusing `StringBuilders` and `List<string>` for line parsing in CSV cmdlets.
- Use `List<string>` instead of `Collection<string>` to get fewer virtual calls and better inlining.
- Reduce allocations by using a preallocated value factory in `TypeTable.GetMembers(ConsolidatedString types)`.
- Replace a `Linq.Any()` with a `List.Count > 0` in binder code.

The main gain is from taking advantage of the fact that all objects created by `Import-Csv` have the same shape (the same properties).

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
        - [x ] Issue filed - Issue link:
    - **OR**

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/powershell/7413)

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
@@ -1234,10 +1234,11 @@ namespace Microsoft.PowerShell.Commands
             {
                 TypeName = ReadTypeInformation();
             }
-
+            var values = new List<string>(16);
+            var builder = new StringBuilder(256);
             while ((Header == null) && (!this.EOF))
             {
-                Collection<string> values = ParseNextRecord();
+                ParseNextRecord(values, builder);
 
                 // Trim all trailing blankspaces and delimiters ( single/multiple ).
                 // If there is only one element in the row and if its a blankspace we dont trim it.
@@ -1279,9 +1280,11 @@ namespace Microsoft.PowerShell.Commands
             _alreadyWarnedUnspecifiedName = alreadyWriteOutWarning;
             ReadHeader();
             var prevalidated = false;
+            var values = new List<string>(16);
+            var builder = new StringBuilder(256);
             while (true)
             {
-                Collection<string> values = ParseNextRecord();
+                ParseNextRecord(values, builder);
                 if (values.Count == 0)
                     break;
 
@@ -1367,13 +1370,12 @@ namespace Microsoft.PowerShell.Commands
         /// <returns>
         /// Parsed collection of strings.
         /// </returns>
-        private Collection<string>
-        ParseNextRecord()
+        private void 
+        ParseNextRecord(List<string> result, StringBuilder current)
         {
-            // Collection of strings to return
-            Collection<string> result = new Collection<string>();
+            result.Clear();
             // current string
-            StringBuilder current = new StringBuilder();
+            current.Clear();
 
             bool seenBeginQuote = false;
             // int i = 0;
@@ -1521,8 +1523,6 @@ namespace Microsoft.PowerShell.Commands
             {
                 result.Add(current.ToString());
             }
-
-            return result;
         }
 
         // If we detect a newline we return it as a string "\r", "\n" or "\r\n"
@@ -1613,7 +1613,7 @@ namespace Microsoft.PowerShell.Commands
 
         private
         PSObject
-        BuildMshobject(string type, IList<string> names, Collection<string> values, char delimiter, bool preValidated = false)
+        BuildMshobject(string type, IList<string> names, List<string> values, char delimiter, bool preValidated = false)
         {
             //string[] namesarray = null;
             PSObject result = new PSObject(names.Count);
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
@@ -1234,8 +1234,15 @@ namespace Microsoft.PowerShell.Commands
             {
                 TypeName = ReadTypeInformation();
             }
-            var values = new List<string>(16);
-            var builder = new StringBuilder(256);
+
+            // initial sizes of the value list and the line stringbuilder
+            // set to reasonable initial sizes. They may grow beyond these,
+            // but this will prevent a few reallocations.
+            const int valueCountGuestimate = 16;
+            const int lineLengthGuestimate = 256;
+
+            var values = new List<string>(valueCountGuestimate);
+            var builder = new StringBuilder(lineLengthGuestimate);
             while ((Header == null) && (!this.EOF))
             {
                 ParseNextRecord(values, builder);
@@ -1370,7 +1377,7 @@ namespace Microsoft.PowerShell.Commands
         /// <returns>
         /// Parsed collection of strings.
         /// </returns>
-        private void 
+        private void
         ParseNextRecord(List<string> result, StringBuilder current)
         {
             result.Clear();
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
@@ -1152,8 +1152,8 @@ namespace Microsoft.PowerShell.Commands
         // Initial sizes of the value list and the line stringbuilder.
         // Set to reasonable initial sizes. They may grow beyond these,
         // but this will prevent a few reallocations.
-        private const int valueCountGuestimate = 16;
-        private const int lineLengthGuestimate = 256;
+        private const int ValueCountGuestimate = 16;
+        private const int LineLengthGuestimate = 256;
 
         internal ImportCsvHelper(PSCmdlet cmdlet, char delimiter, IList<string> header, string typeName, StreamReader streamReader)
         {
@@ -1241,8 +1241,8 @@ namespace Microsoft.PowerShell.Commands
                 TypeName = ReadTypeInformation();
             }
 
-            var values = new List<string>(valueCountGuestimate);
-            var builder = new StringBuilder(lineLengthGuestimate);
+            var values = new List<string>(ValueCountGuestimate);
+            var builder = new StringBuilder(LineLengthGuestimate);
             while ((Header == null) && (!this.EOF))
             {
                 ParseNextRecord(values, builder);
@@ -1287,8 +1287,8 @@ namespace Microsoft.PowerShell.Commands
             _alreadyWarnedUnspecifiedName = alreadyWriteOutWarning;
             ReadHeader();
             var prevalidated = false;
-            var values = new List<string>(valueCountGuestimate);
-            var builder = new StringBuilder(lineLengthGuestimate);
+            var values = new List<string>(ValueCountGuestimate);
+            var builder = new StringBuilder(LineLengthGuestimate);
             while (true)
             {
                 ParseNextRecord(values, builder);
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
@@ -1278,6 +1278,7 @@ namespace Microsoft.PowerShell.Commands
         {
             _alreadyWarnedUnspecifiedName = alreadyWriteOutWarning;
             ReadHeader();
+            var prevalidated = false;
             while (true)
             {
                 Collection<string> values = ParseNextRecord();
@@ -1290,7 +1291,8 @@ namespace Microsoft.PowerShell.Commands
                     continue;
                 }
 
-                PSObject result = BuildMshobject(TypeName, Header, values, _delimiter);
+                PSObject result = BuildMshobject(TypeName, Header, values, _delimiter, prevalidated);
+                prevalidated = true;
                 _cmdlet.WriteObject(result);
             }
             alreadyWriteOutWarning = _alreadyWarnedUnspecifiedName;
@@ -1611,10 +1613,10 @@ namespace Microsoft.PowerShell.Commands
 
         private
         PSObject
-        BuildMshobject(string type, IList<string> names, Collection<string> values, char delimiter)
+        BuildMshobject(string type, IList<string> names, Collection<string> values, char delimiter, bool preValidated = false)
         {
             //string[] namesarray = null;
-            PSObject result = new PSObject();
+            PSObject result = new PSObject(names.Count);
             char delimiterlocal = delimiter;
             int unspecifiedNameIndex = 1;
             for (int i = 0; i <= names.Count - 1; i++)
@@ -1635,7 +1637,8 @@ namespace Microsoft.PowerShell.Commands
                 {
                     value = values[i];
                 }
-                result.Properties.Add(new PSNoteProperty(name, value));
+
+                result.Properties.Add(new PSNoteProperty(name, value), preValidated);
             }
 
             if (!_alreadyWarnedUnspecifiedName && unspecifiedNameIndex != 1)
@@ -1644,7 +1647,7 @@ namespace Microsoft.PowerShell.Commands
                 _alreadyWarnedUnspecifiedName = true;
             }
 
-            if (type != null && type.Length > 0)
+            if (!string.IsNullOrEmpty(type))
             {
                 result.TypeNames.Clear();
                 result.TypeNames.Add(type);
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
@@ -1149,6 +1149,12 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         private readonly StreamReader _sr;
 
+        // Initial sizes of the value list and the line stringbuilder.
+        // Set to reasonable initial sizes. They may grow beyond these,
+        // but this will prevent a few reallocations.
+        private const int valueCountGuestimate = 16;
+        private const int lineLengthGuestimate = 256;
+
         internal ImportCsvHelper(PSCmdlet cmdlet, char delimiter, IList<string> header, string typeName, StreamReader streamReader)
         {
             if (cmdlet == null)
@@ -1235,12 +1241,6 @@ namespace Microsoft.PowerShell.Commands
                 TypeName = ReadTypeInformation();
             }
 
-            // initial sizes of the value list and the line stringbuilder
-            // set to reasonable initial sizes. They may grow beyond these,
-            // but this will prevent a few reallocations.
-            const int valueCountGuestimate = 16;
-            const int lineLengthGuestimate = 256;
-
             var values = new List<string>(valueCountGuestimate);
             var builder = new StringBuilder(lineLengthGuestimate);
             while ((Header == null) && (!this.EOF))
@@ -1287,8 +1287,8 @@ namespace Microsoft.PowerShell.Commands
             _alreadyWarnedUnspecifiedName = alreadyWriteOutWarning;
             ReadHeader();
             var prevalidated = false;
-            var values = new List<string>(16);
-            var builder = new StringBuilder(256);
+            var values = new List<string>(valueCountGuestimate);
+            var builder = new StringBuilder(lineLengthGuestimate);
             while (true)
             {
                 ParseNextRecord(values, builder);
--- a/src/System.Management.Automation/engine/MshMemberInfo.cs
+++ b/src/System.Management.Automation/engine/MshMemberInfo.cs
@@ -4019,6 +4019,14 @@ namespace System.Management.Automation
             _members = new OrderedDictionary(StringComparer.OrdinalIgnoreCase);
         }
 
+        /// <summary>
+        /// Constructs this collection with an initial capacity
+        /// </summary>
+        internal PSMemberInfoInternalCollection(int capacity)
+        {
+            _members = new OrderedDictionary(capacity, StringComparer.OrdinalIgnoreCase);
+        }
+
         private void Replace(T oldMember, T newMember)
         {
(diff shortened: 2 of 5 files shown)


iSazonov:
@powercode Thanks for great PR!

I see that most of the changes (over 95%!) are style. It would be nice if you left only meaningful changes. (I hope you understand the difference between reviewing 1300 and 50 lines of code.)

powercode:
@iSazonov Well, they are separated in different commits.

SteveL-MSFT:
@powercode Thanks for this contribution!  I've fixed your PR description so that the checkboxes show up correctly.  In the future, you can just leave the checkboxes empty then after you submit your PR, you can use your mouse to click them and it'll update the markdown correctly.

SteveL-MSFT:
Since the cleanup is in separate commits, it shouldn't be too difficult to review.  In the future, it'll be better to have the cleanup as a separate PR, although I can certainly understand making the cleanup while making code changes in the same source files :)

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -4894,6 +4902,7 @@ public void Dispose()
         }
     }
 
+
```
SteveL-MSFT: Extra whitespace

src/System.Management.Automation/engine/MshObject.cs
```diff
@@ -492,6 +492,17 @@ public PSObject()
...
+        /// Initializes a new instance of PSObject with an PSCustomObject BaseObject
+        /// with an initial capacity for members
+        /// </summary>
+        /// <param name="instanceMemberCapacity">The initial capacity for the instance member collection.</param>
+        public PSObject(int instanceMemberCapacity) : this()
+        {
+            _instanceMembers = new PSMemberInfoInternalCollection<PSMemberInfo>(instanceMemberCapacity);
+        }
+
+
```
SteveL-MSFT: extra whitespace

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

powercode:
@SteveL-MSFT I did some additional perf work (profiled allocations).

Is it ok to submit a few more commits in this PR?

SteveL-MSFT:
@powercode yes, feel free to submit more commits, will re-review

iSazonov:
Our code guidance demands to separate style and significant changes. Please remove the style commits from the PR - they very large and contain controversial changes. We can discuss them in another PR.

iSazonov:
In the context the line https://github.com/PowerShell/PowerShell/blob/5d03e1653a7d518715fa3f00587cad6b5c78cc89/src/System.Management.Automation/engine/MshMemberInfo.cs#L4561
looks as unneeded although this may be necessary in other scenarios.

powercode:
@PowerShellTeam I'm really getting sick on this endless focus on formatting.

This is a **10x** perf improvement on Import-CSV, and I have spend about as much time on this that I'm willing to spend. 

Feel free to adopt the PR, or ditch it.

iSazonov:
@powercode I am very surprised by "I'm really getting sick ...".
I hope you read our [Coding Guidelines](https://github.com/PowerShell/PowerShell/blob/master/docs/dev-process/coding-guidelines.md#coding-conventions).
_ALL_ contributors (community and MFST) get requests to follow the guideline. (And all the contributors are not happy to receive such requests.)
Any maintainer "SHOULD make sure contributors are following the contributor guidelines" - see [Repository Maintainers docs](https://github.com/PowerShell/PowerShell/tree/master/docs/maintainers).
Again - any maintainer SHOULD. It is not personal requests. It is _well-driven dev process_. We do not want to create chaos in our code.

I absolutely can not accept your "I'm really getting sick". You pushed over 1300 line changes and only 30-50 of them are significant. This absolutely contradicts our Coding Guidelines. This forces reviewers to spend a lot of time looking at insignificant changes. It would be nice if you took care not only of your time. As I wrote earlier if you are a very busy person, then you can always ask a help from maintainers and other contributors and say "I will be busy the whole next month, feel free grab my commits or wait for me to return". This is more creative than writing "I'm really getting sick". 

In recent months, we have done a great job of correcting style issues in the repo in order to facilitate the work of the contributors. And we continue. We are configuring CodeFactor/StyleCop. We have started to use CodeFormatter (a tooling you ask). We have PR for adding `.editorconfig` for VS. 
All this takes a lot of our time, but as a result, it saves _your_ time.

I hope for your understanding, patience and help.

powercode:
Well, I beg to differ. I'm happy to run tools to keep a consistent formatting on *files that already confirms to the standard*. But it is a huge waste of peoples time to nitpick on things that could easily be fixed by tools.

iSazonov:
@powercode The problem is that there is no tool that would fix everything we need.
In addition, fixing which particular problem leads to a change sometimes thousands of lines of code. _This takes a long time to review_. You can see that CodeFactor reports up to 100000 issues. We plan to disable some rules but still we will need a lot of effort and time to complete this work.  Before we finish we are forced to follow the rule - keep the file style.
I tried CodeFormatter on engine folder and get so large PR that we need dozens of reviewers. 
I believe we are moving in the right direction, albeit slowly. So please be patient. 
I really want to make your work comfortable.

SteveL-MSFT:
The best practice is to have separate PRs for style changes than functional changes.  We should try to do that as it makes reviews easier.  I also appreciate @iSazonov's efforts to help clean up the code base.

@powercode I appreciate your contributions and hope you'll continue to contribute, but do try to have style changes as separate PRs next time.

For this specific PR, I don't think it's worth the effort to cherry-pick the style changes to a new PR.  CodeFactor shows 313 issues fixed and when I looked at those specific commits, the style changes all looked good.

TravisEz13 (commented):
> Reviewed 2 of 4 files at r2, 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/powershell/powershell/7413)* status: 0 of 5 files reviewed

src/System.Management.Automation/engine/TypeTable.cs
```diff
@@ -3392,6 +3394,7 @@ internal TypeTable(bool isShared)
         {
             this.isShared = isShared;
             _typeFileList = new List<string>();
+            _memberFactoryFunc = MemberFactory;
```
daxian-dbw: Is the filed `_memberFactoryFunc` really needed? Why not directly use `_consolidatedMembers.GetOrAdd(types.Key, MemberFactory, types)` instead?

daxian-dbw: Got it. You are trying to avoid constructing a `Func<...>` object every time running this method. Nice thought!

There are 4 constructors for `TypeTable`, and adding `_memberFactoryFunc = MemberFactory` here only covers 2 of them, can you add this initialization to the other constructors?

daxian-dbw: Also, if would be helpful to add a comment to the `_memberFactoryFunc` field to explain its purpose. If you cannot get to this PR in the short term, please let me know, I'm happy to make these minor changes for you.

powercode: @daxian-dbw Done. Yes, I profiled allocations and found that this was called so frequently, and generated lots of short lived delegates.

powercode: Nice catch with the ctors - I misread the code and thought they all delegated to a single one. They do now.

daxian-dbw:
Quoted from [`Coding Conventions`](https://github.com/PowerShell/PowerShell/blob/master/docs/dev-process/coding-guidelines.md#coding-conventions):
> please avoid reformatting any existing code when submitting a PR as it obscures the functional changes of the PR. A separate PR should be submitted for style-only changes.

It's OK to have a small set of style changes along with functional changes in a PR, as long as they are in separate commits. But for a large set of style changes, the _best practice_ is to separate them into another PR.

We have this guideline in place actually trying to make the life easier for both the author and reviewer.
- For the reviewer, it's probably harder and cost more time to review such a PR.
- For the author, there are two possibilities when submitting a PR with large set of style changes mixed with functional changes.
   1. No one wants to spend the time going through all the changes, and thus the PR gets left behind with no approval for a long time (I saw this happen before). This will very likely hurt the author's feeling and discourage his/her further contribution.
   1. Since there are large set of style changes, it will be inevitable for reviewers to leave comments about the style changes, which would (1) deviate attention/focus from the more important functional changes; (2) potentially cause dispute because different people may see a style issue differently (you know, it has always been that way when it comes to style and formatting 😄). In the end, this might also discourage the author's further contribution.

That is the context why we have this guideline in place. @powercode Hope you can understand, and thanks for your understanding! :)

However, on the other hand, I totally understand that when making changes in unfamiliar code, style changes may have to be made to understand the code before making changes. This often makes the functional changes depend on the style changes that have already been made, and thus makes it hard to separate the style changes to another PR, especially if the author didn't have the `style-only-changes-in-separate-PR` idea in mind at the very beginning.

It's a blurry line there, and **I think in such a case, as long as style changes are in separate commits, we can be more accepting of the PR while reminding the author to have the `style-only-changes-in-separate-PR` idea in mind for the future contributions.** As reviewers, [`Reviewable`](https://reviewable.io) could make reviewing easier for such a PR.

@iSazonov has spent tons of effort cleaning up the code base and unifying the style/formatting to make the code base more readable and maintainable (I really appreciate your work @iSazonov!). But we are not quite there yet, the tools, rules, and guidance are still work-in-progress and we are looking for input. It's an on-going learning experience for all of us maintainers.

daxian-dbw (approved):

iSazonov:
@daxian-dbw The first two style commits contain controversial changes and gaps. I vote to remove them from the PR. I can grab the commits and open new PR.

daxian-dbw:
> The first two style commits contain controversial changes and gaps.

@iSazonov Can you point out the changes that you are concerned with? You don't need to point out every one of them. One example per category would be good enough.

iSazonov (commented):
@daxian-dbw I can not comment first commit - GitHub returns an error. I see there some style issues too.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -193,13 +213,7 @@ internal void CloneBaseProperties(PSMemberInfo destiny)
...
         /// Gets the member name
         /// </summary>
-        public string Name
-        {
-            get
-            {
-                return this.name;
-            }
-        }
+        public string Name => this.name;
```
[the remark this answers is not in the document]
daxian-dbw: It's a new language feature introduced in C# 7.0 - expression-bodied members. Why can't we use this new language feature?

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -489,29 +500,28 @@ public override bool IsGettable
...
                 return false;
             }
         }
 
         private PSMemberInfo LookupMember(string name)
         {
-            bool hasCycle;
-            PSMemberInfo returnValue;
-            LookupMember(name, new HashSet<string>(StringComparer.OrdinalIgnoreCase), out returnValue, out hasCycle);
+            LookupMember(name, new HashSet<string>(StringComparer.OrdinalIgnoreCase), out var returnValue, out var hasCycle);
```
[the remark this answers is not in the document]
daxian-dbw: Yeah, I noticed those instances, and was planning on fixing them in a separate PR.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -536,18 +546,19 @@ private void LookupMember(string name, HashSet<string> visitedAliases, out PSMem
                     name);
             }
 
-            PSAliasProperty aliasMember = member as PSAliasProperty;
-            if (aliasMember == null)
+            if (!(member is PSAliasProperty aliasMember))
```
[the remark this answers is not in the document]
daxian-dbw: Yeah, I also noticed this, and plan to fix it. (as we discussed in another PR)

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -907,9 +914,10 @@ public override object Value
                         ExtendedTypeSystem.SetWithoutSetterException,
                         this.Name);
                 }
+
                 try
                 {
-                    SetterCodeReference.Invoke(null, new object[2] { this.instance, value });
+                    SetterCodeReference.Invoke(null, new object[] {this.instance, value});
```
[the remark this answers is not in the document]
daxian-dbw: This is not a bug. The length of the array can be inferred from the initializer following `new object[]`. You can even do `new [] { ... }`, though I think this style is less preferred (BTW, I also notice one instance of this style in the changes).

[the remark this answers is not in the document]
powercode: @daxian-dbw Why do you think that style is less preferred?

To me, this is the same discussion we had in the C++ community, where people thought `auto` was bad. Herb Sutter wrote this:
https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/

Try to come up with a situation where 
```
 SetterCodeReference.Invoke(null, new object[] {this.instance, value});
```
is easier to understand than
```
 SetterCodeReference.Invoke(null, new [] {this.instance, value});

```

Types are for the compiler to verify that we are not messing up our logic. 
The essense of both lines are that we are invoking a Setter for a code reference, passing in the parameters. We know that the compiler is happy with the call. We guess that the array is the instance to set the value on, and the actual value. 

This comes from good variable naming.

I would argue that explicit types has fairly limited value for readablility, and gives a lot of churn when doing refactorings.

src/System.Management.Automation/engine/MshMemberInfo.cs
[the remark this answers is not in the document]
iSazonov: The main problem we face is that historically many files have different formatting and style.
This complicates both reading the code and changing it. We are working on fixing this.
If we continue to create different styles, we will never bring our code into good condition.
So if we want to use the new feature, we need to first add it to our codinig guidelines, and then fix the entire code base (in a few commits).

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -1278,9 +1279,12 @@ internal void ReadHeader()
         {
             _alreadyWarnedUnspecifiedName = alreadyWriteOutWarning;
             ReadHeader();
+            var prevalidated = false;
+            var values = new List<string>(16);
+            var builder = new StringBuilder(256);
```
iSazonov: @powercode Please explain why we use this init constants? (Why not 32 and 512?) I believe we need to add a good comment.

Also above we use the same values. I believe we should use `readonly const`.

iSazonov (commented):
@daxian-dbw I could not able comment in first commit so I added comments in full list. (Not all issues was marked as you ask)

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -1412,16 +1390,13 @@ public override string ToString()
         /// <param name="variable">The variable to wrap</param>
         /// <exception cref="ArgumentException">for an empty or null name</exception>
         public PSVariableProperty(PSVariable variable)
-            : base(variable != null ? variable.Name : null, null)
+            : base(variable?.Name, null)
```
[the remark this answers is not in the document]
powercode: I would argue that 
```
base(variable?.Name, null) 
```

is more readable than
```
base(variable != null ? variable.Name : null, null)
```

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
powercode: They are just reasonable guestimates of number of columns and line length - just so that we don't have to resize and reallocate so much.

Agree that we can make constants of them.

src/System.Management.Automation/engine/MshMemberInfo.cs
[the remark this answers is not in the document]
powercode: The first line is read *the Name of variable or null*.
The second line has to be parsed.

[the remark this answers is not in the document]
powercode: And from our coding guidelines:
> Use of new C# language syntax is encouraged.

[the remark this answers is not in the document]
powercode: The syntax was introduced to address just this scenario. Once you have become used to it, it reads a lot better than the old, verbose, syntax.

src/System.Management.Automation/engine/MshMemberInfo.cs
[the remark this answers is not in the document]
powercode: There are so many places is code where you don't see the type. You just don't notice it.

Take this example:

```csharp
void Foo(Bar bar) 
{
    _member.GetInstance(out var instance);
    DoWork(GetData(), _context.Position, bar.Size, instance);
}
```
Of the arguments to DoWork, why do you consider not seeing the type of `instance` as a problem, and not the types of `GetData()`, `_context.Position`, and `bar.Size`?

It makes no sence - In that case, the guideline would have to be to write this as 

```csharp
void Foo(Bar bar) 
{
    _member.GetInstance(out MyInstance instance);
    SomeData someData = GetData();
    XPosition position = _context.Position;
    StrangeSize strangeSize = bar.Size;
    DoWork( someData, position, strangeSize, instance);
}
```

Most of us would reject that as crazy. 

Why do you make a different call for `out var`?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
powercode: Fixing.

iSazonov: I meant that the constants is used in ReadHeader() and Import().
After we read header we already know needed size of the `values` (header column count) and could use it in Import().
ReadHeader() is called once - seems make sense use init value only if we get a header from Header parameter - in the case we know size of the header.

daxian-dbw:
Given that the discussion has been more focused on coding styles instead of the functional improvement, I will reorganize this PR to separate the first two commits to another PR.
@powercode I hope you don't mind, thanks!

powercode:
@daxian-dbw :) I have no objections what-so-ever to people changing code I submit. Less work for me!

I was under the naive impression that the changes was non-controversial, since they were *only formatting* changes, and simplifications that took advantage of more consise ways of expressing the same meaning introduced by newer language versions. 

In the world I work, a *formatting only* change is mostly ignored, and the focus is on the bits that affect the customers, so I didn't perceive it as something that would generate much more work for you guys.

I get that the team is more worried about how code style affect readability than I am - so far I have always been able to read the code, even if the style vary widely, and I frankly don't consider it on the top 10 list of things to prioritize on the project.

What I *do* care about is finding the real issues in the code - like the constructor bug I had introduced. 

I actually do think it is a good workflow to start with a refactoring pass when you are about to do a change to some code. By that I mean a pass where no functionality is changed, only formatting and the way things are expressed. It may be introduction of new methods, etc, but functionally the same.

Having the requirement that that pass should be in it's own PR seems a bit process-heavy to me, but I that is how you want to roll, I'll adapt.

Just my €.02.

daxian-dbw:
OK, I have separated the original first 2 commits to #7446 
All rest commits are kept in this PR. I also added another commit to make `Import` method also use the const int values.

@powercode Please take a look just to make sure I didn't mess up anything :)
@iSazonov @SteveL-MSFT Can you please review again? Thanks!

daxian-dbw (approved):

SteveL-MSFT (approved):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/CSVCommands.cs
```diff
@@ -1235,9 +1241,11 @@ internal void ReadHeader()
                 TypeName = ReadTypeInformation();
             }
 
+            var values = new List<string>(valueCountGuestimate);
+            var builder = new StringBuilder(lineLengthGuestimate);
```
iSazonov: We should exclude the constants here because we read a header once per file - make no sense do the optimization.

After we read the header or get it from Header parameter we know the size of the header and `values` - so we could use the size in `Import()` method below.

As for 256 const in StringBuilder I agree for the PR. Later I want add optimizations in the parser (if you already haven't such plans).

And sorry... `lineLengthGuestimate` should start with upper case `LineLengthGuestimate`.

powercode: Why does it make more sense to do extra work when it can be avoided?

iSazonov (approved):
LGTM with one minor comment.

daxian-dbw:
@powercode Thanks again for the great improvement!

iSazonov:
@powercode Thanks for the great and very useful improvement! I often process logs and I are happy to get a quick Import-Csv cmdlet.  Sorry again for noisy style comments.

@SteveL-MSFT This can be an interesting blog post.

PR #7418, opened 2018-08-01
Improving performance of LanguagePrimitives.TryConvertTo.

### LINKED ISSUE #7420: Improve performance of LanguagePrimitives.TryConvertTo
`LanguagePrimitives.TryConvertTo`  is implemented by wrapping `LanguagePrimitives.ConvertTo` in exception handling, and return false if it failed.

This can be improved upon by using the `ConversionData` return from `FigureConversion` by checking if the `Rank` is `ConversionRank.None` before we try to call the conversion. 

By doing an early exit, return we avoid the exceptions that are always thrown when there are no available conversions.

powercode:
## PR Summary

TryConvertTo is implemented in terms of Convert, but with wrapping exception handling. 
This is costly performance wise, so by inlining the implementation of ConvertTo, and getting access to the FigureConversion output. Using that we can determine upfront on at least some occations that a conversion is not available.

Fixes #7420 

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/LanguagePrimitives.cs
+++ b/src/System.Management.Automation/engine/LanguagePrimitives.cs
@@ -801,9 +801,7 @@ namespace System.Management.Automation
                 }
             }
 
-            string firstString = first as string;
-
-            if (firstString != null)
+            if (first is string firstString)
             {
                 string secondString = second as string;
                 if (secondString == null)
@@ -834,6 +832,7 @@ namespace System.Management.Automation
             try
             {
                 secondConverted = LanguagePrimitives.ConvertTo(second, firstType, culture);
+
             }
             catch (PSInvalidCastException e)
             {
@@ -841,9 +840,7 @@ namespace System.Management.Automation
                                                          first.ToString(), second.ToString(), e.Message);
             }
 
-            IComparable firstComparable = first as IComparable;
-
-            if (firstComparable != null)
+            if (first is IComparable firstComparable)
             {
                 return firstComparable.CompareTo(secondConverted);
             }
@@ -1646,9 +1643,9 @@ namespace System.Management.Automation
         /// <returns></returns>
         public static T ConvertTo<T>(object valueToConvert)
         {
-            if (valueToConvert is T)
+            if (valueToConvert is T variable)
             {
-                return (T)valueToConvert;
+                return variable;
             }
             return (T)ConvertTo(valueToConvert, typeof(T), true, CultureInfo.InvariantCulture, null);
         }
@@ -1664,9 +1661,9 @@ namespace System.Management.Automation
         /// <returns>false for conversion failure, true for success</returns>
         public static bool TryConvertTo<T>(object valueToConvert, out T result)
         {
-            if (valueToConvert is T)
+            if (valueToConvert is T variable)
             {
-                result = (T)valueToConvert;
+                result = variable;
                 return true;
             }
             return TryConvertTo(valueToConvert, CultureInfo.InvariantCulture, out result);
@@ -1684,10 +1681,16 @@ namespace System.Management.Automation
         /// <returns>false for conversion failure, true for success</returns>
         public static bool TryConvertTo<T>(object valueToConvert, IFormatProvider formatProvider, out T result)
         {
-            result = default(T);
+            result = default;
             try
             {
-                result = (T)ConvertTo(valueToConvert, typeof(T), formatProvider);
+                if (TryConvertTo(valueToConvert, typeof(T), formatProvider, out var res))
+                {
+                    result = (T) res;
+                    return true;
+                }
+
+                return false;
             }
             catch (InvalidCastException)
             {
@@ -1697,7 +1700,6 @@ namespace System.Management.Automation
             {
                 return false;
             }
-            return true;
         }
 
         /// <summary>
@@ -1729,20 +1731,31 @@ namespace System.Management.Automation
         /// <returns>false for conversion failure, true for success</returns>
         public static bool TryConvertTo(object valueToConvert, Type resultType, IFormatProvider formatProvider, out object result)
         {
-            result = null;
+            result = default;
             try
             {
-                result = ConvertTo(valueToConvert, resultType, formatProvider);
+                using (typeConversion.TraceScope("Converting \"{0}\" to \"{1}\".", valueToConvert, resultType))
+                {
+                    if (resultType == null)
+                    {
+                        return false;
+                    }
+
+                    var conversion = FigureConversion(valueToConvert, resultType, out var debase);
+                    if (conversion.Rank== ConversionRank.None)
+                    {
+                        return false;
+                    }
+                    result = conversion.Invoke(debase ? PSObject.Base(valueToConvert) : valueToConvert,
+                        resultType, true, debase ? (PSObject)valueToConvert : null,
+                        formatProvider, null);
+                    return true;
+                }
             }
             catch (InvalidCastException)
             {
                 return false;
             }
-            catch (ArgumentException)
-            {
-                return false;
-            }
-            return true;
         }
 
         #endregion public type conversion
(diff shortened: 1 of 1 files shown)


src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -1729,20 +1731,31 @@ public static bool TryConvertTo(object valueToConvert, Type resultType, out obje
...
+                    }
+
+                    var conversion = FigureConversion(valueToConvert, resultType, out var debase);
+                    if (conversion.Rank== ConversionRank.None)
+                    {
+                        return false;
+                    }
+                    result = conversion.Invoke(debase ? PSObject.Base(valueToConvert) : valueToConvert,
+                        resultType, true, debase ? (PSObject)valueToConvert : null,
+                        formatProvider, null);
```
iSazonov: Please fix formatting:
```c#
result = conversion.Invoke(
             debase ? PSObject.Base(valueToConvert) : valueToConvert,
             resultType,
             recurse: true,
...

```

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -1729,20 +1731,31 @@ public static bool TryConvertTo(object valueToConvert, Type resultType, out obje
...
-                result = ConvertTo(valueToConvert, resultType, formatProvider);
+                using (typeConversion.TraceScope("Converting \"{0}\" to \"{1}\".", valueToConvert, resultType))
+                {
+                    if (resultType == null)
+                    {
+                        return false;
+                    }
+
+                    var conversion = FigureConversion(valueToConvert, resultType, out var debase);
+                    if (conversion.Rank== ConversionRank.None)
```
iSazonov: Please add space before `==`.

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -1646,9 +1643,9 @@ public static object ConvertPSObjectToType(PSObject valueToConvert, Type resultT
         /// <returns></returns>
         public static T ConvertTo<T>(object valueToConvert)
         {
-            if (valueToConvert is T)
+            if (valueToConvert is T variable)
             {
-                return (T)valueToConvert;
+                return variable;
             }
             return (T)ConvertTo(valueToConvert, typeof(T), true, CultureInfo.InvariantCulture, null);
```
iSazonov: Make sense to use `return ConvertTo( ... ) as T` as more fast?

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -1684,10 +1681,16 @@ public static bool TryConvertTo<T>(object valueToConvert, out T result)
         /// <returns>false for conversion failure, true for success</returns>
         public static bool TryConvertTo<T>(object valueToConvert, IFormatProvider formatProvider, out T result)
         {
-            result = default(T);
+            result = default;
```
iSazonov: I'd keep old code as more readable for the generic method.

iSazonov (commented):
It would be nice to see examples that we really get a win.

src/System.Management.Automation/engine/LanguagePrimitives.cs
daxian-dbw: I think `ConvertTo() as T` is not faster. And `(T)ConvertTo()` ensures an exception to be thrown when `ConverTo()` doesn't really return a `T` object, which is the expected behavior.

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -1684,10 +1681,16 @@ public static bool TryConvertTo<T>(object valueToConvert, out T result)
...
             try
```
daxian-dbw: The `try/catch` can be removed, as the exceptions are taken care of in `TryConvertTo(valueToConvert, typeof(T), formatProvider, out var res)`.

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -1729,20 +1731,31 @@ public static bool TryConvertTo(object valueToConvert, Type resultType, out obje
         /// <returns>false for conversion failure, true for success</returns>
         public static bool TryConvertTo(object valueToConvert, Type resultType, IFormatProvider formatProvider, out object result)
         {
-            result = null;
+            result = default;
```
daxian-dbw: I think `result = null` is more explicit here, as the type of `result` is `object` here.

src/System.Management.Automation/engine/LanguagePrimitives.cs
iSazonov: I think all exceptions occur in `ConvertTo(valueToConvert, typeof(T), true, CultureInfo.InvariantCulture, null);` and if the method returns a result then the result is always of type T. So formally we can use `as T`. I see there are performance comparisons of  `cast` and `as` operators - last is faster - but I don't know whether it is true for .Net Core 2.1.

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -1646,9 +1641,9 @@ public static object ConvertPSObjectToType(PSObject valueToConvert, Type resultT
         /// <returns></returns>
         public static T ConvertTo<T>(object valueToConvert)
         {
-            if (valueToConvert is T)
+            if (valueToConvert is T variable)
             {
-                return (T)valueToConvert;
+                return variable;
```
iSazonov: Please use `value` name as we do below.

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -1685,19 +1680,14 @@ public static bool TryConvertTo<T>(object valueToConvert, out T result)
...
-            catch (InvalidCastException)
-            {
-                return false;
-            }
-            catch (ArgumentException)
+
+            if (TryConvertTo(valueToConvert, typeof(T), formatProvider, out var res))
             {
-                return false;
+                result = (T) res;
```
iSazonov: Please remove extra space  `(T) res` -> `(T)res`.

src/System.Management.Automation/engine/LanguagePrimitives.cs
iSazonov: Please address the comment and remove extra space 
`conversion.Rank== ConversionRank.None` -> `conversion.Rank == ConversionRank.None`

src/System.Management.Automation/engine/LanguagePrimitives.cs
iSazonov: Please address the comment.

iSazonov (commented):
Please look CodeFactor issues.

iSazonov (approved):

daxian-dbw (approved):
LGTM

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -801,9 +801,7 @@ public static int Compare(object first, object second, bool ignoreCase, IFormatP
                 }
             }
 
-            string firstString = first as string;
-
-            if (firstString != null)
+            if (first is string firstString)
             {
                 string secondString = second as string;
                 if (secondString == null)
```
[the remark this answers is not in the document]
daxian-dbw: @iSazonov I think we should be careful about following one pattern too strictly. I think `as string` is better here because the `if` condition is `secondString == null`.  Changing it to `if (!(second is string secondString))` would make it less readable, which goes against our goal.

daxian-dbw:
@powercode I made a minor update to a few style issues. I hope you don't mind. They are as follows:
- Address @iSazonov's comment about the rename `variable` to `value` to be consistent with other instances.
- Change `out var` to `out object` and `out bool` to make the type of the out parameter a little more explicit.
- Change the indentation of the parameters for the call to `conversion.Invoke`.

Thank you again for the great contribution!

src/System.Management.Automation/engine/LanguagePrimitives.cs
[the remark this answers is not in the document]
iSazonov: Using different patterns (especially next to each other) also reduces readability - I just stumbled over it.
My suggestion was just to roll back the change at all to old pattern or use one pattern.

PR #7434, opened 2018-08-02
Throw error on entering non-existing type in Get-FormatData

### LINKED ISSUE #4235: Get-FormatData quietly ignores non-existing type-name arguments
Steps to reproduce
------------------

```powershell
Get-FormatData NoSuch.Type.Exists.Or.IsLoaded
```

Expected behavior
-----------------

An error that indicates that the type does not exist (in the current session).

Actual behavior
---------------

A quiet no-op. No output, `$?` reports `$True`.

Environment data
----------------

```powershell
PowerShell Core v6.0.0-beta.3 on macOS 10.12.5
PowerShell Core v6.0.0-beta.3 on Ubuntu 16.04.1 LTS
PowerShell Core v6.0.0-beta.3 on Microsoft Windows 10 Pro (64-bit; v10.0.14393)
Windows PowerShell v5.1.15063.413 on Microsoft Windows 10 Pro (64-bit; v10.0.15063)
```

faraazahmad:
Fix #4235 

## PR Summary

Write non-terminating error when `Get-FormatData` doesn't found a type definition for the provided type,

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - [] User-facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
        - [x] Issue filed - Issue link: #4235 
    - **OR**

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
@@ -171,38 +171,40 @@ namespace Microsoft.PowerShell.Commands
                 viewList.Add(formatdef);
             }// foreach(ViewDefinition...
 
+            // If no Type Definitions are found for _typename throw a terminating error
             if (typedefs.Count == 0) {
-                System.ArgumentException exception = new System.ArgumentException("No such type");
                 // All files must have the same extension otherwise throw.
                 ErrorRecord errorRecord = new ErrorRecord(
-                    exception,
+                    new System.ArgumentException("No such type"),
                     "SPECIFIED_TYPE_NOT_FOUND",
                     ErrorCategory.ResourceUnavailable,
                     _typename);
 
                 ThrowTerminatingError(errorRecord);
             }
-            // write out all the available type definitions
-            foreach (var pair in typedefs)
-            {
-                var typeNames = pair.Key;
-
-                if (writeOldWay)
+            else {
+                // write out all the available type definitions
+                foreach (var pair in typedefs)
                 {
-                    foreach (var typeName in typeNames)
+                    var typeNames = pair.Key;
+
+                    if (writeOldWay)
                     {
-                        var etd = new ExtendedTypeDefinition(typeName, pair.Value);
+                        foreach (var typeName in typeNames)
+                        {
+                            var etd = new ExtendedTypeDefinition(typeName, pair.Value);
+                            WriteObject(etd);
+                        }
+                    }
+                    else
+                    {
+                        var etd = new ExtendedTypeDefinition(typeNames[0], pair.Value);
+                        for (int i = 1; i < typeNames.Count; i++)
+                        {
+                            etd.TypeNames.Add(typeNames[i]);
+                        }
                         WriteObject(etd);
                     }
-                }
-                else
-                {
-                    var etd = new ExtendedTypeDefinition(typeNames[0], pair.Value);
-                    for (int i = 1; i < typeNames.Count; i++)
-                    {
-                        etd.TypeNames.Add(typeNames[i]);
-                    }
-                    WriteObject(etd);
                 }
             }
         }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using Microsoft.PowerShell.Commands.Internal.Format;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
@@ -9,6 +8,7 @@ using System.Linq;
 using System.Management.Automation;
 using System.Management.Automation.Internal;
 using System.Management.Automation.Runspaces;
+using Microsoft.PowerShell.Commands.Internal.Format;
 
 namespace Microsoft.PowerShell.Commands
 {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
@@ -173,7 +173,7 @@ namespace Microsoft.PowerShell.Commands
             if (typedefs.Count == 0)
             {
                 ErrorRecord errorRecord = new ErrorRecord(
-                    new TypeLoadException(GetFormatDataStrings.SpecifiedTypeNotFound),
+                    new TypeLoadException(StringUtil.Format(GetFormatDataStrings.SpecifiedTypeNotFound, _typename),
                     "SpecifiedTypeNotFound",
                     ErrorCategory.InvalidOperation,
                     _typename);
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
@@ -175,11 +175,10 @@ namespace Microsoft.PowerShell.Commands
             if (typedefs.Count == 0) {
                 // All files must have the same extension otherwise throw.
                 ErrorRecord errorRecord = new ErrorRecord(
-                    new System.ArgumentException("No such type"),
+                    new System.Exception("No such type"),
                     "SPECIFIED_TYPE_NOT_FOUND",
                     ErrorCategory.ResourceUnavailable,
                     _typename);
-
                 ThrowTerminatingError(errorRecord);
             }
             else {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
@@ -1,15 +1,14 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using Microsoft.PowerShell.Commands.Internal.Format;
 using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Management.Automation;
 using System.Management.Automation.Internal;
-
-using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
 using System.Management.Automation.Runspaces;
-using Microsoft.PowerShell.Commands.Internal.Format;
 
 namespace Microsoft.PowerShell.Commands
 {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
@@ -4,6 +4,8 @@
 using System;
 using System.Linq;
 using System.Management.Automation;
+using System.Management.Automation.Internal;
+
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Management.Automation.Runspaces;
@@ -169,6 +171,17 @@ namespace Microsoft.PowerShell.Commands
                 viewList.Add(formatdef);
             }// foreach(ViewDefinition...
 
+            if (typedefs.Count == 0) {
+                System.ArgumentException exception = new System.ArgumentException("No such type");
+                // All files must have the same extension otherwise throw.
+                ErrorRecord errorRecord = new ErrorRecord(
+                    exception,
+                    "SPECIFIED_TYPE_NOT_FOUND",
+                    ErrorCategory.ResourceUnavailable,
+                    _typename);
+
+                ThrowTerminatingError(errorRecord);
+            }
             // write out all the available type definitions
             foreach (var pair in typedefs)
             {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
@@ -3,8 +3,8 @@
 
 using System;
 using System.Linq;
-using System.Management.Automation;
 using System.Management.Automation.Internal;
+using System.Management.Automation;
 
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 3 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
```diff
@@ -169,27 +171,41 @@ protected override void ProcessRecord()
                 viewList.Add(formatdef);
             }// foreach(ViewDefinition...
 
-            // write out all the available type definitions
-            foreach (var pair in typedefs)
+            // If no Type Definitions are found for _typename throw a terminating error
```
iSazonov: Please remove obvious comment.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
```diff
@@ -169,27 +171,41 @@ protected override void ProcessRecord()
...
+                ErrorRecord errorRecord = new ErrorRecord(
+                    new System.Exception("No such type"),
+                    "SPECIFIED_TYPE_NOT_FOUND",
+                    ErrorCategory.ResourceUnavailable,
+                    _typename);
+                ThrowTerminatingError(errorRecord);
+            }
+            else 
+            {
+                // write out all the available type definitions
```
iSazonov: Please remove obvious comment.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
```diff
@@ -169,27 +171,41 @@ protected override void ProcessRecord()
...
 
-            // write out all the available type definitions
-            foreach (var pair in typedefs)
+            // If no Type Definitions are found for _typename throw a terminating error
+            if (typedefs.Count == 0)
             {
-                var typeNames = pair.Key;
-
-                if (writeOldWay)
+                // All files must have the same extension otherwise throw.
```
iSazonov: Here the comment is unneeded. I believe we should remove it. (Or move to the method description).

src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
```diff
@@ -4,6 +4,8 @@
 using System;
 using System.Linq;
 using System.Management.Automation;
+using System.Management.Automation.Internal;
+
```
iSazonov: Please alphabetically resort all using-s.

iSazonov:
Seems we could [TypeLoadException](https://msdn.microsoft.com/en-us/library/system.typeloadexception(v=vs.110).aspx)

faraazahmad:
Those obvious comments were already there, so I didn't mess with them.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
faraazahmad: Isn't this already sorted?

iSazonov: We have other using-s below.

It is minor comment.

iSazonov (approved):
@faraazahmad Thanks for you contribution!

iSazonov:
@mklement0 Is the fix good for you?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
```diff
@@ -169,27 +170,38 @@ protected override void ProcessRecord()
...
-            // write out all the available type definitions
-            foreach (var pair in typedefs)
+            if (typedefs.Count == 0)
             {
-                var typeNames = pair.Key;
-
-                if (writeOldWay)
+                ErrorRecord errorRecord = new ErrorRecord(
+                    new TypeLoadException("No such type could be found"),
+                    "SPECIFIED_TYPE_NOT_FOUND",
```
SteveL-MSFT: For the FullQualifiedErrorId, the convention is to use Pascal casing: SpecifiedTypeNotFound.

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
mklement0: Good point, @SteveL-MSFT.

Additionally, it should be a _non-terminating_ error, given that the cmdlet accepts multiple inputs and potentially produces multiple outputs.

@faraazahmad: I know I suggested `ResourceUnavailable`  initially, but on further reflection it probably isn't the right category.

We can use `Remove-TypeData` as a model, which already reports an error (also non-terminating), and uses category `InvalidOperation`:

```powershell
PS> Remove-TypeData NoSuchType
Remove-TypeData : Error in TypeData "NoSuchType": The type "NoSuchType" was not found. The type name value must be the full name of the type. Verify the type name and run the command again.
At line:1 char:1
+ Remove-TypeData NoSuchType
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : InvalidOperation: (:) [Remove-TypeData], RuntimeException
+ FullyQualifiedErrorId : TypesDynamicRemoveException,Microsoft.PowerShell.Commands.RemoveTypeDataCommand
```

As an aside: I'm not sure `TypesDynamicRemoveException` was a great choice of error ID.

@SteveL-MSFT, @iSazonov, what do you think?

SteveL-MSFT: Yes, this should be a non-terminating error if multiple are accepted so no need to stop the entire operation.  Agree that `TypesDynamicRemoveException` is a bad error id, but it's a breaking change to change it now as the intent of the FQEI is to make it internet searchable.  For symmetry, it makes sense to use `InvalidOperation` and probably makes more sense for an end user whereas `TypeLoadException` is more dev-centric.

iSazonov:
@SteveL-MSFT I added new commit to address your comment, please update your review.

dantraMSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/FormatAndOutput/common/GetFormatDataCommand.cs
```diff
@@ -169,27 +170,38 @@ protected override void ProcessRecord()
...
 
-            // write out all the available type definitions
-            foreach (var pair in typedefs)
+            if (typedefs.Count == 0)
             {
-                var typeNames = pair.Key;
-
-                if (writeOldWay)
+                ErrorRecord errorRecord = new ErrorRecord(
+                    new TypeLoadException("No such type could be found"),
```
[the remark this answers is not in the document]
iSazonov: Fixed.

iSazonov:
@SteveL-MSFT @dantraMSFT Please update your review.

src/Microsoft.PowerShell.Commands.Utility/resources/GetFormatDataStrings.resx
```diff
@@ -0,0 +1,123 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="SpecifiedTypeNotFound" xml:space="preserve">
+    <value>Cannot find the type.</value>
```
SteveL-MSFT: Should probably be something like:

> The type name '{0}' could not be found.

SteveL-MSFT (changes requested):

dantraMSFT (approved):
Other than Steve's recommendation about including the type name in the error message string, LGTM

Jaykul:
HANG ON ... This is a **breaking** change to the function!!! 

Does `-ErrorAction SilentlyContinue` work, or are you now forcing everyone to wrap this in a try/catch block?

SteveL-MSFT:
@Jaykul this is a non-terminating error

SteveL-MSFT (approved):

Jaykul:
@SteveL-MSFT it still breaks code like this, right?
```powershell
if(!(Get-FormatData $Type)) {
   Update-FormatData -TypeName $Type ...
}
```

mklement0:
@Jaykul:

Yes, for someone who relied on the previous _de facto_ behavior, this would be a breaking change.

That in itself may be reason enough to not make the change - I have no sense of and have made no attempt to assess the real-world impact. What do you think the real-world impact is?

On the other end of the spectrum: _any_ bug fix is a breaking change: what used to break / work nonsensically before now no longer does.

But let us be clear re:

> _deliberately handled the situation_ **correctly** 

_Correctly_ in this situation means:

* inferred from the _de facto_ behavior

* with no supporting evidence from the documentation

* despite the behavior being inconsistent with that of other cmdlets.

And, indeed, the same applies to the current behavior of `Get-Module`: with a non-wildcard argument it should _not_ default to quietly outputting "nothing" (`[System.Management.Automation.Internal.AutomationNull]::Value`)

> or would you accept a PR for Get-Module too?

In that vein: either yes, or  document the surprising existing behavior - see #7498

As an aside: 

> may return multiple entries even without a wildcard.

Unless you're referring to `Get-Module`'s `-ListAvailable` switch - an explicit opt-in to _multiple_ matches - neither `Get-Module` nor `Get-FormatData`  return multiple matches with _non-wildcard_ arguments.

Jaykul:
It's ironic that I'm the one arguing this -- I argued rather passionately against the no-output, no-error design pattern years ago in a discussion on the PowerShell MVP mailing list (although at the time, my argument was _specifically_ against commands that behave differently when you have a simple name, versus a wildcard (thus allowing the possibility of multiple items).

Ultimately I was rather deluged with examples from the broader ecosystem showing examples of commands quietly returning nothing when they could have returned many things.  My point is that this is how PowerShell commands behave, in general:

_If the command **can** return MULTIPLE items_, then when there is nothing to return, they just return nothing. These aren't bugs. They are a deliberate design choice.

And yes, you're wrong @mklement0, when you say that those commands don't return multiple items with simple parameters. Here are some concrete examples.

```posh
# There are about *eight* default rules that affect ScriptInfo, for example:
Get-FormatData System.Management.Automation.ScriptInfo

# Users can have multiple versions of the same module imported because of dependencies.
# We can import multiple versions of Pester as an example...
# (most of us have 3.4.0 and a newer one, right?):
Get-Module Pester -ListAvailable | Import-Module
Get-Module Pester
```

But it's not just those, most of the default `Get` commands do this **if** they _could_ return multiple items:

### During the first command of your session, there's no history (or, do `Clear-History; Get-History`)
```posh
Get-History
```

#### When a folder is empty
```posh
mkdir emptyfolder
Get-ChildItem .\emptyfolder
```

#### When there's no input:
```posh
Get-Unique
# Or even ...
Get-ChildItem .\EmptyFolder\ | Get-Unique
# Or even ...
Get-ChildItem .\EmptyFolder\ | Get-Random
```

#### When there are no sessions
```posh
Get-PSSession
```

#### When there are no commands
```posh
Get-Command -Verb Build
```

#### When there's no TypeData:
```posh
Get-TypeData System.Boolean
```

#### When there's no Jobs:
```posh
Get-Job
```

#### When there's no scheduled jobs
```posh
Get-ScheduledJob
```

#### When the clipboard is empty, as after KeePass
```posh
Set-Clipboard $null 
Get-Clipboard
```

I trust I don't need to go through and test _every_ `Get` command to make the point?

mklement0:
@Jaykul:

> It's ironic that I'm the one arguing this 

I hear you, but let me ask you this before I address your examples:

* Despite your earlier misgivings, do all the behaviors you list now make sense to you?

* Do you believe that all of them apply directly to the case at hand?

Jaykul:
They make sense:

1. They are returning what's there.
2. They are consistent to the principle that if the command can return multiple items, in any particular case where there is nothing to return, they return that -- without error.

Obviously in a lot of these cases, it's very clear that there's not an error. That is (if you understood the situation), you didn't actually expect there to be any child items, history, jobs, content, sessions, etc...

I think all of them are examples of the rule, and serve to show that this cmdlet was correct without the PR.


----

For what it's worth:

**If** we were willing to break the backward compatibility of PowerShell, then I would be happy to argue again this design pattern and try to convince the world that **all** commands should have the same output behavior regardless of _how many_ items they might output:

1. Users (and developers) can't always tell _when_ this will be the case.
2. Although it's a "common" design pattern, it's clearly not well-known or documented.

But in reality, changing this behavior across all the built-in commands would break _too much_, and there's no way to actually "fix" it because too many commands are outside the control of this team.

faraazahmad:
I have to agree with @Jaykul , this makes sense. For example, if I `ls` an empty directory, the expected behaviour would be to not print anything, and not throw an error.

mklement0:
@SteveL-MSFT:

If this fix is ultimately kept, I suggest amending the error message in 2 ways:

* Make it clear that what couldn't be found wasn't a _type per se_, but that type's _formatting data_  _loaded into the current session_.

* Given the need to to always specify the _full_ type name (that is, you cannot omit the `System.` component and you cannot use type accelerators), the error message should provide a hint to that effect.

Therefore, instead of `The type name '{0}' could not be found.`, perhaps we could say:

```none
No formatting data for type name '{0}' has been loaded into the current session.
Note that type names must be specified exactly as defined and are typically full .NET type names.
```

SteveL-MSFT:
@PowerShell/powershell-committee discussed this and we also discussed it in this morning's Community Call.  We agreed that the intent is that if a literal is passed and is not found, it should return a non-terminating error.  Cmdlets that don't have this behavior have bugs.  In cases where it's explicit or implicit filtering, it should return nothing.  We also agreed that we should revert this change and revisit making cmdlets consistent in 6.2.  Finally, the real issue is that we are missing Test-* cmdlets so users had to rely on current behavior with Get-* cmdlets to see if something exists and we should add those cmdlets in the future.

Jaykul:
I think you're going to break so very many things -- that 6.2 needs to be 7.0 to be semantic

swngdnz:
Lord have mercy... once you do this, WPS and CPS are completely different languages that just happen to be related. Forget about moving scripts between the two.

joeyaiello:
I heard today in the @PowerShell/powershell-committee Community Call that we're going to revert this and work towards consistency and `Test-*` cmdlets across the board. 

@SteveL-MSFT if we agree that's the case, will the revert make RC? GA?

SteveL-MSFT:
The @PowerShell/powershell-maintainers will be reverting this commit, but that won't happen for RC where we are already in process of release but will make it for GA.  @iSazonov if you are available, can you revert this as the other maintainers are working on the release?

iSazonov:
@SteveL-MSFT Will do.

iSazonov:
@SteveL-MSFT The commit already reverted by @TravisEz13

SteveL-MSFT:
@iSazonov The discussion is continuing in https://github.com/PowerShell/PowerShell/issues/7498.  I've created https://github.com/PowerShell/PowerShell/issues/7562 for new Test- cmdlets

faraazahmad:
Didn't know my first contribution to Powershell could stir up so much controversy :sweat_smile:. Is it safe to delete the branch?

SteveL-MSFT:
@faraazahmad I think it's a healthy discussion.  Yes, you can delete the branch.

PR #7435, opened 2018-08-02
Remove extra allocations in PSMemberInfoInternalCollection<T>

iSazonov:
## PR Summary

Related #7112.
PSMemberInfoInternalCollection<T> uses internally OrderedDictionary class.
The OrderedDictionary class in .Net Core 2.1 allocates internally ArrayList  and Hashtable  after any first using any property. Even if we check that OrderedDictionary is empty (Count==0) or use indexer it causes allocations.

The commit postpones the internall allocations until this is really necessary.

Taking into account that PSMemberInfoInternalCollection<T> is used in each PSObject and is often empty, this change dramatically reduces memory allocations.

### PerfView results
1. Startup test
- before - GC Stats Total Allocs : 13.274 MB
- after  - GC Stats Total Allocs : 10.082 MB

2. Script test (Import-Csv)
Before fix:
- GC Stats Total Allocs : 4,073.747 MB
gcstatsbeforefix
gcheapallocbeforefix

After the fix:
- GC Stats Total Allocs : 1,186.017 MB
gcstatsafterfix
gcheapallocafterfix

Test script:
```powershell
cd c:\tmp\
function perf_test($source, $result) {
    Write-Host "Measuring Import-Csv performance over time..."

    "index,duration_ms,bytes_consumed" | Out-File $result
    for ($i=0; $i -le 400; $i++) {

            $millis = (Measure-Command { Import-Csv $source }).TotalMilliseconds
            # Uncomment this if you want analize results in Excel
            $memory = [System.GC]::GetTotalMemory($false)
	    $i.ToString() + "," + $millis.ToString() + "," + $memory | Out-File $result -Append
    }
    Write-Host "Done"
}

$fields = 0..19 | ForEach-Object { "random_numbers$_" }
($fields -join ",") | Out-File .\source2.csv
Get-Random -SetSeed 1 | Out-Null
for ($i=0; $i -le 500; $i++) {
    $values = 0..19 | ForEach-Object { (Get-Random).ToString() }
    ($values -join ",") | Out-File .\source2.csv -Append
}

perf_test .\source2.csv .\results.csv
```

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/MshMemberInfo.cs
+++ b/src/System.Management.Automation/engine/MshMemberInfo.cs
@@ -4251,7 +4251,7 @@ namespace System.Management.Automation
 
                 lock (_members)
                 {
-                    return Members.Count;
+                    return _members.Count;
                 }
             }
         }
@@ -4270,7 +4270,7 @@ namespace System.Management.Automation
 
                 lock (_members)
                 {
-                    return Members.Count - _countHidden;
+                    return _members.Count - _countHidden;
                 }
             }
         }
--- a/src/System.Management.Automation/engine/MshMemberInfo.cs
+++ b/src/System.Management.Automation/engine/MshMemberInfo.cs
@@ -4299,6 +4299,11 @@ namespace System.Management.Automation
         /// <returns>the enumerator for this collection</returns>
         public override IEnumerator<T> GetEnumerator()
         {
+            if (_members == null)
+            {
+                return Enumerable.Empty<T>().GetEnumerator();
+            }
+
             lock (_members)
             {
                 // Copy the members to a list so that iteration can be performed without holding a lock.
--- a/src/System.Management.Automation/engine/MshMemberInfo.cs
+++ b/src/System.Management.Automation/engine/MshMemberInfo.cs
@@ -4052,7 +4052,7 @@ namespace System.Management.Automation
         {
             Diagnostics.Assert(newMember != null, "called from internal code that checks for new member not null");
 
-            lock (_members)
+            lock (Members)
             {
                 var oldMember = Members[newMember.Name] as T;
                 Diagnostics.Assert(oldMember != null, "internal code checks member already exists");
@@ -4087,7 +4087,7 @@ namespace System.Management.Automation
                 throw PSTraceSource.NewArgumentNullException("member");
             }
 
-            lock (_members)
+            lock (Members)
             {
                 if (Members[member.Name] is T existingMember)
                 {
@@ -4130,7 +4130,7 @@ namespace System.Management.Automation
                 return;
             }
 
-            lock (_members)
+            lock (Members)
             {
                 if (Members[name] is PSMemberInfo member)
                 {
@@ -4163,7 +4163,7 @@ namespace System.Management.Automation
                     return null;
                 }
 
-                lock (_members)
+                lock (Members)
                 {
                     return Members[name] as T;
                 }
@@ -4223,7 +4223,7 @@ namespace System.Management.Automation
         private PSMemberInfoInternalCollection<T> GetInternalMembers(MshMemberMatchOptions matchOptions)
         {
             PSMemberInfoInternalCollection<T> returnValue = new PSMemberInfoInternalCollection<T>();
-            lock (_members)
+            lock (Members)
             {
                 foreach (T member in Members.Values.OfType<T>())
                 {
@@ -4249,7 +4249,7 @@ namespace System.Management.Automation
                     return 0;
                 }
 
-                lock (_members)
+                lock (Members)
                 {
                     return _members.Count;
                 }
@@ -4268,7 +4268,7 @@ namespace System.Management.Automation
                     return 0;
                 }
 
-                lock (_members)
+                lock (Members)
                 {
                     return _members.Count - _countHidden;
                 }
@@ -4289,7 +4289,7 @@ namespace System.Management.Automation
                     return null;
                 }
 
-                lock (_members)
+                lock (Members)
                 {
                     return _members[index] as T;
                 }
@@ -4309,7 +4309,7 @@ namespace System.Management.Automation
                 return Enumerable.Empty<T>().GetEnumerator();
             }
 
-            lock (_members)
+            lock (Members)
             {
                 // Copy the members to a list so that iteration can be performed without holding a lock.
                 return Members.Values.OfType<T>().ToList().GetEnumerator();
--- a/src/System.Management.Automation/engine/MshMemberInfo.cs
+++ b/src/System.Management.Automation/engine/MshMemberInfo.cs
@@ -4011,6 +4011,10 @@ namespace System.Management.Automation
         private OrderedDictionary _members;
         private int _countHidden;
 
+        /// <summary>
+        /// Get the OrderedDictionary for holding all members.
+        /// We use this property to delay initializing _members until we absolutely need to.
+        /// </summary>
         private OrderedDictionary Members
         {
             get
@@ -4130,15 +4134,15 @@ namespace System.Management.Automation
                 return;
             }
 
-            lock (Members)
+            lock (_members)
             {
-                if (Members[name] is PSMemberInfo member)
+                if (_members[name] is PSMemberInfo member)
                 {
                     if (member.IsHidden)
                     {
                         _countHidden--;
                     }
-                    Members.Remove(name);
+                    _members.Remove(name);
                 }
             }
         }
@@ -4163,9 +4167,9 @@ namespace System.Management.Automation
                     return null;
                 }
 
-                lock (Members)
+                lock (_members)
                 {
-                    return Members[name] as T;
+                    return _members[name] as T;
                 }
             }
         }
@@ -4223,9 +4227,15 @@ namespace System.Management.Automation
         private PSMemberInfoInternalCollection<T> GetInternalMembers(MshMemberMatchOptions matchOptions)
         {
             PSMemberInfoInternalCollection<T> returnValue = new PSMemberInfoInternalCollection<T>();
-            lock (Members)
+
+            if (_members == null)
             {
-                foreach (T member in Members.Values.OfType<T>())
+                return returnValue;
+            }
+
+            lock (_members)
+            {
+                foreach (T member in _members.Values.OfType<T>())
                 {
                     if (member.MatchesOptions(matchOptions))
                     {
@@ -4249,7 +4259,7 @@ namespace System.Management.Automation
                     return 0;
                 }
 
-                lock (Members)
+                lock (_members)
                 {
                     return _members.Count;
                 }
@@ -4268,7 +4278,7 @@ namespace System.Management.Automation
                     return 0;
                 }
 
-                lock (Members)
+                lock (_members)
                 {
                     return _members.Count - _countHidden;
                 }
@@ -4289,7 +4299,7 @@ namespace System.Management.Automation
                     return null;
                 }
 
-                lock (Members)
+                lock (_members)
                 {
                     return _members[index] as T;
                 }
@@ -4309,10 +4319,10 @@ namespace System.Management.Automation
                 return Enumerable.Empty<T>().GetEnumerator();
             }
 
-            lock (Members)
+            lock (_members)
             {
                 // Copy the members to a list so that iteration can be performed without holding a lock.
-                return Members.Values.OfType<T>().ToList().GetEnumerator();
+                return _members.Values.OfType<T>().ToList().GetEnumerator();
             }
         }
     }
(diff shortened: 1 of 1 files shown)


iSazonov:
@SteveL-MSFT If the PR is ok you could consider it for Windows PowerShell 5.1 too.

@daxian-dbw Sorry for ping - the PR can be important.

@markekraus Could you please check json cmdlets with the change if you have a time. I hope we get a win there too.

@vancem Thanks for your great analysis - it helped a lot!

SteveL-MSFT (approved):
LGTM.  At this time we're trying not to make changes to 5.1 as we're trying to make 6 a viable replacement for 5.1 and spending time on that instead.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -4008,20 +4008,32 @@ public virtual IEnumerator<T> GetEnumerator()
...
         private int _countHidden;
 
+        private OrderedDictionary Members
+        {
+            get
+            {
+                if (_members == null)
+                {
+                    _members = new OrderedDictionary(StringComparer.OrdinalIgnoreCase);
+                }
```
daxian-dbw: There could be a race condition here.
In an exaggerated way, thread A checked `_members == null` and entered the if block, then A got swapped out, and thread B came in and used `Members` (so `_member` is initialized). When A got swapped in again, it would re-create the `OrderedDictionary` object, and thus any members B added in will be gone.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -4114,16 +4125,15 @@ public override void Remove(string name)
                     name);
             }
 
-            lock (_members)
+            lock (Members)
```
daxian-dbw: Shall we also check for `_members == null` for the `Remove` action as a shortcut path?

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -4203,9 +4218,9 @@ internal override ReadOnlyPSMemberInfoCollection<T> Match(string name, PSMemberT
         private PSMemberInfoInternalCollection<T> GetInternalMembers(MshMemberMatchOptions matchOptions)
         {
             PSMemberInfoInternalCollection<T> returnValue = new PSMemberInfoInternalCollection<T>();
-            lock (_members)
+            lock (Members)
```
daxian-dbw: Same here, shall we also check for `_members == null` for a shortcut path?

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -4224,9 +4239,14 @@ internal int Count
...
-                lock (_members)
+                if (_members == null)
+                {
+                    return 0;
+                }
+
+                lock (Members)
                 {
-                    return _members.Count;
+                    return Members.Count;
```
daxian-dbw: Same nitpick here, how about just use `_members` for the `lock(...)` block?

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -4238,9 +4258,14 @@ internal int VisibleCount
...
-                lock (_members)
+                if (_members == null)
+                {
+                    return 0;
+                }
+
+                lock (Members)
                 {
-                    return _members.Count - _countHidden;
+                    return Members.Count - _countHidden;
```
daxian-dbw: Same nitpick here.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -4254,9 +4279,14 @@ internal int VisibleCount
...
-                lock (_members)
+                if (_members == null)
+                {
+                    return null;
+                }
+
+                lock (Members)
                 {
-                    return _members[index] as T;
+                    return Members[index] as T;
```
daxian-dbw: Same nitpick here.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -4269,10 +4299,10 @@ internal int VisibleCount
         /// <returns>the enumerator for this collection</returns>
         public override IEnumerator<T> GetEnumerator()
         {
-            lock (_members)
+            lock (Members)
```
daxian-dbw: if we want to go bit far, maybe also check `_members == null` here and if so return `Array.Empty<T>.GetEnumerator()`?
Just an idea.

src/System.Management.Automation/engine/MshMemberInfo.cs
daxian-dbw: I suggest the following code:
```
if (_members == null)
{
    Interlocked.CompareExchange(ref _members, new OrderedDictionary(StringComparer.OrdinalIgnoreCase), null);
}

return _members;
```

daxian-dbw (changes requested):

src/System.Management.Automation/engine/MshMemberInfo.cs
iSazonov: Good idea for the _public_ method.

Fixed as `return Enumerable.Empty<T>().GetEnumerator();`

src/System.Management.Automation/engine/MshMemberInfo.cs
iSazonov: Great catch!

Fixed.

src/System.Management.Automation/engine/MshMemberInfo.cs
iSazonov: I think we can add the check in public methods.

Fixed.

src/System.Management.Automation/engine/MshMemberInfo.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/MshMemberInfo.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/MshMemberInfo.cs
iSazonov: Fixed.

src/System.Management.Automation/engine/MshMemberInfo.cs
iSazonov: Fixed.

iSazonov:
@daxian-dbw I reverted `lock (_members)` to `lock (Members)` because the lock failed on null in Add() and Remove(). I believe that it makes no sense to add the null check in the methods so using `lock (Members)` everywhere looks good..

iSazonov:
I measure pwsh.exe startup and update PR description:
> Startup test
- before - GC Stats Total Allocs : 13.274 MB
- after  - GC Stats Total Allocs : 10.082 MB

markekraus:
@iSazonov The issue with JSON is one of speed more than memory allocation. I don't have he tooling to test the memory performance. but the code I use to test the JSON performance is this:

```powershell
$json = @'
{
    "id": "string",
    "data": [
        {                       
        "accountId": 0,
        "productId": 0,
        "resourceLocationId": 0,
        "consumedServiceId": 0,
        "departmentId": 0,
        "accountOwnerEmail": "string",
        "accountName": "string",
        "serviceAdministratorId": "string",
        "subscriptionId": 0,
        "subscriptionGuid": "string",
        "subscriptionName": "string",
        "date": "2017-04-27T23:01:43.799Z",
        "product": "string",
        "meterId": "string",
        "meterCategory": "string",
        "meterSubCategory": "string",
        "meterRegion": "string",
        "meterName": "string",
        "consumedQuantity": 0,
        "resourceRate": 0,
        "Cost": 0,
        "resourceLocation": "string",
        "consumedService": "string",
        "instanceId": "string",
        "serviceInfo1": "string",
        "serviceInfo2": "string",
        "additionalInfo": "string",
        "tags": "string",
        "storeServiceIdentifier": "string",
        "departmentName": "string",
        "costCenter": "string",
        "unitOfMeasure": "string",
        "resourceGroup": "string"
        }
    ],
    "nextLink": "string"
}
'@
$json = 1..1000 | % { $json }
$json = '[' + ($json -join ',') + ']'
Measure-Command {$Json | ConvertFrom-Json } | % TotalMilliseconds
```

5.1: 375.3416ms
6.0.2: 2055.1092ms
6.1.0-preview.4: 1252.1725ms
6.1.0-preview.10466: 1130.4745ms (your most recent passing build for this PR)

so it is marginally better than 6.1.0-preview.4, better than 6.0.2, but still nowhere near the performance of 5.1. ☹️

iSazonov:
@markekraus Many thanks for your repo script! I'll try to look it deeper next week.

daxian-dbw:
@iSazonov You cannot change all `lock(Members)` to `lock(_members)`. You can change those that `_members == null` has already been checked. For `Add`, there is no check for `_members != null`, so changing to `lock (_members)` would fail for sure.

powercode:
There is a *very* expensive check for duplicate members.

By replacing that check, and by keeping a hashset of preValidated member, I can speed it up by a factor of 3 or so.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -4309,10 +4319,10 @@ public override IEnumerator<T> GetEnumerator()
                 return Enumerable.Empty<T>().GetEnumerator();
             }
 
-            lock (Members)
+            lock (_members)
```
iSazonov: @daxian-dbw Please clarify - is it ok to lock on different objects in different methods? What if different threads call Add() and Remove() in the same time?

iSazonov:
@powercode Using preValidated looks as internal only hack. It will be nice to find more generic solution that we can expose to public.

powercode:
@iSazonov I agree about `preValidated`.

src/System.Management.Automation/engine/MshMemberInfo.cs
daxian-dbw: All locks are essentially on the same object. When `Add` and `Remove` are called at the same time, there are two possibilities:
1. The check of `_members == null` in `Remove` happens **before** the initialization of `_member` triggered by `Add`;
1. The check of `_members == null` in `Remove` happens **after** the initialization of `_member` triggered by `Add`.

In the first case, `Remove` will just return, and `Add` will lock on the newly initialized object.
In the second case, `Remove` will try to lock on `_members` (once it's not null, it will never be set back to null), which points the same object as `Member`, so `Remove` and `Add` will compete locking on the same object.

src/System.Management.Automation/engine/MshMemberInfo.cs
```diff
@@ -4075,16 +4087,15 @@ public override void Add(T member, bool preValidated)
                 throw PSTraceSource.NewArgumentNullException("member");
             }
 
-            lock (_members)
+            lock (Members)
```
powercode: would it be better to save a local reference to Members, instead of calling the Property many times in the same would it be better to save a local reference to the members instead of repeatedly call the `Members` property ?
var member = Members;
lock(members)
{
   ...

powercode (commented):
Would it be better to get a local reference to the Members instead of calling the property repeatedly?
```
            var members = Members;
            lock (members)
            {
                if (members[member.Name] is T existingMember)
                {
                    Replace(existingMember, member);
                }
                else
                {
                    members[member.Name] = member;
                    if (member.IsHidden)
                    {
                        _countHidden++;
                    }
                }
            }
```

daxian-dbw:
One test failed in AppVeyor, but works fine on my dev box. Will address @powercode's comment and trigger the CIs again.

```
[00:15:18]     [-] -WindowStyle Maximized should work on Windows 
[00:15:18]       Expected exactly {Maximized}, but got {Hidden}.
[00:15:18]       636:         $showCmd | Should -BeExactly $WindowStyle
[00:15:18]       at Invoke-Assertion, C:\projects\powershell\src\powershell-win-core\bin\Release\netcoreapp2.1\win7-x64\publish\Modules\Pester\4.2.0\Functions\Assertions\Should.ps1: line 206
[00:15:18]       at <ScriptBlock>, C:\projects\powershell\test\powershell\Host\ConsoleHost.Tests.ps1: line 636
[00:15:18] Processing -WindowStyle 'invalid' failed: Cannot convert value "invalid" to type "System.Diagnostics.ProcessWindowStyle". Error: "Unable to match the identifier name invalid to a valid enumerator name. Specify one of the following enumerator names and try again:
[00:15:18] Normal, Hidden, Minimized, Maximized".
```

daxian-dbw (approved):

iSazonov:
Thanks all for help and useful comments!

iSazonov:
OrderedDictionary was fixed in https://github.com/dotnet/corefx/pull/32001 - some properies now don't allocate if no need.

PR #7438, opened 2018-08-03
Adding LanguagePrimitives.TryCompare to provide faster comparisons

powercode:
## PR Summary

This PR adds the  `TryCompare` equivalent of `Compare`, just as we have a `TryConvert` alternative to `Convert`.
This enables faster comparison in the cases where the convertions throws exception.

As an example:

```powershell
 1..10000 | foreach{$_; [timespan]::FromSeconds($_)}  | sort
```
takes `3 800 ms` before and `280 ms` after.

```powershell
$b = 1..1000000 | %{100; [timespan]::FromSeconds(100)}  | group
```
takes `20 s` before and `15 s` after.

It is implemented equivalently to `LanguagePrimitives.Compare`, but returning false where `Compare` would have thrown an exception.

This is adding a new public API to `LanguagePrimitives`, so it will probably need a committee approval.

The new APIs are:

```CSharp
class LanguagePrimitives {
  public static bool TryCompare(object first, object second, out int result);
  public static bool TryCompare(object first, object second, bool ignoreCase, out int result);
  /// <summary>
  /// Tries to compare first and second, converting second to the type of the first, if necessary.
  /// If a conversion is needed but fails, false is return.
  /// </summary>
  /// <param name="first">First comparison value.</param>
  /// <param name="second">Second comparison value.</param>
  /// <param name="ignoreCase">Used if both values are strings.</param>
  /// <param name="formatProvider">Used in type conversions and if both values are strings.</param>
  /// <param name="result">Less than zero if first is smaller than second, more than  zero if it is greater or zero if they are the same.</param>
  /// <returns>True if the comparison was successful, false otherwise.</returns>
  /// <exception cref="ArgumentException">The parameter <paramref name="formatProvider"/> is not a <see cref="CultureInfo"/>.</exception>
  public static bool TryCompare(
      object first, 
      object second, 
      bool ignoreCase, 
      IFormatProvider formatProvider,
      out int result);
}
```

The new APIs are used by `ObjectCommandPropertyValue` (used by sort-object and group-object) and `FormatGroupManager`.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/ObjectCommandComparer.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/ObjectCommandComparer.cs
@@ -79,31 +79,30 @@ namespace Microsoft.PowerShell.Commands
         {
             ObjectCommandPropertyValue objectCommandPropertyValueObject = inputObject as ObjectCommandPropertyValue;
             if (objectCommandPropertyValueObject == null)
+            {
                 return false;
+            }
 
             object baseObject = PSObject.Base(PropertyValue);
             object inComingbaseObjectPropertyValue = PSObject.Base(objectCommandPropertyValueObject.PropertyValue);
 
-            IComparable baseObjectComparable = baseObject as IComparable;
-
-            if (baseObjectComparable != null)
+            if (baseObject is IComparable)
             {
-                return (LanguagePrimitives.Compare(baseObject, inComingbaseObjectPropertyValue, CaseSensitive, Culture) == 0);
+                var success = LanguagePrimitives.TryCompare(baseObject, inComingbaseObjectPropertyValue, CaseSensitive, Culture, out int result);
+                return success && result == 0;
             }
-            else
-            {
-                if (baseObject == null && inComingbaseObjectPropertyValue == null)
-                {
-                    return true;
-                }
-                if (baseObject != null && inComingbaseObjectPropertyValue != null)
-                {
-                    return baseObject.ToString().Equals(inComingbaseObjectPropertyValue.ToString(), StringComparison.OrdinalIgnoreCase);
-                }
 
-                // One of the property values being compared is null.
-                return false;
+            if (baseObject == null && inComingbaseObjectPropertyValue == null)
+            {
+                return true;
             }
+            if (baseObject != null && inComingbaseObjectPropertyValue != null)
+            {
+                return baseObject.ToString().Equals(inComingbaseObjectPropertyValue.ToString(), StringComparison.OrdinalIgnoreCase);
+            }
+
+            // One of the property values being compared is null.
+            return false;
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


powercode:
The remaining CodeFactor issues are intentionally left that way to match the style of similar code in the same file.

iSazonov:
> This PR adds the Compare equivalent of TryConvert.

Seem typo in the PR description.

test/powershell/engine/Api/LanguagePrimitive.Tests.ps1
```diff
@@ -46,4 +46,60 @@ Describe "Language Primitive Tests" -Tags "CI" {
         [System.Management.Automation.LanguagePrimitives]::GetEnumerable($dt) | ForEach-Object { $count++ }
         $count | Should -Be 2
     }
+
+    It "TryCompare should succeed on int and string" {
+        $result = $null
+        [System.Management.Automation.LanguagePrimitives]::TryCompare(1, "1", [ref] $result) | Should -BeTrue
+        $result | Should -BeExactly 0
```
iSazonov: We use `-BeExactly` for string only. Here we expect an integer and should use `-Be`.
Please review using `-BeExactly` below too.

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -854,6 +854,146 @@ public static int Compare(object first, object second, bool ignoreCase, IFormatP
...
+        /// <exception cref="ArgumentException">The parameter <paramref name="formatProvider"/> is not a <see cref="CultureInfo"/>.</exception>
+        public static bool TryCompare(object first, object second, bool ignoreCase, IFormatProvider formatProvider, out int result)
+        {
+            result = 0;
+            if (formatProvider == null)
+            {
+                formatProvider = CultureInfo.InvariantCulture;
+            }
+
+            if (!(formatProvider is CultureInfo culture))
```
iSazonov: Why we use `IFormatProvider` parameter type if we want get CultureInfo?

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -854,6 +854,146 @@ public static int Compare(object first, object second, bool ignoreCase, IFormatP
...
+            {
+                throw PSTraceSource.NewArgumentException("formatProvider");
+            }
+
+            first = PSObject.Base(first);
+            second = PSObject.Base(second);
+
+            int CompareNullToNumeric(object value, bool numberIsRightHandSide)
```
iSazonov: I think we should use "dry" principle and use the function everywhere in the file.

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -854,6 +854,146 @@ public static int Compare(object first, object second, bool ignoreCase, IFormatP
...
+                // for everything else it's less than zero...
+                result = CompareNullToNumeric(first, false);
+                return true;
+            }
+
+            if (first is string firstString)
+            {
+                if (!(second is string secondString))
+                {
+                    if (!TryConvertTo(second, culture, out secondString))
```
iSazonov: We could replace the two if-s with one.

iSazonov (commented):
Please add some samples that we really get a performance win.

New public API must be approved by @powershell-committee.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and is fine with the api additions.  Thanks.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ObjectCommandComparer.cs
```diff
@@ -190,35 +189,30 @@ public int Compare(object first, object second)
...
-            }
-            catch (InvalidCastException)
-            {
-            }
-            catch (ArgumentException)
+            if (LanguagePrimitives.TryCompare(first, second, !_caseSensitive, _cultureInfo, out int result))
             {
-                // Note that this will occur if the objects do not support
-                // IComparable.  We fall back to comparing as strings.
+                return result * (_ascendingOrder ? 1 : -1);
```
iSazonov: Do we get some win if we exclude multiplication?

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -854,6 +847,129 @@ public static int Compare(object first, object second, bool ignoreCase, IFormatP
...
+            {
+                throw PSTraceSource.NewArgumentException("formatProvider");
+            }
+
+            first = PSObject.Base(first);
+            second = PSObject.Base(second);
+
+
+
+            if (first == null && second == null)
```
iSazonov: Please remove extra empty lines above.

iSazonov (approved):
LGTM with 3 minor comments.

Thanks for the great performance improvements.

daxian-dbw (approved):
LGTM

src/Microsoft.PowerShell.Commands.Utility/commands/utility/ObjectCommandComparer.cs
powercode: I'd leave that to the compiler and the JIT. They usually do a good job on things like this.

src/System.Management.Automation/engine/LanguagePrimitives.cs
```diff
@@ -854,6 +847,127 @@ public static int Compare(object first, object second, bool ignoreCase, IFormatP
...
+                // for everything else it's less than zero...
+                result = CompareObjectToNull(first, false);
+                return true;
+            }
+
+            if (first is string firstString)
+            {
+                if (!(second is string secondString))
+                {
+                    if (!TryConvertTo(second, culture, out secondString))
```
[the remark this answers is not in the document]
powercode: @PaulHigin I recently changed TryConvertTo to check the result of FigureConvertion and exit early, without invoking ConversionData.Invoke() if the ConvertionRank is None.

So, yes, I have avoided the exception in many but not all cases. Interesting take to put it on ConversionData. That is arguably where it belongs.

Issue #7481, opened 2018-08-08
Perf: Add PSObjectBuilder


powercode:
As a part of some perf work I've been doing lately, I looked at the creations of PSObjects.

The creation of them could be significantly faster if we can provide a context for the creation. The code as it exists is very defensive, doing work multiple times that in many cases is only needed once etc.

I propose we add two new types, `ref struct PSObjectBuilder`, and `class PSObjectBuilderUnsafe`.

`PSObjectBuilder` can be more aggressive since we can guarantee that the object it is building is only ever accessible from the same call stack. 
`PSObjectBuilderUnsafe` has to be a bit more defensive since it could potentially be leaked to another runspace/thread.

The PSObjectBuilder guarantees that the object under construction is never exposed to the outside world until it is fully constructed.
It also keep track of what validations has been done, and can use this to skip expensive checks that otherwise has to be performed.

I have prototyped this at https://github.com/powercode/powershell/tree/PSObjectBuilder where I also used it in the JSON cmdlets and the providers.

The results are quite promising.
`ConvertFrom-Json` is 7x faster.
`gci -re -fo c:\windows -ea:0` is 2-3x faster.

But I'm not quite sure about how to move forward here. 

The branch I have contains changes that should go into separate PRs and there are things I would like to get design feedback on.
Should I create a PR anyway or what's the best way to go about this?

iSazonov:
@powercode Feel free to push the PR - it is easiest way to discuss new ideas and code. Many thanks for your investigations!

Currently I don't understand where we could use `PSObjectBuilder` - do we currently create PSObject objects for internal use only in the engine?
Also why do we need new struct and class - can we implement new constructors for PSObject ? or you say about "PSObject fabric"?

dragonwolf83:
Add-Member performance would be good to test as well.

powercode:
There is one use case where it might make sense, and that is the `-NotePropertyMembers @{...}`.

That could potentially be made faster.

iSazonov:
[...]

PR #7493, opened 2018-08-10
WIP: PSObjectBuilder design feedback

powercode:
## PR Summary

Related #7481 and #7112.

I'd like feedback on this idea, it's implementation and maybe see if there are other solutions to the problem that we should explore before going to far in this directions.

The main idea is to capture context of object creation and avoid expensive one-time operations with the PSObjectBuilder struct, or the PSObjectBuilderUnsafe class (can be leaked).

Ignore formatting  and only consider concepts. If this is the way to move forward, I'll provide code of higher conformance quality.

@daxian-dbw @BrucePay I'd love if you have time to take a look.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/MshObject.cs
+++ b/src/System.Management.Automation/engine/MshObject.cs
@@ -2386,7 +2386,7 @@ namespace System.Management.Automation
     /// object being created, and the members that are being added are not
     /// visible from any other location.
     /// This way, defensive copies and otherwise expensive operations can be avoided.
-    public ref struct PSObjectBuilder
+    public ref struct ValuePSObjectBuilder
     {
         const int DefaultHirearchyDepth = 4;
         private readonly HashSet<string> _validatedNames;
@@ -2397,10 +2397,10 @@ namespace System.Management.Automation
         private Stack<PSObject> CurrentObjectStack => _currentObjectStack ?? (_currentObjectStack = new Stack<PSObject>(DefaultHirearchyDepth));
 
         /// <summary>
-        /// Initializes a new instance of PSObjectBuilder.
+        /// Initializes a new instance of ValuePSObjectBuilder.
         /// </summary>
         /// <param name="validatedNameCount">The initial size of the set that tracks validated parameter names.</param>
-        public PSObjectBuilder(int validatedNameCount)
+        public ValuePSObjectBuilder(int validatedNameCount)
         {
             if (validatedNameCount < 1)
             {
@@ -2413,11 +2413,11 @@ namespace System.Management.Automation
         }
 
         /// <summary>
-        /// Initializes a new instance of PSObjectBuilder.
+        /// Initializes a new instance of ValuePSObjectBuilder.
         /// </summary>
         /// <param name="validatedNameCount">The initial size of the set that tracks validated parameter names.</param>
         /// <param name="trustedNames">True if all the names are preValidated.</param>
-        internal PSObjectBuilder(int validatedNameCount, bool trustedNames)
+        internal ValuePSObjectBuilder(int validatedNameCount, bool trustedNames)
         {
             if (validatedNameCount < 1)
             {
@@ -2488,10 +2488,10 @@ namespace System.Management.Automation
     }
 
     /// <summary>
-    /// Unsafe version of PSObjectBuilder that is not guaranteed to be on the same thread.
+    /// Unsafe version of ValuePSObjectBuilder that is not guaranteed to be on the same thread.
     /// Usage must be reviewed carefully.
     /// </summary>
-    internal class PSObjectBuilderLessOptimized
+    internal class PSObjectBuilder
     {
         const int DefaultHirearchyDepth = 4;
         private readonly HashSet<string> _validatedNames;
@@ -2502,10 +2502,10 @@ namespace System.Management.Automation
         private Stack<PSObject> CurrentObjectStack => _currentObjectStack ?? (_currentObjectStack = new Stack<PSObject>(DefaultHirearchyDepth));
 
         /// <summary>
-        /// Initializes a new instance of PSObjectBuilder.
+        /// Initializes a new instance of ValuePSObjectBuilder.
         /// </summary>
         /// <param name="validatedNameCount">The initial size of the set that tracks validated parameter names.</param>
-        public PSObjectBuilderLessOptimized(int validatedNameCount)
+        public PSObjectBuilder(int validatedNameCount)
         {
             if (validatedNameCount < 1)
             {
--- a/src/System.Management.Automation/engine/MshObject.cs
+++ b/src/System.Management.Automation/engine/MshObject.cs
@@ -2491,7 +2491,7 @@ namespace System.Management.Automation
     /// Unsafe version of PSObjectBuilder that is not guaranteed to be on the same thread.
     /// Usage must be reviewed carefully.
     /// </summary>
-    internal class PSObjectBuilderUnsafe
+    internal class PSObjectBuilderLessOptimized
     {
         const int DefaultHirearchyDepth = 4;
         private readonly HashSet<string> _validatedNames;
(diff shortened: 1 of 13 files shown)


src/System.Management.Automation/engine/MshObject.cs
```diff
@@ -2347,6 +2378,204 @@ internal bool IsHelpObject
...
+    /// This way, defensive copies and otherwise expensive operations can be avoided.
+    public ref struct PSObjectBuilder
+    {
+        const int DefaultHirearchyDepth = 4;
+        private readonly HashSet<string> _validatedNames;
+        private Stack<PSObject> _currentObjectStack;
+
+        private PSObject _currentObject;
+
+        private Stack<PSObject> CurrentObjectStack => _currentObjectStack ?? (_currentObjectStack = new Stack<PSObject>(DefaultHirearchyDepth));
```
iSazonov: Ah, we could implement ValueStack struct to allocate init segment in stack. (I see ValueStringBuilder is used internally in CoreFX.) Seems we could exclude tons short-time allocations. The same is true for HashSet ( _validatedNames).

src/System.Management.Automation/engine/MshObject.cs
```diff
@@ -2347,6 +2378,204 @@ internal bool IsHelpObject
...
+        /// Begins the creation of a new PSObject
+        /// </summary>
+        /// <param name="capacity">The initial capacity for the members collection.</param>
+        public void BeginCreateObject(int capacity)
+        {
+            if (_currentObject != null)
+            {
+                CurrentObjectStack.Push(_currentObject);
+            }
+            _currentObject = new PSObject(capacity);
```
iSazonov: Can we by default push the current object on stack? Seem in the case we could initilize  _currentObject in constructor and exclude many  `if (_currentObject != null)`.

src/System.Management.Automation/engine/MshObject.cs
```diff
@@ -2347,6 +2378,204 @@ internal bool IsHelpObject
...
+        /// <param name="name">The name of the property.</param>
+        /// <param name="value">The value of the property.</param>
+        public void AddNoteProperty(string name, object value)
+        {
+            var validated = _validatedNames?.Contains(name) ?? true;
+            if (!validated)
+            {
+                _validatedNames.Add(name);
+            }
+            _currentObject.Properties.Add(new PSNoteProperty(name, value), validated, inObjectBuilderScope:true);
```
iSazonov: Seem we should check  `_currentObject != null`.

For discussion I should mention that we could use `UnsafeAdd()` where internally inObjectBuilderScope:true.

src/System.Management.Automation/engine/MshObject.cs
```diff
@@ -2347,6 +2378,204 @@ internal bool IsHelpObject
...
+        }
+
+        /// <summary>
+        /// Completes the addition of the object.
+        /// </summary>
+        /// <returns></returns>
+        public PSObject EndCreateObject()
```
iSazonov: If we want to follow StringBuilder pattern the method should be `ToPSObject()`.

src/System.Management.Automation/engine/MshObject.cs
```diff
@@ -2347,6 +2378,204 @@ internal bool IsHelpObject
...
+        {
+            var current = _currentObject;
+            PSObject next = null;
+            CurrentObjectStack?.TryPop(out next);
```
iSazonov: No need check CurrentObjectStack != null.

src/System.Management.Automation/engine/MshObject.cs
```diff
@@ -2347,6 +2378,204 @@ internal bool IsHelpObject
...
+
+        private PSObject _currentObject;
+
+        private Stack<PSObject> CurrentObjectStack => _currentObjectStack ?? (_currentObjectStack = new Stack<PSObject>(DefaultHirearchyDepth));
+
+        /// <summary>
+        /// Initializes a new instance of PSObjectBuilder.
+        /// </summary>
+        /// <param name="validatedNameCount">The initial size of the set that tracks validated parameter names.</param>
+        public PSObjectBuilderUnsafe(int validatedNameCount)
```
iSazonov: It should be internal only if it is unsafe. On the other hand, it would be nice to have an api that would work in scripts not only in binaries.

src/System.Management.Automation/engine/TypeTable.cs
```diff
@@ -3549,20 +3549,22 @@ internal Collection<String> GetSpecificProperties(ConsolidatedString types)
             return PSObject.TransformMemberInfoCollection<PSMemberInfo, T>(GetMembers(types));
         }
 
-        internal PSMemberInfoInternalCollection<PSMemberInfo> GetMembers(ConsolidatedString types)
+        internal PSMemberInfoInternalCollection<PSMemberInfo> GetMembers(ConsolidatedString types, bool inObjectBuilderScope = false)
```
iSazonov: The name is very confusing me - in the method we use GetOrAdd() - it is _modify_ operation in _Get_ method! I'd rename the method to GetOrAddMembers().

src/System.Management.Automation/namespaces/ProviderBase.cs
```diff
@@ -64,6 +64,8 @@ public abstract partial class CmdletProvider : IResourceSupplier
         /// </summary>
         private ProviderInfo _providerInformation = null;
 
+        private readonly PSObjectBuilderUnsafe _psObjectBuilderUnsafe = new PSObjectBuilderUnsafe(10);
```
iSazonov: Why should we use unsafe version here?

iSazonov (commented):
Leave a comment

src/System.Management.Automation/engine/MshObject.cs
powercode: I do. That's what ?. does.

src/System.Management.Automation/engine/MshObject.cs
powercode: I think we should have another name. I actually don't mean unsafe, I mean slower since we cannot make as many guarantees as the ref struct.

powercode: So it's a really bad name.

src/System.Management.Automation/namespaces/ProviderBase.cs
powercode: The PSObjectBuilder cannot leave a stack frame. It cannot be a member of a type or allocated on the heap.

So the less optimized version has to be used if it is to survive between calls.

powercode: And again - Think of PSObjectBuilderUnsafe as PSObjectBuilderLessOptimized.

powercode: I will rename it. I see that it causes confusion from the start.

powercode: I like that suggestion! Changing!

src/System.Management.Automation/engine/MshObject.cs
```diff
@@ -2491,7 +2491,7 @@ public PSObject EndCreateObject()
     /// Unsafe version of PSObjectBuilder that is not guaranteed to be on the same thread.
     /// Usage must be reviewed carefully.
     /// </summary>
-    internal class PSObjectBuilderUnsafe
+    internal class PSObjectBuilderLessOptimized
```
iSazonov: It is not clear - less optimized then what? Should we use full name `PSObjectBuilderLessOptimizedThenVakuePSObjectBuilder` ?
I think we can move `LessOptimized` in comments and use simple name `PSObjectBuilder`.

src/System.Management.Automation/engine/MshObject.cs
iSazonov: I meant that CurrentObjectStack do already this internally.

src/System.Management.Automation/engine/MshObject.cs
```diff
@@ -2347,6 +2378,204 @@ internal bool IsHelpObject
...
+            }
+            _currentObject = new PSObject(capacity);
+        }
+
+        /// <summary>
+        /// Begins the creation of a new PSObject
+        /// </summary>
+        /// <param name="instance">The object to wrap in a PSObject.</param>
+        /// <param name="capacity">The initial capacity for the members collection.</param>
+        public void BeginCreateObject(object instance, int capacity)
```
iSazonov: `validatedNameCount` and `capacity` could be one value.

iSazonov:
@powercode Please resolve merge conflicts.

PR #7505, opened 2018-08-12
build: minor syntax fix

kwkam:
## PR Summary

Make Stop-DevPowerShell only call Resolve-Path once
Move Push-Location inside try block

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/build.psm1
+++ b/build.psm1
@@ -447,11 +447,15 @@ function Start-PSBuild {
     }
 
     function Stop-DevPowerShell {
+        if (-not (Test-Path $script:Options.Output)) {
+            return
+        }
+        $pwshDev = Resolve-Path $script:Options.Output
         Get-Process pwsh* |
             Where-Object {
                 $_.Modules |
                 Where-Object {
-                    $_.FileName -eq (Resolve-Path $script:Options.Output).Path
+                    $_.FileName -eq $pwshDev.Path
                 }
             } |
         Stop-Process -Verbose
--- a/build.psm1
+++ b/build.psm1
@@ -463,8 +463,8 @@ function Start-PSBuild {
 
     if ($Clean) {
         Write-Log "Cleaning your working directory. You can also do it with 'git clean -fdx --exclude .vs/PowerShell/v15/Server/sqlite3'"
-        Push-Location $PSScriptRoot
         try {
+            Push-Location $PSScriptRoot
             # Excluded sqlite3 folder is due to this Roslyn issue: https://github.com/dotnet/roslyn/issues/23060
             git clean -fdx --exclude .vs/PowerShell/v15/Server/sqlite3
         } finally {
@@ -963,8 +963,8 @@ function Publish-PSTestTools {
     # Publish tools so it can be run by tests
     foreach ($tool in $tools)
     {
-        Push-Location $tool.Path
         try {
+            Push-Location $tool.Path
             dotnet publish --output bin --configuration $Options.Configuration --framework $Options.Framework --runtime $Options.Runtime
             $toolPath = Join-Path -Path $tool.Path -ChildPath "bin"
 
@@ -1606,9 +1606,9 @@ function Start-PSBootstrap {
 
     Write-Log "Installing PowerShell build dependencies"
 
-    Push-Location $PSScriptRoot/tools
-
     try {
+        Push-Location $PSScriptRoot/tools
+
         if ($Environment.IsLinux -or $Environment.IsMacOS) {
             # This allows sudo install to be optional; needed when running in containers / as root
             # Note that when it is null, Invoke-Expression (but not &) must be used to interpolate properly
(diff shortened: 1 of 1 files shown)


build.psm1
```diff
@@ -447,20 +447,24 @@ function Start-PSBuild {
     }
 
     function Stop-DevPowerShell {
+        if (-not (Test-Path $script:Options.Output)) {
+            return
+        }
+        $pwshDev = Resolve-Path $script:Options.Output
```
kvprasoon: Please change ```$pwshDev``` to Pascal case ```$PwshDev``` .

build.psm1
```diff
@@ -447,20 +447,24 @@ function Start-PSBuild {
...
         Get-Process pwsh* |
             Where-Object {
                 $_.Modules |
                 Where-Object {
-                    $_.FileName -eq (Resolve-Path $script:Options.Output).Path
+                    $_.FileName -eq $pwshDev.Path
```
kvprasoon: Pascal case here too.

kvprasoon (changes requested):

build.psm1
```diff
@@ -447,20 +447,24 @@ function Start-PSBuild {
     }
 
     function Stop-DevPowerShell {
+        if (-not (Test-Path $script:Options.Output)) {
+            return
```
anmenaga: This return path needs to write warning/error.

anmenaga (approved):
LGTM with 1 recommendation.

anmenaga:
Restarted travis-ci for following error:
```
WARNING: The following packages cannot be authenticated!
  powershell
E: There were unauthenticated packages and -y was used without --allow-unauthenticated
/home/travis/build/PowerShell/PowerShell/tools/installpsh-debian.sh: line 177: pwsh: command not found
```

dantraMSFT (changes requested):
Moving the Push-Location changes inside the try block doesn't make sense to me since Pop-Location should only be called if Push succeeded.  Moving it into the try block means Pop-Location is called even when Push-Location fails.

build.psm1
```diff
@@ -447,20 +447,24 @@ function Start-PSBuild {
...
                 }
             } |
         Stop-Process -Verbose
     }
 
     if ($Clean) {
         Write-Log "Cleaning your working directory. You can also do it with 'git clean -fdx --exclude .vs/PowerShell/v15/Server/sqlite3'"
-        Push-Location $PSScriptRoot
         try {
+            Push-Location $PSScriptRoot
```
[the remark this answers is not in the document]
kvprasoon: Since `try` block starts with `Push-Location`, IMO its fine to have it inside.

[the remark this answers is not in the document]
dantraMSFT: @kvprasoon I think if you try this script, you'll find the results are not as you expect.

```powershell
Write-Output (Get-Location)
Push-Location "~\Documents"
Write-Output (Get-Location)
try
{
    Push-Location 'foo'
}
finally
{
    Write-Output (Get-Location)
    Pop-Location
    Write-Output (Get-Location)
}
```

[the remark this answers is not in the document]
kvprasoon: @dantraMSFT 👍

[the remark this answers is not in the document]
dantraMSFT: @kvprasoon - I realized that Push-Location 'foo' does not throw a terminating exception.  So, if you want this to work reliably, you would need -ErrorAction Stop on the Push-Location call.  Regardless, even if you use -Stop in the try block and it fails, Pop-Location is still attempting to pop a location that never got pushed.

At the end of the day, there's no benefit for the Push-Location changes.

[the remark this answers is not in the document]
kwkam: @dantraMSFT 
1. I feel that it's more readable having `Push-Location` inside try block, to make it looks like a section of code wrapped by `Push` and `Pop`. You may disagree.

2. The purpose of moving `Push-Location` is to standardise the logic in `build.psm1`, since both style exists in the current script. I can also move `Push-Location` outside try block:
https://github.com/PowerShell/PowerShell/blob/9947985b87dbcffc5cdafe6e412e07ccd4f3f95c/build.psm1#L249-L250
https://github.com/PowerShell/PowerShell/blob/9947985b87dbcffc5cdafe6e412e07ccd4f3f95c/build.psm1#L472-L473

3. Then you might want to realise one more thing:
```PowerShell
Write-Host $(Get-Location -Stack)
Push-Location '~/Documents'
Write-Host $(Get-Location -Stack)
Push-Location 'foo'
Write-Host $(Get-Location -Stack)
```
Now run `Pop-Location`, will you go back to '~'?

build.psm1
kwkam: Yes. How about
```PowerShell
Write-Warning "Could not find dev powershell at '$($script:Options.Output)'"
```

kwkam: Added warning.

build.psm1
TravisEz13: this is the correct casing for a variable.

build.psm1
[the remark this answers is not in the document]
kwkam: Currently, Push-Location will always push current directory into the stack, and then try to change to new directory, and that's all. Even if the 'change directory' failed, it will not throw a terminating exception nor pop from the stack to revert the action.
If this is the correct behaviour of Push-Location (e.g. by design), then it should be moved inside the try block with option `-ErrorAction Stop` to stop further action and have Pop-Location in finally block to pop the pushed location. Otherwise it should be moved outside of the try block with option `-ErrorAction Stop` to stop further action which also preventing it to "pop a location that never got pushed".

I do not see how there is no benefit for the Push-Location changes. It just depends on whether the current Push-Location behaviour is correct or not.

build.psm1
kwkam: Sorry, I am confused. Is `PwshDev` or `pwshDev` the correct casing?

kvprasoon: @kwkam  `PwshDev`

vexx32: In C# typically `camelCase` is considered correct. Largely in PS `$PascalCase` is considered better for variables, though it can come down to style in some cases.

PR #7509, opened 2018-08-13
Refactor the tokenizer

vexx32:
## PR Summary

Implements 'u' and 's' token case in the Tokenizer.cs and CharTraits.cs parsing which allows the use of uint or short literals as numbers with a 'u' and/or 's' suffix:
```powershell
PS> $Value = 108u
PS> $Value.GetType().FullName
System.UInt32

PS> $Short = 105.5us
PS> $Short
106
PS> $Short.GetType().FullName
System.UInt16
```

I also tidied up some of the code in the vicinity to inline some of the very temporary variable definitions.

Comprehensive Summary:

* Adds type suffixes for `u` and `s` - `uint` and `short`, respectively.
* Adds some logic to test whether `UInt32` or `UInt64` is the appropriate return type by comparing the multiplied value against `UIUnt32.MaxValue` when neither `l` nor `s` is supplied.
* Adds `[short]`, `[ushort]`, `[uint]`, and `[ulong]` type accelerators to help out people coming from C# or other languages. Mapped to `int16`, `uint16`, `uint32`, and `uint64`, respectively.
* Maintains parity with existing `l` / `L` suffix, where it will also perform rounding functions when applied to a `real` literal while also ensuring it still returns an appropriate `uint` or `ulong` value: `105.5u => 106`
* Defines a new flags enum `NumberSuffixFlags` that takes care of one-or-more numerical type suffixes.
* Can explicitly parse `ul`-suffixed numeric values as `ulong` and `us`-suffixed values as `ushort`.
* Explicitly specified length types for `short` and `long` as well as declared `unsigned` types **will** error out and register as an invalid numeric literal if their values after multiplication exceeds the type contraints. This is intentional; if the user specifies a value that cannot be satisfied e.g., `500sgb` (500 short, gigabyte multiplier) then the parser should not be expected to figure out what they meant. 
* Converts `strNum` parameter of `TryGetNumberValue` to `ReadOnlySpan<char>` to hopefully get a little better performance as it flies through all the `TryParse`s.

## PR Checklist

- **User-facing changes**

(not sure on *all* of those points; never worked on this repo before; feel free to let me know about stuff I'm missing)


iSazonov:
@vexx32 Please add tests.

iSazonov:
@SteveL-MSFT Should we get PowerShellCommittee approvement?

vexx32:
I'll get on that this evening. Just realised there's one more case it should handle; native ulong input.

You guys think it's worth looking at adding short type accelerators for `[uint]` and/or `[ulong]` as well to have parity with regular int accelerators?

vexx32:
OK Forgive me for not really.. being aware here. 

Are we talking xUnit or Pester tests in this instance? The latter I can do easily, the former... well, I'll have to do some reading!

vexx32:
@iSazonov hopefully these tests are workable. I tried to place them in as sensible a place as I could find; with the other parser notation tests.

I've also modified the accelerator checks to accept the new `[uint]` and `[ulong]` type accelerators. I think.

vexx32:
Okay, I think I properly fixed both the tests and my ulong parsing. 😊

src/System.Management.Automation/engine/parser/tokenizer.cs
```diff
@@ -3258,10 +3257,36 @@ private static bool TryGetNumberValue(string strNum, bool hex, bool real, char s
                         return false;
                     }
 
+                    if (suffix == 'u' || suffix == 'U')
+                    {
+                        if (UInt32.TryParse(strNum, style, NumberFormatInfo.InvariantInfo, out uint u))
+                        {
+                            ulong testresult = u * (ulong)multiplier;
```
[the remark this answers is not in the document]
vexx32: Without it, it appears to convert to int32 or int64. From what I can tell, `uint * into => int`

test/powershell/Language/Parser/TypeAccelerator.Tests.ps1
```diff
@@ -390,11 +398,11 @@ Describe "Type accelerators" -Tags "CI" {
 
             if ( !$IsWindows )
             {
-                $totalAccelerators = 94
+                $totalAccelerators = 97
```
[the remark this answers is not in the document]
vexx32: Yep. I am very confused on this one. I only added two, but somehow when the count occurs during build, there's an extra one that doesn't seem to exist... Unless the count was wrong somehow to begin with?

src/System.Management.Automation/engine/parser/tokenizer.cs
```diff
@@ -3258,10 +3257,36 @@ private static bool TryGetNumberValue(string strNum, bool hex, bool real, char s
                         return false;
                     }
 
+                    if (suffix == 'u' || suffix == 'U')
```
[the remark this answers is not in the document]
vexx32: But... My inclination for that would almost be that it would be better for it to either throw or produce an unsigned doubl, if there is such a thing.

vexx32:
Appreciate the comments so far!

Okay, so updated the tests to match everything properly, and put in `[ushort]` for `uint16` as well. I'm not sure of the prevalence of (U)int16 usage in PS, but I don't think making them a little more accessible to those coming from other languages is at all a bad idea. :grin:

src/System.Management.Automation/engine/parser/tokenizer.cs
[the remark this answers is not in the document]
vexx32: So take 2:
No, the cast is necessary. C# does not permit implicit casts from long to ulong.

vexx32:
I still don't know what was up with the VSTS build before, but everything's working now. 😄

src/System.Management.Automation/engine/parser/tokenizer.cs
```diff
@@ -3258,10 +3257,36 @@ private static bool TryGetNumberValue(string strNum, bool hex, bool real, char s
...
+                            }
+                            else
+                            {
+                                result = (uint)testresult;
+                            }
+                            return true;
+                        }
+                        else if (UInt64.TryParse(strNum, style, NumberFormatInfo.InvariantInfo, out ulong u2))
+                        {
+                            result = u2 * (ulong)multiplier;
```
iSazonov: We could remove UInt32.TryParse and use only UInt64.TryParse and cast based on MaxValue.

src/System.Management.Automation/engine/parser/tokenizer.cs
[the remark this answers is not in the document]
vexx32: Also, upon reviewing the code paths for `l` suffix... there are *two* of them. It seems a bit like there's a double-up on code paths there.

I'm not sure how the `real` boolean is determined, but `l` suffix has a code path inside the `double` parsing segment for that as well as one outside and after it.

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: Hmm. That is true. And considering that I'm already half doing that having to account for the multiplier... yeah, that makes sense. I'll do that, thanks!

vexx32:
Alright, I officially give up on matching whatever wacky indentation was initially here. It's good enough.

src/System.Management.Automation/engine/parser/tokenizer.cs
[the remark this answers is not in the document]
vexx32: Been thinking on this, and I think that it would *probably* be best to maintain parity with the existing `L` token suffix. 

Adding a small branch for that code path in the `if (real){}` block as well, mimicking the doubling up the `long` parsing has. (I do wish there was a good way to consolidate them, but it looks like we're either gonna be doubling `real` checks for every parse or just doubling a couple of parse checks).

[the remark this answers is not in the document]
vexx32: And done!

Had a slip up where I forgot my `u` token checks were before the check for `real` which caused a slight issue for a moment.

Also added a test or two to check the round-on-parse behaviour.

src/System.Management.Automation/engine/parser/tokenizer.cs
```diff
@@ -3272,6 +3270,16 @@ private static bool TryGetNumberValue(string strNum, bool hex, bool real, char s
                             {
                                 result = ((long)Convert.ChangeType(d, typeof(long), CultureInfo.InvariantCulture)) * multiplier;
                             }
+                            if (suffix == 'u' || suffix == 'U') {
+                                ulong testresult = ((ulong)Convert.ChangeType(d, typeof(ulong), CultureInfo.InvariantCulture)) * (ulong)multiplier;
```
lzybkr: >Convert.ChangeType [](start = 59, length = 18)

This can raise an exception. The parser and tokenizer try hard to avoid any exception.

src/System.Management.Automation/engine/parser/tokenizer.cs
```diff
@@ -3272,6 +3270,16 @@ private static bool TryGetNumberValue(string strNum, bool hex, bool real, char s
                             {
                                 result = ((long)Convert.ChangeType(d, typeof(long), CultureInfo.InvariantCulture)) * multiplier;
                             }
+                            if (suffix == 'u' || suffix == 'U') {
```
lzybkr: >suffix == 'u' || suffix == 'U' [](start = 32, length = 30)

If we are adding `unsigned` support, it seems like we should also support `ul`.

src/System.Management.Automation/engine/parser/TypeResolver.cs
```diff
@@ -739,7 +739,7 @@ internal static class CoreTypes
                     { typeof(Guid),                                        new[] { "guid" } },
                     { typeof(Hashtable),                                   new[] { "hashtable" } },
                     { typeof(int),                                         new[] { "int", "int32" } },
-                    { typeof(Int16),                                       new[] { "int16" } },
+                    { typeof(Int16),                                       new[] { "short", "int16" } },
                     { typeof(long),                                        new[] { "long", "int64" } },
```
lzybkr: Script portability might suffer a bit by adding these type accelerators. I can see C# developers adopting them, and  older versions of PowerShell obviously don't have them. I think there should be some consensus that adding these is a worth it.

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: I simply copied what was already being used in the parser.

Suggested alternatives?

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: I thought it better to simply have the parser infer `ulong` from overlarge `<int>u` literals. But would it be better to have the `ul` explicitly permitted?

Are two-letter suffixes even supported currently?

src/System.Management.Automation/engine/parser/TypeResolver.cs
vexx32: Can you elaborate on why portability might suffer? Are you referring more to backportability?

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: So in other words, what you're asking isn't really plausible without significantly changing how suffix detection works.

I would argue that is a more breaking change.

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: This follows the same code path as the original calls, where the double parse is confirmed before the conversion occurs. Shouldn't be an issue. :)

src/System.Management.Automation/engine/parser/tokenizer.cs
```diff
@@ -3283,6 +3291,28 @@ private static bool TryGetNumberValue(string strNum, bool hex, bool real, char s
                         return false;
                     }
 
+                    if (suffix == 'u' || suffix == 'U')
+                    {
+                        if (UInt64.TryParse(strNum, style, NumberFormatInfo.InvariantInfo, out ulong u))
+                        {
+                            ulong testValue = u * (ulong)multiplier;
```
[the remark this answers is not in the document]
vexx32: From what I can tell `multiplier` is *always* a `long` value in here. All other suffix-specified conversions here do *not* use `biginteger`. That usage only occurs if *no* type specification has occurred and the number is not a `real` value (which I assume means 'contains a decimal point').

So what I've done here essentially reflects how it's already functioning. Should I change this?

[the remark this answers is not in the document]
lzybkr: Try: `18446744073709551615ukb` - that will overflow.

src/System.Management.Automation/engine/parser/TypeResolver.cs
vexx32: Gotcha. I can see that being somewhat of a concern on one hand, but I also feel that if we avoid making QOL changes just because someone may still be on an older version of PS doesn't... really hold that much weight. 

We're talking about PS Core here, after all, and there are already several things similarly basically just quality-of-life changes that have been included. Example that I can think of right this minute would be the character support for range operators. `'a'..'z'` doesn't work in WinPS on any version at all.

Certainly, it may hurt backportability slightly, but this does not remove the existing type accelerators -- if one wishes to use those instead (i.e., `[uint32]` instead of `[uint]`) for compatibility reasons, there is nothing stopping you from doing so for compatibility concerns.

As with many other aspects of PS, if you need to write scripts for older variants of PS, you can test them against those old versions first. Few scripts written in PS 5.1 that I come across are instantly compatible with v3, often requiring some small changes, and forget about v2 entirely. This would simply become another of those things (like the char support for range operator) that you have to spend a little extra typing on if you're concerned about backportability.

src/System.Management.Automation/engine/parser/tokenizer.cs
lzybkr: It's a slightly more involved change, sure, but it's still a simple change and one that people will expect as long as you're adding unsigned support.

As for a breaking change, `u` is a breaking change as well and it's better to cover all expected use cases at once if taking the breaking change.

src/System.Management.Automation/engine/parser/tokenizer.cs
lzybkr: So you're saying `-1.0u` doesn't raise an exception?

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: Touche!

Okay, let me see if I can trace that down... `suffix` is referenced as a `char` value basically everywhere... This is gonna take some thought.

src/System.Management.Automation/engine/parser/tokenizer.cs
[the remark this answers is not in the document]
vexx32: Sure, but the same sort of idea with `l` suffix will also overflow currently. This is no different to existing behaviour.

```
PS> 9223372036854775807lkb
At line:1 char:23
+ 9223372036854775807lkb
+                       ~
The numeric constant 9223372036854775807lkb is not valid.
+ CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
+ FullyQualifiedErrorId : BadNumericConstant

PS > 18446744073709551615ukb
At line:1 char:24
+ 18446744073709551615ukb
+                        ~
The numeric constant 18446744073709551615ukb is not valid.
+ CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
+ FullyQualifiedErrorId : BadNumericConstant
```

This correlates precisely with established behaviours. Should I change both these code paths?

src/System.Management.Automation/engine/parser/tokenizer.cs
iSazonov: c# uses one suffix 'u' for int and long
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/uint
> Integer literals can also include a suffix that denotes the type. The suffix U or 'u' denotes either a uint or a ulong, depending on the numeric value of the literal. The following example uses the u suffix to denote an unsigned integer of both types. Note that the first literal is a uint because its value is less than UInt32.MaxValue, while the second is a ulong because its value is greater than UInt32.MaxValue.

vexx32: @lzybkr consider this one done and dusted.

References replaced, quotation marks swapped out, null terminator accounted for (not even sure why that's necessary, but... it is! apparently! I am unsure why!), and it correctly parses `ul` without allowing strange nonsense suffixes like `ll` or anything even longer that it could theoretically now support.

vexx32: @iSazonov Good to know! We can always revert the two-character implementation of `suffix` if you feel two characters is too many. It *does* open up some possibilities, however. I'll leave that decisions up to you folks; I personally am perfectly happy with a single `u` although I do appreciate the parity of `l` and `ul`.

As for parsing *everything* as `BigInteger`... I don't think that's particularly efficient. Especially since half our numerics are `double` or `decimal`!

Additionally, parsing everything as `BigInteger` unnecessarily would a) go against the decision made by you and others in #7510 and b) likely increase the memory footprint of the parsing method appreciably (although admittedly I can only speculate on that).

src/System.Management.Automation/engine/parser/tokenizer.cs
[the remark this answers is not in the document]
lzybkr: Sorry, it's been a long time since I wrote that code and didn't look at the surrounding code.  The entire function is guarded with by `checked` and `try/catch` - that code didn't show up in the diff and I forgot about it.

The intent of the parser is to never raise an exception, but apparently there were cases I couldn't reasonably avoid.

So you're fine here.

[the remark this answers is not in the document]
lzybkr: Or I should say - it's not a real problem, but you could still use `BigInteger` here to avoid the exception for in theory better performance.

src/System.Management.Automation/engine/parser/tokenizer.cs
lzybkr: You are conflating 2 things - reporting an error does not require raising an exception.

As I stated in another comment, the intent is to never raise an exception while parsing, but this function does sometimes, so I guess it's fine to do so here too.

The parser *by design* avoids exceptions for performance reasons. Exceptions should be exceptional, and because the parser is used in interactive scenarios, errors are not exceptional.

So yes - report an error, but not via an exception.

src/System.Management.Automation/engine/parser/tokenizer.cs
```diff
@@ -3266,14 +3266,14 @@ private static bool TryGetNumberValue(string strNum, bool hex, bool real, char s
...
+                            if (suffix == "l" || suffix == "L")
                             {
                                 result = ((long)Convert.ChangeType(d, typeof(long), CultureInfo.InvariantCulture)) * multiplier;
                             }
-                            if (suffix == 'u' || suffix == 'U') {
+                            if (suffix[0] == 'u' || suffix[0] == 'U') {
                                 ulong testresult = ((ulong)Convert.ChangeType(d, typeof(ulong), CultureInfo.InvariantCulture)) * (ulong)multiplier;
 
-                                if (testresult > UInt32.MaxValue) {
+                                if (testresult > UInt32.MaxValue || String.Compare(suffix, "ul", true) == 0) {
```
lzybkr: We avoid this overload in favor of the one taking `StringComparison.OrdinalIgnoreCase`.

src/System.Management.Automation/engine/parser/tokenizer.cs
```diff
@@ -3492,10 +3534,20 @@ private string ScanNumberHelper(char firstChar, out bool hex, out bool real, out
...
+            if (c.IsTypeSuffix())
+            {
+                SkipChar();
+                suffix += c;
                 c = PeekChar();
             }
 
+            // Due to the suffix being a string instead of char, it must be null-terminated for console
+            // to behave correctly. If someone knows why, I will be happy to add a proper explanation.
+            suffix += "\0";
```
lzybkr: This is horrible for perf - it will allocate a new string for every single number parsed.

It's probably cleaner to fully parse the suffix into flags here (as a flags enum) and pass that instead of a string.

src/System.Management.Automation/engine/parser/tokenizer.cs
iSazonov: We use BigInteger here only for excluding an overflow not for expose it public.

As for memory footprint I think we should use ReadOnlySpan<char> - replace `string strNum` with `ReadOnlySpan<char> strNum` - all TryParse-s use ReadOnlySpan<T> internally.

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: That is likely to be correct.

I guarantee you I do not have the knowledge at this point in time to be able to do that without some significant pointers from you fellas on how you would want that to be implemented.

Char based flag enum? Something like that?  I am game to give it a shot, but I've never worked with flag enums internally before. Need at the very least a good idea of the desired method of implementation to start on something like this.

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: Gotcha. Yeah, I'm not familiar enough with c# or the internals of how PS handles errors vs exceptions to be clear on that difference at the moment. Thank you!

src/System.Management.Automation/engine/parser/TypeResolver.cs
steviecoaster: I think adding this functionality is a welcome move, it does add quite a bit of flexibility.

src/System.Management.Automation/engine/parser/tokenizer.cs
lzybkr: I was thinking of something like:

```c#
[Flags]
enum NumberSuffix
{
    None = 0x0,
    Long = 0x1,
    Double = 0x2,
    Unsigned = 0x4,
}
```

And by parsing, maybe something like:

```c#
            var suffix = NumberSuffix.None;
            c = PeekChar();
            if (c.IsTypeSuffix())
            {
                SkipChar();
                switch c {
                case 'l': case 'L':
                    suffix |= NumberSuffix.Long;
                    break;
                case 'd': case 'D':
                    suffix |= NumberSuffix.Double;
                    break;
                case 'u': case 'U':
                    suffix |= NumberSuffix.Unsigned;
                    break;
                }
                c = PeekChar();
                if (c.IsTypeSuffix())
                {
                    SkipChar();
                    switch c {
                    case 'l': case 'L':
                        suffix |= NumberSuffix.Long;
                        break;
                    case 'u': case 'U':
                        suffix |= NumberSuffix.Unsigned;
                        break;
                    }
                    c = PeekChar();
                }
            }
```

vexx32: Woah, that *is* neat! Thank you! I'll see how that goes!

Would it be most appropriate to have the enum in its own file, or as part of the `tokenizer.cs` file?

src/System.Management.Automation/engine/parser/tokenizer.cs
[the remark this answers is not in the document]
vexx32: I'll look at that possibility. Is using `bigint` really better than throwing the exception and having it caught?

src/System.Management.Automation/engine/parser/tokenizer.cs
lzybkr: I'm a fan of keeping related types in the same file, especially when they are small and not used elsewhere, so I'd keep it in `tokenizer.cs`.

lzybkr: And note my sample code allows something it probably shouldn't - I'll let you figure out what that is and avoid the problem :)

vexx32: I can see it would allow `dl` or `du` suffixes. In fact, any suffix ending in `u` shouldn't be necessary or permitted.

It's also got `double` instead of `decimal` ;)

lzybkr: Of course - `d` is `double` in C, C++, and C#, but not PowerShell.

src/System.Management.Automation/engine/parser/tokenizer.cs
iSazonov: You can look https://source.dot.net/#System.Private.CoreLib/shared/System/Int64.cs,517dc8207c8fc99e
- in .Net Core all parse methods with string parameter is a wrapper for methods with ReadOnlySpan.

ReadOnlySpan is very useful for parsers. Here we can surely get benefits.

vexx32: I'm sure we can, but I'm afraid the changes required for *that* particular wrinkle might be above my abilities at present. I'll revisit the possibility after I take care of the other concerns raised and see what I can do

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: Yeah. Decimal is `m` in those, if I remember my C# correctly. PS just infers double from a non-integer real and only bothers specifying the rest, heh.

vexx32: OK think this is done. Building to check now...

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: @iSazonov from what I can see the parsing methodology is already by-each-character as it would be in that construction.

It's also building the string back up with `StringBuilder` before actually returning it, so it's going to be fairly efficient. 

I think I can say with reasonable confidence that at this point this may be superfluous. The only place that this could reasonably be done would be with the entire `_script` variable here. It may well offer some performant benefit if implemented here instead of making this a string. From what I can see there may be an implicit cast between `string` and `ReadOnlySpan<char>` which may allow me to simply type case the private variable and call it a day.

Everything I'm seeing interact with the _script data itself is actually pulling the data by index anyway, so... this may be a fairly simple thing too. But in terms of using it in *this* parsing segment it's not possible, because it's a read only span and cannot be built in the same way StringBuilder does as far as I know; this is necessary in order to pass the complete token around. 

I don't know that it makes sense to pass the token around as ReadOnlySpan<char> since it's already being built up as string (and I don't think there's an easy way to do such a thing with ReadOnlySpan<char>).

Your thoughts on this one, @lzybkr ?

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: Oooh, fun, it was still parsing cases where I had `dl` and `ll` with... mixed results. Let's... *not* do that. 

Unless we *really* want `ll` -- it could be `long long` aka `biginteger` ;)

lzybkr: I do like what @iSazonov suggested in using `Span` - it might even end up being less code, it's probably worth a look.

src/System.Management.Automation/engine/parser/tokenizer.cs
```diff
@@ -62,6 +62,35 @@ public enum DynamicKeywordBodyMode
         Hashtable = 2,
     }
 
+    /// <summary>
+    /// Defines the permissible suffixes for numerical tokens.
+    /// </summary>
+    [Flags]
+    public enum NumberSuffixFlags
```
lzybkr: I would only make this public if we exposed the flags in the Token api, which could be nice but isn't important.

vexx32: Roger that~! I'll get on it.

Aside: not sure why, but the Get-WinEvent tests are breaking, and I'm *pretty darn sure* I've not changed anything in that code specifically...

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: After talking with some of the folks on the PS discord (who also seem to agree) I may give this a shot.

For clarity: are we talking about putting the entire `_script` into a span instead of string, and then rather than building substrings as happens in the number token helper method, I presume I'd have it pass around smaller pieces of that grabbed by index?

lzybkr: Just the suffix would go in a `Span`.

src/System.Management.Automation/engine/parser/tokenizer.cs
lzybkr: Just remove `public` - you don't need `internal`.

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: Huh. Seems like that would be... minimal in terms of gain, but I'll take a look.

vexx32: @lzybkr had a bit of a fiddle with it, but from what I can see, reworking the suffix as a two-char array is going to make significantly more fragile code. I'd rather not create a mess someone else has to deal with.

As it is, the use of switch statements and flags effectively make it relatively trivial to add an additional type suffix in the event we ever need any others included. Swapping this particular bit out for a `Span<char>`  isn't worth creating the fragile `if (suffix[0] == 'x' && suffix[1] != 0)` bits and pieces that would subsequently litter `TryGetNumberValue`

vexx32: The code is cleaner and more flexible using flags here, I think, unless there's a better way to check the values in the Span<char> that I'm missing.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this.  We have already taken many other changes that impact portability to older PowerShells, but the workaround is (as noted above) not to use these newer features.  Committee accepts this change.

vexx32:
Thanks, guys! I've been poking at those AppVeyor/VSTS build errors, but as far as I can see, I don't think there's anything there that I've broken myself. Looks like some modules aren't loading properly or something?

vexx32:
@iSazonov ; @lzybkr 

I know it's a bit on the late end of things, but it just occurred to me that I could also add an `s` suffix to denote `short` (`(u)int16`) values. Wouldn't really take all that much to add. Do you think it's worth doing?

src/System.Management.Automation/engine/parser/tokenizer.cs
iSazonov: My suggestion is:
```c#
private static bool TryGetNumberValue(ReadOnlySpan<char> strNum, bool hex, bool real, char suffix, long multiplier, out object result)

....
strNum = strNum.Slice(1);
```

src/System.Management.Automation/engine/parser/tokenizer.cs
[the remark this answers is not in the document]
iSazonov: Yes, one additional comparison is better the exception.

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: There is no slicing happening in this method. There is no purpose to that. This would confer little to no benefit; `strNum` here in `TryGetNumberValue` is only passed directly to `TryParse` methods in sequence to try to get a value, and this will rarely be a long string.

The only way I can see this being of use to be more efficient is if we're already passing around slices from a `ReadOnlySpan<char>` that comes all the way from `_script`. If *that* becomes a `Span`, all the strings in this class that get passed around can derive from slices of it, which would work wonderfully, I'm sure.

But short of that I don't see a lot of reasons to change the `TryGetNumberValue` method in this manner. (If you want me to have a crack at converting everything over to the `ReadOnlySpan<char>` type from the `_script` basis onwards I think I can do that.)

src/System.Management.Automation/engine/parser/tokenizer.cs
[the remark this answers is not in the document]
vexx32: Sure, but from the little I know `BigInteger` is generally avoided moreso for memory consumption issues; the documentation itself warns that it can cause *memory* overflow issues. I'm not sure that defaulting to it is the wisest cause of action, in all honesty, not least because when parsing any appreciably large value memory consumption may rise noticeably.

Additionally, we would need to define an arbitrary standard value of where to call it a `double` instead and cast back to that if we don't want to store ginormous integers in memory after parsing...

iSazonov:
I think that this is not so important for the scripting language, but if we have already added a lot of suffixes, then we should have a full set of suffixes.

Please update the PR title and description to describe latest enhancements.

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: Although, since all the TryParse methods have a public overload in .NET Core 2.1 that accepts this data type it would save all of *that* extra data being cloned in memory.... okay. 

Now I see why you were saying that; I misunderstood your intention there. I still think it would be *far* more worthwhile to have `_script` begin as a `ReadOnlySpan` simply for the slicing that already happens on that level to be more efficient, but there may be other limitations with this type that I'm not familiar with as well, so... not sure.

I'll go ahead and get the ReadOnlySpan applied to the TryGetNumberValue...

vexx32:
Roger that, will do. Double checking that it builds correctly with ReadOnlySpan in TryGetNumberValue parameter, and then I'll push that too.

vexx32:
So this builds fine locally and I've tested the packagemanagement imports (the ones giving the build grief) and I'm not seeing issues on my Win10 system.

Not sure what's breaking the build but as far as I can tell it doesn't seem to be my changes here. :/

vexx32:
Awesome, looks like it's just failing on some 'uploading artifacts' tasks on the VSTS builds. I won't pretend to know what that's about but it looks like something being addressed in #7535 😄

lzybkr:
I don't see a need for a `short` suffix - coercion to smaller types isn't usually necessary, but in the interest of completeness, I don't care either.

That said, a byte suffix is needed for completeness, and that needs to be considered by the committee - because `b` could also be used for binary literals.

vexx32:
Actually, the topic of a byte / binary signifier was brought up in discussion of this in the PS Slack.

That idea was mentioned, as well as the possible alternative for `0b000001` type syntax to mirror the hex syntax. Since it looks like that parsing is also done in the tokenizer here, I would be happy to explore either of those options.

SteveL-MSFT:
@vexx32 I would suggesting having the byte and binary in a separate PR with separate discussion

vexx32:
Fair enough. :)

I think I've done as much as is sensible at this point, then!

src/System.Management.Automation/engine/parser/tokenizer.cs
rjmholt: @vexx32 The `Get-WindEvent` tests are currently just volatile and need fixing -- nothing to do with your changes

test/powershell/Language/Parser/Parser.Tests.ps1
```diff
@@ -648,39 +648,50 @@ foo``u{2195}abc
...
+            @{ Script = "0u"; Expected = [UInt32]::MinValue }
+            @{ Script = "15u"; Expected = "15" }
+            @{ Script = "1.5u"; Expected = "2" }
+            @{ Script = "32.1u"; Expected = "32" }
+            @{ Script = "4294967295u"; Expected = $([UInt32]::MaxValue) }
+            @{ Script = "401ul"; Expected = "401" }
+            # Banker's rounding on casting, of course.
+            @{ Script = "1024.5ul"; Expected = "1024" }
+            @{ Script = "4294967301u"; Expected = "4294967301" }
+            @{ Script = "18446744073709551615u"; Expected = $([UInt64]::MaxValue)}
```
rjmholt: Can you add tests for the following cases please:
- `short` literals
- `short unsigned` literals
- `decimal` literals if the tests don't already exist (I know decimals were already supported, but if we're touching them we should test them)
- Various barewords that are "almost" number literals:
   + Numbers that ends with a non-numerical suffix like `7p`
   + Numbers that have a `u` but then have a non-numerical suffix like `20ux`
   + Numbers that use numerical literals but in a bad order or combination like `300ls` or `78su`
- Ideally we'd have some extra tests for case-insensitivity too, but they are less crucial.

I know it's onerous, but the tokenizer is the kind of place where good coverage is crucial. Having read through the code it all looks pristine, but testing like this is most important for picking up regressions later on

test/powershell/Language/Parser/Parser.Tests.ps1
```diff
@@ -648,39 +648,50 @@ foo``u{2195}abc
...
+            @{ Script = "0x0"; Expected = "0" }
+            @{ Script = "0xF"; Expected = "15" }
+            @{ Script = "0x80000000"; Expected = $([int32]::MinValue) }
+            @{ Script = "0xFFFFFFFF"; Expected = "-1" }
+            @{ Script = "0x7fffffff"; Expected = $([int32]::MaxValue) }
+            @{ Script = "0x100000000"; Expected = [int64]0x100000000 }
+            #Tests for uint notation.
+            @{ Script = "0u"; Expected = [UInt32]::MinValue }
+            @{ Script = "15u"; Expected = "15" }
+            @{ Script = "1.5u"; Expected = "2" }
```
rjmholt: I'll just comment and say that the self-rounding literal seems like a very odd concept, but I'm assuming the @PowerShell/powershell-committee already OK'd that

src/System.Management.Automation/engine/parser/tokenizer.cs
```diff
@@ -3260,18 +3294,41 @@ private static bool TryGetNumberValue(string strNum, bool hex, bool real, char s
...
-                            if (suffix == 'l' || suffix == 'L')
+
+                            if (suffix == NumberSuffixFlags.Long)
                             {
                                 result = ((long)Convert.ChangeType(d, typeof(long), CultureInfo.InvariantCulture)) * multiplier;
                             }
+
+                            if (suffix == NumberSuffixFlags.Short)
+                            {
+                                result = (short)((short)Convert.ChangeType(d, typeof(short), CultureInfo.InvariantCulture) * multiplier);
```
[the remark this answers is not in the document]
vexx32: I'll see how it behaves with that conversion in the tokenizer. It's been my experience so far that Convert.ChangeType() seems to do Banker's Rounding whereas casting and some other methods... do different things.

test/powershell/Language/Parser/Parser.Tests.ps1
vexx32: I have designed and interactively tested that incorrectly ordered literals would fail. I'm not sure... how that particular test will respond; the tokenizer throws a 'not a valid numeric literal' when those are entered.

Should I create a new block of tests for ones that throw?

rjmholt: Yeah that test will just flat fail for a failed conversion. Looks like there are two options:
- Turn the body into a `try`/`catch` so that it "returns" the error to compare to.
- Or, more cleanly, write a new test batch for tokenizer errors. The second is also nice because it will encourage that in future, and error cases are just as important to test in my view.

vexx32: Option 2 it is. :)

I'll probably leave off with just testing the tokens I've added and combinations thereof for now, but there'll be plenty of room for adding tests to ensure the others fail when they ought as well. :)

vexx32: @rjmholt turns out you have to try really hard to hide parse errors from the parser that you *want* to throw 😄 

Had to go from string -> script block -> invoke that *in a script block* -> pass to `| Should -Throw`. Otherwise it throws before it gets passed into `Should`!

src/System.Management.Automation/engine/parser/tokenizer.cs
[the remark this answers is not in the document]
vexx32: That's probably also functional. But as far as I can tell that offers no noticeable advantage over `Convert.ToInt16(d * multiplier)` which is what I've gone with at the present.

src/System.Management.Automation/engine/parser/tokenizer.cs
```diff
@@ -3260,41 +3294,100 @@ private static bool TryGetNumberValue(string strNum, bool hex, bool real, char s
...
-                                result = ((long)Convert.ChangeType(d, typeof(long), CultureInfo.InvariantCulture)) * multiplier;
+                                result = Convert.ToInt64(d * multiplier);
+                            }
+                            else if (suffix == NumberSuffixFlags.Short)
+                            {
+                                result = Convert.ToInt16(d * multiplier);
+                            }
+                            else if (suffix.HasFlag(NumberSuffixFlags.Unsigned))
+                            {
+                                ulong testresult = Convert.ToUInt64(d * multiplier);
```
iSazonov: All methods like `Convert.ToUInt64` internally test MaxValue and MinValue. We could expand the methods, do the checks explicitly and exclude exceptions at all.

vexx32: You'll have to elaborate a bit, I'm not sure quite what your intent is. :)

test/powershell/Language/Parser/Parser.Tests.ps1
rjmholt: Yeah, it's a classic gotcha — you can't execute script before you parse it. Same reason why you can't use `Import-Module` to import PowerShell classes to use with other classes.

So you have to parse it at runtime with something like `[scriptblock]::Create("program")` or `[System.Management.Automation.Language.Parser]::Parse(...)`

src/System.Management.Automation/engine/parser/tokenizer.cs
iSazonov: Sample for a short above ( `result = Convert.ToInt16(d * multiplier);`):
```c#
var value = d * multiplier;

if (value < short.MinValue || value > short.MaxValue)
{
    result = null;
    return false;
}

result = (short)value;
return true;
...
```

src/System.Management.Automation/engine/parser/tokenizer.cs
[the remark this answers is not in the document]
rjmholt: Just to go on the record -- since we're looking for specific rounding behaviour here, my preference would be to be explicit about that rather than getting it from casts

src/System.Management.Automation/engine/parser/tokenizer.cs
iSazonov: We could replace the `if`-s with `switch` to get more readable code.

test/powershell/Language/Parser/Parser.Tests.ps1
iSazonov: We have `[system.management.automation.psparser]::Tokenize`.

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: We can't do a direct cast, because that truncates the double instead of rounding it properly. But yes, if we do that and then call Int16.Parse it should work.

vexx32: Re if -> switch: there are parts of this that you can *sort of* get prettier code. If this were PS I could throw the entire thing in a gorgeous switch. But... It's not. 

So the switches are pretty limited in scope here, but I'll see what I can do.

iSazonov: > We can't do a direct cast, 

You can look how CoreFX do this in Convert class for double
https://source.dot.net/#System.Private.CoreLib/shared/System/Convert.cs,938

iSazonov: We can grap from CoreFX needed Convert.Int16/Int32/Int64 methods and put them in Utils.cs (without throw statement).

vexx32: Do we have a `Utils.cs` in this repo? Or am I making a new one?

iSazonov: I think it is best place https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/Utils.cs

rjmholt: As opposed to [PsUtils.cs](https://github.com/PowerShell/PowerShell/blob/5d03e1653a7d518715fa3f00587cad6b5c78cc89/src/System.Management.Automation/utils/PsUtils.cs)

vexx32: Thanks, guys! :D

Okay, got some helper functions down, trying to work on sensible ways to have the code structured.

vexx32: Well let's see how that goes...

vexx32:
Huh, that's interesting...

@lzybkr @rjmholt OK, so now I need to ask. Am I *supposed* to let large hex number patterns read as negatives? Are they *supposed* to only fit into, like, int16 / int32 hex block patterns?

Because if so, I can do a simple length check and see what it needs to get parsed as. Or just flat out parse hex *only* as `[u]int[16|32|64]`

What exactly is the desired behaviour of hex literals? Because in my parsing them as `bigint` either I've made them better or worse, and I don't work with them enough to know which is which!

vexx32:
Okay I talked it over with @jaykul in the Slack and came to the conclusion that:

1. We should probably respect sign bits at int32 and int64 levels, at the very least
2. There may be a potential (if highly edge case) use for larger hex literals, so I haven't limited them to int values; they will happily parse as decimals for the moment or doubles if they get high enough (though at that point their utility is questionable, since there's no accuracy at that high of a value for a double.)
  * The sign bit is every 8 hex characters it would seem, so higher values respect that pattern also, at the present time. 

How do you guys want to handle large hex values?

rjmholt:
Hmmmmmmm I don't have an immediate intuition on how we ought to handle sign in hex. It's the kind of deep language question we should consult a few sources on. Let me do some research 😄

vexx32:
Appreciate it, Robert!

OK, I *think* what I have might be OK.

Currently (if the string is a hex string) it does this:

1. Check length of hex string.
2. If it's not divisible by 8, assume there isn't a sign bit and prepend a `0` to make sure `BigInteger.TryParse` doesn't assume the first character is meant to determine sign.

And now I'm trying to figure some decently uncomplicated logic to work out whether or not to flag the number as `Unsigned` if the sign bit is 7 or less. So far... tricky.

vexx32:
OK So essentially if you have a hex value that is a multiple of 8 digits it expects a sign bit and adheres to it. Otherwise it assumes the value is positive when parsing (if you have a negative sign before the hex literal it is still respected and flips the final value).

It doesn't really make... sense... (in my opinion) for the value to go into doubles territory -- that's kind of against the point of *using* hex (unit precision) so since we already parse as `bigint` in this region (and always have, I've just kind of rolled it all together here) I figured if you're asking for a hex literal *that* big you might actually need it for some insane scenario and thought -- why not just yield the `bigint` value already there for the taking.

I'd love to hear your thoughts on it, though, as always! This is filtered through my own brain, so there may well be stuff in here that we should avoid doing for reasons I haven't thought of yet (yes `bigint` can cause OOM errors but the parser is already using it to check bounds anyway and always has; this won't introduce any new overflow possibilities that weren't already there just by virtue of entering stupidly huge decimals).

@iSazonov @rjmholt @lzybkr

BrucePay:
@SteveL-MSFT I don't care about the back porting issue but I worry a great deal that so many changes to PowerShell's numeric processing will break backwards compatibility in a significant way. I think we need to revisit the impact of this change at next week's committee meeting.

iSazonov:
As for sign:
1. Currently PowerShell implement 
```powershell
-0x1A

-26
```
2. Currently PowerShell CAN consider hex numerics as signed
```powershell
>0xFFFFFFFF

-1

>(0xFFFFFFFF).Gettype()

True     True     Int32                                    System.ValueType
```
3. Current PowerShell behavior is to consider high bit as sign _because target type always is signed_ - int32/64/long/double/BigInteger.

4. If target type is unsigned (ushort,uint,ulong and so on) PowerShell MUST to consider high bit as value not sign
The behavior is in CoreFX https://source.dot.net/#System.Private.CoreLib/shared/System/Number.Parsing.cs,890
```
5. Because we want add binary literals we should consider them here too. PowerShell must process binary literals in the same manner as hex literals.

**Comment by iSazonov** [comment]
@vexx32 The PR is very large (many comments and commits). I suggest split the PR to simplify.
1. Make new PR with new prefixes and tests - it already was approved by PowerShell Committee and we can fast review and merge.
2. Make new PR with NumberSuffixFlags in ScanNumberHelper (using switch) + replace `Convert.ChangeType` with new convert helper methods - the temporary PR _without_ any fuctional changes can be fast reviewed and merged.

After that we can close current PR and open new one with cleaned commits based on PowerShell Committee conclusion - commits with new suffixes and perhaps changes related hex/binary sign.

**Comment by vexx32** [comment]
(renamed for clarity, I went a bit beyond the original intent here) ;)

**Comment by Jaykul** [comment]
This seems like a REALLY good candidate for that feature flag testing that's been discussed before.

Not on the sense that is simple to toggle, but in the sense that this big of a change ought to be tested ... a lot ... In such a way that people can initially opt in to the new parser, then eventually make it opt-out ... And if all goes well, remove the old code?

**Comment by vexx32** [comment]
FWIW; After submitting #7575 I've been toying with possible improvements and have a much cleaner and more polished version of this code at https://github.com/vexx32/PowerShell/tree/Tokenizer/Refactor

It's not yet a PR as it is a branch off of the code in #7575 and includes those changes already.

**Comment by SteveL-MSFT** [comment]
@PowerShell/powershell-committee re-reviewed this and is fine with the proposal (this is not a PR review approval)

File: src/System.Management.Automation/engine/parser/tokenizer.cs [hunk 16]
```diff
@@ -3492,19 +3617,84 @@ private string ScanNumberHelper(char firstChar, out bool hex, out bool real, out
...
             if (c.IsMultiplierStart())
             {
                 SkipChar();
 
-                if (c == 'k' || c == 'K') { multiplier = 1024; }
-                else if (c == 'm' || c == 'M') { multiplier = 1024 * 1024; }
-                else if (c == 'g' || c == 'G') { multiplier = 1024 * 1024 * 1024; }
-                else if (c == 't' || c == 'T') { multiplier = 1024L * 1024 * 1024 * 1024; }
-                else if (c == 'p' || c == 'P') { multiplier = 1024L * 1024 * 1024 * 1024 * 1024; }
+                switch (c)
```
lzybkr: Going from 5 lines to 22 lines of code is an improvement? 🙂 
Theoretically a switch might be faster if the jit generates a jump table, but if performance were the concern, I'd use a statically initialized array and replace these 5 lines of code with:
```
multiplier = multipliers[c];
```

src/System.Management.Automation/engine/parser/tokenizer.cs
```diff
@@ -3237,130 +3272,219 @@ private void ScanNumberAfterDot(StringBuilder sb, ref int signIndex, ref bool no
...
                     {
-                        decimal d;
-                        if (Decimal.TryParse(strNum, style, NumberFormatInfo.InvariantInfo, out d))
+                        if (Decimal.TryParse(strNum, style, NumberFormatInfo.InvariantInfo, out decimal d))
                         {
                             result = d * multiplier;
                             return true;
                         }
 
+                        // Has double suffix but cannot be parsed as double.
```
lzybkr: Has decimal suffix but ...

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: Ooh, that's a good idea! I'll borrow that and submit it in the next PR. Feel free to review this further if you want, but be aware that 90% of the code here has been completely overhauled in the upcoming PR (and it has fancy tidbits with binary parsing and underscore support yay!) for y'all to have fun combing through as I re-learn my C# ;)

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: Haha! That comment is incorrect. It wouldn't get promoted; that would fall through the the result = null; return false; statements.

I'll... quietly make sure I've caught that in my more recent work.... XD

vexx32:
Closing this for now so we don't waste anymore time here. The base of the stuff that was meant to happen here is in #7575, and the rest needs a great deal of clean up that will be in an upcoming PR.

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: Roger that! :)

PR #7633, opened 2018-08-25
Add ITuple checks to PSGetIndexBinder

### LINKED ISSUE #7632: Enable index operations for ITuple implementations
From the discussion in #7471, the following should work

```powershell
PS> $tuple = [Tuple]::Create(1, 'test')
PS> $tuple[0]
# 1
PS> $tuple[1]
# test
PS> $tuple[0..1]
# 1
# test
PS> $tuple[-1]
# test
```

SeeminglyScience:
## PR Summary

Resolves #7632.

This change enables index operations on objects that implement `ITuple`, including slicing and negative indexing.

Add checks in `PSGetIndexBinder` for types that implement an interface with `DefaultMemberAttribute` declared, such as `ITuple`. `Tuple` and `ValueTuple` implement `ITuple` interface, and thus they are made indexable by this PR.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
+++ b/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
@@ -3946,7 +3946,7 @@ namespace System.Management.Automation.Language
                 // This check will catch those cases.
                 if (defaultMember == null)
                 {
-                    defaultMember = i.GetCustomAttribute<DefaultMemberAttribute>(inherit: true);
+                    defaultMember = i.GetCustomAttributes<DefaultMemberAttribute>(inherit: false).FirstOrDefault();
                     if (defaultMember != null)
                     {
                         lengthProperty = i.GetProperty("Count") ?? i.GetProperty("Length");
@@ -4045,6 +4045,17 @@ namespace System.Management.Automation.Language
             MethodInfo indexer,
             ParameterInfo[] getterParams)
         {
+            // PowerShell supports negative indexing for types that meet the following criteria:
+            //      - Indexer method accepts one parameter that is typed as int
+            //      - The int parameter is not a type argument from a constructed generic type
+            //        (this is to exclude indexers for types that could use a negative index as
+            //        a valid key like System.Linq.ILookup)
+            //      - Declares a "Count" or "Length" property
+            //      - Does not inherit from IDictionary<> as that is handled earlier in the binder
+            // For those types, generate special code to check for negative indices, otherwise just generate
+            // the call. Before we test for the above criteria explicitly, we will determine if the
+            // target is of a type known to be compatible. This is done to avoid the call to Module.ResolveMethod
+            // when possible.
             Type limitType = target.LimitType;
             if (limitType.IsArray || limitType == typeof(string) || limitType == typeof(StringBuilder))
             {
@@ -4289,15 +4300,6 @@ namespace System.Management.Automation.Language
 
             if (CanIndexFromEndWithNegativeIndex(target, getter, getterParams))
             {
-                // PowerShell supports negative indexing for types that meet the following criteria:
-                //      - Indexer method accepts one parameter that is typed as int
-                //      - The int parameter is not a type argument from a constructed generic type
-                //        (this is to exclude indexers for types that could use a negative index as
-                //        a valid key like System.Linq.ILookup)
-                //      - Declares a "Count" or "Length" property
-                //      - Does not inherit from IDictionary<> as that is handled earlier in the binder
-                // For those types, generate special code to check for negative indices, otherwise just generate
-                // the call.
                 if (lengthProperty == null)
                 {
                     // Count is declared by most supported types, Length will catch some edge cases like strings.
--- a/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
+++ b/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
@@ -4053,7 +4053,8 @@ namespace System.Management.Automation.Language
             // Get the base method definition of the indexer to determine if the int
             // parameter is a generic type parameter. Module.ResolveMethod is used
             // because the indexer could be a method from a constructed generic type.
-            return indexer.Module.ResolveMethod(indexer.MetadataToken).ContainsGenericParameters;
+            MethodBase baseMethod = indexer.Module.ResolveMethod(indexer.MetadataToken);
+            return !baseMethod.GetParameters()[0].ParameterType.IsGenericParameter;
         }
 
         private DynamicMetaObject IndexWithNegativeChecks(
--- a/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
+++ b/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
@@ -4056,6 +4056,12 @@ namespace System.Management.Automation.Language
             // the call. Before we test for the above criteria explicitly, we will determine if the
             // target is of a type known to be compatible. This is done to avoid the call to Module.ResolveMethod
             // when possible.
+
+            if (getterParams.Length != 1 || getterParams[0].ParameterType != typeof(int))
+            {
+                return false;
+            }
+
             Type limitType = target.LimitType;
             if (limitType.IsArray || limitType == typeof(string) || limitType == typeof(StringBuilder))
             {
@@ -4076,11 +4082,6 @@ namespace System.Management.Automation.Language
             if (limitType.GetInterfaces().Any(i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IList<>)))
             {
                 return true;
-            }
-
-            if (getterParams.Length != 1 || getterParams[0].ParameterType != typeof(int))
-            {
-                return false;
             }
 
             // Get the base method definition of the indexer to determine if the int
--- a/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
+++ b/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
@@ -3946,6 +3946,11 @@ namespace System.Management.Automation.Language
                 return InvokeIndexer(target, indexes, errorSuggestion, defaultMember.MemberName).WriteToDebugLog(this);
             }
 
+            if (typeof(ITuple).IsAssignableFrom(target.LimitType))
+            {
+                return InvokeIndexer(target, indexes, errorSuggestion, "Item").WriteToDebugLog(this);
+            }
+
             return errorSuggestion ?? CannotIndexTarget(target, indexes).WriteToDebugLog(this);
         }
 
@@ -4046,7 +4051,12 @@ namespace System.Management.Automation.Language
             }
 
             // target implements IList<T>?
-            return limitType.GetInterfaces().Any(i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IList<>));
+            if (limitType.GetInterfaces().Any(i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IList<>)))
+            {
+                return true;
+            }
+
+            return typeof(ITuple).IsAssignableFrom(limitType);
         }
 
         private DynamicMetaObject IndexWithNegativeChecks(
@@ -4249,6 +4259,12 @@ namespace System.Management.Automation.Language
                 // the call.
                 PropertyInfo lengthProperty = target.LimitType.GetProperty("Count") ??
                                               target.LimitType.GetProperty("Length"); // for string
+
+                // Default ITuple implementations explicitly implement ITuple.Length.
+                if (lengthProperty == null && typeof(ITuple).IsAssignableFrom(target.LimitType))
+                {
+                    lengthProperty = typeof(ITuple).GetProperty(nameof(ITuple.Length));
+                }
 
                 if (lengthProperty != null)
                 {
--- a/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
+++ b/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
@@ -3941,6 +3941,9 @@ namespace System.Management.Automation.Language
                     }
                 }
 
+                // If the type explicitly implements an indexer specified by an interface
+                // then the DefaultMemberAttribute will not carry over to the implementation.
+                // This check will catch those cases.
                 if (defaultMember == null)
                 {
                     defaultMember = i.GetCustomAttribute<DefaultMemberAttribute>(inherit: true);
--- a/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
+++ b/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
@@ -3946,7 +3946,7 @@ namespace System.Management.Automation.Language
                     defaultMember = i.GetCustomAttribute<DefaultMemberAttribute>(inherit: true);
                     if (defaultMember != null)
                     {
-                        lengthProperty = i.GetProperty("Length") ?? i.GetProperty("Count");
+                        lengthProperty = i.GetProperty("Count") ?? i.GetProperty("Length");
                     }
                 }
             }
--- a/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
+++ b/src/System.Management.Automation/engine/runtime/Binding/Binders.cs
@@ -4267,18 +4267,13 @@ namespace System.Management.Automation.Language
 
             if (CanIndexFromEndWithNegativeIndex(target, getter, getterParams))
             {
-                // PowerShell supports negative indexing for for types that meet the following criteria:
-                //
-                // - Indexer method accepts one parameter that is typed as int
-                //
-                // - The int parameter is not a type argument from a constructed generic type (this is
-                //   to exclude indexers for types that could use a negative index as a valid key
-                //   like System.Linq.ILookup)
-                //
-                // - Declares a "Count" or "Length" property
-                //
-                // - Does not inherit from IDictionary<> as that is handled earlier in the binder
-                //
+                // PowerShell supports negative indexing for types that meet the following criteria:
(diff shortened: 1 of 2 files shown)


src/System.Management.Automation/engine/runtime/Binding/Binders.cs
```diff
@@ -3946,6 +3946,11 @@ public override DynamicMetaObject FallbackGetIndex(DynamicMetaObject target, Dyn
                 return InvokeIndexer(target, indexes, errorSuggestion, defaultMember.MemberName).WriteToDebugLog(this);
             }
 
+            if (typeof(ITuple).IsAssignableFrom(target.LimitType))
```
lzybkr: This doesn't feel like the right fix. `ITuple` specifies a `DefaultMemberAttribute`, but we don't find it above because `GetCustomAttributes` is only checking classes, not implemented interfaces.

The correct fix is to search interfaces for `DefaultMemberAttribute` - there are other interfaces that specify `DefaultMemberAttribute`, and some of those are not currently handled correctly:

```
System.Collections.IDictionary
System.Collections.IList
System.Collections.Generic.IDictionary`2
System.Collections.Generic.IList`1
System.Collections.Generic.IReadOnlyDictionary`2
System.Collections.Generic.IReadOnlyList`1
System.Runtime.CompilerServices.ITuple
System.Runtime.CompilerServices.IRuntimeVariables
System.Collections.Specialized.IOrderedDictionary
System.Linq.ILookup`2
System.Data.IColumnMappingCollection
System.Data.IDataParameterCollection
System.Data.IDataRecord
System.Data.ITableMappingCollection
System.ComponentModel.IDataErrorInfo
Newtonsoft.Json.Linq.IJEnumerable`1
```

SeeminglyScience: @lzybkr Good call! I had checked that none of the default `ITuple` implementations had that attribute but didn't think to check the interface itself.

I've updated the change to look for `DefaultMemberAttribute` in the loop that looks for `IDictionary<,>`.  I tested the update with a custom `ILookup<,>` with an explicitly implemented indexer, that didn't work previously but does now.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
```diff
@@ -4046,7 +4056,12 @@ internal static bool CanIndexFromEndWithNegativeIndex(DynamicMetaObject target)
             }
 
             // target implements IList<T>?
-            return limitType.GetInterfaces().Any(i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IList<>));
+            if (limitType.GetInterfaces().Any(i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IList<>)))
+            {
+                return true;
+            }
+
+            return typeof(ITuple).IsAssignableFrom(limitType);
```
lzybkr: I wonder if this whole method shouldn't just check for a `Count` or `Length` property.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
iSazonov: Do now we support all these interfaces? If so, it's good to have a comment in the code and tests (or even extend the tests)
Also we need to document this.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
SeeminglyScience: I think that would be a breaking change.  For example, if a `Dictionary<,>` could make it that far into the binder (I know it can't, but custom code might) then `-1` could be a valid key.

I did a quick scan of the types visible by default in RC1, looking for types that:

1. Are decorated with `DefaultMemberAttribute`
1. Do not implement `ICollection` or `ICollection<>` (as the `Count` property coming from these make it likely enough to work as expected)
1. Do not implement `IDictionary<,>` (as they'll be handled earlier in the binder)

There's a couple types that jump out at a glance as *possibly* an issue, depending on how imperative it is to preserve that behavior. 

Here's the list in full, a couple obvious ones that jump out are `System.Linq.ILookup<,>`, `ReadOnlyPSMemberInfoCollection<>`, and some CIM related classes like `CimMethodParameterCollection`. 

```
System.String
System.ReadOnlySpan`1[T]
System.Span`1[T]
System.Numerics.Vector`1[T]
System.Text.StringBuilder
System.Runtime.CompilerServices.ITuple
System.Management.Automation.PSEventArgsCollection
System.Management.Automation.ReadOnlyPSMemberInfoCollection`1[T]
System.Management.Automation.Runspaces.InitialSessionStateEntryCollection`1[T]
System.Runtime.CompilerServices.IRuntimeVariables
Microsoft.Management.Infrastructure.CimMethodParametersCollection
Microsoft.Management.Infrastructure.Generic.CimKeyedCollection`1[T]
Microsoft.Management.Infrastructure.Generic.CimReadOnlyKeyedCollection`1[T]
System.Xml.XmlCDataSection
System.Xml.XmlCharacterData
System.Xml.XmlComment
System.Xml.XmlNodeList
System.Xml.XmlSignificantWhitespace
System.Xml.XmlText
System.Xml.XmlWhitespace
System.Xml.Serialization.XmlMembersMapping
System.Xml.Schema.XmlSchemaObjectTable
System.Collections.Specialized.StringDictionary
System.Linq.ILookup`2[TKey,TElement]
System.Linq.Lookup`2[TKey,TElement]
System.Data.SqlTypes.SqlChars
System.Data.SqlTypes.SqlBinary
System.Data.SqlTypes.SqlBytes
System.ComponentModel.MaskedTextProvider
Microsoft.CodeAnalysis.ChildSyntaxList
Microsoft.CodeAnalysis.SeparatedSyntaxList`1[TNode]
Microsoft.CodeAnalysis.SyntaxList`1[TNode]
Microsoft.CodeAnalysis.SyntaxNodeOrTokenList
Microsoft.CodeAnalysis.SyntaxTokenList
Microsoft.CodeAnalysis.SyntaxTriviaList
Microsoft.CodeAnalysis.Text.SourceText
Microsoft.CodeAnalysis.Text.TextLineCollection
System.Reflection.Metadata.GenericParameterHandleCollection
System.Reflection.Metadata.GenericParameterConstraintHandleCollection
Markdig.Helpers.StringSlice
```

While not incredibly likely to actually break existing code, perhaps it should be something the committee discusses?

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
SeeminglyScience: @iSazonov They were already supported.  What *wasn't* supported was indexing into classes that *explicitly* implement the indexer of an interface and define no other indexer.

All `ITuple` implementations explicitly implement the `ITuple` indexer, so while indexing tuples was technically supported, it wasn't working.

iSazonov: @SeeminglyScience Sorry, I did not ask the question correctly. Your explanation is useful and I just believe that it will make it easier to understand the binder if it's in the code.

SeeminglyScience: @iSazonov Ah yes absolutely, I'll add that tonight.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
lzybkr: The risk is that an indexer supports negative values as an index somehow, e.g. as the key in a dictionary.

Based on your list, `StringDictionary`, ``Lookup`2`` and ``ILookup`2`` do look like they'd support a negative value as the index.

Maybe the check should be - the indexer accepts a single `int` and is not a generic parameter. This avoids the problem with `StringDictionary` with the signature `string Item(string key) {get;set;}` and `ILookup[int,T]]` where even though the indexer takes an `int`, it was generic and therefore not always an int.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
SeeminglyScience: @iSazonov I've added a comment explaining purpose of the extra check in the loop.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
```diff
@@ -4027,26 +4040,21 @@ private DynamicMetaObject CannotIndexTarget(DynamicMetaObject target, DynamicMet
...
-                return true;
-            }
-
-            if (typeof(IList).IsAssignableFrom(limitType))
-            {
-                return true;
-            }
-
-            if (typeof(OrderedDictionary).IsAssignableFrom(limitType))
+            if (getterParams.Length != 1 || getterParams[0].ParameterType != typeof(int))
```
iSazonov: Can getterParams be null?

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
```diff
@@ -4242,6 +4249,7 @@ private DynamicMetaObject GetIndexMultiDimensionArray(DynamicMetaObject target,
                 }
             }
 
+
```
iSazonov: Please remove the extra line.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
```diff
@@ -4258,13 +4266,16 @@ private DynamicMetaObject GetIndexMultiDimensionArray(DynamicMetaObject target,
...
             {
                 // PowerShell supports negative indexing for some types (specifically, types implementing IList or IList<T>).
                 // For those types, generate special code to check for negative indices, otherwise just generate
                 // the call.
-                PropertyInfo lengthProperty = target.LimitType.GetProperty("Count") ??
-                                              target.LimitType.GetProperty("Length"); // for string
+                if (lengthProperty == null)
+                {
+                    lengthProperty = target.LimitType.GetProperty("Count") ??
+                                     target.LimitType.GetProperty("Length"); // for string
```
iSazonov: Comment should be on separate lines. Please remove the comment or add enhanced comment above.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
```diff
@@ -4258,7 +4248,7 @@ private DynamicMetaObject GetIndexMultiDimensionArray(DynamicMetaObject target,
                 }
             }
 
-            if (getterParams.Length == 1 && getterParams[0].ParameterType == typeof(int) && CanIndexFromEndWithNegativeIndex(target))
+            if (CanIndexFromEndWithNegativeIndex(target, getter, getterParams))
             {
                 // PowerShell supports negative indexing for some types (specifically, types implementing IList or IList<T>).
```
lzybkr: This comment should be updated.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
lzybkr: No, reflection will return empty arrays but not null.

lzybkr (approved):
LGTM, but it'd be good to have @daxian-dbw sign off before merging.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
SeeminglyScience: Removed the inline comment and added a more description comment above.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
SeeminglyScience: Good catch, I've updated the comment to fully explain the supported scenarios.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
```diff
@@ -4268,13 +4267,25 @@ private DynamicMetaObject GetIndexMultiDimensionArray(DynamicMetaObject target,
 
             if (CanIndexFromEndWithNegativeIndex(target, getter, getterParams))
             {
-                // PowerShell supports negative indexing for some types (specifically, types implementing IList or IList<T>).
+                // PowerShell supports negative indexing for for types that meet the following criteria:
```
iSazonov: Typo - for for.

iSazonov:
LGTM.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
```diff
@@ -4027,26 +4040,21 @@ private DynamicMetaObject CannotIndexTarget(DynamicMetaObject target, DynamicMet
                 bindingRestrictions);
         }
 
-        internal static bool CanIndexFromEndWithNegativeIndex(DynamicMetaObject target)
+        internal static bool CanIndexFromEndWithNegativeIndex(
+            DynamicMetaObject target,
```
daxian-dbw: Turns out this parameter is not used within the method.

I think we probably should keep the existing checks for the common scenarios, since `Module.ResolveMethod` is relatively expensive.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
```diff
@@ -4258,13 +4265,23 @@ private DynamicMetaObject GetIndexMultiDimensionArray(DynamicMetaObject target,
...
+            if (CanIndexFromEndWithNegativeIndex(target, getter, getterParams))
             {
-                // PowerShell supports negative indexing for some types (specifically, types implementing IList or IList<T>).
+                // PowerShell supports negative indexing for types that meet the following criteria:
+                //      - Indexer method accepts one parameter that is typed as int
+                //      - The int parameter is not a type argument from a constructed generic type
+                //        (this is to exclude indexers for types that could use a negative index as
+                //        a valid key like System.Linq.ILookup)
+                //      - Declares a "Count" or "Length" property
+                //      - Does not inherit from IDictionary<> as that is handled earlier in the binder
```
daxian-dbw: Great comments! I think they should be moved to the method `CanIndexFromEndWithNegativeIndex`  😄

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
```diff
@@ -3928,6 +3928,8 @@ public override DynamicMetaObject FallbackGetIndex(DynamicMetaObject target, Dyn
                 return GetIndexArray(target, indexes, errorSuggestion).WriteToDebugLog(this);
             }
 
+            var defaultMember = target.LimitType.GetCustomAttributes<DefaultMemberAttribute>(true).FirstOrDefault();
```
daxian-dbw: Can we change this to also use `GetCustomAttribute<DefaultMemberAttribute>(inherit: true)`?

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
```diff
@@ -3938,12 +3940,23 @@ public override DynamicMetaObject FallbackGetIndex(DynamicMetaObject target, Dyn
                         return result.WriteToDebugLog(this);
                     }
                 }
+
+                // If the type explicitly implements an indexer specified by an interface
+                // then the DefaultMemberAttribute will not carry over to the implementation.
+                // This check will catch those cases.
+                if (defaultMember == null)
+                {
+                    defaultMember = i.GetCustomAttribute<DefaultMemberAttribute>(inherit: true);
```
daxian-dbw: A minor comment: Attributes cannot be inherited from interfaces, so it should be `inherit: false`.

daxian-dbw (changes requested):

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
```diff
@@ -4027,26 +4040,21 @@ private DynamicMetaObject CannotIndexTarget(DynamicMetaObject target, DynamicMet
...
+                return false;
             }
 
-            // target implements IList<T>?
-            return limitType.GetInterfaces().Any(i => i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IList<>));
+            // Get the base method definition of the indexer to determine if the int
+            // parameter is a generic type parameter. Module.ResolveMethod is used
+            // because the indexer could be a method from a constructed generic type.
+            MethodBase baseMethod = indexer.Module.ResolveMethod(indexer.MetadataToken);
+            return !baseMethod.GetParameters()[0].ParameterType.IsGenericParameter;
```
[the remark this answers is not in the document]
SeeminglyScience: That tests return type as well.  The indexer for something like `IList<>` would also be excluded.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
SeeminglyScience: Hmm actually it looks like `DefaultMemberAttribute` can be declared multiple times, and this method will throw in that case.  I believe I need to change the other instance to match this one instead.

daxian-dbw: Oh, that sounds good. Thanks for clarifying.

src/System.Management.Automation/engine/runtime/Binding/Binders.cs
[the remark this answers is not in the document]
daxian-dbw: Good point. `ContainsGenericParameters` indicates if this method comes from a generic type.

daxian-dbw:
@SeeminglyScience I just pushed a commit to re-arrange the order of the checks in `CanIndexFromEndWithNegativeIndex` (parameter check is simpler), and also add `[Feature]` tag to run all the tests. Hope you don't mind!

daxian-dbw (approved):
LGTM

iSazonov:
@SeeminglyScience Thanks for the great enhancement! 
 Please update the PR description to facilitate documentation.

daxian-dbw:
I have updated the PR description. Will merge the PR.

Issue #6697, opened 2018-04-21
New Feature: Commands for string methods & operators on the pipeline


FriedrichWeinmann:
# Introduction
One of the things I have seen that always bugged me in PowerShell was the fact, that string operators cannot be used on a pipeline. A simple example:
```powershell
(2,3,1,4 | Sort-Object) -join "."
```
Not convenient, especially for interactive console use where you often need to go back and add the braces. The default string operators obviously aren't cmdlets and won't work on the pipeline - instead they are fast.

In one of my modules (`PSUtil`) I have functions that emulate this functionality:
```powershell
2,3,1,4 | sort | join "."
```
However I think this would make sense as part of the default command set, as virtually every user of PowerShell has to handle strings.

# Weighing the scales
## Advantages

 - More convenient to use for end user on the pipeline
 - Additional suitable features can be added through parameters (See implementation proposal below)

## Disadvantages

 - Clobber list of commands available by default with more commands.

# On Implementation

I am willing to implement all commands involved, their tests and documentation, all by myself, if this feature is deemed worthwhile.

# Proposed commands & features
## Add-String (Alias: 'wrap')

Does not have an operator equivalent. Command that allows you to easily add to a string on the pipeline.

### Example
```powershell
PS> 1..4 | wrap '"' '"'

"1"
"2"
"3"
"4"
```

### Parameters

 - InputString (`string[]` | Pipeline), the string(s) being added to
 - Before (`string`), the string to add before the input
 - Behind (`string`), the string to add behind the input
 - PadLeft (`char`), character to pad left with
 - PadRight (`char`), character to pad right with
 - PadWidth (`int`), up to how many characters the string should be padded with

## Format-String (Alias: 'format')

Equivalent to `-f`. Command that allows you to use the format operator on the pipeline.

### Example
```powershell
PS> 1..4 | format "{0:N2} - {1:D3}" -Count 2

1,00 - 002
3,00 - 004
```

### Parameters

 - InputObject (`object` | Pipeline), objects to be formatted
 - Format (`string`), the format definition
 - Count (`int`), the number of items to store up before formating them in bulk (optional, default 1)

## Get-Substring

Allows trimming and picking a substring from specified strings.

### Example

```powershell
PS> "abc def ghi" | substring -trim "abi"

c def gh

PS> "abc def ghi" | substring 2 4

c de
```

### Parameters

 - InputString (`string[]` | Pipeline), the strings to pick from
 - Trim (`string`), what characters to trim
[...]

PR #7660, opened 2018-08-29
Join-String cmdlet for creating text from pipeline input

powercode:
## PR Summary

A join-object cmdlet that joins pipeline input to text.

#6697 

As requested by @BrucePay.

Final syntax of the cmdlet:
```
Join-String [[-Property] <pspropertyexpression>] [[-Separator] <string>] [-OutputPrefix <string>] [-OutputSuffix <string>] [-UseCulture] [-InputObject <psobject>] [<CommonParameters>]

Join-String [[-Property] <pspropertyexpression>] [[-Separator] <string>] [-OutputPrefix <string>] [-OutputSuffix <string>] [-SingleQuote] [-UseCulture] [-InputObject <psobject>] [<CommonParameters>]

Join-String [[-Property] <pspropertyexpression>] [[-Separator] <string>] [-OutputPrefix <string>] [-OutputSuffix <string>] [-DoubleQuote] [-UseCulture] [-InputObject <psobject>] [<CommonParameters>]

Join-String [[-Property] <pspropertyexpression>] [[-Separator] <string>] [-OutputPrefix <string>] [-OutputSuffix <string>] [-FormatString <string>] [-UseCulture] [-InputObject <psobject>] [<CommonParameters>]
```
```PowerShell
1..3 | Join-String -OutputPrefix "A " -OutputSuffix " B" -Separator "," -SingleQuote | Should -BeExactly "A '1','2','3' B" 
gci | Join-String Name -Separator "," -DoubleQuote
gci | Join-String {$_.Name * 2} "; "
```

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**


iSazonov:
'Object' in the cmdlet name implies that we can join objects of any type - array + array, array + hash, hash + hash, strings and so on.

powercode:
It is intended to be the pipeline equivalent of the `-join` operator. @BrucePay  suggested the name `Join-Object` in the issue. `Convert` would imply for me that it converted each item to a string. This cmdlet join them together with a delimiter.

RichardSiddaway:
If its going to be the pipeline equivalent of -join wouldn't Join-String be a more descriptive name.  Shouldn't there also be a Split-? cmdlet to match the -split operator

powercode:
It depends. It can join objects, or the properties of objects. 

The output is a string, but it works with object input. Just like Group-Object and Sort-Object.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -5126,7 +5126,7 @@ private static bool IsInDscContext(ExpressionAst expression)
             return Ast.GetAncestorAst<ConfigurationDefinitionAst>(expression) != null;
         }
 
-        private static void CompleteMemberByInferredType(CompletionContext context, IEnumerable<PSTypeName> inferredTypes, List<CompletionResult> results, string memberName, Func<object, bool> filter, bool isStatic)
+        internal static void CompleteMemberByInferredType(TypeInferenceContext context, IEnumerable<PSTypeName> inferredTypes, List<CompletionResult> results, string memberName, Func<object, bool> filter, bool isStatic)
```
PaulHigin: Do we know that CompletionContext.Helper is the same thing has TypeInferenceContext.Helper?

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -6905,4 +6905,83 @@ public object VisitParenExpression(ParenExpressionAst parenExpressionAst)
             return parenExpressionAst.Pipeline.Accept(this);
         }
     }
+
+
+    /// <summary>
+    /// Completes with the property names of the InputObject.
+    /// </summary>
+    public class PropertyNameCompleter : IArgumentCompleter
```
PaulHigin: I think this should be internal access.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -6905,4 +6905,83 @@ public object VisitParenExpression(ParenExpressionAst parenExpressionAst)
...
+            IDictionary fakeBoundParameters)
+        {
+            var pipelineAst = commandAst.Parent as PipelineAst;
+            if (pipelineAst == null)
+            {
+                return null;
+            }
+
+            int i;
+            for (i = 0; i < pipelineAst.PipelineElements.Count; i++)
```
PaulHigin: Have you considered using `Ast.FindAll(...).FirstOrDefault()`?

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
```diff
@@ -121,14 +122,16 @@ public int GetHashCode(PSTypeName obj)
...
+    internal class TypeInferenceContext : IDisposable
     {
         public static readonly PSTypeName[] EmptyPSTypeNameArray = Utils.EmptyArray<PSTypeName>();
+        private readonly bool _ownsPowerShell;
         private readonly PowerShell _powerShell;
 
         public TypeInferenceContext()
         : this(PowerShell.Create(RunspaceMode.CurrentRunspace))
         {
+            _ownsPowerShell = true;
```
PaulHigin: This doesn't look right to me.  This class can own PowerShell/Runspace only if PowerShell is created with `Runspace.NewRunspace` mode.  Otherwise you are disposing a runspace created by someone else (and possibly the thread default runspace).

PaulHigin (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs
```diff
@@ -0,0 +1,211 @@
...
+{
+    /// <summary>
+    /// Group-Object implementation.
+    /// </summary>
+    [Cmdlet(VerbsCommon.Join, "Object", RemotingCapability = RemotingCapability.None, DefaultParameterSetName = "default")]
+    [OutputType(typeof(string))]
+    public class JoinObjectCommand : PSCmdlet
+    {
+        // ReSharper disable once CollectionNeverQueried.Local
+        private readonly List<PSObject> _inputObjects = new List<PSObject>(50);
```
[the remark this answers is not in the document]
powercode: It is cheaper to allocate a list that is a little larger than to reallocate and copy when out-grown.
But the size can always be argued. 

When I profile memory allocations, Array allocations on List.Resize is not uncommon, and when I see them, I try to find a sensible default size do get rid of the most common once.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs
```diff
@@ -0,0 +1,211 @@
...
+using System.Collections.Generic;
+using System.Management.Automation;
+using System.Management.Automation.Internal;
+using System.Management.Automation.Language;
+using System.Text;
+
+namespace Microsoft.PowerShell.Commands.Utility
+{
+    /// <summary>
+    /// Group-Object implementation.
```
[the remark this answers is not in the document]
powercode: oops :)

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
powercode: Only by reading the source. But it was me who refactored the helper into a shared class when I wrote the TypeInferenceContext class.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
powercode: I understand from the perspective of this CL, but in general, isn't it cool if we could get completion for PropertyNames for third party cmdlets? 
Others (like me :) ) can make use of that too.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
powercode: Yes. It's a balance. I try hard not to create extra objects, which Linq tends to do. But it sure make the code uglier and harder to read.

And to be honest, I haven't measured the allocations here. 
But it would mean the allocation of a Func delegate, an AstSearcher, a List for the results, and the enumerator. Granted, the last one is an List.Enumerator which is only on the stack.

I fairly often use Linq in higher level code, but when writing infrastructure like this, I try to go with the boring, ugly, but faster option.

src/System.Management.Automation/engine/parser/TypeInferenceVisitor.cs
powercode: Reverting the whole file.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs
```diff
@@ -0,0 +1,211 @@
...
+    /// </summary>
+    [Cmdlet(VerbsCommon.Join, "Object", RemotingCapability = RemotingCapability.None, DefaultParameterSetName = "default")]
+    [OutputType(typeof(string))]
+    public class JoinObjectCommand : PSCmdlet
```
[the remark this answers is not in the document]
powercode: yes

powercode:
@PaulHigin Thx for the review.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs
```diff
@@ -0,0 +1,211 @@
...
+                }
+                else
+                {
+                    builder.Append(val);
+                }
+            }
+
+            if (PropertyName == null)
+            {
+                if (_inputObjects.Count > 0)
```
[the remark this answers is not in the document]
powercode: Yes, much better!

test/powershell/Modules/Microsoft.PowerShell.Utility/Join-Object.Tests.ps1
```diff
@@ -0,0 +1,87 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+Describe "Join-Object" -Tags "CI" {
```
iSazonov: Please add new line.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs
```diff
@@ -0,0 +1,211 @@
...
+        // ReSharper disable once CollectionNeverQueried.Local
+        private readonly List<PSObject> _inputObjects = new List<PSObject>(Capacity);
+        private DynamicPropertyGetter _propGetter = new DynamicPropertyGetter();
+
+        /// <summary>
+        /// Gets or sets the property name or script block to use as the value to join.
+        /// </summary>
+        [Parameter(Position = 0)]
+        [ArgumentCompleter(typeof(PropertyNameCompleter))]
+        public object PropertyName { get; set; }
```
iSazonov: Can we use PSPropertyExpression type like in #6934?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs
```diff
@@ -0,0 +1,211 @@
...
+    internal class JoinItemCompleter : IArgumentCompleter
+    {
+        public IEnumerable<CompletionResult> CompleteArgument(
+            string commandName,
+            string parameterName,
+            string wordToComplete,
+            CommandAst commandAst,
+            IDictionary fakeBoundParameters)
+        {
+            var res = new List<CompletionResult>(10);
```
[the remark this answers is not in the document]
powercode: Then you end up with races if used from multiple runspaces.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs
powercode: PropertyExpression has some extra semantics, handling hashtables like `@{n="x";ex={2,3,4}}`.

I looked at it, but was not convinced it was a fit. But there's definitely an overlap. Input here is most welcome.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs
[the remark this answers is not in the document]
iSazonov: The list looks as a const for current session. Do I skip something?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs
```diff
@@ -0,0 +1,211 @@
...
+        public string NewLineText
+        {
+            get
+            {
+                switch (Environment.NewLine)
+                {
+                    case "\r": return "`r";
+                    case "\n": return "`n";
+                    case "\r\n": return "`r`n";
+                    default: return Environment.NewLine.Replace("\r", "`r").Replace("\n", "`n");
```
iSazonov: Seems the switch is superfluous. We could leave only
``'c#
return Environment.NewLine.Replace("\r", "`r").Replace("\n", "`n"); 
```

Comment (powercode) [reply src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs]: Yes it could. I wrote it the way I did to reduce allocated objects. May not be worth it. 
The string literals will be loaded from metadata and interned. There will only exist one instance of them. The default line will create two new strings each time.

File: src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs (hunk 11 above)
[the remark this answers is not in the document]
Comment (powercode) [reply src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs]: Yes, it is captured by the local function `AddMatching`.
So the result is depending on `wordToComplete`.

**Review by powercode — COMMENTED** [review]
A general question is how we should handle `null` and `AutomationNull.Value` in the input?

File: src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs (hunk 12 above)
Comment (iSazonov) [reply src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs]: Environment.NewLine is static. So I think we should use the usual method
```c#
public static string NewLineText
{
    get
    {
#if UNIX
        return "`n"
#else
        return "`r`n"
#endif
    }
}
```

powercode: Good point!

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs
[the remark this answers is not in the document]
iSazonov: I see now. :-) It is Friday. I can not catch up with my idea :-)
I am looking at 'new CompletionResult()'. If they isn't modifed later we could make it static and only reference in target List?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs
powercode: But I think MacOS uses ``` `r```

iSazonov:
`AutomationNull.Value` is ignored by default in pipes 
https://github.com/PowerShell/PowerShell/blob/4fc784e840b8df00d2552045c6ed52d027659d2b/src/System.Management.Automation/engine/Pipe.cs#L380

Ex.:
```powershell
$null,$null | % { 1 }
1
1

[System.Management.Automation.Internal.AutomationNull]::Value,[System.Management.Automation.Internal.AutomationNull]::Value | % { 1 }
```

PaulHigin (approved):
LGTM

vexx32:
I'd be inclined to call it Join-String instead, since nothing of the original objects are preserved (counter to things like Group-Object or Select-Object)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// Gets or sets text to include after the joined input text.
+        /// </summary>
+        [Parameter]
+        public string Suffix { get; set; }
+
+        /// <summary>
+        /// Gets or sets if the output items should we wrapped in single quotes.
+        /// </summary>
+        [Parameter(ParameterSetName = "Quote")]
+        public SwitchParameter Quote { get; set; }
```
iSazonov: SingleQuote supplements DoubleQuote.

Another thought is to make one parameter (which replace both Quote and DoubleQuote parameters) with allowed values SingleQuote and DoubleQuote. Seems better UX. This also reduce account of parameter sets.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+        /// Gets or sets if the output items should we wrapped in double quotes.
+        /// </summary>
+        [Parameter(ParameterSetName = "DoubleQuote")]
+        public SwitchParameter DoubleQuote { get; set; }
+
+        /// <summary>
+        /// Gets or sets the input object to join into text.
+        /// </summary>
+        [Parameter(ValueFromPipeline = true, Mandatory = true)]
+        public PSObject InputObject { get; set; }
```
iSazonov: Why only PSObject? Maybe object?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+
+        /// <inheritdoc />
+        protected override void ProcessRecord()
+        {
+            if (InputObject != null && InputObject != AutomationNull.Value)
```
iSazonov: Make sense remove `InputObject != AutomationNull.Value`?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+                _inputObjects.Add(InputObject);
+            }
+        }
+
+        /// <inheritdoc />
+        protected override void EndProcessing()
+        {
+            if (_inputObjects.Count == 0)
+            {
+                return;
```
iSazonov: Should we return null or empty string?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+            }
+
+            if (Property == null)
+            {
+                AppendValue(LanguagePrimitives.ConvertTo<string>(_inputObjects[0]));
+
+                for (var index = 1; index < _inputObjects.Count; index++)
+                {
+                    builder.Append(Delimiter);
+                    AppendValue(LanguagePrimitives.ConvertTo<string>(_inputObjects[index]));
```
iSazonov: we do not need to collect input objects - we need only result strings. So we should move the ConvertTo to the ProcessRecord() method.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+
+                for (var index = 1; index < _inputObjects.Count; index++)
+                {
+                    builder.Append(Delimiter);
+                    AppendValue(GetScriptBlockResultString(_inputObjects[index]));
+                }
+            }
+            else
+            {
+                throw new ArgumentException();
```
iSazonov: We should return useful error message. Or yet use PSPropertyExpression.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+            AddMatching("' '", "Space", "' '  - Space");
+            return res;
+        }
+
+        public string NewLineText
+        {
+            get
+            {
+#if UNIX
+                return Platform.IsMacOS ? "`r" : "`n";
```
iSazonov: It is only for very old MacOs versions. Please remove.
CoreFX doesn't use this https://source.dot.net/#System.Private.CoreLib/src/System/Environment.cs,261
For reference https://en.wikipedia.org/wiki/Newline

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
iSazonov: I suggest to convert to internal for fast merging the PR and open a tracking issue to wait PowerShell Committee approval.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -6905,4 +6906,82 @@ public object VisitParenExpression(ParenExpressionAst parenExpressionAst)
...
+
+        IEnumerable<CompletionResult> IArgumentCompleter.CompleteArgument(
+            string commandName,
+            string parameterName,
+            string wordToComplete,
+            CommandAst commandAst,
+            IDictionary fakeBoundParameters)
+        {
+            var pipelineAst = commandAst.Parent as PipelineAst;
+            if (pipelineAst == null)
```
iSazonov: Why not C# 7.0 syntax?

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -6905,4 +6906,82 @@ public object VisitParenExpression(ParenExpressionAst parenExpressionAst)
...
+            {
+                var parameterAst = (CommandParameterAst)commandAst.Find(ast => ast is CommandParameterAst cpa && cpa.ParameterName == "PropertyName", false);
+                var pseudoBinding = new PseudoParameterBinder().DoPseudoParameterBinding(commandAst, null, parameterAst, PseudoParameterBinder.BindingType.ParameterCompletion);
+                if (!pseudoBinding.BoundArguments.TryGetValue(_parameterNameOfInput, out var pair) || !pair.ArgumentSpecified)
+                {
+                    return null;
+                }
+
+                var astPair = pair as AstPair;
+                if (astPair?.Argument == null)
```
iSazonov: Why not C# 7.0 syntax? Looks more readable.

test/powershell/Modules/Microsoft.PowerShell.Utility/Join-String.Tests.ps1
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    It "Should be called using the InputObject without error with no other switches" {
+        { Join-String -InputObject $testObject } | Should -Not -Throw
+    }
+
+    It "Should return a single string" {
+        $actual = $testObject | Join-String
+
+        $actual.Count       | Should -Be 1
```
iSazonov: Extra spaces.

test/powershell/Modules/Microsoft.PowerShell.Utility/Join-String.Tests.ps1
```diff
@@ -0,0 +1,99 @@
...
+	It "Should tabcomplete InputObject properties" {
+		$cmd = '[io.fileinfo]::new("c:\temp") | Join-String -Property '
+		$res = tabexpansion2 $cmd $cmd.length
+		$completionTexts = $res.CompletionMatches.CompletionText
+		$Propertys = [io.fileinfo]::new($PSScriptRoot).psobject.properties.Name
+		foreach($n in $Propertys) {
+			$n -in $completionTexts | Should -BeTrue
+		}
+	}
+
```
iSazonov: Please remove the extra line.

test/powershell/engine/Basic/DefaultCommands.Tests.ps1
```diff
@@ -319,6 +319,7 @@ Describe "Verify approved aliases list" -Tags "CI" {
 "Cmdlet", "Invoke-WebRequest",                                  ,                     $($FullCLR -or $CoreWindows -or $CoreUnix)
 "Cmdlet", "Invoke-WmiMethod",                                   ,                     $($FullCLR                               )
 "Cmdlet", "Invoke-WSManAction",                                 ,                     $($FullCLR -or $CoreWindows              )
+"Cmdlet", "Join-String",                                        ,                     $(             $CoreWindows -or $CoreUnix)
 "Cmdlet", "Join-Path",                                          ,                     $($FullCLR -or $CoreWindows -or $CoreUnix)
```
iSazonov: Please swap the two lines. Alphabetical order.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+        [ArgumentCompleter(typeof(PropertyNameCompleter))]
+        public object Property { get; set; }
+
+        /// <summary>
+        /// Gets or sets the delimiter to join the output with.
+        /// </summary>
+        [Parameter(Position = 1)]
+        [ArgumentCompleter(typeof(JoinItemCompleter))]
+        [AllowEmptyString]
+        public string Delimiter { get; set; }
```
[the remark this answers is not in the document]
powercode: GetVariableValue is not available in a static context.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+
+            var quoteChar = Quote ? '\'' : DoubleQuote ? '"' : char.MinValue;
+
+            const int DefaultOutputStringCapacity = 256;
+            var builder = new StringBuilder(DefaultOutputStringCapacity);
+            builder.Append(Prefix);
+
+            if (Delimiter == null)
+            {
+                Delimiter = LanguagePrimitives.ConvertTo<string>(GetVariableValue("OFS"));
```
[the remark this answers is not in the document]
powercode: No, that code makes use of the initialization that the engine sets up after the cmdlet objects constructor has been run.

powercode:
What about aliases?

Join has the alias 'j' and String seems to be 's'

Should there be a 'js' alias?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+using System.Management.Automation.Internal;
+using System.Management.Automation.Language;
+using System.Text;
+
+namespace Microsoft.PowerShell.Commands.Utility
+{
+    /// <summary>
+    /// Join-Object implementation.
+    /// </summary>
+    [Cmdlet(VerbsCommon.Join, "String", RemotingCapability = RemotingCapability.None, DefaultParameterSetName = "default")]
```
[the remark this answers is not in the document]
powercode: I don't know much about this. Who does documentation?

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
powercode: It is now internal.

BrucePay (changes requested):
This code really should be using `PSPropertyExpression`.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-object.cs
iSazonov: No, only very old MacOs versions. See my comment below.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
```diff
@@ -6905,4 +6906,80 @@ public object VisitParenExpression(ParenExpressionAst parenExpressionAst)
             return parenExpressionAst.Pipeline.Accept(this);
         }
     }
+
+    /// <summary>
+    /// Completes with the property names of the InputObject.
+    /// </summary>
+    internal class PropertyNameCompleter : IArgumentCompleter
```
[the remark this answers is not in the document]
iSazonov: Is the cmdlet already approved by PowerShell Committee? If so we can merge "as is" and consider this api later.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,211 @@
...
+        {
+            if (_inputObjects.Count == 0)
+            {
+                return;
+            }
+
+            var quoteChar = Quote ? '\'' : DoubleQuote ? '"' : char.MinValue;
+
+            const int DefaultOutputStringCapacity = 256;
+            var builder = new StringBuilder(DefaultOutputStringCapacity);
```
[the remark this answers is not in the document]
iSazonov: We should do the join in ProcessRecord() too and here only add suffix (prefix in BeginProcessing()).

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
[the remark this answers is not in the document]
powercode: I don't think the developer who wrote it excels at documentation in English. Rumor has it that he was willing to write it in the language of the Gods and the Heroes, though.

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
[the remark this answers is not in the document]
powercode: If you are gonna discuss that, please consider making AstTypeInference public too. I need it from EditorServices.

src/Modules/Unix/Microsoft.PowerShell.Utility/Microsoft.PowerShell.Utility.psd1
```diff
@@ -16,7 +16,7 @@ CmdletsToExport= "Format-List", "Format-Custom", "Format-Table", "Format-Wide",
     "Get-Unique", "Export-PSSession", "Import-PSSession", "Import-Alias", "Import-LocalizedData",
     "Select-String", "Measure-Object", "New-Alias", "New-TimeSpan", "Read-Host", "Set-Alias", "Set-Date",
     "Start-Sleep", "Tee-Object", "Measure-Command", "Update-TypeData", "Update-FormatData",
-    "Remove-TypeData", "Get-TypeData", "Write-Host", "Write-Progress", "New-Object", "Select-Object",
+    "Remove-TypeData", "Get-TypeData", "Write-Host", "Write-Progress", "New-Object", "Select-Object", "Join-String",
```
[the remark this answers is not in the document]
powercode: I got lazy when I renamed it from Join-Object. Well, laz*ier* :)

test/powershell/Modules/Microsoft.PowerShell.Utility/Join-String.Tests.ps1
```diff
@@ -0,0 +1,98 @@
...
+    It "Should return a single string" {
+        $actual = $testObject | Join-String
+
+        $actual.Count | Should -Be 1
+        $actual | Should -BeOfType System.String
+    }
+
+    It "Should join property values with default delimiter" {
+        $expected = $testObject.Name -join $ofs
```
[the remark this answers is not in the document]
powercode: Does it make sense to allow wildcards, but not have an array?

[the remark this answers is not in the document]
powercode: I don't get the usecase for the wildcard here. Can you give an example when you think it would be valuable?

test/powershell/Modules/Microsoft.PowerShell.Utility/Join-String.Tests.ps1
```diff
@@ -0,0 +1,98 @@
...
+        $actual | Should -BeExactly $expected
+    }
+
+    It "Should join property values Quoted" {
+        $expected = ($testObject.Name).Foreach{"'$_'"} -join "; "
+        $actual = $testObject | Join-String -Property Name -Delimiter "; " -Quote
+        $actual | Should -BeExactly $expected
+    }
+
+    It "Should join property values DoubleQuoted" {
```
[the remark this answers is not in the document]
powercode: I like that idea!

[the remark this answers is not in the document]
powercode: Do we need an IFormatProvider/Culture too?

src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs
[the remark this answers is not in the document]
powercode: I had it public in my initial commit, as I believed it to be generally useful. Changed it to internal after @PaulHigin feedback.

powercode:
I have addressed most of the feedback.

What I know is remaining is the following:

- Should Property be an Array? If so, why?
- Should Wildcards be allowed? When is this useful?
- Do we need a `-Culture` property for use with format, or should we just use the current culture?
- Should we have an alias for the cmdlet?
- Docs in other languages than Swedish.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
powercode: I don't. I have used this cmdlet in a slightly different but very similar implementation, and the most common use case, by far, is to create strings for use in code.
Having `DoubleQuote` and `SingleQuote` is so much nicer than to have to specify a `Format`. 

I can for sure see the value of Format, but it is not as nice from an interactive perspective as the switches.

iSazonov (commented):
> Should Property be an Array? If so, why?

One scenario I see instantly - very flexible convert to csv.

> Do we need a -Culture property for use with format, or should we just use the current culture?

In Get-Date we have Format parameter without `Cuture`.

> Should we have an alias for the cmdlet?

I believe no - we should delegate this users.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+            if (InputObject != null && InputObject != AutomationNull.Value)
+            {
+                var inputValue = Property == null
+                                    ? InputObject
+                                    : Property.GetValues(InputObject, false, true).FirstOrDefault()?.Result;
+                var stringValue = LanguagePrimitives.ConvertTo<string>(inputValue);
+
+                if (_firstInputObject)
+                {
+                    _firstInputObject = false;
```
iSazonov: Now we do not need this

powercode: Well, we still need a way to determine if we should skip the first delimiter, don't we?

iSazonov: Yes, I forget about the separator.

powercode:
None of the test failures are related to my commits.

anmenaga:
I restarted failed test runs; we'll see how they go...

anmenaga:
@BrucePay can you please take a look again at this PR? Thanks.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+
+        /// <summary>
+        /// Gets or sets if the output items should we wrapped in double quotes.
+        /// </summary>
+        [Parameter(ParameterSetName = "DoubleQuote")]
+        public SwitchParameter DoubleQuote { get; set; }
+
+        /// <summary>
+        /// Gets or sets a format string that is applied to each input object.
+        /// </summary>
```
iSazonov: I'd expand the comment with what is the format and/or add a link on docs.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+namespace Microsoft.PowerShell.Commands.Utility
+{
+    /// <summary>
+    /// Join-Object implementation.
+    /// </summary>
+    [Cmdlet(VerbsCommon.Join, "String", RemotingCapability = RemotingCapability.None, DefaultParameterSetName = "default")]
+    [OutputType(typeof(string))]
+    public sealed class JoinStringCommand : PSCmdlet
+    {
+        /// <summary>A bigger default to not get re-allocations in common use cases.</summary>
```
iSazonov: Please use file pattern:
```c#
/// <Summary>
/// A bigger default to not get re-allocations in common use cases.
/// </Summary> 
```

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+                else
+                {
+                    _outputBuilder.Append(Separator);
+                }
+
+                if (_quoteChar != char.MinValue)
+                {
+                    _outputBuilder.Append(_quoteChar);
+                    _outputBuilder.Append(stringValue);
+                    _outputBuilder.Append(_quoteChar);
```
iSazonov: We could use
```c#
_outputBuilder.Append(_quoteChar).Append(stringValue).(_quoteChar); 
```

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
```diff
@@ -0,0 +1,210 @@
...
+                }
+                else if (string.IsNullOrEmpty(FormatString))
+                {
+                    _outputBuilder.Append(stringValue);
+                }
+                else
+                {
+                    _outputBuilder.AppendFormat(CultureInfo.CurrentCulture, FormatString, stringValue);
```
iSazonov: I'd want to get confirmation that this should be `CurrentCulture` not `InvariantCulture`.
And maybe discuss `-FormatCulture` parameter.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
iSazonov: Please address the comment.

test/powershell/Modules/Microsoft.PowerShell.Utility/Join-String.Tests.ps1
```diff
@@ -0,0 +1,118 @@
...
+
+Describe "Join-String" -Tags "CI" {
+
+    BeforeAll {
+        $testObject = Get-ChildItem
+    }
+
+    It "Should be called using an object as piped without error with no switches" {
+        {$testObject | Join-String } | Should -Not -Throw
+    }
```
iSazonov: We could remove the test - any test below do the test.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
powercode: Same here. Now sure what is right here.

mklement0: PowerShell consistently uses the invariant culture in to/from-string conversions, so it should probably be `InvariantCulture`, at least by default, perhaps with an optional  `-UseCulture` switch (analogous to the `*-Csv` cmdlets).

BrucePay (approved):
LGTM

anmenaga:
@powercode Looks like this PR still needs a couple of minor updates that are mentioned in @iSazonov's review; especially InvariantCulture issue.  Otherwise it seems close to being done.

JamesWTruher (approved):

powercode:
I just realized that when joining dates, I always get the "US", Invariant formatting, which is literally never the output a Swede would want. 

To get the `UseCulture` switch to have effect, I need to do some minor changes in how LanguagePrimitives makes the conversion to strings, and provide an overload that accepts an IFormatProvider.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/join-string.cs
iSazonov:
```powershell
 gcm -ParameterName Culture

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Cmdlet          Compare-Object                                     3.1.0.0    Microsoft.PowerShell.Utility
Cmdlet          Group-Object                                       3.1.0.0    Microsoft.PowerShell.Utility
Cmdlet          New-PSSessionOption                                6.0.0.0    Microsoft.PowerShell.Core
Cmdlet          Sort-Object                                        3.1.0.0    Microsoft.PowerShell.Utility
```

We already have `-Culture` parameter. Make sense re-use the name?

mklement0: @iSazonov: 

I think `-UseCulture` is the correct name, after all:

* The `-Culture <string>` parameters you reference:
    * require an _argument_, namely a _given_ culture
    * are used with cmdlets that are current-culture-_sensitive_ by default, to allow opt-in to a _different_ culture.

* The `-UseCulture` switch, by contrast:
    * is used with cmdlets that are culture-_invariant_ by default, to allow opt-in to the _current_ culture (currently applies only to the `*-Csv` cmdlets).

The latter is what is being implemented here.

iSazonov: @mklement0 Thanks! We need discuss Select-String  too.

daxian-dbw (approved):

daxian-dbw:
The last commit (`"Make TryConvertTo<string> respect the passed formatting provider."`) looks good to me.

iSazonov:
What is current policy: don't merge until doc issue is created?

Issue #3572, opened 2017-04-15
Tab-completion for environmental variable on Linux returns non-existent env. variable


alexandair:
Pressing TAB after `$env:P` returns `env:PATHEXT` that doesn't exist on PowerShell Core on Linux on my machine.

Steps to reproduce
------------------
PS /home/aleksandar> $env:P<tab>

Expected behavior
-----------------
```
PS /home/aleksandar> $env:P<tab>
PATH          PSModulePath  PWD
```

Actual behavior
---------------
```
PS /home/aleksandar> $env:P<tab>
PATH          PSModulePath  PWD           env:PATHEXT
```

Environment data
----------------

```powershell
PS /home/aleksandar> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      6.0.0-alpha
PSEdition                      Core
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   3.0.0.0
GitCommitId                    v6.0.0-alpha.18
CLRVersion
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1

```

chuanjiao10:
$env:pathext or env:pathext return
.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.CPL
no this environmental on linux.

yes! Deleting the variable PATHEXT on linux is better than clearing it,then just press the tab once

iSazonov:
I believe we use `$env:pathext` to find native commands.
Perhaps we would enhance that on Unix.

/cc @mirichmo @andschwa Could you please comment?

lzybkr:
PATHEXT is used to find native commands, but shouldn't be necessary on non-Windows systems.

I believe it's added to a cache of all special variables [here](https://github.com/PowerShell/PowerShell/blob/06020f34e556171b5ebd88155c3ff14e2cc1f5d8/src/System.Management.Automation/engine/CommandCompletion/CompletionCompleters.cs#L4853).

So I think the right fix is probably to add `#if !UNIX` around the "special" variables that are not used on Unix.

iSazonov:
It was removed in #7697 but we continue discussing in #7755

PR #7697, opened 2018-09-03
Do not add PATHEXT environment variable on Unix

iSazonov:
## PR Summary

Address #3572.

On Windows we use file extensions to detect executables. Windows define a list of the extensions in PATHEXT environment variable by default. If the variable is absent Powershell added `env:PATHEXT` special variable on _all_ platforms. A problem is that on Unix the PATHEXT is not defined by default and is not needed. The fix exclude creating PATHEXT on Unix.

It is a breaking change in grace area - no PowerShell scripts should depend on PATHEXT on Unix.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/NativeCommandProcessor.cs
+++ b/src/System.Management.Automation/engine/NativeCommandProcessor.cs
@@ -1301,7 +1301,7 @@ namespace System.Management.Automation
 
             var pathext = Environment.GetEnvironmentVariable("PATHEXT");
             string[] extensionList;
-            if (String.IsNullOrEmpty(pathext))
+            if (string.IsNullOrEmpty(pathext))
             {
 #if UNIX
                 return false;
@@ -1316,7 +1316,7 @@ namespace System.Management.Automation
 
             foreach (string extension in extensionList)
             {
-                if (String.Equals(extension, myExtension, StringComparison.OrdinalIgnoreCase))
+                if (string.Equals(extension, myExtension, StringComparison.OrdinalIgnoreCase))
                 {
                     return true;
                 }
--- a/src/System.Management.Automation/engine/NativeCommandProcessor.cs
+++ b/src/System.Management.Automation/engine/NativeCommandProcessor.cs
@@ -1293,9 +1293,7 @@ namespace System.Management.Automation
             // Now check the extension and see if it's one of the ones in pathext
             string myExtension = System.IO.Path.GetExtension(path);
 
-            string pathext = (string)LanguagePrimitives.ConvertTo(
-                this.Command.Context.GetVariableValue(SpecialVariables.PathExtVarPath),
-                typeof(string), CultureInfo.InvariantCulture);
+            var pathext = Environment.GetEnvironmentVariable("PATHEXT");
             string[] extensionList;
             if (String.IsNullOrEmpty(pathext))
             {
--- a/src/System.Management.Automation/engine/NativeCommandProcessor.cs
+++ b/src/System.Management.Automation/engine/NativeCommandProcessor.cs
@@ -1290,12 +1290,8 @@ namespace System.Management.Automation
         private bool IsExecutable(string path)
         {
 #if UNIX
-
-            if (Platform.NonWindowsIsExecutable(this.Path))
-            {
-                return true;
-            }
-#endif
+            return Platform.NonWindowsIsExecutable(this.Path);
+#else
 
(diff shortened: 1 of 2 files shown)


SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and it seems that $env:PATHEXT isn't needed on Unix systems

src/System.Management.Automation/engine/NativeCommandProcessor.cs
```diff
@@ -1288,14 +1285,21 @@ private void CalculateIORedirection(out bool redirectOutput, out bool redirectEr
...
+                return true;
+            }
+#else
+
             string myExtension = System.IO.Path.GetExtension(path);
 
-            string pathext = (string)LanguagePrimitives.ConvertTo(
-                this.Command.Context.GetVariableValue(SpecialVariables.PathExtVarPath),
-                typeof(string), CultureInfo.InvariantCulture);
+            var pathext = Environment.GetEnvironmentVariable("PATHEXT");
```
rjmholt: ~~It might be messier, but figured I'd ask.~~

~~We have a constant for this -- we could keep that defined on *nix and use it here instead of the string literal. It just might be confusing in SpecialVariables.cs that we have the constant and no variable. But that is what comments are for!~~

Reading through it again I see the constant variable name begins with `env:`. Seems better to me to not mix this line with that file.

rjmholt (approved):
LGTM!

src/System.Management.Automation/engine/NativeCommandProcessor.cs
iSazonov: Sorry, should I fix anything?

rjmholt: No -- I think you've done it the best way!

SteveL-MSFT (changes requested):
Based on @PowerShell/powershell-committee conclusion, we should not make this change as there is utility for an end user to create the PATHEXT env var and use it on non-Windows

iSazonov:
@SteveL-MSFT Your request was addressed.

iSazonov:
CI Appveyor temporary failed.

src/System.Management.Automation/engine/NativeCommandProcessor.cs
```diff
@@ -1288,30 +1285,43 @@ private void CalculateIORedirection(out bool redirectOutput, out bool redirectEr
...
 
-        private bool ValidateExtension(string path)
+        // On Windows, check the extension list and see if we should try to execute this directly.
+        // Otherwise, use the platform library to check executability
+        private bool IsExecutable(string path)
         {
-            // Now check the extension and see if it's one of the ones in pathext
+#if UNIX
+
+            if (Platform.NonWindowsIsExecutable(this.Path))
```
SteveL-MSFT: Why not just:

> return Platform.NonWindowsIsExecutable(this.Path);

SteveL-MSFT (changes requested):

src/System.Management.Automation/engine/NativeCommandProcessor.cs
iSazonov: From PowerShell Committee conclusion we should fallback on Unix to read PATHEXT.

iSazonov:
Reopen the PR to restart CIs.

src/System.Management.Automation/engine/NativeCommandProcessor.cs
SteveL-MSFT: Additional code would need to be added to support something like executing `foo` where `foo.py` exists and `.py` was added to `PATHEXT` on Unix since Windows takes care of this automatically.  That is a feature and not bug fix so certainly out of scope of this PR.  I'd also wait until there's customer request for such a feature.

iSazonov: @SteveL-MSFT I fixed your request. Tracking issue #7755 .

mklement0:
@SteveL-MSFT: I don't think `PATHEXT` fits into the Unix world - please see https://github.com/PowerShell/PowerShell/issues/7755#issuecomment-420307930

iSazonov:
@SteveL-MSFT Please update your review. I believe we can merge.

SteveL-MSFT (approved):

Issue #3832, opened 2017-05-20
Extend Get-Culture with a -Name parameter to allow retrieving a specific culture


mklement0:
`Get-Culture` currently supports returning information about the _current_ culture.

It would be helpful if it supported a `-Name` switch to allow retrieving information about a _specified_ culture.

(This would be analogous to existing `Get-Date` behavior, which by default returns the current date and optionally supports operating on a _specified_ date via `-Date`.)

Also, as @iSazonov  suggests, _listing_ available cultures should be supported: `-ListAvailable`, supplemented by an optional `-Type` parameter that defaults to `SpecificCultures` - (see https://msdn.microsoft.com/en-us/library/system.globalization.culturetypes(v=vs.110).aspx for permissible values, though not all of them may make sense).

Desired behavior
-----------------

```powershell
# Should be the equivalent of: [cultureinfo]::GetCultureInfo('fr-FR')
> Get-Culture fr-FR      # assumes that -Name is the 1st positional param

LCID             Name             DisplayName                                                                                                                                      
----             ----             -----------                                                                                                                                      
1036             fr-FR            French (France)                                                                                                                                  

# Should be the equivalent of [cultureinfo]::GetCultures('SpecificCultures')
> Get-Culture -ListAvailable

LCID             Name             DisplayName                                                                                                                                      
----             ----             -----------                                                                                                                                      
4096             af-NA            Afrikaans (Namibië)                                                                                                                              
1078             af-ZA            Afrikaans (South Africa)                                                                                                                         
# ....
```

Environment data
----------------
```powershell
PowerShell Core v6.0.0-beta (v6.0.0-beta.1)
```

iSazonov:
What is the benefit we get from getting the CultureInfo ?

mklement0:
If there's value in getting the _current_ culture's `[System.Globalization.CultureInfo]` instance - which is currently the sole purpose of `Get-Culture` - then there's obviously value in getting a _different_ culture's information.

For instance, if you wanted to get a given culture's list-separator character, you could then do

```powershell
(Get-Culture it-IT).TextInfo.ListSeparator  # assuming -Name is the 1st positional param
```

instead of

```powershell
[cultureinfo]::GetCultureInfo('it-IT').TextInfo.ListSeparator
```

iSazonov:
Then needed `-ListAvailable`

mklement0:
@iSazonov: Good idea - initial post updated.

Issue #7562, opened 2018-08-17
There should be more `Test-Command` and `Test-Module` cmdlets


SteveL-MSFT:
Specifically, `Test-Command` and `Test-Module` is used heavily in our tests and build module which currently rely on `Get-Command -ErrorAction SIlentlyContinue` which can mask other issues other than testing if a command exists.

- [ ] - Test-Culture (approved in #7702)

joeyaiello:
There is a reasonable design proposal for `Test-Command` in #8755, which is now closed as a duplicate. 

Marking this as `Up-for-Grabs` given that @PowerShell/powershell-committee is fairly confident that we would take something like this into PowerShell .

PR #7702, opened 2018-09-04
Add -Name, -NoUserOverrides and -ListAvailable parameters to Get-Culture cmdlet

iSazonov:
## PR Summary

Address #3832

Add new parameters in Get-Culture cmdlet:
- `-Name` - to allow retrieving a specific culture
- `-NoUserOverrides` - ignore user changes for current culture
- `-ListAvalable` - to allow retrieving all cultures supported on the platform

### Documentation

Please see https://github.com/dotnet/corefx/issues/6374#issuecomment-418827420 and following comments.
In short .Net Core accept all culture names in BCP-47 format and then delegate to OS so result depends on OS behavior and differ on different OS-s.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
@@ -60,7 +60,7 @@ namespace Microsoft.PowerShell.Commands
                 case CurrentCultureParameterSet:
                     if (NoUserOverrides)
                     {
-                        ci = CultureInfo.GetCultureInfo(Host.CurrentCulture.LCID);
+                        ci = CultureInfo.GetCultureInfo(Host.CurrentCulture.Name);
                     }
                     else
                     {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
@@ -24,6 +24,7 @@ namespace Microsoft.PowerShell.Commands
 
         /// <summary>
         /// Gets or sets culture names for which CultureInfo values are returned.
+        /// Empty string matches Invariant culture.
         /// </summary>
         [Parameter(ParameterSetName = NameParameterSet, Position = 0, ValueFromPipeline = true)]
         [ValidateSet(typeof(ValidateCultureNamesGenerator))]
@@ -74,9 +75,9 @@ namespace Microsoft.PowerShell.Commands
                     {
                         foreach (var cultureName in Name)
                         {
-                            if (NoUserOverrides && string.Equals(cultureName, Host.CurrentCulture.Name, StringComparison.CurrentCultureIgnoreCase))
+                            if (!NoUserOverrides && string.Equals(cultureName, Host.CurrentCulture.Name, StringComparison.CurrentCultureIgnoreCase))
                             {
-                                ci = new CultureInfo(cultureName, false);
+                                ci = Host.CurrentCulture;
                             }
                             else
                             {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
@@ -17,14 +17,14 @@ namespace Microsoft.PowerShell.Commands
         private const string ListAvailableParameterSet = "ListAvailable";
 
         /// <summary>
-        /// Gets and sets the specified culture.
+        /// Gets or sets the specified culture name.
         /// </summary>
         [Parameter(ParameterSetName = NameParameterSet)]
         [ValidateNotNull]
         public string Name { get; set; }
 
         /// <summary>
-        /// Gets and sets a switch to list all available cultures.
+        /// Gets or sets a switch to list all available cultures.
         /// </summary>
         [Parameter(ParameterSetName = ListAvailableParameterSet)]
         public SwitchParameter ListAvailable { get; set; }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
@@ -18,7 +18,7 @@ namespace Microsoft.PowerShell.Commands
     {
         private const string DefaultParameterSet = "Default";
         private const string NameParameterSet = "Name";
-        private const string ListAvailableParameterSet = "ListAvailable";
+        private const string ListParameterSet = "List";
 
         /// <summary>
         /// Gets or sets culture names for which CultureInfo values are returned.
@@ -28,10 +28,19 @@ namespace Microsoft.PowerShell.Commands
         public string[] Name { get; set; }
 
         /// <summary>
-        /// Gets or sets a switch to list all available cultures.
+        /// Gets or sets a switch to return predefined cultures.
+        /// By default we return cultures in current state (with custom changes).
+        /// With the switch on, we return predefined, original cultures.
         /// </summary>
-        [Parameter(ParameterSetName = ListAvailableParameterSet)]
-        public SwitchParameter ListAvailable { get; set; }
+        [Parameter(ParameterSetName = NameParameterSet)]
+        [ValidateNotNull]
+        public SwitchParameter Predefined { get; set; }
+
+        /// <summary>
+        /// Gets or sets a filter to list subset or all available cultures.
+        /// </summary>
+        [Parameter(ParameterSetName = ListParameterSet)]
+        public CultureTypes List { get; set; } = CultureTypes.AllCultures;
 
         /// <summary>
         /// Output:
@@ -52,7 +61,17 @@ namespace Microsoft.PowerShell.Commands
                     {
                         foreach (var cultureName in Name)
                         {
-                            CultureInfo ci = CultureInfo.GetCultureInfo(cultureName);
+                            CultureInfo ci;
+
+                            if (Predefined)
+                            {
+                                ci = new CultureInfo(cultureName, true);
+                            }
+                            else
+                            {
+                                ci = CultureInfo.GetCultureInfo(cultureName);
+                            }
+
                             WriteObject(ci);
                         }
                     }
@@ -62,8 +81,8 @@ namespace Microsoft.PowerShell.Commands
                     }
 
                     break;
-                case ListAvailableParameterSet:
-                    foreach (CultureInfo ci in CultureInfo.GetCultures(CultureTypes.AllCultures))
+                case ListParameterSet:
+                    foreach (CultureInfo ci in CultureInfo.GetCultures(List))
                     {
                         WriteObject(ci);
                     }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
@@ -17,11 +17,11 @@ namespace Microsoft.PowerShell.Commands
         private const string ListAvailableParameterSet = "ListAvailable";
 
         /// <summary>
-        /// Gets or sets the specified culture name.
+        /// Gets or sets culture names for which CultureInfo values are returned.
         /// </summary>
-        [Parameter(Position = 0, ParameterSetName = NameParameterSet)]
+        [Parameter(ParameterSetName = NameParameterSet, Position = 0, ValueFromPipeline = true)]
         [ValidateNotNull]
-        public string Name { get; set; }
+        public string[] Name { get; set; }
 
         /// <summary>
         /// Gets or sets a switch to list all available cultures.
@@ -32,7 +32,7 @@ namespace Microsoft.PowerShell.Commands
         /// <summary>
         /// Output the current Culture info object.
         /// </summary>
-        protected override void EndProcessing()
+        protected override void ProcessRecord()
         {
             switch (ParameterSetName)
             {
@@ -43,8 +43,18 @@ namespace Microsoft.PowerShell.Commands
                     }
                     else
                     {
-                        CultureInfo ci = CultureInfo.GetCultureInfo(Name);
-                        WriteObject(ci);
+                        try
+                        {
+                            foreach (var ciName in Name)
+                            {
+                                CultureInfo ci = CultureInfo.GetCultureInfo(ciName);
+                                WriteObject(ci);
+                            }
+                        }
+                        catch (CultureNotFoundException exc)
+                        {
+                            WriteError(new ErrorRecord(exc, "ItemNotFoundException", ErrorCategory.ObjectNotFound, Name));
+                        }
                     }
 
                     break;
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
@@ -17,14 +17,14 @@ namespace Microsoft.PowerShell.Commands
         private const string ListAvailableParameterSet = "ListAvailable";
 
         /// <summary>
-        /// Gets the specified culture.
+        /// Gets and sets the specified culture.
         /// </summary>
         [Parameter(ParameterSetName = NameParameterSet)]
         [ValidateNotNull]
         public string Name { get; set; }
 
         /// <summary>
-        /// List all available cultures.
+        /// Gets and sets a switch to list all available cultures.
         /// </summary>
         [Parameter(ParameterSetName = ListAvailableParameterSet)]
         public SwitchParameter ListAvailable { get; set; }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
@@ -19,7 +19,7 @@ namespace Microsoft.PowerShell.Commands
         /// <summary>
         /// Gets or sets the specified culture name.
         /// </summary>
-        [Parameter(ParameterSetName = NameParameterSet)]
+        [Parameter(Position = 0, ParameterSetName = NameParameterSet)]
         [ValidateNotNull]
         public string Name { get; set; }
 
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
@@ -12,11 +12,11 @@ namespace Microsoft.PowerShell.Commands
     ///     - culture by name
     ///     - list of all supported cultures.
     /// </summary>
-    [Cmdlet(VerbsCommon.Get, "Culture", DefaultParameterSetName = DefaultParameterSet, HelpUri = "https://go.microsoft.com/fwlink/?LinkID=113312")]
+    [Cmdlet(VerbsCommon.Get, "Culture", DefaultParameterSetName = CurrentCultureParameterSet, HelpUri = "https://go.microsoft.com/fwlink/?LinkID=113312")]
     [OutputType(typeof(System.Globalization.CultureInfo))]
     public sealed class GetCultureCommand : PSCmdlet
     {
-        private const string DefaultParameterSet = "Default";
+        private const string CurrentCultureParameterSet = "CurrentCulture";
         private const string NameParameterSet = "Name";
         private const string ListParameterSet = "List";
 
@@ -28,13 +28,12 @@ namespace Microsoft.PowerShell.Commands
         public string[] Name { get; set; }
 
         /// <summary>
-        /// Gets or sets a switch to return predefined cultures.
-        /// By default we return cultures in current state (with custom changes).
-        /// With the switch on, we return predefined, original cultures.
+        /// Gets or sets a switch to return current culture with user overrides (by default).
+        /// With the switch on, we return current culture without user overrides.
         /// </summary>
+        [Parameter(ParameterSetName = CurrentCultureParameterSet)]
         [Parameter(ParameterSetName = NameParameterSet)]
-        [ValidateNotNull]
-        public SwitchParameter Predefined { get; set; }
+        public SwitchParameter NoUserOverrides { get; set; }
 
         /// <summary>
         /// Gets or sets a filter to list subset or all available cultures.
@@ -50,9 +49,20 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         protected override void ProcessRecord()
         {
+            CultureInfo ci;
+
             switch (ParameterSetName)
             {
-                case DefaultParameterSet:
+                case CurrentCultureParameterSet:
+                    if (NoUserOverrides)
+                    {
+                        ci = CultureInfo.GetCultureInfo(Host.CurrentCulture.LCID);
+                    }
+                    else
+                    {
+                        ci = Host.CurrentCulture;
+                    }
+
                     WriteObject(Host.CurrentCulture);
 
                     break;
@@ -61,11 +71,9 @@ namespace Microsoft.PowerShell.Commands
                     {
                         foreach (var cultureName in Name)
                         {
-                            CultureInfo ci;
-
-                            if (Predefined)
+                            if (NoUserOverrides && string.Equals(cultureName, Host.CurrentCulture.Name))
                             {
-                                ci = new CultureInfo(cultureName, true);
+                                ci = new CultureInfo(cultureName, false);
                             }
                             else
                             {
@@ -82,9 +90,9 @@ namespace Microsoft.PowerShell.Commands
 
                     break;
                 case ListParameterSet:
-                    foreach (CultureInfo ci in CultureInfo.GetCultures(List))
+                    foreach (CultureInfo culture in CultureInfo.GetCultures(List))
                     {
-                        WriteObject(ci);
+                        WriteObject(culture);
                     }
 
                     break;
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Globalization;
 using System.Management.Automation;
 
 namespace Microsoft.PowerShell.Commands
@@ -8,16 +9,53 @@ namespace Microsoft.PowerShell.Commands
     /// <summary>
     /// Returns the thread's current culture.
     /// </summary>
-    [Cmdlet(VerbsCommon.Get, "Culture", HelpUri = "https://go.microsoft.com/fwlink/?LinkID=113312")]
+    [Cmdlet(VerbsCommon.Get, "Culture", DefaultParameterSetName = NameParameterSet, HelpUri = "https://go.microsoft.com/fwlink/?LinkID=113312")]
     [OutputType(typeof(System.Globalization.CultureInfo))]
     public sealed class GetCultureCommand : PSCmdlet
     {
+        private const string NameParameterSet = "Name";
+        private const string ListAvailableParameterSet = "ListAvailable";
+
+        /// <summary>
+        /// Gets the specified culture.
+        /// </summary>
+        [Parameter(ParameterSetName = NameParameterSet)]
+        [ValidateNotNull]
+        public string Name { get; set; }
+
+        /// <summary>
+        /// List all available cultures.
+        /// </summary>
+        [Parameter(ParameterSetName = ListAvailableParameterSet)]
+        public SwitchParameter ListAvailable { get; set; }
+
         /// <summary>
         /// Output the current Culture info object.
         /// </summary>
-        protected override void BeginProcessing()
+        protected override void EndProcessing()
         {
-            WriteObject(Host.CurrentCulture);
+            switch (ParameterSetName)
+            {
+                case NameParameterSet:
+                    if (Name == null)
+                    {
+                        WriteObject(Host.CurrentCulture);
+                    }
+                    else
+                    {
+                        CultureInfo ci = CultureInfo.GetCultureInfo(Name);
+                        WriteObject(ci);
+                    }
+
+                    break;
+                case ListAvailableParameterSet:
+                    foreach (CultureInfo ci in CultureInfo.GetCultures(CultureTypes.AllCultures))
+                    {
+                        WriteObject(ci);
+                    }
+
+                    break;
+            }
         }
     }
 }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Management.Automation;
 
@@ -18,12 +20,13 @@ namespace Microsoft.PowerShell.Commands
     {
         private const string CurrentCultureParameterSet = "CurrentCulture";
         private const string NameParameterSet = "Name";
-        private const string ListParameterSet = "List";
+        private const string ListAvailableParameterSet = "ListAvailable";
 
         /// <summary>
         /// Gets or sets culture names for which CultureInfo values are returned.
         /// </summary>
         [Parameter(ParameterSetName = NameParameterSet, Position = 0, ValueFromPipeline = true)]
+        [ValidateSet(typeof(ValidateCultureNamesGenerator))]
         [ValidateNotNull]
         public string[] Name { get; set; }
 
@@ -36,10 +39,10 @@ namespace Microsoft.PowerShell.Commands
         public SwitchParameter NoUserOverrides { get; set; }
 
         /// <summary>
-        /// Gets or sets a filter to list subset or all available cultures.
+        /// Gets or sets a switch to list all available cultures.
         /// </summary>
-        [Parameter(ParameterSetName = ListParameterSet)]
-        public CultureTypes List { get; set; } = CultureTypes.AllCultures;
+        [Parameter(ParameterSetName = ListAvailableParameterSet)]
+        public SwitchParameter ListAvailable { get; set; }
 
         /// <summary>
         /// Output:
@@ -63,7 +66,7 @@ namespace Microsoft.PowerShell.Commands
                         ci = Host.CurrentCulture;
                     }
 
-                    WriteObject(Host.CurrentCulture);
+                    WriteObject(ci);
 
                     break;
                 case NameParameterSet:
@@ -71,7 +74,7 @@ namespace Microsoft.PowerShell.Commands
                     {
                         foreach (var cultureName in Name)
                         {
-                            if (NoUserOverrides && string.Equals(cultureName, Host.CurrentCulture.Name))
+                            if (NoUserOverrides && string.Equals(cultureName, Host.CurrentCulture.Name, StringComparison.CurrentCultureIgnoreCase))
                             {
                                 ci = new CultureInfo(cultureName, false);
                             }
@@ -89,14 +92,32 @@ namespace Microsoft.PowerShell.Commands
                     }
 
                     break;
-                case ListParameterSet:
-                    foreach (CultureInfo culture in CultureInfo.GetCultures(List))
+                case ListAvailableParameterSet:
+                    foreach (var cultureInfo in CultureInfo.GetCultures(CultureTypes.AllCultures))
                     {
-                        WriteObject(culture);
+                        WriteObject(cultureInfo);
                     }
 
                     break;
             }
+        }
+    }
+
+    /// <summary>
+    /// Get list of valid culture names for ValidateSet attribute.
+    /// </summary>
+    public class ValidateCultureNamesGenerator : IValidateSetValuesGenerator
+    {
+        string[] IValidateSetValuesGenerator.GetValidValues()
+        {
+            var cultures = CultureInfo.GetCultures(CultureTypes.AllCultures);
+            var result = new List<string>(cultures.Length);
+            foreach (var cultureInfo in cultures)
+            {
+                result.Add(cultureInfo.Name);
+            }
+
+            return result.ToArray();
         }
     }
 }
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
@@ -9,10 +9,11 @@ namespace Microsoft.PowerShell.Commands
     /// <summary>
     /// Returns the thread's current culture.
     /// </summary>
-    [Cmdlet(VerbsCommon.Get, "Culture", DefaultParameterSetName = NameParameterSet, HelpUri = "https://go.microsoft.com/fwlink/?LinkID=113312")]
+    [Cmdlet(VerbsCommon.Get, "Culture", DefaultParameterSetName = DefaultParameterSet, HelpUri = "https://go.microsoft.com/fwlink/?LinkID=113312")]
     [OutputType(typeof(System.Globalization.CultureInfo))]
     public sealed class GetCultureCommand : PSCmdlet
     {
+        private const string DefaultParameterSet = "Default";
         private const string NameParameterSet = "Name";
         private const string ListAvailableParameterSet = "ListAvailable";
 
@@ -36,25 +37,22 @@ namespace Microsoft.PowerShell.Commands
         {
             switch (ParameterSetName)
             {
+                case DefaultParameterSet:
+                    WriteObject(Host.CurrentCulture);
+
+                    break;
                 case NameParameterSet:
-                    if (Name == null)
+                    try
                     {
-                        WriteObject(Host.CurrentCulture);
+                        foreach (var ciName in Name)
+                        {
+                            CultureInfo ci = CultureInfo.GetCultureInfo(ciName);
+                            WriteObject(ci);
+                        }
                     }
-                    else
+                    catch (CultureNotFoundException exc)
                     {
-                        try
-                        {
-                            foreach (var ciName in Name)
-                            {
-                                CultureInfo ci = CultureInfo.GetCultureInfo(ciName);
-                                WriteObject(ci);
-                            }
-                        }
-                        catch (CultureNotFoundException exc)
-                        {
-                            WriteError(new ErrorRecord(exc, "ItemNotFoundException", ErrorCategory.ObjectNotFound, Name));
-                        }
+                        WriteError(new ErrorRecord(exc, "ItemNotFoundException", ErrorCategory.ObjectNotFound, Name));
                     }
 
                     break;
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
@@ -10,7 +10,7 @@ namespace Microsoft.PowerShell.Commands
     /// Returns:
     ///     - the thread's current culture
     ///     - culture by name
-    ///     - list of all supported cultures
+    ///     - list of all supported cultures.
     /// </summary>
     [Cmdlet(VerbsCommon.Get, "Culture", DefaultParameterSetName = DefaultParameterSet, HelpUri = "https://go.microsoft.com/fwlink/?LinkID=113312")]
     [OutputType(typeof(System.Globalization.CultureInfo))]
@@ -37,7 +37,7 @@ namespace Microsoft.PowerShell.Commands
         /// Output:
         ///     - the thread's current culture
         ///     - culture by name
-        ///     - list of all supported cultures
+        ///     - list of all supported cultures.
         /// </summary>
         protected override void ProcessRecord()
         {
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
@@ -44,9 +44,9 @@ namespace Microsoft.PowerShell.Commands
                 case NameParameterSet:
                     try
                     {
-                        foreach (var ciName in Name)
+                        foreach (var cultureName in Name)
                         {
-                            CultureInfo ci = CultureInfo.GetCultureInfo(ciName);
+                            CultureInfo ci = CultureInfo.GetCultureInfo(cultureName);
                             WriteObject(ci);
                         }
                     }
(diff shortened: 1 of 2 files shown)


src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
```diff
@@ -1,23 +1,61 @@
...
-    [Cmdlet(VerbsCommon.Get, "Culture", HelpUri = "https://go.microsoft.com/fwlink/?LinkID=113312")]
+    [Cmdlet(VerbsCommon.Get, "Culture", DefaultParameterSetName = NameParameterSet, HelpUri = "https://go.microsoft.com/fwlink/?LinkID=113312")]
     [OutputType(typeof(System.Globalization.CultureInfo))]
     public sealed class GetCultureCommand : PSCmdlet
     {
+        private const string NameParameterSet = "Name";
+        private const string ListAvailableParameterSet = "ListAvailable";
+
+        /// <summary>
+        /// Gets and sets the specified culture.
```
mklement0: As an aside: I don't think it's worth reflecting that the property representing the parameter is read/write (that can easily be gleaned from the code) - just stating the purpose of the parameter should be sufficient.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
```diff
@@ -1,23 +1,61 @@
...
+        /// </summary>
+        [Parameter(ParameterSetName = NameParameterSet)]
+        [ValidateNotNull]
+        public string Name { get; set; }
```
mklement0: * Please make the `-Name` parameter _positional_, so that `Get-Culture ru-RU` works too, for instance.

* I suggest making `-Name` accept an _array_ of culture names (`string[]`); as a result, exceptions from non-existent cultures need to be transformed into _non-terminating_ errors.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
iSazonov: The comment should start with "Gets and sets" to fix CodeFactor issue.

mklement0: I see, but in this context such comments are more confusing than helpful. I know nothing about CodeFactor: does this particular rule a default rule, or one that was specifically turned on for this repo?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
```diff
@@ -1,23 +1,61 @@
...
     {
+        private const string NameParameterSet = "Name";
+        private const string ListAvailableParameterSet = "ListAvailable";
+
+        /// <summary>
+        /// Gets or sets the specified culture name.
+        /// </summary>
+        [Parameter(ParameterSetName = NameParameterSet)]
+        [ValidateNotNull]
+        public string Name { get; set; }
```
mklement0: * Please make the `-Name` parameter _positional_, so that `Get-Culture ru-RU` works too, for instance.

* I suggest making `-Name` accept an _array_ of culture names (`string[]`); as a result, exceptions from non-existent cultures need to be transformed into _non-terminating_ errors.

mklement0 (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
```diff
@@ -1,23 +1,61 @@
...
+        /// Gets or sets a switch to list all available cultures.
+        /// </summary>
+        [Parameter(ParameterSetName = ListAvailableParameterSet)]
+        public SwitchParameter ListAvailable { get; set; }
+
         /// <summary>
         /// Output the current Culture info object.
         /// </summary>
-        protected override void BeginProcessing()
+        protected override void EndProcessing()
```
SteveL-MSFT: Is there a reason to do this in `EndProcessing` vs `ProcessRecord`?  Typically, `EndProcessing` is used for cleanup after the pipeline has completed.

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
iSazonov: Done.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
iSazonov: Initially we do not support pipelines and using EndProcessing() was ok. Currently the cmdlet was enhanced to support pipelines and the code was moved to ProcessRecord().

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
mklement0: Thanks, @iSazonov, but you only addressed the positional issue, not making `-Name` an _array_.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
iSazonov: Native language is very flexible - I reworded the comment.

CodeFactor is a service. It uses StypeCop static analyzer too. Yes, we can to turn in/out rules. Currently we have some PRs to turn StypeCop and editors settings. Plan is to complete the work after 6.1 GA.

mklement0:
@iSazonov: That is unfortunate; the problem has since been fixed on _Windows_, but macOS and Linux are still affected.

iSazonov:
I see. I'll skip the test on Linux and MacOs. Should I add a comment that it is a bug in .Net Core? Is there a tracking issue?

iSazonov:
@mklement0 Thanks! I added a comment there.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Culture.Tests.ps1
```diff
@@ -1,25 +1,64 @@
...
+
+    It "Should return the culture array with '-ListAvailable' parameter" {
+
+        $ciArray = Get-Culture -ListAvailable
+        ,$ciArray | Should -BeOfType [System.Array]
+        $ciArray.Count | Should -BeGreaterThan 0
+        $ciArray[0] | Should -BeOfType [CultureInfo]
+    }
+
+    It "Should write an error on unsupported culture name" -Skip:(!$isWindows) {
```
dantraMSFT: I suggest adding the issue link as a comment for -Skip

https://github.com/dotnet/corefx/issues/6374

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
```diff
@@ -1,23 +1,71 @@
...
     public sealed class GetCultureCommand : PSCmdlet
     {
+        private const string NameParameterSet = "Name";
+        private const string ListAvailableParameterSet = "ListAvailable";
+
+        /// <summary>
+        /// Gets or sets culture names for which CultureInfo values are returned.
+        /// </summary>
+        [Parameter(ParameterSetName = NameParameterSet, Position = 0, ValueFromPipeline = true)]
+        [ValidateNotNull]
```
dantraMSFT: Shouldn't this be ValidateNotNullOrEmpty?

dantraMSFT (changes requested):

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and continue to be consistent in our thinking that for literals, a non-terminating error should be returned, and for filtering/wildcards, it should return nothing if not found.  Should also consider adding a Test-Culture cmdlet.

mklement0:
I guess we can work around this by validating names against `CultureInfo.GetCultures(CultureTypes.AllCultures)` while we're waiting for a CoreFX fix - which may or may not come.

@SteveL-MSFT: Currently, the PR has no wildcard support for `-Name` - should we add it? Might be handy, but adds implementation complexity.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
iSazonov: For empty string name `InvariantCulture` is returned.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Culture.Tests.ps1
iSazonov: I updated the test - now it throw on Unix-es too. Link to useful comment added too.

iSazonov:
@SteveL-MSFT 
> Should also consider adding a Test-Culture cmdlet.

We have #7562. I suggest to explicitly track all approved Test-* cmdlets there.

iSazonov:
> I guess we can work around this 

@mklement0 I think it is ok to follow .Net Core behavior "as is" - if they really think it's right, then we can hardly do better.

mklement0:
@iSazonov: Since we're not just _providing access_ to the underlying CoreFx functionality, but packaging it in a PowerShell-appropriate way, I don't think we're obligated to surface the functionality as-is.

I have yet to hear a _good_ reason for the CoreFx behavior (there may be one, but I haven't heard it), and users who really want to implicitly create cultures such as `zz-AA` and `01;2.45"78&$` can use the CoreFx types directly.

iSazonov:
@mklement0 With `-ListAvailable` we get good UX. We could add a filter for `-Name` (based on a completer) but it can be expensive and it seems we do not fully understand the behavior of .Net Core in order to overlap it. I think we shouldn't delay the PR, we can postpone this and fix later if needed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
dantraMSFT: @iSazonov - I think that behavior's too subtle.

Additionally, the ProcessRecord doesn't match the class comment (Return's the thread's current culture).  In short, reviewing the code appears to not support getting the current culture.

NOTE: Line 40 that compares Name to null 'looks' like dead code since name must be non-null for the Name parameter set.  Having an explicit or fall through 'default' case statement will make this clearer.

I suggest the following:

1: Retain the original behavior explicitly
Set the DefaultParameterSetName to 'default' and return the thread's current culture.

2: Name parameter set
Either clearly document that an empty string returns InvariantCulture or don't allow empty strings and require explicit names.  

3: ListAvailable
As implemented.

dantraMSFT (changes requested):

mklement0:
@iSazonov: Postponing wildcard support sounds good to me, but I do think we should get clarity on is whether we truly want to support calls such as `Get-Culture '01;2.45"78&$'`, because if we do so now, we have a breaking change on our hands if we want to change it later.

I see only down-sides to surfacing the baffling CoreFx functionality as-is.

Conversely, if we restrict allowed names to the ones reported by `-ListAvailable` now, it is less problematic to loosen that restriction later, _if_ it turns out that that is really necessary.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
mklement0: I too think that `""` is obscure and suggest requiring explicit use of `Invariant-Culture`, i.e.:  
 
`Get-Culture Invariant-Culture`

iSazonov: 1 - Fixed.
2- Let's consider the scenario - script asks user to input a culture, user press Enter to choose "Default". Without user typing the script get empty string as "default". I guess CoreFX consider InvariantCulture as default if culture is unknown or not selected. We use InvariantCulture widely in PowerShell code too. So allowing empty string in Get-Culture looks good.

iSazonov:
> whether we truly want to support calls such as Get-Culture '01;2.45"78&$', because if we do so now, we have a breaking change on our hands if we want to change it later.

I don't know whether we truly want. The breaking change will be in grace arae - I'd expect users will use right culture names. On the other hand, it is not clear what the filter should be. from the CoreFX discussion we know that there may be some special logic. There can be a situation similar we see with encodings - we get a short list but really supported encodings there is more.
I suggest document the feature and wait feedback/

mklement0:
> from the CoreFX discussion we know that there may be some special logic.
> we get a short list but really supported encodings there is more.

There is an official list of all predefined cultures: `CultureInfo.GetCultures(CultureTypes.AllCultures)`, which we surface via `-ListAvailable`.

The special logic is that you can _create_ cultures on demand, which on Windows allows you to combine any (possibly made-up) language with any (made-up) region.

Surfacing this functionality is not part of the mandate of `Get-Culture` the way I understand it.

As shown in the linked CoreFx discussion, it's not hard to validate names against; in the simplest form (though performance considerations may require a more sophisticated approach):

```powershell
# Returns the requested culture or $null, if not predefined.
[Array]::Find([cultureinfo]::GetCultures('AllCultures'), [Predicate[cultureinfo]] { param($ci) $ci.Name -eq $Name })
```

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
mklement0: Re 2: Curiously, the empty string is not just a placeholder that signals the intent to retrieve the invariant culture when you call `.GetCultureInfo()`, it is the actual value of the `.Name` property  of the culture-info object returned - so I guess it makes sense to surface that.(Unlike what I thought earlier, `"Invariant-Culture"` doesn't actually work; on Unix it implicitly creates an ad-hoc culture, on Windows it breaks.)

mklement0:
Another thing to consider: 

`AllCultures` doesn't mean _all_ cultures; from https://docs.microsoft.com/en-us/dotnet/api/system.globalization.culturetypes?view=netcore-2.1:

> `AllCultures` is a composite field that includes the `NeutralCultures`, `SpecificCultures`, and `InstalledWin32Cultures` values.

Notably, it excludes custom (user-defined) cultures: `UserCustomCulture`, a subset of which appears to be `ReplacementCultures` (custom cultures that override predefined ones - see below).

(On a side note: accidental "creation" of a bogus culture with e.g., `[cultureinfo] 'zz-AA'` is _not_ reflected in the `UserCustomCulture` category).

Sounds like we should include `UserCustomCulture`, right?
Or provide an additional parameter to allow selection of the desired categories explicitly?

Additionally, predefined cultures can be _overridden_ (tweaked on a given system); 
the `.GetCultureInfo()` _method_ seemingly never returns such tweaked cultures,
whereas the `CultureInfo` _constructor_ does by default, and also offers a Boolean parameter to control that.
Seemingly, however, a tweaked version is only returned for the _current_ culture - I couldn't really make sense of the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.-ctor?view=netcore-2.1), which still only reflect the Windows perspective.

Do we need to worry about this?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
```diff
@@ -1,23 +1,69 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Globalization;
 using System.Management.Automation;
 
 namespace Microsoft.PowerShell.Commands
 {
     /// <summary>
     /// Returns the thread's current culture.
```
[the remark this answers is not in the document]
iSazonov: Fixed.

iSazonov:
Full definition
```c#
    //
    // Summary:
    //     Defines the types of culture lists that can be retrieved using the System.Globalization.CultureInfo.GetCultures(System.Globalization.CultureTypes)
    //     method.
    [Flags]
    public enum CultureTypes
    {
        //
        // Summary:
        //     Cultures that are associated with a language but are not specific to a country/region.
        //     The names of .NET Framework cultures consist of the lowercase two-letter code
        //     derived from ISO 639-1. For example: &quot;en&quot; (English) is a neutral culture.
        NeutralCultures = 1,
        //
        // Summary:
        //     Cultures that are specific to a country/region. The names of these cultures follow
        //     RFC 4646 (Windows Vista and later). The format is &quot;&lt;languagecode2&gt;-&lt;country/regioncode2&gt;&quot;,
        //     where &lt;languagecode2&gt; is a lowercase two-letter code derived from ISO 639-1
        //     and &lt;country/regioncode2&gt; is an uppercase two-letter code derived from
        //     ISO 3166. For example, &quot;en-US&quot; for English (United States) is a specific
        //     culture.
        SpecificCultures = 2,
        //
        // Summary:
        //     All cultures that are installed in the Windows operating system. Note that not
        //     all cultures supported by the .NET Framework are installed in the operating system.
        InstalledWin32Cultures = 4,
        //
        // Summary:
        //     All cultures that ship with the .NET Framework, including neutral and specific
        //     cultures, cultures installed in the Windows operating system, and custom cultures
        //     created by the user. System.Globalization.CultureTypes.AllCultures is a composite
        //     field that includes the System.Globalization.CultureTypes.NeutralCultures, System.Globalization.CultureTypes.SpecificCultures,
        //     and System.Globalization.CultureTypes.InstalledWin32Cultures values.
        AllCultures = 7,
        //
        // Summary:
        //     Custom cultures created by the user.
        UserCustomCulture = 8,
        //
        // Summary:
        //     Custom cultures created by the user that replace cultures shipped with the .NET
        //     Framework.
        ReplacementCultures = 16
```

iSazonov:
There is another side. Maybe we will implement Set-Culture. Then the limitations in Get-Culture can limit us. So UserCustomCulture and ReplacementCultures is worth adding.

mklement0:
Thanks for the source-code snippet, @iSazonov: Curiously, part of the comment text contradicts the actual value by suggesting that custom cultures are included in `AllCultures`, whereas value `7` implies that they're not (`1 + 2 + 4`). 

You also have to wonder what `InstalledWin32Cultures` means on other platforms - does it generically refer to _predefined_ cultures? The docs often talk about _cached_ cultures, without clarifying what that means; `.GetCultureInfo()` uses cached cultures, whereas the `CultureInfo` constructors do not.

mklement0:
Agreed, though I'm unclear on what happens if `ReplacementCultures` is also specified: is it the tweaked versions of predefined cultures that are then returned? In the abstract, `ReplacementCultures` is by definition a _subset_ of the cultures returned by `AllCultures`.

If so, and we include it, we shouldn't do so _unconditionally_, because users then have no way to access the _original_ definition (at least not via `Get-Culture`).

Perhaps only add  `UserCustomCulture` by default and offer a `-UseOverride` switch that:

* adds `ReplacementCultures` on demand with `-ListAvailable`, 
* and, for `-Name`, use `CultureInfo()` constructor with `useUserOverride` set to `true`

Again, based on the docs I'm unclear on whether the request to use the override is only honored for the _current_ culture or for any predefined culture requested.

iSazonov:
For reference https://github.com/dotnet/coreclr/blob/030a3ea9b8dbeae89c90d34441d4d9a1cf4a7de6/src/System.Private.CoreLib/src/System/Globalization/CultureInfo.cs#L542

We can expose new parameter for CultureTypes and default AllCultures.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Culture.Tests.ps1
```diff
@@ -1,25 +1,68 @@
...
+        $ci = Get-Culture -Name ru-RU
+        $ci | Should -BeOfType [CultureInfo]
+        $ci.Name | Should -BeExactly "ru-RU"
+    }
+
+    It "Should return specified cultures with '-Name' parameter" {
 
+        $ciArray = Get-Culture "", "ru-RU"
+        $ciArray | Should -BeOfType [CultureInfo]
+        ,$ciArray | Should -BeOfType [System.Array]
```
SteveL-MSFT: prefixing this with a comma always make this an array, so this line doesn't really test anything related to this cmdlet.  I think line 30 is sufficient to validate you got an array

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Culture.Tests.ps1
```diff
@@ -1,25 +1,68 @@
...
+    }
+
+    It "Should write an error on unsupported culture name" {
+
+        # The strange culture name come from the fact
+        # that .Net Core behavior depend on the underlying OS
+        # and can differ on different platforms.
+        # See https://github.com/dotnet/corefx/issues/6374#issuecomment-418827420
+        $ci = Get-Culture -Name "abcdefghijkl" -ErrorAction SilentlyContinue
+        $e = $error[0]
```
SteveL-MSFT: Why not use existing pattern of: $e = { foo -ea stop } | Should -Throw -ErrorId bar -PassThru

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Culture.Tests.ps1
SteveL-MSFT: That is true, but until we have a way in PowerShell and Pester to differentiate, it seems unnecessary to introduce a new pattern.  My preference is to just comment that this is a non-terminating error.  There is an open issue to add a property to determine at runtime terminating vs non-terminating and we can leverage that in the future to update exist tests.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
```diff
@@ -1,23 +1,75 @@
...
+                case DefaultParameterSet:
+                    WriteObject(Host.CurrentCulture);
+
+                    break;
+                case NameParameterSet:
+                    try
+                    {
+                        foreach (var cultureName in Name)
+                        {
+                            CultureInfo ci = CultureInfo.GetCultureInfo(cultureName);
```
mklement0: This should use `new CultureInfo(cultureName, true)` to also reflect overrides for the current culture, for consistency with the other parameter sets. See https://github.com/PowerShell/PowerShell/pull/7702#issuecomment-420048195

mklement0 (changes requested):

iSazonov:
@mklement0 Thanks for the deep investigation!

I think currnet culture API logic looks very tricky. I'd prefer keep the PR and cmdlet as simple as possible and keep a way to enhance the cmdlet later without a breaking change.
So I guess we should rename `ListAvailable` to `ListSystem`. Later we could conside Set-Culture and in the time add `ListAvailable` to return user overrides, filter like `CultureType` and so on.

iSazonov:
> Also, .GetCultures() always returns replacement custom cultures, if any are defined, without the ability to retrieve the original definition.

Seems not always. See https://source.dot.net/#System.Private.CoreLib/src/System/Globalization/CultureInfo.cs,542

> To be consistent with -ListAvailable, -Name too should report user overrides 

Let's consider useful scenarios.
1. User want to get current culture and either revise the setting or use in script:
```powershell
$currentCulture = Get-Culture
```
2. User have currupted culture and want reset:
```powershell
Get-Culture -Name en-Us | Set-Culture
```
3. User want look in depth:
```powershell
Get-Culture -List   # Default is -List AllCultures
Get-Culture -List InstalledWin32Cultures
Get-Culture -List ReplacementCultures
Get-Culture -List SpecificCultures+NeutralCultures
```

mklement0:
>Seems not always. See https://source.dot.net/#System.Private.CoreLib/src/System/Globalization/CultureInfo.cs,542

What I meant was: _If_ replacement cultures are selected for retrieval - and `AllCultures` includes them - then you get them, without being able to retrieve the original definition of the culture that they replace - that's a CoreFx limitation (though one that may never matter).

(Remember that replacement cultures have nothing to do with user overrides.)

---

> Let's consider useful scenarios.

1. is fine: `Get-Culture` returns a `CultureInfo` object for the current culture that includes user overrides, if present.

2. You're correct: I hadn't considered the reset-to-original scenario:

To enable it, `Get-Culture -Name <culture>` would have to return the culture _without overrides_, unlike what I proposed earlier.

(As an aside, it's unclear to me if the concept of overrides, which are registry-based on Windows, even applies to Unix-like platforms.)

That means: 

* consistently using `.GetCultureInfo()` for `-Name` - as you already do - is correct.

* we need to document clearly that, with, say, `en-US` in effect, `Get-Culture` by itself and `Get-Culture en-US` may return _different_ objects (with overrides in the first case, without in the second), and explain _why_.

3.  Your proposed `-List` syntax:

That won't work, because PowerShell doesn't supports switches with optional arguments: `-List` would either have to _always_ require an argument (i.e., be a (non-switch) parameter) or _never support one_ (i.e., be a switch; except for tacking on `:$False` or `:$True`)

Also, according to https://github.com/dotnet/corefx/issues/32207#issuecomment-420372687, `InstalledWin32Cultures` is deprecated, and the only supported flags in Core are `SpecificCultures`, `NeutralCultures`, as well as their combination, `AllCultures`.

So we need to stick with a _switch_ such as `-ListAvailable`, and - if we want to support it (right away) - a _separate_ `-CultureType` parameter that accepts one of `AllCultures`, `SpecificCultures`, `NeutralCultures`.

iSazonov:
@mklement0 Please take another look.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
```diff
@@ -1,23 +1,94 @@
...
         /// <summary>
-        /// Output the current Culture info object.
+        /// Gets or sets a switch to return predefined cultures.
+        /// By default we return cultures in current state (with custom changes).
+        /// With the switch on, we return predefined, original cultures.
         /// </summary>
-        protected override void BeginProcessing()
+        [Parameter(ParameterSetName = NameParameterSet)]
+        [ValidateNotNull]
+        public SwitchParameter Predefined { get; set; }
```
mklement0: As discussed, I don't think we need this switch: user overrides only ever exist for the _current_ culture, so `Get-Culture` without arguments returns the current culture _with_ overrides. 

It's fine to let `Get-Culture` with `-Name` return the  culture _without overrides_, if the argument happens to be the current culture.

That way, user can gain access to either, as desired - we just need to document it as such, along the lines of:

"Use `Get-Culture` to get the current culture _with_ overrides; use `Get-Culture $PSCulture` if you want the _original_ definition."

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
```diff
@@ -1,23 +1,94 @@
...
+
+                    break;
+                case NameParameterSet:
+                    try
+                    {
+                        foreach (var cultureName in Name)
+                        {
+                            CultureInfo ci;
+
+                            if (Predefined)
```
mklement0: As stated, I don't think we need this, after all - just always use `.GetCultureInfo()` to get the _original_ definition, without a user override (if the current culture happens to be requested).
(It is therefore a moot point, but just to be clear it is `new CultureInfo(cultureName, true)` (or just `new CultureInfo(cultureName)` that _does_ give you the overrides, so the logic here is backwards.)

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
```diff
@@ -1,23 +1,94 @@
...
-        protected override void BeginProcessing()
+        [Parameter(ParameterSetName = NameParameterSet)]
+        [ValidateNotNull]
+        public SwitchParameter Predefined { get; set; }
+
+        /// <summary>
+        /// Gets or sets a filter to list subset or all available cultures.
+        /// </summary>
+        [Parameter(ParameterSetName = ListParameterSet)]
+        public CultureTypes List { get; set; } = CultureTypes.AllCultures;
```
mklement0: The default value never takes effect, because it is only by actually using `-List` that this parameter set gets selected, and if use `-List`, specifying an argument is _mandatory_.

As stated, we need

* `-ListAvailable` as a _switch_

* a separate `-CultureType` parameter that defaults to `AllCultures`

Also, I recommend _not_ exposing the enum type `CultureTypes` directly,  given that it contains obsolete values that would only cause confusion; given the statement in https://github.com/dotnet/corefx/issues/32207#issuecomment-420372687, we should limit the values to `AllCultures`, `NeutralCultures`, and `SpecificCultures`.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
```diff
@@ -1,23 +1,94 @@
...
+                        {
+                            CultureInfo ci;
+
+                            if (Predefined)
+                            {
+                                ci = new CultureInfo(cultureName, true);
+                            }
+                            else
+                            {
+                                ci = CultureInfo.GetCultureInfo(cultureName);
```
mklement0: Given my discussions in the CoreFx repo, I still feel strongly that we should validate each `cultureName` against the list of predefined cultures reported by `CultureInfo.GetCultures(CultureTypes.AllCultures)`. 

There is no good reason to follow the - IMO misguided - CoreFx behavior. No one should expect a cmdlet named _Get_-Culture to _create_ cultures on demand, likely unintentionally. It is much more useful to be notified of having accidentally typed the name of a non-predefined culture by way of an error.

We can allow ad-hoc creation on an opt-in basis with another switch, say `-CreateOnDemand`, but I'm not sure that's even necessary.

mklement0 (changes requested):

iSazonov:
Two already discussed questions we are ready to send to PowerShell Committee. /cc @SteveL-MSFT 

1. Filtering input culture names in `-Name` parameter
I think we can pass all formats because we haven't official strong specifications ("delegate to OS") and CoreFX accept them. Currently the cmdlet works with _right_ names and we can not worry about "wrong" names, especially since we do not know exactly what it is and they can differ by platforms (CoreFX returns an exception from OS).

2. `-List` parameter vs `-ListAvailable` switch
I think `-List` parameter with CultureTypes type is better compromise.
`-ListAvailable` switch can confuse users - is there (what is) "no available" cultures? We also use `-ListAvailable` in `Get-Module` and I do not see anything in common between these parameters  although it should have been.
`-ListAvailable`  assume second parameter `-CultureType` for filtering. But we can get the same with one parameter. Also we can add an attribute to filter acceptable values of the enum if we need.

vexx32:
Not sure on your last point there. `Get-Module` lists only *currently loaded* modules by default, so I think it makes some sense for `Get-Culture` to behave the same way?

But I agree; following that pattern means that -ListAvailable *should* be a switch parameter.

iSazonov:
> Get-Culture should list the currently 'loaded' culture.

It is a breaking change. This is completely unacceptable.

mklement0:
@iSazonov 

> Please clarify why we should disable Get-Culture ru-US if Windows allow this?

Because it is not a _predefined_ culture and you're _getting_ something, you're not saying _create this for me if it doesn't exist_.

In 99% of the cases (obviously, a speculation), you'll want to obtain a _predefined_ culture, and in doing so you'll want to know if you've fat-fingered `ru-RU` to `ru_RU` for instance; you don't want to get a fictitious culture in that case, just because the underlying APIs allow you to _create_ them on demand.

_Creating_ a culture, ad hoc, should be a _deliberate act_.
As an advanced use case, we can just leave that to CoreFx for users who truly need it.
If we do want to expose it in PowerShell, a simple `-CreateOnDemand` switch will do.

> when you try to write a script, you most likely want to get something more specific.

Yes, in a script. Interactively, as you're exploring, both listing _all_ predefined cultures _and_ being able to match by, say, `ru*` is helpful  (as also mentioned by @vexx32) - but I thought you wanted to leave implementing the latter for later.

Ultimately, I don't feel strongly about how listing is implemented, however, and `-List <type>` is fine too.

mklement0:
@iSazonov: I think all that @vexx32  meant to say was that you can conceive of the _current_ culture that argument-less `Get-Culture` returns as  the "loaded" one, which makes it analogous to an argument-less `Get-Module` call.

@vexx32: I see your point about _loaded_ vs. _available_; my focus was on the _available_ aspect, and that the name `Get-Culture` makes no reference to _current_. But it's clear that this mustn't change, so it is a moot point.

iSazonov:
> In 99% of the cases (obviously, a speculation), you'll want to obtain a predefined culture

This scenario has never even been considered never in CoreFX until you have created a issue. Although it is clear that this is useful for reset culture scenario.

Now I played with different culture filters. The results look very strange. Now I do not understand why they would be useful to me. This key (-List/-ListAvailable) should be considered as returning "a information list". I am convinced that no one will delve into the depths of our discussion and will simply receive "an information list" to write correctly "zh-MO" not "zh-MK". This is another argument to do nothing in addition to what the OS returns.

mklement0:
> This scenario has never even been considered never in CoreFX until you have created a issue.

What I am suggesting _was  the default scenario before Windows 10_. Basically, the implicit ad-hoc creation crept into the .NET API - to its detriment - based on changes in the underlying Windows API.

This is what happens on Windows 7:

```powershell
PSonWin7> [cultureinfo] 'zz-YY'
Cannot convert value "zz-YY" to type "System.Globalization.CultureInfo". Error: "Culture is not supported.
```

mklement0:
> no one will delve into the depths of our discussion and will simply receive "an information list" to write correctly "zh-MO" not "zh-MK".

I don't understand what you're saying.

mklement0:
To summarize what I think is a sensible resolution:

```powershell
# Given a culture name (only), find it only among the predefined cultures
# (as reported by [cultureinfo]::GetCultures('allcultures')).
# Report a non-terminating error otherwise.
Get-Culture ru-RU 

# Asking for ad-hoc culture definitions requires a new switch, -CreateOnDemand,
# which signals that you indeed want to create a culture name 'fu-ba_r' on demand
# (whatever that results in).
# Note that this will fail on Windows 8.1-, because the on-demand creation wasn't
# introduced to the WinAPI until Windows 10.
Get-Culture fu-ba_r -CreateOnDemand
```

That way, we do not hide any capabilities of the underlying API, we simply surface them differently - in a manner that makes more sense to me.

iSazonov:
>Get-Culture fu-ba_r -CreateOnDemand

"Create" in "Get" cmdlet is very surprising. This once again demonstrates that this API is unique and we cannot (and should not try) do it better than CoreFX team.

My main motivation was:
1. to get information about traditional culture (ex.: datetime format) by `Get-Culture ru_Ru`
2. to get informational list of traditional cultures by `Get-Culture -List`

I don't think that anyone will need more complex scenarios and that anyone will understand the tricks of this API more deeply.

Current design covers this scenarios and doesn't block further possible improvements. I would prefer to stop at this and then continue on after receiving feedback to address specific scenarios.

vexx32:
I almost think it'd be better to instead relegate that functionality to an auxiliary cmdlet e.g `New-Culture` or something.

I agree that the Get cmdlet probably shouldn't be able to do that.

mklement0:
@iSazonov 

> "Create" in "Get" cmdlet is very surprising.

I wholeheartedly agree, yet you're proposing that the creation aspect be _implicit_ and that a user you cannot select between "give me an existing" and "create this for me on demand", which is even worse.

> This once again demonstrates that this API is unique and we cannot (and should not try) do it better than CoreFX team.

> and we cannot (and should not try) do it better than CoreFX team.

Re cannot: Yes, we can, by requiring an explicit switch to signal the intent to create on demand.

Re should not: Yes, we should, namely whenever the underlying API is suboptimal and cannot be changed due to backward compatibility (as is the case here) or if it cannot surfaced in a PowerShell-idiomatic way.

Re your main motivation:

If you think that is sufficient, then you're agreeing with my proposal above, _except_ you don't want to `-CreateOnDemand` switch - and I'm fine with omitting, as I think this will rarely be used; the `New-Cmdlet`  solution that @vexx32 mentions sounds like a viable alternative (although that cmdlet would have to return predefined cultures too).

You current design does _not_ fit your main motivation, in that it'll happily return fictitious culture `fu-bar_n`.

iSazonov:
> You current design does not fit your main motivation, in that it'll happily return fictitious culture fu-bar_n, for instance.

I believe we can ignore this on current stage. Our goal is Set-Culture. But when I think about it, I can not even imagine what difficulties we will encounter. Therefore, I would not do any too "clever" things until the whole chain (Get-Culture, modify, Set-Culture) works. After that we could add useful things even if they fell under "gray breaking changes" that we see is not rare in our dev process here.

SteveL-MSFT:
My recommendation is to focus on the 80% case.  Advanced users can always call the .NET API directly.  I would simplify this to:

1. keep current `Get-Culture` to return current culture
2. `-ListAvailable` does `[cultureinfo]::GetCultures([System.Globalization.CultureTypes]::AllCultures)`
3. defer adding `-CultureType` to the future based on customer demand for it
4. error on cultures that don't match AllCultures (a `New-Culture` may make sense in the future, but it's not clear to me how many people would use it)

iSazonov:
@SteveL-MSFT I replaced `-ListAvailable` switch with `-List` typed parameter because it is not at all clear what is "Available" - different user communities may have different expectations. Even if we indicate in the documentation that this is `AllCultures`, this will not clarify anything to users and the length of this discussion clearly shows that. 
If CultureType enum will get new flags (seeing what kind of transformation undergoes Windows 10 we can expect it) we get them automatically with `-List` typed parameter.
We also did not explore the features of Linux and MacOs in this discussion. There may be its own habits and the only way not to stumble over them is to follow the .Net Core API.
So I'd defer `ListAvailable` switch to the future too in favor to `-List`.

vexx32:
-ListAvailable is consistent with existing things like Get-Module

It seems that Steve is recommending we keep with this pattern.

Parallels:

1. (No Params)
Both Get-Module and Get-Culture - return 'currently loaded / in use items'
2. -ListAvailable
Get-Module enumerates modules that *could* be loaded if needed.
Get-Culture would enumerate All cultures, retrieving items that are currently available to user

iSazonov:
> -ListAvailable is consistent with existing things like Get-Module

@vexx32 These are not comparable things. You will need a lot of time to review the documentation and .Net Core code associated with the cultures. After that, you will no doubt agree that these are incomparable things. :-)

vexx32:
Certainly, they aren't comparable in .NET, but I think the idea is much more for internal consistency in PS's methods of *exposing* such things, regardless of whether they are inherently comparable in their raw form. 😄

mklement0:
@iSazonov:

I'm with @vexx32.

> You will need a lot of time to review the documentation and .Net Core code associated with the cultures.

No, you don't, if you define _available_ as _predefined_, i.e., as `CultureInfo.GetCultures(CultureTypes.AllCultures)`

No platform-specific knowledge required - let CoreFx tell us what the predefined cultures are, on all supported platforms.

iSazonov:
> which sounds perfectly sensible to me

Are you sure? :-) When I reread your posts above, I see that your way was long and not very straight. My fear is that some users might not get what they want with `ListAvailable` whereas using proposed `List` gives them more flexibility. ( We can expect further changes in Windows 10 and especially in different Unix systems there may be different culture models.)

vexx32:
If you're worried about future-proofing, I would think that it's more useful to separate out the culture the specification to a separate parameter, rather than trying to bundle it into a combination -List parameter.

I'd think it makes more sense to have -ListAvailable and then a separate -Type parameter.

SteveL-MSFT:
@PowerShell/powershell-committee reviewed this and agree on:

  - keep current Get-Culture to return current culture
  - -ListAvailable does [cultureinfo]::GetCultures([System.Globalization.CultureTypes]::AllCultures)
  - defer adding -CultureType to the future based on customer demand for it
  - error on cultures that don't match AllCultures (a New-Culture may make sense in the future, but it's not clear to me how many people would use it)

iSazonov:
@SteveL-MSFT @mklement0 The PR is ready for final review.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
SteveL-MSFT: Perhaps add a comment about this in the code

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
SteveL-MSFT: Agree, based on @PowerShell/powershell-committee discussion we should only return cultures from the AllCultures list.  Creation is out of scope of this PR.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Culture.Tests.ps1
```diff
@@ -1,25 +1,64 @@
...
+        $culture = Get-Culture
+        $culture | Should -BeOfType [CultureInfo]
+        ($culture).LCID | Should -Be $host.CurrentCulture.LCID
+
+        Get-Culture -NoUserOverrides | Should -BeOfType [CultureInfo]
+    }
+
+    It "Should have `$PSCulture variable be equivalent to (Get-Culture).Name" {
+
+        (Get-Culture).Name | Should -Be $PsCulture
```
SteveL-MSFT: Based on the test description, it would seem that this should be:

> $PSCulture | Should -BeExactly (Get-Culture).Name

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Culture.Tests.ps1
```diff
@@ -1,25 +1,64 @@
...
+        $ci.Name | Should -BeExactly "ru-RU"
+
+        $ci = Get-Culture -Name ru-RU -NoUserOverrides
+        $ci | Should -BeOfType [CultureInfo]
+        $ci.Name | Should -BeExactly "ru-RU"
+    }
+
+    It "Should return specified cultures with '-Name' parameter" {
+
+        $ciArray = Get-Culture "", "ru-RU"
```
SteveL-MSFT: Seems like we should have a specific test case to handle invariant culture making it clear what is being validated

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
```diff
@@ -1,23 +1,123 @@
...
+
+                    break;
+                case NameParameterSet:
+                    try
+                    {
+                        foreach (var cultureName in Name)
+                        {
+                            if (NoUserOverrides && string.Equals(cultureName, Host.CurrentCulture.Name, StringComparison.CurrentCultureIgnoreCase))
+                            {
+                                ci = new CultureInfo(cultureName, false);
```
SteveL-MSFT: Please use named params here

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
```diff
@@ -1,23 +1,123 @@
...
+        protected override void ProcessRecord()
+        {
+            CultureInfo ci;
+
+            switch (ParameterSetName)
+            {
+                case CurrentCultureParameterSet:
+                    if (NoUserOverrides)
+                    {
+                        ci = CultureInfo.GetCultureInfo(Host.CurrentCulture.LCID);
```
mklement0: Please use `.Name` instead of `.LCID`, and we should probably do the same for all tests.

> **"LCIDs are being deprecated, and implementers are strongly encouraged to use locale names instead.** LCIDs can be used for backward compatibility, but as noted in section 2.2.1, there is no guarantee of LCID uniqueness when used with valid locale names not otherwise associated with an LCID. " - from "[MS-LCID]: Windows Language Code Identifier (LCID) Reference" - https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-LCID/[MS-LCID].pdf

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
```diff
@@ -1,23 +1,123 @@
...
+
+                    WriteObject(ci);
+
+                    break;
+                case NameParameterSet:
+                    try
+                    {
+                        foreach (var cultureName in Name)
+                        {
+                            if (NoUserOverrides && string.Equals(cultureName, Host.CurrentCulture.Name, StringComparison.CurrentCultureIgnoreCase))
```
mklement0: This should be `!NoUserOverrides` and the 2nd `CultureInfo` constructor argument below should be `true` (or omitted), but you could just use `Host.CurrentCulture`, which also reflects the overrides.

mklement0 (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
iSazonov: Fixed.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
iSazonov: I think the change is not related to and not critical for the PR. We can postpone the change until CoreFX obsolete the API.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
iSazonov: Fixed.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Culture.Tests.ps1
iSazonov: Added the check and comment.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Culture.Tests.ps1
iSazonov: I think the order is right because the file is for Get-Culture tests and we should correct the description.
But I'd expect that we check $Host.CurrentCulture instead of `$PsCulture`.
On other side, we haven't tests for special variables at all.

What do you think?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
iSazonov: Added.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/GetCultureCommand.cs
mklement0: `[cultureinfo]::getcultures('allcultures') | group-object lcid | ? count -gt 1` will tell you that hundreds of predefined cultures share the same LCID, so this fix is an absolute must.

iSazonov: Fixed.

test/powershell/Modules/Microsoft.PowerShell.Utility/Get-Culture.Tests.ps1
SteveL-MSFT: I'm fine changing the test description to match the test and adding separate test for $PSCulture

iSazonov: Fixed.

iSazonov:
@SteveL-MSFT Please point me comments if I missed something.

SteveL-MSFT (approved):
LGTM

iSazonov:
@mklement0 Please update your review.

iSazonov:
Reopen to restart all CIs.

mklement0 (approved):

iSazonov:
@mklement0 Many thanks for great help!

mklement0:
My pleasure, @iSazonov, thanks for implementing it all.

PR #7721, opened 2018-09-06
Support calling method with ByRef-like type parameters

### LINKED ISSUE #7596: When calling method, we should be theoretically able to pass an argument to a ByRef-like parameter via implicit casting
The implicit cast to/from a ByRef-like type, such as `Span` or `ReadOnlySpan` doesn't help regular conversion operations in PowerShell because the target ByRef-like type cannot be boxed.

But when calling methods via Expression with jitting, `Expression.Convert` can be used to implicit cast an argument to the `Span` types before passing to the call and no boxing is involved (see an example below).

```c#
var arg = @"e:\abc\def";
var method = typeof(Path).GetMethod(nameof(Path.IsPathRooted), new Type[] { typeof(ReadOnlySpan<char>) });

var body = Expression.Call(method, Expression.Convert(Expression.Constant(arg), typeof(ReadOnlySpan<char>)));
var func = Expression.Lambda<Func<bool>>(body, null).Compile();
var rest = func();
Console.WriteLine(rest);

> True
```

PowerShell evaluates an Expression tree with interpretation by default, which basically translating Expression tree to pre-defined C# code, so the `Expression.Convert` might not work like when Expression tree is jitted (haven't looked into it, will investigate that). If it works with the interpreter too (or we can update the interpreter to make it work), we will also need to update the method resolution. We currently use the same "figuring-out-conversion" method for regular conversion in PowerShell as well as when resolving the best matching method. The implicit cast for ByRef-like target types should continue to be "no-conversion" for regular conversion operation, but theoretically acceptable for method resolution.

daxian-dbw:
## PR Summary

Fix #7596

Support calling methods with ByRef-like type parameters in PowerShell, as long as the argument specified for the parameter can be implicitly/explicitly converted to the ByRef-like type.

We cannot create an instance of a ByRef-like type in PowerShell, but there are types that can be implicitly or explicitly converted to ByRef-like types, such as `T[] -> Span<T>`, `T[] -> ReadOnlySpan<T>`, `String -> ReadOnlySpan<T>`. `ArraySegment<T> -> Span<T>`, `ArraySegment<T> -> ReadOnlySpan<T>`. With changes in this PR, we can call methods with ByRef-like parameter types by providing the arguments of the types that can be casted to the target ByRef-like type. For example:

```
PS:5> $charArray = "F:\".ToCharArray()
PS:6> [System.IO.Path]::IsPathRooted($charArray)
True
```

**What is enabled?**
1. Invoking methods that have ByRef-like parameters, but the return type is not ByRef-like.
2. Invoking constructors with ByRef-like parameters via `[target-type]::new` syntax, but the `target-type` is not ByRef-like.
2. Accessing indexers that have ByRef-like parameters, but the return type is not ByRef-like.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/CoreAdapter.cs
+++ b/src/System.Management.Automation/engine/CoreAdapter.cs
@@ -708,28 +708,46 @@ namespace System.Management.Automation
 
         #region Internal Helper Methods
 
-        private static Type GetArgumentType(object argument)
+        private static Type GetArgumentType(object argument, bool isByRefParameter)
         {
             if (argument == null)
             {
                 return typeof(LanguagePrimitives.Null);
             }
-            PSReference psref = argument as PSReference;
-            if (psref != null)
+
+            if (isByRefParameter && argument is PSReference psref)
             {
-                return GetArgumentType(PSObject.Base(psref.Value));
+                return GetArgumentType(PSObject.Base(psref.Value), isByRefParameter: false);
             }
+
             return argument.GetType();
         }
 
-        internal static ConversionRank GetArgumentConversionRank(object argument, Type parameterType)
+        internal static ConversionRank GetArgumentConversionRank(object argument, Type parameterType, bool isByRef, bool allowCastingToByRefLikeType)
         {
-            Type fromType = GetArgumentType(argument);
-            ConversionRank rank = LanguagePrimitives.GetConversionRank(fromType, parameterType);
+            Type fromType = null;
+            ConversionRank rank = ConversionRank.None;
+
+            if (allowCastingToByRefLikeType && parameterType.IsByRefLike)
+            {
+                // When resolving best method for use in binders, we can accept implicit/explicit casting conversions to
+                // a ByRef-like target type, because when generating IL from a call site with the binder, the IL includes
+                // the casting operation. However, we don't accept such conversions when it's for invoking the method via
+                // reflection, because reflection just doesn't support ByRef-like type.
+                fromType = GetArgumentType(PSObject.Base(argument), isByRefParameter: false);
+                if (fromType != typeof(LanguagePrimitives.Null))
+                {
+                    LanguagePrimitives.FigureCastConversion(fromType, parameterType, ref rank);
+                }
+                return rank;
+            }
+
+            fromType = GetArgumentType(argument, isByRef);
+            rank = LanguagePrimitives.GetConversionRank(fromType, parameterType);
 
             if (rank == ConversionRank.None)
             {
-                fromType = GetArgumentType(PSObject.Base(argument));
+                fromType = GetArgumentType(PSObject.Base(argument), isByRef);
                 rank = LanguagePrimitives.GetConversionRank(fromType, parameterType);
             }
 
@@ -1186,6 +1204,7 @@ namespace System.Management.Automation
         /// </summary>
         /// <param name="methods">different overloads for a method</param>
         /// <param name="invocationConstraints">invocation constraints</param>
+        /// <param name="allowCastingToByRefLikeType">true if we accept implicit/explicit casting conversion to a ByRef-like parameter type for method resolution</param>
         /// <param name="arguments">arguments to check against the overloads</param>
         /// <param name="errorId">if no best method, the error id to use in the error message</param>
         /// <param name="errorMsg">if no best method, the error message (format string) to use in the error message</param>
@@ -1194,6 +1213,7 @@ namespace System.Management.Automation
         internal static MethodInformation FindBestMethod(
             MethodInformation[] methods,
             PSMethodInvocationConstraints invocationConstraints,
+            bool allowCastingToByRefLikeType,
             object[] arguments,
             ref string errorId,
             ref string errorMsg,
@@ -1201,7 +1221,7 @@ namespace System.Management.Automation
             out bool callNonVirtually)
         {
             callNonVirtually = false;
-            var methodInfo = FindBestMethodImpl(methods, invocationConstraints, arguments, ref errorId, ref errorMsg, out expandParamsOnBest);
+            var methodInfo = FindBestMethodImpl(methods, invocationConstraints, allowCastingToByRefLikeType, arguments, ref errorId, ref errorMsg, out expandParamsOnBest);
             if (methodInfo == null)
             {
                 return null;
@@ -1243,6 +1263,7 @@ namespace System.Management.Automation
         private static MethodInformation FindBestMethodImpl(
             MethodInformation[] methods,
             PSMethodInvocationConstraints invocationConstraints,
+            bool allowCastingToByRefLikeType,
             object[] arguments,
             ref string errorId,
             ref string errorMsg,
@@ -1364,8 +1385,18 @@ namespace System.Management.Automation
                         Type elementType = parameter.parameterType.GetElementType();
                         if (parameters.Length == arguments.Length)
                         {
-                            ConversionRank arrayConv = GetArgumentConversionRank(arguments[j], parameter.parameterType);
-                            ConversionRank elemConv = GetArgumentConversionRank(arguments[j], elementType);
+                            ConversionRank arrayConv = GetArgumentConversionRank(
+                                arguments[j],
+                                parameter.parameterType,
+                                isByRef: false,
+                                allowCastingToByRefLikeType: false);
+
+                            ConversionRank elemConv = GetArgumentConversionRank(
+                                arguments[j],
+                                elementType,
+                                isByRef: false,
+                                allowCastingToByRefLikeType: false);
+
                             if (elemConv > arrayConv)
                             {
                                 candidate.expandedParameters = ExpandParameters(arguments.Length, parameters, elementType);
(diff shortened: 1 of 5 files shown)


test/powershell/Language/Interop/DotNet/DotNetInterop.Tests.ps1
```diff
@@ -180,7 +225,95 @@ namespace DotNetInterop
...
+        }
+
+        AfterAll {
+            $ps.Dispose()
+        }
+
+        It "Support method calls with ByRef-like parameter as long as the argument can be casted to the ByRef-like type" {
+            $testObj.PrintMySpan("abc", "def") | Should -BeExactly "abc3"
+            $testObj.PrintMySpan("abc", "Hello".ToCharArray()) | Should -BeExactly "abc5"
+            { $testObj.PrintMySpan("abc", 12) } | Should -Throw -ErrorId "MethodArgumentConversionInvalidCastArgument"
```
iSazonov: Ah, looks like bad UX. Moreover, now C# supports "value arguments by ref".

test/powershell/Language/Interop/DotNet/DotNetInterop.Tests.ps1
```diff
@@ -180,7 +225,95 @@ namespace DotNetInterop
...
+
+            $path = [System.IO.Path]::GetTempPath()
+            [System.IO.Path]::IsPathRooted($path.ToCharArray()) | Should -Be $true
+        }
+
+        It "Support constructor calls with ByRef-like parameter as long as the argument can be casted to the ByRef-like type" {
+            $result = [DotNetInterop.Test2]::new("abc")
+            $result.Name | Should -BeExactly "Number of chars: 3"
+
+            { [DotNetInterop.Test2]::new(12) } | Should -Throw -ErrorId "MethodCountCouldNotFindBest"
```
iSazonov: The same.

iSazonov (commented):
LGTM with one comment.

Also it might make sense to show in the tests that if we pass an argument by reference, then we can return its changed value.

test/powershell/Language/Interop/DotNet/DotNetInterop.Tests.ps1
daxian-dbw: > now C# supports "value arguments by ref".

Can you please elaborate that? Do you mean `MethodName(ref int id)`? If so, then powershell already supports it via `[ref]`.

test/powershell/Language/Interop/DotNet/DotNetInterop.Tests.ps1
daxian-dbw: `new(12)` is to use an integer to test the "not castable" scenario. What is your suggestion to change?

test/powershell/Language/Interop/DotNet/DotNetInterop.Tests.ps1
iSazonov: It seems I did not understand my idea myself:-)

PowerShell like to convert to/from string - `1 + "2"` returns `3`. So if we accept a string in the test I'd expect that we convert the int to string and `ReadOnlySpan<char>`. 
Perhaps there is more complex scenario. If a parameter type is Span<Int> I'd expect that we can able to convert the int to Span<Int> too and accept array of int-s.

If this is possible, you can do this in the next PR.

daxian-dbw: I don't think that's a good idea. Passing arguments to a ByRef-like parameter should be explicit, and I don't think we should do a second layer of conversion from the argument to the types that might be able to implicitly/explicitly cast to the target ByRef-like parameter type.

src/System.Management.Automation/engine/CoreAdapter.cs
```diff
@@ -708,28 +708,46 @@ internal string BaseParameterizedPropertyToString(PSParameterizedProperty proper
...
-            ConversionRank rank = LanguagePrimitives.GetConversionRank(fromType, parameterType);
+            Type fromType = null;
+            ConversionRank rank = ConversionRank.None;
+
+            if (allowCastingToByRefLikeType && parameterType.IsByRefLike)
+            {
+                // When resolving best method for use in binders, we can accept implicit/explicit casting conversions to
+                // a ByRef-like target type, because when generating IL from a call site with the binder, the IL includes
+                // the casting operation. However, we don't accept such conversions when it's for invoking the method via
+                // reflection, because reflection just doesn't support ByRef-like type.
```
lzybkr: Does PowerShell run on a platform without JIT support? In that case, the expression tree generated by the binder would be interpreted.

test/powershell/Language/Interop/DotNet/DotNetInterop.Tests.ps1
```diff
@@ -180,7 +225,95 @@ namespace DotNetInterop
...
+                </SetCodeReference>
+            </CodeProperty>
+        </Members>
+    </Type>
+</Types>
+'@
+            $typeXmlFile = Join-Path -Path $TestDrive -ChildPath "TestType.ps1xml"
+            Set-Content -Path $typeXmlFile -Value $typeXml -Encoding Ascii
+            $ps = [powershell]::Create()
+            $ps.AddCommand("Update-TypeData").AddParameter("AppendPath", $typeXmlFile).Invoke()
```
lzybkr: Update-TypeData can be used without creating a file if you'd prefer.

test/powershell/Language/Interop/DotNet/DotNetInterop.Tests.ps1
iSazonov: With the implicit cast in C# what is a difference `string` from `ReadOnlySpan<char>`? Currently CoreFX has overloads with string parameter as wrapper on methods with `ReadOnlySpan<char>`. We can free use any of the overloads - no difference in result.
Now let's look PowerShell behavior.
`function Test1([string] $var)` and `function Test1([int] $var)` both accepts `123` and `"123"`.
If they will be methods we see the same.
Then if it is a binary method with string parameter we see the same. But if it is `ReadOnlySpan<char>` we get an error.
Seems as minimum we should consider a string as special case.
As maximum perhaps we can discover that there is more common scenarios.

test/powershell/Language/Interop/DotNet/DotNetInterop.Tests.ps1
BrucePay: Indeed - much easier. Or is there some reason to use the file?

BrucePay (approved):
LGTM

test/powershell/Language/Interop/DotNet/DotNetInterop.Tests.ps1
daxian-dbw: The purpose of this PR is to make PowerShell able to interact with libraries that only exposes APIs with ByRef-like parameters. When there is an equivalent wrapper API, there is no point to call the ByRef-like parameter based API in PowerShell.

As for the conversion, `string` and `ReadOnlySpan<char>` are two different types with different semantics, and having implicit casing support between them doesn't change that. PowerShell doesn't support two-layer indirect conversion, for example, PowerShell supports `[FileInfo] -> [string]`, and `[string] -> [type]`, but `[FileInfo] -> [type]` will fail, even if the `FileInfo` object can be converted to a string value that can be converted to a type:
```
$a = [System.IO.FileInfo]::new("string")
$b = [string] $a
$b
> string

$c = [type] $b ## success
$d = [type] $a ## fail
Cannot convert the "string" value of type "System.IO.FileInfo" to type "System.Type".
At line:1 char:1
+ $d = [type] $a
+ ~~~~~~~~~~~~~~
+ FullyQualifiedErrorId : ConvertToFinalInvalidCastException
```

src/System.Management.Automation/engine/CoreAdapter.cs
daxian-dbw: I think .NET Core is a runtime that always has JIT support. .NET Core framework comes with the reflection emit APIs, which depends on JIT support to work.

test/powershell/Language/Interop/DotNet/DotNetInterop.Tests.ps1
daxian-dbw: No particular reason, I just went with the file without considering other options ;) I will change to simply the test.

src/System.Management.Automation/engine/CoreAdapter.cs
lzybkr: In addition to the expression interpreter, there is an IL interpreter, so I guess it's possible to have no JIT but also support reflection emit.

test/powershell/Language/Interop/DotNet/DotNetInterop.Tests.ps1
iSazonov: Well, then it's best to explicitly indicate in the documentation why this does not work otherwise there will be a lot of questions.

test/powershell/Language/Interop/DotNet/DotNetInterop.Tests.ps1
daxian-dbw: Updated to use `Update-TypeData -TypeName -MemberType ...`

daxian-dbw:
The Travis-CI builds have been successful, but somehow the status was not reflected in the checks below. Here are the CI builds:
https://travis-ci.org/PowerShell/PowerShell/builds/426886335

iSazonov:
@daxian-dbw Thanks for the very useful fix!

Issue #2230, opened 2016-09-10
Expression interpreter has received a lot of improvements in CoreFx


bartdesmet:
As a heads-up, the expression interpreter under `src\System.Management.Automation\engine\interpreter` seems to be a clone from the original code in the DLR. Many improvements to it were made in the CoreFx repo at http://github.com/dotnet/corefx. It'd be good to consider either porting these changes or eliminating the copy altogether once the `Compile(bool preferInterpretation)` overload is available on all platforms by PowerShell. Note that the regular `Compile()` overload will be available on all platforms that support `System.Linq.Expressions` and will use the interpreter in case the compiler is not available.

PR #7729, opened 2018-09-07
Build and package framework dependent package

adityapatwardhan:
## PR Summary

Add the functionality to build a framework dependent (shared framework) package for PowerShell.

The changes create two packages, one for Windows and other for Linux, due to `#if` code.

## PR Checklist

- [X] [PR has a meaningful title](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
    - Use the present tense and imperative mood when describing your changes
- [X] [Summarized changes](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [X] [Change is not breaking](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#making-breaking-changes)
- [X] [Make sure all `.h`, `.cpp`, `.cs`, `.ps1` and `.psm1` files have the correct copyright header](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
- [x] This PR is ready to merge and is not [Work in Progress](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---work-in-progress).
    - If the PR is work in progress, please add the prefix `WIP:` to the beginning of the title and remove the prefix when the PR is ready.
- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
    - [ ] User-facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
        - [ ] Issue filed - Issue link:
- **Testing - New and feature**
    - [ ] Not Applicable or can only be tested interactively
    - **OR**
    - [ ] [Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
        - [ ] [Add `[feature]` if the change is significant or affects feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

PR #7731, opened 2018-09-07
WIP: Use new Compile() overload in CompileTree()

iSazonov:
## PR Summary

Related #2230.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/PowerShell.Common.props
+++ b/PowerShell.Common.props
@@ -96,6 +96,7 @@
     <TargetFramework>netcoreapp2.1</TargetFramework>
     <RuntimeFrameworkVersion>2.1.3</RuntimeFrameworkVersion>
     <LangVersion>Latest</LangVersion>
+    <TieredCompilation>true</TieredCompilation>
 
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
--- a/src/System.Management.Automation/engine/Utils.cs
+++ b/src/System.Management.Automation/engine/Utils.cs
@@ -1395,6 +1395,8 @@ namespace System.Management.Automation.Internal
 
         internal static bool ShowMarkdownOutputBypass;
 
+        internal static bool ExpressionCompile;
+
         /// <summary>This member is used for internal test purposes.</summary>
         public static void SetTestHook(string property, object value)
         {
--- a/src/System.Management.Automation/engine/parser/Compiler.cs
+++ b/src/System.Management.Automation/engine/parser/Compiler.cs
@@ -1780,12 +1780,7 @@ namespace System.Management.Automation.Language
                 default:
                     // Here compileInterpretChoice == CompileInterpretChoice.CompileOnDemand
                     //
-                    // compilationThreshold is # of times the script must run before we decide to compile
-                    // NeverCompile sets the threshold to int.MaxValue, so theoretically we might compile
-                    // at some point, but it's very unlikely.
-                    // -1 is to use default (32).
-                    var deleg = new LightCompiler(compilationThreshold: -1).CompileTop(lambda).CreateDelegate();
-                    return (Action<FunctionContext>)deleg;
+                    return lambda.Compile(preferInterpretation: false);
             }
         }
 
--- a/src/System.Management.Automation/engine/parser/Compiler.cs
+++ b/src/System.Management.Automation/engine/parser/Compiler.cs
@@ -1780,7 +1780,14 @@ namespace System.Management.Automation.Language
                 default:
                     // Here compileInterpretChoice == CompileInterpretChoice.CompileOnDemand
                     //
-                    return lambda.Compile(preferInterpretation: false);
+                    if (Internal.InternalTestHooks.ExpressionCompile)
+                    {
+                        return lambda.Compile(preferInterpretation: false);
+                    }
+                    else
+                    {
+                        return lambda.Compile(preferInterpretation: true);
+                    }
             }
         }
 
--- a/src/System.Management.Automation/engine/parser/Compiler.cs
+++ b/src/System.Management.Automation/engine/parser/Compiler.cs
@@ -1535,7 +1535,7 @@ namespace System.Management.Automation.Language
 
                 if (attribute is ExperimentalAttribute expAttribute)
                 {
-                    // Only honor the first seen experimental attribute, ignore the others. 
+                    // Only honor the first seen experimental attribute, ignore the others.
                     if (!hasSeenExpAttribute && expAttribute.ToHide) { return null; }
 
                     // Do not add experimental attributes to the attribute list.
@@ -1763,19 +1763,30 @@ namespace System.Management.Automation.Language
         private Action<FunctionContext> CompileTree(Expression<Action<FunctionContext>> lambda, CompileInterpretChoice compileInterpretChoice)
         {
             if (lambda == null)
-                return null;
-
-            if (compileInterpretChoice == CompileInterpretChoice.AlwaysCompile)
             {
-                return lambda.Compile();
+                return null;
             }
 
-            // threshold is # of times the script must run before we decide to compile
-            // NeverCompile sets the threshold to int.MaxValue, so theoretically we might compile
-            // at some point, but it's very unlikely.
-            int threshold = (compileInterpretChoice == CompileInterpretChoice.NeverCompile) ? int.MaxValue : -1;
-            var deleg = new LightCompiler(threshold).CompileTop(lambda).CreateDelegate();
-            return (Action<FunctionContext>)deleg;
+            switch (compileInterpretChoice)
+            {
+                case CompileInterpretChoice.AlwaysCompile:
+
+                    return lambda.Compile(preferInterpretation: false);
+
+                case CompileInterpretChoice.NeverCompile:
+
+                    return lambda.Compile(preferInterpretation: true);
+
+                default:
+                    // Here compileInterpretChoice == CompileInterpretChoice.CompileOnDemand
+                    //
+                    // compilationThreshold is # of times the script must run before we decide to compile
+                    // NeverCompile sets the threshold to int.MaxValue, so theoretically we might compile
+                    // at some point, but it's very unlikely.
+                    // -1 is to use default (32).
+                    var deleg = new LightCompiler(compilationThreshold: -1).CompileTop(lambda).CreateDelegate();
+                    return (Action<FunctionContext>)deleg;
+            }
         }
 
         internal static object GetExpressionValue(ExpressionAst expressionAst, bool isTrustedInput, ExecutionContext context, IDictionary usingValues = null)


lzybkr:
I'm not sure this is a good idea - why we would want to use 2 different interpreters?

iSazonov:
I think that it does not make sense to port the code from CoreFX here and it's better to migrate to CoreFX interpreter. I see that it doesn't support Compile overload with threshold. So first step is to use `Compile(preferInterpretation: true);`.
The second step is in question. We could use tiered compilation. 
https://blogs.msdn.microsoft.com/dotnet/2018/08/02/tiered-compilation-preview-in-net-core-2-1/
There already is TieredCompilation_Tier1CallCountThreshold = 30
https://github.com/dotnet/coreclr/blob/f6174b93d100d46f4641f040b6de5fa254c1ee71/Documentation/project-docs/clr-configuration-knobs.md

iSazonov:
I see #7729 - grossgen doesn't work with framework-dependent deployment and tired compilation will come in handy.

iSazonov:
I checked with PerfView that TC works. Seems the full set of tests is performed on CIs about the same time.

lzybkr:
Why do you think it's better to migrate? Do you have data that shows it's faster?

It does not jit compile, so loops will be ~50X slower.

daxian-dbw:
@iSazonov The interpreter in PowerShell was updated to act like the tiered compilation. When the script block contains less than 300 statements, the delegate created by the interpreter is able to generate a compiled delegate on demand to replace itself, when the script block or loops in it runs for a certain number of times.

The tiered compilation cannot replace this optimization. The tiered compilation will optimize the `Run` methods from some instructions, but no matter how JIT is able to optimize those individual methods, the script is still being evaluated in an interpreted way -- fetch an instruction, data gets pushed to a stack in the interpreter, run some C# code, pop the data, save to local variable list, fetch the next instruction, etc. However, after a compiled delegate gets created on demand, the script will be running directly in the jitted native code, and it's possible for the compiled delegate to further benefit from the tiered compilation to be even more optimial.

You can see it as a 3-tiered compilation.

lzybkr:
Are you certain your experiment is valid? It's  possible you only execute one code path because of caching.

iSazonov:
@lzybkr What cache do you mean? I added a test hook to explicitly switch from interpreter (CoreFX) to compiler.

lzybkr:
PowerShell caches script block definitions to avoid recompiling, so I was just asking that you confirm your experiment hits the code paths you expect.

I measured a 50X slowdown when switching to the CoreFx interpreter and this did not surprise me because their interpreter no longer supports JIT.

Your results do surprise me - if JIT is happening at some point I'm happy, but I'd like pointers to where that happens or at least a solid explanation of what magic is making the new interpreter faster.

iSazonov:
I run the test script in interactive session by manually copy-paste and use debugger to confirm that Compile(true)/Compile(false) is called after each the copy-paste. 
I tied to run the script in cycle and get other results (many times faster). There I guess was a cache.
I am also surprised by these results. So far I'm inclined not to trust myself. Maybe I'm doing something wrong.

daxian-dbw:
The measurement is too specific and doesn't reflect real scenarios, here are the reasons:
1. The testing script contains minimal statements, so the cost for JIT compiling is low and it doesn't reflect the cost you would get for a real scenario script which contains relatively many statements.

2. When there are too many statements in a script block, it would be too expensive to JIT compile it (cause very slow startup). The tradeoff PowerShell takes is to [`NeverCompile`](https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/parser/Compiler.cs#L1733) the script block in that case. With the current `CompileOnDemand` policy, even though the script block as a whole will never be compiled, the loops in it will still be JIT compiled after running for certian times, as long as the statements in them are not too many. However, with your change, nothing will be JIT compiled in that case, and the performance will for sure decrease comparied to the current `CompileOnDemand`.

And BTW, I guess tiered compilation was turned on in your local builds when compared with PS 6.1-RC. That would be another factor that changes the numbers you get from the measurements, even though I don't know how much difference that would make.

iSazonov:
Yes, tiered compilation was turned. Now I tested without it and see ~20%  decrease in performance in the scenario.

My main concern was that the interpreter would be much faster in the interactive session, but this fear did not materialize. It seems we could use the compilation of even small scripts in an interactive session. We can even get some benefits because even one line script can be an CPU expensive cycle and compile + tiered compilation seem to bring improvements.

Next test I did for Parser.
```powershell
$text = ""
foreach ($file in dir -Recurse -Path C:\Users\sie\Documents\GitHub\iSazonov\PowerShell\test\powershell\ -Filter "*.ps1") {
    $text = dir -Recurse -Path C:\Users\sie\Documents\GitHub\iSazonov\PowerShell\test\powershell\ -Filter "*.ps1" | Get-Content -Raw 
}


Invoke-Command  -ScriptBlock { for ($j = 0; $j -lt 100; $j++) {

$step3 = measure-command {
foreach ($t in $text) {
    $tokens = $null
    $errors = $null
    [Management.Automation.Language.Parser]::ParseInput($t, [ref]$tokens, [ref]$errors) | Out-Null
}
}
$step3.Milliseconds

}} | Measure-Object -Average
```
Results 659 ms for RC1 and 662 ms for TC build - slower ~0.5%.
(I should note that the variance of th TC build is greater in al tests.)

In the scenario it seems TC does not give any advantages, but it's more likely that the parser is a high-quality code.

iSazonov:
Test for small script:
```powershell
[System.Management.Automation.Internal.InternalTestHooks]::SetTestHook("ExpressionCompile", $false)

$step4a = measure-command {

for ($j1 = 0l; $j1 -lt 30000; $j1++) {

Invoke-Command  -ScriptBlock {

$no = $false
if ($no) {
    # 1
    $a += $a + 1
    $b -= $b - 1
    $c *= $c * 1Mb * 1Kb
    $d /= $d / 1Tb
    $e = "a,b,c,d" -split ","
    $f = "a", "b", "c", "d" -join ";"
    $g = New-Guid
    $i = [Math]::Max(1234567890, 12345678901234567890)
    $j = (1, 2, 3, 4, 5, 6, 7, 8, 9, 0)[9]
    $k = 1, 2, 3, 4, 5, 6, 7, 8, 9, 0 | Measure-Object
}

} # end Invoke-Command

} # end for

} # end measure-command


[System.Management.Automation.Internal.InternalTestHooks]::SetTestHook("ExpressionCompile", $true)

$step4b = measure-command {

for ($j1 = 0l; $j1 -lt 30000; $j1++) {

Invoke-Command  -ScriptBlock {

$no = $false
if ($no) {
    # 1
    $a += $a + 1
    $b -= $b - 1
    $c *= $c * 1Mb * 1Kb
    $d /= $d / 1Tb
    $e = "a,b,c,d" -split ","
    $f = "a", "b", "c", "d" -join ";"
    $g = New-Guid
    $i = [Math]::Max(1234567890, 12345678901234567890)
    $j = (1, 2, 3, 4, 5, 6, 7, 8, 9, 0)[9]
    $k = 1, 2, 3, 4, 5, 6, 7, 8, 9, 0 | Measure-Object
}

} # end Invoke-Command

} # end for

} # end measure-command


$step4a.Milliseconds
$step4b.Milliseconds
```
Results:

| Iterations | RC1 | Crossgened Interpretator | Crossgened Compile |
| :-: | :-: | :-: | :-: |
| 30 | 10 | 11 | 20 |
| 300 | 30 | 38 | 36 |
| 3000 | 189 | 270 | 200 |
| 30000 | 820 | 590 | 820 |

This result shows that even on RC1 we do possibly not need to compile small scripts. Perhaps this is due to a change between .Net Core and Framework.

iSazonov:
Another test:
```powershell
[System.Management.Automation.Internal.InternalTestHooks]::SetTestHook("ExpressionCompile", $false)
$step1 = measure-command {
for ($ii = 0l; $ii -lt 300; $ii++) {
    $a+=1
    $a += $a + 1
    $b -= $b - 1
    $c *= $c * 1Mb * 1Kb
    $d = 1;$d /= $d / 1Tb
    $e = "a,b,c,d" -split ","
    $f = "a", "b", "c", "d" -join ";"
    $g = New-Guid
    $i = [Math]::Max(123, 1234)
    $j = (1, 2, 3, 4, 5, 6, 7, 8, 9, 0)[9]
    $k = 1, 2, 3, 4, 5, 6, 7, 8, 9, 0 | Measure-Object}
}


[System.Management.Automation.Internal.InternalTestHooks]::SetTestHook("ExpressionCompile", $true)
$step2 = measure-command {
for ($ii = 0l; $ii -lt 300; $ii++) {
    $a+=1
    $a += $a + 1
    $b -= $b - 1
    $c *= $c * 1Mb * 1Kb
    $d = 1;$d /= $d / 1Tb
    $e = "a,b,c,d" -split ","
    $f = "a", "b", "c", "d" -join ";"
    $g = New-Guid
    $i = [Math]::Max(123, 1234)
    $j = (1, 2, 3, 4, 5, 6, 7, 8, 9, 0)[9]
    $k = 1, 2, 3, 4, 5, 6, 7, 8, 9, 0 | Measure-Object}
}

$step1.TotalMilliseconds
$step2.TotalMilliseconds

```
Results:

| Iterations | RC1 | Crossgened Interpretator | Crossgened Compile |
| :-: | :-: | :-: | :-: |
| 30 | 11 | 23 | 19 |
| 300 | 50 | 78 | 53 |
| 3000 | 386 | 594 | 411 |
| 5000 | 634 | 978 | 670 |
| 10000 | 1256 | 1923 | 1299 |
| 20000 | 2534 | 3684 | 2522 |
| 30000 | 3747 | 5560 | 3819 |
```

iSazonov:
I updated previous test results bacause of a bug in test script.
Now I see that most likely neither compile nor CoreFX interpreter will not give an improvement.
On the other hand, TC gives an improvement in performance.

iSazonov:
.Net Core team announced .Net Core 2.2.0 Preview2 with TC enabled by default.
https://blogs.msdn.microsoft.com/dotnet/2018/09/12/announcing-net-core-2-2-preview-2/

We should definitely continue to investigate the effect of TC on PowerShell Core.

daxian-dbw:
@iSazonov I asked for details of the measurement, and it turned out the measurement was taken with PSCore 6.0 code base built with `release` configuration without crossgen. PowerShell was used as a Fx Dependent application in the measurement and executed by `dotnet .\bin\release\netcoreapp2.0\win7-x64\pwsh.dll -command exit`.

So, the result of the measurements indicates that with a snapshot of the code base at 6.0 timeframe, the perf improvement in .NET Core 2.1 runtime and tiered compilation combined together offer a 20% startup improvement in the Fx Dependent scenario.

However, we got some degradation in startup time in the 6.1 timeframe, and the rough sources include `TaskbarJumpList`, `Experimental Feautre Flag (configuration file access)` and more. From my naive measurement, without tiered compilation, 6.1 is about 7% slower than 6.0 in startup time (crossgen'ed), and after turning on tiered compilation, 6.1 is about the same as 6.0. I'm looking in to the degradation.

iSazonov:
@daxian-dbw Thanks! It is interesting!

> TaskbarJumpList

I think we could move this to install phase (to msi custom action).

iSazonov:
Make sense ask CoreFX team to implement `LightCompiler(threshold) ( 3-tiered compilation)`?

lzybkr:
You can certainly ask, though it might be enough to add support for custom instructions in the interpreter.

Today, PowerShell loops are implemented as a custom instruction which switches to the jit compiled version of the loop after sufficient iterations. Our core interpreter works similarly, but PowerShell loops were just special enough that it was a little easier to create a new instruction.

If CoreFX allowed this sort of extension, both loops and entire functions could implement their own tiered compilation strategy.

PR #7742, opened 2018-09-08
WIP: Change WMF4 pre-req check in MSI installer to a warning

SteveL-MSFT:
## PR Summary

Currently, the MSI installer does a check to see if version of pwrshplugin.dll is compatible with PSCore6.  This requires WMF4 or newer.  If the check fails, the installer fails.  However, there are valid use cases where the user may not care about WinRM based remoting.  Fix is to change this to a warning message and also disable the checkbox to enable PSRemoting.

Screenshot

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/assets/Product.wxs
+++ b/assets/Product.wxs
@@ -251,7 +251,7 @@
         </Condition>
       </Control>
       <Control Id="ContextMenuOpenPowerShell" Type="CheckBox" X="20" Y="120" Width="290" Height="17" Property="ADD_EXPLORER_CONTEXT_MENU_OPENPOWERSHELL" CheckBoxValue="0" Text="Add '$(var.ExplorerContextSubMenuDialogText)' context menus to Explorer"/>
-      <Control Id="WMFWarning" Type="Text" X="20" Y= "140" Width="290" Height="32" Transparent="yes" NoPrefix="yes" Text="{\WMFWarningFont}Windows Management Framework 4.0 or newer is required enable remoting over WinRM. See: https://aka.ms/pscore6-prereq">
+      <Control Id="WMFWarning" Type="Text" X="20" Y= "140" Width="290" Height="32" Transparent="yes" NoPrefix="yes" Text="{\WMFWarningFont}Windows Management Framework 4.0 or newer is required to enable remoting over WinRM. See: https://aka.ms/pscore6-prereq">
         <Condition Action="hide">
             <![CDATA[PWRSHPLUGIN_VERSION]]>
         </Condition>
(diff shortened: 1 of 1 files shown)


SteveL-MSFT:
cc @bergmeister to help review

alexandair:
Warning message:
"...required to enable PowerShell remoting over..."

iSazonov:
Could we make the link clickable?

bergmeister (approved):
Looks OK from a WiX perspective but I am wondering if it would be better to move the checkbox to the bottom to allow for alignment of the text with the checkbox because to an end user it is not obvious that the checkbox is disabled due to that text. Also, if you consider people who are not interested in remoting: They will get alerted due to the red text and waste time reading and thinking about the text. From a UX perspective, popping a dialog box when attempting to check the EnableRemoting checkbox would be an alternative.

bergmeister:
@iSazonov I remember getting hyperlinks to work in the past, which was not possible, but [this](https://stackoverflow.com/a/45813037/1810304) Thread indicates that WiX 1.11 added this capability.

iSazonov:
If the PR is targeted on 6.1.0 GA I believe we should merge "as is". Any other enhancement carries a risk to break something.

SteveL-MSFT:
@iSazonov this is not required for GA.  We can take it for 6.2.  Anyone who wants to install PSCore6 and not WMF4 on downlevel can use the .zip pkg today.

iSazonov:
Then I suggest to insert pre-request check step on first place and accumulate all pre-request checks there like WMF4, libssl and so on. Results can be in text log. With a button to show. With blocking Next if the checks failed.

anmenaga:
I restarted 'PowerShell-CI-linux' test run.

SteveL-MSFT:
@iSazonov that's a good suggestion, we can use a warning icon for warnings and a red X for any blocking pre-reqs.  Let me think more about this after 6.1 GA.

iSazonov:
@SteveL-MSFT I know Exchenge Server team use that in their installer.

iSazonov:
We could look WiX Bootstrapper.

SteveL-MSFT:
WiX Bootstrapper appears to help install dependencies.  However, it seems there's a set of users who want to use PSCore6 w/o installing WMF4 so I don't think the Bootstrapper will help here.  It would help with other dependencies that are required for .Net Core.

iSazonov:
> there's a set of users who want to use PSCore6 w/o installing WMF4

Is this supported? I think such users should install from zip, yes?

SteveL-MSFT:
@iSazonov actually, that is an excellent point!  I agree that non-standard use should use the zip.  So in that case, we should look at Bootstrapper to help with pre-reqs.  I'm closing this PR and opened issue https://github.com/PowerShell/PowerShell/issues/8176

SteveL-MSFT:
@bergmeister that is a good point about offline installs.  The Bootstrapper has capability to check for files before doing an install of a pre-req.  This means that if the customer has pre-installed the pre-reqs, then the Bootstrapper becomes a no-op.  It seems like the zip is sufficient for edge cases like where the user doesn't care about remoting.  If this is really a mainstream case and not an edge case, we can certainly revisit this.

PR #7819, opened 2018-09-19
Fix race condition in tests

iSazonov:
## PR Summary

Last days the tests often failed due to race condition.
See https://powershell.visualstudio.com/PowerShell/_build/results?buildId=2022&view=logs

Fix is to use test hook (write verbose message at cmdlet start) and Wait-UntilTrue function.

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
@@ -109,14 +109,14 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         protected override void ProcessRecord()
         {
-            if (InternalTestHooks.ActivateSleepForStoppingTest)
+            if (InternalTestHooks.ActivateSleepForStoppingTest != 0)
             {
                 // We'll wait the verbose stream marker in tests and then call Stop().
                 // The cmdlet is very fast and likely to end before the test sees the marker
                 // so we have to slow down it by Sleep(50).
                 // We can not sleep for a long time because Stop() doesn't work on sleeping runspace.
                 WriteVerbose("ConvertTo-Json started");
-                System.Threading.Thread.Sleep(50);
+                System.Threading.Thread.Sleep(InternalTestHooks.ActivateSleepForStoppingTest);
             }
 
             if (InputObject != null)
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
@@ -109,6 +109,12 @@ namespace Microsoft.PowerShell.Commands
         /// </summary>
         protected override void ProcessRecord()
         {
+            if (InternalTestHooks.ActivateSleepForStoppingTest)
+            {
+                WriteVerbose("ConvertTo-Json started");
+                System.Threading.Thread.Sleep(50);
+            }
+
             if (InputObject != null)
             {
                 _inputObjects.Add(InputObject);
--- a/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
+++ b/src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
@@ -111,6 +111,10 @@ namespace Microsoft.PowerShell.Commands
         {
             if (InternalTestHooks.ActivateSleepForStoppingTest)
             {
+                // We'll wait the verbose stream marker in tests and then call Stop().
+                // The cmdlet is very fast and likely to end before the test sees the marker
+                // so we have to slow down it by Sleep(50).
+                // We can not sleep for a long time because Stop() doesn't work on sleeping runspace.
                 WriteVerbose("ConvertTo-Json started");
                 System.Threading.Thread.Sleep(50);
             }
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Xml.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Xml.Tests.ps1
@@ -59,19 +59,7 @@ Describe "ConvertTo-Xml DRT Unit Tests" -Tags "CI" {
     }
 
     It "StopProcessing should work" {
-        try {
-            $ps = [PowerShell]::Create()
-            $null = $ps.AddScript({ Get-Process | ConvertTo-Xml -Depth 2 -Verbose })
-            [System.Management.Automation.Internal.InternalTestHooks]::SetTestHook('ActivateSleepForStoppingTest', $true)
-            $null = $ps.BeginInvoke()
-            Wait-UntilTrue { $ps.Streams.Verbose.Count -gt 0 } -IntervalInMilliseconds 50
-            $null = $ps.BeginStop($null, $null)
-            Wait-UntilTrue { $ps.InvocationStateInfo.State -eq "Stopped" } -IntervalInMilliseconds 50
-            $ps.InvocationStateInfo.State | Should -BeExactly "Stopped"
-        } finally {
-            $ps.Dispose()
-            [System.Management.Automation.Internal.InternalTestHooks]::SetTestHook('ActivateSleepForStoppingTest', $false)
-        }
+        Test-Stopping { Get-Process | ConvertTo-Xml -Depth 2 -Verbose } -IntervalInMilliseconds 50
     }
 
     # these tests just cover aspects that aren't normally exercised being used as a cmdlet
--- a/test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Xml.Tests.ps1
+++ b/test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Xml.Tests.ps1
@@ -59,12 +59,18 @@ Describe "ConvertTo-Xml DRT Unit Tests" -Tags "CI" {
     }
 
     It "StopProcessing should work" {
-        $ps = [PowerShell]::Create()
-        $null = $ps.AddScript({ Get-Process | ConvertTo-Xml -Depth 2 -Verbose })
-        $null = $ps.BeginInvoke()
-        Wait-UntilTrue { $ps.Streams.Verbose.Count -gt 0 } -IntervalInMilliseconds 50
-        $null = $ps.Stop()
-        $ps.InvocationStateInfo.State | Should -BeExactly "Stopped"
+        try {
+            $ps = [PowerShell]::Create()
+            $null = $ps.AddScript({ Get-Process | ConvertTo-Xml -Depth 2 -Verbose })
+            [System.Management.Automation.Internal.InternalTestHooks]::SetTestHook('ActivateSleepForStoppingTest', $true)
+            $null = $ps.BeginInvoke()
+            Wait-UntilTrue { $ps.Streams.Verbose.Count -gt 0 } -IntervalInMilliseconds 50
+            $null = $ps.Stop()
+            $ps.InvocationStateInfo.State | Should -BeExactly "Stopped"
+        } finally {
+            $ps.Dispose()
+            [System.Management.Automation.Internal.InternalTestHooks]::SetTestHook('ActivateSleepForStoppingTest', $false)
+        }
     }
 
     # these tests just cover aspects that aren't normally exercised being used as a cmdlet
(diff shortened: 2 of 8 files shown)


test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Xml.Tests.ps1
```diff
@@ -59,22 +59,21 @@ Describe "ConvertTo-Xml DRT Unit Tests" -Tags "CI" {
...
-		$ps.AddParameter("Depth", 2)
-		$ps.BeginInvoke()
-		$ps.Stop()
-		$ps.InvocationStateInfo.State | Should -BeExactly "Stopped"
+        $ps = [PowerShell]::Create()
+        $null = $ps.AddScript({ Get-Process | ConvertTo-Xml -Depth 2 -Verbose })
+        $null = $ps.BeginInvoke()
+        Wait-UntilTrue { $ps.Streams.Verbose.Count -gt 0 } -IntervalInMilliseconds 50
+        $null = $ps.Stop()
+        $ps.InvocationStateInfo.State | Should -BeExactly "Stopped"
```
TravisEz13: these is no guarantee that this will take a significant about of time.  It may finish before you stop it.

TravisEz13: @JamesWTruher  Can you review?

test/powershell/Modules/Microsoft.PowerShell.Utility/XMLCommand.Tests.ps1
```diff
@@ -24,139 +24,141 @@ Describe "XmlCommand DRT basic functionality Tests" -Tags "CI" {
...
         $ps = [PowerShell]::Create()
         $null = $ps.AddScript("1..10")
         $null = $ps.AddCommand("foreach-object")
         $null = $ps.AddParameter("Process", { $_; start-sleep 1 })
         $null = $ps.AddCommand("Export-CliXml")
         $null = $ps.AddParameter("Path", $testfile)
+        $null = $ps.AddParameter("Verbose")
         $null = $ps.BeginInvoke()
-        Start-Sleep 1
+        Wait-UntilTrue { $ps.Streams.Verbose.Count -gt 0 } -IntervalInMilliseconds 50
```
TravisEz13: I have similar concerns here about test reliability

test/powershell/Modules/Microsoft.PowerShell.Utility/ConvertTo-Xml.Tests.ps1
iSazonov: > these is no guarantee 

Yes, I'd remove the tests - I do not see a way to make these tests reliable. If we just add a hook with an infinite loop...

SteveL-MSFT: The intent of the test is to cover the StopProcessing code path as identified by code coverage.  I wonder if we can just have at test hook in the cmdlet so that ProcessRecord() doesn't return if the test hook is in place.  Therefore we can call `Stop()` so that the StopProcessing code path gets exercised?

TravisEz13: Seems like the implementation would be for `ProcessRecord` to Sleep for enough time that we are certain enough that we can stop it.  This could also mean that we can get one object and process it instead of trying to process a lot of records trying to make it take a long time.

iSazonov: Test hook added.
Sleep suspend current thread. So I used only short sleep to slow down a cmdlet.

iSazonov:
Reopen to restart Ci Appveyor.

iSazonov:
Reopen to restart Ci Appveyor.

TravisEz13:
@iSazonov Do you have a Microsoft account?  I can give you permissions to restart the Azure Pipelines if you give me a Microsoft account.  feel free to send it to Steve or my work email.

iSazonov:
@TravisEz13 I sent.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -100,6 +100,12 @@ protected override void BeginProcessing()
         /// </summary>
         protected override void ProcessRecord()
         {
+            if (InternalTestHooks.ActivateSleepForStoppingTest)
+            {
+                WriteVerbose("ConvertTo-Json started");
+                System.Threading.Thread.Sleep(50);
```
SteveL-MSFT: I don't think this sleep is needed.  There was a concern that the cmdlet would finish successfully before the stop of the pipeline got issued.  I think they way to solve that is to have another check at the end of ProcessRecord() where if the test hook is enabled, just have it sleep and not finish.

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
iSazonov: Stopping doesn't work on sleeping runspace. So idea is only slow down the cycle and make many interations.

SteveL-MSFT: Ok, perhaps add a comment?

SteveL-MSFT (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
iSazonov: Done.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -100,6 +100,16 @@ protected override void BeginProcessing()
...
         protected override void ProcessRecord()
         {
+            if (InternalTestHooks.ActivateSleepForStoppingTest)
+            {
+                // We'll wait the verbose stream marker in tests and then call Stop().
+                // The cmdlet is very fast and likely to end before the test sees the marker
+                // so we have to slow down it by Sleep(50).
+                // We can not sleep for a long time because Stop() doesn't work on sleeping runspace.
+                WriteVerbose("ConvertTo-Json started");
+                System.Threading.Thread.Sleep(50);
```
JamesWTruher: I would be happier if this value was in the testhook (ala TestStopComputerResults). This would enable us to change the duration easily, and even specific to a particular test.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
```diff
@@ -100,6 +100,16 @@ protected override void BeginProcessing()
         /// </summary>
         protected override void ProcessRecord()
         {
+            if (InternalTestHooks.ActivateSleepForStoppingTest)
+            {
+                // We'll wait the verbose stream marker in tests and then call Stop().
+                // The cmdlet is very fast and likely to end before the test sees the marker
+                // so we have to slow down it by Sleep(50).
+                // We can not sleep for a long time because Stop() doesn't work on sleeping runspace.
+                WriteVerbose("ConvertTo-Json started");
```
JamesWTruher: what are you trying to achieve with the WriteVerbose? I'm not sure there's much value here and I would rather reduce the chatty output.

test/tools/Modules/HelpersCommon/HelpersCommon.psm1
```diff
@@ -312,4 +313,29 @@ function Start-NativeExecution
...
+    try {
+        $ps = [PowerShell]::Create()
+        $null = $ps.AddScript($sb)
+        Enable-Testhook 'ActivateSleepForStoppingTest'
+        $null = $ps.BeginInvoke()
+        Wait-UntilTrue { $ps.Streams.Verbose.Count -gt 0 } -TimeoutInMilliseconds $TimeoutInMilliseconds -IntervalInMilliseconds $IntervalInMilliseconds
+        $null = $ps.BeginStop($null, $null)
+        Wait-UntilTrue { $ps.InvocationStateInfo.State -eq "Stopped" } -TimeoutInMilliseconds $TimeoutInMilliseconds -IntervalInMilliseconds $IntervalInMilliseconds
+
+        $ps.InvocationStateInfo.State | Should -Be "Stopped"
```
JamesWTruher: As this is a general purpose helper, I'm wondering whether this should be `| Should -Not -Be "Running"`
some of the states might be "failed" or "disconnected".

JamesWTruher (commented):
Generally, I'm in favor of more test hooks, but I'm not a fan of extra verbosity. My only real objection is the hard coding of the timeout duration. I think that should be testhook supplied value.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
SteveL-MSFT: The verbose message here is used to tell the test that the cmdlet has started so that it can issue the stop.  Otherwise there's a window where stop is issued before the cmdlet under test has started.

iSazonov: Yes, we need an event to start the stopping.

test/tools/Modules/HelpersCommon/HelpersCommon.psm1
iSazonov: If the stopping doesn't work we can get "Completed".

src/Microsoft.PowerShell.Commands.Utility/commands/utility/WebCmdlet/ConvertToJsonCommand.cs
iSazonov: Done.

iSazonov:
@SteveL-MSFT Please update your review.

test/powershell/Modules/Microsoft.PowerShell.Utility/XMLCommand.Tests.ps1
```diff
@@ -24,139 +24,125 @@ Describe "XmlCommand DRT basic functionality Tests" -Tags "CI" {
...
+        $property2 = "abcdef"
+        $isHiddenTestType = [IsHiddenTestType]::New($property1,$property2)
+        $isHiddenTestType | Export-Clixml $testfile
+        $results = Import-Clixml $testfile
+        $results.Property1 | Should -Be $property1
+        $results.Property2 | Should -Be $property2
+    }
+
+    It "Export-Clixml StopProcessing should succeed" {
+        $sb = [scriptblock]::Create("1..10000 | Export-CliXml -Path $testfile -Verbose")
```
SteveL-MSFT: Isn't it better to use:

```powershell
$sb = { 1..10000 | Export-CliXml -Path $testfile -Verbose }
```

This will give syntax coloring and can be analyzed by psscriptanalyzer in vscode

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/XMLCommand.Tests.ps1
iSazonov: Done.

test/powershell/Modules/Microsoft.PowerShell.Utility/XMLCommand.Tests.ps1
```diff
@@ -24,139 +24,125 @@ Describe "XmlCommand DRT basic functionality Tests" -Tags "CI" {
...
+    }
+
+    It "Export-Clixml StopProcessing should succeed" {
+        $sb = { 1..10000 | Export-CliXml -Path $testfile -Verbose }
+        Test-Stopping $sb -IntervalInMilliseconds 50
+    }
+
+    It "Import-Clixml StopProcessing should succeed" {
+        1..10000 | Export-Clixml -Path $testfile
+        $sb = [scriptblock]::Create("Import-CliXml -Path $testfile,$testfile,$testfile,$testfile,$testfile -Verbose")
```
SteveL-MSFT: Another one to change to `{ ... }`

SteveL-MSFT (changes requested):

test/powershell/Modules/Microsoft.PowerShell.Utility/XMLCommand.Tests.ps1
iSazonov: Sorry. Fixed.

SteveL-MSFT (approved):

SteveL-MSFT (changes requested):
The test failures indicates that the pipeline finishes before stop is processed.  Perhaps we need to rethink how to make the stop predictable.

iSazonov:
No ideas have I how make it more stable so close.

Issue #7749, opened 2018-09-10
Improve authoring experience for Format.ps1xml and Types.ps1xml files


felixfbecker:
The Format.ps1xml and Types.ps1xml system in PowerShell is incredibly powerful. It's awesome how we can output objects, then totally separate from that define how these should be displayed in similar ways to how UIs are declared in XAML or JSX.

However, authoring these files is incredibly frustrating. There is no autocompletion, hovers or validation for these files in your editor. The [reference documentation](https://docs.microsoft.com/en-us/powershell/developer/format/format-schema-xml-reference) is incredibly hard to find, browse and search (a bunch of duplicate `element Foo for Bar for Baz for Quz (Format)`. [Error messages are often very confusing](https://github.com/PowerShell/PowerShell/issues/6788).

The way this could be improved a _lot_ would be with an XSD for these files, which would allow validation, hovers and autocompletion. I started working on this:

![image](https://user-images.githubusercontent.com/10532611/45297479-cdfa5880-b505-11e8-88d4-20cf464532f8.png)

This is with the [vscode-xml](https://github.com/redhat-developer/vscode-xml) extension.
Schemas are here: https://github.com/felixfbecker/PowerShellXSD

I currently just autogenerated this schema from the Format file in that repo.

I will investigate writing a script that parses the official schema markdown reference [here](https://github.com/PowerShell/PowerShell-Docs/blob/staging/developer/format/format-schema-xml-reference.md) and converts it into an XSD (probably with manual edits afterwards).

I couldn't find reference docs for Types.ps1xml files unfortunately, but they also seem simpler.

Another unfortunate thing to mention is that schema validation won't work if you want to use ANSI color codes in Format files because of https://github.com/PowerShell/PowerShell/issues/7526

I am posting this here in case someone wants to help out and because it would be great if eventually these schemas would be versioned in the PowerShell repo and hosted by Microsoft.

felixfbecker:
The schema for Format.ps1xml files is complete and autogenerated from the reference documentation: https://github.com/felixfbecker/PowerShellXSD/blob/master/Format.xsd

lzybkr:
Many years ago we produced xsd files for both format and types ps1xml, I've shared them [here](https://gist.github.com/lzybkr/b2586418259caa5fa2379cb57a93925c), and we should probably include them in this repo.  I encourage you to compare to what you've produced to check correctness,

[...]

PR #7832, opened 2018-09-20
Add XSDs for Format and Types.ps1xml files

felixfbecker:
## PR Summary

As proposed in #7749.

These can be used for autocompletion and validation in editors like VS Code. It makes writing Format and Types files ***so*** much more pleasant to write.

autocompletion

The schemas were taken from #7749 but extended with documentation parsed from the reference docs markdown [with a script](https://github.com/felixfbecker/PowerShellXSD).

Question: Should this be hosted on http://schemas.microsoft.com/ like the MAML schemas? How do the MAML schemas get published there?

cc @lzybkr @SteveL-MSFT

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - User-facing [Documentation needed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#pull-request---submission)
        - Issue filed - Issue link:
    - **OR**
    - Make sure you've added a new test if existing tests do not effectively test the code changed](https://github.com/PowerShell/PowerShell/blob/master/.github/CONTRIBUTING.md#before-submitting)
        - [Add `[feature]` if the change is significant or affects feature tests](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/testing-guidelines.md#requesting-additional-tests-for-a-pr)

### DIFF
--- a/src/Schemas/Format.xsd
+++ b/src/Schemas/Format.xsd
@@ -1,0 +1,758 @@
+<?xml version="1.0" encoding="utf-8"?>
+<xs:schema attributeFormDefault="unqualified" elementFormDefault="qualified" xsi:schemaLocation="http://www.w3.org/2001/XMLSchema http://www.w3.org/2001/XMLSchema.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xs="http://www.w3.org/2001/XMLSchema" xml:lang="en">
+  <xs:element name="Configuration">
+    <xs:annotation>
+      <xs:documentation>Represents the top-level element of a formatting file.</xs:documentation>
+    </xs:annotation>
+    <xs:complexType>
+      <xs:all>
+        <xs:element minOccurs="0" name="DefaultSettings" maxOccurs="1" type="DefaultSettings">
+          <xs:annotation>
+            <xs:documentation>Optional element. Defines common settings that apply to all the views of the formatting file.</xs:documentation>
+          </xs:annotation>
+        </xs:element>
+        <xs:element minOccurs="0" name="SelectionSets" maxOccurs="1" type="SelectionSets">
+          <xs:annotation>
+            <xs:documentation>Optional element. Defines the common sets of .NET objects that can be used by all views of the formatting file.</xs:documentation>
+          </xs:annotation>
+        </xs:element>
+        <xs:element minOccurs="0" name="Controls" maxOccurs="1" type="Controls">
+          <xs:annotation>
+            <xs:documentation>Optional element. Defines the common controls that can be used by all views of the formatting file.</xs:documentation>
+          </xs:annotation>
+        </xs:element>
+        <xs:element minOccurs="0" name="ViewDefinitions" maxOccurs="1" type="ViewDefinitions">
+          <xs:annotation>
+            <xs:documentation>Optional element. Defines the views used to display objects.</xs:documentation>
+          </xs:annotation>
+        </xs:element>
+      </xs:all>
+    </xs:complexType>
+  </xs:element>
+  <xs:complexType name="DefaultSettings">
+    <xs:annotation>
+      <xs:documentation>Defines common settings that apply to all the views of the formatting file. Common settings include displaying errors, wrapping text in tables, defining how collections are expanded, and more.</xs:documentation>
+    </xs:annotation>
+    <xs:all>
+      <xs:element default="false" minOccurs="0" name="ShowError" maxOccurs="1" type="Boolean">
+        <xs:annotation>
+          <xs:documentation>Optional element. Specifies that the full error record is displayed when an error occurs while displaying a piece of data.</xs:documentation>
+        </xs:annotation>
+      </xs:element>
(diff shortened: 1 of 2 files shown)


src/Schemas/Format.xsd
```diff
@@ -0,0 +1,758 @@
...
+    <xs:complexType>
+      <xs:all>
+        <xs:element minOccurs="0" name="DefaultSettings" maxOccurs="1" type="DefaultSettings">
+          <xs:annotation>
+            <xs:documentation>Optional element. Defines common settings that apply to all the views of the formatting file.</xs:documentation>
+          </xs:annotation>
+        </xs:element>
+        <xs:element minOccurs="0" name="SelectionSets" maxOccurs="1" type="SelectionSets">
+          <xs:annotation>
+            <xs:documentation>Optional element. Defines the common sets of .NET Framework objects that can be used by all views of the formatting file.</xs:documentation>
```
SteveL-MSFT: `.NET Framework` is specifically not `.NET Core`.  Since this applies to both, I would suggest just changing all occurrences of `.NET Framework` to simply `.NET`

src/Schemas/Format.xsd
```diff
@@ -0,0 +1,758 @@
...
+      <xs:element minOccurs="0" name="TableColumnHeader" maxOccurs="unbounded" type="TableColumnHeader">
+        <xs:annotation>
+          <xs:documentation>Optional element. Defines the label, the width, and the alignment of the data for a column of a table view.</xs:documentation>
+        </xs:annotation>
+      </xs:element>
+    </xs:sequence>
+  </xs:complexType>
+  <xs:complexType name="TableColumnHeader">
+    <xs:annotation>
+      <xs:documentation>Defines the label, the width of the column, and the alignment of the label for a column of the table.</xs:documentation>
```
SteveL-MSFT: Perhaps make this a bit more generic in case this get extended:

> Defines the properties of the column.  Examples: label, width, and alignment.

SteveL-MSFT (changes requested):
Great contribution!  Didn't review the schema itself in detail, but overall it looks fine.  We can address as bugs if issues are reported.

felixfbecker:
@SteveL-MSFT as mentioned, all documentation is 1:1 unmodified from the Format XML schema reference: https://docs.microsoft.com/en-us/powershell/developer/format/format-schema-xml-reference

I can easily update these schemas with new documentation by running the script again, but I don't think it makes sense to apply fixes to the documentation only to the schema? Most people will probably read the docs online. We should decide which one is the source of truth

SteveL-MSFT:
@felixfbecker I submitted [https://github.com/PowerShell/PowerShell-Docs/pull/2943] to fix the online documentation.  I'm ok leaving the second comment until we add more to the schema.  You can wait for that to be merged and rerun your script.

lzybkr:
I like the idea of a single source of truth. Maybe we should include your script in this PR so anyone can update the schemas from the docs.

felixfbecker:
In an ideal world the docs would rather be generated from the schema (there are probably tools out there that do this). Parsing markdown with regexes is really hacky so I'm not sure if we want to rely on that. The current docs are not very well-structured in general so generating them differently would likely do them good.

@SteveL-MSFT your change actually didn't affect my script because it only updated the table of contents, I sent a follow-up PR: https://github.com/PowerShell/PowerShell-Docs/pull/2944

felixfbecker:
My script also detected many cases in the docs where the schema you gave me @lzybkr 	directly contradicted the reference docs, e.g. an explicit `minOccurs="0"` in the schema, but an explicit "This element is required" in the docs. I tried out a few cases by omitting the element, loading the file into PowerShell and checking if it errors, and found that in all cases the docs were wrong. So the docs are a bad single source of truth.

lzybkr:
Generating the markdown does sound easier. It doesn't need to be in this PR, but I do like that idea.

And I don't see every truly having a single source of truth without generating PowerShell's parser from the xsd - but having done that once before, we'd want the generated code to be much better than what I saw a few years ago.

felixfbecker:
@lzybkr @SteveL-MSFT updated schemas with latest docs, should be good to merge now.

src/Schemas/Types.xsd
```diff
@@ -164,4 +164,4 @@
       </xs:element>
     </xs:sequence>
   </xs:complexType>
-</xs:schema>
+</xs:schema>
```
SteveL-MSFT: Please add a trailing newline at the end

SteveL-MSFT (changes requested):

SteveL-MSFT:
@felixfbecker please rebase off master to pick up fixes causing the CI failures

felixfbecker:
@SteveL-MSFT all green ✅

SteveL-MSFT (approved):

felixfbecker:
@adityapatwardhan?

adityapatwardhan:
@felixfbecker Sorry for the delay. I will review this today.

adityapatwardhan (approved):

adityapatwardhan:
@felixfbecker Thanks for your contribution!

PR #7859, opened 2018-09-25
Ensure NestedModules property gets populated by Test-ModuleManifest

### LINKED ISSUE #7833: Test-ModuleManifest does not populate the NestedModules field
When I run the Test-ModuleManifest cmdlet with a psd1 file that has a NestedModules field, it returns null for NestedModules (even though the dll exists and is in the correct location).  This results in Update-ModuleManifest commenting out the existing NestedModule fields when I try to update just the version.  This is causing massive issues with our versioning scripts when trying to update from PowerShell 5.1 to PowerShell 6.0.  Running this in PowerShell 5.1 with the same files works as expected.  When I pulled down PowerShell to try to debug this, I saw that NestedModules was set correctly here: https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs#L2041, but not set in moduleInfo before moduleInfo was returned here: https://github.com/PowerShell/PowerShell/blob/master/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs#L2800.

Steps to reproduce
------------------

```powershell
$a = Test-ModuleManifest -Path <path to module with filled out nestedmodules field>
$a.NestedModule # Returns nothing
```

Expected behavior
-----------------

```none
$a = Test-ModuleManifest -Path <path to module with filled out nestedmodules field>
$a.NestedModule # Returns correct Nested Modules
```

Actual behavior
---------------
Returns nothing
```none

```

Environment data
----------------

```powershell
> $PSVersionTable
PSVersion                      6.1.0-preview.4
GitCommitId                    6.1.0-preview.4
```

SteveL-MSFT:
## PR Summary

This is a regression introduced by https://github.com/PowerShell/PowerShell/pull/7145 as an optimization where no further analysis of the module is done if none of the exported capabilities have a wildcard.  However, the NestedModules property hasn't been populated yet, so it remains empty.  This is a regression as previously this property is always populated if the manifest contains a value.

Fix is to create PSModuleInfo entries for each specified NestedModule even though we don't do actual loading and analysis.  Creating the fake NestedModule is similar to how the code already creates fake RequiredModules when not importing.

Also some minor cleanup of repetitive code in the tests.

Fix https://github.com/PowerShell/PowerShell/issues/7833

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**

### DIFF
--- a/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
@@ -2797,6 +2797,20 @@ namespace Microsoft.PowerShell.Commands
 
                 if (!needToAnalyzeScriptModules)
                 {
+                    // need to add nested modules to the manifestInfo when no more analysis needs to be done
+                    PSModuleInfo fakeNestedModuleInfo = null;
+                    foreach (ModuleSpecification nestedModule in nestedModules)
+                    {
+                        fakeNestedModuleInfo = new PSModuleInfo(nestedModule.Name, Context, null);
+                        if (nestedModule.Guid.HasValue)
+                        {
+                            fakeNestedModuleInfo.SetGuid(nestedModule.Guid.Value);
+                        }
+                        fakeNestedModuleInfo.SetVersion(nestedModule.RequiredVersion ?? nestedModule.Version);
+
+                        manifestInfo.AddNestedModule(fakeNestedModuleInfo);
+                    }
+
                     return manifestInfo;
                 }
             }
--- a/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
@@ -1339,7 +1339,7 @@ namespace Microsoft.PowerShell.Commands
         }
 
         /// <summary>
-        /// Helper function to generate fake PSModuleInfo objects from ModuleSpecification objects
+        /// Helper function to generate fake PSModuleInfo objects from ModuleSpecification objects.
         /// </summary>
         /// <param name="moduleSpecs">Collection of ModuleSpecification objects</param>
         /// <returns>Collection of fake PSModuleInfo objects</returns>
--- a/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
+++ b/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
@@ -1338,6 +1338,25 @@ namespace Microsoft.PowerShell.Commands
             return true;
         }
 
+        /// <summary>
+        /// Helper function to generate fake PSModuleInfo objects from ModuleSpecification objects
+        /// </summary>
+        /// <param name="moduleSpecs">Collection of ModuleSpecification objects</param>
+        /// <returns>Collection of fake PSModuleInfo objects</returns>
+        private IEnumerable<PSModuleInfo> CreateFakeModuleObject(IEnumerable<ModuleSpecification> moduleSpecs)
+        {
+            foreach (ModuleSpecification moduleSpec in moduleSpecs)
+            {
+                var fakeModuleInfo = new PSModuleInfo(moduleSpec.Name, Context, null);
+                if (moduleSpec.Guid.HasValue)
+                {
+                    fakeModuleInfo.SetGuid(moduleSpec.Guid.Value);
+                }
+                fakeModuleInfo.SetVersion(moduleSpec.RequiredVersion ?? moduleSpec.Version);
+                yield return fakeModuleInfo;
+            }
+        }
+
         private ErrorRecord GetErrorRecordIfUnsupportedRootCdxmlAndNestedModuleScenario(
             Hashtable data,
             string moduleManifestPath,
@@ -1977,16 +1996,8 @@ namespace Microsoft.PowerShell.Commands
                 }
                 else
                 {
-                    PSModuleInfo fakeRequiredModuleInfo = null;
-                    foreach (ModuleSpecification requiredModule in requiredModules)
+                    foreach (PSModuleInfo fakeRequiredModuleInfo in CreateFakeModuleObject(requiredModules))
                     {
-                        fakeRequiredModuleInfo = new PSModuleInfo(requiredModule.Name, Context, null);
-                        if (requiredModule.Guid.HasValue)
-                        {
-                            fakeRequiredModuleInfo.SetGuid(requiredModule.Guid.Value);
-                        }
-                        fakeRequiredModuleInfo.SetVersion(requiredModule.RequiredVersion ?? requiredModule.Version);
-
                         requiredModulesSpecifiedInModuleManifest.Add(fakeRequiredModuleInfo);
                     }
                 }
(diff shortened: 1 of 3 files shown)


rjmholt:
This seems like the right way to efficiently build/return the nested module info objects. But for the sake of having considered other ideas, wouldn't users expect a fully live nested module info object here?

Since the fake module info is the same type that would normally be returned by `Test-ModuleManifest` but is hollow, the fact that we took an optimised path means you now don't get back fields like `ModuleBase` and have no idea why. Unlike `RequiredModules`, `NestedModules` really forms a tree, and you might want to traverse that -- but now you wouldn't be able to consistently expect to be able to use that behaviour.

[Further down in the same method](https://github.com/PowerShell/PowerShell/blob/76c9da189dd5990010fbce3d44890d9659be9b0e/src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs#L2927-L2949) when nested modules are read, we actually go and read their manifest and load a module info.

Creating fake info objects from the nested module specifications is certainly more efficient, but I think it's a breaking change from previous (pre-regression) behaviour. See Windows PowerShell's version of the same:
```
PS C:\Users\roholt\Documents\Dev\sandbox\mod> $x = Test-ModuleManifest .\mod.psd1
PS C:\Users\roholt\Documents\Dev\sandbox\mod> $x.NestedModules |fl *


LogPipelineExecutionDetails : False
Name                        : x
Path                        : C:\Users\roholt\Documents\Dev\sandbox\mod\x.psm1
ImplementingAssembly        :
Definition                  :
Description                 :
Guid                        : 00000000-0000-0000-0000-000000000000
HelpInfoUri                 :
ModuleBase                  : C:\Users\roholt\Documents\Dev\sandbox\mod
PrivateData                 :
Tags                        : {}
ProjectUri                  :
IconUri                     :
LicenseUri                  :
ReleaseNotes                :
RepositorySourceLocation    :
Version                     : 0.0
ModuleType                  : Script
Author                      :
AccessMode                  : ReadWrite
ClrVersion                  :
CompanyName                 :
Copyright                   :
DotNetFrameworkVersion      :
ExportedFunctions           : {[Banana, Banana]}
Prefix                      :
ExportedCmdlets             : {}
ExportedCommands            : {[Banana, Banana]}
FileList                    : {}
CompatiblePSEditions        : {}
ModuleList                  : {}
NestedModules               : {}
PowerShellHostName          :
PowerShellHostVersion       :
PowerShellVersion           :
ProcessorArchitecture       : None
Scripts                     : {}
RequiredAssemblies          : {}
RequiredModules             : {}
RootModule                  :
ExportedVariables           : {}
ExportedAliases             : {}
ExportedWorkflows           : {}
ExportedDscResources        : {}
SessionState                :
OnRemove                    :
ExportedFormatFiles         : {}
ExportedTypeFiles           : {}
```

I think we may be committed to doing the inefficient thing.

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -2797,6 +2797,20 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
 
                 if (!needToAnalyzeScriptModules)
                 {
+                    // need to add nested modules to the manifestInfo when no more analysis needs to be done
```
PaulHigin: // Add nested modules ...

PaulHigin (commented):
LGTM

SteveL-MSFT:
@rjmholt You raise a fair point.  However, I think most cases the user doesn't need a complete object.  In the issue, the problem with the current behavior is that the `NestedModules` property gets cleared out when used with `Update-ModuleManifest` which is dataloss for the customer.  The fake moduleinfo is sufficient as the modulemanifest is really just a hashtable.  Although it seems I should add a test to Update-ModuleManifest.

I would suggest deferring whether we want to add capability to get the entire tree of objects until there is a customer request for it.

daxian-dbw:
@SteveL-MSFT Thanks for fixing this regression! It looks good to me, but since the same code was both here and for `requiredModules`, would it make sense to have a private helper method to encapsulate the code for creating the fake `ModuleInfo` object?

The helper method can be a iterator method, which can be used for both here an `requiredModules`:
```
private IEnumerable<ModuleInfo> CreateFakeModuleObject(IEnumerable<ModuleSpecification> moduleSpecs)
{
    foreach (ModuleSpecification moduleSpec in moduleSpecs)
    {
        var fakeModuleInfo = new PSModuleInfo(moduleSpec.Name, Context, null);
        if (moduleSpec.Guid.HasValue)
        {
            fakeModuleInfo.SetGuid(moduleSpec.Guid.Value);
        }
        fakeModuleInfo.SetVersion(moduleSpec.RequiredVersion ?? moduleSpec.Version);
        yield fakeModuleInfo;
    }
}
```

SteveL-MSFT:
@daxian-dbw yes, I'll make that change.  Thanks!

daxian-dbw (approved):
LGTM

src/System.Management.Automation/engine/Modules/ModuleCmdletBase.cs
```diff
@@ -1338,6 +1338,25 @@ internal bool LoadModuleManifestData(ExternalScriptInfo scriptInfo, ManifestProc
             return true;
         }
 
+        /// <summary>
+        /// Helper function to generate fake PSModuleInfo objects from ModuleSpecification objects
```
iSazonov: Please add final dot.

test/powershell/engine/Module/TestModuleManifest.Tests.ps1
```diff
@@ -2,13 +2,17 @@
 # Licensed under the MIT License.
 Describe "Test-ModuleManifest tests" -tags "CI" {
 
+    BeforeEach {
+        $testModulePath = "testdrive:/module/test.psd1"
+        New-Item -ItemType Directory -Path testdrive:/module
```
iSazonov: Please suppress output `New-Item -ItemType Directory -Path testdrive:/module > $null`.

test/powershell/engine/Module/UpdateModuleManifest.Tests.ps1
```diff
@@ -0,0 +1,33 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+Describe "Update-ModuleManifest tests" -tags "CI" {
+
+    BeforeEach {
+        $testModulePath = "testdrive:/module/test.psd1"
+        New-Item -ItemType Directory -Path testdrive:/module
```
iSazonov: The same.

test/powershell/engine/Module/UpdateModuleManifest.Tests.ps1
```diff
@@ -0,0 +1,33 @@
...
+        Remove-Item -Recurse -Force -ErrorAction SilentlyContinue testdrive:/module
+    }
+
+    It "Update should not clear out NestedModules: <variation>" -TestCases @(
+        @{ variation = "export with wildcards"; exportValue = "*" },
+        @{ variation = "export without wildcards"; exportValue = "@()"}
+    ) {
+        param($exportValue)
+
+        New-Item -ItemType File -Path testdrive:/module/foo.psm1
```
iSazonov: The same.

iSazonov (approved):

PR #2518, opened 2016-10-21
Added -Top/-Bottom params to Sort-Object for Top/Bottom N sort and many Pester tests

KirkMunro:
This changeset addresses issue #2462, along with all feedback received in the comments for that issue.

The original issue was identified as an enhancement to add -Top <int> support to Sort-Object, so that (a) you would get results back faster when you only wanted the top N, and (b) you would get code elegance by not having to pipe to Select-Object. Several commenters wanted -Bottom N support added for completeness, so an additional parameter set was created with a -Bottom <ìnt> parameter.  Over 40 tests were written for Pester to compare the sort results using Sort -Top <int> or Sort -Bottom <int> with sort results when -Top and -Bottom are not used to ensure that we always get back the data we want according with how Sort-Object sorts data without -Top/-Bottom.  That means that if Sort-Object is broken (if it returns results in a different order), these tests will pass; however, the intent of these tests is to compare the sort results with those of a basic Sort-Object invocation, and other tests exist to verify the order of the items returned from Sort-Object, so this is all good.

Additionally, performance testing was done and -Top/-Bottom sorts on very large data sets (1M integers) returned results about twice as fast as when using sort | select -First/-Last, as long as -Unique is not used.  This performance improvement is partly because of the reduced pipeline and partly because of a faster sort algorithm (an indexed min-/max-heap sort is used to process Top N/Bottom N results more quickly).

Worth noting: Using -Top/-Bottom sorts using -Unique are really, really slow with very large data sets, just as sort | select -First/-Last are slow with very large data sets. This code is old, and while looking at it I felt it would be worth looking at SortedSet for -Unique sorts (SortedSet did not exist when sort-object was first written) to see if it would return unique sorted results more quickly. That investigation was outside of the scope of this PR though, so I'm simply mentioning it here so that I don't forget about the item. A separate issue should probably be created to investigate SortedSet with Sort-Object -Unique to see if it improves performance like I hope it will.

PR #7862, opened 2018-09-26
Add `-Stable` to `Sort-Object` and related tests

### LINKED ISSUE #7846: Sort-Object inappropriately sacrifices stability for performance
PR #2518 by @KirkMunro adds the ability to select the top or bottom `N` items of a sorted set, introducing stability to the sorting method. However, this behavior is undermined whenever `N` equals or exceeds the input size.

This deviance can be observed when sorting by congruence modulo for example:

Steps to reproduce
------------------

```powershell
1..20 |Sort-Object {$_ % 3} -Top 20
```

Expected behavior
-----------------

```none
3
6
9
12
15
18
1
4
7
10
13
16
19
2
5
8
11
14
17
20
```

Actual behavior
---------------

```none
18
3
15
6
12
9
1
16
13
10
7
4
19
11
8
14
5
17
2
20
```

The culprit seems to be the last two conditions in this branch: https://github.com/KirkMunro/PowerShell/blob/master/src/Microsoft.PowerShell.Commands.Utility/commands/utility/sort-object.cs#L247

```csharp
            // If -Top & -Bottom were not used, or if -Top or -Bottom would return all objects, invoke
            // an in-place full sort
            if ((Top == 0 && Bottom == 0) || Top >= dataToProcess.Count || Bottom >= dataToProcess.Count)
            {
                sortedItemCount = FullSort(dataToProcess, comparer);
            }
            // Otherwise, use an indexed min-/max-heap to perform an in-place sort of all objects
            else
            {
                sortedItemCount = Heapify(dataToProcess, comparer);
            }
```

Suggested fix would be to either add a `-Stable` switch parameter that forces `Sort-Object` to use the stable min/max heap sort implementation, or change the branch above to only default to `FullSort()` when `-Top`/`-Bottom` are omitted completely.

My personal preference would be to go for _both_ changes - `-Stable` would be nice for when you don't know input size up front but want to maintain input order for equal items

Environment data
----------------

```powershell
> $PSVersionTable

PSVersion                      6.1.0
GitCommitId                    6.1.0-57-g0f0c46dfe51440470b86f67b258e14aea32fa0aa
```

KirkMunro:
## PR Summary

Fix #7846.

- added -Stable switch parameter to Sort-Object
- moved -Top parameter from Default parameter set to Top parameter set and made it mandatory for that parameter set
- added new Pester tests for -Stable switch (ascending, descending, with/without unique)

Regarding the breaking change, this is clearly in [Bucket 3](https://github.com/PowerShell/PowerShell/blob/master/docs/dev-process/breaking-change-contract.md#bucket-3-unlikely-grey-area) (technically it is breaking, but it is highly unlikely that anyone will be impacted by this change). For details on how it is breaking, see [here](#issuecomment-424757855).

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [X] Issue filed - Issue link: https://github.com/PowerShell/PowerShell-Docs/issues/3246
    - **OR**


src/Microsoft.PowerShell.Commands.Utility/commands/utility/sort-object.cs
```diff
@@ -17,6 +18,17 @@ public sealed class SortObjectCommand : OrderObjectBase
...
+        /// This param specifies if a stable sort is required.
+        /// </summary>
+        /// <value></value>
+        [Parameter(ParameterSetName="Default")]
+        public SwitchParameter Stable
+        {
+            get { return _stable; }
+            set { _stable = value; }
+        }
+        private bool _stable;
```
iSazonov: Please use auto-property.

test/powershell/Modules/Microsoft.PowerShell.Utility/Sort-Object.Tests.ps1
```diff
@@ -225,6 +225,80 @@ Describe "Sort-Object DRT Unit Tests" -Tags "CI" {
 	}
 }
 
+Describe 'Sort-Object Stable Unit Tests' -Tags 'CI' {
+
+	Context 'Modulo stable sort' {
+
+		$unsortedData = 1..20
```
iSazonov: We could use shorter range  `1..10`.

iSazonov (commented):
Please look CodeFactor issues.

test/powershell/Modules/Microsoft.PowerShell.Utility/Sort-Object.Tests.ps1
KirkMunro: Using a shorter range invalidates the test. A longer range is required for the test to be able to properly identify that the stable sort is working. There is no difference between a stable and an unstable short with a range of 1..10.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/sort-object.cs
KirkMunro: My preference is absolutely for auto-properties; however, the reason I chose to use this format was simply to follow existing style for other SwitchParameter properties. My assumption is that whoever implemented sort had a reason for creating a boolean field to accompany a SwitchParameter property instead of using an auto-property. If we are to change this cmdlet to use auto-properties for parameters of type SwitchParameter, I think it should be done for all the switch parameters in the cmdlet, not just the one that I have added. Thoughts?

KirkMunro:
@iSazonov Why did you add the Breaking-Change label? Because the Top parameter was moved to a separate parameter set (which would invalidate code someone had written specifically against the Default parameter set if they expected Top to be in there)? Or is there a functional breaking change you see here that I don't?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/sort-object.cs
KirkMunro: You know what, nevermind, I don't see any advantage to using a boolean field with a switchparameter parameter, so I made this change for both -Stable and -Unique parameters.

iSazonov:
@KirkMunro Formally it is a breaking change because (1) a parameter set is changed, (2) sort order is changed and some scripts can fail. I believe we should inform users in change log about this. Also we need PowerShell Committee approvement.

iSazonov:
@KirkMunro Please update Markdig.Signed package version 0.15.2 -> 0.15.3 to pass CI tests.

anmenaga:
@SteveL-MSFT we need to make sure this breaking change gets reviewed by PS Committee...
`Review - Committee` label was added almost a month ago.

SteveL-MSFT:
@anmenaga yes, this is in the queue for review but we have so many items now it's taking a bit longer

iSazonov:
@KirkMunro Please resolve merge conflicts.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/sort-object.cs
```diff
@@ -239,13 +245,12 @@ protected override void EndProcessing()
             // Track the number of items that will be output from the data once it is sorted
             int sortedItemCount = dataToProcess.Count;
 
-            // If -Top & -Bottom were not used, or if -Top or -Bottom would return all objects, invoke
-            // an in-place full sort
-            if ((Top == 0 && Bottom == 0) || Top >= dataToProcess.Count || Bottom >= dataToProcess.Count)
```
PaulHigin: Why did you remove `Top >= dataToProcess.Count || Bottom >= dataToProcess.Count` condition?

src/Microsoft.PowerShell.Commands.Utility/commands/utility/sort-object.cs
```diff
@@ -239,13 +245,12 @@ protected override void EndProcessing()
...
+            // If -Stable, -Top & -Bottom were not used, invoke an in-place full sort
+            if (!Stable && Top == 0 && Bottom == 0)
             {
                 sortedItemCount = FullSort(dataToProcess, comparer);
             }
-            // Otherwise, use an indexed min-/max-heap to perform an in-place sort of all objects
+            // Otherwise, use an indexed min-/max-heap to perform an in-place, stable sort of all objects
```
PaulHigin: Please add to comment that Heap sort performs a stable sort, and provide short description of stable sort.

PaulHigin (changes requested):

src/Microsoft.PowerShell.Commands.Utility/commands/utility/sort-object.cs
```diff
@@ -16,37 +16,42 @@ namespace Microsoft.PowerShell.Commands
     public sealed class SortObjectCommand : OrderObjectBase
     {
         #region Command Line Switches
+
+        /// <summary>
+        /// Gets or sets a value indicating whether a stable sort is required.
```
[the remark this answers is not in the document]
KirkMunro: Done and done. Commit coming shortly.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/sort-object.cs
KirkMunro: Done and coming in a commit shortly.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/sort-object.cs
KirkMunro: The discussion in bug #7846 describes this in detail (and this was the original driver for this PR). With the old logic, when Top or Bottom exceed the number of items being sorted, you wouldn't end up with a heap sort, which you should when you use Top or Bottom. The end result was a less performant, unstable full sort when you expect a stable heap sort to be performed. Removing these conditions fixes that issue (one part of this PR). Adding -Stable addresses the need for stable sorting when you don't care about the original data size (the other part of this PR).

PaulHigin (approved):
LGTM

SteveL-MSFT:
@BrucePay I believe you had an action item to review this PR on behalf of the @PowerShell/powershell-committee

adityapatwardhan (approved):

adityapatwardhan:
Change looks good, awaiting review from committee.

sdwheeler:
Docs issue https://github.com/PowerShell/PowerShell-Docs/issues/3021

BrucePay:
@KirkMunro @anmenaga My main question is this: do we really need a custom implementation of a heap sort? LINQ OrderBy and OrderByDescrnding are stable sorts. Why not just use them? It would be a lot less code to maintain.

iSazonov:
There has been a significant performance increase in # 2518 + support Top/Bottom.
There is open issues in CoreFX for new API that can replace the code.

KirkMunro:
@BrucePay That question has nothing to do with this PR. This PR is to correct issue #7846, while also allowing full stable sorts using the same algorithm. The sorting algorithm isn't new here. It was added for PowerShell 6.0 in PR #2518 to greatly improve Sort-Object performance issues while adding -Top and -Bottom to Sort-Object.

To get the answer to your question, I think it would be better to open another issue requesting a comparison between LINQ's OrderBy and OrderByDescending and the existing indexed min-/max-heap sort algorithm, which should include an in depth analysis of results to ensure sorting does not change as well as a thorough performance analysis. Based on past experience investigating the performance issues with Sort-Object, and figuring out the algorithm to add -Top and -Bottom sort capabilities. Either that or compare with the CoreFX issue that @iSazonov just mentioned (Ilya, can you provide a link to that please?).

BrucePay (approved):
Hmm - sorry - I clicked the link from the issue and started looking at the wrong PR. Given that the code I'm concerned about has already been committed I agree that a separate issue is the right way to approach it. Other than that, you might consider checking PSBoundParameters to see if the top/bottom/stable parameters have been _specified_ instead of depending on signal values but I'll leave that up to you. (Sooner or later someone is probably going to suggest using `sort -Top -10` to get all _but_ the top ten :-) ) . Anyway I approve these changes.

src/Microsoft.PowerShell.Commands.Utility/commands/utility/sort-object.cs
```diff
@@ -16,37 +16,46 @@ namespace Microsoft.PowerShell.Commands
...
+        /// </summary>
+        /// <value></value>
+        /// <remarks>
+        /// Items that are duplicate according to the sort algorithm will appear
```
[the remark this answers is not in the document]
KirkMunro: Fixed.

iSazonov:
Waiting PowerShell Committee review.

sdwheeler:
@SteveL-MSFT What's the status on this PR?

adityapatwardhan:
@KirkMunro Fix the merge conflict.

adityapatwardhan:
@KirkMunro Thank you for your contribution!

KirkMunro:
Thanks @adityapatwardhan, for fixing the merge conflict and merging the PR. I missed the earlier comments about the merge conflict.

PR #7875, opened 2018-09-26
Replace test certificates with self-signed certificate generating command

rjmholt:
## PR Summary

Adds a dependency on the newly published [SelfSignedCertificate](https://github.com/rjmholt/SelfSignedCertificate) module so that we can construct our own self-signed certificates at test time rather than checking them in.

To do this, the following happens:

- `Publish-PSTestTools` will `Save-Module` the SelfSignedCertificate from the gallery to
  `$RepoRoot/test/tools/Modules/`
- In `WebListener.psm1` we import the SelfSignedCertificate module
  - I tried both `using module` at the top of the psm1 and `NestedModules` in the psd1 here but neither seemed to work
- `Start-WebListener` now builds both client and server certificates when it is called

## PR Checklist

- **User-facing changes**
    - [x] Not Applicable
    - **OR**
    - **OR**


.gitignore
```diff
@@ -6,6 +6,7 @@ project.lock.json
 /staging/
 /Packages/
 *.nuget.props
+/test/tools/Modules/SelfSignedCertificate/
```
SteveL-MSFT: Even though this module is for test, perhaps it's ok to just use `install-module -scope currentuser`?  Adding it to a repo path breaks our work towards separating tests from dependencies on the repo.

.gitignore
```diff
@@ -46,6 +47,9 @@ dotnet-uninstall-debian-packages.sh
 *.nupkg
 *.AppImage
 
+# Ignore certificates
+*.pfx
```
SteveL-MSFT: Similarly, the pfx files should be generated outside of the repo structure

test/tools/Modules/WebListener/WebListener.psm1
```diff
@@ -18,6 +21,61 @@ Class WebListener
 
 [WebListener]$WebListener
 
+function New-ClientCertificate
+{
+    param([string]$CertificatePath)
+
+    $password = ConvertTo-SecureString -Force -AsPlainText 'password'
```
SteveL-MSFT: We shouldn't have a hardcoded password.  Looks like the `GeneratePassword()` method isn't available in corefx.  I think it's fine to generate a random alphanumeric password.

test/tools/Modules/WebListener/WebListener.psm1
```diff
@@ -61,10 +119,14 @@ function Start-WebListener
         $appDll              = 'WebListener.dll'
         $serverPfx           = 'ServerCert.pfx'
         $serverPfxPassword   = 'password'
```
SteveL-MSFT: Maybe be better to store in env var the generated password rather than hardcode

SteveL-MSFT (changes requested):

.gitignore
rjmholt: I had that exact discussion with @JamesWTruher and argued that we shouldn't pollute the installed modules because PowerShell was build on a machine. But I will make the change

.gitignore
rjmholt: The question is where. We don't have a cross-platform temp dir and I would very much prefer not to put it in `$HOME`. Ideally it would be on the `$TestDrive`, but anything that tries to operate across `Describe` blocks with a generated certificate will fail.

Arguably we should refactor things to do exactly that, but I think that is out of scope of this PR.

SteveL-MSFT: For now, `Path::GetTempPath()` is good enough I think

iSazonov:
We slow down CIs. Why we need the change?

SteveL-MSFT:
@iSazonov as a Microsoft project, we can't have secrets (even test secrets) stored statically in the repo

SteveL-MSFT (approved):

.gitignore
JamesWTruher: we should *not* be polluting the user's space with anything that isn't explicitly testing our behavior for the current user. These tests are very likely to be run on a dev box and we should *not* be adding stuff to the current user's scope. Further, this is a tool *only* for our tests. If the user happens to have this already, it still shouldn't use it, we should be using the module that we know we should be using in the location that we control without molesting their environment.

build.psm1
```diff
@@ -5,6 +5,10 @@
 # On Windows paths is separated by semicolon
 $script:TestModulePathSeparator = [System.IO.Path]::PathSeparator
 
+# Path to a directory to store modules we temporarily need to test PowerShell
+$script:TestModuleDirPath = Join-Path ([System.IO.Path]::GetTempPath()) 'PwshTestModules'
+New-Item -Force -ItemType Directory -Path $script:TestModuleDirPath
```
iSazonov: Do we need output to console? If no we should `New-Item > $null`.

build.psm1
```diff
@@ -1139,7 +1146,7 @@ function Start-PSPester {
     }
 
     # Autoload (in subprocess) temporary modules used in our tests
-    $command += '$env:PSModulePath = '+"'$TestModulePath$TestModulePathSeparator'" + '+$($env:PSModulePath);'
+    $command += '$env:PSModulePath = '+"'$TestModulePath$TestModulePathSeparator$script:TestModuleDirPath$TestModulePathSeparator'" + '+$env:PSModulePath;'
```
iSazonov: I'm afraid of such expressions. Maybe use Join-Path before the line?

iSazonov (commented):
@SteveL-MSFT  I found some samples of hard coded passwords in test files by key "password". (And in New-TestHost too).

build.psm1
rjmholt: `Join-Path` won't work since we want `;` and not `\` (or `:` and not `/` on *nix). But I'll break this expression up a bit

iSazonov (approved):

build.psm1
```diff
@@ -1139,7 +1146,8 @@ function Start-PSPester {
     }
 
     # Autoload (in subprocess) temporary modules used in our tests
-    $command += '$env:PSModulePath = '+"'$TestModulePath$TestModulePathSeparator'" + '+$($env:PSModulePath);'
+    $newPathFragment = $TestModulePath + $TestModulePathSeparator + $script:TestModuleDirPath + $TestModulePathSeparator
```
daxian-dbw: `$TestModulePath` and `TestModuleDirPath` look a bit confusing. It would be nice if we can pull the `SelfSignedCertificate` module to the `test\tools\Modules` folder, and update `.gitignore` to ignore files in the `SelfSignedCertificate` folder. Only update/overwrite it in `Start-PSPester` when we bump the version.

This is not a blocking comment for this PR, but it would be great if it can be addressed in a follow-up PR.

test/powershell/Modules/Microsoft.PowerShell.Security/certificateCommon.psm1
```diff
@@ -78,7 +78,17 @@ Function New-ProtectedCertificate
     Password: "password"
     #>
 
-    $certLocation = ".\test\tools\Modules\WebListener\ServerCert.pfx"
+    $certLocation = Join-Path ([System.IO.Path]::GetTempPath()) 'protectedCert.pfx'
+
+    # Hackery to create a new 10 digit random hex string as a password
+    $rand = [Random]::new()
+    $global:protectedCertPassword = ConvertTo-SecureString -Force -AsPlainText (((1..10).ForEach{ '{0:x}' -f $rand.Next(0xf) }) -join '')
```
daxian-dbw: Better expose it through a function, instead of a global variable. The global variable will linger around even after you remove the module from a session.

build.psm1
SteveL-MSFT: @daxian-dbw polluting the git directories is counter-productive to the work to enable the tests to run outside of the git repo

rjmholt: See https://github.com/PowerShell/PowerShell/pull/7875#discussion_r220750974 and https://github.com/PowerShell/PowerShell/pull/7875#pullrequestreview-159232521.

daxian-dbw: Good to know that thoughts have been put on this issue. Closed.

test/powershell/Modules/Microsoft.PowerShell.Security/certificateCommon.psm1
```diff
@@ -68,6 +68,19 @@ OksttXT1kXf+aez9EzDlsgQU4ck78h0WTy01zHLwSKNWK4wFFQM=
...
+{
+    # Hackery to create a new 10 digit random hex string as a password
+    $rand = [Random]::new()
+    $script:protectedCertPassword = ConvertTo-SecureString -Force -AsPlainText (((1..10).ForEach{ '{0:x}' -f $rand.Next(0xf) }) -join '')
+    return $script:protectedCertPassword
+}
+
+Function Get-CertificatePassword
+{
+    return $script:protectedCertPassword
```
adityapatwardhan: Should throw if $script:protectedCertPassword is not set or create a new one and return.

test/tools/Modules/WebListener/WebListener.psm1
```diff
@@ -18,6 +19,76 @@ Class WebListener
 
 [WebListener]$WebListener
 
+function New-RandomHexString
+{
+    param([int]$Length = 10)
+
+    $random = [Random]::new()
+
+    return ((1..$Length).ForEach{ '{0:x}' -f $random.Next(0xf) }) -join ''
```
adityapatwardhan: This code is the same as `New-CertificatePassword` can we re-use the code?

adityapatwardhan (changes requested):

test/tools/Modules/WebListener/WebListener.psm1
adityapatwardhan: @rjmholt How about in here: https://github.com/PowerShell/PowerShell/blob/master/test/tools/Modules/HelpersCommon/HelpersCommon.psm1

adityapatwardhan (approved):

Issue #7557, opened 2018-08-17
Allow binary numbers, byte literals, and BigInteger?


vexx32:
I will attempt to summarise here the primary points of discussion that have ensued in #7993 as it has spiraled into many threads, and I suspect a bullet-point summary of questions to be answered will be significantly easier on the committee. 

## Regarding Binary & Hex Parsing

As part of the refactor & introduction of binary parsing, methodology of hex parsing has also been altered a bit. Parsing currently _results_ in the same as it currently works, with the caveat that literals with values above `Int64.MaxValue` are also now acceptable.

### Parse as C# Literals

With that in mind, @mklement brought up the point that we may want to simply _change_ how hex and binary parsing work. That is, mimic C#'s behaviour of these literals in source, which would mean parsing _all_ hexadecimal literals as strictly positive (no more `0xFFFFFFFF -eq -1` &mdash; instead, `0xFFFFFFFF -eq UInt32.MaxValue`) **and** having such literals smoothly convert up to `UInt` values. 

With that in mind, the code patterns for hex or binary literals would seek out the lowest available parsed value type (when no type is specified) in the following order: `Int32`, `UInt32`, `Int64`, `UInt64`, `Decimal`, and possibly finally `BigInteger`.

#### Hex & BigInteger

If we elect to _keep_ current hex behaviour, we need to consider how it would behave in ranges higher than `Decimal`. `BigInteger`'s default parser for hex numerals will simply assume the highest bit of a byte is indicative of the sign. As a result any numeral treated as signed that begins with `0x8` or higher will be considered the negative two's complement representation when we enter ranges that can only be parsed as `BigInteger`. This could be overridden easily, if this behaviour is considered to be undesirable.

#### Binary Parsing with Sign Bits

Then we face the issue of what to do about binary parsing. I doubt most folks working with binary directly will be working in ranges above 32-bit numbers, but I could be very wrong on that. They are, however, easier to work with in the `byte`, `short`, and `int` value ranges (8, 16, 32-length literals), and behaviour of a sign bit in this case is _also_ entirely up to the parser here due to the custom implementation of binary parsing for speed concerns.

[...]

PR #7901, opened 2018-10-01
Add support for Byte Literals

vexx32:
## PR Summary

* Adds `y` suffix that is used to specify a numeric literal as the `sbyte` data type.
  * Can be combined with the existing `u` suffix as `uy` to specify the `byte` data type
* Adds tests to ensure the parsing for these works correctly.

Examples:
```powershell
PS> 100y
100
PS> 100y.GetType().FullName
System.SByte
PS> 100uy
100
100uy.GetType().FullName
System.Byte
PS> 300uy
# Error: Invalid numeric literal
```

N.B.: I believe this will be a breaking change, as it means some command names (specifically, numerals suffixed with a y such as `100y`) will now be interpreted as numerals instead of potential function or command names.

Ref: #7557 -- This will resolve the first part of that issue. The second portion of that issue will be solved in a follow-up PR where binary literal support is added.

## PR Checklist

- **User-facing changes**
    - [ ] Not Applicable
    - **OR**
        - [X] Issue filed - Issue link: https://github.com/PowerShell/PowerShell-Docs/issues/2965
    - **OR**


iSazonov:
@vexx32 Please open Issue in docs repo and add link to the PR description

src/System.Management.Automation/engine/parser/tokenizer.cs
```diff
@@ -3355,6 +3365,15 @@ private static bool TryGetNumberValue(string strNum, bool hex, bool real, Number
                                 return true;
                             }
 
+                            result = null;
+                            return false;
+                        case NumberSuffixFlags.SignedByte:
+                            if (sbyte.TryParse(strNum, style, NumberFormatInfo.InvariantInfo, out sbyte sb))
+                            {
+                                result = (sbyte)(sb * multiplier);
```
iSazonov: Seems we could return false if multiplier > 1 and SignedByte or UnsignedByte. I do not know whether it makes sense to do this test explicitly. Certainly this will eliminate the exception.

test/powershell/Language/Parser/Parser.Tests.ps1
```diff
@@ -709,6 +709,32 @@ foo``u{2195}abc
...
+                @{ Script = "-3e1y"; ExpectedValue = "-30"; ExpectedType = [sbyte] }
+                #Hexadecimal
+                @{ Script = "0x0y"; ExpectedValue = "0"; ExpectedType = [sbyte] }
+                @{ Script = "0x41y"; ExpectedValue = "65"; ExpectedType = [sbyte] }
+                @{ Script = "-0x41y"; ExpectedValue = "-65"; ExpectedType = [sbyte] }
+                #Binary
+                @{ Script = "0b0y"; ExpectedValue = "0"; ExpectedType = [sbyte] }
+                @{ Script = "0b10y"; ExpectedValue = "2"; ExpectedType = [sbyte] }
+                @{ Script = "-0b10y"; ExpectedValue = "-2"; ExpectedType = [sbyte] }
+                @{ Script = "0b11111111y"; ExpectedValue = "-1"; ExpectedType = [sbyte] }
```
iSazonov: Please remove the tests for binary suffix.

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: Well, still needs to be parsed as (s)byte regardless, no?

But yes, taking the shortcut of skipping that if there is a multiplier applied here is a code call, thank you!

iSazonov: >  still needs to be parsed as (s)byte regardless

I think yes. It allow "recovery" and continue parsing.

vexx32: Aaaand patched in, thank you!

test/powershell/Language/Parser/Parser.Tests.ps1
vexx32: Done! 😄

src/System.Management.Automation/engine/parser/tokenizer.cs
vexx32: And again, good catch.

The multiplier can be *negative* in valid signed parsing cases because the hex parsing ignores negative sign, deferring that to the multiplier to be applied after parsing of the hex literal. I assume this is because the hex parsing methods of `.TryParse()` do not allow negative signs, which is probably because hex parsing permits entering signed values by utilising the sign bit.

iSazonov: It is not obvious behavior so please add the comment in code.

src/System.Management.Automation/engine/parser/tokenizer.cs
```diff
@@ -3355,6 +3365,15 @@ private static bool TryGetNumberValue(string strNum, bool hex, bool real, Number
                                 return true;
                             }
 
+                            result = null;
+                            return false;
+                        case NumberSuffixFlags.SignedByte:
+                            if (Math.Abs(multiplier) == 1L && sbyte.TryParse(strNum, style, NumberFormatInfo.InvariantInfo, out sbyte sb))
```
iSazonov: Below we have such code - please add the comment there too.

Why we use `1L`? I think we should use "universal" const `1` without suffix. Below too.

vexx32: Different comment there, but yes.

Wasn't sure if c# would care about type of numeric value when testing for equality, actually. :)

src/System.Management.Automation/engine/parser/tokenizer.cs
iSazonov: Iexpect `result = sb;`

vexx32: Impossible to completely ignore multiplier; if a hex literal is negative, multiplier will be `-1` by this point thanks to code referenced above. 

We must still either multiply by multiplier value or else account for it in a more complex expression, e.g., `result = multiplier < 0 ? -sb : sb`

I think simply multiplying by the already-verified multiplier value is the most straightforward method.

iSazonov: I agrre with this in the PR but I'd consider the possible optimization in follow PR.

vexx32: Yeah, this will be resolved more cleanly in the follow up PR. 😄

iSazonov (approved):

vexx32:
CodeFactor issues remaining are just in missing documentation in the enum; I can either add that to this PR or address it in follow-up PR with the rest of the changes. I would prefer to keep this one minimal, however. 😄

iSazonov:
@vexx32 Feel free to address CodeFactor issues.

vexx32:
Rebased for CIs && added `<summary>` headers for CodeFactor

iSazonov:
@daxian-dbw Could you please review the next step PR? After that we could continue and implement previously planned performance optimizations for the code in follow PR.

TravisEz13 (approved):

vexx32:
@TravisEz13 you fellas are free to pore over it one more time, but just in case you weren't aware / didn't recall, it has been discussed a good bit and reviewed by the committee in terms of the concept over in #7557 😄

TravisEz13:
@SteveL-MSFT This is marked as a breaking change.  The issue linked is not marked as reviewed.  Looking back at the issue, it looks like this portion of the design was reviewed.  Is that correct?

vexx32:
@TravisEz13  sorry for the confusion; I requested the committee further review the ensuing discussion regarding possibly supporting underscore literal syntax and native BigInteger values. This portion was reviewed, aye. 😃

iSazonov:
@daxian-dbw Could you please look the PR before we merge?

